PR #82, opened 2015-01-09
WifI wifi.xx.getnetmask & wifi.xx.getbroadcast methods added

rkoffer:
two methods for obtain netmask and broadcast address
will be used for some kind of discoverable process like:

local discoveryBroadcastConnection = net.createConnection(net.UDP)

tmr.alarm(1,10000,1, function() 
   myip = wifi.sta.getip()

   if myip ~= nil then
	 discoveryBroadcastConnection:connect(8888, wifi.sta.getbroadcast()) 
	 discoveryBroadcastConnection:send( string.format("%08x",node.chipid()))
   end
end)

please review
thnx
Rafal

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -191,6 +191,49 @@ static int wifi_getip( lua_State* L, uint8_t mode )
   }
 }
 
+// Lua: broadcast = wifi.xx.getbroadcast()
+static int wifi_getbroadcast( lua_State* L, uint8_t mode )
+{
+  struct ip_info pTempIp;
+  char temp[64];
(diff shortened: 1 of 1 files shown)


nodemcu:
very appreciate. thank you very much @rkoffer .

I need some advise. any one see this please give me some advise.
can we combine getnetmask and getip like this?
```lua
ip, netmask, gw = wifi.sta.getip()
```

or just merge this PR.

vowstar:
To save RAM, combine 3 in 1 is better. But the API may becomes confused.

vowstar:
I have saw the setip() API. In order to make the API symmetry, combine 3 in 1 as getip() is a good idea.

PR #89, opened 2015-01-10
Vastly simplified generating and flashing firmware.

sej7278:
Uses the new elf2image features of esptool.py to do away with the myriad of shell scripts and binaries.

Fixed permissions on esptool.py

To compile and flash we just call:

```
make
cd app/
make flash
```


nodemcu:
I will test this in windows

nodemcu:
in both win 7 eclipse toolchain and win xp vm.
```
# xtensa-lx106-elf-objcopy -O binary .output/eagle/debug/image/eagle.app.v6.out .output/eagle/debug/bin/eagle.app.v6.bin
../tools/esptool.py elf2image .output/eagle/debug/image/eagle.app.v6.out -o .output/eagle/debug/bin/
env: python2.7: No such file or directory
make: *** [.output/eagle/debug/bin/eagle.app.v6.bin] Error 127
```

if I remove python2.7 from the first line of esptool.py. no error, but
```
../tools/esptool.py elf2image .output/eagle/debug/image/eagle.app.v6.out -o .output/eagle/debug/bin/
```
will hang forever.


lubuntu vm works great. after install python-serial.

and esptool.py can burn firmware to devkit and runs to prompt.(which not under windows)
is esptool.py have some issue under windows?

sej7278:
no idea why it doesn't seem to work on windows, maybe its the paths - hidden directories or different types of directory slashes e.g. / vs \ ? maybe @themadinventor can comment?

esptool.py shouldn't really use ```#!/usr/bin/env python2.7``` it should just be ```#!/usr/bin/env python``` or ```#!/usr/bin/python```

also on windows you'll need to pass in ```--port COM3``` or something

themadinventor:
I was under the impression that an (version-wise) unqualified shebang line could invoke Python 3 (which esptool is currently incompatible with) on some systems that don't have Python 2.7 installed. The change origins from a pull request (themadinventor/esptool#7), so I though somebody was having problems on some configuration that I haven't tested, and the current form works on the platforms that I'm currently testing on (Ubuntu 14.04 and MacOS 10.9). However, reading up on the subject, it seems that skipping the version always should default to a Python 2 implementation, so I'm probably reverting to that.

sej7278:
@themadinventor yes, debian for example never includes the version string, but i believe it defaults to 2.x even if 3.x is installed - 3.x won't be the default anywhere for quite some time i expect.

but aside from that, do you have any idea why it seems to hang on windows?

nodemcu:
after some test, it IS the slashes.
linux-style path in makefile is ok. 
but the makefile pass the linux-style path to esptool.py on windows cause the issue.
and only the path after -o cause the issue.
"$<" before -o is ok.

sej7278:
does ```../bin/``` work on windows?

```diff
diff --git a/app/Makefile b/app/Makefile
index e498fad..9a5aa0c 100644
--- a/app/Makefile
+++ b/app/Makefile
@@ -159,10 +159,10 @@ sinclude $(PDIR)Makefile
 
 $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
        @mkdir -p $(BINODIR)
-       ../tools/esptool.py elf2image $< -o $(BINODIR)/
+       ../tools/esptool.py elf2image $< -o ../bin/
 
 flash:
-       ../tools/esptool.py write_flash 0x00000 $(BINODIR)/0x00000.bin 0x10000 $(BINODIR)/0x10000.bin
+       ../tools/esptool.py write_flash 0x00000 bin/0x00000.bin 0x10000 bin/0x10000.bin
 
 .PHONY: FORCE
 FORCE:
```

nodemcu:
../tools/esptool.py elf2image $< -o ..\\bin\\
works.

nodemcu:
../tools/esptool.py elf2image $< -o ..\\bin\\

nodemcu:
../tools/esptool.py elf2image $< -o ..\\\\bin\\\\

sej7278:
I've added PR #93 to allow users to specify the serial port the ESP is connected to, defaults to /dev/ttyUSB0 for ```make flash```

So do we need to do a platform check to fix the slashes?

hamishcunningham:
I confirm that the new setup works for me on (rebuilding on Ubuntu 14.04 and flashing from Raspbian on a Pi). Thanks folks!

PR #91, opened 2015-01-10
lua module loader

Hoksmur:
Hi. What do you think about loadable binary (or precompiled?) modules? You have a file system. It helps save memory and use only needed functional.
> yes, some elf loader should be implemented. and there is some out there. contiki and rtthread has elf loader.
> if we have a elf loader. modify lua loadlib, it can use
> ```lua
require("mod")
```
> to load a binary lib. 
> and it would be awesome if you can help.
> thanks.

### CODE REVIEW

**Comment by Hoksmur** [comment]
Ok. I will try. But I'm not a skilled in c programming. And I haven't esp8266's SDK yet.

**Comment by Hoksmur** [comment]
Ops.. Possible, we didn't understand each other. For example - there is a file (on embedded file system) *.lc . It doesn't use runtime memory. If you need this one - 
```lua 
requre('modname') 
```
It can be precompiled block for Lua VM or C-compiled. I don't know, this idea is good or bad. 
How to place file on FS - it is another question, for example - decided by *.lua script.

Hoksmur:
I fount it today. Maybe it will be usefull for you, maybe - you know it clean. Description Lua VM, how it works.
http://goo.gl/L8Bzl6

PR #93, opened 2015-01-10
Allow user to specify the serial port the ESP is attached to

sej7278:
For example on Windows it would be:

```make flash PORT=COM3```

On Linux you can use the default /dev/ttyUSB0:

```make flash```


nodemcu:
oh, add this already, thanks anyway.

sej7278:
lol, cool!

PR #94, opened 2015-01-10
adjustments for esptool.py.

kmpm:
Can now do make && make flash directly from root.
Also removed unnecessary stuff from app/Makefile.

It might need some adjustments for windows though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ ifeq ($(OS),Windows_NT)
 		CPP = xtensa-lx106-elf-cpp
 		OBJCOPY = xtensa-lx106-elf-objcopy
 	endif
-	FIRMWAREDIR = ..\\bin\\
+	ESPPORT = com1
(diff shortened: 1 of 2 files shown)


nodemcu:
thanks, currently make and make flash is needed within app/ for eclipse toolchain.
because the project root in eclipse is /app.
and for now, a \\ can not pass to esptool.py as a para.

kmpm:
ok
* make and make app works within app even with this change, as well as from repository root. I can not test with eclipse and or windows at the moment though.
* So you are saying that esptool.py needs backslash on windows but we can't pass it because of something? Is there something I can do about it?

vowstar:
Could we use eclipse project file in SDK root folder?

nodemcu:
@kmpm 
my mistake. a linux-style(./../../)  path can not pass to esptool.py as a para on windows.

PR #119, opened 2015-01-21
DHT22 module

javieryanez:


lua_modules/dht22/dht22.lua
```diff
@@ -0,0 +1,101 @@
...
+local checksum
+local checksumTest
+
+function M.read(pin)
+  humidity = 0
+  temperature = 0
+  checksum = 0
+
+  -- Use Markus Gritsch trick to speed up read/write on GPIO
+  gpio_read = gpio.read
```
dvv: this wastes to global. is it intended?

lua_modules/dht22/dht22.lua
```diff
@@ -0,0 +1,101 @@
...
+  gpio_write = gpio.write
```
dvv: a standard trick is to make everything "local", that is why _seemingly_ insane headers such as
```lua
local tonumber = tonumber
```

lua_modules/dht22/dht22.lua
javieryanez: It is a speed optimization. I do not understand at all, but this loop:

```javascript
while (gpio_read(pin) == 1 and c < 100) do c = c + 1 end
```
It is faster than this:

```javascript
while (gpio.read(pin) == 1 and c < 100) do c = c + 1 end
```

dvv: It's just due to `gpio.read()` requires: access to global table, access to its field, call vs. `gpio_read()` requires: access to global (because you forgot to prepend with `local`) function, call.

It will be (apriori) even faster if you cache `gpio.read` to a _local_ variable, as access to local variables is the fastest. So try to use `local gpio_read, gpio_write = gpio.read, gpio.write`. That is what I meant by this and the next  comment.

nodemcu:
indeed, and also the global table is in flash, even slower

lua_modules/dht22/dht22.lua
javieryanez: I'll try it.

lua_modules/dht22/dht22.lua
vowstar: Yes, I agree. May

lua_modules/dht22/dht22.lua
```diff
@@ -0,0 +1,101 @@
...
+  
+  bitStream = {}
```
dvv: Always try to use locals. If you omit `local` keyword it becomes global -- this is standard catch.
Also think in closure terms. Try to hide everything private. Say, a cosher module

```lua
local M
do
  -- private stuff
  local not4all = { ... }
  local f1 = function(...) ... end
  local f2 = function(x, y) not4all[x] = y end
  -- explicitly expose
  M = {
    f1 = f1,
  }
end
return M
```

This way you hide details, returning just what explicitly exposed.

NB: _All above is my imho_.

lua_modules/dht22/dht22.lua
dvv: Well, https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_modules/ds18b20/ds18b20.lua#L124 assigns a global called `read`. Please, confirm. This is hardly what you wanted, as it may clash with others.
Please, consider enclosing private stuff in `do ... end` and explicit exposure, as I've written below.

lua_modules/dht22/dht22.lua
javieryanez: I made a test:

```lua
function test()
  pin = 4
  gpio_read = gpio.read
  local gpio_read_local = gpio.read

  c = 0
  time1 = tmr.now() 
  while (gpio.read(pin) == 1 and c < 100) do c = c + 1 end
  time2 = tmr.now()
  print("Time: "..(time2 - time1))
  print(c)

  c = 0
  time1 = tmr.now() 
  while (gpio_read(pin) == 1 and c < 100) do c = c + 1 end
  time2 = tmr.now()
  print("Time: "..(time2 - time1))
  print(c)
  
  c = 0
  time1 = tmr.now() 
  while (gpio_read_local(pin) == 1 and c < 100) do c = c + 1 end
  time2 = tmr.now()
  print("Time: "..(time2 - time1))
  print(c)
end

test()
```

The result has been that:

>Time: 6170
100
Time: 2191
100
Time: 2048
100

As you you said, it's a little faster in a local variable. If c is local then the speed increase is great.

As the loop speed is critical, I will check if you need some adjustment before making the change.

PR #136, opened 2015-01-26
Support floating point LUA.

vowstar:
Now, define macro LUA_NUMBER_INTEGRAL in user_config.h can remove this feature.


baracudaz:
Just wondering what impact does this change have on available RAM and heap size?

vowstar:
Before use floating point, it is 24480 Byte available. After use floating point, it is 23360 Bytes left.

baracudaz:
OK that's around 1kB less. Thanks for your response.

PR #137, opened 2015-01-26
for reading internal vdd33 pin voltage.

alonewolfx2:
return value=mV
example usage and result
```
print(node.readvdd33())
3348
> 
```


vowstar:
https://github.com/nodemcu/nodemcu-firmware/commit/fc8fa2cc915aea3687b068eb0355bb84b2fc113d

PR #142, opened 2015-01-28
IR send module added

dvv:
Please, consider applying

### DIFF
--- a/lua_examples/irsend.lua
+++ b/lua_examples/irsend.lua
@@ -1,0 +1,74 @@
+------------------------------------------------------------------------------
+-- IR send module
+--
+-- LICENCE: http://opensource.org/licenses/MIT
+-- Vladimir Dronnikov <dronnikov@gmail.com>
+--
+-- Example:
+-- require("irsend").nec(4, 0x00ff00ff)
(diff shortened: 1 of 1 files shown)


baracudaz:
Hi dvv,
Do you have any further example how your module works? Do you drive IR LED directly from ESP pin or via transistor? I am not quite clear what the TSOP is? Thanks

dvv:
My ESP was not capable to drive IR led (that drove me almost crazy yesterday), so I had to come out to get some Darlington to keep going.
In TSOP mode ESP is meant to drive an external IR receiver as if ESP is a TSOP1838, say. In my particular case, I hacked a chinese rgb led stripe which is controlled via IR remote. I desoldered the TSOP, merged both grounds and connected driving pin to TSOP input pin (in fact I did that with atmega328p but since I succeeded I wanted to clone this to ESP).
So now the stripe can be controlled via WiFi which is far better )

dvv:
To illustrate (different scales, sorry):
Vanilla mode: `nec(4, 0x00ff00ff)`
image
TSOP mode: `nec(4, 0x00ff00ff)`
image

PR #143, opened 2015-01-28
new sleep set options added (from 0.9.5 sdk)

alonewolfx2:
usage:
-- no wakeup until rst pin to low 
```
node.dsleep(0) 
```
-- 4 mean disable RF after deepsleep or restart. (esp8266 consuming ~13mA in this mode.)
-- 1 mean enable RF after deepsleep or restart.
```
node.dsleepsetoption(4)
```

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -26,12 +26,24 @@ static int node_deepsleep( lua_State* L )
 {
   s32 us;
   us = luaL_checkinteger( L, 1 );
-  if ( us <= 0 )
+ // if ( us <= 0 )
+ if ( us < 0 )
     return luaL_error( L, "wrong arg range" );
   system_deep_sleep( us );
   return 0;  
 }
 
+// Lua: dsleep_set_options
+static int node_deepsleep_setoption( lua_State* L )
+{
+  s32 option;
+  option = luaL_checkinteger( L, 1 );
+  if ( option < 0 || option > 4)
+    return luaL_error( L, "wrong arg range" );
+  else
+	  deep_sleep_set_option( option );
+  return 0;
+}
 // Lua: info()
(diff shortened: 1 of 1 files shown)


app/modules/node.c
```diff
@@ -295,6 +307,7 @@ const LUA_REG_TYPE node_map[] =
   { LSTRKEY( "input" ), LFUNCVAL( node_input ) },
   { LSTRKEY( "output" ), LFUNCVAL( node_output ) },
   { LSTRKEY( "readvdd33" ), LFUNCVAL( node_readvdd33) },
+  { LSTRKEY( "dsleepsetoption" ), LFUNCVAL( node_deepsleep_setoption) },
```
vowstar: Maybe can use dsleep(us, option) one time sovle every thing. Default option is 0. It is more easy and save ram.

alonewolfx2:
you are right but deepsleepsetoption command working with deepsleep and restart function. mean if i can use node.dsleepsetoption(4) and just node.restart(), its starting with no RF. thats why I created separately. by the way dsleepsetoption automatically changing with default value every restart on new at example on 0.9.5 but its not changing automatically in nodemcu. i dont know why

vowstar:
Yes, it also confused me, the SDK is very strange.
I have combine it, and node.dsleep() can have these usage:
```lua
--do nothing
node.dsleep()
--sleep μs
node.dsleep(1000000)
--set sleep option, then sleep μs
node.dsleep(1000000, 4)
--set sleep option only
node.dsleep(nil,4)
```

alonewolfx2:
hmm good. also it have some bug in deepsleep. sometimes it didnt wake up correctly for long time use. i didnt figure out yet.

Issue #155, opened 2015-01-30
mqtt LWT bad QoS and retain


Suxsem:
Hi. I found a bug relative to lwt feature of mqtt module:

this code
    m:lwt("offline", "offline", 2, 0)
SHOULD send a message with QoS = 2 and retain = false

instead... the mqtt broker receive it as QoS = 0 and retain = true

no problem with normal publishing...only with lwt

Suxsem:
Sorry, the bug is still there. The lwt is published with Retain bit on also if I set it to 0
e.g.
`m:lwt("status", "offline", 2, 0)`
is sent as
`m:lwt("status", "offline", 2, 1)`

PR #158, opened 2015-01-31
fixes nodemcu/nodemcu-firmware#155 lwt qos, retain

tuanpmt:
change header support MQTT 3.1.1 valid by https://eclipse.org/paho/clients/testing/
fixes error if username or password = null,


nzmichaelh:
Note that this makes nodemcu incompatible with Mosquitto v0.15 which ships with Debian.  See http://mosquitto.org/2013/01/mosquitto-debian-repository/ for updated packages which should fix the problem,

PR #170, opened 2015-02-03
Cle69 patch 1

cle69:
Add  Bit rate 1200, 2400,4800 to support old Serial link


cle69:
Add Bit rate 1200, 2400,4800 to support old Serial link

cle69:
Add Bit rate 1200, 2400,4800 to support old Serial link

cle69:
Add Bit rate 1200, 2400,4800 to support old Serial link

PR #182, opened 2015-02-06
Use python2 instead of python(2/3)

lukas2511:
On modern systems "python" refers to python3, and the script won't work, so I suggest changing it to python2 with this patch.

This patch shouldn't break compatibility with older systems.


nodemcu:
has anybody test this PR in windows Eclipse?
please review this commit: 2709258c44b4fa2397df736b12b9c2ef86dc4d89
change from python2.7 to python

Issue #176, opened 2015-02-04
Please push PR to branch "dev" instead of master.


nodemcu:
Thank you everyone for contributing to this project.
I have made branch "dev" up-to-date with master branch.
new features or some major fixes can first push to this "dev" branch.
note that pre_build bin in this "dev" branch is never up-to-date.
for ones want to test new features, please pull "dev" branch and compile.

Suxsem:
thanks! :) please don't forget to build new pre_build! :)

vowstar:
Ok, I agree with it =)

funshine:
I would like to open this issue again.
there's many PRs in master ahead of dev branch. we are now going to sort things out.

nodemcu:
I would like to open this issue again.
there's many PRs in master ahead of dev branch. we are now going to sort things out.

karrots:
#575 & #675 were cherry picked over to 'dev'.

nodemcu:
great, once all done, we will add this to guideline and can be closed.

TerryE:
I'll sort out any residual manual merges tomorrow.

marcelstoer:
@karrots Thanks for checking.
@nodemcu It already _is_ in the guidelines, see https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md.

TerryE:
We now have this as a contribution guideline which is being actively enforced by the committers so we can close this as an issue.

PR #187, opened 2015-02-06
Disabled automatic detection of larger flash sizes

Manawyrm:
The autodetection is broken and causes a number of strange reliability issues, see here: 
http://www.esp8266.com/viewtopic.php?f=18&t=1418&p=9082#p9082
and here: 
http://www.esp8266.com/viewtopic.php?f=18&t=1180

and numerous other threads in the esp8266 forum. 

Of course this is not a fix, but it will work as a workaround to fix the crashes on short term.

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -9,11 +9,11 @@
 #define NODE_VERSION	"NodeMCU 0.9.5"
 #define BUILD_DATE	    "build 20150127"
 
-// #define FLASH_512K
+#define FLASH_512K
 // #define FLASH_1M
 // #define FLASH_2M
 // #define FLASH_4M
-#define FLASH_AUTOSIZE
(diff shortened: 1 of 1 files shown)


vowstar:
Sorry for that we can only accept PR to dev instead master.
https://github.com/nodemcu/nodemcu-firmware/issues/176
This issue we will consider, because many people use the 4MByte or other flash larger than 512kBytes.

MarsTechHAN:
How about the first time auto detection, And save data in somewhere in Flash?

vowstar:
I guess I have find the solution, waiting a PR

Manawyrm:
Sounds good. Shall I then close this PR?

vowstar:
Keep this as an issue, waiting us final fix.

vowstar:
I think this issue is fixed. Please test.

PR #192, opened 2015-02-08
Add support for WS2812 based LED strips

dereulenspiegel:
This pull request adds the command ws2812.write(string) to control WS2812 based LED strips. This module can also be disabled in user_conf.h.


clixx-io:
is there any example code for this yet ?

dereulenspiegel:
I updated the README.md with some examples. You can find them at the end.

PR #196, opened 2015-02-09
Added the hold and unhold methods to tcp socket  

Point-less:
While implementing a simple echo server (https://gist.github.com/Point-less/27ec0a2284f2d82d6a9c) i found out that the data was been lost because of espconn_sent new implementation ignores calls been done before the data can be sent. The only way of avoiding this is having some kind of callback which tells us when we can send new data. While I was trying to implement the echo server using that callback I came to realize that there is was no way of stopping data from coming in and be ACKed or throttling the TCP connection, so I added the support for hold/unhold in the tcp socket, which lets us stop the ack of incomming data by dropping the payload of the tcp packets (so that the protocol "looses" some packets and adjust the speed according to our needs).

This is a really simple approach and can be enhanced (buffering for a small time the payload for example, so when we call unhold we can have data right on time). The espresiff sdk has the same two functions in it's api, so i used the same interface.

With this feature i can have a working echo server (https://gist.github.com/Point-less/51eafb0e2d48021de03e).

### DIFF
--- a/app/include/lwip/app/espconn.h
+++ b/app/include/lwip/app/espconn.h
@@ -394,5 +394,21 @@ extern sint8 espconn_igmp_join(ip_addr_t *host_ip, ip_addr_t *multicast_ip);
 *******************************************************************************/
 extern sint8 espconn_igmp_leave(ip_addr_t *host_ip, ip_addr_t *multicast_ip);
 
+/******************************************************************************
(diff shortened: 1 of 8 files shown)


Point-less:
Oh shoot, i got an extra commit on the pull request, sorry, should i do it again?

PR #197, opened 2015-02-09
Revert "Use python2 instead of python(2/3)"

nodemcu:
Reverts nodemcu/nodemcu-firmware#182
reported that not work in mac.


ignaworn:
I have experienced the same issue on Windows and MiniGW, had to revert from python2 to python to successfully compile nodemcu.

Issue #206, opened 2015-02-12
Compilation fails


slaff:
As of today, 12th of February, 2015 the compilation of the master branch fails with the following error message:

make[2]: Entering directory `/home/vagrant/dev/nodemcu-firmware/app/libc'
....
xtensa-lx106-elf-gcc -Os -Os -ffunction-sections -fno-jump-tables -g -O2 -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals  -D__ets__ -DICACHE_FLASH -DLWIP_OPEN_SRC -DPBUF_RSV_FOR_WLAN -DEBUF_LWIP   -I include -I ./ -I ../wofs -I ../include -I ./ -I ../../include -I ../../include/eagle  -o .output/eagle/debug/obj/c_stdio.o -c c_stdio.c
c_stdio.c:284:0: error: "__P" redefined [-Werror]
 #define __P(x) x
 ^
In file included from /opt/Espressif/xtensa-lx106-elf/xtensa-lx106-elf/include/string.h:12:0,
                 from ../../include/osapi.h:8,
                 from c_stdio.h:7,
                 from c_stdio.c:1:
/opt/Espressif/xtensa-lx106-elf/xtensa-lx106-elf/include/sys/cdefs.h:53:0: note: this is the location of the previous definition
 #define __P(args) args
 ^
cc1: all warnings being treated as errors
make[2]: *** [.output/eagle/debug/obj/c_stdio.o] Error 1

slaff:
Great, thanks!

PR #207, opened 2015-02-12
Update c_stdio.c

eyaleb:
The macro '__P' is redefined here which fails the build.


eyaleb:
building ('make' in nodemcu-firmware) fails as the "redefined" warning is treated as an error due to compile options. One should not use the '__P' macro name here.

nodemcu:
hi, this has been fixed in dev branch, will merge to master soon

nodemcu:
ref issue #206

PR #212, opened 2015-02-13
Update irsend.lua

ksankark:
The for loop was reverse because of that when i transfer 0xfff000ff i have received 0xff000fff. This update works perfect (I have tested this by receiving through Arduino IRreceiver).

### DIFF
--- a/lua_examples/irsend.lua
+++ b/lua_examples/irsend.lua
@@ -57,7 +57,7 @@ do
     pulse(pin, NEC_HDR_MARK)
     waitus(NEC_HDR_SPACE)
     -- sequence, lsb first
-    for i = 0, 31 do
+    for i = 31, 0, -1 do
       pulse(pin, NEC_BIT_MARK)
       waitus(isset(code, i) and NEC_ONE_SPACE or NEC_ZERO_SPACE)
     end


dvv:
Hey, such merges do break things )

nodemcu:
:( sorry, didn't look into this carefully..
should I revert it? 
and more test?

dvv:
NP. Let's first wait for @ksankark 's reaction.

dvv:
Well, https://github.com/shirriff/Arduino-IRremote/blob/master/IRremote.cpp#L82 shows that arduino library sends nec codes (and others as well) most significant bit first, while collecting them in the opposite way as https://github.com/shirriff/Arduino-IRremote/blob/master/IRremote.cpp#L524 reads.

This is inconsistent, I believe. So I'd rather keep my version. Probably accompanied with a reversed one.

What do you guys think?

ksankark:
“But since ESP8266 is little-ending arch (right?).” Me too have a doubt here. If something is mismatching it could be only here.

According to this document http://0x04.net/~mwk/doc/xtensa.pdf in page 41 under section 2.1 it is given like this “However, this notation convention is independent of how an Xtensa
processor actually numbers bits, because a given processor can be configured for either
little- or big-endian byte and bit ordering”.

Now i have tested this setup with my TV, there also it works perfect after this modification.
Please correct me if i am wrong.

Issue #225, opened 2015-02-15
Problems with DHT22 library and 2015-02-13 binary (introduced in 2015-01-27)


robcazzaro:
[Solved: please see comment below]

I was using 2015-01-26 binary (prebuilt binary, didn't recompile anything) and started using the DHT22 library, which works very reliably. 

Updated to 2015-02-13, and updated the library to use floating point. But I started noticing a high rate of checksum errors (roughly 30-40% of the reads result in a checksum error). Went back to 2015-01-26 and, with the same library (the new one, compatible with both floating point and integers) and the exact same code, I can read the DHT22 very reliably, not a single checksum error in 50 tries

I tried flashing 2015-01-27, and the problem reproduces there as well. So 2015-01-26 is the last reliable build for the DHT22 library 

It looks as if the timing of the loops reading the DHT22 signal is slightly off, somehow, or a float math error in the processing of the packets from the DHT22. I know it's not a float math error in the checksum itself because when the checksum is off, either the temperature or humidity reading (usually humidity) is off significantly as well (i.e. while reading 45% humidity with the right checksum, might read 20% humidity when the checksum is off). I did put extra print() statements in the DHT22 library to see the values being read, btw, and not relying on the values returned by the function (which sets humidity to nil when the checksum is off)

robcazzaro:
I spent a bit more time understanding the DHT22 library, and I figured out how to solve the issue

This code reads the data, measuring the length of the pulse.

  for j = 1, 40, 1 do
    while (gpio_read(pin) == 1 and bitlength < 10 ) do
      bitlength = bitlength + 1
    end
    bitStream[j] = bitlength
    bitlength = 0
    -- bus will always let up eventually, don't bother with timeout
    while (gpio_read(pin) == 0) do end
  end

Short pulse=low, long pulse=high (from the DHT22 dayasheet)

In 2015-01-26, the values of bitStream[j] are between 0 and 2 for low (usually 1 and 2, 0 is only when the first bit is low), and 6 and 7 for high. For low, you see many 2, never a 5 for high

In newer builds (starting from 2015-01-27 up to and including 2015-02-13), the values of bitStream[j]  are 0-2 for low (here as well usually 1 and 2), but 4 and 5 for high. Very few 2 for low, and never a 6 for high. Very few 4 as well (that's why it rarely failed), most of the high seems 5, pointing to a slight timing difference in the loop

[...]

PR #227, opened 2015-02-16
Update dht22.lua to address issue #225

robcazzaro:
Addresses issue #225, slower loop timing due to the use of floats. Should be safe to use with integer-only builds, since on integer-only builds I could never see "3" as a valid value for either low or high

### DIFF
--- a/lua_modules/dht22/dht22.lua
+++ b/lua_modules/dht22/dht22.lua
@@ -62,17 +62,17 @@ function M.read(pin)
 
   --DHT data acquired, process.
   for i = 1, 16, 1 do
-    if (bitStream[i] > 4) then
+    if (bitStream[i] > 3) then
       humidity = humidity + 2 ^ (16 - i)
     end
   end
   for i = 1, 16, 1 do
-    if (bitStream[i + 16] > 4) then
+    if (bitStream[i + 16] > 3) then
       temperature = temperature + 2 ^ (16 - i)
     end
   end
   for i = 1, 8, 1 do
-    if (bitStream[i + 32] > 4) then
+    if (bitStream[i + 32] > 3) then
       checksum = checksum + 2 ^ (8 - i)
     end
   end


vowstar:
May be integer-only builds is better than float builds.

dvv:
I believe one could use the following to accomodate for whether the build has floats or no:
```lua
if 1/2 ~= 0 then
  print("Floating. Needa rely on another timings")
else
  print("Integers only. Behave just like before")
end
```

marcoskirsch:
I used

if math ~= null then
   ....

Seemed a little bit (just a little bit) cleaner.

GeoNomad:
or  

if math then

dvv:
Wonder if we could have driver written in C, exposed as dht.read (pin,
dht_type)

morlac:
+1

2015-02-26 5:52 GMT+01:00 Vladimir Dronnikov <notifications@github.com>:

> Wonder if we could have driver written in C, exposed as dht.read (pin,
> dht_type)

robcazzaro:
It would be a lot better, agreed. As things are, the DHT22 library uses a lot of very precious memory. In my case, I had to optimize a few things to fit everything in memory. Removing the long DHT22 function would be great.

Alas, while fixing the problem in LUA was within my abilities, porting one of the many DHT22 libraries to nodeMCU in C++ is not

dvv:
As a workaround, you might want to try my https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_examples/yet-another-dht22.lua -- it uses only a preallocated table of 5 octets and should not be touched by floating point timing issue.

robcazzaro:
I'll give yours a try, thanks, and see if it uses less memory. But please note that after my proposed change, the LUA function now works both for float match and integer math, so that's not an issue anymore. Saving memory for other LUA code would be nice

Implementing the library in C++, would save a lot of LUA memory when someone uses a DHTxx

marcoskirsch:
I think we need to be careful about adding drivers for specific peripherals to the firmware. It saves memory for those who want to use them. But it wastes memory for everyone else. Feature creep is a very bad thing on such a limited device. 

If it ends up being added, it should probably be disabled by default (like imho floating point support) and it should be simple and clear how to enable it. Maybe a confirmation header file where a #define is changed for each optional feature. 

I'm hesitant to ask this since it could hijack the thread, but you mention C++. As far as I know nodemcu-firmware is C. Right? Do the tools even support C++?

robcazzaro:
Agree on being careful adding rarely used libraries.

What is needed, is a way to load C libraries dynamically, ideally a wrapper that allows to load a binary library and have dofile() or (require() also be able to load those, similarly to how today require() recognizes *.lc files in bytecode. 

Being able to dynamically load C libraries would allow the creation of any C library without weighting down everyone else not using it

I tried using bytecode compressed files, but the savings are really not that much 

(and I misspoke: my fingers type C++ automatically these days :-), I meant C)

PR #250, opened 2015-02-27
ws2812: change colors to accept RGB values

flokli:
Nobody would expect GRB input, and converting from RGB to GRB via Lua is too slow,
so convert it in place before sending to the hardware.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -36,14 +36,18 @@ static void ICACHE_FLASH_ATTR send_ws_1(uint8_t gpio) {
 }
 
 // Lua: ws2812.write(pin, "string")
-// Byte triples in the string are interpreted as G R B values.
-// ws2812.write(4, string.char(0, 255, 0)) uses GPIO2 and sets the first LED red.
+// Byte triples in the string are interpreted as R G B values
+// and sent to the hardware as G R B.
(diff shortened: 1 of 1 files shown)


markusgritsch:
Please delete your pull request, it breakes the WS2812 timing.

It took quite some time and experimentation to get the timing right, for all possible combinations of the values sent and used compiler optimization settings. Every change to this code has to be checked carfully with a logic analyzer or scope.

I would not mind changing the byte order to RGB, but the way you've done it breakes the timing. See here: http://www.esp8266.com/viewtopic.php?f=21&t=1143&p=11024#p11024

flokli:
@markusgritsch: You're right. I thought the led was broken somehow, and for this reason green.
I will try to fix the timing before I reopen the pull-request.

markusgritsch:
Thanks.  And please do really verify your changes with a simple scope or a cheap LA and make sure, the timing does not change.  Merely "LEDs seem to work for me" is not enough, due to different LEDs being more or less sensitive to border-line timings :)

And keep in mind that due to compiler optimizations it might make a difference if you send a 0 or a 1 as the first value, and also if it is the first time the code is called after a reset, or if you  just check subsequent calls (caching).

Be careful and rather do not make a pull request if you don't have the tools to check the timings.  Things get too easily broken otherwise.

PR #258, opened 2015-03-02
I2C Driver Module For the MCP23008 I/O Expander

AllAboutEE:
This is a driver for an I/O expander which lets you control 8 additional GPIOs via I2C. 

This is a video of the LEDs example if you'd like to see how it works.
https://www.youtube.com/watch?v=KphAJMZZed0


dvv:
FYI: https://github.com/eadf/esp8266_mcp23017_example/blob/master/driver/mcp23017/mcp23017.c written in C hence can guarantee timings

PR #260, opened 2015-03-03
Changed the ws2812 byte order from GRB to RGB

markusgritsch:
There seems to be interest in using the more intuitive RGB byte order to specify the LED color. Also fixed the documentation.


clixx-io:
Hello,

is there a trick to using this?

> =ws2812.write(3,string.char(0,255,0))


stdin:1: attempt to index global 'ws2812' (a nil value)

>

markusgritsch:
To reflect the change in the byte order, write() has been renamed to writergb().  It's documented at the very bottom of this page: https://github.com/nodemcu/nodemcu-firmware/tree/dev

clixx-io:
That's not working either:

> =ws2812.writergb(3,string.char(0,255,0))

stdin:1: attempt to index global 'ws2812' (a nil value)

>

markusgritsch:
The module ws2812 is disabled in the master branch. It is only enabled in the current dev branch. Check your app/include/user_config.h file.

clixx-io:
Ok - fair enough.

I'm just flashing the .bin file from the repo. I'm not building it.

clixx-io:
Just wondering,

does somebody have a quick link to the URL that tells me how I can build
the firmware so I can get the WS2812 functionality ?

Just that I have all the hardware rigged up and ready to go.

Also, I can't see in the API function where you set the number of LED that
you want to address.

Thanks in advance

PR #267, opened 2015-03-06
Remove binaries from repository

GeorgeHahn:
Replaced by [CI builds](https://github.com/nodemcu/nodemcu-firmware/releases)


nodemcu:
the pre_build/latest/nodemcu_latest.bin is the firmware lately build,
as asked by many people, because they have a ref link to it.
any way to preserve a hard link with a "static" name in CI builds system?
all can we re-direct link pre_build/latest/nodemcu_latest.bin to some where else?

GeorgeHahn:
Linking to https://github.com/nodemcu/nodemcu-firmware/releases/latest will always give the latest release as built by the CI, is that what you want?

nodemcu:
actually they are using this link:
https://github.com/nodemcu/nodemcu-firmware/blob/master/pre_build/latest/nodemcu_latest.bin
can this link be re-directed? say click on this and re-direct to https://github.com/nodemcu/nodemcu-firmware/releases/latest

GeorgeHahn:
Ahh, gotcha. I'm not aware of any way to do that. This PR would just break that link and give 404 errors to anyone who tried it.

nodemcu:
the binary will be there for some days, until most of us get used to the new download location.
thank you any way.

PR #272, opened 2015-03-07
Add document for file module and a shell script for python2 detection on Archlinux.

jiegec:
I will continue making other documents for other modules.


Makefile
```diff
@@ -75,7 +75,7 @@ else
     endif
 endif
 #############################################################
-ESPTOOL ?= ../tools/esptool.py
+ESPTOOL ?= ../tools/esptool.sh
```
GeorgeHahn: This looks like it may fail on Windows; maybe it belongs in the `if` statements above?

jiegec: I think if you use this, you should be in cygwin/msys environment.

nodemcu:
on windows means that something like eclipse toolchain, seems it does not use cygwin.
is it possible to modify esptool.py that it can be used both with python2 and python3?

jiegec:
emm.. yes, but the license of esptool.py is GPL, so it need to be fixed by the original author, but he seems no more working on it..

jiegec:
If i can edit it, according to the https://www.python.org/dev/peps/pep-0394/ which is not well implemented, i would change the hashbang to #!/bin/usr/env python2. Mac and ubuntu is tested.

nodemcu:
I mean, modify esptool.py to compatible with python3.
formerly it IS #!/bin/usr/env python2, but it can not used in some os.
then it changed to python. ref #197, #182

jiegec:
Ok. I am working on it.

PR #289, opened 2015-03-17
Added node.setcpufreq(mhz) function.

markusgritsch:
This function allows setting the CPU frequency to node.CPU80MHZ or node.CPU160MHZ.

Forum: http://www.esp8266.com/viewtopic.php?f=21&t=1369


eliabieri:
Nice, waited a long time for this.

-- 

Elia Bieri
whateveryouwant@eliabieri.com

nodemcu:
tested and wouldn't smoke 8266 right? :)

markusgritsch:
No worries. It's safe :)

markusgritsch:
While we're at it, has anyone an idea why the linker gives me an out of memory error when compiling the dev branch in float mode?  http://www.esp8266.com/viewtopic.php?f=23&t=2030
I mean, it seems that travis-ci does not have this problem.

PR #329, opened 2015-04-03
Add support to DHT11 and add some comments.

MarsTechHAN:


MarsTechHAN:
Add support to DHT11 and add some comments.

PR #332, opened 2015-04-05
Fix DHT11 10times BUG, add auto-select feature, fix program bug in md

MarsTechHAN:


MarsTechHAN:
But it use several lines of extra codes... So maybe not a good idea? I thought

MarsTechHAN:
Sorry for mistake, re-process the pull request to dev branch

chadouming:
You were right, this does use a lot more memory. I will try to find a way to make it with less line of codes

MarsTechHAN:
I think, it may not use that much of mem...? If you think it really is, how about just pre-calculate all the bytes? and if it is DHT22, combine those bytes later?

chadouming:
Why not calculate all the byte dht22 style and replace empty one by zeros ?

PR #342, opened 2015-04-08
Update tmr.c

klukonin:
I don't know C so good, but looks like we can decrease tmr.delay latency


klukonin:
New Pull is here
https://github.com/nodemcu/nodemcu-firmware/pull/344

PR #344, opened 2015-04-09
tmr.delay latency decrease

klukonin:
I'm adding watchdog clear to avoid sustem reboot.

Looks like it should work correct now.

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -71,16 +71,8 @@ static int tmr_delay( lua_State* L )
   us = luaL_checkinteger( L, 1 );
   if ( us <= 0 )
     return luaL_error( L, "wrong arg range" );
-  unsigned sec = (unsigned)us / 1000000;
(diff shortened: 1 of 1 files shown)


nodemcu:
if the delay us > 3000000, about 2~3 seconds.
the watch dog will fire and reset the chip.
please verify.

klukonin:
Confirmed.

As I think, it will lbe better to add some instructions to the API page, rather than add a check to the delay function. So If someone need to make a delay about 1-3 seconds, let it be 1000 us delay, but 1000-3000 times. Unusual scenario. IMHO.

Second way is to devide tmr.delay into tmr.delay.s, tmr.delay.ms, and tmr.delay.us for second, milisecond and microsecond values. How do you think?

klukonin:
Sorry, I'll add future commits to the dev branch.

klukonin:
https://github.com/nodemcu/nodemcu-firmware/pull/346

PR #346, opened 2015-04-09
tmr.delay devided into tmr.delay.us, tmr.delay.ms and tmr.delay.s

klukonin:
Just a simple ugly code to add different values for tmr.delay and decrease latency a little bit.

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -71,15 +71,11 @@ static int tmr_delay( lua_State* L )
   us = luaL_checkinteger( L, 1 );
   if ( us <= 0 )
     return luaL_error( L, "wrong arg range" );
-  unsigned sec = (unsigned)us / 1000000;
-  unsigned remain = (unsigned)us % 1000000;
-  int i = 0;
-  for(i=0;i<sec;i++){
-    os_delay_us( 1000000 );
-    WRITE_PERI_REG(0x60000914, 0x73);
-  }
-  if(remain>0)
-    os_delay_us( remain );
+
+  os_delay_us( us );
+  
+  WRITE_PERI_REG(0x60000914, 0x73); //Watchdog clear to aviod system reboot
+
   return 0;  
 }
 
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -65,16 +65,53 @@ typedef void (*alarm_timer_callback)(void *arg);
 static alarm_timer_callback alarm_timer_cb[NUM_TMR] = {alarm_timer_cb0,alarm_timer_cb1,alarm_timer_cb2,alarm_timer_cb3,alarm_timer_cb4,alarm_timer_cb5,alarm_timer_cb6};
 
 // Lua: delay( us )
-static int tmr_delay( lua_State* L )
+static int tmr_delay_us( lua_State* L )
 {
   s32 us;
   us = luaL_checkinteger( L, 1 );
   if ( us <= 0 )
     return luaL_error( L, "wrong arg range" );
 
-  os_delay_us( us );
-  
-  WRITE_PERI_REG(0x60000914, 0x73); //Watchdog clear to aviod system reboot
+     os_delay_us( us );
+     WRITE_PERI_REG(0x60000914, 0x73); //Watchdog clear to aviod system reboot
+
+  return 0;  
+}
+
+// Lua: delay( ms )
+static int tmr_delay_ms( lua_State* L )
+{
+  s32 ms;
+  ms = luaL_checkinteger( L, 1 );
+  if ( ms <= 0 )
+    return luaL_error( L, "wrong arg range" );
+    
+  int i = 0;
+  for(i=0;i<1000;i++) // 1ms == 1000us. Simple stupid and should work.
+                      //May be later this will need some calibration or refactoring.
+    {
+     os_delay_us( ms );
+     WRITE_PERI_REG(0x60000914, 0x73); //Watchdog clear to aviod system reboot
+    }
+
+  return 0;  
+}
+
+// Lua: delay( s )
+static int tmr_delay_s( lua_State* L )
+{
+  s32 ms;
+  ms = luaL_checkinteger( L, 1 );
+  if ( ms <= 0 )
+    return luaL_error( L, "wrong arg range" );
+
+    int i = 0;
+     for(i=0;i<1000000;i++) // 1s == 1000000us. Simple stupid and should work.
+                            //May be later this will need some calibration or refactoring.
+       {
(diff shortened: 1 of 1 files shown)


app/modules/tmr.c
```diff
@@ -65,21 +65,54 @@ typedef void (*alarm_timer_callback)(void *arg);
...
+    return luaL_error( L, "wrong arg range" );
+    
   int i = 0;
-  for(i=0;i<sec;i++){
-    os_delay_us( 1000000 );
-    WRITE_PERI_REG(0x60000914, 0x73);
-  }
-  if(remain>0)
-    os_delay_us( remain );
+  for(i=0;i<1000;i++) // 1ms == 1000us. Simple stupid and should work.
```
ronyb29: If I ask it to sleep for 30ms, it will try to sleep for 30us, 1000 times, which is 30 ms. It would be better to do the opposite; sleep for 1000us 30 times, in order to skip a couple of evaluations of the for loop.

In fact, a better solution would be to calculate (with a simple division) exactly how many times you can sleep for a full second (which seems to be a sensible interval to reset the watchdog) and then sleep for the remainder. In the example i gave you, that would be 0 full seconds and then 30000us.

In the case of implementing the second solution, it would be exactly as implemented in the original `tmr_delay`.

app/modules/tmr.c
```diff
@@ -180,7 +213,9 @@ static int tmr_time( lua_State* L )
 #include "lrodefs.h"
 const LUA_REG_TYPE tmr_map[] = 
 {
-  { LSTRKEY( "delay" ), LFUNCVAL( tmr_delay ) },
+  { LSTRKEY( "delay.us" ), LFUNCVAL( tmr_delay_us ) },
+  { LSTRKEY( "delay.ms" ), LFUNCVAL( tmr_delay_ms ) },
+  { LSTRKEY( "delay.s" ), LFUNCVAL( tmr_delay_s ) },
```
ronyb29: This would break the current api for tmr.delay.

ronyb29:
Are you trying to get a slightly more precise API for very small amount of micro second delays?

Maybe it would be better to add a single method to do that (just like your `tmr_delay_us`), and then test it to see what kind of latency you get in each one (I'd love to help you write the script for that).

app/modules/tmr.c
klukonin: The main idea was to remove the original code with a lot of math. Just because of a huge latency for us operations. For example, if we are talking about ir protocols. So if I want to make a delay about 5us, I don't need any additional calculations. But I underdtand your idea about ms, it's good. I agree.

app/modules/tmr.c
klukonin: I know. Is it so bad?

klukonin:
Yes, all about micro second delays. I just want to make IR communication module with stable baud rate.

Okay, let's make some testing =)
I'll add a new method without  breaking the current API
Please, can you consult me a little bit about compiling this firmware?

nodemcu:
Hi, if you want to a faster micro second delays
maybe put code below at the beginning of the function:
```c
if(us<1000000){
    os_delay_us( us);
    WRITE_PERI_REG(0x60000914, 0x73);
    return;
}
```

and if you want a precise us_delay to do GPIO toggle.
I suggest the new api gpio.serout(), added recently.
sorry not doc this api yet.

nodemcu:
https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/gpio.c#L154
```lua
gpio.serout(1,1,{8,18},8)	-- serial 30% pwm 38k, lasts 8 cycles
```
I hope the example can demonstrate the usage clearly?

```c
// Lua: serout( pin, firstLevel, delay_table, [repeatNum] )
// -- serout( pin, firstLevel, delay_table, [repeatNum] )
// gpio.mode(1,gpio.OUTPUT,gpio.PULLUP)
// gpio.serout(1,1,{30,30,60,60,30,30})  -- serial one byte, b10110010
// gpio.serout(1,1,{30,70},8)  -- serial 30% pwm 10k, lasts 8 cycles
// gpio.serout(1,1,{3,7},8)  -- serial 30% pwm 100k, lasts 8 cycles
// gpio.serout(1,1,{0,0},8)  -- serial 50% pwm as fast as possible, lasts 8 cycles
// gpio.mode(1,gpio.OUTPUT,gpio.PULLUP)
// gpio.serout(1,0,{20,10,10,20,10,10,10,100}) -- sim uart one byte 0x5A at about 100kbps
// gpio.serout(1,1,{8,18},8) -- serial 30% pwm 38k, lasts 8 cycles
static int lgpio_serout( lua_State* L )
{
  unsigned level;
  unsigned pin;
  unsigned table_len = 0;
  unsigned repeat = 0;
  int delay_table[DELAY_TABLE_MAX_LEN];
  
  pin = luaL_checkinteger( L, 1 );
  MOD_CHECK_ID( gpio, pin );
  level = luaL_checkinteger( L, 2 );
  if ( level!=HIGH && level!=LOW )
    return luaL_error( L, "wrong arg type" );
  if( lua_istable( L, 3 ) )
  {
    table_len = lua_objlen( L, 3 );
    if (table_len <= 0 || table_len>DELAY_TABLE_MAX_LEN)
      return luaL_error( L, "wrong arg range" );
    int i;
    for( i = 0; i < table_len; i ++ )
    {
      lua_rawgeti( L, 3, i + 1 );
      delay_table[i] = ( int )luaL_checkinteger( L, -1 );
      lua_pop( L, 1 );
      if( delay_table[i] < 0 || delay_table[i] > 1000000 )    // can not delay more than 1000000 us
        return luaL_error( L, "delay must < 1000000 us" );
    }
  } else {
    return luaL_error( L, "wrong arg range" );
  } 

  if(lua_isnumber(L, 4))
    repeat = lua_tointeger( L, 4 );
  if( repeat < 0 || repeat > DELAY_TABLE_MAX_LEN )
    return luaL_error( L, "delay must < 256" );

  if(repeat==0)
    repeat = 1;
  int j;
  bool skip_loop = true;
  do
  {
    if(skip_loop){    // skip the first loop.
      skip_loop = false;
      continue;
    }
    for(j=0;j<table_len;j++){
      noInterrupts();
      // platform_gpio_write(pin, level);
      DIRECT_WRITE(pin, level);
      interrupts();
      delayMicroseconds(delay_table[j]);
      level=!level;
    }
    repeat--;
  } while (repeat>0); 

  return 0;  
}
```

nodemcu:
this gpio.serout() api is made for serializing data out of a pin.
hope it worked for IR.

klukonin:
Thank you very much!
I'll write a test lua module for it. This should be awesome for IR

PR #390, opened 2015-04-27
Add setphymode and getphymode to wifi module

dnc40085:
useful for changing max transfer rate and output power of pa


klukonin:
dnc40085, Great job!

Can you please add some information about API to your code?

dnc40085:
does that work?

klukonin:
dnc40085, Yes!
That's great =)

Issue #356, opened 2015-04-13
which SSID is station currently connected


timoline:
Can you add a command where you can see which SSID is station currently connected ?
or which SSID is used by wifi.sta.connect() ?

devsaurus:
Please check whether #404 resolves your request and comment or close this issue accordingly.

devsaurus:
Closed due to inactivity. Feel free to reopen in case the issue persists.

PR #404, opened 2015-05-09
Add function wifi.sta.getconfig() and more parameters to wifi.sta.config()

dnc40085:
I added to the function wifi.sta.config the option to set whether or not auto-connect is enabled upon configuration and also the option to connect to different APs that have the same SSID


dnc40085:
Added bonus, this is also a fix for issue  #356

PR #406, opened 2015-05-09
Add wifi.sta.getrssi()

dnc40085:
To conserve space I modified wifi_scan_done so it could be used with both wifi.sta.getap() and wifi.sta.getrssi()

The function wifi.sta.getrssi() first gets the current station configuration, then it gets the current wifi channel, then uses that information to setup a scan configuration to use with wifi_station_scan to get the AP info, then the C callback wifi_scan_done takes the RSSI value and forwards it to the users callback function.

This is complementary to my previous pull request https://github.com/nodemcu/nodemcu-firmware/pull/404


dnc40085:
I decided it would be easier to just to expose the option to set scan configuration in wifi.sta.getap() and use Lua script to get RSSI info.

Issue #416, opened 2015-05-14
I2C address limited to 0x7F 


saper-2:
Hello,

I can't scan whole i2c bus because ```i2c.address``` is limited to 127(dec) - 0x7F. I have TMP101 which have base address 0x90 so I can't access it because I got "wrong arg range"... This is just a simple fix ( I think that at least :D ):
In file:
```nodemcu-firmware/app/modules/i2c.c ```
change line 59 in function 
```static int i2c_address( lua_State *L )```
To this:
```C
if ( address < 0 || address > 255 )
```
I would be very grateful for quick fix (and .bin to flash :) )

And is possible to ```print()``` without trailing new line?
nodemcu is a great piece of work!

marcelstoer:
> I would be very grateful for quick fix

I'm sure you could speed up the process if you provided a pull request.

> (and .bin to flash :) )

Once it's integrated into one of the branches you could use http://frightanic.com/nodemcu-custom-build/ to have your own custom firmware built.

pipi61:
Hi!
i2c first address byte 8 bit, bit 0:r/w flag, bit 7...bit1: i2c address bit6..0
http://www.ti.com/lit/ds/sbos231h/sbos231h.pdf

saper-2:
TMP101: address bits are [7:1], [0] is R/W .
WIth ADD0=0 , address is [7:1]1001000 (0x90/0x91), with ADD0=float address is [7:1]1001001 (0x92/0x93) , with ADD0=1 address is [7:1]1001010 (0x94/0x95) . I have also AM2321 (humidity & temp. sensor on I2C) and it's address is 0xB8=[7:1]1011100 ...

I don't really know how to operate github (I think I'm too stupid for this...) I will try add pull request (I don't know how to do it *right now*...)

marcelstoer:
> I don't really know how to operate github (I think I'm too stupid for this...) I will try add pull request (I don't know how to do it right now...)

There's a first time for pretty much anything. Fortunately GitHub is extensive easy-to-follow documentation: https://help.github.com/articles/using-pull-requests/, https://help.github.com/articles/creating-a-pull-request/

saper-2:
Ok, I think I did it :smile: before you posted :smile: . The first link I'm have to read, but the 2nd I found and tried to follow it :smile: hope I did everything all right :smiley:

saper-2:
[...]

PR #417, opened 2015-05-14
Fix i2c.address range from 127 to 255 (Issue #416)

saper-2:
Fixed ```i2c.address``` to accept I2C addresses from 0 to 255 (previously was from 0 to 127) .
Please see if this is all need to be done to this fix.
Issue: https://github.com/nodemcu/nodemcu-firmware/issues/416


saper-2:
i2c library need more fix because it can't handle higher addresses (over 0x80). I'll see what I can do...

PR #424, opened 2015-05-17
added net.dns.setdnsserver('ip',[index]),  net.dns.getdnsserver([inde…

jorgempy:
…x]),  net.dns.resolve('domain.com', function cb('ip' or nil if fail) )

modified:   app/modules/net.c


vowstar:
TODO:FIX THE DOCUMENT
added:
net.dns.setdnsserver('ip',[index])
net.dns.getdnsserver([index]), 
net.dns.resolve('domain.com', function cb('ip' or nil if fail) )
modified: app/modules/net.c
@vowstar

Issue #433, opened 2015-05-19
adc.readvdd33() kills network?


errolt:
Hi NodeMCU.

I am new to lua so this might be my fault.

This code does not run as expected:
v=(adc.readvdd33())
sk=net.createConnection(net.TCP, 0)
    sk:on("receive", function(sck, c) print(c) end )
    sk:connect(80,"184.106.153.149")
    sk:send("GET /update?key=yourkey&field1=0 \r\n")
    sk:send("HTTP/1.1\r\nHost: 192.168.0.66\r\nConnection: keep-alive\r\nAccept: */*\r\n\r\n")

This should print "0" as thingspeak does not know the key. But uploading it via ESPlorer to the nodemcu board causes some kind of crash and a reboot. Running it again after the reboot does nothing and prints nothing.

If you remove the first line, containing adc.readvdd() then the code runs as expected and there are no crashes.

Running "NodeMCU 0.9.6 build 20150406  powered by Lua 5.1.4", think it is float version.

Thank you,
Errol

d-a-v:
Check this, this has been debuggued (search readvdd33)

https://github.com/nekromant/esp8266-frankenstein/blob/master/src/cmd_common.c

vowstar:
https://github.com/nodemcu/nodemcu-firmware/commit/fc8fa2cc915aea3687b068eb0355bb84b2fc113d
Applied nekromant‘s method.

dnc40085:
adc.vdd33() still causes reset when connected to AP...

Before the fix fc8fa2c I was testing different things out for a potential fix and I found that in stationAP and SoftAP mode the adc.readvdd33() took ~240 microseconds to execute. However in station mode it took ~225mS milliseconds.

cal101:
Moin,

did you ever read something somewhere and had problems to find it again?
I finally found it.

Here is a foot note

> Disable interrupts during bit-banging the WS2812. Avoid disabled interrupts for more then 10ms or 
> wireless connections will act wired.

from

http://naberius.de/2015/05/14/esp8266-gpio-output-performance/

So it may not be a good idea to disable interrupts in this case.

Cal

dnc40085:
so far the only way I can see to prevent crashing if the chip is in Station mode, is to verify it is disconnected before checking vdd33

errolt:
[...]

PR #437, opened 2015-05-23
Added fix to adc.readvdd33() to prevent crashing in station mode 

dnc40085:
concerning issue #433, when station is connected to an Access Point, a call to adc.readvdd33() causes a reset.
I believe this issue is either a limitation of the SDK or the esp8266ex hardware it's self.

I changed the adc.readvdd33 function to first check if the chip is in station mode and if so it will next check if the station is connected and if the station is connected it will return an error. otherwise it will read vdd33.

### DIFF
--- a/app/modules/adc.c
+++ b/app/modules/adc.c
@@ -24,11 +24,23 @@ static int adc_sample( lua_State* L )
 static int adc_readvdd33( lua_State* L )
 {
   uint32_t vdd33 = 0;
-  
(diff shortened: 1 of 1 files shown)


dnc40085:
thought about it after I submitted the PR,
should this return an error if connected or return nil?

vowstar:
I think print nil and write a detailed document about this is better. issue #433 may no solution, we are thinking how to do with it in special ways.

PR #443, opened 2015-05-26
Expose sha1 from rom 

creationix:
I'm just exposing a simple SHA1 function to lua that will make implementing websockets in lua much easier.

I also implemented a base64 encoder as needed by websockets.

### DIFF
--- a/app/include/user_modules.h
+++ b/app/include/user_modules.h
@@ -31,6 +31,7 @@
 #define LUA_USE_MODULES_U8G
 #define LUA_USE_MODULES_WS2812
 #define LUA_USE_MODULES_CJSON
+#define LUA_USE_MODULES_CRYPTO
 #endif /* LUA_USE_MODULES */
 
 #endif	/* __USER_MODULES_H__ */
--- a/app/modules/crypto.c
+++ b/app/modules/crypto.c
@@ -1,0 +1,90 @@
+// Module for cryptography
+
+//#include "lua.h"
+#include "lualib.h"
+#include "lauxlib.h"
+#include "platform.h"
+#include "auxmods.h"
+#include "lrotable.h"
+
+#include "c_types.h"
+#include "c_stdlib.h"
+
+#include "user_interface.h"
+
+/**
+  * hash = crypto.sha1(input)
+  *
+  *	Calculates raw SHA1 hash of input string.
+  * Input is arbitrary string, output is raw 20-byte hash as string.
+  */
+static int crypto_sha1( lua_State* L )
+{
+  // We only need a data buffer large enough to match SHA1_CTX in the rom.
+  // I *think* this is a 92-byte netbsd struct.
+  uint8_t ctx[100];
+  uint8_t digest[20];
+  // Read the string from lua (with length)
+  int len;
+  const char* msg = luaL_checklstring(L, 1, &len);
+  // Use the SHA* functions in the rom
+  SHA1Init(ctx);
+  SHA1Update(ctx, msg, len);
+  SHA1Final(digest, ctx);
+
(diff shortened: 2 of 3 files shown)


pastukhov:
Can you provide some usage examples?

creationix:
@pastukhov Sure, but I'm not sure where these functions go or where to put the docs.

```lua
-- The websocket GUID is part of the protocol spec
local GUID =  "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
-- The key comes in the request as the Sec-Websocket-Key header
local websocketKey = "SGVsbG8gV29ybGQ="
-- We send the response as Sec-Websocket-Accept
local websocketAccept = node.base64(node.sha1(websocketKey .. GUID))
```

The `node.sha1(data)` function simply returns a raw 20-byte sha1 without hex encoding since lua strings are binary safe.

The `node.base64(data)` function accepts binary data and returns the base64 encoded version.

pastukhov:
Great!

TerryE:
You are operating on a string and returning a string.  I think that the precedent here is `string.dump()`.  So if you want to place this in an existing module, then propose string. Or you add a new module. 

Node is definitely the wrong module to place this functionality, as this  provides system information.

creationix:
@TerryE string is an interesting place, but it's a lua builtin.  What's the opinion on changing builtins in elua?  Assuming I'm able to move it, this would make for some interesting syntax assuming it would go on string's metatable.

```lua
print(("Hello"):sha1():base64())
```

dnc40085:
@creationix maybe it could go in a new module...maybe we could name the new module "crypto".

I've been thinking of implementing a Time-based One Time Password (TOTP) token generator, which if I'm not mistaken uses the HMAC-SHA1 algorithm to generate the tokens. If I were to get it working I was thinking of making a new module to put it in since it wouldn't quite seem right in any of the preexisting modules.

TerryE:
@creationix, Tim my main point was that **node** is definitely the wrong place.  If you look at the node methods -- **restart**, **dsleep**, **info**, **chipid**, **flashid**, **heap**, **key**, **led**, **input**, **output**, **readvdd33** -- these are all to do with the hardware or system info.  I admit that **compile** is a late addition and doesn't really fit here, but it is doing esp8266-specific manipulation of dumped globs (it removes all of the debug info to reduce the runtime footprint).

The strong argument for making this a *separate* module is that this overhead would be optional for any users using Marcel's build approach. If we put this code in a core module, then it will be a mandatory runtime overhead.

creationix:
Ok, I think I figured out how to use the sha1 code in the rom https://github.com/creationix/nodemcu-firmware/commit/d7a8bd0c1afa809004352e8565fc9a6141f957cf.

I'm still trying to guess the signature of the base64_encode function.

I'll happily move this to a new module, but I'm not quite sure how to do that.

dnc40085:
@creationix I submitted a pull request to your dev096 branch with the necessary modifications to create a new module.
creationix/nodemcu-firmware#1

creationix:
Ok, cleaned up and moved code to new crypto module @dnc40085 helped make.

The `crypto.base64Encode` function is my own implementation for now till I figure out the signature of base64_encode in the rom.

dnc40085:
@creationix When I made the crypto module, I copied the adc.c and unlike node.c, the line  '#include "c_stdlib.h"' was missing, if you add that line to the top of crypto.c, it should compile correctly

creationix:
Since the `dev096` branch is so broken, I made a new pull request #447 against `dev`.  This other PR has proper docs in the details.

dnc40085:
I added a pull request that should bring the file system back to life on the dev096 branch.

PR #447, opened 2015-05-28
Add crypto module.

creationix:
Initial implementation of crypto module.  This contains the main helpers need to make an efficient websocket server in lua.

# Crypto

Contains basic string functions to perform basic crypto related actions.  The internal ROM is used 
when possible and minimal implementations in C are used otherwise.

Here is an example of calculating the `Sec-WebSocket-Accept` header using the value from `Sec-WebSocket-Key` and the GUID from the spec:

```lua
local key = "mZWV+am1H2aUtiiY+bN1mo+lgNQ="
local guid = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
local accept = crypto.toBase64(crypto.sha1(key .. guid))
```

Some messages in websocket are masked for security (prevent cache poisoning).  Here is an example of using the `crypto.mask` function to apply or remove the mask.

```lua
local message = "Hello World"
local mask = string.char(math.random(256) - 1, math.random(256) - 1, math.random(256) - 1, math.random(256) - 1)
local masked = crypto.mask(message, mask)
local original = crypto.mask(masked, mask)
```

## crypto.sha1(input) -> rawHash

Calculate the raw sha1 hash of a value.  If you wish to print the hash use either `hexEncode` or `base64Encode`.

## crypto.toHex(raw) -> hex

Encode a raw string using hex encoding.

## crypto.toBase64(raw) -> base64

Encode a raw string using base64 encoding.

## crypto.mask(message, mask) -> masked

Apply a mask to a message.  The mask is repeated using modulo so that each byte of the message is XORed with the `mask[index % maskLength]`

### DIFF
--- a/app/include/rom.h
+++ b/app/include/rom.h
@@ -1,0 +1,25 @@
+// Headers to the various functions in the rom (as we discover them)
+
+// SHA1 is assumed to match the netbsd sha1.h headers
+#define SHA1_DIGEST_LENGTH		20
+#define SHA1_DIGEST_STRING_LENGTH	41
+
+typedef struct {
+	uint32_t state[5];
+	uint32_t count[2];
+	uint8_t buffer[64];
+} SHA1_CTX;
+
+extern void SHA1Transform(uint32_t[5], const uint8_t[64]);
+extern void SHA1Init(SHA1_CTX *);
+extern void SHA1Final(uint8_t[SHA1_DIGEST_LENGTH], SHA1_CTX *);
+extern void SHA1Update(SHA1_CTX *, const uint8_t *, unsigned int);
+
+// base64_encode/decode derived by Cal
+// Appears to match base64.h from netbsd wpa utils.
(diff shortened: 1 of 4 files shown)


marcelstoer:
Great, as soon as it's been merged I'll add it to http://frightanic.com/nodemcu-custom-build/.

creationix:
Oops, didn't mean to commit my config changes, will force push revert soon.

creationix:
I made a sample websocket echo server that uses this new module.
https://gist.github.com/creationix/079e11345249525c7e2e

marcelstoer:
Just learned that it's each contributor's own responsibility to add new stuff to https://github.com/nodemcu/nodemcu-firmware/wiki/nodemcu_api_en-(dev096).

mikewen:
@Tim, great work! This is amazing! Now I can code-sign the Lua code and push to the NodeUSB.

Issue #444, opened 2015-05-26
File system isn't usable on dev096


pastukhov:
```lua
>     file.open("init.lua", "a+")
>     file.writeline('foo bar')
stdin:1: open a file first
```

creationix:
Do you have one of the new nodemcu devboards?  They require special flashing tools in windows to get a working filesystem.

pastukhov:
No. I have esp-01, esp-12, esp-201 and nodemcu devboard v0.9. And i am on ubuntu.

jonn26:
+1 I am seeing this as well on esp01 & esp-12.

drv53b6:
file.format()   -- ?

dnc40085:
+1 confirmed on esp-01 esp-07, compiled on lubuntu with esp-open-sdk.

I cloned the current dev096 branch, compiled, and burned, then ran file.format() then used file.fsinfo() and it returned `stdin:1: file system error` on both the esp-01 and the esp-07

pastukhov:
file.format() doesn't help
The problem  is in 5261fb194853f7f56c9578fa582f9984c3cd955e
Without it everything is fine.

pastukhov:
file.format() log with debug:
```
> file.format()
_flash_used_end:40268eb6
fs.start:4026c000,max:10000
CACHE_ALLO: allocated cache page 0
page index byte len:         512
object lookup pages:         1
page pages per block:        128
page header length:          5
object header index entries: 234
object index entries:        253
available file descriptors:  3
free blocks:                 1
mount res: 0
erase 4026c000:00001000
erase 4026d000:00001000
erase 4026e000:00001000
erase 4026f000:00001000
erase 40270000:00001000
erase 40271000:00001000
erase 40272000:00001000
erase 40273000:00001000
erase 40274000:00001000
erase 40275000:00001000
erase 40276000:00001000
erase 40277000:00001000
erase 40278000:00001000
erase 40279000:00001000
erase 4027a000:00001000
erase 4027b000:00001000
CACHE_GET: have cache page 0 for 0000
CACHE_GET: have cache page 0 for 0000
_flash_used_end:40268eb6
fs.start:4026c000,max:10000
CACHE_ALLO: allocated cache page 0
page index byte len:         512
object lookup pages:         1
page pages per block:        128
page header length:          5
object header index entries: 234
object index entries:        253
available file descriptors:  3
free blocks:                 1
mount res: 0
format done.
```
file.fsinfo()
[...]

PR #450, opened 2015-05-29
Revert "Update spiffs to latest." on dev096 branch

dnc40085:
This reverts commit 5261fb1 which breaks filesystem (#444)


dnc40085:
upon further testing found that this still doesn't fix the problem

dnc40085:
I replaced the folder with an older version (commit: 8f925e6) and it seems to fix the problem with the filesystem

PR #460, opened 2015-06-01
Added option to wifi.sta.getap to allow changing output format

dnc40085:
It has come to my attention that some people would like the option to select between the old style and new style output format for the table returned by wifi.sta.getap().

This PR adds that functionality.

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -599,9 +599,14 @@ static int wifi_station_setauto( lua_State* L )
   * 	scan and get ap list as a lua table into callback function.
   * Syntax:
   * 	wifi.sta.getap(function(table))
+  * 	wifi.sta.getap(format, function(table))
   * 	wifi.sta.getap(cfg, function(table))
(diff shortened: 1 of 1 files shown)


dnc40085:
When I was coding the addition of the `cfg` parameter, I initially was going to add it at the end to make the source more streamlined and to easily make it backwards compatible. 
I figured it would confuse people if the extra parameters were at the end, especially if the example was written like this:
```lua
    wifi.sta.getap(function(t)
     print("\n\t\t\tSSID\t\t\t\t\tBSSID\t\t\t  RSSI\t\tAUTHMODE\t\tCHANNEL")
      for bssid,v in pairs(t) do
       local ssid, rssi, authmode, channel = string.match(v, "(.*),(-?%d+),(%d+),(%d+)")
        print(string.format("%32.s",ssid).."\t"..bssid.."\t  "..rssi.."\t\t"..authmode.."\t\t\t"..channel)
      end
    end, scan_cfg, 1)
```
That's why I went with the messier source code.

PR #462, opened 2015-06-02
Crypto module

jmattsson:
Hi there,

I don't know if there's any interest in this, but we've built a small crypto module for exposing MD2/MD5/SHA1 and optionally SHA{256,384,512} in Lua, as well as the HMAC versions of these.

I imported Aaron Gifford's SHA256 code into libssl since that seemed the most appropriate location, though currently the SSL library does not make use of it. Feel free to move it somewhere else if you want. The crypto module (standalone commit 20bc8f4) can be built without the SHA256 support if WITHOUT_SHA2 is defined, making it useful even without the Gifford code.

Example usage:

    local hash = crypto.hash('MD5', "abc")
    local signature = crypto.hmac('SHA256', "abc", "mysecretkey")

Known limitations:
- HMAC keys must not exceed the block length of the chosen hash function
- While I'm familiar with Lua, I'm new to NodeMCU so a certain measure of "monkey see, monkey do" was used in the making of this module...


creationix:
I recently added a crypto module in the dev branch.  Hashes are already in the device ROM and can be used directly without duplicating code and using extra memory.

I only exposed sha1 from the hashes since that was all I needed to get websocket working, but it shouldn't be hard to add the other hashes in the rom exposed in similar manner.

https://github.com/nodemcu/nodemcu-firmware/pull/447

jmattsson:
Cool! I hadn't seen that one since I've been over on master! Would you be opposed to me plugging in my hash&hmac functions and deprecate or remove the .sha1() function?

I'm not sure why you're saying I'm duplicating code and using extra memory - objdumping both the included libssl.a and the newly built libssl.a both have the hash functions sitting in the .irom0.text segment. This includes the SHA256/384/512 functions which are not available in the default libssl (and SHA256-HMAC is what I need for posting to a particular IoT service). If it's preferable I could of course put the SHA256-512 support in the crypto module code.

jmattsson:
Closing this pull request - will work on extending creationix's module instead.

creationix:
I'm not expert on memory usage, I'm mostly repeating what I've been told.  But if you look at my PR, I'm not using my SHA1 implementation or the one in the ssl lib.  I'm just calling the SHA1 that's already in the hardware.  The trick is guessing the proper signature from assembly dumps.

I don't think it affects free heap much either way, but I would imagine including an external sha1 implementation would make the code size larger and use up more flash.  I'm not sure how

cal101:
Moin,

maybe I can shed some light on the memory issues.
Esp8266 has a modified harvard architecture. That means code and memory are separated.
So code size does not have an effect on heap memory. (dram0_0_seg)

Executed code comes from 3 places:
- Firmware ROM (0x40000000), That contains the SHA1 code @creationix uses.
- irom0_0_seg :                         org = 0x40210000, len = 0x60000
  This is code you flashed which is directly executed frpm flash speeding thinks up
  using 32k (I think) of cache ram. This reduces your size of a flash file system if increased.
- iram1_0_seg :                         org = 0x40100000, len = 0x8000
  This is another section in the code you flashed which is copied to instruction ram
  on startup. This reduces your size of a flash file system if increased.

Interrupt code and (I think) code using the SPI interface where the flash is connected to
must reside in Firmware ROM or iram because otherwise it will get in conflict with
loading from flash.

Cal

TerryE:
Thanks Cal, a very useful summary.  The ESP8266 is a _closed source_ SoC, so the details are quite murky. It has at least two CPUs on the die. The main CPU is an Xtensa LX.  This is comparable to the bottom end of the ARM range of CPUs.  (There's also a I/O CPU for the WiFI)  The system runs at 80MHz and ROM access will run pretty much at full throttle.  Flash access is pretty slow in comparison, but there is as Cal says a 32KB RAM cache in front of this (again details are murky).

creationix:
So using the hash functions in the ROM will save space in the cache and save more space for the filesystem.  Plus there is a chance they might be faster since they are part of the ROM?

@jmattsson how do you feel about using more of the ROM functions?

creationix:
From <http://hackaday.com/2015/03/18/how-to-directly-program-an-inexpensive-esp8266-wifi-module/>

> There’s a few things to take note of when programming the ESP8266. One is that there are a bunch of functions that are built into the ROM you can leverage. This means you don’t need to (and shouldn’t) re-implement MD5 (even with hmac), SHA1, comms with the external flash, SPI, UART functions, software floating point functions, AES, memory commands, printf, low-level IO tools, even a real-time event scheduler, and, most importantly… 802.11!

> The full list of functions that reside in the on-chip ROM can be found in [eagle.rom.addr.v6.ld](https://github.com/nodemcu/nodemcu-firmware/blob/master/ld/eagle.rom.addr.v6.ld). It would behoove you to take a look in there now and see the wonderland of awesome that’s provided out of the box.

It doesn't appear to have the SHA2 family, but it does have SHA1 and MD5 with and without HMAC.

jmattsson:
The SHA1{Init,Update,Final} prototypes look compatible with the SHA1_{Init,Update,Final} so they should work as a drop-in replacement I think. Presumably the MD5 functions follow suit. I'll see if I can get a bit of time tomorrow to test this. If it works, I'll switch them over.

creationix:
@jmattsson make sure to use the header I started for ROM functions and add signatures as you discover them.  This way we'll have at lease some semblance of sanity and typechecking.

jmattsson:
@creationix All updated! I've opened a new pull request: https://github.com/nodemcu/nodemcu-firmware/pull/467

One thing I learned while doing this was that any and all access to the mapped flash (irom0.text) segment is only possible in 32bit loads. Byte and dword loads generate an exception and a reboot.  This behaviour was confirmed to match the LX106 configuration, once we found that documentation. Use of ICACHE_RODATA_ATTR to get read-only constants out of RAM thus comes with some rather annoying limitations.

cal101:
Moin,

reading your post my first reaction was: Sure it must be aligned, works as expected because its instruction area. 
But we *don't* access the flash I think. We access an area that is *configured*
to be mapped to the flash and to use some memory as cache.
But I think I read somewhere (ISA specs?) that this configuration includes the purpose
of the area (instruction in our case) and that it's valid to map something to different areas
with different configurations.
So it MAY be possible to map that differently.

Maybe a direction to explore.

I can see that int4 aligned access may be awkward for strings. 
(Hmm why do my typical strings work? Need to take a look which area they get into.)
What usage are you thinking of that would have major problems with aligments?

Cal

jmattsson:
As far as we can tell, the type of memory (data vs instructions) is determined when the LX106 silicon itself is configured.

It's not merely alignment which is enforced, it's accesswidth as well. For example, while placing and using this struct in instruction memory (be it flash or ram):

    struct {
      uint32_t a;
      uint32_t b;
    } foo;

would work fine, the following dies as soon as b or c is accessed:

    struct {
      uint32_t a;
      uint16_t b;
      uint16_t c;
    } bar;

This is due to the instruction areas only allowing access via L32x instruction(s?), so even though b above is properly aligned, it would be accessed via an L16x instruction, which is not possible in instruction areas, and instead generates an exception and causes a reboot.

Your strings most likely work because they end up in the .rodata segment, which is loaded into data RAM. The actual pointer to the string is 32bit aligned and loaded, so accessing it in instruction areas works well.

What we were trying to do was to shift the entire .rodata segment into irom0.text, since well, it's read-only, it doesn't have to be in writable RAM. We were looking at replacing the regular entry-point (call_user_main) with a trampoline which mapped the flash before calling through to call_user_main. Alas, it would seem this approach can not be made to work generically due to the L32x constraint. Stashing constants in irom will continue to be a manual game, unless of course someone has tons of time and energy and manages to build a toolchain which is clever enough to always to 32bit loads from these areas and just throw away the unwanted bits.

Annoyingly enough it also seems the silicon is going to extra lengths to only allow 32bit loads from these areas.

cal101:
Some yes. Some no. Some maybe.
Its not a pure harvard architecture.
WITLB / WDTLB play a role (see reset handler)
and some additional unknown cache configuration stuff (Cache_Read_Enable) adds to the mix.
I understand what you want to archieve and try to do the same.
I hinted to the rom symbols and their usage.
I know a bit or two about the chip but in the esp8266 forum are people I ask for support.

If you really want to access the data unaligned install a handler to support that
and it may be worth the handler overhead for 80% of the cases.
For the rest 20% the usage maybe improved by changing the code.

I feel discussing this in some pull request is non productive.
Please open a topic about this in the forum and so others can notice and join.
I would like to join that discussion.

Over and out,
Cal

PR #463, opened 2015-06-02
Crypto module, take 2

jmattsson:
Hello again,

Since there's clearly interest in a Lua crypto module, I reworked my earlier stuff to integrate with creationix's crypto module instead. This branch now extends it to provide the .hash() and .hmac() functions, and also introduces support for SHA{256,384,512} through the newly created libcrypto.a. This is using Aaron Gifford's SHA256 implementation. The whole thing can be built without the SHA256 support if WITHOUT_SHA2 is defined, for those wishing to conserve space and not needing any of the SHA2 stuff.

Example usage:

local rawhash = crypto.hash('MD5', "abc")
local rawsignature = crypto.hmac('SHA256', "abc", "mysecretkey")

### DIFF
--- a/app/crypto/digests.c
+++ b/app/crypto/digests.c
@@ -1,0 +1,168 @@
+/*
+ * Copyright (c) 2015, DiUS Computing Pty Ltd (jmattsson@dius.com.au)
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTOR(S) ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTOR(S) BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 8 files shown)


marcelstoer:
> The whole thing can be built without the SHA256 support if WITHOUT_SHA2 is defined

Shouldn't you add that to `app/include/user_config.h` then? And invert to logic so that one explicitly needs to request SHA2? For reference I looked at `CLIENT_SSL_ENABLE` in  https://github.com/nodemcu/nodemcu-firmware/blob/master/app/include/user_config.h#L42. So, I suggest to call that flag `SHA2_ENABLE` or so instead.

dnc40085:
heres the prototype for `PROVIDE ( hmac_sha1 = 0x4000ba28 );` 
`int     hmac_sha1 (const u8 *key, size_t key_len, const u8 *data, size_t data_len, u8 *mac);` in case you want to use some of the built in functions to save space

app/crypto/digests.c
```diff
@@ -0,0 +1,168 @@
...
+const digest_mech_info_t *crypto_digest_mech (const char *mech)
+{
+  if (!mech)
+    return 0;
+
+  size_t i;
+  for (i = 0; i < (sizeof (hash_mechs) / sizeof (digest_mech_info_t)); ++i)
+  {
+    const digest_mech_info_t *mi = hash_mechs + i;
+    if (strcasecmp (mech, mi->name) == 0)
```
creationix: I've noticed that very few of the existing nodemcu functions accept strings from lua for anything other than string data.  Perhaps the heap overhead of all those interned strings for things like `"SHA1"` is expensive?  

Or perhaps the other functions are just lazy and not very script friendly.

app/crypto/digests.c
```diff
@@ -0,0 +1,168 @@
...
+      return mi;
+  }
+  return 0;
+}
+
+static const char hex[] = "0123456789abcdef";
```
creationix: FWIW, I already had a hex encoder in my crypto PR (though as a lua function and not internal C function like this one).

Is this used internally by other C functions?

jmattsson:
@marcelstoer Sure, can do! I'd had forgotten about the existence of user_config.h. I've made the change and added to this branch.

@dnc40085 Thanks for the prototype. In this case it would actually mean more code to custom-case  use hmac_sha1() and/or hmac_md5(). I have a single hmac function which can use any hashing algorithm - not one hmac_xxx() function for each algorithm.

app/crypto/digests.c
jmattsson: Fair point. Maybe I should change them them to use numeric constants and add them as crypto.MD2, crypto.SHA1 etc like the net module does with TCP/UDP? Though on the other hand that's likely less memory efficient in uncompiled Lua. Happy to accept guidance here on which way is preferred.

app/crypto/digests.c
jmattsson: Yeah, I'm using it from another module I'm writing. I did see your hex encoder function, but it looked small enough that it didn't seem worthwhile to replace its internals with a call to mine. I figured in most cases mine would go unused and the linker would omit it altogether.

Issue #459, opened 2015-05-31
Problem with multiple MQTT subscriptions


devsaurus:
I'm setting up an application where the nodemcu firmware is supposed to subscribe to 3 topics as shown here:
```lua
m = mqtt.Client("esp8266", 120, "", "")

m:on("message", function(conn, topic, data) 
  print(topic .. ":" ) 
  if data ~= nil then
    print(data)
  end
end)

m:connect("192.168.0.254", 1883, 0, 0, function(conn) print("connected") end)

m:subscribe("/topic1", 0, nil)
m:subscribe("/topic2", 0, nil)
m:subscribe("/topic3", 0, nil)
```
All seems fine as long as the messages come in asynchronously. But when all three topics deliver messages (more or less) at the same time, the third topic is not processed properly. Eg publishing on another client as follows (srv is on IP 192.168.0.254):
```
mosquitto_pub -h srv -i cli -q 0 -t /topic1 -m 1; mosquitto_pub -h srv -i cli -q 0 -t /topic2 -m 2; mosquitto_pub -h srv -i cli -q 0 -t /topic3 -m 3
```
yields:
```
/topic1:
1
/topic2:
2
/topic2:
2
```
Two messages at once is no problem, no matter which topics are used. But having all 3 in one row causes the last message on the 3rd channel being overwritten by the 2nd message.

Am I doing something wrong here? I'd guess the callback is somehow interlocked, but it seems strange. BTW the other method to subscribe to multiple channles in the README.md doesn't work at all for me:
```lua
m:subscribe({["/topic1"]=0, ["/topic2"]=0, ["/topic3"]=0, nil})
```
It spawns a huge number of prints for a single publish...

TerryE:
@devsaurus, this isn't "I found a problem / bug with the firmware"; it's an "I don't know how to get mqtt to work".  This issue list is the former; try www.esp8266.com for the latter and we'll help you there :smile:

drv53b6:
how to report spam? this @TerryE  alweys spam with hes www.esp8266.com

TerryE:
@drv53b6, esp8266 is my forum and I have no affiliation with it other than it's the only forum that I know of which offers such free support to ESP8266 Lua developers, so I provide help there to developers, just as I do on StackOverflow which doesn't have an active eLua / ESP8266 tag.  Even if you don't agree with my suggesting this developer forum and would recommend a _free_ alternative, my underlying point that this isn't a bug with the firmware still stands.

[...]

PR #465, opened 2015-06-04
Fix Problem with multiple MQTT subscriptions #459

cal101:
Problem with multiple MQTT subscriptions #459
	https://github.com/nodemcu/nodemcu-firmware/issues/459

Fixed  wrong buffer pointer increment.


klukonin:
Ahhah, Great job.

Really helps to fix this bug!

devsaurus:
This issue is also present in `dev` branch. Do you prefer a separate PR for that?

PR #467, opened 2015-06-05
Crypto module improvements

jmattsson:
As per discussion over on https://github.com/nodemcu/nodemcu-firmware/pull/462 I've updated the crypto module to use the ROM MD5 & SHA1 functions instead of the libssl versions. Prototypes for the MD5 functions have been added to rom.h.

Additionally most of the static const data used by this module has been relocated to irom0.text in order to free up precious RAM.

### DIFF
--- a/app/crypto/digests.c
+++ b/app/crypto/digests.c
@@ -29,35 +29,43 @@
  */
 #include "digests.h"
 #include "user_config.h"
+#include "rom.h"
 #include "lwip/mem.h"
-#include "lwip/arch.h"
-#include "ssl/ssl_crypto.h"
-#include "sha2.h"
 #include <string.h>
 #include <c_errno.h>
+
+#ifdef MD2_ENABLE
+#include "ssl/ssl_crypto.h"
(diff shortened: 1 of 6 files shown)


creationix:
+1

@vowstar, I wonder how hard it would be to set up some CI (continuous integration) service where every commit is tested to see how much heap is free upon startup?  This way we could track memory regressions easily.  I guess the tricky part is actually running the code on actual hardware.  That might be hard to automate.

jmattsson:
One thing that should be fairly easy to do is track the size of the .rodata segment. All const data ends up there by default and chews up RAM. As I mentioned over in the pull/462 thread it would not seem we can relegate that segment to flash alone, though you're certainly welcome to try to find a loophole or clever way which we haven't thought of over there!

As for running on actual hardware, that should be doable if someone is willing to invest some time in it. We've got a number of systems in CI where we flash new builds and fire up test harnesses on the devices' serial ports to test against regressions. None for esp8266 nodes yet, though.

Issue #394, opened 2015-04-28
Add support for MAX7219 8x8 LED matrix


marcelstoer:
#50 brought SPI support to the firmware, nice. After the issue was closed a [discussion in the comments](https://github.com/nodemcu/nodemcu-firmware/issues/50#issuecomment-75832880) was started about support for the popular MAX7219 8x8 LED matrix.
Since the results were inconclusive and the discussion kinda faded out with @morganrallen stating 

> I think neither the NodeMCU nor the ESP8266 SPI interfaces can handle the MAX7219

I'd like to see this handled in a dedicated issue.

If you can prove the above statement false then it'd be nice to at least see an example for the MAX7219 8x8 LED matrix added.

morganrallen:
I did get it working and I even released a demo. The last comment in that thread has a link to my demo code that I was able to control 4 devices.

marcelstoer:
> I did get it working and I even released a demo.

I know, well done. However, making a custom build against the ESP SDK operates a few layers below writing a few dozen lines of Lua code against the NodeMCU firmware. I'm more interested in the latter.

morganrallen:
Yes true, but the demo is a good starting point and shows the differences
between how NodeMCU and the SDKs SPI interfaces differ from what the MAX
expects. I've been away from all this for about a month, but off the top of
my head the primary problem was the MAX sends two bytes at a time, this is
how it addresses down the chain (I think).
Hopefully I will be back on my project using these boards early next week
and I'll wrap my head around it again.

On Tue, Apr 28, 2015 at 3:39 PM, Marcel Stör <notifications@github.com>
wrote:

> I did get it working and I even released a demo.
>
> I know, well done. However, making a custom build against the ESP SDK
> operates a few layers below writing a few dozen lines of Lua code against
> the NodeMCU firmware. I'm more interested in the latter.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nodemcu/nodemcu-firmware/issues/394#issuecomment-97246462>
> .
>



-- 
http://ithoughtyouweretherobot.com
Metal and Wire

http://nolonelyguineapigs.com/
Wandering and Rambling

http://morglog.org
Old and Neglected.

-----BEGIN PGP PUBLIC KEY BLOCK-----
Version: OpenPGP.js v.1.20121007
Comment: http://openpgpjs.org

[...]

PR #496, opened 2015-06-23
Create PCD8544.c

aguaviva:
Quick and simple module to drive a pcd8544 without needing the ug8.

More info here: http://codinglab.blogspot.be/2015/06/simple-nodemcu-module-to-drive-nokia.html

Also note that i wrote a lua module to drive the display, although it works it takes too much mem.


marcelstoer:
I don't think this should be added to the firmware for the following reasons:
- it's not (yet) a proper module that can be turned on/off in `app/include/user_modules.h`
- the pins are not configurable
- I'd argue that the use case (pcd8544) is too specific

aguaviva:
I can add the missing files app/include/user_modules.h and make the pins configurable, no problem.

As for the pcd8544 (Nokia5110) LCD, these babies are everywhere and cost about 2 dollars. 

How about I add the missing files, and then we let the users decide through frightanic?

marcelstoer:
> How about I add the missing files, and then we let the users decide

I guess this is more of a general question that should be answered by @vowstar or @MarsTechHAN. Should the firmware provide "bindings" (of some sort) for widely used input/output components? If so, when is a component considered "widely used"? 

In issue #394, adding support for the MAX7219 8x8 LED matrix, I raised this issue before.

aguaviva:
Understood. Ideally I'd like nodemcu to support as many I/O componets as possible (provided I can disable them in the build). The same thing happens in the linux kernel, you have a mainstream build that has most of the stuff people need, and you can also optimize/customize it by removing/adding the bits you don't need.

MarsTechHAN:
We can just add them to he code, and defualtly turn it off. So the guys who use it can turn it on...How about this?

MarsTechHAN:
As for I myself, I will agree to add that modules. I believe the best way is defaultly turn it off, and for some one who needs it, he can open that. And the coming ACC (auto cloud compileing) system can solve it, by just opening the modules the users will use...

Many moudles like ws2812, dht, rc, crypto...lib someone may dont need them, but default lib turn tem on, which I believe is a bad idea.

As a conclusion, more lib? Absolutely yes. But remember to turn them off by default, and as marcelstoer saying
- [ ] Add turned on/off function in app/include/user_modules.h (And default close)
- [ ] Make the pins configurable

aguaviva:
Agreed! I'll be glad to make those changes.

MarsTechHAN:
I cannot decide it, the user of NodeMCU will decide it...both 8544 and 7219 are useful(At least for me) for someone and that's enough. Dont cause problem for others and make the user feel right, that's enough.

vowstar:
Thank you very much, and we will help modify it after merge, we will make it can be turned on/off in app/include/user_modules.h and pins are configurable. Welcome every PR!

dvv:
Wonder if it is feasible to take frankenstein code and build system and salt it with just Lua interpreter?

vowstar:
May be using jmattsson's code https://github.com/nodemcu/nodemcu-firmware/pull/504 very possible ;)

vowstar:
Web build service is useful, if possible we will develop a local build tool and let user choose what modules on/off, just like linux kernel.

aguaviva:
awesome!

goliatone:
is this going to get merged?

jmattsson:
Will this PR see the requested updates as requested, or should I close this for now?

marcelstoer:
@aguaviva hasn't responded to last request over a month ago. I'm, therefore, closing this PR.

PR #501, opened 2015-06-24
Newbase64

md5crypt:
well, I've wrote a working base64, probably useless, but I needed it.


vowstar:
We re-pull it to nodemcu/dev https://github.com/nodemcu/nodemcu-firmware/pull/507
And will into master after testing.
Thanks for support NodeMCU!

vowstar:
Finished push.

PR #503, opened 2015-06-25
fix data conflict of UART RX & node.input()

cs8425:
sorry for my poor Engilsh and I'm a noob of git.

#### Why & How
data conflict if UART has data come in and lua call node.input() at same time.
split data to 2 buffer fix this.

trade off: some RAM
 - UART RX buffer(interrupt): 2048 Byte @ app/include/driver/uart.h:8
 - UART RX buffer(for lua readline): 1024 Byte @ app/include/user_config.h:19

org:
 - UART_RX >> interrupt buffer >> lua.c:gLoad->line[] >> lua.c:readline() >> exec or call  uart.c:uart_on_data_cb()
 - node.input() >> lua.c:gLoad->line[] >> lua.c:readline() >> exec or call uart.c:uart_on_data_cb()

mod:
 - UART_RX >> interrupt buffer >> lua.c:UART_buf[] >> lua.c:readline()
 - node.input() >> lua.c:gLoad->line[] >> lua.c:readline()


TerryE:
> trade off: some RAM

No this is taking a relatively large hit on RAM for something that few users want or need.  This change is not for the common good.  Maybe we should add some sizing parameters into user_config.h and expose them as a customisable table in the cloud build, but no, this shouldn't be the default.

cs8425:
The default interrupt buffer size is 256 Byte.
Using extra 256 Byte as lua readline buffer will be acceptable?

(I have already add sizing parameters into user_config.h

jorgempy:
plus this should be merged to dev, not master.

PR #504, opened 2015-06-26
Move const data to ROM, freeing up ~16k of RAM

jmattsson:
This branch contains the necessary support for moving all const data (.rodata* sections) out of RAM into IROM. The barrier to doing this before was that the IROM is only capable of being accessed by the 32bit load instructions, so any 8bit or 16bit wide constants caused the ESP to reboot.

To address this, we install a user exception handler as the first thing on boot, which gets invoked whenever a sub-32bit wide load is performed and in those circumstances synthesize the load instruction by reading the whole 32bits and then masking/shifting appropriately.

The upside of this is that we free up a bit over 16k of RAM, almost doubling the available Lua heap. The tradeoff is that 8bit and 16bit const data access gets a fair bit slower. There is room for a reasonable amount of optimization in this area if it turns out to be required - to start off I've focused on clarity and maintainability. If it's really necessary, it should be possible to bypass the XTHAL completely and write a hand-optimized handler in assembler and put that straight into the user exception vector. It's also possible to explicitly put performance critical 8/16bit constants back into RAM using the RAM_CONST_ATTR #define on them, and of course in some cases it might be better to simply widen the constants to 32bits.

One thing to be mindful of is that with this change, rodata is only accessible once the IROM has been mapped by Cache_Read_Enable() which is called by the SDK inside call_user_start(). Currently call_user_start() does not use any const data before it enables the mapping, but if a future release used by NodeMCU does so, the solution is to instruct the linker script to put the libmain.a .rodata sections into RAM. Having the IROM mapped before calling into call_user_start() does not work, as that function assumes an umapped state.

This change is relatively small and well contained; The tricky bit really was working out all the small individual steps that had to be put together to make it all work. Well, that and groking lots of disassemblies...

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -99,7 +99,7 @@ LINKFLAGS_eagle.app.v6 = 			\
 	-nostdlib 				\
     -T$(LD_FILE) 				\
 	-Wl,--no-check-sections 		\
-    -u call_user_start 				\
+	-Wl,--wrap=_xtos_set_exception_handler \
 	-Wl,-static 				\
 	-Wl,--start-group 			\
 	-lc 					\
--- a/app/include/rom.h
+++ b/app/include/rom.h
@@ -37,3 +37,57 @@ extern unsigned char * base64_decode(const unsigned char *src, size_t len, size_
 // initialized because it uses mem_malloc
 
 extern void mem_init(void * start_addr);
+
+
+// Hardware exception handling
+struct exception_frame
+{
+  uint32_t epc;
+  uint32_t ps;
+  uint32_t sar;
+  uint32_t unused;
+  union {
+    struct {
+      uint32_t a0;
+      // note: no a1 here!
+      uint32_t a2;
+      uint32_t a3;
+      uint32_t a4;
+      uint32_t a5;
+      uint32_t a6;
+      uint32_t a7;
+      uint32_t a8;
+      uint32_t a9;
+      uint32_t a10;
+      uint32_t a11;
+      uint32_t a12;
+      uint32_t a13;
+      uint32_t a14;
+      uint32_t a15;
+    };
+    uint32_t a_reg[15];
+  };
+  uint32_t cause;
+};
+
+/**
+ * C-level exception handler callback prototype.
+ *
+ * Does not need an RFE instruction - it is called through a wrapper which
+ * performs state capture & restore, as well as the actual RFE.
+ *
+ * @param ef An exception frame containing the relevant state from the
+ *           exception triggering. This state may be manipulated and will
+ *           be applied on return.
+ * @param cause The exception cause number.
+ */
+typedef void (*exception_handler_fn) (struct exception_frame *ef, uint32_t cause);
+
+/**
+ * Sets the exception handler function for a particular exception cause.
+ * @param handler The exception handler to install.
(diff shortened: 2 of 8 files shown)


vudangngoc:
great news (y)

vowstar:
Fantastic

pastukhov:
and to dev096 ?

pastukhov:
```lua
> =node.heap()
35736
```
wow!

vowstar:
Yes, Very cool. This commit saved NodeMCU project.

pastukhov:
I have cherry-picked this commit to my local dev096 and it is working.

marcelstoer:
I guess I can take the down my cloud build service now ;-)

TerryE:
> I guess I can take the down my cloud build service now ;-)
Nah, someone will always find a way of using up all the RAM.  :smile:

vowstar:
Based this commit many impossible things become possible, and NodeMCU have much ram can optimize.

nodemcu:
@jmattsson Great Work!!!~~~Thank you! 
you made me a year!

matteovigni:
@jmattsson You did a great work with this commit but I think i found a bug in it. If I put my nodemcu in softap or stationap mode it reboots as soon as a client tries to connect to the ap.

jmattsson:
@matteovigni  Looks like there is a bug in the dhcpserver code. It os_memcpy()s into the const array "xid", which obviously isn't possible now that the constants really are constant.

I'll get a pull request ready in a minute - thanks for spotting the problem!

jiegec:
What's the difference between this and ICACHE_RODATA_ATTR ICACHE_FLASH_ATTR? Are those still needed? Which way is faster?

jmattsson:
@jiegec Those aren't strictly needed any more. The linker file used by NodeMCU now places both code and const data into irom by default. The only time you need to use an explicit section attribute is if you have code which runs before the irom is mapped (very unlikely), or if you have performance critical 8bit or 16bit tables (also unlikely). If you run into the latter, you could either do what the u8g modules used to do, and manually load & chunk the const data from a 32bit load (this avoids the overhead of triggering the load exception), or mark the const data to be placed in RAM (which is the fastest performance wise, but very costly in terms of RAM).

Actually, come to think of it, there is also a third option: Manually cache the const data to RAM on an as-needed basis. Just os_malloc() a buffer and bulk copy your const data into it before the performance critical part, use the constants from RAM, then free it once you're done.

My recommendation is to not do anything special - in most cases that will already be the optimal.

jiegec:
@jmattsson Thank you very much!

TerryE:
@jmattsson  Johnny, a random thought that (a) might be wrong and/or (b) not be worth all the hassle.  I suspect that most of the non 32bit aligned fetches from .rodata are operations on string constants.  90+% of these will probably be `c_memcpy()`,  `c_strcpy()` or `c_strcmp()`, so maybe replacing these by smart versions would remove most of the runtime overhead of this patch.

jmattsson:
If there is a genuine need for the extra performance, that might be an easy way to go about it. I'd like to see some performance stats first though. Also, as I mentioned somewhere in the initial stuff - there is plenty of room to optimise the exception handler itself, but again, there'd need to be a really good need for the extra performance to warrant the extra work (and overhead in lack of maintainability of a heavily optimised handler).

Issue #510, opened 2015-06-26
Cannot connect to the device via access point on latest pre-build Integer 2015.06.25


saitejalakkimsetty:
The device seems to crash repeatedly when we try to connect to ESP via the access point mode on the latest pre-build (Integer 20150625)
```
apcfg = {}
apcfg.ssid = "1234567000"
apcfg.pwd = "123456789"
wifi.ap.config(apcfg)
```

The output: (It repeats)
```
NodeMCU 0.9.6 build 20150625  powered by Lua 5.1.4
lua: cannot open init.lua
> ¤„FêÊâˆÖD¦ýDH
ÊŒ÷ìô

NodeMCU 0.9.6 build 20150625  powered by Lua 5.1.4
lua: cannot open init.lua
> ?)1ê&âˆÖDŒâõJHŽDLäô

NodeMCU 0.9.6 build 20150625  powered by Lua 5.1.4
lua: cannot open init.lua
> 
```

PR #511, opened 2015-06-27
Unbreak AP mode after rodata changes

jmattsson:
This fixes the problem where nodeMCU crashes when in SOFTAP/STATIONAP mode after the changes to the rodata sections.

Turns out the dhcpserver code was os_memcpy()ing into a const array...

### DIFF
--- a/app/lwip/app/dhcpserver.c
+++ b/app/lwip/app/dhcpserver.c
@@ -13,7 +13,7 @@
 #include "user_interface.h"
 
 ////////////////////////////////////////////////////////////////////////////////////
-static const uint8_t xid[4] = {0xad, 0xde, 0x12, 0x23};
+static uint8_t xid[4] = {0xad, 0xde, 0x12, 0x23};
 static u8_t old_xid[4] = {0};
 static const uint8_t magic_cookie[4] = {99, 130, 83, 99};
 static struct udp_pcb *pcb_dhcps = NULL;


MarsTechHAN:
Fixed #510

saitejalakkimsetty:
@MarsTechHAN The problem still exists. When I freshly flash the firmware, I am able to connect. But after leaving the ESP to idle for some time (like an hour) and then try to connect back, if simply does not connect anymore.

jmattsson:
I'm unable to reproduce this problem here.

@saitejalakkimsetty I've added some extra debugging prints to the branch at https://github.com/DiUS/nodemcu-firmware/tree/loadnon32-debug . Could you please try reproducing the problem on that firmware? If the issue is related to the const change, it should repeatedly spit out diagnostics on the serial before it reboots, which should allow us to pinpoint the error.

saitejalakkimsetty:
@jmattsson Would it be possible to link me the firmware itself? I dont exactly know how to build form the source.

jmattsson:
@saitejalakkimsetty Okay, I've built the loadnon32-debug branch for you and released here: https://github.com/jmattsson/nodemcu-firmware/releases/tag/loadnon32-debug

If you're connected to the serial port it will be *very* obvious if you hit a bad load.

saitejalakkimsetty:
@jmattsson I have downloaded the firmware from the link (Thank you!). I flashed it at 0x00000 and  0x10000.

The serial console gives: (In Esplorer)
```
PORT OPEN 9600

Communication with MCU...
Got answer! AutoDetect firmware...

Can't autodetect firmware, because proper answer not received.
ü|vñh¶,�äz¶ØÑI$bzùÀ¤>ÈèÀl 8ñ
```

It is not accepting any commands from the console.

saitejalakkimsetty:
@jmattsson I have flashed the firmware from the build. (Thank you!)
When i try to connect through the serial (using Esplorer), it does throw some gibberish on the console.
```
PORT OPEN 9600

Communication with MCU..Got answer! AutoDetect firmware...

Can't autodetect firmware, because proper answer not received.
Eü|vñ¬l´øl B³èøì¤D:ìTÏÍ�À$Zø
```
But it does not continuously reboot.

Issue #442, opened 2015-05-25
Check restart reason


vudangngoc:
I think it's good to know why NodeMCU restart, this information will help developer handle behavior of NodeMCU more easily, avoid causes make NodeMCU restart and improve reliable of ESP.

dperlman:
My understanding is that nodeMCU restarts because the watchdog timer times out. This means either that user code entered a blocking loop which didn't allow the timer to reset; or else the firmware was rendered non-functional by memory corruption resulting from heap overflow. In either case, there's no practical way for the firmware to track the cause of its own failure. However, if you watch the serial console output, there is almost always an error message before the restart and you can figure out what happened from that.

vudangngoc:
Can we determine restart cause is power cutting or watchdog time out?

cal101:
Moin,

watchdog resets will print "wdt reset".

Cal

drv53b6:
node.restart(restart_code)
print(node.rcode())
or
print(rcode) rcode=nil

TerryE:
@drv53b6 -- except that `node.restart()` doesn't take any parameters and the **node** library doesn't contain a  `node.rcode()` method.

drv53b6:
@TerryE must do it, it will be very useful

vudangngoc:
@cal101 : where it print "wdt reset"? any news about this issue?

jmattsson:
There is also a ROM function called rtc_get_reset_reason() which returns the reset reason. 4 is the code for wdt reset, 3 for reset via the reset pin. I can't remember what the code for a cold boot is off the top of my head.

vudangngoc:
@jmattsson : could you please give me a piece of example code?

MarsTechHAN:
@vudangngoc When ESP8266 restart it will print information at 74880bps

Air000:
@jmattsson where to find the definitions of rtc_get_reset_reason()'s return value?

jorgempy:
this should be closed, as #512 has been merged. also it should be documented somewhere.

PR #512, opened 2015-06-27
Expose boot reason in Lua

jmattsson:
Since someone was asking for this over in #442 and I happen to know how to get it, here is a small patch to provide a node.bootreason() function.


vudangngoc:
thank you very much ^_^

smartynov:
@jmattsson thanks a lot! I came across this merge when investigating causes of my nodemcu app restarts. I think this feature requires a bit of documentation: could you please write all the possible return values and their meaning?

Issue #494, opened 2015-06-22
Filling & Fixing the document  [Here is a log]


MarsTechHAN:
<h3> If you find any bug inside the document, comment at here </h3> 
<h4>Now we are taking step fixing the</h4>
</br>
We will list them up here

- [ ] Add Crypto Modules
- [ ] Add RC Modules
- [x] Add DHT Modules

- [ ] Add 'hidden' function to wiki(like file.fsinfo()...those undoumented function)
- [ ] Fix the bug inside the existance Document

1.If some one comment a issue or unlist function in the NodeMCU, I will add them into the waiting list.
2.Everybody can go finishing the issue on the list, and after you have finished, comment that you have done some part of waiting list.

marcelstoer:
Once you're done you can close #486.

TerryE:
@MarsTechHAN, I am not quite sure what you mean by the title.  Are you asking for corrections to the user [API document](/nodemcu/nodemcu-firmware/wiki/nodemcu_api_en) in the github wiki?  I am sure that a few of the community would be happy to support you here, but we just need to understand what you asking for.  Regards Terry

MarsTechHAN:
@TerryE Yes, nowaday, we really need to take out some time to updata the document after serval month's developing. Because the API document is really a mess... many powerful function are un-document, and many of the demo program are wrong. That will really consume new developer's a lot of time, which is what everybody dont want to see~

So...This page will work like this->
1.If some one comment a issue or unlist function in the NodeMCU, I will add them into the waiting list.
2.Everybody can go finishing the issue on the list, and after you have finished, comment that you have done some part of waiting list.

TerryE:
@MarsTechHAN, Some of the examples are really not good examples of how to program applications in Lua, and we have lots of overlaps eg.in fragment.lua with other examples.  A big job :worried:

MarsTechHAN:
@TerryE Yeah... That's the reason why we need to make a to-do-list

vowstar:
Yes, very big job.
so much examples need fix, we  just finished a little.

marcelstoer:
Documentation is now tracked in other issues.

PR #513, opened 2015-06-27
update 433MHz transmission comment, add static ICACHE_FLASH_ATTR flag

mikewen:
update 433MHz transmission comment on how to use rc module in Lua,
added static ICACHE_FLASH_ATTR flag (Not required, but good for consistence)


MarsTechHAN:
#494 Can you add the RC module's document, and add some example on NodeMCU's wiki?
```
https://github.com/nodemcu/nodemcu-firmware/wiki/nodemcu_api_en
```

Issue #517, opened 2015-06-29
RTC accuracy


patrickhwood:
I've run some experiments with the RTC (tmr.time()) and found that it's not very accurate when the device is active (most of my tests have been with the dht driver).  It drifts by up to 10 seconds per hour in some cases.

I've also found that the internal usec clock, system_get_time(), is far more accurate (at least on one test board; I will try others soon), presumably because it's connected directly to a divider on the xtal.  I replaced the code for tmr.time() to use this clock, compensating for when it rolls over every ~4000 sec, and found it drifted ~2 seconds over 10 hours, when compared against an NTP synchronized linux system.

Has anyone done any real tests of the various clocks available in the SDK?

marcelstoer:
Maybe PR #519 fixes this?

md5crypt:
ok, there is a problem with my code, value of system_rtc_clock_cali_proc is not constant over time. So the error does not accumulate but the time looks like this:
```
tmr.alarm(0,1000,1,function() print(tmr.time()) end)
570
577
573
574
576
```
I thinking of applying a low pass or median filter on it, but the question is how much memory can I waste on the filter? More memory -> more stable time

patrickhwood:
I would suggest using system_get_time() instead.  It is very stable (~2 sec over 10 hours on my board), and doesn't require normalization with system_rtc_clock_cali_proc.  It probably runs off the same xtal input as the RTC anyway.

md5crypt:
no, from what I read the RTC clock is a separate low-power circuit (probably a low-frequency RLC oscillator, that would explain the need of calibration) and it keeps on running even in deep sleep while the system time doesn't. I propose:
* tmr.systime() returns system_get_time()
* tmr.setrtcfactor(low,high) where low is a 31 bit fractional part and high 31 bit decimal part
* tmr.rtctime() returns rtc * factor

patrickhwood:
An RLC is probably going to drift with temperature changes, and maybe with voltage drift, depending on the quality of the regulator.  You're right about deep sleep mode, though, which is unfortunate.

Maybe by adding an optional byte string parameter to tmr.time(), it could be periodically synched from an NTP source's 64-bit timestamp.

saitejalakkimsetty:
@patrickhwood Where is the `system_get_time()` located and how do I use it?

md5crypt:
[...]

Issue #516, opened 2015-06-29
CI and automated testing


marcelstoer:
I know core-contributors are very busy (SDK changes, docs, etc.). Still, I believe at some point this project should invest in CI and automated testing.

I'm eagerly waiting for the [WiPy MicroPython boards from Kickstarter](https://www.kickstarter.com/projects/wipy/the-wipy-the-internet-of-things-taken-to-the-next). They just recently described how they test.

> Rock solid software
> 
> We just finished building a professional integration test setup for the WiPy. This setup consists of a Linux box (a Raspberry Pi 2B in this case) running Jenkins and is connected to a WiPy via USB and also via WiFi. Every time a new software change is pushed to the MicroPython repository the following happens:
> - The latest revision of the software is checked out and built.
> - The WiPy is updated via FTP with the recently built software.
> - The MicroPython test suite is executed on the WiPy.
> - In case of any failure, tests results are emailed to the developers.
>   ... and we are busy right now writing specific tests for all the WiPy peripherals. Rock solid software

source: https://www.kickstarter.com/projects/wipy/the-wipy-the-internet-of-things-taken-to-the-next/posts/1277814

IMO that's the way to go (i.e. that's how I develop software in the non-embedded world).

MarsTechHAN:
I guess in the future, NodeMCU Dev Kit V2, we will enable some dramtic function (like modfible on-board automatic test system).
For now... I guess we can use a computer(like respe pie) with script to deal with this.
But how to co-wroking with Travis CI, do you have some idea?

marcelstoer:
> Travis CI is a wonderful system

Oh yes, truly marvelous!

> But how to co-wroking with Travis CI, do you have some idea?

My initial ideas....

The current build has one major flaw in my eyes: the binaries built by Travis are not use for any further processing, they just sit there until Travis dumps the VM it created to run the build.

So, the first step really would be to get the binaries off that Travis VM. In the `after_success` phase you could:
- `curl` them to a location of your choice
- `FTP` them to a location of your choice
- use the [Amazon S3 deployer](http://docs.travis-ci.com/user/deployment/s3/) provided by Travis
- use the wonderful [GitHub API](https://developer.github.com/v3/) to push the binaries back to your GitHub repo
[...]

PR #519, opened 2015-06-29
rewrote the timer module from scratch

md5crypt:
#### new timer module
* rewritten from scratch
* 6 new api calls
* old api still works
* better rtc precision (arithmetic errors do not accumulate)
* software watchdog
* clean code

#### api doc:
```
tmr.wdclr() -- not changed
tmr.now()   -- not changed
tmr.time()  -- not changed
tmr.delay() -- not changed
tmr.alarm() -- not changed
tmr.stop()  -- changed, see below. use tmr.unregister for old functionality

tmr.register(id, interval, mode, function)
	bind function with timer and set the interval in ms
	the mode can be:
		tmr.ALARM_SINGLE for a single run alarm
		tmr.ALARM_SEMI for a multiple single run alarm
		tmr.ALARM_AUTO for a repating alarm
	tmr.register does NOT start the timer
	tmr.alarm is a tmr.register & tmr.start macro
tmr.unregister(id)
	stop alarm, unbind function and clean up memory
	not needed for ALARM_SINGLE, as it unregisters itself
tmr.start(id)
	ret: bool
	start a alarm, returns true on success
tmr.stop(id)
	ret: bool
	stops a alarm, returns true on success
	this call dose not free any memory, to do so use tmr.unregister
	stopped alarms can be started with start
tmr.interval(id, interval)
	set alarm interval, running alarm will be restarted
tmr.state(id)
	ret: (bool, int) or nil
	returns alarm status (true=started/false=stopped) and mode
	nil if timer is unregistered
tmr.softwd(int)
	set a negative value to stop the timer
	any other value starts the timer, when the
	countdown reaches zero, the device restarts
	the timer units are seconds
```

and why, why, why are compilation warnings off? I had to #pragma them at the top of the file

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -1,232 +1,332 @@
-// Module for interfacing with timer
+/*guys, srsly, turn on warnings in the makefile*/
+#if defined(__GNUC__)
+#pragma GCC diagnostic warning "-Wall"
+#pragma GCC diagnostic warning "-Wextra"
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#endif
 
-//#include "lua.h"
+/*-------------------------------------
+NEW TIMER API
+---------------------------------------
+
+tmr.wdclr() -- not changed
+tmr.now()   -- not changed
+tmr.time()  -- not changed
+tmr.delay() -- not changed
+tmr.alarm() -- not changed
+tmr.stop()  -- changed, see below. use tmr.unregister for old functionality
+
+tmr.register(id, interval, mode, function)
+	bind function with timer and set the intercal in ms
+	the mode can be:
+		tmr.ALARM_SINGLE for a single run alarm
+		tmr.ALARM_SEMI for a multiple single run alarm
+		tmr.ALARM_AUTO for a repating alarm
+	tmr.register does NOT start the timer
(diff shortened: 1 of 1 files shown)


app/modules/tmr.c
```diff
@@ -1,232 +1,358 @@
-// Module for interfacing with timer
+/*guys, srsly, turn on warnings in the makefile*/
```
creationix: +1

TerryE:
This issue of not having tighter compiler checking (e.g. `-Wall -Wextra -Wunused-parameter`) by default  and error-free makes is a excellent one and one where we the community could contribute.

However given that this @md5crypt version is a complete rewrite with potentially its own set of bugs and possible subtle changes to the the existing functionality, I would suggest that we should defer merging it into dev until at least one other developer has done a complete code review and validated testing.

md5crypt:
well, I was shocked that it just got merged without any discussion whatsoever ^^
As I already wrote in https://github.com/nodemcu/nodemcu-firmware/issues/517 tmr.time() is broken, but that's the esp's api fault. I think a per-chip long-run RTC clock factor calculation is needed.

marcelstoer:
> I was shocked that it just got merged without any discussion whatsoever

+1

It made me create #516.

PR #522, opened 2015-07-01
Added WS2801 support

kbeckmann:
I wrote a module to support the ws2801 LED driver. It's not that commonly used anymore but I guess people still have some old LED-strips that they would still like to use (I do which is why I wrote the code).

The api-difference compared with the ws2812 driver is that one has to call ws2801.init() and that ws2801.write() takes RGB-strings.

Please let me know if I have to change anything.


klukonin:
COOL!

But I think that such modules should be additional and not included by default.
Just IMHO.

kbeckmann:
I agree with you! Feel free to disable it as a default.

Off topic, but is there a backlog with stuff to do? I currently have a lot of free time and would love to contribute more.

klukonin:
Shure.
//offtop 
I'm afraid this project does not have a TODO list.
For this moment a lot of people wait for on-the-air update.
Current bootloader is able to do this.
So if you want to contribute more, it's a great case =)

marcelstoer:
> I agree with you! Feel free to disable it as a default.

As it's already been merged please create a new PR to correct this. I'll later add the new module to my [cloud build](http://frightanic.com/nodemcu-custom-build/).

> Off topic, but is there a backlog with stuff to do? I currently have a lot of free time and would love to contribute more.

Great! The [issue list](https://github.com/nodemcu/nodemcu-firmware/issues) is quite full. Since you seem to be familiar with writing drivers I suggest you look into fixing #394, adding support for the popular MAX7219 8x8 LED matrix panels.

devsaurus:
> Off topic, but is there a backlog with stuff to do? I currently have a lot of free time and would love to contribute more.

[U8glib](https://github.com/olikraus/u8glib/wiki) comes with a whole bunch of [display drivers](https://github.com/olikraus/u8glib/wiki/device). Just a few of them already have Lua bindings in nodemcu. Adding more would be nice - though verifying them requires the actual hardware.

PR #531, opened 2015-07-04
ws2812: implement brightness for ws2812,keep buffer for rgb

makefu:
brightness may between 0 and 1 and will be applied with the next ws2812.write
ws2812.write now takes R,G,B instead of grb (was undocumumented before) by
allocating a transfer buffer and freeing it afterwards.

ws2812.writergb now becomes obsolete with it but currently remains in the
module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,12 +440,18 @@ In contrast to the source code based inclusion of XBMs into u8glib, it's require
 
 ####Control a WS2812 based light strip
 ```lua
+  
 	-- set the color of one LED on GPIO2 to red
-	ws2812.writergb(4, string.char(255, 0, 0))
-	-- set the color of 10 LEDs on GPIO0 to blue
-	ws2812.writergb(3, string.char(0, 0, 255):rep(10))
-	-- first LED green, second LED white
-	ws2812.writergb(4, string.char(0, 255, 0, 255, 255, 255))
+	ws2812.write(4, string.char(255, 0, 0))
+  -- set the brightness to 50%
+  ws2812.set_brightness(0.5)
+	-- set the color of 10 LEDs on GPIO0 to blue with half brightness
+	ws2812.write(3, string.char(0, 0, 255):rep(10))
+  --returns the current brightness
+  print(ws2812.brightness())
+  0.5
+	-- first LED green, second LED white at half brightness
+	ws2812.write(4, string.char(0, 255, 0, 255, 255, 255))
 ```
 
 ####coap client and server
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -5,6 +5,15 @@
 #include "lrotable.h"
 #include "c_stdlib.h"
 #include "c_string.h"
+#include "c_math.h"
+
+#include "c_stdio.h"
+#include "c_stdlib.h"
+#include "c_stdarg.h"
+#include "c_string.h"
+
+#include "strbuf.h"
+
 /**
  * All this code is mostly from http://www.esp8266.com/viewtopic.php?f=21&t=1143&sid=a620a377672cfe9f666d672398415fcb
  * from user Markus Gritsch.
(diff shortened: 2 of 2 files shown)


jiegec:
But。。。Is the floating support always open?

makefu:
you are right, it will not work with the integer firmware. The transfer buffer will work on both versions but i just do not care enough about INT firmware to make the other parts work ...
`hsv2rgb` will be a huge hassle to transfer, brightness as well.

It may be a good idea to think about a  module management so i can maintain these changes and provide users a way to install it easily.

marcelstoer:
> you are right, it will not work with the integer firmware.

I don't think this is a big issue - as long as the docs are very clear about this limitation.

TerryE:
Surely if this module doesn't work with integer builds, then it should include the necessary preprocessor conditionals to generate `#error` statement on integer builds then the docs on this issue would be the code.

creationix:
FWIW, I use integer builds with ws2812 all the time.  I have an integer
based HSL to RBG conversion I wrote the other day for another project.  I
can share the logic if desired.  It obviously requires larger input
ranges.  I would probably do 0-767 for hue (mod 768) and 0-511 for
lightness. for 24-bit RGB (8x8x8) outputs.

klukonin:
@creationix, That's cool. Please, share your logic =)
integer firmware is more popular for this moment.

mikewen:
+1 for integer logic for ws2812. Integer firmware just make so much sense.

creationix:
It's for a VM I'm designing, The syntax is basically lisp without parens (since all opcodes have fixed number of arguments).  It should be trivial to port to C.  All math assumes integers only https://gist.github.com/creationix/88c159db3c97a170ddec

creationix:
The numbers would probably need to be made smaller.  I can port it to C and change the ranges to match the 24-bit color used by ws2812 if desired.

creationix:
I wrote a hcl conversion for integer lua today.

```lua
function hcl(h, c, l)
  h = h % 768
  local x = c * (128 - math.abs((h % 256) - 128)) / 128
  local r, g, b
  if h < 128 then
    r, g, b = c, x, 0
  elseif h < 256 then
    r, g, b = x, c, 0
  elseif h < 384 then
    r, g, b = 0, c, x
  elseif h < 512 then
    r, g, b = 0, x, c
  elseif h < 640 then
    r, g, b = x, 0, c
  else
    r, g, b = c, 0, x
  end
  local m = l - (77 * r + 151 * g + 28 * b) / 256
  r, g, b = r + m, g + m, b + m
  return 
      g < 0 and 0 or g > 255 and 255 or g,
      r < 0 and 0 or r > 255 and 255 or r,
      b < 0 and 0 or b > 255 and 255 or b
end
```

It accepts hue in the range 0-767, chroma roughly in the range 0-255 and luma in the range 0-255.  If has clamps at the end since it's gamut is much wider than 8-bit rgb.  This color model is much better than HSL or HSI for color accuracy, but harder on the math and less convenient with the mismatching gamuts.

It's rather slow in lua and especially in float mode.  Porting to C would help a lot.

I'm able to get it fast enough by flashing the integer version of nodemcu and overclocking to 160Mhz.

```lua
node.setcpufreq(node.CPU160MHZ)
h = 1
local pixels = {}
tmr.alarm(0, 66, 1, function ()
  h = (h + 1) % 768
  for x = 1, 8 do
    local b = math.abs(((h + x) % 16) - 8) * 5 - 60
    for y = 1, 8 do
      pixels[x + y * 8 - 8] = string.char(hcl(h, 70, b + y * 10))
    end
  end
  ws2812.write(8, table.concat(pixels))
end)
```

kbeckmann:
@creationix I'd happily help out with a c module port. Which APIs are desirable?
hcl.to_rgb(string_with_hcl_data) -- returns a string with rgb data
hcl.to_rgb(h, c, l) -- returns rgb values just like your function.
Is hcl.from_rgb() interesting at all or should we just skip it?

felixcameron:
Hcl from grb is very useful for me... Great for communication with life
lights!!! Please implement it :)

felixcameron:
*useful for lifx lights

dirkmb:
I also did some work with keeping a buffer for the ws2812. its in https://github.com/dirkmb/nodemcu-firmware I'll merge makefu's and my changes in the next days and share the result.

marcelstoer:
@makefu Do you plan to address the issues mentioned in the comments and to resolve the merge conflicts? If not then please close it.

makefu:
@marcelstoer i do not know what the status is regarding @dirkmb. i was hoping the merge will happen there :)
i am currently not interested in merging into upstream due to lack of interest. maybe i will look into sometime in the future.
Keep up the good work :+1:

dirkmb:
I have not done anything since the chaos communication camp. I'll open a merge request if I found the time to clean up the code and completely integrate makefu's changes.

PR #542, opened 2015-07-09
Add func for verbose info in node module

jiegec:
It shows definition and modules included in this build.

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -24,6 +24,8 @@
 #include "flash_api.h"
 #include "flash_fs.h"
 #include "user_version.h"
+#include "user_config.h"
+#include "user_modules.h"
 
 #define CPU80MHZ 80
 #define CPU160MHZ 160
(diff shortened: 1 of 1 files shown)


app/modules/node.c
```diff
@@ -93,6 +95,116 @@ static int node_info( lua_State* L )
...
+#ifdef LUA_NUMBER_INTEGRAL
+                   "number integral enabled "
+#else
+                   "number integral disabled " 
+#endif
+
+#ifdef LUA_OPTRAM
+                   "optimize memory enabled "
+#else 
+                   "otiimize memory disabled "
```
kbeckmann: typo

kbeckmann:
Nice. People who implement new modules need to know that they have to add their module to this list as well.

nodemcu:
I am not sure about this PR, this can tell whether some options are enabled, but for modules, seems lua has a global table showing module names?

marcelstoer:
> People who implement new modules need to know that they have to add their module to this list as well.

Can't this be detected at runtime somehow? Anything manual is error-prone. It'd be yet another place a new module needs to be "registered".

jmattsson:
I concur. Module names should be automatically detected. They're available in the static variable `lualibs` in modules/linit.c, so look at pulling them out from there.

nodemcu:
```lua
for k,v in pairs(_G.package.loaded) do print(k) end
```
this can print out module names

jmattsson:
That only shows Lua modules loaded via require(), not the compiled-in modules though?

nodemcu:
really? I should have test this first. sorry.

TerryE:
The way that this info routine lists off loaded ROM table denormalises this meta data and therefore creates an unnecessarymaintenance headache.  Would it just be a lot easier to expose the lualibs table in `init.c` via a `luaL_Reg *lib LuaL_getlualibs (lua_State *L)` and the we can loop over this list.  The first entry in each row is the char * of the module name.

That way maintainers don't even need to know about this info routine.

Issue #534, opened 2015-07-06
dev: ws2812 O2 optimize turns first led green


makefu:
with the addition of `__attribute__((optimize("O2")))` the ws2812 library does not work reliably anymore, the first LED is always green for me.

https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/ws2812.c#L19

I tried building the firmware without the attribute flag and the behavior is correct again

creationix:
Duplicate of #487.  Thanks for helping find the cause.

MarsTechHAN:
@makefu 
@creationix 
Maybe the version cause the problem. Can you guys show your NodeMCU's Version
(Like clone from which version which branch and using which kind of board)

makefu:
I tried most current 'dev' branch (as of 2015-07-05)  with the new `__attribute__((optimize("O2")))` flag in the `ws2812.c` file.
This creates the issue.
When i remove the attribute tag from the file the problem goes away.

DR-AU:
I use the pre-built firmware and from NodeMCU 0.9.6 build 20150625 the problem exists

TerryE:
Closed as duplicate of #487.

Issue #487, opened 2015-06-18
Neopixel library no longer works properly.


creationix:
Some time recently (not sure, maybe a few couple weeks ago), the dev or dev096 branch (and now master) stopped working for my neopixels (ws2812).

It mostly works, but the first pixel will light up bright green every other call to `ws2812:write` or `ws2812:writergb`.

I have to work around this by sending just the first pixel after sending a full refresh.  Sometimes it still flickers visibly because the two lua calls are slow enough.

Is anyone else seeing this?

creationix:
At first I thought I broke my pixels, but I've tried several different strips and they all act the same.

vowstar:
I think I'll buy some ws2812 for testing ... :D

creationix:
@vowstar for what it's worth, I'm testing with two different kinds.

The newer ones like http://www.adafruit.com/products/1487 seem to have a slightly different protocol than the older ones like http://www.adafruit.com/products/1734 (rgb vs grb ordering)

jorgempy:
i am seeing this as well, it is very strange.

MarsTechHAN:
We try to make the program work : )
Wish you can sucessfully remote your WS2812

creationix:
I think this may be related to printing the `>` on the serial console.
When using the serial console to send commands if I prefix the line with a single `"\0\0\0"` write and then do the real write, the colors are correct.  I think the first tiny write will reset whatever bad state the neopixel is in.

But if I do this in two lines, it doesn't help.  It has to be one command.

This will result in the first pixel being wrong with 255 for the first byte.
```
> ws2812.write(3, "\0\0\0")
> ws2812.write(3, "\1\2\3\4\5\6")
```

But this will work
```
ws2812.write(3, "\0\0\0")ws2812.write(3, "\1\2\3\4\5\6")
```

kbeckmann:
Has anyone else tried the new code? Feel free to close this issue if the issue is solved.

creationix:
I'm having a hard time getting this to work with my ws2811 neopixels.  They worked before with the old ws2812 code except red and green were swapped compared to my ws2812 neopixels. (This made the `ws2812.write` function very convenient)

creationix:
Hmm, not working with my ws2812 neopixels either (the adafruit 8x8 array).  Am I doing something wrong.  I'm testing with the nodemcu devboard 1.0 flashed in dio mode with latest esptool.py.

creationix:
[...]

PR #548, opened 2015-07-12
Improved ws2812 implementation

kbeckmann:
- Fixes the "first led problem" (#534 #487).
- Adds support for 160MHz mode.
- Optimized timings. Time to write pixels is reduced by about 25%
  This is useful when having a large amount of chained LEDs (>500).

Please review the code and come with comments before merging it. I have tested the code a lot and it works for me in both 80 and 160MHz mode. But it would be great if somebody else could verify it.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -5,27 +5,55 @@
 #include "lrotable.h"
 #include "c_stdlib.h"
 #include "c_string.h"
-/**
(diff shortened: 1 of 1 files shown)


creationix:
I love the new code.  Nice work.

robertfoss:
Thank dood! Excellent stuff!

kbeckmann:
Happy to contribute :smile:

PR #550, opened 2015-07-13
Remove pre_build directory containing very old binaries

marcelstoer:
I feel that keeping that directory with the very old binaries around only leads to confusion. I fail to see how those artifacts could still be relevant and I suggest you remove `pre_build` from all branches (can create a PR if you like).


marcelstoer:
The second commit shouldn't have gone into this branch/PR. I'll prepare a new PR.

PR #560, opened 2015-07-17
Remove pre_build directory containing very old binaries

marcelstoer:
I feel that keeping that directory with the very old binaries around only leads to confusion. I fail to see how those artifacts could still be relevant and I suggest you remove pre_build from all branches (can create a PR if you like).


zerog2k:
+1

dnc40085:
+1

marcelstoer:
Thanks, but the request was to delete the `pre_build` folder from all branches. Are you going to merge that any time soon or shall I create individual PRs as offered?

Issue #563, opened 2015-07-17
dev120: depending on number of selected modules build fails


marcelstoer:
Not to be impatient but just to keep track of bugs...

Depending on the (number of) modules you select in `user_modules.h` the dev120 build may fail with 

>  `section `.text' will not fit in region `iram1_0_seg'

``` 
xtensa-lx106-elf-gcc  -L../lib -Wl,--gc-sections -Xlinker -Map=mapfile -nostdlib -T../ld/eagle.app.v6.ld -Wl,--no-check-sections -Wl,--wrap=_xtos_set_exception_handler -Wl,-static -Wl,--start-group -lc -lgcc -lhal -lphy -lpp -lnet80211 -llwip -lwpa -lmain -ljson -lsmartconfig -lssl user/.output/eagle/debug/lib/libuser.a driver/.output/eagle/debug/lib/libdriver.a json/.output/eagle/debug/lib/libjson.a upgrade/.output/eagle/debug/lib/libupgrade.a platform/.output/eagle/debug/lib/libplatform.a libc/.output/eagle/debug/lib/liblibc.a lua/.output/eagle/debug/lib/liblua.a coap/.output/eagle/debug/lib/coap.a mqtt/.output/eagle/debug/lib/mqtt.a u8glib/.output/eagle/debug/lib/u8glib.a smart/.output/eagle/debug/lib/smart.a wofs/.output/eagle/debug/lib/wofs.a spiffs/.output/eagle/debug/lib/spiffs.a cjson/.output/eagle/debug/lib/libcjson.a crypto/.output/eagle/debug/lib/libcrypto.a dhtlib/.output/eagle/debug/lib/libdhtlib.a modules/.output/eagle/debug/lib/libmodules.a -Wl,--end-group -lm -o .output/eagle/debug/image/eagle.app.v6.out 
/esp-open-sdk/xtensa-lx106-elf/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld: .output/eagle/debug/image/eagle.app.v6.out section `.text' will not fit in region `iram1_0_seg'
collect2: error: ld returned 1 exit status
make[1]: *** [.output/eagle/debug/image/eagle.app.v6.out] Error 1
make[1]: Leaving directory `app'
make: *** [.subdirs] Error 2
```
OK example: node,file,gpio,wifi,net,pwm,tmr,adc,uart
[...]

PR #573, opened 2015-07-22
Dev120 need to be able build all closed modules

3dot3:
If open a module like CJSON, the build will failed currently.


MarsTechHAN:
Because dev120 use many good features which could provide enormous of RAM and extra function. Some function are dead because of some compatibility issue. If you someone can deal with them...That will be wonderful.

marcelstoer:
I reported this in issue #563.

georgkreimer:
Running this revision on Adafruits HUZZAH I get this:
screen shot 2015-07-31 at 10 46 43 am
Same also happens with my ESP-01, ESP-12, ESP-12E modules from other sources.

Issue #359, opened 2015-04-14
MQTT connection closes when not publishing


FrankX0:
I am playing around with mqtt on 0.9.6-dev_20150406.
It seems that when I only subscribe and not publish, the connection will be closed after the keepalive time expires. When I do publish, the connection is kept alive.

Is this correct behaviour?

zerog2k:
+1
I also have this problem with NodeMCU 0.9.6 build 20150704.
This seems to be a bug.
Also, one could have a timer task which reconnects, however I dont see any mqtt client methods for checking if the client is connected or not, so not sure how to handle this issue.

zerog2k:
A quick look at the mqtt code appears that mud->keep_alive_tick is the counter which needs to be reset.
I see it's getting reset in several places:

mqtt_socket_received
https://github.com/nodemcu/nodemcu-firmware/blob/dceed526c90aaa63a4ec2aaf11d8ce313f40c6ff/app/modules/mqtt.c#L394

mqtt_socket_publish
https://github.com/nodemcu/nodemcu-firmware/blob/dceed526c90aaa63a4ec2aaf11d8ce313f40c6ff/app/modules/mqtt.c#L1215

mqtt_socket_sent
https://github.com/nodemcu/nodemcu-firmware/blob/dceed526c90aaa63a4ec2aaf11d8ce313f40c6ff/app/modules/mqtt.c#L413

I'm suspecting that mud->keep_alive_tick may be missing getting reset to zero in mqtt_socket_received.
I'm trying to build a debug version of nodemcu to load and try this out...

zerog2k:
@marcelstoer 
Hi Marcel,
I'm trying to build a debug version of nodemcu to troubleshoot this issue, but so far, I'm unable to get a working build. I was thinking about using your wonderful hosted build:
http://frightanic.com/nodemcu-custom-build/

Can you add an option to make a debug build? I assume it would be uncommenting "#define DEVELOP_VERSION" in app/include/user_config.h

zerog2k:
Ok I think the reason why DEVELOP_VERSION was not working is that it also does some other stuff which I dont want like setting uart port speed to 78800 instead of 9600.

zerog2k:
Notes (mainly to self ;):
just building a version with NODE_DEBUG (and dev120 with mqtt module enabled) worked for me.
I'll post this here for later analysis of what might be happening:
Here is the lua script:
[...]

PR #571, opened 2015-07-21
Switch to Travis containerized builds

marcelstoer:
Containerized Travis CI builds have many advantages over traditional builds: http://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F

One of them is caching which will soon include caching of APT packages by default (not yet enabled): http://docs.travis-ci.com/user/caching/#Caching-Ubuntu-packages

And again, as with PR #560, I ask you: where should this PR go? dev, master...? Do you want a separate PR for every branch or are you going to merge it to all of them yourself?

PR #551, opened 2015-07-14
wifi station connection event monitor

dnc40085:
The event monitor watches the station's connection state, executing a user callback when the connection state changes.

If anybody has any suggestions for changes(function name, default interval, etc.) or any questions please feel free to comment comments.

CONTRIBUTORS: Before merging, please allow a few days for others to comment.

This is also a fix for issue #173 

Here are the functions to be added:

#### Description
  Register callback for wifi station status event
#### Syntax
  wifi.sta.eventMonReg(wifi_status, function)
  wifi.sta.eventMonReg(wifi.status, "unreg")
#### Parameters
- wifi_status: wifi status you would like to set callback for
  - Valid wifi states:
    - wifi.STA_IDLE
    - wifi.STA_CONNECTING
    - wifi.STA_WRONGPWD
    - wifi.STA_APNOTFOUND
    - wifi.STA_FAIL
    - wifi.STA_GOTIP
- function: function to perform when event occurs
- "unreg": unregister previously registered callback

#### Returns
Nothing.
#### Example
```lua 
  --register callback
  wifi.sta.eventMonReg(wifi.STA_IDLE, function() print("STATION_IDLE") end)
  wifi.sta.eventMonReg(wifi.STA_CONNECTING, function() print("STATION_CONNECTING") end)
  wifi.sta.eventMonReg(wifi.STA_WRONGPWD, function() print("STATION_WRONG_PASSWORD") end)
  wifi.sta.eventMonReg(wifi.STA_APNOTFOUND, function() print("STATION_NO_AP_FOUND") end)
  wifi.sta.eventMonReg(wifi.STA_FAIL, function() print("STATION_CONNECT_FAIL") end)
  wifi.sta.eventMonReg(wifi.STA_GOTIP, function() print("STATION_GOT_IP") end)
  
  --unregister callback
  wifi.sta.eventMonReg(0, "unreg")
```

#### Description
Start wifi station event monitor
#### Syntax
 wifi.sta.eventMonStart()
 wifi.sta.eventMonStart(ms)

### Parameters
- ms: interval between checks in milliseconds. defaults to 150 ms if not provided.

#### Returns
  Nothing.

#### Example
```lua
  --start wifi event monitor with default interval
  wifi.sta.eventMonStart()

  --start wifi event monitor with 100 ms interval
  wifi.sta.eventMonStart(100)
```

#### Description
Stop wifi station event monitor
#### Syntax
wifi.sta.eventMonStop()
wifi.sta.eventMonStop("unreg all")

[...]

Issue #449, opened 2015-05-28
Creating a nodeMCU Lua (unofficial) FAQ


TerryE:
_Forgive this post as an issue but I wanted feedback and comment from this audience which only has part overlap with the equivalent audience to this [topic on esp8266.com](http://www.esp8266.com/viewtopic.php?f=24&t=3311&p=18770)._

This is really by way of an RFC.  I've started drafting an unofficial FAQ on a local MediaWiki instance.  It's only about 30% complete and I am still in rough brain-dump mode.  It still needs proper proofing and polishing, but I am maintaining a copy on the esp8266.com wiki: [nodeMCU Unofficial FAQ](http://www.esp8266.com/wiki/doku.php?id=nodemcu-unofficial-faq)  for those that might want to take a skim.   I don't think that an ultimate master should belong here so at some point I will can it , if and when we transfer the then content.

My inspiration for this is the incredibly useful [Lua Unofficial FAQ \(uFAQ\)](http://www.luafaq.org/).  What I want to put in here are Q&A for all of the questions that new to nodeMCU Lua programmers reqularly trip over and ask on this issues list and on the esp8266.com developer forum.

*    Please use the esp8266.com topic reference above if you have an account there  to discuss / correct the content / suggest extra additions.  Even a list of placeholder questions would be good.
*    Alternatively post here on this issue (as long as the nodeMCU guys are comfortable with this).
*    Does anyone have a better suggestion (i) where we should do initial work, and (ii) where the list should ultimately reside? Should something like this be part of the standard node MCU documentation?

As you will see I am still on a voyage of discovery on how this Lua RTS works, but there's detail in here that I just haven't found anywhere else and have only worked out through re-engineering the source.

If anyone wants to chat / skype / email on specific points then drop me a mail.  My mail domain is as per my website on my profile less the "blog." and my mailbox is just Terry at this.

TerryE:
I've relocated this extended and updated page to the esp8266 site and changed the first post reference.  I won't post to this issue again unless others feel that some version of this content should be pulled into the formal nodeMCU documentation, and if I get no other comments, I'll close this issue in a week or so.

creationix:
Thanks for making this.  I see lots of useful stuff in there.

marcelstoer:
[...]

PR #575, opened 2015-07-24
fix mqtt connection closing when not publishing 

zerog2k:
should not reset keep_alive_tick when receiving as this can prevent
sending keepalive (PINGREQ), which can result in server disconnect.
fixes issue #359

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -65,7 +65,7 @@ typedef struct lmqtt_userdata
   int cb_puback_ref;
   mqtt_state_t  mqtt_state;
   mqtt_connect_info_t connect_info;
-  uint32_t keep_alive_tick;
+  uint16_t keep_alive_tick;
   uint32_t event_timeout;
   uint8_t secure;
   bool connected;     // indicate socket connected, not mqtt prot connected.
@@ -391,7 +391,6 @@ READPACKET:
     else
       espconn_sent( pesp_conn, node->msg.data, node->msg.length );
   }
-  mud->keep_alive_tick = 0;
   NODE_DBG("receive, queue size: %d\n", msg_size(&(mud->mqtt_state.pending_msg_q)));
   NODE_DBG("leave mqtt_socket_received.\n");
   return;
@@ -545,6 +544,7 @@ void mqtt_socket_timer(void *arg)
       }
     }
   }
+  NODE_DBG("keep_alive_tick: %d\n", mud->keep_alive_tick);
   NODE_DBG("leave mqtt_socket_timer.\n");
 }
 


marcelstoer:
Not sure if it should be merged into master rather than dev or dev096?

zerog2k:
hi @marcelstoer 
I tested this on master branch, and it seems working.
Question: I did not see it documented somewhere - what is the point of dev, dev096, dev120.
I assume these are related to the esp sdk somehow, but for example, master branch builds ok against esp sdk 1.2.0 for me (using pfalcon/esp-open-sdk repo)

marcelstoer:
> I did not see it documented somewhere - what is the point of dev, dev096, dev120.

Sigh... You know how many times I asked this question here: #560, #571, https://github.com/nodemcu/nodemcu-firmware/issues/449#issuecomment-109043189, https://github.com/nodemcu/nodemcu-firmware/pull/560#issuecomment-122804550, https://github.com/nodemcu/nodemcu-firmware/pull/551#issuecomment-121160394

zerog2k:
@vowstar 
Hi Vowstar,
Can you please advise which branches *fixes* should (versus new *feature* development)?
Should this all go to dev?
Is there a process which core developers occasionally merge dev->master ?
I'm totally confused with all of the dev, dev096, dev120 branches. I don't see it explained anywhere what they are for, and how they should be used or updated.

Can you please explain for us?

jorgempy:
@zerog2k why the 32bits to 16bits change of the ticker? to save memory?

zerog2k:
Regarding branches questions, I think we have our answers in the discussion on issue #449 
As a simple fix for something that is broken, i think it belongs to master, but also needs to be PR'ed forward to dev (and maybe dev120) as well.

@jorgempy 
The mqtt keep-alive interval is 16-bits per spec:
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Keep_Alive
>The Keep Alive is a time interval measured in seconds. Expressed as a 16-bit word, it is the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one Control Packet and the point it starts sending the next. It is the responsibility of the Client to ensure that the interval between Control Packets being sent does not exceed the Keep Alive value. In the absence of sending any other Control Packets, the Client MUST send a PINGREQ Packet [MQTT-3.1.2-23].

A nice side-effect is to save one byte of ram.

jorgempy:
ok, thanks! +1 to merge to master...

zerog2k:
@vowstar Can this be merged to master?

kri5:
:+1:

jorgempy:
:pray:

PR #581, opened 2015-07-28
Added USE_DNS define.

robertfoss:
USE_DNS is used to flag DNS capability in the DHCP offer message.

If the DNS capability is flagged in the DHCP offer message clients will send DNS messages to the ESP.
This allows for captive portal type functionality where every DNS request can be made to point to the address of the ESP.


jmattsson:
Oooh, this sounds good!

Issue #579, opened 2015-07-27
disable spiffs cache code to free ram+flash


aguaviva:

BesitzeRuf:
?

dnc40085:
@aguaviva Can you please elaborate? Have you disabled the cache on your own? If so, does it cause a increase/decrease performance? How much RAM/flash is freed if the cache is disabled?

aguaviva:
The spiffs comes with a cache, thing that consumes ram and flash. There are a few #defines that can be used to disable it, I tried but it didn't work. Writing seems to work (wont crash) but reading resets the unit I didn't look much indeep. Will provide more details (going to bed now)

TerryE:
@aguaviva, uncached writing to SPI flash is strongly discouraged as the performance is terrible without the cache and the life of the Flash memory severely degraded.  This is not an option that will get support from the community.  There are a lot better ways to save memory, and we're in the process of implementing most of these.

aguaviva:
@TerryE As always... it depends! you may need the cache others may need more RAM. As everything, it should be configurable and we should let the user choose whatever fit better their needs.

TerryE:
> As everything, it should be configurable and we should let the user choose whatever fit better their needs. 

As I said, I don't think that this will get any support from the dev community.  Do the work yourself and take the performance measures.  It might then get merged.

marcelstoer:
> This is not an option that will get support from the community.

Fully agree.

aguaviva:
TerryE, I am not asking you to do it, I am putting this here so we know we have this option. And if anyone needs it can explore this path and implement it. That you dont like a feature doesnt mean others won't like it. Just dont implement it, others or myself might do it. Also any idea that can be ifdef'ed should be welcome.

PR #586, opened 2015-07-31
Reclaim ~1.5k RAM by tuning down SPIFFS cache, use dynamic DNS table names

jmattsson:
Here are a couple of things that might be of interest:

  1. I tuned the SPIFFS cache down a bit, freeing up half a K of RAM in the process. I also resolved the issue preventing the cache from being disabled, and made disabling the cache optional (as requested in #579). Disabling the SPIFFS cache is really only for those of us who have a read-only scenario and need as much RAM as we can get. This patch leaves the cache *enabled*.
  1. I switched the DNS table over to use dynamically allocated storage for the dns names. These buffers were taking up a whopping 1k of RAM, most of it which would never be used.

Both of these items have only received a basic level of testing here so far, but we're hoping the community might hammer this and point out any issues :)

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -47,6 +47,8 @@
 // #define BUILD_WOFS		1
 #define BUILD_SPIFFS	1
 
+#define SPIFFS_CACHE 1
+
 // #define LUA_NUMBER_INTEGRAL
 
 #define LUA_OPTRAM
--- a/app/lwip/core/dns.c
+++ b/app/lwip/core/dns.c
@@ -173,7 +173,7 @@ struct dns_table_entry {
   u8_t  seqno;
   u8_t  err;
   u32_t ttl;
-  char name[DNS_MAX_NAME_LENGTH];
+  char *name;
   ip_addr_t ipaddr;
   /* pointer to callback on DNS query done */
   dns_found_callback found;
@@ -673,6 +673,8 @@ dns_check_entry(u8_t i)
             if (pEntry->found)
               (*pEntry->found)(pEntry->name, NULL, pEntry->arg);
             /* flush this entry */
+            mem_free (pEntry->name);
+            pEntry->name    = NULL;
             pEntry->state   = DNS_STATE_UNUSED;
             pEntry->found   = NULL;
             break;
@@ -697,6 +699,8 @@ dns_check_entry(u8_t i)
       if (--pEntry->ttl == 0) {
         LWIP_DEBUGF(DNS_DEBUG, ("dns_check_entry: \"%s\": flush\n", pEntry->name));
         /* flush this entry */
+        mem_free (pEntry->name);
+        pEntry->name    = NULL;
         pEntry->state = DNS_STATE_UNUSED;
         pEntry->found = NULL;
       }
@@ -839,6 +843,8 @@ responseerr:
     (*pEntry->found)(pEntry->name, NULL, pEntry->arg);
   }
   /* flush this entry */
+  mem_free (pEntry->name);
+  pEntry->name  = NULL;
   pEntry->state = DNS_STATE_UNUSED;
   pEntry->found = NULL;
 
@@ -898,11 +904,16 @@ dns_enqueue(const char *name, dns_found_callback found, void *callback_arg)
   LWIP_DEBUGF(DNS_DEBUG, ("dns_enqueue: \"%s\": use DNS entry %"U16_F"\n", name, (u16_t)(i)));
 
   /* fill the entry */
+  namelen = LWIP_MIN(strlen(name), DNS_MAX_NAME_LENGTH-1);
+  char *namebuf = (char *)mem_zalloc (namelen);
+  if (!namebuf)
+    return ERR_MEM;
   pEntry->state = DNS_STATE_NEW;
   pEntry->seqno = dns_seqno++;
   pEntry->found = found;
   pEntry->arg   = callback_arg;
-  namelen = LWIP_MIN(strlen(name), DNS_MAX_NAME_LENGTH-1);
+  mem_free (pEntry->name);
+  pEntry->name = namebuf;
   MEMCPY(pEntry->name, name, namelen);
   pEntry->name[namelen] = 0;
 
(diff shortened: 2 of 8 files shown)


kbeckmann:
Awesome! It's great to be able to disable the cache and free some extra RAM for those who need it.

TerryE:
@jmattsson Johny,  the issue that I have with having the cache trim option is that this hits SPIFFS I/O performance.  We already have enough problems with support calls and "issues" from posters whose man issue is that they don't research what they are doing properly and blunder into all sorts of beartraps.  They then raise a bugrep which really says "I don't know what I am doing, so my bug must be a bug with the firmware".  I'll leave you to field the ones relating SPIFFS!! :sunglasses:

aguaviva:
@TerriE The 'I don't know what I am doing' gets solved by documenting the features. I think this one is pretty simple to understand though. As long as we have it on by default we should be newbie-proof. 

In general we shouldn't object to any features that can be enabled/disabled. I am all about having choices, especially when resources are scarce.

aguaviva:
@jmattson That was quick :) Thank you!

jmattsson:
@TerryE When you say "cache trim" do you refer to the fact that I tuned down the total cache size or the fix that makes it possible to disable the cache altogether at compile-time?

Allow me to elaborate somewhat; The cache settings as used in NodeMCU were the example ones straight out of the INTEGRATION doc in the SPIFFS project, which cater to having multiple files open for writing at the same time. NodeMCU on the other hand provides an interface which only allows having one file open for writing at any given time. Write caching is the key issue, I'm sure you can agree with me on. With the tuned down settings there is effectively space for two open files in the cache. The worst case scenario I can think of for NodeMCU is if the user has a file open for read/write, and then does nested dofile() calls. Dofile() is on the other hand is not very dependent on the caching, as it does very large reads. To minimise the potential impact of the reduced cache size for file.read() operations, I changed that from using getc() (which admittedly would have been pretty horrible on an uncached file) to doing block reads and thereby getting rid of a lot of function calling overhead.

When it comes to the option that allows disabling the SPIFFS cache altogether, well, it's a compile time option, and the default is to have the cache *enabled* (for obvious reasons). If you're skilled enough to build your own firmwares and load them, chances are you have some idea what you're doing :)  If people are raising bug reports on such things they should probably be closed with prejudice and a pointer to the forum (wasn't there a discussion about this somewhere recently, anyway?). My point here is that there are legitimate use cases where disabling the cache is a useful thing to be able to do if you're effectively using a read-only file system. I am by no means arguing that this should be the default or even a run-time configurable thing (ick!). And of course, the NodeMCU project is free to accept or reject anything I submit - we're just trying to be good open source citizens and contribute back things which we've found useful.

@aguaviva Contrary to what it may look like, I didn't actually do this for you ;)  We already had this on our to-do list here at work for our current project, and it was pure luck I saw this thread before I'd finished preparing the pull-request. As for debugging the crash, I already had a pretty good idea where to look from when I upgraded SPIFFS to 0.3.2 the other week. In this case it was an unaligned read of the data just after the page header, and that was pretty easy to find with a light sprinkling of printf()s in the flash read/write functions. I've been discussing the various alignment issues with the guy who created SPIFFS, and while he's a busy person hopefully at some point we'll be able to pull in a clean up-stream version without any extra tweaks needed.

aguaviva:
@jmattsson LOL! I know you didn't do it for me :) Thanks for the explanation, i would never have figured out that my crashes were coming from unaligned read. Good stuff!

TerryE:
@jmattsson, Johny, (as with all OS projects) you gain karma here by the value of your contributions so you already score highly IMO.    
I've been testing my compressed debug info patch with Ierusalimschy's Lua 5.1 test suite and this has thrown up issues with coroutining tests (which nodeMCU nominally supports)  I haven't checked out whether the limit is one file open per thread or one file open globally, but the nodeMCU implementation generally manages resources globally rather than on a per thread basis.  I haven't got time to do a full design review on this one so I will take your analysis as is.  Thanks :smile:

jmattsson:
@TerryE All good, I don't mind a robust technical discussion, and I'm open to people pointing out when I'm doing stupid things!

For example, it turns out that disabling the SPIFFS cache exposed some "interesting" behaviour in the lower-levels of the flash *read* access API, which explains the issues we've had with alignment-related crashes (and I can't actually blame SPIFFS for it, so I'm going to have to go eat crow over there too).

The reason for the mess around getting SPIFFS to run boils down to platform_s_flash_read() not mirroring the capabilities of platform_s_flash_write() in regard to alignment leniency. The read path required an aligned destination buffer, which was not always the case due to the layout of the filesystem. Having the SPIFFS cache enabled masked a lot of these issues, and with some shims to force certain alignments in various parts of SPIFFS the whole thing could be put together in a way that made it work. It was really just a bunch of band-aids on top of the real problem though. Fixing platform_s_flash_read() to correctly handle an unaligned destination buffer removed the need for all of those workarounds, and also removes a number of hidden issue where things previously worked largely by chance. There are places where reads from flash go into char-arrays which just-so-happen to start on a 4-aligned boundary, but where adding e.g. a char or short on the stack just before it would have caused things to blow up. This is now no longer the case.

Another "fun" thing that came to light was that the flash write path was unable to deal with the case where the source was a flash location, despite it having a perfectly good bounce buffer capability. Largely, this wasn't much of an issue as nothing in NodeMCU made use of such a path until it became possible to disable the SPIFFS cache (and you then were brave enough to still want to do writes, and happened to write some string literals straight from rom). Trivial fix though, since it already had the bounce buffer capability. Definite head-scratcher however.

TL;DR:
  Bernie fixed the platform_s_flash_read(), making everything work better.
  Keeping SPIFFS up to date should be a lot less painful.
  Running without SPIFFS cache now works in many more scenarios (possibly all?)

jorgempy:
+1 to merge this!

jmattsson:
Oh, and if anyone had any doubts as to whether using node.compile() without the SPIFFS cache, let me assure you it is *not* a good idea! You will very likely make the SPIFFS garbage collector very very unhappy in a very short amount of time. Not to mention it puts a lot of stress/erases on the flash, and a compile can take up to 3 minutes or so...

Just a fair warning for those of you who want to play without the cache! Treat the fs as read-only, and if you want to do anything beyond that, test, test and test more. Or, you know, don't mess with the cache in the first place - it doesn't chew up that much RAM any more ;)

aguaviva:
Definitely with a big power comes a big responsibility :)

What would be the preferred way to document all this? Wiki?

PR #588, opened 2015-08-02
Use #if instead of #ifdef.

robertfoss:
Fixes issue where #define USE_DNS 0 is interpreted as true and enables DNS in the DHCP Offer message.

### DIFF
--- a/app/lwip/app/dhcpserver.c
+++ b/app/lwip/app/dhcpserver.c
@@ -145,7 +145,7 @@ static uint8_t* ICACHE_FLASH_ATTR add_offer_options(uint8_t *optptr)
         *optptr++ = ip4_addr3( &ipadd);
         *optptr++ = ip4_addr4( &ipadd);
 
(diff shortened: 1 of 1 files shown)


marcelstoer:
Not sure if GitHub sent you a notification about the build/compile errors...

> dhcpserver.c: In function 'add_offer_options':
dhcpserver.c:111:5: error: "USE_CLASS_B_NET" is not defined [-Werror=undef]
 #if USE_CLASS_B_NET
     ^
dhcpserver.c:157:5: error: "USE_CLASS_B_NET" is not defined [-Werror=undef]
 #if USE_CLASS_B_NET
     ^
dhcpserver.c:175:5: error: "USE_CLASS_B_NET" is not defined [-Werror=undef]
 #if USE_CLASS_B_NET

robertfoss:
Thanks @marcelstoer, I removed the #ifdefs that weren't needed. It now compiles.

Issue #594, opened 2015-08-07
Download corrupted for large flash images


devsaurus:
I was banging my head against this the last days while adding a new module until I found this: themadinventor/esptool#53

It seems that the ROM code for download contains an issue which leads to a corrupted flash contents once the image grows beyond flash address 0x70000. For details refer to https://cesanta.com/blog/esp8266-flash-erasing-issue.shtml
It happens with current themadinvetor/esptool as well as the embedded esptool here. Since dev branch exceeds this limit for the float version, we're already in trouble now (integer version is briefly below the threshold).

This issue can't be resolved without a fix or workaround in esptool and other flashers.

MarsTechHAN:
@marcelstoer Do you have better idea how to download in Linux & MacOS, Espressif's tool only work under windows?

marcelstoer:
@MarsTechHAN Sorry, I can't help you with this, it really isn't my area of expertise.

MarsTechHAN:
@marcelstoer But...I do think before we fix this bug or find a better way, we should not delete the present methods...Because after it, most of the man will have no idea how to download bin in Linux or Mac OS.

marcelstoer:
@MarsTechHAN I guess your comments refer primarily to my PR-in-the-making #595, right?
- If so, let's continue the discussion there. I have the feeling your concerns are already addressed in the comments there.
- If not, then please elaborate as I might not understand your comments. I don't want to get rid off `esptool.py`, not at all (I use it frequently myself), but rather make sure NodeMCU always uses the up-to-date one from the themadinventor repository.

jiegec:
Well, but themadinventor seldom updates the esptool.

marcoskirsch:
> Well, but themadinventor seldom updates the esptool.

I think that's a different issue. If it becomes a huge problem, the community can look into branching esptool into its own repository. But it still makes sense to manage that separately from nodemcu-firmware.

devsaurus:
> Well, but themadinventor seldom updates the esptool.

If we'd follow @marcelstoer's proposal in #595 we could decide to pull in a submodule even from a different fork/branch. Eg the workaround in jkent's [erase_fix branch](https://github.com/jkent/esptool/tree/erase_fix) works fine for me. ...and link back once the fix is merged to upstream.

Don't know what the git experts think about this...

marcelstoer:
[...]

PR #595, opened 2015-08-08
Delete esptool.py

marcelstoer:
I really don't think it makes much sense to host a replica of `esptool.py` in this repository. As #594 showed once again it isn't efficient if bugs have to be fixed/PR'ed in two repositories. We had problems with this not too long ago with the DIO issue (#523 and https://github.com/themadinventor/esptool/issues/40).


zerog2k:
Should it be a git submodule reference to the other project instead?
This should ensure that any dependency projects are pulled in.

marcelstoer:
Ouch, just burned myself badly. Sorry, I spoke too soon. I didn't realize the makefile depends on this.

> ../tools/esptool.py elf2image .output/eagle/debug/image/eagle.app.v6.out -o ../bin/

How to best go about this?

marcelstoer:
> Should it be a git submodule reference to the other project instead?

Yeah, that sounds like a reasonable plan:
- add the `themadinventor/esptool` repository as a submodule into the `/tools` folder
- add the following commands to the makefile to make sure people can still just clone this repo and start the build (rather than using a variation of `git clone` which also inits submodules):
 - `cd tools/esptool`
 - `git submodule init`
 - `git submodule update`

Ok?

zerog2k:
Looks like it might work. I can checkout your branch and test it out.

If nodemcu repo starts using submodule (which is the right way to depend on other projects), may also want to update some of the docs/readmes, e.g.,:
1. initial repo clone should include --recursive command
2. updating/rebuilding will need some submodule update (i think?)

Fine examples from esp-open-sdk:
https://github.com/pfalcon/esp-open-sdk

marcelstoer:
> 1. initial repo clone should include --recursive command

Assumption I: you only need `esptool.py` if you want to build the firmware (and for flashing of course). 
Assumption II: if you flash with `esptool.py` rather than any other flashing tool you download a copy independently of the NodeMCU binaries i.e. it can/should be decoupled from this firmware.

IMO there's no need to clone with `--recursive`. My intention was to keep as much as possible as-is. People building the firmware themselves shouldn't notice `esptool.py` is no longer replicated in this repository. That's why I proposed to add `git submodule init/update` to the makefile.

jmattsson:
The problem I see is that upstream esptool.py can be really slow to merge in fixes (such as the DIO, or lately the SPIEraseArea bug stuff). Perhaps more importantly, there was also no guarantee that we'd get the entrypoint change into upstream, yet that change was what made it possible for me to kick the constants out of RAM into ROM. At this point I'm somewhat hesitant to give up the control that having our own esptool.py brings. Just my opinion, of course.

marcelstoer:
> The problem I see is that upstream esptool.py can be really slow to merge in fixes

Yeah, I also noticed...see https://github.com/nodemcu/nodemcu-firmware/issues/594#issuecomment-129601586

devsaurus:
> Assumption I: you only need esptool.py if you want to build the firmware (and for flashing of course).
> Assumption II: if you flash with esptool.py rather than any other flashing tool you download a copy independently of the NodeMCU binaries i.e. it can/should be decoupled from this firmware.

> Perhaps more importantly, there was also no guarantee that we'd get the entrypoint change into upstream.

I agree with @marcelstoer and @jmattsson. We're facing the fact that esptool is dual-purpose (not a good approach IMO, [dotadiw](https://en.wikipedia.org/wiki/Unix_philosophy#Do_One_Thing_and_Do_It_Well)).

nodemcu firmware is quite innovative on the compile-time features and has specific requirements for section usage eg. We'll probably have to care for this part within nodemcu community. But downloading should be decoupled from that and moved to upstream or forks.

zerog2k:
+1 for fork and submodule

jmattsson:
Sounds like what's wanted is to trim down the in-tree esptool.py to only contain the image building parts (presumably under a different name, e.g. mkespimage.py), and leave out the flashing functionality, but potentially maintain a fork of esptool.py where we can merge in fixes quicker. Is that a fair summary so far?

In this case there wouldn't even be a need to submodule it. The fork would just be something that sits "next" to nodemcu-flasher, available for those who don't use Windows.

marcelstoer:
> Sounds like what's wanted is to trim down the in-tree esptool.py to only contain the image building parts (presumably under a different name, e.g. mkespimage.py), and leave out the flashing functionality, but potentially maintain a fork of esptool.py where we can merge in fixes quicker.

I wouldn't have a problem with that even though I'd still rather submodule the original esptool.py. But if there's some consensus in the community that your proposal is more desirable I wouldn't mind closing this PR.

udif:
@jmattsson, you can always fork themadinventor/esptool into your own esptool repository, THEN submodule it inside nodemcu. What this brings you is the ability to do both the things you want:
1. Maintain your own patches for esptool.py on your fork, and even send a pull request to the original repository, hopefully your PR will be accepted.
2. Easily merge upstream fixes and updates by rebasing your patches once in a while, when you feel like.

jmattsson:
@marcelstoer As per @devsaurus' suggestion of dotadiw, I [trimmed down](https://github.com/DiUS/nodemcu-firmware/tree/mkespimage)  esptool.py to only contain image creation/info capabilities (under the name of tools/mkespimage.py), and also made it able to produce a combined image which should help with the issues around flash erase reported elsewhere. I.e. you only get one larger 0x00000.bin image by default, rather than the 0x00000.bin + 0x10000.bin. Do you want to give it a spin and see if you like this approach?

This effectively cuts the dependency on esptool as far as the nodemcu firmware building itself is concerned, and avoids having to introduce a submodule. The image creation logic is unlikely to change much given how limited in scope it is.

Oh, and I noticed two more references to esptool - both the 'flash' target in the Makefile, and a mention in the README.md which would need to be updated if you remove esptool.

marcelstoer:
> Do you want to give it a spin and see if you like this approach?

I don't understand in detail _what_ the script does (not my area of expertise) but I see merit in such an approach, thanks.
I could cherry-pick your commit into my branch but...

> I noticed two more references to esptool - both the 'flash' target in the Makefile

Grrr...another road block. No `esptool.py` means no `flash` target in the Makefile which in turn means no need for the `ESPPORT` variable in there.

_I_ have absolutely no need for the `flash` target but I don't know about others.

jmattsson:
To be honest, the `flash` target is less useful these days than it was. Even just with the three modules on my desk (ESP-12E, NodeMCU-dev-0.9, ESP-01), I need to use three different flash settings (baud, mode, size), which the Makefile does not cater for. And when I do flash, I don't actually use the NodeMCU bundled esptool.py. I guess what I'm getting at is, while there is an argument to be made for having a `flash` target, there are now a couple of arguments *against* it too, and not just the ideological/best-practice should-we-submodule-it.

This isn't my project, but for what it's worth, I wouldn't miss the `flash` target either.

marcelstoer:
https://travis-ci.org/nodemcu/nodemcu-firmware/builds/77910079 failed after `esptool.py` was replaced with `mkespimage.py` (I cherry-picked https://github.com/dius/nodemcu-firmware/commit/0ee772312fe5179653f07782fb437c2d467afd07 into this PR branch) but I need to come back to this at a later date, no time right now.

zerog2k:
I use flash target, but just because it saves typing due to being two parts. If we have a monolithic image to flash starting from 0x00000, it makes flashing slightly easier.
On the other hand, fork and submodule seems to be the approach with the least overall impact to existing setup.
I would imagine most users who are pulling and building nodemcu will:
1. likely want to flash from same machine (not always, but I'd wager, most of the time)
2. are probably savvy enough to install or already have installed esptool elsewhere.

As a user of nodemcu, I would vote for fork and submodule for completeness for those who are not so savvy with esp sdk, etc.

jmattsson:
@marcelstoer That build failed because the srec conversion you have there expected two bin files, whereas my mkespimage branch generated a combined bin image.

marcelstoer:
@jmattsson I realize I ventured deeply into unknown territory with the turns this PRs is taking. AFAIU we can do away with `srec_cat` all together if using `mkespimage.py`. All the [current instructions](https://github.com/nodemcu/nodemcu-firmware/blob/master/.travis.yml#L15) seem to be doing is concatenating to two `.bin` files into one, padding/filling the gaps with a constant value without actually _transforming_ the input files. So, with only 1 input file left there's nothing to concatenate or pad, right?

More importantly though...I'd really like to get feedback from @nodemcu, @vowstar  or @MarsTechHAN what the chances are this PR will eventually get merged. Do you guys support the direction this PR is going?

jmattsson:
@marcelstoer Yep, that's my understanding too.

TerryE:
Yup, I like this approach. :smiley:

marcelstoer:
Guys, thanks for all the feedback. I learned a fair share. I'm closing this PR because I need start over and create a PR against `dev` rather than `master`.

PR #604, opened 2015-08-16
BMP085: Fix B5 calculation

kbeckmann:
B5 was calculated with some precision loss before. Thanks @aguaviva for spotting the error.


aguaviva:
thanks to you for contributing with your module :)

PR #618, opened 2015-08-27
Replace Dev120 to Dev130 with latest sdk1.3

3dot3:
sdk 1.3 fixed some wifi bug, it is worth to replace Dev120 with it.


dnc40085:
Shouldn't this be posted under issues?

Issue #624, opened 2015-08-31
Minimal Debug Library Support


TerryE:
I needed to add a minimal debug support for my luac patch to provide a traceback, and it struck me that it's crazy not supporting  this as standard so I created this patch to provide this function and the debug.getregistry() discussed in #452.  See my commit dddf46ebade5bfb9ef15351b508c1016c9da3a60.  I'll do a pull request in a day or so if anyone is interested.

I've commented out the unused elements if built with `LUA_USE_BUILTIN_DEBUG_MINIMAL`.  

Incidentally I've also worked out why the full debug library fails to build --this  is due to to differences in standard readline vs the SDK UART handling.  It would probably be fairly straightforward to code around this, but it's not a high priority for me at the moment.

TerryE:
Patch now incorporated into master.

TerryE:
I've noticed that the traceback is missing the bottom level in the stack.  I just need to validate what is going on here and why.

TerryE:
Nope.  Just rechecked.  I am not losing the bottom level.  Mindfart.

PR #626, opened 2015-08-31
Add minimum debug library providing traceback() and getregistry() functions

TerryE:
As discussed in nodemcu/nodemcu-firmware#624, this patch implements a minimal debug library to provide correct trace back error messages and also the getregistry() function.


TerryE:
In correcting pull from wrong branch, I've used the wrong patch.  I will fix and redo

PR #630, opened 2015-09-01
Add Luac.cross capability to nodeMCU

TerryE:
A standard Lua build generates two executables: `lua`, the standalone runtime envirnment; and `luac`, the lua compiler  

eLua has been designed to be built against a newlib-based ELF library.  This library largely maintains the standard C API naming and include hierarchies but these have been stripped down to enable easier embedded use / implementation (rather than being ran over a standard operating system).  eLua also splits the two standard executables into `lua` which uses the ELF libraries to run on the target and `luac.cross` which builds on the host using the standard host-provided includes and runtime.

nodeMCU has adopted a different convention and modified the Lua sources accordingly by implementing two main changes:

*    Include files are "" pathed rather than <> pathed and are prefixed by `c_` so `<stdio.h>` becomes `"c_stdio.h"`, etc.
*    Standard C runtime functions have been prefixed with `c_` so `printf` becomes `c_printf`, etc.

I am not at all sure of the logic for doing this as it seems to have involved a lot of changes for no benefit, but we shouldn't change this at this stage in the project development. However one consequence is enabling the building of `luac.cross` becomes problematic and has therefore been dropped.

`luac.cross` is a really "nice top have", so this patch migrates across the standard `lua-cross.lua` build process into the `tools` directory.  It also enables this build process by introducing a set of C prepocessor macros which: 

*   Generate the current source if `LUA_CROSS_COMPILER` is not defined
*   Generates standard host compatible source if `LUA_CROSS_COMPILER` is defined
 
I have done this by adopting the following rules:

*    All c modules used in the luac build define `LUAC_CROSS_FILE` and include `lua.h` as their first include.  (They all included `lua.h` anyway, but sometimes later in the include chain.  This change does not impact the generated source in these cases.)
*    `lua.h` includes `luac_cross.h` if `LUA_CROSS__FILE` is defined and this include defines the required macros.  This enables any  `LUA_CROSS__FILE` enabled C files to use these conditional macros.
*    These macros enable the correct C header to be mapped in depending on whether `LUA_CROSS_COMPILER` is defined and also unmap `c_printf` to `printf`, etc. if needed.
*    Any includes which require `LUA_CROSS_COMPILER` dependent conditional include logic have this inline depending on an `#ifdef LUA_CROSS_COMPILER` test, so that none-luac components build unchanged in a standard build.  (This avoids a bunch of regression test issues.)
*    The eLua source contained some conditional code depending on `LUA_CROSS_COMPILER` which has been replaced by `#if 0` in the nodeMCU build.  I've converted this back to the correct format.

As a result of the changes: this patch generates an unchanged binary if built with the standard make, and it also generates a host exectable `luac.cross` if the standard eLua luac cross build process is executed. 

Note: This patch also depends on my last minimum debug library patch hence this commit being included here as well.

### DIFF
--- a/app/include/user_modules.h
+++ b/app/include/user_modules.h
@@ -8,7 +8,8 @@
 // #define LUA_USE_BUILTIN_IO 			// for io.xxx(), partially work
 
 // #define LUA_USE_BUILTIN_OS			// for os.xxx(), not work
-// #define LUA_USE_BUILTIN_DEBUG		// for debug.xxx(), not work
+// #define LUA_USE_BUILTIN_DEBUG
+#define LUA_USE_BUILTIN_DEBUG_MINIMAL // for debug.getregistry() and debug.traceback()
 
 #define LUA_USE_MODULES
 
--- a/app/include/user_modules.h
+++ b/app/include/user_modules.h
@@ -11,6 +11,7 @@
 // #define LUA_USE_BUILTIN_DEBUG
 #define LUA_USE_BUILTIN_DEBUG_MINIMAL // for debug.getregistry() and debug.traceback()
 
+#ifndef LUA_CROSS_COMPILER
 #define LUA_USE_MODULES
 
 #ifdef LUA_USE_MODULES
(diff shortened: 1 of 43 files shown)


vowstar:
Very cool and save us form headache

PR #643, opened 2015-09-10
Added HW SPI and UART ports

lionello:
Added the port assignments as they appear on the NodeMCU dev kit


robertfoss:
That's useful information to have around. Thanks!

Issue #520, opened 2015-06-30
HTTPS (SSL) issue


kroegerama:
The following code does not seem to work as expected ***(0.9.6-dev_20150627)***:
```
sk=net.createConnection(net.TCP, 1)
sk:on("receive", function(sck, c) print(c) end )
sk:on("connection", function(conn)  print("connected") end )
sk:on("disconnection", function(conn)  print("disconnected") end )
sk:connect(443,"api.pushbullet.com")
```
This instantly displays the "disconnected" message.
Same for:
```
sk:connect(443,"google.com")
```

pastukhov:
I'm getting 
```lua
DNS retry 1!
DNS retry 2!
DNS retry 3!
DNS retry 4!
DNS Fail!
````

**Comment by MarsTechHAN** [comment]
If it just show disconnection...
It means NM even not connected...

Check the address


As for DNS Retry, the IP address is not exist in DNS Server

Both of question is not related much to SSL

**Comment by kroegerama** [comment]
My question is indeed related to SSL. The error happens only, if I try to use a secure connection.

This code works just fine:

```
sk=net.createConnection(net.TCP, 0)
sk:on("receive", function(sck, c) print(c) end )
sk:on("connection", function(conn)  print("connected") end )
sk:on("disconnection", function(conn)  print("disconnected") end )
sk:connect(80,"api.pushbullet.com")
```

(changes: **0** in line 1, **80** in line 5)

@MarsTechHAN: Did you actually try connecting to a https service?

MarsTechHAN:
I will try it later, but does the server accept SSL connection？

kroegerama:
Thank you for trying to help.

Yes, the server does accept SSL connections. In fact, it requires SSL.
My non-ssl request just gets an answer from the Server like "please use SSL to connect to this service".

kroegerama:
Just a quick update: I built NodeMCU from the latest dev096-branch and still have the same issue.

marcelstoer:
I went as far back as NodeMCU 0.9.5 build 20150318 and can confirm this fails for SSL.

georgkreimer:
AFAIK dev096 branch is built on Espressif SDK 0.9.6_b1. It's using a 5month old libssl (https://github.com/nodemcu/nodemcu-firmware/commits/dev096/lib/libssl.a) SSL was properly fixed in a much later SDK version. I believe 1.1.0

abl:
[...]

PR #644, opened 2015-09-10
Added ENDUSER_SETUP module.

robertfoss:
The module will start a SoftAP, DNS, HTTP server acting as a captive portal.
The portal will have the SSID SetupGadget and will serve a website where
credentials for the end users WLAN can be submitted.

Once valid credentials have been submitted and used to connect to the
local WLAN this module will stop.

This module requires USE_DNS to enabled in dhcpserver.h

Start the captive portal by running enduser_setup.start()


pastukhov:
It's possible to improve this module to add the list of wifi networks that module found?

marcelstoer:
Nice...what happens if I move the ESP to a different location and none of the previously used/configured networks are available?

pastukhov:
I'm using
```
    wifi.setmode(wifi.SOFTAP)
    enduser_setup.start()
```
and getting
```
> dofile('captive.lua')

-------------------
enduser_setup_start
-------------------

enduser_setup_http_load_payload unable to load file "index.html", loading backup HTML.
> 
WiFi Credentials Stored
-----------------------
name: "WiFi"
pass: "12345676789"
bssid_set: 0
bssid: ""
-----------------------

Connected to "WiFi" as 192.168.100.103

------------------
enduser_setup_stop
------------------

enduser_setup_http_stop failed. espconn_delete returned ERROR #4294967291.
```

pastukhov:
Please make wifi network name configurable.

robertfoss:
pastukhov:
wifi.setmode(wifi.SOFTAP) is run from the module and does not have to be run separately.
The SSID of the captive portal could easily be made configurable as an optional part of the start() call.

marcelstoer:
Nothing happens. Once the captive portal has stopped, it won't start by itself again.
The portal could be made to run whenever the ESP has been unable to enter STATION mode for 60 seconds or so. But that might be a bit intrusive.
How would you prefer that functionality to be implemented?

zunzu:
Im getting:
NodeMCU 0.9.6 build 20150627  powered by Lua 5.1.4
Opening WiFi credentials portal
lua: not enough memory

TerryE:
Why are you tagging loads of program and constant data in`ICACHE_RAM_ATTR`.  This means any anyone compiling in this module statically allocate a big chunk of RAM for no good reason and this will kill their RAM.  What is the logic for doing this?  Something like this belongs in Flash even without the JM patch.

We should really have a quality checklist and developer guidelines to stop this sort of "antisocial" library module killing the system.

robertfoss:
@TerryE Things which are latency or throughput sensitive have been tagged as *ICACHE_RAM_ATTR*. This is mostly functions. And it was done to make sure the we respond to incoming DNS/HTTP traffic in the promptest way possible.

Things that aren't as sensitive has been tagged with *ICACHE_FLASH_ATTR*.

Enabling this module where it isn't intended to be used is a big waste. Even if it is all moved to flash it will use about 5kb of flash. Which indeed is a lot. Especially for the ESP-01 modules. So I wouldn't suggest that this module is enabled by default.

So if this module only is compiled into images where it is intended to be used, the trade-off between performance and ram I consciously did is a lot more reasonable.

Do you agree?

marcelstoer:
> The portal could be made to run whenever the ESP has been unable to enter STATION mode for 60 seconds or so.

That or it could
- try to connect at most n times, possibly with some exponential backoff factor between retries
- listen for some reset action on a pin, that's what https://github.com/sebastianhodapp/ESPbootloader/blob/master/init.lua#L12 does

robertfoss:
@marcelstoer so the way the portal works now is that once it's been enabled it won't disable itself until it has acquired a STATION mode IP. I just want to clarify that.

I'm not sure where functionality relating to this should be written. To some extent things will still have to be done in lua.
The pin listening I feel is too specific to put into a c module. It could be done, but it probably only would help a small fraction of users. But maybe I'm wrong.

I don't think I understand how an exponential backoff would work. The problem you're wanting to solve is how the ESP8266 would try to stay connected when moved from one location to another?

marcelstoer:
> The problem you're wanting to solve is how the ESP8266 would try to stay connected when moved from one location to another?

Moving yes, staying connected no. Sorry for not being specific enough and for not studying your code thoroughly (it's been 15 years since I last wrote C code).

Were you saying that with your module the ESP stays in portal mode whenever it boots until WiFi is configured? I thought the WiFi configuration was persisted somehow once it's set? If so, there has to be a way to reset it if the ESP is moved to a different WiFi.
Let's say I tinker on a module at home i.e. using my home WiFi. The next day I take it to work. Nothing changes except the WiFi configuration. I want to plug it in, configure WiFi and be good to go.

robertfoss:
@marcelstoer So *enduser_setup.start()* will have to be called somewhere from lua before the module does anything. But once called the portal will keep running until a working set of SSID+PASSWORD has been submitted. Once connected to that SSID the module will exit.

So if the ESP is started and *enduser_setup.start()* called the portal will run until a working set of SSID+PASSWORD has been found. Once connected to that SSID the module will exit. After the module has exited it will not restart itself unless told to do so from lua space.

TerryE:
> Things which are latency or throughput sensitive have been tagged as ICACHE_RAM_ATTR. This is mostly functions. And it was done to make sure the we respond to incoming DNS/HTTP traffic in the promptest way possible.  Things that aren't as sensitive has been tagged with ICACHE_FLASH_ATTR.

Sorry, but you fundamentally misunderstand the epoches involved.  If you are writing a device driver or bit diddling some physical hardware where µS precision is important *then* you might want to cache code and R/O data in RAM.  The latencies that you are dealing with are 10s of mSec if not 100s.  Most of the helper routines that you are calling run out of flash anyway, so what advantage is there is moving thie code into RAM?  Does it really matter materially whether your code runs at 20 MIPS or 5 MIPS?  

`http_html_backup` is about 3Kb which going to be split into 2 IP packets anyway.  If you want to avoid the exception handler overhead then malloc the RAM, memcpy it to RAM and free the RAM when done.  White a couple of macros to encapsulate it and then doing this is largely transparent at a source level.   OK, so this adds 3-4µS to the runtime of `enduser_setup_http_load_payload()`, but again this isn't material.  You've just freed up 3K RAM for use in the rest of the execution.

> Enabling this module where it isn't intended to be used is a big waste. Even if it is all moved to flash it will use about 5kb of flash. Which indeed is a lot. Especially for the ESP-01 modules. So I wouldn't suggest that this module is enabled by default.  So if this module only is compiled into images where it is intended to be used, the trade-off between performance and ram I consciously did is a lot more reasonable.

So when the new-to-esp8266 programmer includes this in her or his "standard" build and then posts new issues saying "my app is running out of RAM", are you going to field these new issues?  IMO, you've blown away a third of the available RAM on the 0.95 builds for no rational reason.  If every other core and lbrary developer did this then we'd need an extra 128Kb RAM just to boot.  So I would vote against including this patch on any nodemcu branch whilst this patch uses RAM_CONST_ATTR.  And so I clearly don't agree with your analysis.

This is even more emphasised by the discussion with Marcel on setup delays which can be 100s if not 1000s of mSec.  Will the user even notice?

robertfoss:
@TerryE Tell me if I'm reading this wrong 

    ./include/c_types.h:88: #define RAM_CONST_ATTR __attribute__((section(".rodata")))
    
    ./ld/eagle.app.v6.ld:75 /* Trade some performance for lots of ram. At the time of writing the
                             * available Lua heap went from 18248 to 34704. */
                             *(.rodata*)

According to these linker scripts (which I read before adding RAM_CONST_ATTR to anything) everything in ".rodata" will be put into flash. Is this not the case?

>http_html_backup is about 3Kb which going to be split into 2 IP packets anyway. If you want to avoid the exception handler overhead then malloc the RAM, memcpy it to RAM and free the RAM when done. >White a couple of macros to encapsulate it and then doing this is largely transparent at a source level. >OK, so this adds 3-4µS to the runtime of enduser_setup_http_load_payload(), but again this isn't material. You've just freed up 3K RAM for use in the rest of the execution.

Re-allocating the same slab of memory for every http request is a waste. Having the complete HTTP response allocated while the module is _running_ is quite reasonable.
I'm not sure if you've read the code. But the way it works now is that the complete HTTP payload is malloc'd and assembled upon starting the module. That basically means combining *http_header_404* with *http_html_backup* in a buffer. This buffer is freed as soon as the module has completed received a working set of credentials.

Are you suggesting doing this assembling for every HTTP request?

>So when the new-to-esp8266 programmer includes this in her or his "standard" build and then posts new issues saying "my app is running out of RAM", are you going to field these new issues? IMO, you've blown away a third of the available RAM on the 0.95 builds for no rational reason. If every other core and lbrary developer did this then we'd need an extra 128Kb RAM just to boot. So I would vote against including this patch on any nodemcu branch whilst this patch uses RAM_CONST_ATTR. And so I clearly don't agree with your analysis.
This is a bit hyperbolic isn't it? Out of the free 35kb of memory 33kb are still available when the module is running. I'm not sure what numbers you've seen. But if they are as alarming as you are suggesting, please give them to me.

TerryE:
You've added `ICACHE_RAM_ATTR` pretty much everywhere including on your code segments.

What's wrong  with "wasting" a few microseconds of processing on a per request basis? You aren't paying for them per ¢.

I don't field any issues or Qs on the forums saying "Gosh, I've got loads of memory available", but I do continually have to explain how to fit apps in " too little" RAM. You can write this using a *lot* less RAM, and my vote is thatc this module doesn't belong in this repository if you don't want to.

robertfoss:
@TerryE I asked you in my last post if _ICACHE_RAM_ATTR_ wasn't stored in flash. Like the linked scripts actually suggest. Is it or isn't it stored in ram?

TerryE:
>> What's wrong with "wasting" a few microseconds of processing on a per request basis? You aren't paying for them per ¢.

> You are. With performance and battery life. Which are the same thing. All the time spent allocating and then de-allocating the same memory is just a waste. Because in reality we'll actually rely on that memory being available to us when we receive an HTTP request.

I used to take my specs  off at night when I was a young kid because I was afraid that if I wore them when asleep then they'd not have a chance to let them rest and they'd quickly wear out.  You seem to be using the same thought process here.  Don't anthropomorphise a piece of silicon.  If you have a target response of 100mSec, say, and you are running at 40mHz then your code has 4M clock cycles to complete.   Whether it completes in 1.1M vs 1.2M doesn't matter in the slightest. You aren't going to save anything.  Certainly in terms of overall power consumption is makes bugger all difference.  If you are running of battery then you need to use sleep and deep sleep plus the RTC RAM to store sleep-to-sleep context. *That* will save battery life. 

>I asked you in my last post if `ICACHE_RAM_ATTR` was or wasn't stored in flash. Like the linked scripts actually suggest. Is it or isn't it stored in ram?

It's in RAM not Flash; the RAM in the name is a bit of a give away. I pulled your patch into my current dev setup which is based on dev120 and tried to make it.  It failed with the error: section `'.text' will not fit in region 'iram1_0_seg'`, which means that this module breaks any SDK 1.2 based builds. (I tried commenting out all of the non-core module, but the build still failed).

Iram1_0 is the bit of Flash (see this [ESP8266 Memory Map](http://www.esp8266.com/wiki/doku.php?id=esp8266_memory_map#memory_layout)) that is mirrored into a fixed 32 kB onchip RAM window.  It is used to store time critical interrupt code and constants which need to be in RAM to avoid Flash fetch overhead.  It is a fixed size, and the SDK takes up a lot of it.  With each SDK adding new functionality, it's getting tighter and tighter -- hence Johnny's patch.  If you want to see the impact of your implementation then do your build with and without your module enabled and run

```
xtensa-lx106-elf-objdump -h app/.output/eagle/debug/image/eagle.app.v6.out
```
to get a summary report of the usage, and get your hex calculator out to compare the two.

robertfoss:
### Re-allocating the same buffer for each request
>>>What's wrong with "wasting" a few microseconds of processing on a per request basis? You aren't paying for them per ¢.

>>You are. With performance and battery life. Which are the same thing. All the time spent allocating and then de-allocating the same memory is just a waste. Because in reality we'll actually rely on that memory being available to us when we receive an HTTP request.

>I used to take my specs off at night when I was a young kid because I was afraid that if I wore them when asleep then they'd not have a chance to let them rest and they'd quickly wear out. You seem to be using the same thought process here. Don't anthropomorphise a piece of silicon. If you have a target response of 100mSec, say, and you are running at 40mHz then your code has 4M clock cycles to complete. Whether it completes in 1.1M vs 1.2M doesn't matter in the slightest. You aren't going to save anything. Certainly in terms of overall power consumption this makes bugger all difference. If you are running off battery then you need to use sleep and deep sleep plus the RTC RAM to store sleep-to-sleep context. That will save battery life. 

Any CPU of sane design will indeed use less energy if it is executing NOPs than if it is doing memory and register accesses. Yes or no? And by doing one allocation instead of the same allocation over and over we will certainly do less work.

You are indeed correct in saying that going into deep sleep would save power. But that is not a feature that has been implemented (nor is likely to be). Which means that saving power in other ways is still meaningful.

Additionally you didn't address the fact that changing the malloc from being run on start up to whenever a HTTP request is run is received does not lower the memory requirements for this module. It will just make it fail whenever a HTTP request is received.

### Selecting linker space for large && non-performance sensitive data/functions.
>>    ./include/c_types.h:88: #define RAM_CONST_ATTR __attribute__((section(".rodata")))
>>    
>>    ./ld/eagle.app.v6.ld:75 /* Trade some performance for lots of ram. At the time of writing the
>>                             * available Lua heap went from 18248 to 34704. */
>>                             *(.rodata*)
>>
>>According to these linker scripts (which I read before adding RAM_CONST_ATTR to anything) everything in ".rodata" will be put into flash. Is this not the case?

I repeated you dev120 build experiment and reached the same conclusion you did. RAM_CONST_ATTR does indeed force things into ram, which means I've misread the LD files.

What I intended to do was force CONST data into flash, but unwittingly forced it into ram. I'm going to supply a commit that adds a ROM_CONST_ATTR for forcing content into flash. As well as remove all occurrences of RAM_CONST_ATTR.

TerryE:
Look, I am getting bored with the quote war, and the willy waving.  I am just trying to help, and give you info to do so.  Have a look at the memory map.  The 32Kb **iram1**slot at 0x40100000 is where the Flash `ICACHE_RAM_ATTR` etc gets statically mapped.  Malloc doesn't touch this; it used the 40Kb **dram0** slot starting at 0x3FFE8000.  The Flash itself is mapped to 0x40200000.  (So these are apples and oranges.)  There may or may not be an L1 cache in front of this Flash Window; the chip is closed source so no one has unequivocally confirmed this.  I personally don't think so, but I'd need to some C benchmarking to find out.   The CPU can certainly fetch memory mapped code and word aligned data from this Flash, albeit with the extra 12 (IIRC) clocks overhead on QIO Flash and 24 clocks on the DIO Flash.

robertfoss:
I've fixed what you primarily were objecting to. And now agree with attributing anything with RAM_CONST_ATTR being wrong. The previous 2 commits reflect that fact.

However I still don't agree with re-allocating memory for every HTTP request. It's just plain silly. It does provide us with any benefits, and it costs performance/battery. So I would rather not make that change.

Are you still objecting to this PR being merged? If so why exactly.

TerryE:
Robert, extensions should be layered on the Lua core firmware not changing it and changing a file as fundamental as `include/c_types.h` counts as such a change IMO, especially when existing  declaratives and defaults exist to enable the extension to work as intended.   So yes, I still object to the this PR.

Just doing a quick walk through the code, there's also a load of other stuff that needs thinking about.

There is a standard style for writing Lua extensions and a module should conform to this.  Read my [FAQ](http://www.esp8266.com/wiki/doku.php?id=nodemcu-unofficial-faq) which gives your the links for the references.  Use the PiL as the style guide and follow some of the Lua module examples.  

For example nearly all of your error processing does a printf followed by a return.  Do a grep -R looking for **printf**s.  It's not done.  That's because the UART is not connected to anything on "production" embedded devices, so these errors getting dumped into the great ether bit-bucket.  

The applications programmer (in general in Lua extensions) treats errors as if they are divided into two categories: "normal errors" e.g writing to a full file system where the app logic could be expected to handle it and in this case a status and an optional status parameter are used.  Or unexpected errors are thrown, and `pcall()` can be used to catch these (and which `pcall()` itself is a case of the former). 

In reality this more like a classic terminal connected C program that has been minimally restructured around the SDK.  It has had a thin layer added so that is can be called from Lua.  I am not sure what it is but it doesn't fit into the category of module which belong in the core repository.

marcelstoer:
Btw Terry, the following comment of yours reminded me of an IMHO fundamental bug in dev120.

>  It failed with the error: section '.text' will not fit in region 'iram1_0_seg', which means that this module breaks any SDK 1.2 based builds. (I tried commenting out all of the non-core module, but the build still failed).

I don't intend to divert the discussion you're having with Robert but as per #563 the same happens with a lot of other modules in dev120.

TerryE:
@Marcel, good point, but I'll pick up on your issue.

robertfoss:
### Selecting linker space for large && non-performance sensitive data/functions.
> Robert, extensions should be layered on the Lua core firmware not changing it and changing a file as fundamental as include/c_types.h counts as such a change IMO, especially when existing declaratives and defaults exist to enable the extension to work as intended. So yes, I still object to the this PR.

So if I'm understanding you correctly what you don't like is adding ROM_CONST_ATTR. To my knowledge there is no other way of doing this. Which means that its a potentially valuable addition.
RAM_CONST_ATTR was added 2015-07-08, and ROM_CONST_ATTR is basically the same thing, but for flash how is one valid but the other one not?

I'm at a bit of a loss as to where you actually want to store these things. Which attributes do you want them to have? I'm very open to suggestions here. But if you want to store them irom0.rodata, there probably should be a define for it somewhere. Preferably where other defines like it are.

> There is a standard style for writing Lua extensions and a module should conform to this. Read my FAQ which gives your the links for the references. Use the PiL as the style guide and follow some of the Lua module examples. 

### LUA extension style guide
I searched through the page for "lua extension", "extension" and "extend"  but couldn't find a link. Could you give the link to me?

### Error handling
> For example nearly all of your error processing does a printf followed by a return. Do a grep -R looking for printfs. It's not done. That's because the UART is not connected to anything on "production" embedded devices, so these errors getting dumped into the great ether bit-bucket. 

You do have a point here. They could be hidden behind a debug macro. Like NODE_DEBUG, but something else, since enabling NODE_DEBUG often has catastrophic consequences for successfully building working images.

The asynchronous nature of receiving and handling UDP/TCP packets prevents pcall from being used to catch errors that occur in the callback functions. So I don't think pcall is a solution to the issue of conveying errors from c-space to lua-space.
I guess a LUA onError callback could be a parameter to the start() function and could handle transmit the errors to lua-space, where they can be handled appropriately. But that is a lot of complexity for doing something that really shouldn't be so hard.

Personally I don't agree with the philosophy of failing silently. If something is confusing to a newbie it is modules that appear to be working at a surface level, but are failing in ways they refuse to tell you about.
But if it is the policy of the nodemcu project at large is to fail silently I will adapt the code to do so.
Should I do that?

TerryE:
Read PiL.  "throw" != "fail silently".  This is not a Lua extension.

robertfoss:
>Read PiL. "throw" != "fail silently". This is not a Lua extension.
I read http://www.lua.org/pil/8.4.html is that what you're referring to? To my understanding it does nothing for asynchronous tasks. Am I misunderstanding something?

How would an error in for example tcp_recv_cb be handled using throw/pcall?

I was hoping for a response to:
 * LUA extension style guide
 * Selecting linker space for large && non-performance sensitive data/functions

TerryE:
PiL -- try the book BTW -- gives the general philosophy.  Lua was written by a bunch of guy with Portuguese as their first language, nodemcu Mandarin or some or Chinese language language.  Ierusalimschy is fluent in English, but like many FLOSS projects the style guide is the implementation itself. 

For example, have a look at how `app/modules/net.c` approaches this.  It exposes the asynchronous model to the Lua application layer, but even if you want to architecture this as almost an extra C thread then if you get tot he point where you error asynchronously then the obvious approach is to stash a Lua errror callback in the Lua registry and use the nodemcu model: pass the problem back to the lua application layer. Think about your NFRs here (non-functional requirements) that is because there are what should constrain your implementation architecture: what is your synchronous vs async model and approach?  How does Lua, the nodemcu firmware, the SDK and the esp8266 chip architecture constrain you?  How do you want to interact at an applications level with programmers who are attempting to use your library?  One approach here could be to mirror how the **net** and **socket** modules works: separate out the configure, execute and on(event callback) methods and store the callback closures in the Lua registry.

As far as your last point see previous reference to #563:  At the moment we seem to be at an empasse.

jmattsson:
>According to these linker scripts (which I read before adding RAM_CONST_ATTR to anything) everything in ".rodata" will be put into flash. Is this not the case?

No. You're confusing symbol names with section names, which is easy enough to do given the "interesting" approach on the eagle soc.

The ".rodata" *section* is located into dram0, i.e. RAM. If you really, really, *really* need a constant to reside in RAM, you can use the RAM_CONST_ATTR macro to force it into this section. Currently only the very low-level symbols _bss_start and _bss_end live there. We have yet to encounter a constant which can't sit pretty in flash.

The irom0.text section is, as the name suggests, located in irom0, i.e. flash. The linker script pulls all *symbols* with names matching .rodata* into that section by default.

I hope that clarifies thing a bit.

robertfoss:
@jmattsson Thank you! So _#define ROM_CONST_ATTR __attribute__((section("irom0.rodata")))_ should be fine and indeed force blobs to be stored in flash? And in that case is that a reasonable addition to c_types.h in your mind?

@TerryE 
Re **Selecting linker space for large && non-performance sensitive data/functions**:
Could I ask you to repeat/clarify you objections?

Re **Error handling**:
So you'd like to see _luaL_error_ based error handling?
Regarding how I'd like the module to interact with application programmers (and the lua/nodemcu interface), beyond a start and a stop function I would like there to be no interaction. This was a conscious design choice from the start. At a later time further interactions could be added if someone finds a need for them. But the idea is for this module to be very self-contained.

TerryE:
@robertfoss: `c_types.h` is an SDK include file.  *Any* changes to it create a long term regression cost, unless we can get Espressif to support and integrate the change.  My preference would be to backout all of the current nodemcu changes and move them into nodemcu build files.

Please read #563.  I have asked you do do this twice. This already contains the discussion that you are asking me to repeat. 

I will repeat: you are writing a Lua extension. Lua programmers will expect it to behave like an extension.  If you wan it to be self contained then don't expose any errors -- handle them all internally.  If you are exposing errors then it isn't self-contained, is it?  A UART based interface is absolutely no good if the esp6288 doesn't have a connected serial interface.  Programmers will want the option of catching errors programmatically through the standard Lua mechansims, e.g. pcall or callbacks with error strings.

robertfoss:
@TerryE
Re **#define ROM_CONST_ATTR __attribute_((section("irom0.rodata"))):**
While I don't agree with you regarding not allowing any changes into ctypes.h. How would you prefer that I convey the *__attribute_((section("irom0.rodata")))* attribute?

Re **Selecting linker space for large && non-performance sensitive data/functions:**
_0f883b1_ and _ 3a05c3e_ made the module build without any of those complications. I had assumed you tested those patches. So when you are referring to #563, are you saying that dev120 does not build for you when you run the latest version of this PR? Or are you saying something else? In that case what?

Re **Error handling**:
Alright, so would defining a debug macro for printing the errors over serial be acceptable to you? The macro would naturally be disabled by default. That way they would 'behave like an extension'.

TerryE:
See `ld/eagle.app.v6.ld` line 199 onwards, and other discussion.  Why invent a new section name?

I haven't test your patches.  I am only one committer on this project and not a member of the core nodemcu team.   It's just that I seem to be the only one that as taken a look at your code.  I don't see my job as to act as your testing service.  

I feel that your extension or some development thereof could be a worthwhile snippet for use by the community, but as it currently stands it is not something on which I would attempt to response to issues and support Qs.

As far as error handling does, if it is only intended as a diagnostic for the extension maintainer then you should follow the `NODE_DBG()` macro model and only generate the debug code if its going to be used.  But there is still the issue of how the Lua developer who wants to use your extension gets meaningful error diagnostics and notification programmatically when using the extension on a non-UART connected IoT device.

We seem to be going around in circles.  I think that I will leave further responses to the nodeMVU team and other committers.

robertfoss:
@TerryE 
Re **#define ROM_CONST_ATTR __attribute_((section("irom0.rodata"))):**
So you're unhappy about the _irom0.rodata_ section being unused and not the _irom0.text_ section?

Re **Error handling:**
>As far as error handling does, if it is only intended as a diagnostic for the extension maintainer then you should follow the NODE_DBG() macro model and only generate the debug code if its going to be used. But there is still the issue of how the Lua developer who wants to use your extension gets meaningful error diagnostics and notification programmatically when using the extension on a non-UART connected IoT device.

I agree with you that hiding debug info behind a macro is not an ideal situation and that going the full luaL_error route would be ideal. For now I'll happily go the route that _mqtt.c_ et al. have been going.
A commit moving the prints to live behind NODE_DEBUG will be submitted in a few minutes.

> I feel that your extension or some development thereof could be a worthwhile snippet for use by the community, but as it currently stands it is not something on which I would attempt to response to issues and support Qs.

>We seem to be going around in circles. I think that I will leave further responses to the nodeMVU team and other committers. 

That's unfortunate. I'll iterating on this module as long as there appears to be interest in this PR.

jmattsson:
> How would you prefer that I convey the __attribute_((section("irom0.rodata"))) attribute?

@robertfoss There is no need to do anything special at all to convey this, const-data-in-flash is the default. Having a bunch of __attribute__((section("blah"))) sprinkled across any code base is not a very good thing (even if it's hidden behind macros). There are a handful of special cases where it is needed, link-time arrays and interrupt vectors being the most common ones I'd say, but they are truly the exception rather than the norm.

With NodeMCU now you automatically get what is the best fit for most things, no magic needed (unlike the plain SDK with all its ICACHE_FLASH_ATTRs). I consider it unfortunate that I had to touch both the linker files and c_types.h to achieve that, as they're both part of the SDK and having changes in there adds to the pain of upgrading the SDK. Keeping the changes minimal is key. Seeing as we haven't encountered a need for the RAM_CONST_ATTR macro yet, I'm tempted to back that one out.

When it comes to UART and debugging/errors, what I'll say is this:
  - Printing things directly to the UART is a generally a bad idea. People might be talking other protocols to peripherals there, and would likely get a wee bit miffed if they got protocol gibberish interjected there.
  - As @TerryE already mentioned, users really should have some way of responding to error conditions. Having had a (very) quick look at this module/pr, given its non-interactive nature I'd say that passing an optional error callback to the .start() function would be appropriate. You could even do .start(err_cb, dbg_cb) if you wanted to easily hook up debugging output too. The arguments passed to those callbacks should ideally be a string and/or error code, so it's dead-easy to just do a .start(print,print) while you're developing and want some quick output. There are plenty of examples in the modules on how to stash and invoke such a callback.

robertfoss:
@jmattsson Thanks for the feedback.

I'll get working no a callback based error mechanism then.

robertfoss:
Support has now been added for supplying 3 callbacks to the start call.
Connected, error and debug.

Debug is not enabled by default but can be enabled during a recompile.

    #define ENDUSER_SETUP_DEBUG_ENABLE 1

The start call could now look something like this for example:

    enduser_setup.start(
    function()
      print("IP address acquired");
    end,
    function(err, str)
      print("ERROR #" .. err .. ": " .. str);
    end,
    function(str)
      print("DEBUG: " .. str);
      end
    )

jmattsson:
This has shaped up nicely I reckon! A couple of further comments & ideas:
- The SSID seems hardcoded. It would be good to have that as a #define over in user_config.h. At the very least the SSID should include (the lower) part of the ESP's MAC address so the networks don't conflict if you're starting up several ESPs at the same time.
- Maybe allow serving of additional files to allow a user to include a logo or some such on their customized index.html?
- You know what would be fantastic? If it was possible to register a dynamic handler for certain paths, e.g. by passing in a table of the form ``{ 'GET': { '/settings': func_read_extra_cfg }, 'POST': { '/apply_settings': func_do_extra_cfg, '/restart': function() node.restart() end }}`` which would make it possible for a user to not just reskin the index.html, but also augment it with some extra settings (e.g. allow naming of the node).

Thoughts?

robertfoss:
* Dynamic and MAC suffixed SSID
  - Probably a good idea to add for before merging.
* Customized logo support
 - From my perspective that can be added during a recompile as base64 inlined in the html.
 - Adding dynamic HTML would add a decent bit of complexity and with regards to feature requests is a bit of a slippery slope
 * Dynamic handlers
  - I think there are better and more proper http servers out there for customizing content. If not, I do think that should be a separate module.

As it stands now both the DNS and the HTTP parts of this module are implemented in the simplest possible way the could possibly work. And for this module, I think that is the correct way of doing it.

Does that sounds about right to you?

jmattsson:
Yep, just the SSID improvement is probably a good call at this point.

grahamehorner:
@robertfoss please consider adding mesh network configuration as an option; with regards to the web UI I would like these elements/layouts that can be customisable to be done a CSS that can be included at compile time.

grahamehorner:
Oh and not to be critical but HTTPS and HTTP POST is a must I think for this given nature off the data getting transmitted.

marcelstoer:
> HTTPS and HTTP POST is a must I think

Ideally yes, but due to #520 you're currently out of luck.

robertfoss:
@marcelstoer  @grahamehorner: Addind SSL support sounds reasonable to me. Realistically this module won't be imported into a branch earlier than dev120, which has a working SSL implementation as far as I know.

grahamehorner:
@marcelstoer #520 is client connection to https? not running an https server? or have I missed something here?

marcelstoer:
> #520 is client connection to https?

That's correct. However, it's my understanding (i.e. my assumption) that you'll run into similar issues if you run an HTTPS server.

grahamehorner:
@marcelstoer that could be the case; however until verified that hosting HTTPS servers have an issues to assume that HTTPS server hosting doesn't work and we have an issue (I believe) wrong. 

HTTPS server hosting should in theory be simpler then HTTPS client communication given a client requires to exchange/validate the certificate and certificate chain etc. to establish the trusted connection.

robertfoss:
I've looked into running a HTTPS server and serving the captive portal from it. There appears to be two main issues with doing that.

 * Both a HTTP and HTTPS server would need to be run as a enduser is likely to type in a non-https url in the browser. This means that the HTTP server would have to serve redirects to the HTTPS server.

 * The HTTPS server would have to serve a self-signed certificate. Actually it's worse than that, since the the domain we're hijacking never will correspond with the domain in our certificate. This is pretty much the kind of stuff that SSL is meant to protect against.

As it stands now, I see no reasonable way of introducing SSL support.

grahamehorner:
@robertfoss routers and access points do this without issue, they use a self cert which maps to the ip address, changing the ip regenerates the certificate and cause a reboot of the roofer or access point.  I personally don't consider a using typing http:// ... instead of https:// and not getting redirected an issue as we are focused on technical users doing config of some sort.

I would think if the initial start up of the esp8266 is in access point (no password) mode with https on a fixed up which allows for a user configuration webpage to be supplied like routes etc this is a good starter for 10 and later addition like redirection and self cert generation or cert uploading may be added.

grahamehorner:
a simple top socket listener on port 80 should be able to redirect with minimum of resources.

robertfoss:
@grahamehorner: This configuration module is very much not focused on technical users. It is focused on any enduser of an IoT product. Ideally you shouldn't have to know anything to use this.

About the security benefits of pushing data through SSL:
 * The esp is unable to generate certificates on the fly. This has to be done on separate computer and then transferred, which means it can't be done on the fly.
 * The certificate will be self-signed since certificate authorities (CAs) don't issue certs for LAN IP addresses. And the cert would have to be issued to the ESP SOFTAP gateway IP. Which typically is 192.168.1.4. 
 * Self-signed certificates are trivially MITM-able, which means that they don't really offer any security
 * Self-signed certificates only offer any security if a site is visited more than once (through certificate pinning), which won't be the general case for these configure once IoT products.

If you actually are worried about someone snooping the HTTP traffic between you and the ESP, running in a non-open SOFTAP_MODE makes more sense. That is however still worse for usability and the target user group here. 

So to make this clear; I will not implement an SSL mode. It would provide us with no security benefits and make the enduser experience worse.

robertfoss:
On a separate note, does anyone have any objections against this PR being merged?

grahamehorner:
@robertfoss that's your view and I would agree to disagree on this; any security is better than no security, the IoT space has allows had issues with security and processor/solutions that don't have security have been quickly compromised and slated by the press resulting in low user confidence.

of course weak security also has an impact in that; it give a false sense of security, and hence why the suggestion was to use a self-cert (browser with give warnings) and allow the users to upload a certificate into flash via the configuration page; that would give greater security as a middle option.

I still think that using HTTP GET with sensitive data in the URL as parameters is VERY RISKY and in general bad practice, regardless of weather over HTTP or HTTPS and this should be changed to a HTTP POST at a minimum.

robertfoss:
@grahamehorner Security that you and I know how to break is at the very best security by obscurity, which is not security at all.

Which benefits would using HTTP POST have? What makes a GET more risky than a post?

grahamehorner:
@robertfoss see google or stackoverflow HTTP GET and sensitive data.  better have some lock on a door than a door with no lock and wide open!

grahamehorner:
http://192.168.1.4/config?username=me.here&password=letmein no anti forgery token or any check = no border control BAD news!

robertfoss:
I'm not finding any differences between POST and GET in terms of security. And nor would I expect there to be any.

We really don't have to worry about forgery attacks here. And even if someone were to submit garbage data as PWD and SSID, it simply would be discarded since no-one is likely to guess our wifi credentials.

So what exactly is the attack you had in mind where a token would help us?

grahamehorner:
just check your browser history?! your not even considering securing the data using rsa public/private key which can be done simply via javascript; and has many many implantation on the world wide web.

I'll leave this topic now; but in my opinion and many others posting sensitive data which are in a URL is wrong; can be faked and presents a real risk.

A RSA key pair consist in the following:
•The modulus n, a big integer which is equal to the product of two big prime integers, p and q.
•The public exponent e.
•The private exponent d. d is such that ed = 1 when taken modulo p-1, and also when taken modulo q-1.
•The first factor p.
•The second factor q.
•The value dp = d mod p-1.
•The value dq = d mod q-1.
•The value q' = q-1 mod p.

The public key contains only n and e.

The private key contains all of the values specified above. only n and d are needed for a perfunctory RSA implementation, but the other values allow for faster operation, and also "masking" (protection against side-channel leaks), so the RSA standard defines that a private key contains all the values (see appendix A.1).

From the value you have, assuming that your "m" is really the RSA modulus, the other values can be recomputed, at some effort (less than one second worth of CPU, but more human time than I care to invest in it). From all these values, you could then encode standard-compliant RSA public and private keys; but this requires some knowledge of ASN.1. Finally, a 512-bit RSA key is weak; use at least 1024 bits, preferably more (2048 bits ought to be fine). Some RSA implementations will refuse to work with RSA keys shorter than 1024 bits.

If you and the other developers are happy then I've no problem with your PR being merged, and I will then fork and implement security onto for review.

robertfoss:
So you want to do javascript RSA in the captive portal?

I guess if an asymmetric cipher (like RSA) was to be implemented on both the client and server for enduser_setup this could actually work. And be secure. 

It would however require:
 * A JS RSA implementation that isn't 10KB large. 
 * A C RSA implementation (this might actually exist in app/ssl/crypto/ssl_rsa.c)
  - ssl_rsa.c might be tricky to link too. I'm not sure.
  - ssl_rsa.c might not be doing a straight up RSA, which might make it incomaptible with what we're doing
 * Storing a 2-4KB key-pair. 2048-bit RSA is the minimum key-length that is considered secure.
   - If the key is pre-generated anyone can decrypt this communication and the security gains are lost. We're back to security by obscurity.
   - If the key generation is done on the ESP, it would have to be implemented as a C-module somewhere. No support currently exists.

So to summarize; Doing asymmetric crypto would work. And be secure. But is a _lot_ of work. And far beyond the scope of this module. When the prerequisites for adding RSA to this module exist, we can talk about adding support for them.

marcelstoer:
Kudos to you Robert for sticking with this and improving the module based on input from Terry & Johny.

> So to make this clear; I will not implement an SSL mode. It would provide us with no security benefits and make the enduser experience worse.

I agree with you.

However, I disagree with you regarding GET vs. POST. From a security perspective POST is marginally better (e.g. query string params show up in access log files) but for the given use case it doesn't matter much. As processing POST payload isn't really more complicated than processing GET parameters I'd still use POST, though.

grahamehorner:
@marcelstoer @robertfoss while I don't agree with the outcome; I can honestly say in all my 35yrs as a developer and 15yrs as cisco CCNE I've never come across the developers that are so reluctant to simply apply some common-sense and give some consideration to applying and proving security to a service that receives sensitive information. but hay-ho that is open source for you majority rule even if people disagree; guess I don't have to use it and I'll implement my own solution that is secure.

marcelstoer:
> apply some common-sense and give some consideration to applying and proving security to a service that receives sensitive information

I can't speak for Robert but I believe we both use commons-sense and give this serious consideration. It may simply be a matter of personal risk assessment of the environment we se this module being deployed into.

- Given the constrained resources on the ESP and the restrictions the captive portal imposes I don't see a feasible way to provide SSL support. None of us disagrees that it'd be sensible to have that, though.
- As stated I agree with you on the GET vs. POST issue.
- As for public/private key encryption, yes, that'd be good to have I agree. However, some of the ESPs are so bare bone that it simply may not be possible (not sure, though). Robert made a pretty good analysis of the requirements. Given enough interest it may still be added to the module without tearing much of the existing code base apart.

robertfoss:
By popular demand I'll revisit using POST instead of GET. But it won't be today, I don't have access to a board for testing now.

jmattsson:
And to answer @robertfoss' earlier question, yes I'd quite like to get this merged in. Perhaps after the POST vs GET has concluded? And Robert, would you mind moving the ENDUSER_SETUP_AP_SSID define into app/include/user_config.h? That is meant to be the canonical place for such tweaks.

@grahamehorner I'm really not sure which security aspect you are worried about here? This module is intended to be used, very briefly, during an initial deployment of a device. During that time the end-user will specifically join the device's AP, and do a one-off configuration. Using SSL here would at best provide some obfuscation - there would still be nothing stopping an "attacker" simply joining the same network and doing the configuration him/herself. The overhead for that bit of obfuscation is simply not warranted on a limited device such as this under what I would consider normal circumstances. To provide actual security, one would need to:
  - run the AP in WPA2 mode
  - ideally run SSL, and with a properly signed certificate
  - ideally also do HTTP-level authentication (or SSL client cert)

For protocol correctness HTTP POST or PUT should be used, but since the nature of the entire setup guarantees that there are no proxies that could log the credentials, I would consider that to be largely stylistic.

All of the above items negatively impact on the usability of the device, as is often the case with security features. If there is anything that I would consider "worthwhile low-hanging fruit" in the security area, it would be to have an *option* in the module to use WPA2, with credentials loaded from flash. That would provide both user authentication as well as encryption, and assuming the device shuts down the AP once the configuration is done would not be subject to idle WPA-cracking.

@marcelstoer, does the CI builder do really clever things like enable modules which are off by default and test build those too? I'm a complete noob when it comes to Travis.

marcelstoer:
Thanks Johny for spelling out everything that I was too lazy to put down.

> does the CI builder do really clever things

Let's discuss that elsewhere, ok (e.g. PM me)? That issue's comment list is long enough already ;-)

grahamehorner:
please consider the use case where you have many 10s 100s or 1000s of IOT devices getting deployed; initially these devices will all be insecure with little or known WPA configured passwords; remote configuration of these devices without securing the information could result in leaked sensitive information above and beyond that of WEP keys etc.

I would also suggest hiding the said while in configuration mode and/or so only the end user is aware the AP exists.

This of cases becomes more critical when doing mass distributed deployments and configuration over public networks that may have additional config items that a hack cause issues with

TerryE:
@grahamehorner, Grahame most of the points that you make are valid, but we are trying to square a circle here.  The esp8266 is extremely resource limited. We've 3 RAM slots -- one used for mirrored ROM and **.bss**; one to cache the firmware being loaded of SPI flash; one for heap and C stack.  There is also the SPI flash itself.  On the current ESP-01 class modules we are against the limits of all of them.  Using with approach with SSL would require loading this module into the firmware build as well as the crypto and ssl stacks.  This will bankrupt our hardware resources, and the build will be unusable.

However, my personal preference is for a different provisioning architecture which doesn't impose additional firmware demands, or require end-users to connect to devices with some HTTP-serviced configuration screen .  I build my chips with a preconfigured SPIFFS.  This includes an initialisation Lua module which "calls home" to a predefined IP address / port passing info including the CPU  ID.  This then downloads specific to device modules and configuration data which is loaded into the SPIFFS to configure the  device to purpose.  All this requires is simple Lua provisioning script running on something like an RPi on the network and everything is done automatically when any uninitialised esp8266 is first powered up.  No encryption, but content authentication using an MD5 signature based on content and a shared secret.  

However, each to his own.  I can see the arguments for this type of approach that may suit many users, so I whilst I don't support it, I will not blackball it.

grahamehorner:
@TerryE thanks; I was starting to think I was talking to a brick wall to with regards the issues raised to be honest; I do understand the constraints and memory footprint of these devices; and was hoping a sensible discussion around alternatives and options on securing data in a different way to result; and a wider topic of IOT security models where SSL may not be available I feel is needed across many projects.

I personally agree with you assessment with regards to configuration approach; a phone home style approach is/would be more secure and scalable; MD5 of some shared secret + (salt) CPU ID + date time would be good enough an access-token which can then be used to do an OTA update of configuration that is secured by the shared secret + CPU ID

I'm not black balling the approach either way; just I felt it needed more thought on the type of information sent over the HTTP GET and weather it was correct to do this without ANY sort of security or protection of the information?  The best risk to systems is after all the internal/semi-trust users messing with things, and when looking to place the ESP8266 in the wild on public networks; hmm... we have many user cases to think about, maybe looking at remote configuration/deployments needs to be done on a case-by-case basis.

robertfoss:
So, the SSID define has now been moved to user_config.h and the HTTP GET has been replaced with a POST.

jmattsson:
Great! But doesn't the `c_strncmp("GET "...)` in `enduser_setup_http_recvcb()` prevent any POST requests from being handled? I admit I haven't actually *tried* it, so I could be off the mark here...

robertfoss:
Yeah, something is wrong here. I tested the code. And it did indeed work.
But the code does not look right to me either.

robertfoss:
@jmattsson Glad you spotted it, because I couldn't reproduce a working version using POST. I must've done something stupid during testing. The patch also cleans up HTTP request handling a slight bit.

jmattsson:
Merged. Thanks for your work! Please add usage docs on [the wiki](https://github.com/nodemcu/nodemcu-firmware/wiki/nodemcu_api_en-%28dev096%29) so people know how to enjoy your efforts :)

marcelstoer:
@robertfoss Did you document this anywhere? I didn't find it in the wiki.

robertfoss:
I documented it in the link provided by Johny.

robertfoss:
I added some info to https://github.com/nodemcu/nodemcu-firmware/wiki/nodemcu_api_en-%28dev%29---partial!#enduser_setup-module

PR #661, opened 2015-09-22
Added tweetnacl module

abgoyal:
Currently only crypto_box and crypto_secretbox apis are supported.


marcelstoer:
Can you please explain why you think this should be available on a resource-constrained IoT device? I don't see the use case yet. Would you agree that this module should be disabled in `user_modules.h` by default?

TerryE:
@marcelstoer, @abgoyal, this again underlines the need for a PECL-style approach to non-core modules.  If someone has gone to the effort to develop or port a module, test it and has a usecase, then it seems a waste not to make it available as some form of snippet if the contributor is willing to support it.

However, we should separate all of these modules and the support issues and distraction that thy cause from the mailine core nodeMCU Lua development.

abgoyal:
* @marcelstoer , I am perfectly fine with this being disabled by default.

* On my build this adds 7104 bytes - less than 7KB - to the final .bin, so its not exceptionally large. That was the entire idea behind using (tweet)nacl. A known-good, fully-featured authenticated encryption library with a very small footprint. (as an aside: I think in its (admittedly short) history, there has been exactly one bug found in it, unlike ssl).

* I need it myself, and while searching around, I came across a fair few other comments expressing interest in a nacl/esp8266 port - someone had even tried, but that project is defunct. So I know its needed by others. I coded it up and offer it up in the hope that it will be useful to someone else - its got quite a large number of use cases not served by ssl.

* @TerryE I agree. I submitted a pull-request to the mainline cause that seems to be the only choice at present. If there were a "modules" repository - I'd be happy to move it there. As I am also working on making this work with main-line Lua, I'd anyway have to host it independently. 

I'd just like to add that reworking the build system to allow "overlays" of modules (the android "repo" tool is a perfect example) would help in such scenarios.

TerryE:
> I'd just like to add that reworking the build system to allow "overlays" of modules (the android "repo" tool is a perfect example) would help in such scenarios.

+1, but I've already got half a dozen priority items on my TODO list, and only limited hobby time. :disappointed: It's a pity thet the nodeMCU crew seem to have gone fishin' !

marcelstoer:
> If someone has gone to the effort to develop or port a module, test it and has a usecase, then it seems a waste not to make it available

Absolutely, I fully agree.

I'm just concerned that in the not too distant future the default binary available from this repository won't be useful to anyone because it leaves no memory for program code. It doesn't matter to me personally because I use my cloud build service but it may matter to lots of others (even though my service seems popular).

devsaurus:
@marcelstoer I share your concern. We'll run into this situation sooner than later. The creativity and imagination of engineers, devs and users is unlimited while the ESP's memory is...
I'd also gladly support a modularization approach. If we'd had a clear road to follow, though.

abgoyal:
I have added in a second commit, disabling the module in user_modules.h. 

This should allow merging the PR without impacting the default build.

Issue #676, opened 2015-10-04
onewire power problems - bus left in bad state


nickandrew:
I'm working on a pull request to fix this.

The onewire_write() call with power set to 0 changes the pin mode to OUTPUT and a low voltage. This is bad because the bus should normally be high, and low is used to signal the start of a transaction.

Examples:

```c
void onewire_skip(uint8_t pin)
{
    onewire_write(pin, 0xCC, 0);           // Skip ROM
}
```

Looking at the code for onewire_write() ...

```c
  if ( !power) {
    noInterrupts();
    DIRECT_MODE_INPUT(pin);
    DIRECT_WRITE_LOW(pin);
    interrupts();
  }
```

There seems to be a problem here. DIRECT_MODE_INPUT is defined as `GPIO_DIS_OUTPUT(pin_num[pin])` which disables output for the pin. But DIRECT_WRITE_LOW calls GPIO_OUTPUT_SET which re-enables output!

Here's my plan to fix. In the code blocks which test `power`:

```diff
   if (!power) {
     noInterrupts();
-    DIRECT_MODE_INPUT(pin);
-    DIRECT_WRITE_LOW(pin);
+       platform_gpio_mode(pin, PLATFORM_GPIO_INPUT, PLATFORM_GPIO_FLOAT);
     interrupts();
   }
```

Which I think was what was intended. The onewire_write() calls with power set to 0 in the firmware should change to 1. If you really need no power, call them yourself. There's also onewire_depower() which I assume is intended to remove all power from the bus. This won't work if there's a recommended weak pullup resistor on the bus.

Attached is my logic analyser trace of the problem. 1wire is the 1-wire bus. Channel-4 is used to trigger the trace, and distinguish the test phases.

Phase 1: bus reset until about 800 us.

Phase 2: ow.write(pin, 0xcc, 1) leaving bus high at about the 1.9 ms mark

Phase 3: ow.write(pin, 0xcc, 0) leaving the bus low. All 1-wire devices on the bus will reset which is not good, because 0xcc is supposed to leave all devices selected.

Here's the test script I used:

```lua
function owtest2()
    pin = 1
    ow.setup(pin)
    tmr.delay(50000)
    gpio.mode(4, gpio.OUTPUT)
    gpio.write(4, gpio.LOW)
    ow.reset(pin)
    ow.write(pin, 0xcc, 1)
    gpio.write(4, gpio.HIGH)
    ow.write(pin, 0xcc, 0)
[...]

PR #680, opened 2015-10-05
Fix bus power problems in onewire

nickandrew:
See #676

The 1-wire pin was being set back to output low, which reset all 1-wire
devices on the bus.

Change the power == 0 behaviour to tri-state the pin, as I believe was
intended. Make this not the default, because the bus should be high
when a bit/byte is not being read/written. A low bus for a long time
is a reset condition.


nickandrew:
I don't have a LA trace of the exact same test, but here is one of the successful reset, skip rom, convert T commands.
onewire-start-conversion

PR #699, opened 2015-10-22
Implement SDK function wifi_status_led_install

Aeprox:
This PR implements the new SDK functions wifi_status_led_install(pin,muxname,pinfunc ) and wifi_status_led_uninstall(). I've only tested some of these pins, can't test them all, but they should work. In all honesty it's quite lackluster (but energy efficient) at the moment, just blinks while trying to connect, then stays off. Might get upgraded in newer SDKs though

### Description
Enables/disables output of wifi status to an LED connected to the specified pin
### Syntax
wifi.setstatusled(pin)

### Parameters
pin: Pin number (only allows 0,4,5,12,13 and 14). Use nil to remove the LED

### Returns
nil

### Example

```lua
    wifi.setstatusled(4) -- Connect LED to GPIO4, it should flash when trying to connect

    wifi.setstatusled(nil) -- disables status output
```


eku:
I suggest to change the LUA api. wifi.uninstallstatusled() is not needed, if wifi.installstatusled() would accept ´´´nil´´´ as uninstall.

Aeprox:
I fully agree, see newest commit. I kept using the names given in the SDK to prevent confusion but it's a simple feature anyway.

eku:
@Aeprox please request, since app/modules/mdns.c should not be part of the PR.

Aeprox:
Heh, sorry about that! It's been a long week.

PR #708, opened 2015-10-30
Added ability to use #DEFINE to select default baud rate.

karrots:
Adds ability to use a C preprocessor define to select baud rate for non-development versions.

Conflicts:
	app/user/user_main.c


dnc40085:
where is BIT_RATE_DEF supposed to be defined? in user_modules.h?

TerryE:
+1. We should keep the code simple and add a line to `ùser_config.h`(as this is nothing to do with module functionatity)
```C
#ifdef DEVELOP_VERSION
#define BIT_RATE_DEFAULT  BIT_RATE_74880
#else
#define BIT_RATE_DEFAULT BIT_RATE_9600 
#endif
````
and unconditionally use `BIT_RATE_DEFAULT` in `user_main.c` in the uart_init.  There no need to include `uart.h` anywhere else as this is only bound in compilation in user_main.

**Comment by karrots** [comment]
Isn't this considered redefining a macro and thus a compile error?

**Comment by karrots** [comment]
Threw in a #undef before the second #define feels a bit dirty but it compiles.

**Comment by TerryE** [comment]
??

```C
#if SOME_PREPROC_COND
#define SOME_DEF a
#else
#define SOME_DEF b
#endif
```

is perfectly valid.  The define is only executed once since only once path is taken.  Unless of course you've been tweaking the make file to use  `-DSOME_DEF=a`.  **Don't do this** with the nodeMCU make.  You tailor the build through the user files.

karrots:
Sorry you are correct I did copy a `-DSOME_DEV=a` from my build script.

The original thought behind the mod was to enable cli level bit rate selection. It was to enable easier iteration in a build script which was building a couple different firmwares with different rates for testing a serial issue. I think I'll close this PR and open a new one with all the lessons learned and discussed here.

PR #713, opened 2015-10-31
Added support for the hx711 load cell ADC.

christakahashi:
Adds basic support (chA@128gain only, which is the most useful mode anyway) for the hx711 load cell ADC.  These are pretty cheap and popular for hacking weight scales so I figure the community at large might benefit from my code.


devsaurus:
What prevents you from using the SPI port instead of SW bit-banging? The comm protocol looks very similar to SPI.

christakahashi:
The HX711 doesn't tristate it's output and has no CS pin (also you need to send 25,26, or 27 bits which wasn't possible when i started, but looks to be possible now).  The data out is always driven while awake and the datasheet doesn't say if the output is high or low impedance during powerdown.  Also since there's no CS (power down command is given by pulling and maintaining clk high for 60uS) any clk signals not intended for it would still wake the device and it would start to drive MISO.

Also, My understanding is the HW SPI port is in use by the EEPROM, so it would for sure have to share.

If there's a better solution that doesn't require adding code that would be optimal IMO.  But I don't see it right now.  I did notice that the gpio module has an undocumented but more general bitbangy mode ("serout" function) but it doesn't look like that can do any reading.

devsaurus:
The HSPI module would restrict you to specific pins, for sure.

Apart from that, please have a look at your WDT handling. This PR added another instance of `WRITE_PERI_REG(0x60000914, 0x73);` which were meant to be replaced recently in #703.

marcelstoer:
Pity this was already merged by @vowstar even though this discussion was still ongoing.

PR #722, opened 2015-11-03
Moving luac cross from master to dev

TerryE:
Note that I've had to do a few tweaks to interoperate with the SDK 1.4 changes, but other than that  it seems to work fine.  You need lua-5.1 and lua-filesystem installed on your build host, but other than that doing a `lua tools/cross-lua.lua` builds luac.cross fine.

### DIFF
--- a/app/include/user_modules.h
+++ b/app/include/user_modules.h
@@ -8,7 +8,8 @@
 // #define LUA_USE_BUILTIN_IO 			// for io.xxx(), partially work
 
 // #define LUA_USE_BUILTIN_OS			// for os.xxx(), not work
-// #define LUA_USE_BUILTIN_DEBUG		// for debug.xxx(), not work
+// #define LUA_USE_BUILTIN_DEBUG
+#define LUA_USE_BUILTIN_DEBUG_MINIMAL // for debug.getregistry() and debug.traceback()
 
 #define LUA_USE_MODULES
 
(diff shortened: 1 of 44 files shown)


jmattsson:
Before I merge it, a couple of points:
  - There are a couple of added #if 0 statements. Would it be better to simply nuke that code instead to keep things clean? It's all under version control after all, and easy to pull back if needed.
  - I'd love to see the luac.cross build integrated into the CI build too, so we get ongoing coverage of that. In my experience, things that don't get regularly built somehow develop bit-rot.

TerryE:
> There are a couple of added #if 0 statements.
The `#if 0` get mixed reactions from developers.  I like you would tend to deprecate this, but it is the coding style that the ndeMCU guys used when removing chunks of the eLua code to create the first ndeMCU baseline.  If you look some of the changes were replacing `#if 0` with conditionals for the cross compile variant.  To be honest this reconciliation was an easier exercise with the `#if 0` technique than having removed the lines.

The couple of `#if 0`s that I introduced were in the reduced `math` library that the firmware supports.  These came out of a cross reference analysis as part of our effort to removed unused or duplicated generate code in the firmware.   If you look zeroday used this `#if 0` technique to comment out the mah API calls that weren't being implemented in the firmware, but left in a couple of the called static routines that were no longer being used but were still generating code and more to the point pulling in C library math routines.  Again I felt it a lot less confusing to stick with zerday's practice than to use mixed styles in the same module.

jmattsson:
Cool, sounds like there's a good reason for sticking with #if 0. Are you wanting to add the travis support to this PR, or a new PR later? I'm happy with either.

TerryE:
No.  Commit this one when you have a chance.  I'll do the travis support as a separate pass.

TerryE:
Thanks Johny, I'll rebaseline and do the `strip_debug()` PR next and then sort out travis for `luac.cross`

Issue #721, opened 2015-11-03
Problem with powersaving


NiklausS:
Hi,
just upgraded to from 0.9.5 to 0.9.6 and found that esp does not go to LIGHT_SLEEP or MODEM_SLEEP after dsleep -> the powerconsumption remains at 50mA with 0.9.5 this was working fine.

tested with:
ok - nodemcu_float_0.9.5_20150318.bin  (NodeMCU 0.9.5 build 20150318  powered by Lua 5.1.4)
bad - nodemcu-dev-18-modules-2015-10-27-15-11-48-float.bin (Lua 5.1.4 on SDK 1.4.0)
bad - nodemcu-dev140-19-modules-2015-10-28-20-33-03-float.bin

```Lua
tmr.alarm(1, gsleep*1000, 0, function()
print("timer1 trigger dsleep - "..node.heap())
node.dsleep(1000000, 4)
end)
```

Thanks for the greate software.
 Nik

dnc40085:
I'm having the same issue with the current dev branch(SDK ver 1.4.0).
The SDK function `deep_sleep_set_option()` used by `node.dsleep()`, is returning a 0, which indicates a failure according to the SDK programming guide.

dnc40085:
The new function `wifi.sleep` should help you accomplish your task until the `node.dsleep` issue is resolved

NiklausS:
Thanks,
already tested it yesterday and it works so far.
I've places a wifi.sleep(0) at line 1 in the init.lua. I see a small  peek of approx. 60mA. 
So far this works as workaround.
I guess this goes in line with the readvdd33(), returning always 65535 in dev branches.

PR #725, opened 2015-11-04
Add function wifi.sleep

dnc40085:
This pull request adds functionality to force wifi radio into sleep mode without restart.

This is also workaround for the deep sleep issue #721 where executing `node.dsleep(us, 4)` does not disable radio upon waking up

#### Description
Force wifi radio into sleep mode.

Note:
* if wifi radio was disabled using node.dsleep(us, 4) executing wifi.sleep(0) WILL NOT re-enable wifi radio.
* Wifi mode MUST BE SET after re-enabling wifi radio, since wifi.sleep(1) sets wifi mode to NULL.

#### Syntax
wifi.sleep()
wifi.sleep(desired_sleep_state)

#### Parameters
desired_sleep_state:  (if sleep state was changed, return new sleep state. otherwise return current sleep state)
* 0 : Disable sleep mode
* 1 : Enable sleep mode
* nil : Do not change sleep state, return current sleep state.

#### Returns
Current WiFi sleep mode

#### Example

```lua
    --get current wifi sleep state
    print("wifi sleep mode is currently set to"..wifi.sleep())
    
    --enable forced wifi radio sleep
    print("forced wifi sleep mode is now set to "..wifi.sleep(1))

    --disable forced wifi radio sleep
    print("forced wifi sleep mode is now set to "..wifi.sleep(0))
    wifi.setmode(wifi.STATION) --WIFI MODE MUST BE SET AFTER WIFI RADIO IS RE-ENABLED
    wifi.sta.connect()
```

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -20,7 +20,7 @@ static uint8 getap_output_format=0;
 
 //wifi.sleep variables
 #define FPM_SLEEP_MAX_TIME 0xFFFFFFF
-bool FLAG_wifi_force_sleep_enabled=0;
+static bool FLAG_wifi_force_sleep_enabled=0;
 
 
 //variables for wifi event monitor
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -1383,7 +1383,7 @@ const LUA_REG_TYPE wifi_map[] =
   { LSTRKEY( "sta" ), LROVAL( wifi_station_map ) },
   { LSTRKEY( "ap" ), LROVAL( wifi_ap_map ) },
 
-  // { LSTRKEY( "NULLMODE" ), LNUMVAL( NULL_MODE ) },
+  { LSTRKEY( "NULLMODE" ), LNUMVAL( NULL_MODE ) },
   { LSTRKEY( "STATION" ), LNUMVAL( STATION_MODE ) },
   { LSTRKEY( "SOFTAP" ), LNUMVAL( SOFTAP_MODE ) },
   { LSTRKEY( "STATIONAP" ), LNUMVAL( STATIONAP_MODE ) },
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -18,6 +18,11 @@
 static int wifi_smart_succeed = LUA_NOREF;
 static uint8 getap_output_format=0;
 
+//wifi.sleep variables
+#define FPM_SLEEP_MAX_TIME 0xFFFFFFF
+bool FLAG_wifi_force_sleep_enabled=0;
+
+
 //variables for wifi event monitor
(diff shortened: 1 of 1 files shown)


jmattsson:
I like the idea here, but it feels like there's a mismatch between the exposed Lua API and the underlying SDK API. Using wifi.sleep(1) will only put the radio to sleep for a maximum of 268 seconds (less than 5min) if I'm reading this correctly. The expectation I would have as a user would be that the wifi is put to sleep until I call wifi.sleep(0).

dnc40085:
Judging by what the SDK programming guide(v 1.4.0) says on page 86 under the function `wifi_fpm_do_sleep` if the sleep time is set to 0xFFFFFFF and `wifi_fpm_set_sleep_type` is set to MODEM_SLEEP_T. The wifi radio will only wake by executing the function `wifi_fpm_do_wakeup`.

Also, I've found that in my own testing the wifi radio stays asleep much longer than 268.4 seconds.

jmattsson:
That's what I get for not RTFM'ing the full thing and only looking at user_interface.h. This is looking pretty good then! My only question is whether there is any particular reason why the new flag variable isn't declared static?

Unless someone has objections, I'd be happy to merge this in this week.

TerryE:
+1 It should be `static bool FLAG_wifi_force_sleep_enabled=0;`  I'd better go through my luac.cross patch and make sure that I've not missed any! :smile:

dnc40085:
> My only question is whether there is any particular reason why the new flag variable isn't declared static?

Oops, I forgot to add the keyword, thanks for pointing that out.

I was thinking of adding code to return an error if `wifi_fpm_do_sleep` does not return a 0. I left it out to keep the code as short as possible, maybe it could be added for debugging purposes with `#ifdef DEVELOP_VERSION`.

Any other ideas or suggestions?

TerryE:
> I was thinking of adding code to return an error if wifi_fpm_do_sleep does not return a 0.

As a general rule if a system routine returns a status, then we should not ignore this.  Doing so causes silent failures that are incredibly difficult to ignore.  So either a lua_error() or a return status IMO, and the latter is probably easier for everybody.

dnc40085:
> So either a lua_error() or a return status IMO, and the latter is probably easier for everybody.

how about adding the return value so it is also returned with current wifi status similar to this: `new_wifi_sleep_state, wifi_fpm_do_sleep_retval=wifi.sleep(1)`

TerryE:
My ideas was similar. Why not follow the `pcall()`  model.  Either the return is `new_wifi_sleep_state, nil` on a normal return or `nil, error_return` on an invalid return.  This is upwards compatible with sloppy programs that only do a:

```Lua
new_wifi_sleep_state = wifi.sleep(1)
```

dnc40085:
@TerryE I made the changes you suggested and submitted a PR(#731)

PR #729, opened 2015-11-05
Fix to do TCP port number randomization at boot.

jmattsson:
This addresses the issue where a server would otherwise reject a valid
connection attempt from a rebooted node, due to the server thinking that
packet belonged to a recently closed session. By starting off the local
TCP port numbers at random* points each boot the risk of port reuse
across a reboot is significantly reduced.

*) As random as can be had via the lower 12 bits of the ccounter.

I had hoped Espressif would fix this in the SDK and open-source LWIP, but I haven't managed to convince them to do so, so I guess NodeMCU will be the only firmware with this issue fixed...

### DIFF
--- a/app/lwip/core/tcp.c
+++ b/app/lwip/core/tcp.c
@@ -635,7 +635,7 @@ tcp_new_port(void)
 #define TCP_LOCAL_PORT_RANGE_START 4096
 #define TCP_LOCAL_PORT_RANGE_END   0x7fff
 #endif
-  static u16_t port = TCP_LOCAL_PORT_RANGE_START;
+  static u16_t port __attribute__((section(".port"))) = TCP_LOCAL_PORT_RANGE_START;
   
  again:
   if (++port >= TCP_LOCAL_PORT_RANGE_END) {
--- a/app/user/user_main.c
+++ b/app/user/user_main.c
@@ -51,6 +51,21 @@ void TEXT_SECTION_ATTR user_start_trampoline (void)
 }
 
 
+/* To avoid accidentally losing the fix for the TCP port randomization
+ * during an LWIP upgrade, we've implemented most it outside the LWIP
+ * source itself. This enables us to test for the presence of the fix
+ * /at link time/ and error out if it's been lost.
+ * The fix itself consists of putting the function-static 'port' variable
+ * into its own section, and get the linker to provide an alias for it.
+ * From there we can then manually randomize it at boot.
+ */
+static inline void tcp_random_port_init (void)
+{
+  extern uint16_t _tcp_new_port_port; // provided by the linker script
+  _tcp_new_port_port += xthal_get_ccount () % 4096;
+}
+
+
 void task_lua(os_event_t *e){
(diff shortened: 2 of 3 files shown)


jmattsson:
Any objections to merging this?

TerryE:
I would prefer a non-pseudo stochastic fix, but beggers can't be choosers :wink:

jmattsson:
If/when I find the time to pick off a certain item on my TODO list, I might revisit this to improve it a little bit further.

devsaurus:
I've been forwarding my local repo past this merge and ended up with a non-functional build. At 748800 baud I get nothing but junk:
```
PORT OPEN 74880

 ets Jan  8 2013,rst cause:1, boot mode:(3,0)

load 0x40100000, len 28248, room 16 
tail 8
chksum 0x50
load 0x3ffe8000, len 2368, room 0 
tail 0
chksum 0x55
load 0x3ffe8940, len 8, room 8 
tail 0
chksum 0xf5
csum 0xf5

SDK v��
SDK v��
SDK v��
...
```
This fault can be avoided when `tcp_random_port_init ()` is commented in `nodemcu_init()`. But, what's going wrong? I couldn't find any obvious issue in the merge.

devsaurus:
I get this error from frightanic's build service as well.
> http://frightanic.com/nodemcu-custom-build/builds/nodemcu-dev-11-modules-2015-11-15-20-23-43-integer.bin
> This was built against the dev branch and includes the following modules: node, file, gpio, wifi, net, i2c, spi, tmr, uart, bit, ws2812.

Seems that current dev branch is somehow broken.

jmattsson:
I just checked out & built current dev 340edbbe2fe89fcef60a77e76d8f58163315bfe4, which runs without any issues on my devkit 0.9 here, both float & integer versions, default modules.

However, the pre-built firmware you linked to goes cyclic on boot for me too. Have you tried with fewer/different modules, and does it make a difference? Could give me access to your non-working firmware, the app/mapfile, and your user_modules.h + user_config.h somewhere? I'll try to reproduce the issue here (and fix if I can reproduce). It seems most strange that the `tcp_random_port_init()` could cause this, but clearly *something* is breaking *somehow*. The failure mode reminds me of when I marginally outgrew the space on my ESP-01, but that shouldn't be an issue on the ESP-12... /headscratch

devsaurus:
I see this on a plain ESP-03.

It happens for a very basic module selection of `node, file, gpio, wifi, net, i2c, spi, tmr, adc, uart, bit`. Integer build with `CLIENT_SSL_ENABLE` disabled.
Strangely, same build with active `CLIENT_SSL_ENABLE` works doesn't exhibit this issue!

Files of a failing configuration with a more complex setup are available from here (though again without SSL):
https://gist.github.com/devsaurus/9033a48d6f5c923408b9

jmattsson:
Alright, I can reproduce the problem if I comment out CLIENT_SSL_ENABLE. Now thoroughly confused. Will fix. Somehow.

jmattsson:
Okay, it's an alignment issue. This should be easy...

PR #731, opened 2015-11-05
Update function wifi.sleep

dnc40085:
Update function wifi.sleep to return any error codes to user and improve code readability

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -357,7 +357,7 @@ static int wifi_sleep(lua_State* L)
 	{
 	  FLAG_wifi_force_sleep_enabled = TRUE;
 	}
-	else if (wifi_fpm_do_sleep_return_value != 0)
+	else
 	{
 		wifi_fpm_close();
 		FLAG_wifi_force_sleep_enabled = FALSE;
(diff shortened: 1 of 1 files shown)


app/modules/wifi.c
```diff
@@ -352,17 +352,37 @@ static int wifi_sleep(lua_State* L)
 	// set force sleep type
 	wifi_fpm_set_sleep_type(MODEM_SLEEP_T);
 	wifi_fpm_open();
-	wifi_fpm_do_sleep(FPM_SLEEP_MAX_TIME);
+	wifi_fpm_do_sleep_return_value = wifi_fpm_do_sleep(FPM_SLEEP_MAX_TIME);
+	if (wifi_fpm_do_sleep_return_value == 0)
+	{
+	  FLAG_wifi_force_sleep_enabled = TRUE;
+	}
+	else if (wifi_fpm_do_sleep_return_value != 0)
```
jmattsson: Minor comment, this should probably be a plain "else", not an "else if".

jmattsson:
This is looking fine to me. I will merge unless someone objects.

PR #733, opened 2015-11-05
Contributing guideline update

marcelstoer:
Feel free to suggest corrections or merge and add them yourself :wink: .

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,78 @@
-## Contributing
+# Contributing to NodeMCU
 
-Thank you everyone for contributing to this project.
+:+1::tada: First off, thanks for taking the time to contribute! :tada::+1:
 
-Pull requests for new features and major fixes should be opened against the `dev` branch.
+We're really glad you're reading this, because we need volunteer developers to help this project come to fruition.
 
-Avoid intermediate merge commits. Rebase your feature branch onto `dev` to pull updates and
-verify your local changes against them before placing the pull request.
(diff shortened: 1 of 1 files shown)


TerryE:
@marcelstoer Marcel, one suggestion reading this. I think that this is one very valid case where the contributor should raise an issue, discussing the proposed change; emphasising that the contributor is proposing to develop the patch her or himself, and outling the strategy for implementation.  This type of discussion is what we *should* be doing on the issues list and it is better to do this before or in parallel to developing the patch rather than having "you should have done it this way" type of feedback on the PR itself.

jmattsson:
Terry raises an interesting point, though I'm in favour of merging in this update sooner rather than later. Small steps / release often, and all that :)

devsaurus:
@marcelstoer Very helpful guide to contribution! Would have saved me a lot of headaches while learning git &  earlier this year.

@TerryE Thanks for clarifying this. Good to set a common approach here as well.

marcelstoer:
Thanks for the feedback, much appreciated.

@TerryE I don't fully agree with you but I'll of course comply if there's consensus among the collaborators. In any case, your comments (which I could add to this PR) should IMO only apply to "major" - someone define that - changes to either code, documentation or process.

@jmattsson yeah, yeah...exactly my world :smile: 

@devsaurus I'm still learning the Git & GitHub part myself - the hard way. Often I feel like a toddler learning to walk. I doesn't really help that there are many different ways to achieve more or less the same. Git's just got too many options for that.

TerryE:
@marcelstoer I am happy with the stepwise approach, and we have broad consensus hence the commit.  I also agree that raising an early issue should be optional.

PR #742, opened 2015-11-09
Port of strip debug patch to the SDK 1.4-baseline dev

TerryE:
I begin to see the hassle of keeping the documentation in a wiki which doesn't share common branches with the repository.   I am mainly issuing this PR so that developers can evaluate it.  Even through its not up to master (or what should be master threshold), it is dev quality so I leave it to the other committers to decide whether to pull it into dev in its current form.  

There is a whitepaper explaining the background and use on the esp8266 site here: [Lua Compact Debug (LCD)](http://www.esp8266.com/wiki/doku.php?id=nodemcu-lcd) and the APi hook is  `node.stripdebug()`

```
 node.stripdebug([level[, function]]). 
 level:    1 don't discard debug
           2 discard Local and Upvalue debug info
           3 discard Local, Upvalue and lineno debug info.
 function: Function to be stripped as per setfenv except 0 not permitted.
 If no arguments then the current default setting is returned.
 If function is omitted, this is the default setting for future compiles
 The function returns an estimated integer count of the bytes stripped.
```
You need to set `LUA_OPTIMIZE_DEBUG` as per whitepaper to include this code.  Essentially I recommend level 2 which gives you almost the same memory savings as node.compile but with full line info and stack traceback: a very small memory  overhead for an extremely useful diagnostic capability.


jmattsson:
I think you're really the expert here on the Lua internals Terry, so I'm very inclined to follow your recommendation. What about this patch set makes you say it's only dev quality? Nothing stood out when I had a quick look through.

TerryE:
> What about this patch set makes you say it's only dev quality?

Testing it was a bitch because of all the edge cases.  Have a quick read through my LCD paper.  I used Bogdan Marinescu's [LTR](http://www.eluaproject.net/doc/master/en_arch_ltr.html) patch as an example, but doing decent path analysis on nodemcu or eLua proved totally impractical so I first developed the patch on a standard Lua build so I could use the full [Lua 5.1 Test suite](http://www.lua.org/tests/) to shake down the edge cases (like how is handles backout during compile error).  I am pretty sure its solid.   In my view we should:

*    **promote to `dev`** when the patch is simple / low risk and or reviewed by another developer with karma.
*     **promote fro `dev` to `master` only if (i) low risk and priority or (ii) after its been in `dev` a few weeks minimum to have allowed other to have downloaded and tried out the code.

Of course the current `master` actually contains a lot of dubious patches and we need to review and rebaseline `master` and `dev` to get this relationship baselined but that's a separate one-off exercise on my TODO list.

jmattsson:
I don't think it will get more extensive testing by not getting it onto dev though :)

In terms of promotion I'm personally wary of doing selective promotion. My preferred approach is more along the FreeBSD approach of a period of code freeze (except for bug fixing, naturally) before a release. Just my $0.02 to keep things simple.

PR #746, opened 2015-11-10
Remove adc.readvdd33() limitation.

### LINKED ISSUE #433: adc.readvdd33() kills network?

jmattsson:
With the change to use the new system_get_vdd33() function for reading vdd33, it became possible to read it while connected to WiFi. As such, the explicit prevention of doing so is no longer needed nor wanted. Getting vdd33 is now possible regardless of whether the WiFi is connected (since system_get_vdd33 internally suspends WiFi while reading the vdd).

Note that system_get_vdd33() checks byte 107 of esp_init_data_default.bin, so this byte MUST be changed to 0xFF to get actual ADC readings (otherwise you'll only see 65535 returned). For whatever reason the SDK-provided esp_init_data_default.bin does not have this set.

This should thoroughly close #433, btw.

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -11,11 +11,13 @@
 // #define FLASH_16M
 #define FLASH_AUTOSIZE
 #define FLASH_SAFE_API
+
+// Byte 107 of esp_init_data_default, only one of these 3 can be picked
+#define ESP_INIT_DATA_ENABLE_READVDD33
+//#define ESP_INIT_DATA_ENABLE_READADC
+//#define ESP_INIT_DATA_FIXED_VDD33_VALUE 33
+
 // #define DEVELOP_VERSION
-#define FULL_VERSION_FOR_USER
-
-#define USE_OPTIMIZE_PRINTF
-
 #ifdef DEVELOP_VERSION
 #define NODE_DEBUG
 #define COAP_DEBUG
--- a/app/platform/flash_api.c
+++ b/app/platform/flash_api.c
@@ -8,7 +8,17 @@
 #include "spi_flash.h"
 #include "c_stdio.h"
 
-static volatile const uint8_t flash_init_data[128] ICACHE_STORE_ATTR ICACHE_RODATA_ATTR =
+#if defined(ESP_INIT_DATA_ENABLE_READVDD33)
+# define INIT_107 0xff
+#elif defined(ESP_INIT_DATA_ENABLE_READADC)
+# define INIT_107 0x00
+#elif defined(ESP_INIT_DATA_FIXED_VDD33_VALUE)
+# define INIT_107 ESP_INIT_DATA_FIXED_VDD33_VALUE
+#else
+# define INIT_107 0xff
+#endif
+
+static const uint8_t flash_init_data[128] =
 {
     0x05, 0x00, 0x04, 0x02, 0x05, 0x05, 0x05, 0x02, 0x05, 0x00, 0x04, 0x05, 0x05, 0x04, 0x05, 0x05,
     0x04, 0xFE, 0xFD, 0xFF, 0xF0, 0xF0, 0xF0, 0xE0, 0xE0, 0xE0, 0xE1, 0x0A, 0xFF, 0xFF, 0xF8, 0x00,
@@ -16,7 +26,7 @@ static volatile const uint8_t flash_init_data[128] ICACHE_STORE_ATTR ICACHE_RODA
     0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     0xE1, 0x0A, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x93, 0x43, 0x00, 0x00, 0x00,
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, INIT_107, 0x00, 0x00, 0x00, 0x00,
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
 };
 
@@ -316,26 +326,13 @@ bool flash_rom_set_speed(uint32_t speed)
     return true;
 }
 
-bool flash_init_data_written(void)
-{
-    // FLASH SEC - 4
-    uint32_t data[2] ICACHE_STORE_ATTR;
-#if defined(FLASH_SAFE_API)
-    if (SPI_FLASH_RESULT_OK == flash_safe_read((flash_safe_get_sec_num() - 4) * SPI_FLASH_SEC_SIZE, (uint32 *)data, sizeof(data)))
-#else
-    if (SPI_FLASH_RESULT_OK == spi_flash_read((flash_rom_get_sec_num() - 4) * SPI_FLASH_SEC_SIZE, (uint32 *)data, sizeof(data)))
-#endif // defined(FLASH_SAFE_API)
-    {
-        if (data[0] == 0xFFFFFFFF && data[1] == 0xFFFFFFFF)
-        {
-            return false;
-        }
-    }
-    return true;
-}
-
 bool flash_init_data_default(void)
 {
+    /* Can't copy directly from flash (which is where the default data lives)
+     * due to it being unmapped during the write, so bounce via ram buffer. */
+    uint8_t init_data[128];
+    os_memcpy (init_data, flash_init_data, 128);
+
     // FLASH SEC - 4
     // Dangerous, here are dinosaur infested!!!!!
     // Reboot required!!!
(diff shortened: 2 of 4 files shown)


dnc40085:
Shouldn't byte 107 of `flash_init_data` in `flash_api.c` also be changed to reflect this requirement?

jmattsson:
I wasn't even aware that there was a default block included in NodeMCU - I thought it was the SDK itself that provided it. Thanks for pointing that out!

Hmm, it does pose a bit of a dilemma given the mutually exclusive nature of readvdd33 vs adc.read. What do you think about making byte 107 depend on a #define in user_config.h?

dnc40085:
I found it when I was (unsuccessfully) trying to figure out how to implement spiffs in my own project.
The init data is written by `nodemcu_init()` in `main.c`, I'm guessing that it's done when the file system is initially setup.

I think it would be a good idea to make this option configurable, maybe readvdd33 could be the default setting.

jmattsson:
It turns out that the current SDK *also* lays down its own init data if there's none, before the NodeMCU `flash_init_data_written()` gets run. Thus, we don't ever end up using our default data. I'll see if I can untangle this and get a check in before the SDK does. If I can't, then we can at least simplify our code by removing all the init_data related stuff... :/

jmattsson:
Okay, the esp_init_data should now be both fully functional as well as fully consistent.

jmattsson:
Considering that the value is used as (x/10) Volts for the RF calibration, decimal seems the better choice (33 => 3.3V). The C compiler is happy dealing with either hex or decimal after all, so might as well make it as human readable as possible. Also, the ESP8266_RF_init.xls reference document uses decimal for the limits.

app/platform/flash_api.c
```diff
@@ -8,7 +8,17 @@
 #include "spi_flash.h"
 #include "c_stdio.h"
 
-static volatile const uint8_t flash_init_data[128] ICACHE_STORE_ATTR ICACHE_RODATA_ATTR =
```
vowstar: Are you sure remove  ICACHE_RODATA_ATTR ?

jmattsson:
Yes, since we these days:
  - place all const data in flash automatically, and
  - the flash read functions have been improved to handle unaligned reads

vowstar:
very cool!

TerryE:
> the flash read functions have been improved to handle unaligned reads

Yes, but at the (small) cost of an exception handler call in the case of an unaligned read, but in this case we are doing and os_memcpy of the data and the whole block defaults to aligned anyway, so the accesses will be aligned.

jmattsson:
I was actually thinking of the improvement made back in 1259f8d776a49f5daf1dbb3aff0c3643e377e8e0, but that of course is in the read-path, not the write path.

The NodeMCU built-in esp_init_data hasn't been possible to write out since the consts-to-flash change due to the flash being unmapped during the write (and thus reading+writing zeroes rather than the actual data). I'm not sure why nobody noticed this. Maybe even SDK 0.9.5 was writing out its own copy and thus masked this problem? Anyway, that too is fixed by this PR. Sorry about the breakage!

jmattsson:
Any further comments on this, or can we merge it?

eyaleb:
I started using this build and now I cannot `adc.read(0)`. I already had a patch to use `system_get_vdd33()` for Vdd which worked. I do need to read both.

 I now kept the default config of
```#define ESP_INIT_DATA_ENABLE_READVDD33```
and could not read the `adc`. Should I use a different setting if I want to read both? I tried the others and with both I could read the `adc` but not `vdd`.

I wonder why the exclusivity, I had no problem reading both in the past.

TIA

**Comment by jmattsson** [comment]
Please see the official SDK documentation. The VDD can only be read provided:
  - the TOUT (aka ADC) pin is floating (i.e. not connected); and
  - your esp_init_data_default has the correct setting in byte 107

In the old 0.9.6 SDK it was "possible" to read both, but neither could/would have been returning valid values afaiu.

**Comment by eyaleb** [comment]
@jmattsson 
Yes, this was always the story, but I was in fact getting good readings (checked against a multimeter). And I still am, having a number of modules still running now with the older fw (as recent as dev from 25/Nov).

Is there any doco on what actually goes on with the internal ADC (hwr wiring) and the content of byte 107 (swr)?

**Comment by jmattsson** [comment]
So you had something hooked up to TOUT all along, and still managed to get accurate vdd readings?

I wish Espressif were more forthcoming with documentation. The closest I've come in this area are a thread or two on their bbs, plus their SDK documents of course.

**Comment by eyaleb** [comment]
Did further testing. With this config
```
// Byte 107 of esp_init_data_default, only one of these 3 can be picked
//#define ESP_INIT_DATA_ENABLE_READVDD33
#define ESP_INIT_DATA_ENABLE_READADC
//#define ESP_INIT_DATA_FIXED_VDD33_VALUE 33
```
I am unable to read vdd (read 65535). However, if I change `app/modules/adc.c`
```
lua_pushinteger(L, system_get_vdd33 ());
```
to
```
lua_pushinteger(L, readvdd33 ());
```
I can read both. I was running this way for a few months without a problem.

So the question is: Do we really need to use `system_get_vdd33 ()`?
I know what `2C-ESP8266__SDK__Programming\ Guide__EN_v1.4.pdf` says, yet...

**Comment by jmattsson** [comment]
I'm on leave at the moment so I don't have access to an oscilloscope to confirm, but if so it would seem that:
  - we should expose `readvdd33()` somehow, possibly by evicting `system_get_vdd33()` again; and
  - we probably also should wrap `readvdd33()` in the same wifi on/off that `system_get_vdd33()` does to avoid crashing.

Any other takers on this front?

**Comment by eyaleb** [comment]
As I was investigating issues with OW (https://github.com/nodemcu/nodemcu-firmware/issues/859) I slowly reverted my changes (there was only one) and now it looks like the abovementioned call to `readvdd33()` is the cause of the problem (bad ow readings).

I do not know enough about the internals of the SDK (or the hwr) but not calling this function made my `ds18b20` readings reliable.

In short: do not proceed with the above proposal. I still need a way to read both `vdd33` and `tout`.

**Comment by fixingthingsguy** [comment]
Contributors All - If I was using a build service such as frightanic, how would I request an enable of either TOUT or VDD. I imagine choosing both is not possible, and is not necessary.
Regards

File: app/include/user_config.h [hunk 2]
```diff
@@ -11,11 +11,13 @@
 // #define FLASH_16M
 #define FLASH_AUTOSIZE
 #define FLASH_SAFE_API
-// #define DEVELOP_VERSION
-#define FULL_VERSION_FOR_USER
 
-#define USE_OPTIMIZE_PRINTF
+// Byte 107 of esp_init_data_default, only one of these 3 can be picked
+#define ESP_INIT_DATA_ENABLE_READVDD33
+//#define ESP_INIT_DATA_ENABLE_READADC
```
JarrodMP: I would like to build the firmware with this option , what or how is the best root to take .

fixingthingsguy: My experience with trying to build was extremely memorable, with an unsuccessful outcome. I spent a lot of time  , and that's because I'm no kind of expert. I finally gave up, there were too many gotcha's with setting up Docker, VM, gitHub etc. Lots of hidden stuff an expert would catch, but a novice doesn't even know to look for. Its best to avoid that route.In my opinion, its far more preferable to stay with a frightanic build and then use the esp tool to change the so-called byte 107 to the way you want it.Based on expert suggestions I wrote down what I did in #1007. It enables me to stay consistent with the SDK and Dev/Master in one fell swoop as you don't have to load the SDK 1.4  separately. If you already have the right SDK, the notes in #911 walk you through the byte 107 change.

marcelstoer: > I would like to build the firmware...what or how is the best root to take?

Not to ask such questions in a 2-months old commit comment would be a good start. Then you can familiarize yourself with http://nodemcu.readthedocs.org/en/dev/en/build/.

JarrodMP: Hi thanks very much ,
I see i can use the Flash download tool and change the settings which 
effectively changes this setting.
from

#define ESP_INIT_DATA_ENABLE_READVDD33
//#define ESP_INIT_DATA_ENABLE_READADC

to

//#define ESP_INIT_DATA_ENABLE_READVDD33
#define ESP_INIT_DATA_ENABLE_READADC

I went the route yesterday afternoon of setting up Oracle VM ware and 
installing Linux increasing drive space and getting the firmware to 
compile ,
Changing a standard firmware source download to be the setting i wanted 
above in the user_config.h file and i compiled it and put the firmware 
on one of my units ,Surprisingly it worked ,

I would like to thank you although  I  haven't tried the change the bit 
at position 107 yet I will try it, just so I have 2 options ,
It seems that you can use a GUI and flag and burn it which seems much 
simpler than compiling the firmware from source.
I looked at the manual(3A_ESP8266_Flash_Tool_user_Manual_EN_v1.0.pdf) 
and on point 2.2 and it seems simple enough to change a flag , as well 
as other interesting settings i have not read about yet.

We should keep in touch.
regard JarrodMP

JarrodMP: Hi
Thanks another helpful user pointed me in the right direction.

I loaded Oracle VM Virtual box Ubuntu all the Compilers and git hubbed 
the source,
Changed the setting in config.h and compiled ,FaBBed it and its all 
Working --   ( FaBBed ) Flag and Basic Burn)

Another user also pointed out the route of using the the EMU Flash tool 
, And that was very helpfully

I was asking from the point of the experienced users which is option 
they find better to use and has the least problems .

Thank you gave me all the options , I was hoping for an opinion on which 
was for you the better option.
It did help.

Thanks I should pay attention to what I am doing more often.
Missed that it was 2 months old .

I should read the manual more often as I was merely FaBBing and burning 
Lua_ing my ESP when i came across the 65535 read glitch when I had just 
read the VCC of my unit and needed an external Current read before I 
sent the data to my rest service and before sending the unit to sleep   .

So I blindly asked a question,

*I am mere Padawan*

Respectfully I bid you adieu

Regards JarrodMP

Issue #223, opened 2015-02-14
udp and getpeer


hreintke:
Used the dev branch to compile a version with getpeer() option.
Works OK for tcp connections but get error  
"PANIC: unprotected error in call to Lua API (servertest1.lua:2: attempt to call method 'getpeer' (a nil value))"
when using it on an UDP server connection.

hreintke:
Have been investigating further on. 
This is the lua file I am using  : 
function pr(c,pl)
  print ("parameter info c")
  print(type(c), " ",getmetatable(c));
  print ("metadatainfo c");  
  mt = getmetatable(c);
  for n,s in pairs(mt) do print(n," ",s)  end
end

if srv ~= nil then srv:close() end
if udpsrv ~= nil then udpsrv:close() end

counter=1

srv=net.createServer(net.TCP)


srv:listen(80,function(conn) 
                conn:on("receive",pr)   
                conn:on("sent",function(conn) print("tcp package sent, Heap = "..node.heap()) conn:close() end)
               end)

udpsrv=net.createServer(net.UDP)
udpsrv:on("receive",pr)
udpsrv:on("sent", function(conn) print("udp packet sent \n") end)
udpsrv:listen(8000)

which results for UDP in 
parameter info c
userdata	 	romtable: 0x4023d100
metadatainfo c
listen	 	lightfunction: 0x4025d98c
close	 	lightfunction: 0x4025d03c
on	 	lightfunction: 0x4025ce04
send	 	lightfunction: 0x4025cb04
__gc	 	lightfunction: 0x4025c52c
__index	 	romtable: 0x4023d100

And for TCP in : 
parameter info c
userdata	 	romtable: 0x4023cff8
metadatainfo c
connect	 	lightfunction: 0x4025d974
close	 	lightfunction: 0x4025d024
on	 	lightfunction: 0x4025cdec
send	 	lightfunction: 0x4025caec
hold	 	lightfunction: 0x4025da10
unhold	 	lightfunction: 0x4025d9a4
dns	 	lightfunction: 0x4025dbf4
getpeer	 	lightfunction: 0x4025c54c
__gc	 	lightfunction: 0x4025c514
__index	 	romtable: 0x4023cff8

Still trying to find why there is this difference. 
Herman

hreintke:
LS,
I have not enough experience with Pull Request to get that done but the updates in this comment lead to a working getpeer() for servers (UDP).
1: add getpeer to the server metatable
```lua
[...]

PR #750, opened 2015-11-10
Fix sending UDP response using new SDK 1.4.0

lukaszo:
From release notes of new SDK 
`Modify the espconn pointer in the receive callback of UDP. Parameters remote_ip and remote_port in it are the remote IP and port set by espconn_create. If users want to obtain IP and ports of the current sender, please call espconn_get_connection_info to get relevant information. `

This it copy paste from net module: https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/net.c#L1080-L1088


jmattsson:
I believe we might need the same fix over in enduser_setup.c too then. @robertfoss do you want to weigh in?

TerryE:
Also in net udp server probably. and see #223

robertfoss:
jmattson: Having a look.

PR #751, opened 2015-11-11
Allow to set content type in COAP (fix #658)

### LINKED ISSUE #658: COAP unusable
The primary problem is that there is no documentation but there also seem to be a number of bugs. I looked through the code but I am not expert enough in Lua/C interfaces to debug the code myself. As this is a very significant protocol for IOT for small devices like this, I am stalled until it is fixed. 

The COAP server will only respond with numbers. It may have something to do with a mixup between "payload" and "content", but I didn't have time to analyze the code in detail. Without the ability to respond with text, especially JSON, it really cannot be used.

It also does not seem to support the JSON data type yet although I could use a text data type for now.

The client code doesn't seem to handle responses (but this may just be a doc issue).

There may be more but that is as far as I was able to get. I am happy to help get this fixed but it needs an expert on the current code.

Was the code written from scratch or adapted from another COAP implementation?

lukaszo:
To test this change on dev branch this PR should also be merged https://github.com/nodemcu/nodemcu-firmware/pull/750

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,6 +545,9 @@ cs:listen(5683)
 myvar=1
 cs:var("myvar") -- get coap://192.168.18.103:5683/v1/v/myvar will return the value of myvar: 1
 
+all='[1,2,3]'
+cs:var("all", coap.JSON) -- sets content type to json
+
 -- function should tack one string, return one string.
 function myfun(payload)
   print("myfun called")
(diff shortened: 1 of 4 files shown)


jmattsson:
This looks like a good addition to me. I'm happy to merge. Any other opinions? :)

I doubt anyone would be ~~foolish~~ brave enough to attempt proper EXI encoding though.

PR #753, opened 2015-11-11
Change esptool.py default mode to DIO.

vowstar:
Change esptool.py default mode to DIO to support more board by default.


jmattsson:
This raises the old question of whether we really should be maintaining our own copy of esptool (probably not), but in the interim this is a sensible change.

PR #764, opened 2015-11-15
Integrate an extra build step for PRs to build a special firmware

### LINKED ISSUE #756: Run an all-modules build for PRs automatically
#747 showed once again that we need to run an all-modules build for PRs automatically.

marcelstoer:
If the build is triggered for a PR:
- run a script at the very end that
 - enables all modules in `user_modules.h`
 - disables SSL in `user_config.h`
 - runs `make` again with the new parameters

I opted to directly change the configuration values in the `.h` files rather than trying to pass parameters to `make`. It's easier to debug IMO because I can simply dump the content of the files _after_ manipulation to the "console" and analyze it if necessary. Also, it allows for more complex manipulations should the need for that ever arise.
Any additional PR sanity checks can be added to `pr-build.sh` in the future.

Fixes #756

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ script:
 - cd bin/
 - file_name_integer="nodemcu_integer_${TRAVIS_TAG}.bin"
 - srec_cat -output ${file_name_integer} -binary 0x00000.bin -binary -fill 0xff 0x00000 0x10000 0x10000.bin -binary -offset 0x10000
+# http://docs.travis-ci.com/user/environment-variables/#Convenience-Variables
+- if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash "$TRAVIS_BUILD_DIR"/tools/pr-build.sh; fi
 deploy:
   provider: releases
   api_key:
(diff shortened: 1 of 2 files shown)


jmattsson:
I like! Happy to merge unless someone objects (or merges!) first :)

marcelstoer:
I modified the modules regex slightly as suggested by Terry in https://github.com/marcelstoer/nodemcu-firmware/commit/0ce965d22adffe4161525ed7424a4eb4f04d32cf#commitcomment-14410614.

> My only suggestion is that you use `#define *LUA_USE_MODULES_` just in case there are multiple spaces after the define.

However, since I squashed the commits into one and force-pushed it that comment is no longer accessible from here.

jmattsson:
Minor comment - the `|| exit`s shouldn't be necessary since you use `set -e`.

marcelstoer:
Oh, riiiight...I added the `|| exit`s because the awesome http://www.shellcheck.net/ complained about their absence. However, I only checked the commands inside the sub shell `()`. There's an interesting comment at https://github.com/koalaman/shellcheck/wiki/SC2164

> Having a set -e command anywhere in the script will disable this message, even though it won't necessarily prevent the issue.

-> I'd prefer to keep them.

jmattsson:
Done!

PR #767, opened 2015-11-16
Support DS18S20 and fix negative temp handling

### LINKED ISSUE #610: ds18b20.lua | Handling difference between DS18S20 and DS18B20
Looks like DS18B20 only works proper with ds18b20 (having FAMILY CODE == 0x28).
DS18**B**20  (FAMILY CODE == 0x28) has a 12bit resolution w/ 2^(-4) lsb = 0,0625 which is correclty applied.

*derived from http://datasheets.maximintegrated.com/en/ds/DS18B20.pdf*

DS18**S**20 (FAMILY CODE == 0x10) has a 9bit resolution w/ 2^(-1) lsb = 0,5 which is not handled correctly.

*derived from http://datasheets.maximintegrated.com/en/ds/DS18S20.pdf*

This factor should be handled according to the FAMILY CODE.

nickandrew:
[2015-11-16: Fixes cleaned up and rebased against dev branch.]

The DS18S20 has only 1 fractional bit whereas DS18B20 has 4, and their
temperature register alignment differs. Check the family code to choose
the correct multiplier for both devices.

Closes #610

Negative temperatures (less than 0°C) are returned as a sign-extended
two's complement number. Subtract 0x10000 to recover the proper
negative value.

### DIFF
--- a/lua_examples/onewire-ds18b20.lua
+++ b/lua_examples/onewire-ds18b20.lua
@@ -1,5 +1,5 @@
 --'
--- 18b20 one wire example for NODEMCU
+-- ds18b20 one wire example for NODEMCU (Integer firmware only)
 -- NODEMCU TEAM
 -- LICENCE: http://opensource.org/licenses/MIT
 -- Vowstar <vowstar@nodemcu.com>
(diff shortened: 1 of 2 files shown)


jmattsson:
I'm not in a position to verify this, but it *looks* good :)

nickandrew:
Hi @jmattsson, I discussed earlier versions with @TerryE about 6 weeks ago, before the move to 1.4.0 SDK.

TerryE:
+1 on Nick's comment

PR #769, opened 2015-11-16
Rework the README file considerably

nickandrew:
Nobody had updated the README file from 0.9.5 to 1.4.0 and so that was my starting point.

Looking at the README from the point of a new user, information you need to know wasn't there or was presented in the wrong order.

- Restructured to remove a lot of irrelevant information (e.g changelog)
- Reordered to show general features first, then show off the platform capabilities, then explain how to build it, how to configure it, how to flash it and then how to run it after flashing
- Added info about esptool (the lack of doco confused me initially as a Linux user, as I couldn't use nodemcu-flasher and there was no note in the README to say it was for Windows only)
- Updated info about Marcel's cloud build server (you *do* need to tell people to build the dev branch or they will build from master)
- Added a link to the docker containerised build, which is the easiest way to build a firmware with your own patches
- Added sufficient detailed build instructions for anybody who really wants to install the toolchain
- Updated the list of optional modules in user_modules.h
- Added instructions on tagging your own builds
- Added instructions on changing the boot time serial interface rate
- Described the GPIO0 conditions for normal boot and for flash upgrade
- Gave more detailed instructions on what needs to be flashed
- And how to connect the device
- And the least interesting code snippets went down the bottom of the document where they can be cleaned up later.

After merging this pull request into dev I'd suggest also merging it into master, or probably nobody will see it until dev itself becomes master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,44 +34,6 @@ Tencent QQ group: 309957875<br />
 - add coap module (done)
 - cross compiler (done)
 
-# Change log
-2015-06-27<br />
-fixed ap/station-ap cannot connect to the device.<br />
-added wifi.ap.getconfig().<br />
-fixed net.dns.getdnsserver().<br />
-added new base64 lua example.<br />
-added node.bootreason() to inspect boot cause.<br />
-optimization of u8g.<br />
-
-# Change log
-2015-06-25<br />
-move constants to ROM. Frees up 16k+ of RAM.<br />
-add dhtlib for DHT11/21/22/33/44, port from Arduino.<br />
-add 433MHz transmission.<br />
-add crypto library.<br />
-re-add ws2812.write().<br />
-add wifi.getchannel.<br />
-changed wifi_setip() to allow setting SoftAP gateway to 0.0.0.0.<br />
-added net.dns.setdnsserver and net.dns.getdnsserver.<br />
-add support for lm92 temperature sensor.<br />
-implement getStrWidth() and setFontLineSpacingFactor().<br />
-add -Os flag to release and debug builds.<br />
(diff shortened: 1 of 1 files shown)


marcelstoer:
Thanks for the PR.

Please consider https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md next time. It would have helped to learn about your intention before you started. Among the collaborators we have been discussing what to do with the README not too long ago and I've been working on it already.

Also, please squash your commits into one (e.g. by following http://www.andrewconnell.com/blog/squash-multiple-git-commits-into-one) as requested in CONTRIBUTING.md.

nickandrew:
Hi Marcel,

I did read CONTRIBUTING.md before starting, but I noted that it said

```
This is just one way of doing things. If you're proficient in Git matters you're free to choose your own.
```

I've been working in git for about the last 8 years. I didn't squash my commits because they are logical units of change, and they're all changing the same file. Chunks of text were moved around. If I had presented a single large commit you'd have had no idea what had changed. As it is you can understand the individual changes, and when it comes to review, you might decide not to pick up all the changes I made. Squashing my commits prevents that happening.

One thing I forgot to say in the PR is, if you want to see what the README looks like after all the changes, it's here: https://github.com/nickandrew/nodemcu-firmware/tree/tidy-readme

marcelstoer:
Fair enough...I don't necessarily need to agree to understand your reasoning :wink:. Thanks for explaining, I see your point.

marcelstoer:
Even though once #771 and #774 are completed the `README.md` will look quite different, this is still valuable to have.

PR #770, opened 2015-11-17
Ensure ld-provided port alias is properly aligned.

jmattsson:
This fixes the problem observed after #729 without CLIENT_SSL_ENABLE defined.

### DIFF
--- a/ld/nodemcu.ld
+++ b/ld/nodemcu.ld
@@ -115,6 +115,7 @@ SECTIONS
     *(.data)
 
     /* Hook for randomizing TCP start numbers */
+    . = ALIGN(2);
     _tcp_new_port_port = ABSOLUTE(.);
     */liblwip.a:tcp.o(.port)
     /* Verify that the LWIP source has been patched as needed, or fail with


devsaurus:
Can confirm the fix here.

jmattsson:
Sorry, @TerryE this one was a bit too high-profile for my liking to hold off on (non-SSL builds not booting). You may now tell me off for merging my own PR :)

TerryE:
Nope, no probs.  An exception just means that.  I should have been around, but 9 days ago I had a bare slab on my plot and now there's a complete weatherproof timber frame of our new house, with doors and windows fitted after a 9-day roller coaster and the longest working day was 14 hours!  That's my excuse for being a little distracted this last week!!

jmattsson:
Whoa, you're making good progress on that house!

TerryE:
The planners have stipulated a stone skin and slate roof to match in with the village, plus 1st and 2nd fit so we won't be finished until next summer.  Not that this has naythinto do with data alignment :laughing:

jmattsson:
Oh ouch... I hope that slate roof won't cause you too many troubles down the line. But you know, aligning tiles and aligning bytes, there are similarities - bad things happen in both cases if you get it wrong!

PR #775, opened 2015-11-18
Update dht_lib.lua

iWladi:
misprint


TerryE:
The way that Martin coded this seems peculiar to me, but to your specific correction, this does seem to be a cut and paste error.

marcelstoer:
Why was this merged to `master` rather than `dev`?

TerryE:
Because I am a prat :oops:

TerryE:
This is now dead. See #789

PR #776, opened 2015-11-18
Make net.dns.resolve and net.dns work with an IP address as the hostname

askainet:
The functions *net.dns* and *net.dns.resolve* call *espconn_gethostbyname* with a callback function *net_dns_found* to resolve a hostname to an IP address, but if this call returns inmediately an OK status (because the hostname is in fact already an IP address) they need to run the *net_dns_found* right away to deal with the result

Also, the function *net_dns_found* was always expecting the *host_ip.addr* to be 0, but *espconn_gethostbyname* could return inmediately and assign it the IP address


jmattsson:
This looks good to me. Unless someone has a reason not to, I'll merge this in day or so. If someone wants it in before then, go right ahead!

PR #779, opened 2015-11-19
Make sntp.sync work with DNS hostnames

askainet:
Added a call to *dns_gethostbyname* to resolve the server name to an IP address:

If server is already an IP address it returns inmediately.

If it's a DNS hostname it uses the new callback function *sntp_dns_found* that will get the IP address and run *sntp_dosend* after DNS resolution is OK.


jmattsson:
Thanks!

PR #782, opened 2015-11-20
Provide GIT_HASH as version information

eku:
Not having any particular version information when your are searching for bugs in a module that has been flashed months ago. The git hash helps to identify the version in use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,9 @@ install:
 - tar -zxvf tools/esp-open-sdk.tar.gz
 - export PATH=$PATH:$PWD/esp-open-sdk/xtensa-lx106-elf/bin
 script:
+- export GIT_HASH=$(git log --pretty=format:%h -n 1)
 - export BUILD_DATE=$(date +%Y%m%d)
-- make EXTRA_CCFLAGS="-DBUILD_DATE='\"'$BUILD_DATE'\"'" all
+- make EXTRA_CCFLAGS="-DBUILD_DATE='\"'$BUILD_DATE'\"' -DGIT_HASH='\"'$GIT_HASH'\"'" all
 - cd bin/
 - file_name_float="nodemcu_float_${TRAVIS_TAG}.bin"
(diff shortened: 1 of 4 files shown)


TerryE:
A date is a date and not a git hash.  It might be a good idea to add a `BUILD_HASH` parameter, but we should stick with a date for a date field.  Why not 

```bash
BUILD_DATE=$(shell git log --pretty=format:%hci-n 1)
```
You can find the hash from the commit date anyway.  You also need to fix the travis issue.

marcelstoer:
> A date is a date and not a Git hash.

+1

> It might be a good idea to add a BUILD_HASH parameter

If so then it should IMO be called `COMMIT_HASH`.

> but we should stick with a date for a date field.

+1. As long as it's called `BUILD_DATE` it should be exactly that - a _build_ date.

eku:
Anyway, who needs a hard coded build date?

BTW, the link to the failed travis build gives me white page. No content at all.

TerryE:
> As long as it's called `BUILD_DATE` it should be exactly that - a _build_ date.

The slight disadvantage of using a literal build date is in the context of the cloud build service.  You could have two otherwise identical binaries with different build dates, which is why there is an argument for using the last commit date.

marcelstoer:
> there is an argument for using the last commit date

There definitely is...if the property is named accordingly.

eku:
You want me to rename GIT_HASH to what?

TerryE:
@eku Eric, it's not `GIT_HASH` that we're talking about but `BUILD_DATE`.  Marcel thinks that this is misleading if we use the last commit date.  His point is that the build date is the date that the make was run.  Maybe `GIT_LAST_COMMIT_DATE` says what this really is.

eku:
@TerryE I understand. But this is probably another PR.

jmattsson:
Even a commit hash does not clearly define the firmware, considering the plethora of user_config that is available.

TerryE:
@eku Eric,  we think that this one isn't quite the way to go, and I don't want to leave the PR open in this case.  If you want to raise an issue to discuss options referencing this and proposing a way ahead, then please do so.

PR #787, opened 2015-11-23
Merge Onewire changes from Master into Dev

TerryE:
Master contained two OW changes one of which was already reimplemented in dev using the
correct macros.  The second was from hazarkarabay (f6d0c0c) that onewire_search()
returns a device address vector even when no device is found.  I've reimplemented this
but using the correct index, rom_byte_number, rather than declaring a new temp index i.

### DIFF
--- a/app/driver/onewire.c
+++ b/app/driver/onewire.c
@@ -407,10 +407,15 @@ uint8_t onewire_search(uint8_t pin, uint8_t *newAddr)
       LastFamilyDiscrepancy[pin] = 0;
       search_result = FALSE;
    }
-   int i;
-   for (i = 0; i < 8; i++) newAddr[i] = ROM_NO[pin][i];
+   else
+   {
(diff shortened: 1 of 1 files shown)


TerryE:
@nickandrew, can you please review.  Thanks. Terry

nickandrew:
I'll test it. I didn't actually notice the search function was missing that fix when I put 'dev' on a device recently. So I'll do some tests with the current version of 'dev' and with this pull request.

nickandrew:
Ok, I tested everything using old and new firmware using this script: https://gist.github.com/nickandrew/54d2bfaaefeba65b1e7a

The search function wasn't actually broken before the fix; it was merely redundantly copying 8 bytes and the lua interface subsequently ignored it.

Testing worked with 0, 1, 2 and 3 devices on the bus. It's good to merge.

PR #793, opened 2015-11-28
Add uart.alt function to swap to using RTS and CTS as RX and TX (pins…

btsimonh:
I wanted to use the alternate serial pins (GPIO13 & GPIO15) on the ESP03, but the code did not seem to be there.
I've added some defines for the IOSWAP register in uart_register.h, and then used them in a new function uart0_alt(), plus implemented calling through to the function from Lua.
With these changes, the 'normal' serial can remain in place for flash, diagnostics, etc. while serial peripherals don't get hassled with bootup output.
I've tested that the serial no longer comes out of the normal pins, but not wired to the new pins yet!

ignore the config changes..... did not know how to exclude them from the pull request

### DIFF
--- a/app/driver/uart.c
+++ b/app/driver/uart.c
@@ -66,6 +66,39 @@ uart_config(uint8 uart_no)
     SET_PERI_REG_MASK(UART_INT_ENA(uart_no), UART_RXFIFO_FULL_INT_ENA);
 }
 
+
+
+/******************************************************************************
+ * FunctionName : uart0_alt
+ * Description  : Internal used function
+ *                UART0 pins changed to 13,15 if 'on' is set, else set to normal pins
+ * Parameters   : on - 1 = use alternate pins, 0 = use normal pins
+ * Returns      : NONE
+*******************************************************************************/
+void ICACHE_FLASH_ATTR
+uart0_alt(uint8 on)
+{
+    if (on)
+    {
+        PIN_PULLUP_DIS(PERIPHS_IO_MUX_MTDO_U);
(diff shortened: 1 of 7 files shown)


marcelstoer:
I don't understand the C code but the changes to the three `app/include/user_*.h` files must definitely be reverted. Then [squash and force-push](http://www.andrewconnell.com/blog/squash-multiple-git-commits-into-one) your commits.

btsimonh:
Really confused now....
rebased and did some other stuff i did not understand, and now down to 7 files changed.  6 of which i changed :(  one of which is unchanged...  so i assume it change in the master branch...  anyway, it's now a clean pull request without the user configs, etc.....

marcelstoer:
Oh, looks like there's one essential part from https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md that you missed: PRs should be created against `dev` rather than `master`!
Most likely you didn't even get to see this file because it hasn't been propagated to `master` yet :disappointed: .

btsimonh:
yeah; really having fun here :(.
That file which 'modified' itself; well it was wrong anyway!!!!  so it had changed from *wrong* to *worse*, and now i've made it *better*, but since i've never used or looked at that bit I can't know if it is *correct*.  Let's hope someone cares.
I wish i'd read the CONTRIBUTING.md file, it would have saved a lot of trouble.  maybe I start again....

btsimonh:
Please ignore this one for the moment.  I'm trying to clarify exactly what the IOSWAP register does....

btsimonh:
Does the new pull request look better?

PR #794, opened 2015-11-28
Added uart.alt(n) to swap to use alternate rx/tx lines.

btsimonh:
if n = 0 -> std pinout.
if n != 0 -> alternate pinout (gpio13 and gpio15)

uses IOSWAP register at 0x3ff00028 - not documented?

### DIFF
--- a/app/driver/uart.c
+++ b/app/driver/uart.c
@@ -21,6 +21,10 @@
 #ifndef FUNC_U0RXD
 #define FUNC_U0RXD 0
 #endif
+#ifndef FUNC_U0CTS
+#define FUNC_U0CTS                      4
+#endif
+
 
 // For event signalling
 static uint8 task = USER_TASK_PRIO_MAX;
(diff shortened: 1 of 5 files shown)


nickandrew:
+1

jmattsson:
Seems like very useful functionality! One question though - what state are the previous RX/TX pins left in after a swap? Can we get that documented and/or clarified somehow?

jmattsson:
Also, how does this compare to using the system_uart_swap() function?

btsimonh:
reply on a question to expressif highlighted this post:
http://bbs.espressif.com/viewtopic.php?f=7&t=22
for comparison.
Ref state of normal RX/TX pins, they MAY become RTS/CTS (as it indicates it's a 'swap'). One assumes that using GPIO functions, they could then be reassigned?
p.s. fully tested working last night, as far as sending and receiving from the new pins.

btsimonh:
ref system_uart_swap() and system_uart_deswap() - no got the SDK to hand right now, but http://neilkolban.com/tech/wp-content/uploads/2015/09/Kolbans-Book-on-the-ESP8266-October-2015.pdf seems to indicate this would swap the UART in use not the pins....?

PR #796, opened 2015-11-29
Fix compilation errors when LUA_OPTIMIZE_MEMORY == 0

Alkorin:
nodemcu doesn't compile when LUA_OPTRAM is not defined.

This patch fix compilation errors


Alkorin:
Yes,

My first goal was to implement/test the ```LUA_OPTIMIZE_MEMORY == 0``` part of my ws2812 library patchs https://github.com/Alkorin/nodemcu-firmware/commit/0c7be8712fbbc3c03d1a3584477037d2433d6594

So, I've undef ```LUA_OPTRAM``` in app/include/user_config.h and saw all these compilation errors. I've fixed them, built a firmware, uploaded it on my esp8266 and tested it: it worked with ~18kB of free heap instead of ~42kB and my ws2812 was working.

My patch should be a no-op for LUA_OPTIMIZE_MEMORY or I missed something ?

TerryE:
Read the eLua LTR whitepaper which expalins how LTR works.  In essence as many libraries and library constants are moved into ROMable address space.  This is essential when working with a processor that only has 48Kb of effective RAM.  Standard Lua uses dynamic (that is runtime) callbacks to initialise these, but to do this means that you need to be able to write to the address space -- which is a no-no for ROM.  Hence `user_config.c` and `user_modules.c` which convert these call backs into static declarations.  What your patch is doing it to try to back these out.  The result will never work on an ESP8266.  So look at the other modules and see why they work and mirror this. 

Don't propose a patch unless you have working code / firmware image that you have thoroughly tested. :smile:

Alkorin:
I've a firmware image which works (at least boot) with ```LUA_OPTIMIZE_MEMORY == 0```

```
> print(node.heap())
19304
> print(node.CPU160MHZ) 
160
```

So the patch with ```MOD_REG_NUMBER(L, "CPU160MHZ", CPU160MHZ);``` works.

If I build it with LUA_OPTIMIZE_MEMORY == 2, it works too
```
> print(node.heap())
42744
> print(node.CPU160MHZ) 
160
```

So, except if I missed something my MOD_REG_NUMBER patches are good and are no-op if ```LUA_OPTIMIZE_MEMORY == 2``` no ?
I've just moved LNUMVAL definitions in ```#if LUA_OPTIMIZE_MEMORY``` sections, so it does nothing if ```LUA_OPTIMIZE_MEMORY``` is set, and I've defined them in the ```#else``` section, so it again does nothing if ```LUA_OPTIMIZE_MEMORY``` is set.

Other are just typo/missing definitions fixes...
I know I'm new and I've to learn, but ... which part is wrong ?

jmattsson:
This discussing is raising an interesting point - should we even support something other than LUA_OPTIMIZE_MEMORY == 2?

TerryE:
>This discussing is raising an interesting point - should we even support something other than LUA_OPTIMIZE_MEMORY == 2?

 IMO, _only_ reason that we retain `LUA_OPTIMIZE_MEMORY` is for alignment to the eLua repo, and to allow non-LTR running for core testing. As far as module writers are concerned, we should mandate LTR, and not support non -LTR PRs

jmattsson:
@TerryE Would you like me to do a clean-up sweep of all the modules and centralize the -DLUA_OPTIMIZE_MEMORY=2 to have a consistent approach in the tree? I could put that on my to-do list for the holiday break.

TerryE:
Thanks. Good idea.

TerryE:
Superseded by #810

PR #797, opened 2015-11-29
Fix crash in implementation of concat if a GC happens at a bad moment

pjsg:
This fixes a panic if (I think) a GC happens during a tostring() in the handling of concat. The original assert firing was:

```
lvm.c:332: (((top-1)->tt) == 6)
```

This problem goes away (for me) if I reset 'top' after each call to tostring().

### DIFF
--- a/app/lua/lvm.c
+++ b/app/lua/lvm.c
@@ -327,31 +327,36 @@ void luaV_concat (lua_State *L, int total, int last) {
         top = L->base + last + 1;
         luaG_concaterror(L, top-2, top-1);
       }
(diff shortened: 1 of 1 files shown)


TerryE:
Philip, see my #813, which is an alternative fix staying within the EGC paradigm.  If you are happy with my fix then I will close this PR.

TerryE:
Since Johny merged my alternative patch, I suggest we close this, but I just want to thank you Patrick.  Without your work, we probably wouldn't have found this.

pjsg:
Good. I'll build with your committed version and check it out....

Issue #801, opened 2015-11-30
PANIC messages should not cause hardware fatal exceptions


pjsg:
The following test case triggers assertion failures in lvm.c reliably:

```
function badfont()
   local x = u8g.ssd1306_128x64_i2c(0x40)
   
   x:setFont(0)
end

tmr.alarm(0, 100, 0, badfont)
```

with this error:
```
PANIC: unprotected error in call to Lua API (stdin:3: bad argument #1 to 'setFont' (font data expected))
lvm.c:470: base <= L->top && L->top <= L->stack + L->stacksize
Fatal exception 29(StoreProhibitedCause):
epc1=0x4024adc5, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000004, depc=0x00000000
```

Now, this should generate a PANIC as I did pass in a bad argument to setFont, but it isn't actually clear what should happen next. However, I would argue that an assertion error followed by a hardware trap is incorrect.

I'll try and fix this (once I understand what _ought_ to happen as a result of a PANIC)

pjsg:
It looks as though @nickandrew is going to add the enablement of asserts.... I'll probably use another ticket to fix the PANIC issues.

jmattsson:
Looking at that `29(StoreProhibitedCause)` and `excvaddr=0x00000004` I'd say something in the assert code tried to write through a null-pointer.

nickandrew:
> it isn't actually clear what should happen next.

'Programming in Lua' section 8.4 (which I just read for the first time) indicates that a panic is appropriate, because the code is not being executed under pcall(). What I'd *like* to see here, as well as the panic, is a stack traceback.

I'm guessing the firmware is soldiering on regardless after the panic, and that's why the null pointer exception happens.

pjsg:
The panic is appropriate and the lua manual indicates that the app should exit. I don't know what that means in an embedded environment. It could cleanly restart the firmware (reboot), or it could just abort that particular call stack -- essentially leaving the app running.

I think that the requirements for embedded systems are different to that of a standalone program. In a system which mixes C and has a bit of lua where there is significant work done by the C code (e.g. a firewall that supports LUA for customization), it would be very unfriendly to reboot the firewall because somebody has a typo in their lua code. 

I also agree that a stack trace would be nice as well.

pjsg:
[...]

PR #804, opened 2015-12-01
Do not allow execution to continue after a PANIC

jmattsson:
As discussed in #801, I thought I'd already pushed this.

### DIFF
--- a/app/lua/lauxlib.c
+++ b/app/lua/lauxlib.c
@@ -816,6 +816,7 @@ static int panic (lua_State *L) {
   luai_writestringerror("PANIC: unprotected error in call to Lua API (%s)\n",
                    lua_tostring(L, -1));
 #endif
+  while (1) {}
   return 0;
 }
 


TerryE:
Johnny, will this do what you intend? Surely a `while(1) {}` loop will simply trigger the WDT? I am thinking  aloud on my tablet rather than my laptop, but  IMO the best approach would be to dump some valedictory diagnostic into the RTC and latch a "PANIC" restart reason.

jmattsson:
Aye Terry, it [does precisely that](https://github.com/nodemcu/nodemcu-firmware/issues/801#issuecomment-160844777). This isn't a fix for @pjsg's *actual* issue, this is only to prevent the subsequent flood of Weird Stuff(tm) that otherwise happens after you've hit a panic and can lead you on a wild goose chase. Continuing Lua execution after a panic is (very very) wrong, and that''s what this one-liner ensures doesn't happen.

I'm not opposed to extending it with further diagnostics, though having a #define to switch it off would be preferrable (we use every word of available RTC memory for actual data storage on some of our projects,
so we'd rather sacrifice the diagnostic capability). Knowing what information to squirrel away to be useful is more your area of expertise than mine, so I'll let you do that bit ;)

pjsg:
Could this while(1) loop be placed in place of the c_exit that was commented out. If someone (in the future) uses the lua_atpanic call to replace the panic handler, then this problem will happen again.

TerryE:
@jmattsson Johny, one of the changes that I added along with my strip debug was to re-implement proper tracebacks.  See `lua.c`, reporter `docall()` calls `traceback()` which now calls `debug.traceback()` as this is now implemented.  It would be a good enhancement if the panic routine did a proper traceback before going into the shutdown death loop.

Possibly the simplest way would be to move the static traceback() to `lauxlib.c:luaL_traceback()` and add the `docall()` logic inline in `panic()`.

jmattsson:
Both good ideas, I'll put both on my to-do list!

nickandrew:
I modified it to optionally halt on panic rather than reboot. Just for use while debugging.

```diff
diff --git a/app/include/user_config.h b/app/include/user_config.h
index b4e7e5b..53b119e 100644
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -26,6 +26,8 @@
 #define BIT_RATE_DEFAULT BIT_RATE_9600
 #endif /* DEVELOP_VERSION */
 
+/* Define this to halt forever on panic, rather than automatic restart */
+#define PANIC_HALT
 
 #define NODE_ERROR
 
diff --git a/app/lua/lauxlib.c b/app/lua/lauxlib.c
index 1ce21f3..75a2fdd 100644
--- a/app/lua/lauxlib.c
+++ b/app/lua/lauxlib.c
@@ -816,7 +816,11 @@ static int panic (lua_State *L) {
   luai_writestringerror("PANIC: unprotected error in call to Lua API (%s)\n",
                    lua_tostring(L, -1));
 #endif
+#if defined PANIC_HALT
+  while (1) { system_soft_wdt_feed(); }
+#else
   while (1) {}
+#endif
   return 0;
 }
 
```

Could be useful to add?

TerryE:
Nick, I am not sure what this would add other than avoiding the boot message, as the ESP to all intents and purposes would be locked up and need a H/W reboot.  

But as to implementing this type of option, there is now a minimal debug compiled in by default.  This provides `traceback` and `getregistry`.   It would be simple add, say, `degug.setPanicAction()` to make this configurable at runtime.  You might want to wrap this code in a `DEBUG_PANIC_STALL` conditional.

nickandrew:
> Nick, I am not sure what this would add other than avoiding the boot message, as the ESP to all intents and purposes would be locked up and need a H/W reboot.

That's right, it's strictly for debugging use. I'm seeing the console messages on ESPlorer and when it panics I can scroll back through the full console output at my leisure. ESPlorer doesn't allow scroll back while the device is still outputting text, so it's difficult to review the logs if the device is continually rebooting.

I'll check out the minimal debug. Not sure how I can use it (I suppose it's undocumented) but I'd need to call it from C, because it's the C functions in which I am detecting the error. Changing the lua code tends to make the panic disappear.

TerryE:
See my last comment on #813.  My patch shows you how to do this.

TerryE:
@nickandrew Nick, I am going to merge this one as it is a lot better than the current continue and then die messily and confusingly as is the current behaviour.  This is the best option for typical Lua developers.  I am currently developing a more sophisticated firmware and module developer variant which gives better info, but will implement this as a subsequent PR.

nickandrew:
@TerryE No worries! Looking forward to it. I probably did something wrong following your comment in #813 cause I've been getting reboots rather than tracebacks. I'll have another go at it. And I'll keep feeding the WDT in my local branch to halt rather than reboot, as it's so useful to me to have the boot-run-crash cycle be a one shot.

PR #812, opened 2015-12-03
Fix ECG case in concat where stack top not recalculated

TerryE:
The standard Lua GC can't be run from an arbitrary point in the VM.  The ECG is a patch which allows GC to be run after before or any arbitrary allocation.  Part of this collection can shrink the stack,  thus moving its location in memory.  Any functions which take local pointers to stack elements must either set a flag to fix the stack or recompute the relevant local pointers after any allocation.  The patch missed this one case.


jmattsson:
Sounds like this is something we should also submit back to the eLua project?

PR #815, opened 2015-12-05
Remove the lobject.c:88 assertion failures

nickandrew:
When lua assertions are enabled, normal operation results in many:

```
lobject.c:88: (((t1)->tt) == 4)
lobject.c:88: (((t2)->tt) == 4)
lobject.c:88: (((t1)->tt) == 4)
lobject.c:88: (((t2)->tt) == 4)
lobject.c:88: (((t1)->tt) == 4)
lobject.c:88: (((t2)->tt) == 4)
```

It comes from using the pvalue() macro for 3 pointer types, where pvalue() also checks the type of pointer and complains through the assertion where the type == 4 (TLIGHTUSERDATA).

Use the correct macro according to the type of data being compared to eliminate this assertion error.


TerryE:
+1 I found this as well, but thanks for the patch Nick

PR #841, opened 2015-12-12
Cleanup/minor correction: Flash addr cleanup

jmattsson:
The platform flash API has been somewhat confused as to what offsets it should be using, and making some unsafe assumptions around the mapped portion of the flash. The first commit fixes the offset part and should make the API considerable more portable (esp32, perhaps?) and less confusing to use.

After some digging around in the flash cache control register I also managed to find a way to resolve the unsafe assumptions part too. The flash API now correctly deduces the mapped portion and adjusts accordingly.

### DIFF
--- a/app/platform/common.c
+++ b/app/platform/common.c
@@ -149,7 +149,7 @@ uint32_t platform_flash_get_first_free_block_address( uint32_t *psect )
   uint32_t start, end, sect;
   NODE_DBG("_flash_used_end:%08x\n", (uint32_t)_flash_used_end);
   if(_flash_used_end>0){ // find the used sector
-    sect = flashh_find_sector( ( uint32_t )_flash_used_end - INTERNAL_FLASH_MAPPED_ADDRESS - 1, NULL, &end );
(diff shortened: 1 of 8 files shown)


jmattsson:
@nickandrew I agree that looks a bit off, but I have yet to understand the reasoning behind the 4-sector stuff in the first place. I don't see a reason why changing 0x3000 to 0x3FFF wouldn't be a good thing though.

nickandrew:
I'm assuming that an unlucky firmware build could end up with a size which causes the start of the spiffs to overlap the end of the firmware by up to 4K; this should be easy to test by adding the right amount of static data and showing it to be overwritten by spiffs.

jmattsson:
Yep, I've just added that fix to this PR. Thanks for pointing it out!

TerryE:
Yup, I remember reading in the spiffs documentation that is supposed to start on a 4x4k page boundary above the end of the flash ROM. If we are making then contiguous, excepting this 4k alignment then the extra 0xFFF makes sense. 

+1 to the other changes :)

Issue #810, opened 2015-12-02
Clean up modules for consistent LTR approach


jmattsson:
Note to self: As discussed in #796, I need to do a sweep of all the modules and tidy up the LTR aspects to get some consistency, and make it easier for new devs to Do The Right Thing.

TerryE:
@jmattsson Johny, this is a big patch. I need to review it properly, and not when I am having a Saturday morning cuppa in bed and working on my tablet! :laughing: 

But, I've replied on the issue, because I wanted to air the principle for wider comment amongst the other active contributors. I am sure that most will trust that we can age the details between us, following those agreed principles.

I suggest that these principles are:

*     We are mandating LTR at optimisation level 2 for all standard modules within our distribution. What individual developers do for their own add-on libraries I'd up to them.
*     We are including your non-aligned flash memory patch, and making us of it by moving all default .text segments into flash.
*     We are including your refactoring of the common string and allocation macros to remove unnecessarily duplicated routines.
*     We are including my LCD (packed info) patch at a default stripdebug level 2. That is source compiles dump local and upvalue debug info, and retain packed line debug information. As this drops the overhead of running from source compared to **lc** files to less than 10%.

Since your are tidying up the source, I would like to propose another principle, and that is that we kept the differences between the eLua code base and nodeMCU to a necessary minimum.  An example here is that nodeMCU has an almost global replace of strlen to c_strlen, but to avoid pulling in two different strlen functions we now have strlen as a linker synonym for c_strlen, so why do this bulk edit and make 100s of unneeded difference?

Anyway, I would appreciate your and others thoughts.

jmattsson:
The patch is mostly large in size rather than content. A pattern will stand out very quickly once you start reading it... but a tablet might not be the best device for that!

As for the listed principles, I'm in agreement. Plus I'd love to see the addition of the EGC changes/tweaks you're working on once they're ready - the performance improvement seemed quite remarkable from what you mentioned so far.

I'm all for keeping differences minimal between NodeMCU and upstream eLua (hence my comment in the PR). When large changes can be avoided through smaller changes via the pre-processor or linker-script, that's typically where I'll lean.

TerryE:
Johny, This patch is sufficiently complex that it merits some review comments and possibly a second iteration.  

[...]

PR #842, opened 2015-12-12
Cleanup: LTR module registration

jmattsson:
This is the PR corresponding to #810.

@TerryE There are a bunch of things in app/lua that insist on setting their own MIN_OPT_LEVEL back and forth. I see no reason why it would really need to do so, but I'm not versed enough in the Lua internals to want to make that call. For now I opted for minimal changes to keep it easier to apply any upgrades/patches from the eLua project.


TerryE:
Johny, There's somethig going wrong with the make and linker magic.  I think that the issues is that the old init.o generated loader references to the rom tables that then resulted in a pull cascade bringing in the these modules This just isn't happening so most modules aren't getting loaded..  When I get to the bottom of this I'll update his comment.

jmattsson:
Argh, I can reproduce that here too now! Looks like I broke something while I was tidying up for the commit *grumble*

TerryE:
Give me a 'mo.  Have a look at this gist: [generate_user_modules.lua ](https://gist.github.com/TerryE/e52402ac2926438dfacc#file-gistfile1-txt)

TerryE:
You plonk a config file in your suer folder and have a make rule with generates the user_includes.h, for example:
```
;
; Minimal biuld for testing.
;
builtin = string,table,coroutine,math,debug_minimal

modules = node,file,gpio,wifi,net,,spi,tmr,adc,uart,ow,bit
```

jmattsson:
Doing it with outside scripts are easy, but would break the cloud build until it's fixed. The idea with just using the compiler & linker was to avoid both external scripts and being backwards compatible.

The issue we're seeing is that despite the KEEP() statement in the linker file, the linker is happily ignoring those sections in a whole bunch of files. I'm trying to understand why; I'm sure I had this working just fine yesterday.

jmattsson:
All the pieces are in the right place in the .o / .a files, it's the final linking which incorrectly chucks some things out.

TerryE:
> it's the final linking which incorrectly chucks some things out

Not true.  It's doing its job -- which only to include modules that are explicitly include and their dependency cascades.  What other difference is there between an included module and one that's not required.  The defines don't generate anything, so the linker is blind to them.  You need to generate a strong link per module to get it included.  You can't do that in the module; it has to be in a central place like init.c.

Travis can run lua etc. as part of the make so long as its configured correctly.

jmattsson:
The ld KEEP() statement is meant to mark section contents to be included in the dependency tree building, even it there are no direct references to it. It's mostly used to ensure raw ISR functions are included in output, since they're fixed to a particular address which only the hardware ever references, and the linker would otherwise be blind to, as you put it.

This is why I needed to drop included vs excluded module inits into different sections, so that I could KEEP() only the included ones.

jmattsson:
Right. So, apparently GNU ld does an extra discard pass early on for objects in a static archive. The only mention of that I've found was over at [Stack Overflow](http://stackoverflow.com/a/2610815). If you specify the plain .o files on the command-line instead of the archive, the linker does the expected thing. Wrapping the archive in a --whole-archive also makes it do the right thing. Don't be fooled by the impression --whole-archive gives, of including unnecessary things; The linker still chucks out anything unreferenced, only it does so later, after any explicit KEEP() statements in the linker script have had a chance to take effect. For the curious, the default-build firmware is actually 16 bytes smaller after the change to the NODEMCU_MODULE() approach.

I was sure I had this all working yesterday, but based on what I've learned this morning, I must've been dreaming yesterday.

TerryE:
Johny,

Just out of interest I've got my scripted version working, but the secret was adding this into `init.c`:
```C
#ifdef LUA_MODULES_EXTERNS
   LUA_MODULES_EXTERNS
static const __attribute__((used)) luaR_table  *reference_lua_libs[] = {
   LUA_MODULES_REFS
};
#endif
```
I've dropped the static from the NODEMCU_MODULE macro lua_rotable entry and `LUA_MODULES_REFS` is just a list of the addresses of these entries.  You need the (used) attribute and this triggers the include cascade.  I am not sure how to do this without the central list.  

As far as Travis CI is concerned, it supports Perl as standard doesn't it?  Why not just use Perl 5.18 for the utility.

jmattsson:
> Why not just use Perl 5.18 for the utility.

- Because there is no need for it, it can all be done with the compiler & linker
- Not everyone might have perl installed (especially on a Windows box)
- Just because CI can handle it, it doesn't necessarily make it nice & easy for other dev environments (yes, I'm thinking about the impact this will have on our build-setup at $work too here)

As for having a central macro which knows about all the table entries, that's what I wanted to get away from, because something has to generate/update that list.

TerryE:
Johny, the standard way to approach this type of build configuration issue is to use a configure / autoconf script.  These problems really arise here because we're trying to use a makefile hierarchy to do this entirely, and we're just pushing this to the limits of this capability.  We could everything with a Universal Turing Machine, but that doesn't mean that it's sensible to do it this way.  There is a wealth of information on the Internet on using configure systems, but getting to the bowels of KEEP and other loader directives and their scope, effects and side effects is extremely esoteric.

What I am trying to get away from is relying on individual module developers update some central core file, which is a different and in my view more practical objective.   As to the Windows point we are using the open-sdk toolchain, which requires a GNU/POSIX system (Linux, BSD, MacOSX, Windows with Cygwin); Perl is in practice universal on these systems which is why I suggested it.

Anyway, I have simple working interim, and await your solution with interest :-)

jmattsson:
It's pushed, no need to be kept in suspense ;)

TerryE:
Sorry, I did an edit but forgot to post hence we cross posted.  I'll have a look at your new push. :-)

jmattsson:
Oh don't get me started on the Espressif makefile hierarchy - as you say, just because it can be done doesn't mean it should be done. So much foot-shooting resulting from not sticking with the functional approach which make is built around... Yes, make is hard to grok at first, but trying to use it in an imperative manner only adds to the confusion.

I'm not sure why you think my aim isn't to get away from module developers having to update some central core file. That is precisely what this does. A module developer only needs to write the module C file, and to enable it do the appropriate #define in user_config.h (for full backwards compatibility, both developers & cloud builder).

TerryE:
> I'm not sure why you think my aim isn't to get away from module developers having to update some central core file.

I did say that -- or if I did I didn't mean it sorry.  We both agree on this one.  It's just that there are different ways to skin the same cat.  In this case, if your approach is working, then I am happy to roll with it.

TerryE:
Sorry, but it's still not working.  I tried a custom build commenting out all non-core modules, including u8g, for example.  The bin file was far too big.  If you do, say:
```
grep u8g app/mapfile | grep 0x | grep -vP '\.comment|\.debug|\.xtensa\.info|0x0\b'
```
then you will see that the image has a load of u8g components in it. Ditto for all of the other modules that were supposed to be skipped.  Essential as far as I can see, the 
```
MODULES_LIB = $(filter %modules.a, $(DEP_LIBS_eagle.app.v6))
```
hauls in not only the stuff that we wanted but was being skipped but also a whole load of stuff that we didn't want and was being skipped.  

Round 3 or back to bash?

jmattsson:
Well I'll be....

Let me ponder it for a bit? I need a break anyway from the SDK/build rejig stuff which is really not giving me the warm and fuzzies at the moment.

TerryE:
I am knackered and off to bed.  You can have all (my) night!! maybe swap notes in (my) morning :smiley:

jmattsson:
Looks like I didn't need quite all (the|your) night, Terry! After a considerate amount of pondering, pontification, and rumination, I arrived at the conclusion that unless I wanted to do something outlandishly convoluted (no thanks), I had to at least move half-way towards your approach.

The third attempt still avoids needing a common file which knows about all the modules (be it auto-generated or not), avoids introducing a dependency on another interpreter, and still keeps the modules self-contained. Instead of trying to use section names to pick out the enabled modules, I switched to using the variable name of the Lua rotable for the module. In the Makefile before the final link make just invokes nm on libmodules.a and filters these names to obtain a list of symbols. With that in hand, it's (finally) trivial to tell gcc/ld to please include those in the link. A possibly useful side-effect of this is that you can easily see in the build log which modules were enabled in that particular build - there's a `-u MODNAME_module_enabled` on the linker line for each.

I've been testing this back, forth, up, down and sideways and staring at the mapfile output. As far as I can see this now does do the right thing (though I feel like a clown for it taking so many iterations). So, over to you, @TerryE and anyone else who'd like to point out where I've stuffed this up! :)

TerryE:
OK, A simple sanity check on the following: 

*  **builtin** = string, table, coroutine, math, debug_minimal
*  **modules** = bit, file, gpio, i2c, net, node, ow, rtcfifo, rtcmem, spi, tmr, uart, wifi

With current dev:
```
Sections:
Idx Name          Size      VMA       LMA       File off  Algn
  0 .irom0.text   00046ddb  40210000  40210000  00007de0  2**4 CONTENTS, ALLOC, LOAD, CODE 
  1 .data         00000898  3ffe8000  3ffe8000  000000e0  2**4 CONTENTS, ALLOC, LOAD, DATA
  2 .rodata       00000008  3ffe8898  3ffe8898  00000978  2**2 CONTENTS, ALLOC, LOAD, DATA
  3 .bss          000070c0  3ffe88a0  3ffe88a0  00000980  2**4 ALLOC
  4 .text         0000745f  40100000  40100000  00000980  2**2 CONTENTS, ALLOC, LOAD, READONLY, CODE
```
and with the new build process:
```
Sections:
Idx Name          Size      VMA       LMA       File off  Algn
  0 .irom0.text   00062ef3  40210000  40210000  00007e80  2**4 CONTENTS, ALLOC, LOAD, CODE
  1 .data         00000904  3ffe8000  3ffe8000  000000e0  2**4 CONTENTS, ALLOC, LOAD, DATA
  2 .rodata       00000008  3ffe8904  3ffe8904  000009e4  2**2 CONTENTS, ALLOC, LOAD, DATA
  3 .bss          000075a0  3ffe8910  3ffe8910  000009f0  2**4 ALLOC
  4 .text         0000748b  40100000  40100000  000009ec  2**2 CONTENTS, ALLOC, LOAD, READONLY, CODE
``` 
Note that the **.irom0.text** segment has risen by 0x46ddb to 0x62ef3, that's about 104Kb extra.
grepping the two mapfiles for this segment hightlights that smartconfig, ssl, gpio16 and i2C_master components are being pulled in even though these aren't used with this configuration, hence the extra 100 odd Kb.

One more try? or shall we try my shell version with a 2 line file containing the above builtin and modules list to replace the user_config.h file.  Hint: just look at the size of the 0x00001.bin file for this configuration. If its not around 290K then something has gone wrong with the link.

jmattsson:
... but that's [impossible](http://plasmasturm.org/log/6debug/)! I'm really going to have to dig into this to understand what's happening here, or it's going to drive me nuts not knowing.

I'm not disputing that your script might not be quicker to get to work, but by now this is [personal](https://www.youtube.com/watch?v=k5gisePRSIE) - between me and the toolchain. It is ON! I foresee two cref's and a diff in my near future...

Oh, and I'm pretty sure you mean user_modules.h, not user_config.h, Terry ;)

jmattsson:
Moving on to stage 2 - it doesn't happen on my machine. Two builds, default config. Rev 9003d3e8fb283cc242e1d1962cdc17efe3749119 (ltr) and 6d82f8da9d42763172e6bd3d7ac02d20ba515930 (dev):
```
$ ls -l bin.dev/
total 492
-rw-r--r-- 1 johny johny  32800 Dec 19 12:30 0x00000.bin
-rw-r--r-- 1 johny johny 463843 Dec 19 12:30 0x10000.bin
$ ls -l bin.ltr
total 492
-rw-r--r-- 1 johny johny  32800 Dec 19 12:36 0x00000.bin
-rw-r--r-- 1 johny johny 463851 Dec 19 12:36 0x10000.bin
johny@silence:~/src/nodemcu-firmware
```
Two builds with the same modules you listed:
```
$ ls -l bin.*
bin.dev:
total 428
-rw-r--r-- 1 johny johny  32192 Dec 19 12:41 0x00000.bin
-rw-r--r-- 1 johny johny 405259 Dec 19 12:41 0x10000.bin

bin.ltr:
total 428
-rw-r--r-- 1 johny johny  32192 Dec 19 12:39 0x00000.bin
-rw-r--r-- 1 johny johny 405267 Dec 19 12:39 0x10000.bin
```

You mentioned that you're seeing ssl as an extra item being linked in after my changes. Well, it *should* linked in - it's used by the `net` module! And smartconfig is used by `wifi`. i2c_master is surely related to the enabled `i2c` module? Same as gpio16 being for the `gpio` module? Are you sure you're not comparing apples with oranges here Terry?

TerryE:
It was very late when I did this, and I was tired, but I simply swapped doing a switch between the dev and module-ltr-cleanup branches with a make clean and make between each. If sound as is if one build has SSL disabled. Not sure why the other two modules were dropped. 

But I am testing on a 512 Kb ESP-01 so the default build won't even load. So I have to use a much reduced selection to get a set of binaries that will actually boot. I suppose that I could use an ESP-12E, but you have the double whammy of having to load larger binaries at half the speed, so I prefer the 01 for this type of repeated load tests.

Just to check: you are taking the builds through to loading onto an ESP target and running a basic Lua health check?

TerryE:
OK, so what am I doing wrong: 
```
terry@nodemcu:~/nodemcu-firmware$ git status
On branch dev
Your branch is up-to-date with 'nodemcu/dev'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.build/
	luac.cross

nothing added to commit but untracked files present (use "git add" to track)
terry@nodemcu:~/nodemcu-firmware$ git log --pretty="format:%h¬%ai¬%aN¬%s" -n 10 dev
4708b2a¬2015-12-18 00:39:32 +0000¬Terry Ellison¬Merge pull request #858 from marcelstoer/dev-new-docker-reference
da5834e¬2015-12-17 20:21:47 +0100¬Marcel Stör¬Add a few small corrections and new Docker link
6d82f8d¬2015-12-17 12:57:25 +1100¬Johny Mattsson¬Merge pull request #857 from devsaurus/dev-spi_cpol
5e9bf95¬2015-12-16 22:23:04 +0100¬devsaurus¬cpol=high is not implemented due to insufficient hw docs
53ee80e¬2015-12-16 10:53:58 +0000¬Terry Ellison¬Merge pull request #841 from jmattsson/flash-addr-cleanup
fd91131¬2015-12-16 18:59:55 +1100¬Johny Mattsson¬Corrected the 4-page alignment for start of fs.
4461fcd¬2015-12-15 22:30:04 +0100¬Arnim Läuger¬Merge pull request #853 from pastukhov/dev
5cb27f8¬2015-12-15 09:35:20 +0300¬Artem Pastukhov¬Add ImageMagic info.
6009602¬2015-12-14 11:41:10 +0300¬Artem Pastukhov¬Some polish
4b0cec1¬2015-12-14 11:34:11 +0300¬Artem Pastukhov¬* Add u8glib readme.
terry@nodemcu:~/nodemcu-firmware$ cp {/tmp,app/include}/user_modules.h
terry@nodemcu:~/nodemcu-firmware$ git diff -- app/include/user_modules.h
diff --git a/app/include/user_modules.h b/app/include/user_modules.h
index 1a4646e..3004a14 100644
--- a/app/include/user_modules.h
+++ b/app/include/user_modules.h
@@ -12,42 +12,39 @@
 #define LUA_USE_BUILTIN_DEBUG_MINIMAL // for debug.getregistry() and debug.traceback()
 
 #ifndef LUA_CROSS_COMPILER
-#define LUA_USE_MODULES
 
-#ifdef LUA_USE_MODULES
-#define LUA_USE_MODULES_ADC
+//#define LUA_USE_MODULES_ADC
 #define LUA_USE_MODULES_BIT
 //#define LUA_USE_MODULES_BMP085
-#define LUA_USE_MODULES_CJSON
-#define LUA_USE_MODULES_COAP
-#define LUA_USE_MODULES_CRYPTO
-#define LUA_USE_MODULES_DHT
+//#define LUA_USE_MODULES_CJSON
+//#define LUA_USE_MODULES_COAP
+//#define LUA_USE_MODULES_CRYPTO
+//#define LUA_USE_MODULES_DHT
 //#define LUA_USE_MODULES_ENDUSER_SETUP // USE_DNS in dhcpserver.h needs to be enabled for this module to work.
 #define LUA_USE_MODULES_FILE
 #define LUA_USE_MODULES_GPIO
 //#define LUA_USE_MODULES_HX711
 #define LUA_USE_MODULES_I2C
-#define LUA_USE_MODULES_MQTT
+//#define LUA_USE_MODULES_MQTT
 #define LUA_USE_MODULES_NET
 #define LUA_USE_MODULES_NODE
 #define LUA_USE_MODULES_OW
-#define LUA_USE_MODULES_PWM
-#define LUA_USE_MODULES_RC
+//#define LUA_USE_MODULES_PWM
+//#define LUA_USE_MODULES_RC
 #define LUA_USE_MODULES_RTCFIFO
 #define LUA_USE_MODULES_RTCMEM
-#define LUA_USE_MODULES_RTCTIME
-#define LUA_USE_MODULES_SNTP
+//#define LUA_USE_MODULES_RTCTIME
+//#define LUA_USE_MODULES_SNTP
 #define LUA_USE_MODULES_SPI
 #define LUA_USE_MODULES_TMR
-#define LUA_USE_MODULES_TSL2561
-#define LUA_USE_MODULES_U8G
+//#define LUA_USE_MODULES_TSL2561
+//#define LUA_USE_MODULES_U8G
 #define LUA_USE_MODULES_UART
 //#define LUA_USE_MODULES_UCG
 #define LUA_USE_MODULES_WIFI
-#define LUA_USE_MODULES_WS2812
+//#define LUA_USE_MODULES_WS2812
 
-#endif /* LUA_USE_MODULES */
-#endif
 
+#endif  /* LUA_CROSS_COMPILER */
 #endif /* __USER_MODULES_H__ */
terry@nodemcu:~/nodemcu-firmware$ make clean
make -C app clean;
make[1]: Entering directory `/home/terry/nodemcu-firmware/app'
make -C user clean;  make -C driver clean;  make -C json clean;  make -C upgrade clean;  make -C platform clean;  make -C libc clean;  make -C lua clean;  make -C lwip clean;  make -C coap clean;  make -C mqtt clean;  make -C u8glib clean;  make -C ucglib clean;  make -C smart clean;  make -C wofs clean;  make -C modules clean;  make -C spiffs clean;  make -C cjson clean;  make -C crypto clean;  make -C dhtlib clean;  make -C tsl2561 clean;
make[2]: Entering directory `/home/terry/nodemcu-firmware/app/user'
#
# ....  Rest of make O/P trimmed
#
terry@nodemcu:~/nodemcu-firmware$ make
make[1]: Entering directory `/home/terry/nodemcu-firmware/app'
make[2]: Entering directory `/home/terry/nodemcu-firmware/app/user'
DEPEND: xtensa-lx106-elf-gcc -M -I../../sdk-overrides/include -I../../sdk/esp_iot_sdk_v1.4.0/include -Os -ffunction-sections -fno-jump-tables -fdata-sections -g -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals -D__ets__ -DICACHE_FLASH -DLWIP_OPEN_SRC -DPBUF_RSV_FOR_WLAN -DEBUF_LWIP -I include -I ./ -I ../../include/ets -I ../libc -I ../platform -I ../lua -I ../wofs -I ../spiffs -I ../include -I ./ -I ../../include -I ../../include/eagle user_main.c
DEPEND: xtensa-lx106-elf-gcc -M -I../../sdk-overrides/include -I../../sdk/esp_iot_sdk_v1.4.0/include -Os -ffunction-sections -fno-jump-tables -fdata-sections -g -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals -D__ets__ -DICACHE_FLASH -DLWIP_OPEN_SRC -DPBUF_RSV_FOR_WLAN -DEBUF_LWIP -I include -I ./ -I ../../include/ets -I ../libc -I ../platform -I ../lua -I ../wofs -I ../spiffs -I ../include -I ./ -I ../../include -I ../../include/eagle user_exceptions.c

make[2]: Leaving directory `/home/terry/nodemcu-firmware/app/tsl2561'
xtensa-lx106-elf-gcc -L../sdk/esp_iot_sdk_v1.4.0/lib -L../sdk/esp_iot_sdk_v1.4.0/ld  -Wl,--gc-sections -Wl,-Map=mapfile -nostdlib -T../ld/nodemcu.ld -Wl,@../ld/defsym.rom -Wl,--no-check-sections -Wl,--wrap=_xtos_set_exception_handler -Wl,-static -Wl,--start-group -lc -lgcc -lhal -lphy -lpp -lnet80211 -lwpa -lmain -ljson -lsmartconfig -lssl user/.output/eagle/debug/lib/libuser.a driver/.output/eagle/debug/lib/libdriver.a json/.output/eagle/debug/lib/libjson.a upgrade/.output/eagle/debug/lib/libupgrade.a platform/.output/eagle/debug/lib/libplatform.a libc/.output/eagle/debug/lib/liblibc.a lua/.output/eagle/debug/lib/liblua.a lwip/.output/eagle/debug/lib/liblwip.a coap/.output/eagle/debug/lib/coap.a mqtt/.output/eagle/debug/lib/mqtt.a u8glib/.output/eagle/debug/lib/u8glib.a ucglib/.output/eagle/debug/lib/ucglib.a smart/.output/eagle/debug/lib/smart.a wofs/.output/eagle/debug/lib/wofs.a spiffs/.output/eagle/debug/lib/spiffs.a cjson/.output/eagle/debug/lib/libcjson.a crypto/.output/eagle/debug/lib/libcrypto.a dhtlib/.output/eagle/debug/lib/libdhtlib.a tsl2561/.output/eagle/debug/lib/tsl2561lib.a modules/.output/eagle/debug/lib/libmodules.a -Wl,--end-group -lm -o .output/eagle/debug/image/eagle.app.v6.out 
../tools/esptool.py elf2image .output/eagle/debug/image/eagle.app.v6.out -o ../bin/
make[1]: Leaving directory `/home/terry/nodemcu-firmware/app'
#
# ....  Rest of make O/P trimmed
#
terry@nodemcu:~/nodemcu-firmware$ ls bin
0x00000.bin  0x10000.bin
terry@nodemcu:~/nodemcu-firmware$ ls bin -l
total 316
-rw-rw-r-- 1 terry terry  32048 Dec 19 15:51 0x00000.bin
-rw-rw-r-- 1 terry terry 290267 Dec 19 15:51 0x10000.bin
terry@nodemcu:~/nodemcu-firmware$ 
```

TerryE:
OK, I've found it.  I'd taken out the following from `app/include/user_modules.h` so all of the module defines were being made, but `LUA_USE_MODULES` itself wasn't
```
#define LUA_USE_MODULES
#ifdef LUA_USE_MODULES
...
#endif /* LUA_USE_MODULES */
```
and was copying back the user_modules.h.  This is only used once in linit.c with an \#ifdef around `#include "modules.h".   What I don't understand is that one of the aims of this tidy up is that we still need `modules.h` for it to work, but we're trying to get rid of this.

I am still interest in your reply to "what's wrong with a simple config file rather than this complexity?"

TerryE:
I don't like the choice of  programmer interface (needing to keep and edit the `user_modules.h` file), but this a stepwise improvement and better the current approach. It works.  So I am committing this, but might wish to revisit this in the future.  For now we have other priorities.

PR #849, opened 2015-12-13
Adds a module for the APA102 "SuperLED"

Nervengift:
Basically what is says in the caption: module for the [APA102 "SuperLED"](https://cpldcpu.wordpress.com/2014/08/27/apa102/) using the same API as the WS28xx modules (with the only difference that "pin" is now the number of the SPI interface to use)


pastukhov:
Please provide some lua code example.

marcelstoer:
PRs must be created against `dev`. See https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md, GitHub actually points you to it when you open a PR.

PR #852, opened 2015-12-14
Added support for the APA102 LED.

robertfoss:
This patch provides support for the APA102 LED.

The protocol is SPI based, however the platform wide SPI implementation is slightly incompatible with the particular brand of SPI that APA102 speaks.
Not relying on the generic SPI implementation also has the benefit of supporting more than one APA102 strip.

In terms of optimization this code is fairly optimized, but there is probably room for improvements to the bitbanging. The bitbanging has been tested in both 80MHz and 160MHz modes.


marcelstoer:
PRs must be created against `dev`. See https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md, GitHub actually points you to it when you open a PR.

robertfoss:
Opening a new PR against DEV.

PR #853, opened 2015-12-14
Readme for u8glib app to dev

pastukhov:
Readme for u8glib app.


devsaurus:
Nice approach - thanks!
Please add a reference what `convert` is supposed to be. I assume it's a component of the ImageMagick package?

pastukhov:
I have add an info about ImageMagic package

PR #854, opened 2015-12-14
Added support for the APA102 LED.

robertfoss:
This PR adds support for the APA102 LED.

The support is not based around the platform provided SPI implementation since it isn't flexible enough to offer meaningful improvements, and would impose limitations for how many and where APA102 strips are connected. Looking forward there might be a reason to go back to SPI if ESP32 support is added to NodeMCU since it actually has SPI with DMA support.
Note that the brand of SPI spoken by the APA102 is somewhat non-standard in that it requires 50% more clock cycles than the number of bits of data that are being sent, so SPI support might even with an ESP32 not be viable.

This API also exposes the per LED PWN brightness setting that APA102 LEDs provide unlike WS2812 LEDs.

The actual bitbanging could potentially be done faster (and has been tested with 160MHz), but is currently reasonably optimized.

**Lua API**

```
Lua: apa102.write(data_pin, clock_pin, "string")
Byte quads in the string are interpreted as (brightness, B, G, R) values.
Only the first 5 bits of the brightness value is actually used (0-31).

apa102.write(1, 3, string.char(31, 0, 255, 0)) uses GPIO5 for DATA and GPIO0 for CLOCK and sets the first LED green, with the brightness 31 (out of 0-32)
apa102.write(5, 6, string.char(255, 0, 0, 255):rep(10)) uses GPIO14 for DATA and GPIO12 for CLOCK and sets ten LED to red, with the brightness 31 (out of 0-32).
                                                        Brightness values are clamped to 0-31.
```


TerryE:
Supercedes #852

jmattsson:
Just a couple of small points:
  - Given the loops, is there any real potential of getting bitten by the watchdog? I have no concept of what the runtime bounds would actually be.
  - ICACHE_FLASH_ATTR doesn't do anything in NodeMCU. The linker script places functions into flash automatically.
  - It potentially collides with #842.

robertfoss:
If a buffer that is too long is trying to be sent the watchdog might catch it. However this is true for the current ws2812 and ws2801 modules too.

ICACHE_FLASH_ATTR removed.

TerryE:
The WDT is there for a reason. There is a guideline that no task or cb should take no more than 10-15 mSec, or the espconn stack could experience internal timeouts and fail. You should either ensure that the overall timing of less than 10 mSec, or allow block updates, or include a health warning that of you used this module then the Wifi could fail.

robertfoss:
@TerryE An error could be returned if the buffer provided is longer than length X. Currently I don't have any access to an oscilloscope. So let's conservatively make the assumption that pushing a single LED (32 bits) out takes 1000 cycles. 80MHz * 10mSec / 1000 Hz/LED = 800k LEDs.
I could put that check into the code if you guys feel like it, but if 800k is correct we have quite a bit of headroom. Also tell me if my math or assumptions are wrong.

TerryE:
@robertfoss, Robert that type of sanity check is all that I was wanting.  If the general duration for the sorts of led strings that the esp would be driving is well under a mSec, then this isn't an issue :smiley:.

robertfoss:
@TerryE 100k LED check added.

robertfoss:
Are there any issues I could help close to have this PR merged?

TerryE:
@robertfoss Robert we are in the process of trying to put out an SDK 1.5 release which also simplifies the way you integrate modules into the firmware core and sorts of an architectural issue with how ISRs trigger Lua functionality. We will need to changes this module slightly, so I would prefer to hold off merging this for a week or so. Is there any reason why we can't wait?

robertfoss:
Sure, holding off for a bit is fine. I just didn't know what was going on.

TerryE:
Thanks Robert.  Much Appreciated.  Merry Christmas if we don't speak before. :)

robertfoss:
What's left on the todo list for getting this module merged?

devsaurus:
The main items would be:
* rebase on current `dev`
* adapt to the new module registration that was introduced with 6b6456be478e8b72d7a5af2ae1f375ae8d85151e et al.
* add API documentation in `docs/en/modules`

Unfortunately, this PR missed some major reworks of the code base...

robertfoss:
Pushed fixes to comply with the above bullets.

marcelstoer:
A small detail, the documentation for any new module should be registered in `mkdocs.yml` e.g. https://github.com/nodemcu/nodemcu-firmware/blob/dev/mkdocs.yml#L34. I'll do that once it's merged.

robertfoss:
Thanks!

Issue #863, opened 2015-12-19
ENDUSER_SETUP_AP_SSID must be defined even if the module is not being used.


TerryE:
@robertfoss, If you are not using End User Setup (i.e `LUA_USE_MODULES_ENDUSER_SETUP` is not defined in user_modules) then the make still fails if `ENDUSER_SETUP_AP_SSID` is not defined.  This shouldn't happen.  Before its first use: `ENDUSER_SETUP_AP_SSID:643` there should be conditional logic to define it to be an empty string or equiv if undefined.

robertfoss:
Agreed. I will supply a PR.

TerryE:
Thanks Robert :smile:

Issue #719, opened 2015-11-03
Policy for maintaining open issues


TerryE:
The nodeMCU team have suspended active contributions to this project until further notice.  However, the support and maintenance of the project is being continued by a small number of independent contributors until such a time as the nodeMCU team can re-engage.  Our priority is to produce a stable release:

*    Based on the current SDK (now at 1.5.1 ot time of posting).   
*    Including Johny Mattson's unaligned access to flash patch.
*    Including Terry Ellison's strip_debug patch
*    John and Terry's refactoring of duplicated string and memory utilities.
*    Improved integration with Marcel Stör's, [Custom Build Service](http://nodemcu-build.com/)
*    Including improved user documentation and Lua examples.
 
The nodfeMCU firmware is a fairly thin [eLua](http://www.eluaproject.net/overview)-based implementation layered over this SDK, and by rebaselining to this current version, this will address many report firmware issues that are in practice a result of known SDK issues (see [SDK Release notes](http://espressif.com/new-sdk-release/)). Our other enhancements will significantly increase the effective RAM available to  Lua programs.

However, in order to achieve these aims with our limited resources we have to enforce a strict scope on issues that we will accept here.  We will specifically close:

*    Any questions of a "how do I ..." or a "I can't get this to work ..." (Please refer these application developer forums such as StackOverflow with the [`esp8266`](http://stackoverflow.com/tags/esp8266), [`nodemcu`](http://stackoverflow.com/tags/nodemcu) or [`Lua`](http://stackoverflow.com/tags/lua) tags, or [esp8266.com](http://www.esp8266.com/viewforum.php?f=17))
*    Any enhancement requests where the originating poster is not proposing to resource the development him or herself as we don't have the effort to resource additional enhancements
*    Any issues relating to builds prior to 1.5 as all work on these has stopped.  However, we encourage posters to upgrade their firmware to a SDK-1.5.1 `dev` build and if the issues still exists then we will investigate it as per the following item. 
*    Any bug reports with a SDK 1.5.1 based build that does not define the build baseline and include a standalone example (in the form of a gist or inline code) which clearly demonstrates the bug.

Notwithstanding this, if we can give constructive feedback then we may still provide guidance on such closed issues, but we reserve the right not to comment further.

[...]

PR #866, opened 2015-12-21
Fixed compilation error when the enduser_setup module is disabled

robertfoss:
Fixed compilation error when the enduser_setup module is disabled yet requires ENDUSER_SETUP_AP_SSID to be defined.

As discussed in #863.

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -640,6 +640,10 @@ static void enduser_setup_ap_start(void)
   struct softap_config cnf;
   c_memset(&(cnf), 0, sizeof(struct softap_config));
 
+#ifndef ENDUSER_SETUP_AP_SSID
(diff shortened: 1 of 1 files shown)


JanGoe:
hello,

you have any lua code sampel how i can use the enduser function's ?

wenn i use the sampel in the doku, and i go with my iphone to "http://192.168.4.1" and i enter my wlan ssid and password and click submit i get in safarie a Pars error

TerryE:
@JanGoe Jan, this list is for the firmware developers.  Please so #719 for help on application programmers forums.  I will close this call, but perhaps @robertfoss Robert might give you a quick pointer.

robertfoss:
@JanGoe, I've had similar issues myself since I switched to a new phone.
And will look into it. Not in the very near future though.
In the meantime I'd suggest trying to connect using a different device,
like a laptop.

Issue #790, opened 2015-11-24
ws2812 buffer


Alkorin:
Dunno if it's only me, but coming from C, playing with immutable strings with the ws2812 library is painful: when you have a long strip of leds, it's easier to have a buffer which represent your current strip and be able to set any byte in it.

I've written a PoC (my first hack with a lua library...) of a such buffer:
https://github.com/nodemcu/nodemcu-firmware/compare/dev...Alkorin:ws2812-buffer

Current API:
```
buffer = ws2812.newBuffer(nbLeds)
ws2812.setBuffer(buffer, led, G, R, B)
ws2812.writeBuffer(pin, buffer)
```

Next step, be able to call set on the buffer and have:
```
buffer = ws2812.newBuffer(nbLeds)
buffer.set(led, G, R, B)
ws2812.writeBuffer(pin, buffer)
```

Any advice are welcome :)

devsaurus:
I like your idea of having a persistent and structured buffer.

You allocate the buffer as Lua userdata - `ws2812_writergb()` could detect and accept this as payload parameter for the LEDs in parallel to the current string type. Then there's no need to have `ws2812.writeBuffer()` at all.

Merging the set functionality into the buffer object is for sure quite elegant and keeps the API lean.

Alkorin:
Thanks @devsaurus will do it :)

Alkorin:
Done, new API:

```
buffer = ws2812.newBuffer(nbLeds)
buffer:fill(G, R, B)
buffer:set(led, G, R, B)
ws2812.write(pin, buffer)
```

Simple led chaser:

```
i=0;
a = ws2812.newBuffer(16);
tmr.alarm(0, 100, 1, function() i=i+1; a:fill(0, 0, 0); a:set(i%16, 0, 255, 0); ws2812.write(4, a); end)
```

Still have to understand how to declare my metatable without `LUA_OPTIMIZE_MEMORY`

jmattsson:
@Alkorin Try [this link](http://www.eluaproject.net/doc/v0.9/en_arch_ltr.html) for the eLua LTR internals.

TerryE:
@Alkorin, Thomas, a small Lua style optimization.  You should avoid using globals as locals / upvalues generate shorter and more efficient runtime code. Also have you thought of provided a `write(pin)` method to the buffer so that this could be written:

```Lua
local i, a = 0, ws2812.newBuffer(16)
[...]

PR #869, opened 2015-12-22
Add support for ws2812 to accept table as a data source.

bbktsk:
Add support for ws2812 to accept table as a data source. Currently, it supports strings only and since lua strings are read only, this causes serious performance issues when working with larger number of LEDs. This commit adds new function write_table_rgb, which accepts a table of three-element rgb tables as an argument.
Also removed some code duplication to clean-up.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -23,6 +23,16 @@ static void ICACHE_RAM_ATTR ws2812_write(uint8_t pin, uint8_t *pixels, uint32_t 
   uint8_t *p, *end, pixel, mask;
   uint32_t t, t0h, t1h, ttot, c, start_time, pin_mask;
 
+  // Initialize the output pin and wait a bit
+  platform_gpio_mode(pin, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
+  platform_gpio_write(pin, 0);
+
+  // from now on, use translated pin number.
+  pin = pin_num[pin];
+
+  // Send the buffer
+  ets_intr_lock();
+
   pin_mask = 1 << pin;
   p =  pixels;
   end =  p + length;
@@ -52,6 +62,7 @@ static void ICACHE_RAM_ATTR ws2812_write(uint8_t pin, uint8_t *pixels, uint32_t 
       mask = 0x80;
     }
   }
+  ets_intr_unlock();
 }
 
 // Lua: ws2812.writergb(pin, "string")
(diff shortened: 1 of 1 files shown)


TerryE:
> this causes serious performance issues when working with larger number of LEDs.

Only true if you are using a continuous pallet.  If you are sticking to a fixed set of 16 colours then you will only use 16 strings no matter how many lights.  I would suspect that you will often want runs or repeat patterns so you could consider enriching the vector syntax.  Just a suggestion.

bbktsk:
@TerryE I do not follow. First, limiting myself to 16 colors with device capable of showing 16 millions is, well, limiting. Second, I'm driving a strip of 144 leds and although reusing predefined strings will be faster, I think that overhead of constructing a new control string out of 144 substrings every time will be still much worse than just overwriting a table. Or am I missing something?

TerryE:
At the moment, the EGC is fired after every alloc, so this can slow down apps up to 10x. All I an saying is that sometimes getting good solid performance isn't very intuitive, so it's week with experimenting and benchmarking.

bbktsk:
I found issue #790. Although the solution is different than mine, there's one interesting thing it does: instead of two separate functions for different argument types it uses a single function that dispatches based on argument type. Is this considered a good Lua style? Should I change my pull request to do the same thing?

TerryE:
Why don't you two discuss this and come to a consensus.

bbktsk:
I do not feel strongly one way or the other. My Lua experience is quite
limited and I do not know what is considered a good style with regard to
"overloaded" functions.

On Tue, Dec 22, 2015 at 7:23 PM, Terry Ellison <notifications@github.com>
wrote:

> Why don't you two discuss this and come to a consensus.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nodemcu/nodemcu-firmware/pull/869#issuecomment-166696236>
> .
>



-- 
--
Bob Koutsky

Alkorin:
@bbktsk I've a current dev which accept a {g,r,b} table on the set function, I'll push it after some code cleaning.

For the moment I achive a rainbow with a 300 leds strips at ~60fps

Alkorin:
Done: https://github.com/Alkorin/nodemcu-firmware/commits/ws2812-buffer

I currently have performance issues due to luaL_checkudata which is slow, if I replace it by lua_touserdata (same thing but without type check...) it's ~5 times faster :)

I've a precalculated colorMap
```
rainbow = {}
for i = 0, 179 do
  rainbow[i] = {makeColor(i*2, 100, 10)}
end
```

And here is my loop:
```
local i, buf = 0, ws2812.newBuffer(300); local set = ws2812.set; tmr.alarm(0, 20, 1, function()
    i = i + 1
    
    local rainbow = rainbow;
    for j = 0,299 do
        set(buf, j, rainbow[(j+i)%180])
    end
    buf:write(4)
end)
```

Current timings are : ~6ms for the loop + 10ms streaming to the strip => >60fps

marcelstoer:
@bbktsk @Alkorin what's your plan with this? At least the code should be rebased on `dev` before we can merge it.

Issue #875, opened 2015-12-23
DHCP Device name contains illegal underscore in development build


boatmon0:
In the development build the default device name my dhcp router shows for me is ESP_058CDC.  The underscore in the name seems to mess up the router and the device is not reachable by that name. (ping or http://)  I built 12/22/2015 at nodemcu-build.com, installing their default packages.  
Please replace the underscore with a dash.

Is it possible for me to change the device name?  I would like to provide my own name. This device name should probably default to the device id, to maintain uniqueness.

![image 1](https://cloud.githubusercontent.com/assets/377449/11979485/63a6028c-a959-11e5-9ff4-0eba67f388eb.png)

devsaurus:
The default dhcp hostname is constructed automatically without means to influence the composition. It could be set to an arbitrary string via [wifi_station_set_hostname()](https://github.com/nodemcu/nodemcu-firmware/blob/master/sdk/esp_iot_sdk_v1.4.0/include/user_interface.h#L252) but there's no Lua binding for this function yet. This might involve some trial and error to get it working.
Since we're currently focusing on basic topics in the firmware, there's probably no activity at the moment to integrate new SDK functions. A pull request with such an enhancement is always welcome, though.

udif:
http://stackoverflow.com/questions/4919889/is-there-a-standard-that-defines-what-is-a-valid-ssid-and-password

It seems that there is no such thing as an "illegal" character in an SSID. As such, the nodemcu behavior is completely legal. you may want to override the default name, but it is definitely not "illegal".

TerryE:
As @devsaurus says, this is SDK functionality so you really need to report the bug to Espressif.  It's an enhancement to add an extra API to override the default.

Where in RFC2131 does it forbid the use of _ ?

udif:
sorry, I misread your issue, because the SSID is also of the same form, ESP_xxyyzz, where xxyyzz are the 3 lower bytes of the MAC address.

You are talking ABOUT the DHCP name, not SSID.

boatmon0:
Wow, what an active group.  Great stuff, and you have at least pointed me in the right direction.  

I found the problem when I named a device using the underscore.  Here is my reference.

https://en.wikipedia.org/wiki/Hostname

[...]

PR #878, opened 2015-12-26
Add functions: wifi.sta.sethostname and wifi.sta.gethostname

dnc40085:
This PR exposes SDK functions `wifi_station_set_hostname` and `wifi_station_get_hostname` referred to in issue #875.

I also implemented the option to set hostname at boot by adding `luaopen_wifi` and have it set the hostname according to `user_config.h` and defaults to `NODE-xxxxxx`(xxxxxx being the last digits of mac address) if a hostname is not specified.

#### Description
Set station hostname.

#### Syntax
wifi.sta.sethostname(hostname)

#### Parameters
 - `hostname`: Hostname must only contain letters, numbers and hyphens(`-`) and be 32 characters or less with first and last character being alphanumeric.

#### Returns
 - `true`: hostname was successfully set
 - `false`: hostname was not successfully set

#### Example
   
```lua
  if (wifi.sta.sethostname("NodeMCU")==true) then
    print("hostname was successfully changed")
  else
    print("hostname was not changed")
  end

```

#### Description
Get current station hostname.

#### Syntax
wifi.sta.gethostname()

#### Parameters
 - `nil`

#### Returns
 - `hostname`: Currently configured hostname.

#### Example
   
```lua
  print("Current hostname is: \""..wifi.sta.gethostname().."\"")
```

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -83,6 +83,17 @@
 
 #define ENDUSER_SETUP_AP_SSID "SetupGadget"
 
+/*
+ * A valid hostname only contains alphanumeric and hyphen(-) characters, with no hyphens at first or last char
+ * if WIFI_STA_HOSTNAME not defined: hostname will default to NODE-xxxxxx (xxxxxx being last 3 octets of MAC address)
+ * if WIFI_STA_HOSTNAME defined: hostname must only contain alphanumeric characters
+ * if WIFI_STA_HOSTNAME_APPEND_MAC not defined: Hostname MUST be 32 chars or less
+ * if WIFI_STA_HOSTNAME_APPEND_MAC defined: Hostname MUST be 26 chars or less, since last 3 octets of MAC address will be appended
+ * if defined hostname is invalid: hostname will default to NODE-xxxxxx (xxxxxx being last 3 octets of MAC address)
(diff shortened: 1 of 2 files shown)


TerryE:
Thanks. I'll do the dev 1.5 merge before implementing this.

dnc40085:
I was wondering, since both this and the event monitor add `luaopen_wifi` to `wifi.c`and would inevitably cause conflicts when merged. Would it be better to just combine them in to one pull request?

dnc40085:
anybody?

jmattsson:
@TerryE any reasons why we shouldn't get this merged in at this point? Other than people having been busy, I mean?

TerryE:
This implementation is fine. The general issue as we discussed with #935 is that we are used to coding for clarity rather than generated code size. IMO, this is fine for a PC implementation, but here we are trying to make best use of a reasonably limited resource, the flash rom size, so this tilts the balance slightly to also thinking about generated code size, e.g. using isaplha rather than reimplementing this and folding conditional branches where practical.  However we also have another scarse resource: the time available to our contributors.

marcelstoer:
*/* will add this to `wifi.md` in `newdocs`.

PR #879, opened 2015-12-27
support airkiss and esptouch at the same time

SmartArduino:
in /app/modules/wifi.c
change if ( smart_type > 1 ) return luaL_error( L, "wrong arg range" );
to if ( smart_type > 2 ) return luaL_error( L, "wrong arg range" );


TerryE:
@SmartArduino, please read our rules for contribution. We don't accept PRs to master. Dev is the correct branch to use. We would also extect a discussion of the proposed change, tests carriedout etc.

I leave it to you to close this request and prepare this for dev

PR #886, opened 2015-12-29
Add HTTP client module and documents.

vowstar:
HTTP client Demo code
```lua
-- Support HTTP and HTTPS, For example
-- HTTP POST Example with JSON header and body
http.post("http://yourtesting.url/",
           "Content-Type: application/json\r\n", 
           "{\"hello\":\"world\"}", 
            function(code, data) 
                print(code) 
                print(data) 
            end)
-- HTTPS GET Example with NULL header
http.get("https://www.vowstar.com/nodemcu/","",
            function(code, data) 
                print(code) 
                print(data) 
            end)
-- You will get
-- > 200 
-- hello nodemcu
-- HTTPS DELETE Example with NULL header and body
http.delete("https://10.0.0.2:443","","",
            function(code, data) 
                print(code) 
                print(data) 
            end)
-- HTTPS PUT Example with NULL header and body
http.put("https://testput.somewhere/somewhereyouput.php","","",
            function(code, data) 
                print(code) 
                print(data) 
            end)
-- HTTP RAW Request Example, use more HTTP/HTTPS request method
http.request("http://www.apple.com:80/library/test/success.html","GET","","",
            function(code, data) 
                print(code) 
                print(data) 
            end)

```
Please try!  :)

Issue #771, opened 2015-11-17
Re-write the README.md


marcelstoer:
#769 is the first attempt at getting the `README.md` right. I'd like to use this issue to discuss what direction we should be taking once that PR is merged.

We (i.e. the current collaborators) already agreed that:
- The code samples need to be removed. Examples belong in the examples folder and at best the README should contain a link or reference to these.
- It should be shorter i.e. more concise.
- It should contain a summary of the project status referring to #719 for our policy on accepting new issues.

jmattsson:
- Link to the FAQ
- Link to recommended forums for getting user help (closely related to 719)

nickandrew:
I suggest a .md file for each module, fully documenting what the module does. Example code snippets can help here.

Code samples in README.md can also assist somebody who's thinking whether they should use NodeMCU, by showing how *easy* it is to program.

marcelstoer:
@nickandrew API documentation is handled in #774, I just created it.

ghusson2:
Maybe there should be some advises about how to use nodemcu in general. I mean for example, adding a timer before running scripts at startup (in order to have the time to upload a code update in case of big bug), how to cut long instructions sets in different scripts and so on.
Maybe some advises from http://www.esp8266.com/wiki/doku.php?id=nodemcu-unofficial-faq.
And maybe and most important some "begin from zero here" (i did it for my fablab in french : http://fil.crepp.org/pages/view/51/nodemcu). On this last, I can translate it if you want (but linux oriented).

marcelstoer:
> Maybe some advises from http://www.esp8266.com/wiki/doku.php?id=nodemcu-unofficial-faq.

My plan is to integrate the FAQ into the wiki replacement with #774 (and move it from unofficial to official finally :wink: ).

> On this last, I can translate it if you want

Once the documentation is based on RTD serving the docs in multiple languages should be a breeze.

ghusson2:
Ok, just perfect :+1:

nickandrew:
@ghusson2 I'll compare your "begin from zero" document with the latest README; there might be places I can add detail. No need to translate from French; it's obvious what most of the document means :-)

One last thing, please delete your reference to nodemcu_float_0.9.6 as that version is deprecated. Current README.md covers: cloud build (as your doc does), building in a Docker container, and installing the build toolchain directly.

ghusson2:
@nickandrew : "latest README" : oh yes, indeed :)
Thank you I updated my doc.
[...]

PR #887, opened 2015-12-29
Remove all version data and combine 33 commits to 1.

vowstar:
Remove all version data and combine 33 commits to 1. Add http module and documention, including fix httpclient bug. etc.
https://github.com/nodemcu/nodemcu-firmware/pull/886#issuecomment-167763167

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ Comment-out the #define statement for unused modules. Example:
 // #define LUA_USE_MODULES_BMP085
 #define LUA_USE_MODULES_TSL2561
 // #define LUA_USE_MODULES_HX711
+#define LUA_USE_MODULES_HTTP
 
 #endif /* LUA_USE_MODULES */
 ```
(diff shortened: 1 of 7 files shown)


jmattsson:
I'm really keen on getting an actual HTTP client implementation into NodeMCU. I haven't had a chance to play/test this one, but :+1:  in theory! :)

jmattsson:
Well, it's not without its shortcomings, notably in its error reporting. It took me some time to work out that a status of -1 was because the page I was fetching was too big. That said, I'd rather have this module than not have it, and it seems like a good starting point for further improvements. The module appears sound and I haven't run into any memory leaks. Any objections before I merge this?

marcelstoer:
> Any objections before I merge this?

Too late :wink:
- no proper docs (as you apparently learned the hard way)
- examples in `README.md`, they don't belong there anymore (see #771)

Also, it needs to be added to https://github.com/nodemcu/nodemcu-firmware/wiki/Module-origin

PR #888, opened 2015-12-29
Fix problem in wifi.sta.getap where invidual result is lost.

dnc40085:
In my project, I use `wifi.sta.getap()` to search for a specific access point and check it against a list before connecting. 
Recently, with a firmware upgrade, it stopped returning the ap's info.
I determined that somewhere along the line with one of the SDK upgrades, the first entry in bss_link, which is supposed to be blank, was dropped and the now commented-out line was dropping the individual result returned causing a blank table to be returned to the lua CB.
This PR is the fix for the issue.

Lua code to reproduce issue:
```lua
wifi.sta.getap({ssid="Known_SSID"}, function(T) print("\r") for k,v in pairs(T) do print(k.." : "..v) end end)
```

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -97,7 +97,6 @@ static void wifi_scan_done(void *arg, STATUS status)
   if (status == OK)
   {
     struct bss_info *bss_link = (struct bss_info *)arg;
-    bss_link = bss_link->next.stqe_next;//ignore first
     lua_newtable( gL );
 
     while (bss_link != NULL)


jmattsson:
Verified here. Looks good, thanks! The first entry is most certainly not blank these days (was it ever?)

It has however pointed out to me that our current Lua interface is not capable of returning information about SSIDs that is present on multiple channels, since we key off the SSID. We'd need to turn the returned entity into an array rather than a table to handle this. I'll raise an issue for that, but I'm not sure it's something we'll change.

PR #890, opened 2015-12-30
Update flash_api due to flash size map is redefined by espressif.

vowstar:
SIZE_64MBIT and SIZE_128MBIT is not supported by espressif.

### DIFF
--- a/app/platform/flash_api.c
+++ b/app/platform/flash_api.c
@@ -132,13 +132,13 @@ uint32_t flash_rom_get_size_byte(void)
             // 32Mbit, 4MByte
             flash_size = 4 * 1024 * 1024;
             break;
-        case SIZE_64MBIT:
-            // 64Mbit, 8MByte
-            flash_size = 8 * 1024 * 1024;
+        case SIZE_16MBIT_8M_8M:
+            // 16Mbit, 2MByte
+            flash_size = 2 * 1024 * 1024;
             break;
-        case SIZE_128MBIT:
-            // 128Mbit, 16MByte
-            flash_size = 16 * 1024 * 1024;
+        case SIZE_32MBIT_8M_8M:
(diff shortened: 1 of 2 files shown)


TerryE:
@vowstar, What this patch seems to do is to remove the 8Mb and 16Mb options from our flash code.  As far as I know, no one actually produces and modules with these sized flash chips anyway, but if Espressif has explicitly withdrawn support then (i) where is the reference for this and (ii) shouldn't just treat these as an error rather than introducing the unused extra enumeration constants?

vowstar:
1. espressif's engineer said to me that esp8266 not support 64Mbit and 128Mbit flash by this way, if we are using their SDK.
2. see espressif's official sdk's root makefile, you will find the detailed information about flash size map. this file is too old and I write this before SDK v0.9.5, now after sdk v1.0.0 the parameter 64M and 128M ismapped to 16M and 32M. please check carefully

TerryE:
Thanks, What you are saying is that the reference is that you have to retro-engineer this from the sdk Makefile, and the `SPI_SIZE_MAP` for the V1.x Makefiles, has replaced the 5th and 6th entries (previously 64Mb and 128Mb) with new two 2Mb and 4Mb but with an addr of 0x101000 instead of 0x81000.

vowstar:
Yes, all from new makefile(>sdk v1.0.0). And I tested where espressif write their wi-fi configuration, It may dangerous for our filesystem(overwrite). And sorry for my poor English, thanks.

TerryE:
Your English is understandable and a _lot_ better than my Mandarin!! My vocabulary stretches to "thank-you" and even then I suspect that you would laugh at my horrible pronunciation. :open_mouth:

vowstar:
:smile:

PR #894, opened 2015-12-31
Add support for basic trigonometric functions

grytole:
Now math lib has disabled trigonometric functions, so I decided to create simple module until Lua will be able to handle this work by default.


TerryE:
@grytole, whilst this is an interesting exercise, the C libraries already provide these functions and the runtime code has only been `#ifdef`fed out to save Flash memory space for the 95%+ of developers that don't need these. Perhaps the best approach here would simply be to host this `#ifdef` constant into user_config.h rather than reimplementing existing C / assembler source in Lua.

PR #895, opened 2016-01-01
Add method to retrieve panic string from previous boot of nodemcu

pjsg:
This PR adds 
```lua
node.panicstr(offset)
```
which sets the offset in the RTC memory to save any panic string. It also returns the previous panic string (if any). The previous panic must have been saved with the same offset.

For example:
```lua
print ("Last panic", node.panicstr(64))

tmr.alarm(0, 1000, 0, function() 
  foobar() 
end)
```

This prints:
```
Last panic	stdin:2: attempt to call global 'foobar' (a nil value)
```

This only adds a single unsigned char of memory.


app/lua/lauxlib.c
```diff
@@ -806,16 +809,45 @@ static void *l_alloc (void *ud, void *ptr, size_t osize, size_t nsize) {
...
+                   msg);
 #endif
+  if (panic_str_offset != 0) {
+    size_t len = c_strlen(msg) + 1;
+    size_t avail = (RTC_SYSTEM_MEM_NUM_DWORDS + RTC_USER_MEM_NUM_DWORDS - panic_str_offset - 1) * 4;
+    if (len > avail) {
+      len = avail;
+    }
+    system_rtc_mem_write(panic_str_offset + 1, msg, len);
+    rtc_mem_write(panic_str_offset - RTC_SYSTEM_MEM_NUM_DWORDS, 0x31415926);
```
jmattsson: I'm really not keen on the mixing of system_rtc_mem_(read|write) and rtc_mem_(read|write). The SDK functions don't have a well-defined interface - in some versions of the SDK offset zero points to system rtc mem, and in others it points into the user area (and you need negative indices to reach the system area). It's to avoid that mess that we ended up rolling our own accessors.

I'd much rather see an addition to rtcaccess.h that supports sub-4byte read/writes if you really need those. It would also get rid of all these +/- RTC_SYSTEM_MEM_NUM_DWORDS.

app/lua/lauxlib.c
```diff
@@ -806,16 +809,45 @@ static void *l_alloc (void *ud, void *ptr, size_t osize, size_t nsize) {
...
   c_fprintf(c_stderr, "PANIC: unprotected error in call to Lua API (%s)\n",
-                   lua_tostring(L, -1));
+                   msg);
 #else
   luai_writestringerror("PANIC: unprotected error in call to Lua API (%s)\n",
-                   lua_tostring(L, -1));
+                   msg);
 #endif
+  if (panic_str_offset != 0) {
+    size_t len = c_strlen(msg) + 1;
```
jmattsson: Please consider using lua_tolstring() instead of needing a separate strlen() here.

app/lua/lauxlib.c
pjsg: OK -- I didn't know about the SDK history.... It is easy enough to fix.

app/lua/lauxlib.c
pjsg: Will do.

TerryE:
Philip, I personally don't support the idea of adding this as a standard option. Let me explain why. The RTC memory is a very scarce resource and should be reserved for uses where there is no practical alternative. If you want to store context between boots or deep sleeps, then you need to use this or Flash,  but the latter has a limited write cycle limit, so applications will want to leave volatile state in RTC  memory.

What you are doing here is to try to debug specific issues,, so this trick is fine if you aren't already using the memory. But this isn't something that a typical Lua developer with want to do, IMO.

TerryE:
Philip, I've been brooding about this one and I really think that you are over-engneering this.  You get a panic because as the error says: _PANIC: unprotected error in call to Lua API_.  If you want to capture the error, then surely isn't is simply to use an xpcall handler to protect your code?  You can have your own little Lua routine to do what you want with the error in this error handler routine.  Why modify the runtime node library to reimplement something that Lua provides as standard?

pjsg:
If the user doesn't want to use this feature, then they don't have to use it -- and then it doesn't touch the rtc memory at all (it only uses the rtc memory if there is a panic *and* the offset has been provided by calling node.panicstr). 

I find that the typical place that a panic happens is when there is an error in a callback (which are typically unprotected by default). 

I note that less than 10% of the frightanic users actually include rtcmem in their builds....

TerryE:
If you have a cb panicking what is wrong with just doing an xpcall here?  I know xpcall is a little clunky as you have to manipulate the arguments manually in Lua 5.1
```Lua
sk:on('receive', function(sk, data) XXX end)  -- becomes
sk:on('receive', function(sk, data) xpcall( function() XXX end, error_handler) end)  -- sk and data are upvals

pjsg:
Yes, you could replace every callback with an xpcall wrapper -- but it means that you have to go into library code when you include other lua modules. 

I'm coming round to the view that really what is required is a circular log buffer (in rtc memory) of all the console output. This way it would capture asserts (if enabled) and any other diagnostic output. Obviously this would be most useful in a headless environment. This would need to be specifically enabled to start recording....

What do you think?

TerryE:
I am happy for you to modify your own fork, as that's entirely up to you.  However, this falls short of what I would accept for a modification to the _core_ nodemcu code.   I make similar patches to help me when I am debugging complex issues, but I apply this same test to my own code so they never leave my local instance let alone get promoted to my public fork.  (Note my above edit which we cross posted on).   So still I give this a big **-1**. 

If you have a panicking application, then the easiest thing is just to accept that you need to hook up a diagnostic serial port.

The RTC memory just isn't large enough to support any sensible wrap-around buffer anyway.  What might work some form of WALFS (a Wrap-Around Logging File System) that is Flash aware and uses a fixed cyclic buffer and a tail pointer in RTC mem, with a logging function which can be hooked up to a `node.output()` callback.  And also possibly a spool-out function which will flush the log contents out to some URI using some standard protocol such as a HTTP/POST.  This could be allocated along side SPIFFS. 
 
If you want to suggest a modification to the core, then you really must think through the entire implications, including benefits, disadvantages, risks and opportunities. Creating an add-on library is a different matter entirely, such as creating your own library module with two entry points in it: `protect.setHandler(error_cb)` and `protect.call(function, args)`?

OK, when this code was cut, we didn't have call tracebacks implemented.  Maybe what is valid is to suggest that we include a proper Lua traceback as part of the PANIC.  That why, it is easy to establish which cb's need to be wrapped.

pjsg:
OK. I made the tradeoff that said that using a minimal amount of RAM (i.e. one byte) and some code space was better than using more RAM and less code (i.e. if I took the hooking approach). However, if you are happy with maybe an extra 4 or 8 bytes of RAM consumption, then I'll add hooks and a new module.

TerryE:
Philip, 

  *  If this is for you then you can do anything that you want.  
  *  If you are wanting to develop your own library then as long as t is coded soundly and falls in with our guidelines, then there is no reason why you shouldn't offer it so long as it meets a purpose.
  *  If you are proposing changes to the core then this is another ballgame entirely.  We need a compelling reason, and the changes must be architecturally sound.  

As I said, I've done diagnostic patches myself, but they don't tick these boxes so they've never left my machine.  This isn't about 1 vs 4 vs 8 bytes RAM; it's about requirements, breadth of need and architecture.  This isn't about how do I tweak my implementation; it's more about have I got the requirement right? Does my approach fit in with the Lua way of doing things? 

What is the requirement here?  I personally think that its a need to get decent crash diagnostics (like a decent traceback) after any error, including PANICs and lua_assert errors when debugging.  At a minimum these should be presented on a UART output as this will meet 90% of development scenarios.

Some form of library add-in or even core support (configurable through `user_config.h`) to implement some wrapping log partition on flash would be nice.  Taking over all or a big chunk of RTC mem is _not_ nice.   Having this as a mandatory patch is _not_ nice.  

Have you compared your suggested approach to standard Lua's or to eLua's?  

Please read through this PR's comments and reflect on this.  I doubt whether any of the committers would support this current implementation.

TerryE:
Philip,  Perhaps a middle ground is that we introduce the following API constants `LUA_MASKUNPROTCALL` and `LUA_MASKASSERT` into the core code so that the `lua_sethook()` API call can be used to establish hooks for these events.  Any developer is then free to set these hooks with hist own custom code, and it becomes a simple matter for you to develop your own module implementing your own strategy without imposing this on all developers who wish to use the nodeMCU firmware.

TerryE:
See #899 which will enable what you need in the Lua core.  I cannot support the PR in this current form.  Should any of the other committers disagree and feel that this is a better approach than #899, then please reopen,

Issue #893, opened 2015-12-31
How to sleep a longer time?


chegongcai:
How to sleep a longer time? there are my code:
time=2150000000     
node.dsleep(time,2) 
It seems  the most long time is 35 minutes,I want to let it sleep 4 hours, how can I do it ?

TerryE:
Please see #719.

PR #896, opened 2016-01-02
Fix for node.dsleep to increase maximum sleep time

dnc40085:
`node.dsleep()` is currently using the type sint32 for the variable `us` to hold the sleep time, which limits maximum sleep time to 2,147,483,647 us(~35min).
This PR changes that variable's type to uint32 to match the type used in the SDK function `void system_deep_sleep(uint32 time_in_us)` increasing maximum sleep time to 4,294,967,295 us(~71 min)

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -39,7 +39,8 @@ static int node_restart( lua_State* L )
 // Lua: dsleep( us, option )
 static int node_deepsleep( lua_State* L )
 {
-  s32 us, option;
+  uint32 us;
+  uint8 option;
   //us = luaL_checkinteger( L, 1 );
   // Set deleep option, skip if nil
   if ( lua_isnumber(L, 2) )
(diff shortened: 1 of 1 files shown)


marcelstoer:
I guess this was inspired by #893?

dnc40085:
@marcelstoer I didn't want to say that it was a fix for that issue, since they were trying to make it goto sleep for four hours. But yeah that sounds about right.

On a side note, I wonder if Espressif is eventually going to change it to use uint64, so if we want, the ESP could sleep for half a million years.

chegongcai:
I also found that the node.dsleep() just can sleep about 35min, whether using the type of int or float.This will waste a lot of power.

TerryE:
@chegongcai, waking up from deep sleep every 35 minds rather than 70 mins for maybe a few mSec isn't going to save that much battery. The 70 mins is a H/W or SDK constraint.

PR #902, opened 2016-01-04
ws2812 strip data buffer

Alkorin:
Following #790, here is my current work almost ready to merge.

Feel free to comment or ask if something seems missing.

Goal was to efficiently store a RW state buffer of the led strip. At each iteration loop it is then easier to update only a portion of it, or compute the step in multiple calls without doing insane string concatenations.

Led chaser example:
```lua
local i, dir, buf = 0, 1, ws2812.newBuffer(300); buf:fill(0, 0, 0); tmr.alarm(0, 15, 1, function()
    i=i+dir
    if i == 299 then
      dir = -1
    elseif i == 0 then
      dir = 1
    end
    buf:fade(2)
    buf:set(i, 0, 255, 0)
    buf:write(4)
end)
```

Currently, :set allows (G, R, B) tuple, { G, R B } table and "GRBGRB..." string.
In the future, I plan to add table of strings { "GRB", "GRB", "GRB", ... } and table of tables { { G, R B }, { G, R B }, { G, R B }, ... } to be able to store precomputed patterns.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -4,54 +4,46 @@
 #include "c_stdlib.h"
 #include "c_string.h"
 #include "user_interface.h"
+#include "driver/uart.h"
 
-static inline uint32_t _getCycleCount(void) {
-  uint32_t cycles;
-  __asm__ __volatile__("rsr %0,ccount":"=a" (cycles));
-  return cycles;
-}
-
-// This algorithm reads the cpu clock cycles to calculate the correct
-// pulse widths. It works in both 80 and 160 MHz mode.
-// The values for t0h, t1h, ttot have been tweaked and it doesn't get faster than this.
-// The datasheet is confusing and one might think that a shorter pulse time can be achieved.
-// The period has to be at least 1.25us, even if the datasheet says:
-//   T0H: 0.35 (+- 0.15) + T0L: 0.8 (+- 0.15), which is 0.85<->1.45 us.
-//   T1H: 0.70 (+- 0.15) + T1L: 0.6 (+- 0.15), which is 1.00<->1.60 us.
-// Anything lower than 1.25us will glitch in the long run.
+// Stream data using UART1 routed to GPIO2
+// NODE_DEBUG should not be activated because it also uses UART1
 static void ICACHE_RAM_ATTR ws2812_write(uint8_t pin, uint8_t *pixels, uint32_t length) {
-  uint8_t *p, *end, pixel, mask;
-  uint32_t t, t0h, t1h, ttot, c, start_time, pin_mask;
 
-  pin_mask = 1 << pin;
-  p =  pixels;
-  end =  p + length;
-  pixel = *p++;
-  mask = 0x80;
-  start_time = 0;
-  t0h  = (1000 * system_get_cpu_freq()) / 3022;  // 0.35us (spec=0.35 +- 0.15)
-  t1h  = (1000 * system_get_cpu_freq()) / 1477;  // 0.70us (spec=0.70 +- 0.15)
-  ttot = (1000 * system_get_cpu_freq()) /  800;  // 1.25us (MUST be >= 1.25)
+  // Data are sent LSB first, with a start bit at 0, an end bit at 1 and all inverted
+  // 0b00110111 => 110111 => [0]111011[1] => 10001000 => 00
+  // 0b00000111 => 000111 => [0]111000[1] => 10001110 => 01
+  // 0b00110100 => 110100 => [0]001011[1] => 11101000 => 10
+  // 0b00000100 => 000100 => [0]001000[1] => 11101110 => 11
+  uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
 
-  while (true) {
-    if (pixel & mask) {
-        t = t1h;
-    } else {
-        t = t0h;
-    }
-    while (((c = _getCycleCount()) - start_time) < ttot); // Wait for the previous bit to finish
-    GPIO_REG_WRITE(GPIO_OUT_W1TS_ADDRESS, pin_mask);      // Set pin high
-    start_time = c;                                       // Save the start time
-    while (((c = _getCycleCount()) - start_time) < t);    // Wait for high time to finish
-    GPIO_REG_WRITE(GPIO_OUT_W1TC_ADDRESS, pin_mask);      // Set pin low
-    if (!(mask >>= 1)) {                                  // Next bit/byte
-      if (p >= end) {
-        break;
-      }
-      pixel= *p++;
-      mask = 0x80;
-    }
-  }
+  // Configure UART1
+  // Set baudrate of UART1 to 3200000
+  WRITE_PERI_REG(UART_CLKDIV(1), UART_CLK_FREQ / 3200000);
+  // Set UART Configuration No parity / 6 DataBits / 1 StopBits / Invert TX
+  WRITE_PERI_REG(UART_CONF0(1), UART_TXD_INV | (1 << UART_STOP_BIT_NUM_S) | (1 << UART_BIT_NUM_S));
+
+  // Redirect UART1 to GPIO2
+  // Disable GPIO2
+  GPIO_REG_WRITE(GPIO_ENABLE_W1TC_ADDRESS, BIT2);
+  // Enable Function 2 for GPIO2 (U1TXD)
+  PIN_FUNC_SELECT(PERIPHS_IO_MUX_GPIO2_U, FUNC_U1TXD_BK);
+
+  uint8_t *end = pixels + length;
+  do {
+    uint8_t value = *pixels++;
+
+    // Wait enough space in the FIFO buffer
+    // (Less than 124 bytes in the buffer)
+    while (((READ_PERI_REG(UART_STATUS(1)) >> UART_TXFIFO_CNT_S) & UART_TXFIFO_CNT) > 124);
+
+    // Fill the buffer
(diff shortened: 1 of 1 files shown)


TerryE:
This is a general comment about all bit banging drivers but including this one. IMO, the general approach is unstable with the ESP SDK. The general guidelines are that interrupt handlers and disabling of interrupts should be no more than 50uS and any cb task less than 50-100 msec. We've discussed this second but not the first. You are doing a `ets_intr_lock` ...`ets_intr_unlock` around your entire routine, and if you do this there is no guarantee that anything that depends on interrupts will still be working at the end of it. OK, the bit timing within the 24 bit triplet is tight,  but this is only ~27 uS. The reset on most real devices seems to be around 50 mSec, so why not limit your lock to around the 24 bit sequence so that any pending interrupts can be delivered between 3 byte sets? The library will then be a lot more stable.

.

Alkorin:
Reset timing is 50us, not 50ms, so I should have the guarantee that the SDK will not interrupt me more than 50us.

I'll give it a try but I have already seen +/- 1ms deviation on some timing benchmarks due to interrupts :/

TerryE:
You are working within a set PC ground rules set by Espressif and their SDK. You are free to ignore them for your own code, but a Lua developer using your module might expect the Wifi to work for example with this driver and start raising issues if it doesn't.

That 1ms pf other

Alkorin:
Yes I'll check it this evening @TerryE 

I need wifi/network too, so I'll take care of it ;)

TerryE:
I realise that the datasheet says that the WS2812 reset interval is > 50µS but it doesn't say how much.  I was looking at a couple of detailed posts which retro-engineer the operational characteristics and they are very inconsistent. (For example [this one](http://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/) suggests 50 mSec!)  As fas a I can see, the sequence goes **<reset>(<RGB>)\*** where the reset is passed down the line and the first chip pops and latches the first **RGB** and passes the remaining (?low?) + (n-1) **RGB** sequence to the next chip and so on.  If this is correct then the tail-end Charlie on a 1000 long strip will see <reset+8991µS low>**RGB**.

So if you are correct in asserting the <reset> **RBG RBG RBG** <interrupt delay>** RGB . . . ** will be seen as a <reset> 3×RGB <reset> 997×RGB then we can't allow interrupts in the ~9mSec window when we are bit banging.  But in this case we should also assume that the Wifi stack _might_ have died or hung in the meantime and therefore the programming guidelines for the driver should explain this constraint end give an application example of how the Lua programmer can reset and recover the Wifi.

Such are the joys of a single core CPU and a closed source HW architecture + SDK. :cry:

TerryE:
Just brooding about this a bit more. The time for a single RGB is 27 uSec. So we can clearly disable interrupts for this. But even a stop of 10, let alone 100, LEDs gives an end to end interval that is to long to disable interrupts safely so I suggest that the obvious option is to enable interrupts between RGB codes. So we really have three options:

  *  Ignore the SDK timing guideline, disable interrupts for the entire RGB sequence, and then try to repair the damage when done (e.g. restart the WiFi and TCP stacks).

  *  Enable interrupts between RBG sets, but ignore the potential consequences of an interrupt having been delivered.

*  Enable interrupts between RBG sets, and add detection logic around the disable..enable to see if an interrupt has been delivered and have some recovery rules -- for example if the delay was more than X uSec, then reset and restart the sequence.

This last option might be the way to go if the probability of a disruptive interrupt is small and the visual impact of the restart isn't perceivable.

nickandrew:
A technique I have used very successfully on the _ATTiny85_ is that an interrupt handler does the actual bit banging (controlled by a timer, for precision) under the direction of non-interrupt code.

To describe it informally, the mainline code would start an operation (such as sending a character down the software emulated serial port) and then setup the 2nd operation (the next character to be sent) and then wait until the first operation had finished, then setup the 3rd operation, and so forth. So the interrupt code would always have 1 buffered character to send, and it would be able to start on the next character immediately after finishing the previous. There'd be no delay proceeding from one operation to the next.

Alkorin:
Another solution for my own project would be to add another µC to do the timing critical stuff, but it's not the question here.

I'll check interrupts durations to see if we could let them enabled or not. If not, we should let a note in the documentation which says that the wifi stack may be broken if the led strip is too long.

What are the guidelines for the maximum time that interrupts can be disabled ?

karrots:
Not sure if this will help things. This guy has an extensive write up on the timing needs of the WS2812 protocol

https://cpldcpu.wordpress.com/2014/01/14/light_ws2812-library-v2-0-part-i-understanding-the-ws2812/

Alkorin:
Thanks a lot @karrots , it looks like a 1 could be a lot longer, so ... we could try to deliver interrupts while we try to send a 1 :)

0 is a "short pulse", 1 is a "long pulse". If we send a "0", it's timing sensitive. If we send a 1 ... we maybe have a winner there !, Will try this tomorrow !!!

Alkorin:
Hum, no :/
Total cycle time can't exceed 9µs, so if an interrupt occurs, it should not be more than 9µs

TerryE:
@Alkorin Thomas, reread my last comment and consider my option(3).  Let's say for the sake of argument you are getting 10 interrupts per second on you  ws2812 chip.  We should treat this Poisson but let's keep this simple an assume that this is 1 every 100 mSec or so.  If you are driving a 100 LED strip then the time to relatch the RGB data is 2.7mSec, so if you are enabling interrupts every 27µSec with collision detection then roughly 1 in 40 RGB loads will have a detected collision.  Let's say this occurs after the 16 LED load. What is wrong with just letting the sequence reset and then restarting the sequence. So instead of:

*    <reset>100×RBG

You do

*    <reset>16×RBG (interrupt detected) <reset>100×RBG

Alkorin:
Yes, it works @TerryE, but as expected it resets :)

And it resets fast ... even at 160MHz where interrupts are handled in 6.5us, I've 2 or 3 currupted send per seconds. I'll implement the interrupt detection and see.

Do we have some "known" values for the maximum time that interrupts can be disabled ?

TerryE:
non-OS V1.5 SDK Programming Guide Section 2 -- Overview

> Notice:
>
>  *  Using non-OS SDK which is single-threaded, the CPU should not take long to execute tasks:
>     *   If interrupt is disabled, CPU can only be occupied in us range and the time should not bemore than 10 µs; if interrupt is not disabled, it is suggested that CPU should not be occupied more than 500 ms.
>     *   We suggest using a timer to check periodically, if users need to call os_delay_us or function
while , or function for in timer callback, please do not occupy CPU more than 15 ms.
>     *   Using non-OS SDK, please do not call any function defined with ICACHE_FLASH_ATTR in the
interrupt handler.

Alkorin:
Ok, so I discard my idea to let a "safe area" at the end of the stream. 10µs is the time to send 8 bits... :)

TerryE:
It's a guideline. Masking interrupts around an RGB set is less than 3x this, so probably won't cause problems and will  ensure that only complete RGB sets get shifted out. Allowing interrupts between sets and checking the clock delta around the enable - disable to detect for interrupt delivered and restart on interrupt detected is what I'd try.

matgoebl:
How about using hardware support on the ESP8266? There is a I2S-output WS2812(B) Driver available at https://github.com/cnlohr/esp8266ws2812i2s (however for the arduino sdk). Can we integrate that into nodemcu?
Another option is uart driven, see https://github.com/Makuna/NeoPixelBus/tree/UartDriven

I'm using the current ws2812 driver together with wifi and a tcp server. The ws2812 strip is updated in a timer function.
Only running the timed updates (at e.g. 30ms) works fine with 240 LEDs.
With concurrent tcp communication I get resets even with 24 LEDs. With just 1 LED it works fine.

Alkorin:
Hum, UartDriven is a good approach, I will take a look :)

TerryE:
IMO, UartDriven sounds like a horrible botch that could further compromise the SDK-imposed architecture. I think that you'd need compelling arguments to get support for this.

skycoders:
and what about the i2s variant? Does that sound more feasible? I failed with longer LED strips and WiFi as well...

TerryE:
> IMO, UartDriven sounds like a horrible botch that could further compromise the SDK-imposed architecture. I think that you'd need compelling arguments to get support for this.

I've had a look at this.  As far as I can see this is using some form of H/W serial to WS2812 encoding so that the ESP8266 on chip UART fifo can be used to latch out the **RGB** data bytes.  It could really do with an application note as to how to implement it, but I withdraw my "horrible botch" comment.  Sorry.

Alkorin:
I'll try to implement it.
I've to convert the arduino helpers into SDK/Register calls, will be fun :)

skycoders:
I offer extensive testing with the longest strip I could get (300 LEDs waiting for input). Once there is something to try, let me know. I have the docker image running to build own firmware.

Alkorin:
@skycoders have fun, it works on my 300 leds strip.
Timings measured with my scope: 
0 => ~320ns 
1 => ~944ns
Period => 1.25us

It's a first draft, I've to clean a little bit :D

Alkorin:
I've done some cleanup and will split my branch in two.

I'll make a new PR for the UART part

Alkorin:
@skycoders see PR #951

makefu:
@Alkorin i just wanted to give you a big Thanks for your contribution. I tried the merged code in nodemcu:dev yesterday and it worked very smoothly.

What i did was a fade from one buffer to another ( https://github.com/makefu/rgbcubes/blob/master/nodemcu-fader/standalone/fade.lua ) using algorthmic decay. With your addition the code became very clean, no need to stitch together long strings anymore. 

Keep on rocking!

PR #906, opened 2016-01-06
Remove left-over cruft regarding Lua mem opt.

jmattsson:
I must've overlooked this when we moved LUA_OPTIMIZE_MEMORY=2 into the app/Makefile.


Alkorin:
:thumbsup:

PR #908, opened 2016-01-06
Add support for Temperature Comparison register set/get functions

elgarbe:
Add support for Temperature Comparison register set/get functions


nickandrew:
Looks good to me.

jmattsson:
Thank you!

Issue #907, opened 2016-01-06
Strange tmr.alarm() operation


0kulist:
After hundreds of lines of code - something that drives me crazy. 
So, the story. Same problem on two devkits of type D1 mini:
1. Original NodeMCU 0.9.6 build 20150704  powered by Lua 5.1.4
2. NodeMCU custom build by frightanic.com, branch: master,	modules: node,file,gpio,wifi,net, pwm,tmr,uart, built on: 2016-01-05 16:31, powered by Lua 5.1.4 on SDK 1.4.0

I need alarm timer for periodic (up to 24h) events. So "interval", after tmr.c - uint32_t milliseconds.
Observed that example from api with a special interval fires alarm immediately - twice.

    NodeMCU 0.9.6 build 20150704  powered by Lua 5.1.4
    lua: cannot open init.lua
    > tmr.alarm(2, 11126000, 1, function() print("hello world") end )
    > hello world
    hello world

And 11126000 is only 185.43 minutes. How is that possible?

jmattsson:
I can confirm this behaviour. Something is thoroughly broken.

Not only does it fire right away, it doesn't re-fire as one would expect with a repeating timer. After testing with falling values for interval, around 6872000 I start seeing auto-repeats, but rather quick ones at that. It gets really weird once I hit 6871968, where I get autorepeat for a handful of prints, then it stops. The actual number of prints varies from run to run.
`tmr.alarm(2, 6871968, 1, function() print("hello world") end )`

`tmr.alarm(2, 6871955, 1, function() print("hello world") end )` on the other hand only gives me one print, consistently, quickly. Same all the way down to 6871949.

However, once I get to 6871948 it doesn't trigger early.

In all cases tmr.state() returns `true,1`.

Even though the interface (including in the SDK) is in ms, I believe internally the SDK uses us. Even in that scenario I'm not seeing any obvious wrap-arounds or overflows.

TerryE:
Yes but is this a bug in our firmware or in the SDK? I cant recall any special code in the firmware which could create this as we essentially wrap around the SDK timer functions. If this is an SDK bug, then should we be trying to code around it in our firmware. I think not; it would be time to claim a $1000 :)

jmattsson:
I'm poking at it... given the irregularity I'm thinking there's a good chance we're somehow clobbering the os_timer structure and thereby messing things up.

nickandrew:
Check the value of `tmr->interval` each time the timer fires?

[...]

PR #917, opened 2016-01-10
Switch to SDK downloading + replacement libmain.a for timer fix

jmattsson:
Just in case it is wanted, here's a branch for the replacement libmain.a to address the timer issue in #907. Given earlier discussion about the -dl branch, I shifted the SDK zip to be downloaded as well as the replacement libmain.a zip. I didn't want to add more binaries to the repo unnecessarily...

If the final call is to not use the replacement libmain, just close this PR :)


TerryE:
To download or not to download, that is the question. 

With a retry count of 20, and caching this should be pretty resilient. I haven't tried this but what happens if the 20th retry download is short rather than a total failure?  Won't we get into a latched failure where we have a short tarball in the cache? Should we do the wget if the tarball doesn't exist _or_ is too short? That way retrying a failed download should recover. This should address Marcel's resillience concerns. 

I know others have stronger views on this.

My only other comment is that if we agree on the dl approach then I would do the same for the esp-open-sdk.tar.gz tarball, but this is pretty low priority so doesn't need to be done now.

marcelstoer:
> Should we do the wget if the tarball doesn't exist *or* is too short?

+1

jmattsson:
Does wget leave short files in place? I can easily add a `|| { rm -f $@; exit 1; }` to guard against that. Automatically re-downloading would take a bit more make file logic that I'm not sure would be worth it vs just having to re-run make. I can have a play tonight if you want (subject to time availability).

TerryE:
Just thinking about this, the main reason for the brown outs on download is that we are downloading the SDKs from **bbs.espressif.com** through the CN firewalls. This would be a lot faster and resilient if we used a github dedicated user account to hold these zips.

marcelstoer:
Those are valid concerns but to me this discussion is getting a bit absurd. Why would you *not* want to use Git submodules but then go to great length to manually build something reliable and resilient that (in essence) is not too different from what Git would do?

jmattsson:
Because a sub module is a complete git repository, so you would end up with all the historical versions, for good and bad. And if we want that, we might as well just stick with keeping it all in this repo.

marcelstoer:
Riiight :flushed:.. and that's where the debate started, you want to reduce the amount of data pulled from where ever.
There's no way to tell Git to only clone revision X of a sub module, that's true. If we picked up Terry's suggestion of a dedicated GitHub repo we could still do `git clone --depth 1 --branch <SDK-version> <repo>` to reduce amount data to transfer. I suspect, but haven't tested, that using `git` over `wget`/`curl` would get us better stability as Git should ensure consistency of its operations.

jmattsson:
SDK 1.5.1 was [just released](http://bbs.espressif.com/viewtopic.php?f=46&t=1609) but it doesn't appear to have the timer fix. Either that or they copy/pasted the wrong fix description. I'll follow up with Espressif; I'd say hold off merging this for now at least.

jmattsson:
Gah, I forgot to post the follow up earlier. Alright, so the timer fix isn't in 1.5.1, which pretty much gives us three options:
  1. Use this PR and fix a reported issue in NodeMCU right away.
  1. Upgrade to 1.5.1 and get those improvements but leave the reported issue outstanding.
  1. Refuse to pick one of those and instead wait for SDK 1.5.2 or 1.6.0.

All have merit. Suggestions?

TerryE:
IMO, In the great scheme of things, it makes little difference whether the maximum period is 30 odd minutes, 60 odd minutes or somewhere in between.  The overall epoch is the same and it's not days, so if you are waiting on ToD then you are going to have to use a timer loop and it makes bugger all difference in terms of power consumption or coding complexity whether you wake up every 30 mins or 60 mins to count down.  

I suggest that you either toss a coin or do the easiest for you.  We've got far higher priority stuff to worry about.

PR #923, opened 2016-01-12
Ugly workaround for broken SPIFFS GC alignment.

### LINKED ISSUE #915: multiple file.write cmd break spiffs
I'm using the latest master build firmware.

When I'm writing a file to esp8266, after a while it would rst and no longer able to write or delete files.  I would need to issue file.format() cmd to fix the spiff system.  Basically, with trail and error, I think the cause is when the previous file write cmd didn't finish before the next file write cmd.  Is there a why we can add a hook to check when the file.write cmd is done?

Test:
```Lua
-- Loop this
file.open("testfile","w+")
file.write(str1)
file.write(str2)
file.write(str3)
file.write(str4)
file.write(str5)
-- etc.
```
After a few times the above will reset and won't be able to create new files or delete old ones.

If I added some delays between file.write, the problem occurs less frequent.

jmattsson:
This fixes #915, for now.

### DIFF
--- a/app/spiffs/spiffs_gc.c
+++ b/app/spiffs/spiffs_gc.c
@@ -250,6 +250,10 @@ s32_t spiffs_gc_find_candidate(
 
   // divide up work area into block indices and scores
   // todo alignment?
+  // YES DO PROPER ALIGNMENT !^@#%!@%!
+  if (max_candidates & 1)
+    ++max_candidates; // HACK WORKAROUND ICK for sizeof(spiffs_block_idx)==2
+
   spiffs_block_ix *cand_blocks = (spiffs_block_ix *)fs->work;
   s32_t *cand_scores = (s32_t *)(fs->work + max_candidates * sizeof(spiffs_block_ix));
 


Alkorin:
Alignment patch from upstream:

```
diff --git a/app/spiffs/spiffs_gc.c b/app/spiffs/spiffs_gc.c
index a86f812..a0837c3 100644
--- a/app/spiffs/spiffs_gc.c
+++ b/app/spiffs/spiffs_gc.c
@@ -36,7 +36,7 @@ s32_t spiffs_gc_quick(
   int cur_entry = 0;
   spiffs_obj_id *obj_lu_buf = (spiffs_obj_id *)fs->lu_work;
 
-  SPIFFS_GC_DBG("gc_quick: running\n", cur_block);
+  SPIFFS_GC_DBG("gc_quick: running\n");
 #if SPIFFS_GC_STATS
   fs->stats_gc_runs++;
 #endif
@@ -249,10 +249,15 @@ s32_t spiffs_gc_find_candidate(
   c_memset(fs->work, 0xff, SPIFFS_CFG_LOG_PAGE_SZ(fs));
 
   // divide up work area into block indices and scores
-  // todo alignment?
   spiffs_block_ix *cand_blocks = (spiffs_block_ix *)fs->work;
   s32_t *cand_scores = (s32_t *)(fs->work + max_candidates * sizeof(spiffs_block_ix));
 
+   // align cand_scores on s32_t boundary
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpointer-to-int-cast"
+  cand_scores = (s32_t*)(((ptrdiff_t)cand_scores + sizeof(ptrdiff_t) - 1) & ~(sizeof(ptrdiff_t) - 1));
+#pragma GCC diagnostic pop
+
```

jmattsson:
Let's see what other gremlins hide in 0.3.3 before we start merging fixes then...

And he *really* means `intptr_t`, not `ptrdiff_t`.

TerryE:
Sorry Johny, I thought that I'd merged this. Apologies for not doing so.

Issue #848, opened 2015-12-13
Add node.rst_info()


nickandrew:
This *doesn't work* because it always returns reason code 6 "external system reset", but it would be good if it can be fixed.

I noticed the SDK implements system_get_rst_info() which I gather returns the same information as the message which sometimes appears after an exception:

```
Fatal exception 28(LoadProhibitedCause):
epc1=0x402566d5, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00004655, depc=0x00000000
```

It would be good to capture this and make it available to lua code. So I implemented it in my branch here: https://github.com/nickandrew/nodemcu-firmware/commit/dccd9ca4cd6bfb3aa578610c352502eab4fc3966

But it doesn't work. Reading comments here http://bbs.espressif.com/viewtopic.php?t=1230 and here https://github.com/esp8266/Arduino/issues/1098 indicates it is fixed in SDK 1.4.0 or fixed in 1.4.0+patch or fixed in 1.5.0

I haven't tried the patch yet but perhaps we can leave this issue open for testing pending 1.5.0 SDK support (add a milestone?).

devsaurus:
Interestingly, this function seems to provide more information than `rtc_get_reset_reason()` which is currently [exposed as node.bootreason()](https://github.com/nodemcu/nodemcu-firmware/blob/7512e628b6b39a8ab4f85e73c13ca216a1734eaa/app/modules/node.c#L474). Both should be merged / unified once `system_get_rst_info()` can be considered working.

TerryE:
@nickandrew Nick I think that your patch is making this implementation overly complicated.  

*    The ROM / reboot code dumps a 7 word info block into the RTC RAM.
*    The SDK copies this into a 7 word RAM structure during its startup and makes the address of this available through the `system_get_rst_info()` API call.  There is no point in the firmware taking a second RAM copy.
*   As Arnim says we already have a node. bootreason() call, so why not simply extend this to return the extended list.  This has the advantage of reusing the exting firmware API call in an upwardly compatible way.
*    I dislike system calls dumping info to the uart as a side effect.  If they print then that should be the function and the call should be `node.printbootreason()` or whatever, but what's the point of this if you can simply do:

``Lua
  print( ('reset reason: %x exception: %d epc1=0x%08x, epc2=0x%08x, epc3=0x%08x, excvaddr=0x%08x, depc=0x%08x'
       ):format(node.bootreason()) )
```

[...]

PR #924, opened 2016-01-12
Expose extended reset info via node.bootreason().

jmattsson:
I opted against providing constants for all the various reset reasons, as this is a bit of a niche function and the space overhead did not seem appropriate. I'll update `newdocs` if/when merged.

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -472,8 +472,20 @@ static int node_setcpufreq(lua_State* L)
 // Lua: code = bootreason()
 static int node_bootreason (lua_State *L)
 {
+  struct rst_info *ri = system_get_rst_info ();
   lua_pushnumber (L, rtc_get_reset_reason ());
-  return 1;
+  lua_pushnumber (L, ri->reason);
+  if (ri->reason == REASON_EXCEPTION_RST)
+  {
+    lua_pushnumber (L, ri->epc1);
+    lua_pushnumber (L, ri->epc2);
+    lua_pushnumber (L, ri->epc3);
+    lua_pushnumber (L, ri->excvaddr);
+    lua_pushnumber (L, ri->depc);
+    return 7;
+  }
+  else
+    return 2;
 }
 
 // Lua: restore()


TerryE:
Johny, I know that I didn't get around to doing my PR first but was there any reason for not using [my suggested version](nodemcu/nodemcu-firmware/issues/848#issuecomment-164295405) on #848?  It just that this approach takes up less firmware memory.

jmattsson:
Yes, three reasons:
  - I hadn't seen that.
  - I already had this one as a local patch in my git stash.
  - Your version makes dangerous assumptions about the format of that structure. Willy-nilly hard-casting from structs to non-char pointers is the sort of thing which is causing us grief in SPIFFS at the moment. The C standard is very clear on this topic these days - type punning is only permissible through a union or a char pointer. Anything else is undefined (but just happens to work often enough that people keep using that approach).

TerryE:
Woops, put the knife back in its sheaf! :laughing:  I observed that the rst_info is a structure comprising only uint32 fields (which it is), so I am casting a contiguous set of uint32s onto a uint32[]. I am not casting across types.  

By binding all of the fields to a `Lua_number` parameter, you assume that they are implicitly castable to a double (or a long in the case on an integer build).  

Maybe I am getting too old, but why is one heresy and the other whiter than white?  I am certainly too old to have a falling out over this.

jmattsson:
Sorry if I came off a bit strong - I'd just spent a couple of hours digging up the SPIFFS alignment stuff so nerves were a bit raw. Here are a couple of links which hopefully are less acidic:

http://www.cocoawithlove.com/2008/04/using-pointers-to-recast-in-c-is-bad.html

http://blog.qt.io/blog/2011/06/10/type-punning-and-strict-aliasing/

I'll say one other thing though - my approach will break noisily if/when Espressif catch on to epc2 & epc3 not actually being used and they remove them from the struct ;)

TerryE:
Yup, except that it's not really punning is it?  _I saw a dog; it was really a dog -- HaHa, did you get the pun?_ 

`rst_info` is a structure which for architectural reasons is a vector of 32 bit values -- which Espressif decided to name the fields separately as uint32s rather than by positional index.  Punning is where you overload the types e.g. casting `uint16 x, y` collectively as *((uint32 *) &x) , but casting a uint32 x onto *((uint32 *) &x) is perfectly OK, IMO.  

Yes if they decided to a char field in the middle then we'd be punning, and we need to change this, but if they added or dropped uint32 fields, then yours would need to be changed, but my version would still be OK.

We don't live a black and white world.  Just lots of colours which are ordered on different grey scales depending on would spectrum you use to illuminate them.  Anyway the fun here is in the debate, not the outcome. :wink:

jmattsson:
That was a dog of a joke, Terry... ;)

(But I'm thankful we don't have to discuss ducks at least!)

You're still violating the aliasing rules by casting to a uint32. The types here are `uint32` and `struct rst_info`. Yes, they are (currently) compatible in a machine sense, but not for aliasing purposes (see standard $6.5 7). Moreover, and given our current compilation flags, more relevant, you're effectively discarding information and preventing the compiler from sanity checking. Member changes to the rst_info struct won't get bitched about by the compiler this way. Silent errors are always Very Bad(tm) in my books. I want things to blow up at compile time, or if that's not possible, at link time. Never at runtime. A compiler is a wonderful type error checking tool, and I miss it whenever I'm in a dynamic type environment.

All that being argued, I did look at the disassembly and I can see why you're unhappy. Heck, *I'm* unhappy with what the compiler generated! Can I suggest a middle ground improvement?

```
  struct rst_info *ri = system_get_rst_info ();
  lua_pushnumber (L, rtc_get_reset_reason ());
  lua_pushnumber (L, ri->reason);
  if (ri->reason == REASON_EXCEPTION_RST)
  {
    uint32 arr[5] = { ri->epc1, ri->epc2, ri->epc3, ri->excvaddr, ri->depc };
    int i;
    for (i = 0; i < 5; ++i)
      lua_pushnumber (L, arr[i]);
    return 7;
  }
  else
    return 2;
```
With the above body the resulting code is only 23 bytes larger than your version, while still maintaining the benefits of using named members and clean typing. If you prefer, we could also change `lua_pushnumber()` to `lua_pushunsigned()` for additional clarity.

And yeah, if I wanted real black & white, I'd become a mathematician!

TerryE:
Or even (does the ROM reason add anything to the SDK one?):
```C
struct rst_info *ri = system_get_rst_info ();
uint32 arr[6] = { ri->reason,  ri->epc1, ri->epc2, ri->epc3, ri->excvaddr, ri->depc };
int i,n = ((ri->reason == REASON_EXCEPTION_RST) ? 1 : 6;
for (i = 0; i < n; ++i) lua_pushunsigned (L, arr[i]);
return n;
```
:eyes:

jmattsson:
"Just" backwards compatibility.

And sure, I could get behind going all 6 rather than 5. It's a bit less clear, but still fully readable.

jmattsson:
How many bytes did that take off, btw?

TerryE:
Haven't checked yet.  Too busy answering Qs !! And now bedtime is calling :laughing:

Issue #901, opened 2016-01-04
Crash in lmacProcessAckTimeout


pjsg:
This is just a placeholder ticket -- the bug is not in the nodemcu code, but (I think) in the espressif SDK. I have opened a bug with them, although the bug is not very reproducible.

There are other people (try https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=lmacprocessacktimeout), with non-nodemcu codebases, and with older versions of the SDK that have seen the same crash -- lmacProcessAckTimeout+39 (decimal).

It causes my app to crash roughly once every 4 to 8 hours. It appears to be position dependent -- if I move the board from one location to another, the failure rate goes up or down. The AP is positioned a long way away, so the signal is weak to begin with.

If you have unexplained crashes, then use the improved node.bootreason() from issue #900 to get the failing PC. Then you can look in your mapfile to find the failing function. 

Just to repeat, this is *probably* not a bug in the nodemcu code, but it is a bug that affects the users of nodemcu.

pjsg:
Argh! They rejected the bug report with

```
But please report your issue to nodemcu instead, here we accept the bug report based on our original SDK. 
You can report your issue to nodemcu, if they find that is not due to their codes, they will come to us and help us duplicate the issue.
```

nickandrew:
@pjsg "they will come to us" :smile: I think "they" (you) already went to "us" (Espressif). Perhaps Terry needs to give the bug report his stamp of approval.

pjsg:
I really want to be able to reproduce it easily. The same bug happens with the Arduino / ESP8266 code. The person there has now reported the bug: https://github.com/esp8266/Arduino/issues/1329

pjsg:
I just got email from espressif:

Hi,

Please have a try with our latest ESP8266_NONOS_SDK_V1.5.1 http://bbs.espressif.com/viewtopic.php?f=46&p=5315 .

If your problem is still unsolved, please feel free to let us know.

I'm going to try and build with this version and see if the problem has gone away......

pjsg:
Yes, the problem seems to have been resolved. My code has been running for 12 hours with no crash.

pjsg:
The upgrade solved it.

PR #928, opened 2016-01-13
Upgrade to SDK version 1.5.1 from Espressif

pjsg:
This PR just replaces the SDK .zip with the newest version from Espressif. It also updates the version numbers in various places.

This solves issue #901.

Verified by running previously failing code for 12 hours without failure. Previously would crash within 2 hours.

Only non-obvious change is to alter the '.extracted' file to mark which version of the SDK was extracted. This allows upgrading the SDK.


jmattsson:
Nice, thanks for all the testing with #901!

You missed changing `NODE_VERSION_REVISION` to a 1 in user_version.h here though ;)

@TerryE @marcelstoer Seeing as #901 seems a bit more important than the timer issue in #907, I'd say we should probably just close the #917 PR. The remaining question then is whether we keep piling in zip files, or switch to downloading & caching them.

TerryE:
My +1 is to download and cache them. Also +1 on #901 as there is a simple workaround for #907 and this code is going to be needed in the general case anyway

jmattsson:
Alright, @pjsg would you mind updating this PR to do the download like #917? Just ignore the binpatch/libmain* parts in the makefile. Or throw it over the fence to me and I'll look at it tomorrow, that's fine too. Thanks!

pjsg:
This now does the download. *HOWEVER* this means that when you run this for the first time with a new SDK, you need network access. This means that it does not work if you are running one of the standard docker images -- which appear to have been setup without network access. You can just run 'make' outside the docker image, and then you have to do the rest of the build inside. I'm not sure what the right approach is here.

I did change the size check to use wc rather than parse the output of ls -- this is more portable.

TerryE:
> Wouldn't a MD5 or SHA1 verification be more reliable at detecting errors in the download vs a size check?

Yes MD5 or SHA1 is sufficient, but the failure mode that we are trying to detect is a TCP link failure.  All that is necessary in this instance is a length check; neither MD5 nor SHA1 is necessary.

If we want end-to-end link integrity then we should just switch _all_ of our http connections to https, or use a MD5 / SHA1 signature on all http transferred data. IMO, there's little point in assuring just one file.

TerryE:
> This that it does not work if you are running one of the standard docker images -- which appear to have been setup without network access.

Hummmnn, perhaps you should swap from docker to a standard Ubuntu LTS 12.04-5 minimal server image as I describe in this [esp8266 wiki page](http://www.esp8266.com/wiki/doku.php?id=toolchain#how_to_setup_a_vm_to_host_your_toolchain).  I am a VirtualBox man; @marcelstoer is the docker expert, so he will need to comment on this.

pjsg:
I converted to using a sha1 sum -- it makes the tests slightly simpler....

TerryE:
@pjsg  As I said sufficient, but not necessary :smile:  Did you note my cross post on ufw?

pjsg:
Actually, it turns out to be a local configuration problem on my system.... So I am much happier now.

marcelstoer:
> it does not work if you are running one of the standard docker images -- which appear to have been setup without network access.

There's no such thing as a "standard Docker image", they're all different. The image creator decides what features it contains and how they're configured.
[My image for example](https://github.com/marcelstoer/docker-nodemcu-build/blob/master/Dockerfile) extends the [Ubuntu image](https://hub.docker.com/_/ubuntu/) and runs `apt-get` commands. It definitely does have network access - if the underlying OS or VM does.

TerryE:
Hey, Marcel, that's just nitpicking, IMO. you have to use LTS 12.04-5 so the toolchain tarball is compatible with Travis CI.  I agree that you've extended it.  Good on you :)

jmattsson:
Looks good to me, thanks Philip!

PR #934, opened 2016-01-14
Completing DS3231 driver

Takkuz:
I've written some functions I need to use DS3231 RTC on NodeMCU.
I think the cover all of DS3231 capabilities:

- setting an alarm
- enable/disable an alarm
- reaload a triggered alarm
- watch if time has been compromised
- get CTRL and STATUS bytes from the board that could be read according to page 11 and follow of the [datasheet](http://datasheets.maximintegrated.com/en/ds/DS3231.pdf)

I've also rewrite docs to explain those methods.

### DIFF
--- a/lua_modules/ds3231/ds3231.EN.md
+++ b/lua_modules/ds3231/ds3231.EN.md
@@ -39,10 +39,11 @@ package.loaded["ds3231"]=nil
 <a id="ds3231_settime"></a>
 ## setTime()
 ####Description
-Sets the current date and time.
+Sets the current date and time.  
+if _disableOscillator_ is set to 1 the oscillator will **stops** on battery.
 
 ####Syntax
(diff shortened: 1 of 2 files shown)


marcelstoer:
Thanks and kudos for writing documentation. A few issues:
- PRs must be created against `dev`, see https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md
- squash your commits, again see CONTRIBUTING.md
- please use the new Markdown syntax/template for the documentation, see https://github.com/nodemcu/nodemcu-firmware/blob/newdocs/docs/en/modules/node.md for an example

So, pls fix and resubmit.

Takkuz:
I apologize. I red the guideline only after made the request. I'm really sorry. I'll make it compliant, thank you.

PR #935, opened 2016-01-14
Add the exccause parameter into the bootreason block

pjsg:
It turns out that the exccause parameter was missing from the returned set of values from node.bootreason(). This adds that parameter in. [This is the parameter that tells you which type of exception it was. There is a good table in the xtensa documentation]

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -477,12 +477,13 @@ static int node_bootreason (lua_State *L)
   lua_pushnumber (L, ri->reason);
   if (ri->reason == REASON_EXCEPTION_RST)
   {
+    lua_pushnumber (L, ri->exccause);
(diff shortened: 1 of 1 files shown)


TerryE:
Philip, good catch. We also need to update the node documentation.

jmattsson:
After all our debate over that function @TerryE we both overlooked that - wow :laughing:

jmattsson:
I've updated `newdocs` to list the exception cause value.

TerryE:
> After all our debate over that function @TerryE we both overlooked that 

But my original version wouldn't have had this bug :wink:

jmattsson:
>But my original version wouldn't have had this bug :wink:

Sure, but now we have the best of both worlds!

PR #938, opened 2016-01-15
Optimized node.bootreason() for size.

jmattsson:
Reduces the size from 179 to 102 bytes.


jmattsson:
Here you go, @TerryE :)

TerryE:
Thanks sir.

PR #942, opened 2016-01-16
Add docs for coap module.

funshine:
Add docs for coap module.
Note that the coap module is not complete, it's only in a early stage.


funshine:
shoot, seems make 2 commits not into 1 but 4...I will redo...

PR #944, opened 2016-01-16
Dev - Completing DS3231 driver

Takkuz:
I've written some functions I need to use DS3231 RTC on NodeMCU.
I think the cover all of DS3231 capabilities:

- setting an alarm
- enable/disable an alarm
- reaload a triggered alarm
- watch if time has been compromised
- get CTRL and STATUS bytes from the board that could be read according to page 11 and follow of the [datasheet](http://datasheets.maximintegrated.com/en/ds/DS3231.pdf)

I've also rewrite docs to explain those methods.


marcelstoer:
As with #934 you again picked the wrong branch. PRs go into `dev` and not `master`. If a PR has lots of commits and a long list of files there's usually something wrong. That's something you can verify before you create it.

Takkuz:
Damn... I apologize again. I delete everything and start again following the tutorial. I don't really know why it doesn't work. I'll try again keeping more attention. Sorry.

marcelstoer:
I'm happy to hear your ideas for improvements if you feel CONTRIBUTING.md doesn't explain things well enough. I know first hand that it might take a well until it all feels natural as I still struggle with Git every now and then.

PR #945, opened 2016-01-16
Fix unit constants to be consistent with code

SebastianSchildt:
The unit is decided by 'C', 'F' 'K', therefore the numeric constants broke all example code like

```
print(t.readNumber(nil,t.F))
```

This patch fixes that. Can be applied to master and dev


marcelstoer:
PRs go into `dev` not `master`and commits should be squashed. See https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#working-with-git-and-github

Please resubmit.

Issue #903, opened 2016-01-05
How safe is masking all interrupts.


TerryE:
We have our bit banging code splattered with `ets_intr_lock()`, `ets_intr_unlock()` sequences, where
-  the locked code can run to 100s uSec or mSec in some cases.
-  the code is in Flash so we could get a cache miss interrupt during a locked code sequence.
-  the code might trigger an unaligned data fetch and the HW will need to dispatch an unaligned data fetch exception to execute the code.

Does anyone know where I can get a detailed description of the rules here? Should we be moving these routines into irom0? @jmattsson, @nodemcu, anyone any comments?

jmattsson:
I believe the document you're looking for is the [Xtensa Instruction Set Architecture](http://0x04.net/~mwk/doc/xtensa.pdf).

Based on the testing I did it is possible to hit the unaligned-data-access exception inside an ets_intr_lock() block and have it do its thing fine. Of course, there is a timing penalty, as is there for running code from the flash.

Moving things into iram0 is problematic, as the SDK has gobbled up most of it by now. When I did the 1.4.0 upgrade I used a good portion of my bag'o'tricks to get the NodeMCU bits to fit alongside the SDK bits.

TerryE:
The amount of code that truly _needs_ to do interrupt locking is a few 100 bytes. In a few cases we have locking just for the hell of it, such as in the GPIO module where where we locks around a call to a ROM routine which when you look at the code already has locks around the critical region. IMO, any PRs which include this type of code should be rejected unless the author includes a proper impact assessment and justification -- that is if we care about getting the firmware stable.

jmattsson:
There should be very few legitimate causes for disabling interrupts. It sounds like it's largely bitbanging which is doing this, so could we get away with a common big-banging routine that takes a list of ops in RAM to bang out, and then have that common routine in iram? I'm thinking along the lines of the Linux SPI driver interface.

TerryE:
Now that sounds like an approach that I'd definitely +1. A single unified approach within the core for a common mode issue that a number of libraries face.

nodemcu:
+1 too, a function/api that bit bang out with a limited delay(baudrate), in iram.
a little bit late, happy new year!

TerryE:
[...]

PR #951, opened 2016-01-18
Rewrite ws2812 serialization code to use UART1 instead of bit-banging.

Alkorin:
It allows keeping interrupts enabled but force to use GPIO2.

It's a first draft, some work still need to be done because it brokes compatibility.
What's the best solution ?
 - Depreciate the old module and create a new one with new function signatures because bit-banging is just a bad idea and we should stop allowing people using it ?
 - Offering the choice between bit-banging-which-breaks-wifi and uart-but-only-gpio2 ?
 - if(pin == gpio2) { uart } else { bitbanging } with a notice which says that using another pin than GPIO2 is at your own risk ?

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -4,54 +4,46 @@
 #include "c_stdlib.h"
 #include "c_string.h"
 #include "user_interface.h"
+#include "driver/uart.h"
 
-static inline uint32_t _getCycleCount(void) {
-  uint32_t cycles;
-  __asm__ __volatile__("rsr %0,ccount":"=a" (cycles));
-  return cycles;
-}
-
-// This algorithm reads the cpu clock cycles to calculate the correct
-// pulse widths. It works in both 80 and 160 MHz mode.
-// The values for t0h, t1h, ttot have been tweaked and it doesn't get faster than this.
-// The datasheet is confusing and one might think that a shorter pulse time can be achieved.
-// The period has to be at least 1.25us, even if the datasheet says:
-//   T0H: 0.35 (+- 0.15) + T0L: 0.8 (+- 0.15), which is 0.85<->1.45 us.
-//   T1H: 0.70 (+- 0.15) + T1L: 0.6 (+- 0.15), which is 1.00<->1.60 us.
-// Anything lower than 1.25us will glitch in the long run.
+// Stream data using UART1 routed to GPIO2
+// NODE_DEBUG should not be activated because it also uses UART1
 static void ICACHE_RAM_ATTR ws2812_write(uint8_t pin, uint8_t *pixels, uint32_t length) {
-  uint8_t *p, *end, pixel, mask;
-  uint32_t t, t0h, t1h, ttot, c, start_time, pin_mask;
 
-  pin_mask = 1 << pin;
-  p =  pixels;
-  end =  p + length;
-  pixel = *p++;
-  mask = 0x80;
-  start_time = 0;
-  t0h  = (1000 * system_get_cpu_freq()) / 3022;  // 0.35us (spec=0.35 +- 0.15)
-  t1h  = (1000 * system_get_cpu_freq()) / 1477;  // 0.70us (spec=0.70 +- 0.15)
-  ttot = (1000 * system_get_cpu_freq()) /  800;  // 1.25us (MUST be >= 1.25)
+  // Data are sent LSB first, with a start bit at 0, an end bit at 1 and all inverted
+  // 0b00110111 => 110111 => [0]111011[1] => 10001000 => 00
+  // 0b00000111 => 000111 => [0]111000[1] => 10001110 => 01
+  // 0b00110100 => 110100 => [0]001011[1] => 11101000 => 10
+  // 0b00000100 => 000100 => [0]001000[1] => 11101110 => 11
+  uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
 
-  while (true) {
-    if (pixel & mask) {
-        t = t1h;
-    } else {
-        t = t0h;
-    }
-    while (((c = _getCycleCount()) - start_time) < ttot); // Wait for the previous bit to finish
-    GPIO_REG_WRITE(GPIO_OUT_W1TS_ADDRESS, pin_mask);      // Set pin high
-    start_time = c;                                       // Save the start time
-    while (((c = _getCycleCount()) - start_time) < t);    // Wait for high time to finish
-    GPIO_REG_WRITE(GPIO_OUT_W1TC_ADDRESS, pin_mask);      // Set pin low
-    if (!(mask >>= 1)) {                                  // Next bit/byte
-      if (p >= end) {
-        break;
-      }
-      pixel= *p++;
-      mask = 0x80;
-    }
-  }
+  // Configure UART1
+  // Set baudrate of UART1 to 3200000
+  WRITE_PERI_REG(UART_CLKDIV(1), UART_CLK_FREQ / 3200000);
+  // Set UART Configuration No parity / 6 DataBits / 1 StopBits / Invert TX
+  WRITE_PERI_REG(UART_CONF0(1), UART_TXD_INV | (1 << UART_STOP_BIT_NUM_S) | (1 << UART_BIT_NUM_S));
+
+  // Redirect UART1 to GPIO2
+  // Disable GPIO2
+  GPIO_REG_WRITE(GPIO_ENABLE_W1TC_ADDRESS, BIT2);
+  // Enable Function 2 for GPIO2 (U1TXD)
+  PIN_FUNC_SELECT(PERIPHS_IO_MUX_GPIO2_U, FUNC_U1TXD_BK);
+
+  uint8_t *end = pixels + length;
+  do {
+    uint8_t value = *pixels++;
+
+    // Wait enough space in the FIFO buffer
+    // (Less than 124 bytes in the buffer)
+    while (((READ_PERI_REG(UART_STATUS(1)) >> UART_TXFIFO_CNT_S) & UART_TXFIFO_CNT) > 124);
+
+    // Fill the buffer
+    WRITE_PERI_REG(UART_FIFO(1), _uartData[(value >> 6) & 3]);
+    WRITE_PERI_REG(UART_FIFO(1), _uartData[(value >> 4) & 3]);
+    WRITE_PERI_REG(UART_FIFO(1), _uartData[(value >> 2) & 3]);
+    WRITE_PERI_REG(UART_FIFO(1), _uartData[(value >> 0) & 3]);
+
+  } while(pixels < end);
 }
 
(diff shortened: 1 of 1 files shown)


TerryE:
@Alkorin Tomas, this really is a completely new approach.  So my suggestion is not to try to PR too quickly.  Find another  WS2812 developer to buddy with and act as independent tester and hammer it so that you both confident.  

Also out of interest, why are you reallocating a temporary buffer to process the GRB -> RGB triples. why not just pass an RGB flag into  `ws2812_write()`? Something like 
```C
  int i, j,  o[3];
  o[0] = rgb_flag ? 1: 0; o[1] = 1- o[0]; o[2] = 2;

  for (i = 0; i < length; i += 3) {
    for (j = 0 ; j <2; j++) {
      uint8_t value = pixels[i+o[j]];
      // . . . Wait for 4 free bytes and write out bit doubles
    }
  }
```
Also why wait at half full threshold, rather than 252? @Alkorin Tomas, this really is a completely new approach.  So my suggestion is not to try to PR too quickly.  Find another  WS2812 developer to buddy with and act as independent tester and hammer it so that you both confident.  

Also out of interest, why are you reallocating a temporary buffer to process the GRB -> RGB triples. why not just pass an RGB flag into  `ws2812_write()`? Something like 
```C
  int i, j,  o[3];
  o[0] = rgb_flag ? 1: 0; o[1] = 1- o[0]; o[2] = 2;

  for (i = 0; i < length; i += 3) {
    for (j = 0 ; j <2; j++) {
      uint8_t value = pixels[i+o[j]];
      // . . . Wait for 4 free bytes and write out 4 x bit doubles
    }
  }
```
Also why wait at half full threshold, rather than 252? Also use UART1 rather than 1.  And this is one point where you could bung in a delay between the tests of the  fifo count.

Alkorin:
Yes it need to be tested, it's running here for more than 1h with with my scope attached to see any failures.

matgoebl:
@Alkorin Thanks a lot for your quick PR.
At a first try I was able to drive my ws2812 strip.
I will do more testing in a few days: re-run the tcp tests, that crashed before.

kbeckmann:
Always interesting with a new approach! I don't mean to sound negative but what's the benefit using UART? Handling interrupts while pushing pixels? I see the CPU usage is still high.
Using I2S you can push pixels using DMA and offload the cpu, someone did that here https://github.com/cnlohr/esp8266ws2812i2s but it messes with the tx or rx pin (don't remember which one right now) so debugging becomes messy.

I would vote for two solutions, one which can handle any gpio but no interrups, and another solution that prioritizes performance but is limited to a specific pin. This could also be a build flag if flash size is critical.

kbeckmann:
@Alkorin A bit off topic for this PR maybe since you're trying to move away from bit banging. But if you are concerned about the time it takes to push pixel data to the ledsrip, you could use two output pins simultaneously. I experimented with it and got it to work - resulted in double frame rate. http://kbeckmann.github.io/2015/07/25/Bit-banging-two-pins-simultaneously-on-ESP8266/

Alkorin:
I2S uses U0RXD. So if you want to use your strip, you'll loose access to Lua's interpreter via serial :/ (and will need a little more deep patch ...)
UART uses U1TXD which is accessible via GPIO2.

The benefit of using UART instead of bit-banging is to let the wifi stack alive ! According to Espressif's documentation we should not disable interrupts more than 10µs. 10µs is the time used to serialize ... 8 bits, a third led... So the library will never be compliant with Espressif's recommendations :/

We could let the bit-banging version but with a notice which says "use at your own risks - wifi stack will die if you use it for more than 0 led... :)". If you don't need wifi, why not ;)

Alkorin:
@kbeckmann I already saw your work to use two pins, here I tried to address #903. We have a more robust library compatible with Espressif's recommendations. The drawback is to have a specific pin :/ (on an esp-01... :relaxed:)

skycoders:
Hi, first tests on my end look good as well, I will continue testing extensively. My build currently does not connect to WiFi but that might be a different story.
Over all, I'd prefer a single pin solution as well if you look at it from a usability point of view. It is much easier for people to connect their strip instead of trying to synchronize two strips in a way that keeps animations etc. working. My plan was to provide a simple http based control of nice WS2812 features on typical strips you can get at the dealer of your choice. I expect that if we get this working, the WS2812 library of NodeMCU will get much more useful for anyone. Nevertheless, leaving the PIN choice to people by maybe also keeping the bit-banging implementation active sounds reasonable as well. If we provide the new implementation as its own module, that could easily be integrated through the online NodeMCU firmware builder if people need it. Just as a guess...

kbeckmann:
Oh, thanks for the explanaition @Alkorin. This probably explains the issues I've had when pushing pixel data coming from wifi for longer time (hours). It works surprisingly well considering the 10us recommendation, hehe.

If this patch makes the ESP more stable, then by all means, go for it! I'll also try the code and let you know how it goes.

skycoders:
I have now set up my 300 LEDs strip, together with WiFi. I have this script https://github.com/geekscape/nodemcu_esp8266/blob/master/examples/ws2812.lua running at an update rate of 20ms and it runs great. Even at only 5ms timer interval before each update it runs. This never worked with the old driver. I have to stop for now, but tests will continue tomorrow evening. Thanks @Alkorin for implementing this so quickly!

TerryE:
All (probably less Tomas who gets this) the wider architectural issue arises because the ESP8266 is a single core XT06, that also needs to run the Wifi stack.  Yes the chip has a Wifi controller on the die, but this is pretty low level and needs a general purpose CPU to handle interrupts and Wifi / IP / TCP housekeeping tasks.  If you mask off interrupts so that you bit bang out to the WS2812, then the Wifi stack dies, and recovery isn't straight forward.

The UART FIFO is latched out under H/W control and a 128 byte FIFO gives you a H/W buffer of 32×9µS = 288 µSec which mean that you don't need to mask interrupts and thus the network ISRs can be served without causing a ws2812 reset.

The only other option as I see it would be to mass interrupts; accept that the wifi is going to die during the ws2812 output and restart the CPU when done to recover the Wifi (storing any context in the RTC memory), but this restart would mean that you'd get a pretty crappy frame rate!

matgoebl:
Fantastic results - very big thanx to @Alkorin :thumbsup:
I've been running a load test for over 26.9 hours now: 300 LEDs updated at about 69.4fps,  with 24k tcp connections in between. Im on the dev branch with this ws2812.c by @Alkorin.
Using the default ws2812.c I got an reboot after a few seconds.

My test setup:
In the ESP side I have a [telnethttpd.lua](https://github.com/matgoebl/luatool/blob/master/luatool/init_wifi_telnethttpd.lua) running and in addition I run the command:
`counter=0 function count() counter=counter+1 ws2812.writergb(1,string.char(counter%256):rep(300*3)) tmr.alarm(6,1,0,count) end count()`
On the unix side I run:
`while sleep 1;do while sleep 1;do echo =counter|nc -w1 $ESPIP 8266;echo;done`

I'd stronly suggest to provide this improved driver to our users.

TerryE:
Given the enthusiastic feedback, I don't think that we should delay moving this into dev further. However we do need to do usual checks on backwards compatability, documentation, etc. @Alkorin  Thomas, can I ask you to do this?

marcelstoer:
Thomas, thanks a lot for all you efforts.

> I don't think that we should delay moving this into dev further. However we do need to do usual checks on backwards compatability, documentation, etc.

For what it's worth I disagree with this procedure. The whole point of moving the docs back to the repository was to ensure consistency with the code. Hence, if PRs change API or behavior they should include doc updates before we merge them. Authors shouldn't think about backwards compatibility *after* code was merged but before. Right now we're no better than before because the documentation says X but the code does Y.
Sooner or later someone will report a "bug" claiming that `ws2812.writegrb(pin, string)` didn't work anymore. Not what we want; certainly not what I want.

Alkorin:
I agree @marcelstoer!
That's why I opened #966 to discuss/fix this asap... !

skycoders:
@Alkorin I saw that you are also working on an extended version of the ws2812 module which allows to send tables instead of strings. Maybe in the course of making the ws2812 module better, this could be included as well? I realized that building a 300 LEDs string by appending characters to a string is not working well, having a table and then create the string by concat works, but wastes memory. As the charming part of these strips is individual addressing, a table would be a great fit from my point of view. I'd love to see it :)

TerryE:
@marcelstoer, I agree that this should be an absolute gate for hoisting from `dev` to `master`, but this thread is extremely active and the main WS2812 users are now actively participating.  Why not just revert the change?

TerryE:
@Alkorin, Incidentally if you agree that this shouldn't have been pulled, then why did you issue a PR in the first place? If you have a change that is not ready for being integrated, then just push it to you local fork and reference the commit in your issue discussions.

Alkorin:
Hum, I could be wrong, but for me it's how PR works. As per github documentation, *Pull requests let you tell others about changes you've pushed to a repository on GitHub. Once a pull request is sent, interested parties can review the set of changes, discuss potential modifications, and even push follow-up commits if necessary.*

It's what I've done here, tell what I've done and discuss about the final road. Next time I can open an issue to discuss about it, but ... issues are not made to track code changes.

Anyway, we have to choose how we handle this compatibility change in #966 and we will close this asap :)

TerryE:
Thomas, I am just getting moderator fatigue. Sorry for this. I clearly made a mistake. Marcel can sort this out. He has my blessing. Any work that I do on nodemcu over the next few weeks at least will relate to my own contributions.

devsaurus:
> Hum, I could be wrong, but for me it's how PR works.

As always, there are two sides of the same coin. The intention in this project is that design discussions and evolution of an implementation is handled in an issue and not in a PR. There's a related recommendation in [CONTRIBUTION.md](https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#contributing-to-nodemcu).
Personally, I made similar experiences like you did and took with me back then: Place a PR if I really *want* the code to be merged; now. But open an issue in case I need constructive feedback to polish the code towards a PR.

On to technical topics...
I've included this update in my project and found the following (maybe already reported somewhere else). After reset, the first call to `ws2812.writergb()` always switches the first LED to green. The next call works then as expected.
```lua
node.restart()
-- wait for reboot
-- all LEDs off

ws2812.writergb(2, string.char(0,0,0, 0,0,0, 0,0,0))
-- 1st LED is green

ws2812.writergb(2, string.char(0,0,0, 0,0,0, 0,0,0))
-- 1st LED is now off
```

Do you see the same?

skycoders:
@devsaurus yes, this is the same here, and also noted by @matgoebl in the backwards-compat ticket #966. Something must be different with the first call

Alkorin:
Will take a look this evening

Alkorin:
Here seems to be the bug:


Pin is high at reboot... I will force it to 0 before starting to stream to see if it fixes the problem.

Alkorin:
@devsaurus I've pushed a fix on my branch

Issue #540, opened 2015-07-08
expose ssl aes


sauloal:
Would it be possible to expose:
app/ssl/crypto/ssl_aes.c

so that we can use public key encryption to send data under insecure connections?

regards

peon374:
I would like to do encrypted auth to cloud service and this would help me too.

marcelstoer:
It's IMHO a bit unfortunate AES isn't part of the [crypto module](https://github.com/nodemcu/nodemcu-firmware/wiki/nodemcu_api_en#crypto-module) but buried in the SSL stuff.

dfeinzeig:
I thought I saw mention that the esp8266 has optimizations for aes?

TerryE:
@dfeinzeig, yes it does, but the API is not documented.  If you wish to retro-engineer it and develop the enhancement then we would welcome this, but in the meantime those actually developing the core firmware need to set their own priorities.

devsaurus:
See related PR against `dev`.

PR #952, opened 2016-01-19
Added AES support to crypto module.

jmattsson:
Merely a wrapper around the (undocumented) internal SDK AES functions.

Possibly sufficient to satisfy the enhancement request in #540. Both ECB and CBC modes supported.

### DIFF
--- a/app/crypto/mech.c
+++ b/app/crypto/mech.c
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2016 Dius Computing Pty Ltd. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * - Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the
+ *   distribution.
+ * - Neither the name of the copyright holders nor the names of
+ *   its contributors may be used to endorse or promote products derived
+ *   from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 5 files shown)


TerryE:
Thanks Johny.  A really useful addition,

robertfoss:
ECB should _never_ be used for anything but schoolbook examples as it does not actually provide any worthwhile security for data sizes larger than one block (16bytes).

CBC is also weak and very prone to security issues (like the recent string of SSL issues). But unlike ECB it offers some sort of meaningful security.

I would prefer ECB to be excluded from this patch.

jmattsson:
It really comes down to risk assessment, as with all security. The issue with ECB is that all identical plaintext blocks result in the same cipher block. If your usage scope is small enough, that can very well be an acceptable trade-off.

If you want to make a note in the docs about CBC being (much) preferred to ECB and a link or two to educational articles, that would certainly be welcome.

TerryE:
Johny beat me to this. This is really ROM functionality that has a small firmware wrapper around it.  Yes maybe we should caveat it's use in our documentation with a health warning, but that is somewhat short of disabling it.  @robertfoss, why not draft third documentation PR.

robertfoss:
@Johny The only secure usecase is for plaintexts of less 16bytes (or
perhaps exactly 16bytes depending on padding). Every other usecase is
non-secure.

I don't intend to be unnecessarily coarse, but ECB mode is not a version of
AES that is actually intended for use other than teach people about the
internals of actual AES algorithms.

This is indeed ROM functionality, but it was also never exposed nor
intended to be called from userspace.

There should be no practical application for ECB in a context like NodeMCU.
If there is, it should be discouraged. Not encouraged.
More functionality is not better in this case. Consider not including ECB
support a feature, rather than the opposite.

TerryE:
I found [this](http://stackoverflow.com/questions/1220751/how-to-choose-an-aes-encryption-mode-cbc-ecb-ctr-ocb-cfb#22958889) interesting discussion on SO.  

Robert, you are arguing that we should not be exposing this AES implementation to Lua-land.  Do you have an alternative proposal?

robertfoss:
I'm arguing that we shouldn't expose ECB mode. It might be implemented
already, but nothing good can come from exposing it.
CBC mode is a lot better, and actually can be secure if used properly.

PR #955, opened 2016-01-20
Fix example code in WiFi Module Doc

zhujunsan:


TerryE:
This correction is fine. However some of the example snippets fall short of Lua best practice. We really need to do an end-to-end revue of all these examples at some point.

Issue #927, opened 2016-01-12
Switch on more compiler warnings; improve codebase accordingly


jmattsson:
Currently we're building NodeMCU with the inherited compiler settings from Espressif. They're very, very lax, and not doing much to help us get a high quality code base. While [some code](https://github.com/nodemcu/nodemcu-firmware/blame/master/app/modules/tmr.c#L1) is compiled with higher warning levels, most isn't.

Getting to -Wall -Wextra with -std=gnu99 or gnu11 for app/modules, app/platform and app/user should be our minimum aim. Third party code we clearly have less influence over, but at the very least we should build that with their recommended warnings enabled.

I'm hopeful that getting us standards clean will also improve the code generation and thus give us a reduced flash foot print.

devsaurus:
Fully agree and consider me in for any overhaul of the code base.
I've been bitten a couple of times just because the compiler silently accepted crappy code in the first place.

TerryE:
+1, but its going to be a pain in the arse clearing all of those warnings.  Divide and conquer?

devsaurus:
> Divide and conquer?

We had good experience with this approach for the doc transfer job. Ground rules were:
* central "job table" [Module list](https://github.com/nodemcu/nodemcu-firmware/wiki/Module-list) where WIP status is published to avoid duplicate / conflicting efforts
* direct commits to a dedicated branch
* have an issue for discussion

Looking at #958 the division lines would be defined by the Makefiles, further boiling down to directories or even whole subtrees.
@jmattsson how about a similar approach? Set up a wiki page containing all "Makefiles" to take away for WIP and check mark when done. Commits go into a dedicated branch at nodemcu-firmware repo (*insert snappy name here*) until it's finally merged to `dev` - Or do you prefer a PR per developer and merge these one by one?
I would set up branch and wiki page if you're ok with that.

jmattsson:
I would really prefer we get incremental improvements into `dev` as we go along. If we don't do that we'll discover by the time we're ready to merge back that there's another ton of things that need to be fixed that have crept in in the meantime.

We don't need a big-bang approach here, and in fact it might be outright helpful to do this piecemeal. There will be a substantial amount of changes resulting from this, and if we get testing on each chunk as we go along it will be a lot easier to pinpoint any regressions that might be introduced.

If people are interested in working on this, having a "job table" on the wiki could be handy.  Something like this perhaps:

Dir | gnu11 | Wall | Wextra | notes
[...]

PR #958, opened 2016-01-20
Enable -std=gnu11 for modules dir.

jmattsson:
Plus boat-load of fixes to actually make that build. Who needs prototypes... *sigh*

This is the first step on the long path of #927 ...


jmattsson:
Thanks for pointing those out @devsaurus. I changed to use STD_CFLAGS rather than EXTRA_CCFLAGS, to avoid it being clobbered by the command line. CI now builds with the expected flags.

I also added -Wimplicit, and sorted out the other small things you pointed out.

devsaurus:
Looks very good and also passed some basic use cases here.
:+1: from my side - I'd merge it soon if nobody objects.

TerryE:
+1

PR #972, opened 2016-01-24
Fix redefine error on some other version build environment.

vowstar:
This macro will cause EFAULT redefine error while compile in espressif's Lubuntu build environment and ai-thinker's build environment. So add #ifndef EFAULT to check it is more safe.

### DIFF
--- a/app/include/arch/cc.h
+++ b/app/include/arch/cc.h
@@ -38,7 +38,10 @@
 #include "c_types.h"
 #include "ets_sys.h"
 #include "osapi.h"
+
+#ifndef EFAULT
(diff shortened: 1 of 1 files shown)


jmattsson:
Note that this header is part of LWIP, so this fix is likely to go missing at the time of the next SDK+LWIP upgrade.

vowstar:
I have another (not good) idea, check the environment in Makefile and define in CCFLAGS while using espressif's Lubuntu. The espressif's lubuntu environment defined COMPILE=gcc and windows XP environment defined lx106.

PR #977, opened 2016-01-26
Fix HTTP response bug while server only support HTTP/1.0

vowstar:
- Fix HTTP response bug while server response HTTP/1.0.
When server use HTTP/1.0 API, NodeMCU will receive error. So fix it, and now support both  HTTP/1.0 and HTTP/1.1,
- Change BUFFER_SIZE_MAX from 1024 to 0x2000. This doesn't consume more RAM because it is only a limit value, the buffer is allocated when data received. (but Lua in firmware not support such a long string, I don't find better way to hold it, may write it to filesystem directly?)

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -354,14 +354,16 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
 		else if ( req->buffer[0] != '\0' )
 		{
 			/* FIXME: make sure this is not a partial response, using the Content-Length header. */
-			const char * version = "HTTP/1.1 ";
-			if ( os_strncmp( req->buffer, version, strlen( version ) ) != 0 )
+			const char * version_1_0 = "HTTP/1.0 ";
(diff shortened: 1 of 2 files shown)


dvv:
Wonder if anyone ever tried to port https://github.com/nodejs/http-parser here? As they state, "Depending on your architecture, it only requires about 40 bytes of data per message stream".

I ago did similar https://github.com/dvv/ray-http/blob/master/parser/lua-http-parser.c

jmattsson:
These changes look good. I'll raise a separate PR for the http module docs to note the limit change.

As for long strings, I don't think writing to the filesystem directly is a (good) answer here. If anything, perhaps having an advanced streaming mode option in the API to allow the callback to be called incrementally, much like the `net` module does with `on('receive',...)`.

vowstar:
@dvv These code is very cool, it can be used not only in HTTP client but also in HTTP server. I will try it and let this module support both client and server.
@jmattsson Yes, a better solution is streaming method, when long data is come, it can split data and callback while every segment received.

PR #981, opened 2016-01-27
Fixed issue where default ESP_xxxxxx SSID was appearing

jfollas:
...for enduser_setup

Looks like SDK 1.5.1 may have changed behavior: AP config is not saved unless the wifi is in SoftAP (or Station+AP) mode.

Without this change, the SSID is not set to the "SetupGadget" naming convention (or whatever prefix you may have defined in user_config.h). Instead, you'll see the last-saved SSID being used, and by default, that looks like "ESP_xxxxxx".


jmattsson:
Thanks!

Issue #920, opened 2016-01-11
TM1829 module missing in master & dev


marcelstoer:
Just noticed that #686 is missing in master & dev. I couldn't find any notes as to why this could be intentional. Was it simply forgotten in the recent 1.4 upgrade and the necessary branch/PR clean-up (PR was directly merged to master)?

jmattsson:
I'm not aware of any reason why it should be excluded. I'd say it was simply missed in the cleanup.

TerryE:
+1 My bad.

devsaurus:
Thought it was still awaiting some kind of check/verification since it never got in touch with a recent SDK on `dev`.

TerryE:
Yup, I did ask the contributors to resubmit, but I should have followed it up.

marcelstoer:
@h3ndrik can you please take care of that.

h3ndrik:
thanks. submitted new pull requests for master #983 and dev-branch #984

PR #983, opened 2016-01-28
add TM1829 support

h3ndrik:
This code is taken from the WS2812 module with modified timings and inverted logic levels

see #920: this is a resubmission after the module cleanup


marcelstoer:
Thanks for the effort but there's no need to create a PR against `master`. The module will land here once `dev` is promoted to `master`. Sorry if the description in #920 was misleading, my bad.

Issue #774, opened 2015-11-18
Fix the API documentation mess


marcelstoer:
I want to drop the Wiki documentation and move the content to the repository so we can serve it with Read The Docs. See http://read-the-docs.readthedocs.org/en/latest/features.html for details. This would solve two major pain points:

- Documentation not in line with the code

  If the documentation is moved to the code repository we can verify with every commit/PR that the documentation is updated accordingly. Furthermore, the documentation is merged automatically with the code if it moves from branch X to Y. The human-readable docs are built automatically with every commit.

- L10n

  Read The Docs (and similar solutions) offer L10n to easily serve the documentation in multiple languages. This would replace the inconsistent Wiki pages for non-EN languages.

jmattsson:
I've never used Read The Docs, but as I've mentioned elsewhere I'm all in favour of switching to *something* like this. I had a look through the docs but I didn't see much in terms of what mark-up/down/sideways could/would/should be used. I'm familiar with JavaDoc / Doxygen syntax which works pretty well for documenting code. For user-facing docs it's probably not as appropriate (though the easy API cross-referencing is most handy).

Anyway, I guess what I'm trying to say is that I'd be happy to give anything a go as long as it seems reasonable, and especially if you have experience working with this particular flavour and can set a good example for the rest of us :)

marcelstoer:
> I had a look through the docs but I didn't see much in terms of what mark-up/down/sideways could/would/should be used.

It supports reStructuredText and Markdown.

jorgempy:
+1

andyhu:
+2

TerryE:
I've never come across Read The Docs before but I have used and am familiar with Doxygen both for end-user and programming documentation, so it would get my vote simply on the grounds that it is one less system to learn, but what the heck if someone sets up a usable system then I'll use it.  eLua uses another system -- can't recall which off the top of my head.

marcelstoer:
I just found out that the SmartArduino/DOIT/MXCHIP.INC folks do a pretty good job with the introductory documentation for WiFiMCU at https://www.gitbook.com/book/fineshang/wifimcu-based-on-emw3165-user-manual/details. As you can see they're on GitBook which is veeery similar to RTD.

Others are in a similar situation as we are: https://civicrm.org/node/3268, http://j.mp/1YqisPD, http://wp.me/p2YeUz-1sG.

jmattsson:
[...]

PR #984, opened 2016-01-28
add tm1829 module to dev branch

h3ndrik:
This code is taken from the WS2812 module with modified timings and inverted logic levels

see #920: this is a resubmission after the module cleanup
also add some API description  (#774)


marcelstoer:
Thanks a lot! I'd like to see two additional things:
- add your doc file at the right position (alphabetical) to https://github.com/h3ndrik/nodemcu-firmware/blob/dev/mkdocs.yml
- squash all your commits into one as per https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#general-flow

@devsaurus in https://github.com/nodemcu/nodemcu-firmware/issues/920#issuecomment-170825882 you mentioned some verification you wanted to perform?

TerryE:
Can I give this a :-1:  or at least a hold.  We shouldn't be adding more modules which break the Espressif timing guidelines and cause the CPU to crater, until we agree a policy for addressing these issues.  In this case interrupts are disabled for ~40N µSec where N is the number of LEDs.

devsaurus:
> @devsaurus in #920 (comment) you mentioned some verification you wanted to perform?

Not me. In general verification that the code in `old-master` SDK 0.9.6 is functional with a recent SDK. That's been the reason why the code wasn't pulled into `dev` as is.
Can be considered as done, now that @h3ndrik rebased the code onto `dev`.

Alkorin:
Copy-and-paste programming, fonctions comments were wrong in ws2812 module, and ... same here :/

It looks like a similar protocol than ws2812, we could try to share code / use similar implementation.
I don't have any TM1829 here to test :/

TerryE:
The main difference is that the logic is inverted that is each data bit is a low-high rather than high-low, so  whilst your encoding scheme can easily be mapped onto the 1xxxxxxxx0 bit sequences output by the uart, this will need to be shifted one sub-bit for the TM1289. We could get this to work, I think, but test HW and an oscilloscope will be needed.

marcelstoer:
@h3ndrik do you intend to fix the code based on the the comments here? If not then please close he PR.

marcelstoer:
No feedback in more than two weeks -> closing.

PR #987, opened 2016-01-29
Enhancements to enduser_setup module

jmattsson:
The major enhancement is the support for scanning for available WiFi networks, and picking from the resulting list. Additionally there is a bunch of smaller fixes. See aef41c6 for a detailed breakdown on the changes.

Tested on Windows Phone 8.1 (IE), recent iOS, Android, Firefox, Chrome and Edge.

@robertfoss it'd be great if you could have a look over this. I tried to keep the clean look you had going, but I'll readily admit that I'm not a web front-end person :)

This PR also contains a couple of supporting fixes related to libc usage & linking.

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -29,12 +29,15 @@
  * OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  * @author Robert Foss <dev@robertfoss.se>
+ *
+ * Additions & fixes: Johny Mattsson <jmattsson@dius.com.au>
  */
 
 #include "module.h"
 #include "lauxlib.h"
 #include "platform.h"
 #include "c_stdlib.h"
+#include "c_stdio.h"
 #include "c_string.h"
 #include "ctype.h"
 #include "user_interface.h"
@@ -42,6 +45,7 @@
 #include "flash_fs.h"
 
 #define MIN(x, y)  (((x) < (y)) ? (x) : (y))
+#define LITLEN(strliteral) (sizeof (strliteral) -1)
 
 #define ENDUSER_SETUP_ERR_FATAL        (1 << 0)
 #define ENDUSER_SETUP_ERR_NONFATAL     (1 << 1)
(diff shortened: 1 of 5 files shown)


robertfoss:
This looks like a solid patch to me. I've tested the code and it works well.

I have one usability improvement to suggest.
When a WiFi has been connected to, instead of shutting down SOFT_AP and the HTTP server right away, the user should probably be shown a page saying that the wifi has been connected to and that the AP is going to shut down. Just to minimize confusion.
So possibly one would want to keep SOFT_AP and HTTP on for another second or so just to serve this message.

Anyway, thanks for all the improvements.

jfollas:
The timing of this pull request coincides with similar work that I've been doing to the same module, so thanks for your improvements! 

One improvement that I made in my own codebase was to compress the HTML and JavaScript even further: I renamed CSS classes and element ids to be as short as possible, and used a JavaScript compressor that replaced variable names with single characters, etc. Once this HTML was as tiny as I could get it, I gzip/zopfli compressed the file and stored those bytes instead of the HTML itself in the C code (and then on the response, I also set Content-Encoding: gzip when I served that particular file). I mainly did this so that I could embed a company logo onto that web page yet still work with a manageable memory footprint. 

While I can't submit my own page to the project because it's branded, I may repeat this exercise with your code after the pull request has been merged, if you think there's value in that.

jmattsson:
@robertfoss  Good suggestions! I'll fix up that sentence when I'm on my workstation again. I probably won't have time to add the wait-until-a-few-seconds-later-before-stopping, but I like the idea.

@jfollas I contemplated gzip, but I wasn't sure how widely supported it was these days (did I mention I'm not a web guy?)

jfollas:
@jmattsson In short, there's less support for XMLHttpRequest than gzip encoding in browsers (going back as far as IE5 timeframe). Downside for something like enduser_setup is that you have to generate the gzip payload byte array yourself in order to embed into the source, and that process is not as straightforward as editing a string literal. The upside, though, is the 40% size savings on top of the HTML/CSS/JS minification (makes the ESP happy), which is why I do it in my fork.

app/modules/enduser_setup.c
```diff
@@ -478,90 +665,263 @@ static int enduser_setup_http_serve_html(struct espconn *http_client)
...
+      bool already = (state->scan_listeners != NULL);
+
+      l->next = state->scan_listeners;
+      state->scan_listeners = l;
 
-    retval = enduser_setup_http_serve_html(http_client);
-    if (retval != 0)
+      if (!already)
+      {
+        if (!wifi_station_scan (NULL, on_scan_done))
```
jfollas: This will fail if WiFi is not in Station (or Station+SoftAP) mode. I had a situation today where I entered into enduser setup with wifi mode=0. The AP was correctly turned on when start() was called, but nothing in code considered turning on the Station from within enduser_setup.

jmattsson:
@jfollas Right. That's something that happens when you're not using `manual` mode? As you might imagine we're not actually using that mode here at $work, and I'm not entirely comfortable I understand the intents of the other code path. Could you and/or @robertfoss give me an idea of how this should work in that code path? I'm assuming there's a reason why it doesn't simply drop into STA+AP mode in `start()`, but I haven't worked out why...

jfollas:
I can't really speak to manual mode - @robertfoss will need to explain that use case. 

My normal workflow is that Lua script tries to connect to the network, and then if it can't, I start up enduser_setup. In this one case today, I was testing my fork of EUS directly without the benefit of Lua scripts trying to start up networking on the STA first, which meant that the STA interface was not active ahead of time. This made the /aplist request not return anything to JavaScript (and the code that checks for IP Address in order to shut down EUS also failed to work with the STA off).

My workaround, at least, was to just make sure that we were in STA+AP mode when the AP started up in EUS because we need both for the code to work correctly.

robertfoss:
@jmattsson I don't actually think there is a reasons for STA+AP not being set straight away.
As far as I see, there's nothing preventing us from going that route.

jfollas:
@robertfoss @jmattsson Thoughts on this:

When I connect to EUS from my laptop, which may have a bunch of open web pages and services running, etc, the first thing that happens is everything wants to try to reconnect to their servers. So I see a massive amount of DNS and HTTP requests happening all at once.

```espconn_send``` will only allow 8 operations to be buffered, so in the scenario above, which is often less than the number of active requests that my laptop is making of the captive portal. ```enduser_setup_dns_recv_callback``` will treat the ```ESPCONN_MAXNUM``` error code as a fatal error, and shut down EUS.

I'm proposing that this scenario be treated as a non-fatal error, because DNS operates on UDP anyways, which has no guaranteed delivery.

robertfoss:
That sounds very reasonable to me.

jmattsson:
Alright, I've fixed up the odd sentence in the docs, and changed to drop into station+ap mode right away. Any objections to getting this merged, so you guys can work on the other improvements on top of this?

jfollas:
Ship it! :-)

robertfoss:
I'm happy.

Issue #872, opened 2015-12-23
Changing the default user settings.


TerryE:
At the moment the "default" `dev` compile doesn't load on the vast majority of ESP modules out there (the 512Kb parts).  My recommendation is that we change the default in user_config.h and user_modules.h so that if  a developer downloads the repo dev branch or zip and does a make then the binary is usable on a 512Kb part.  

Anyone who needs to add anything that needs an additional module can uncomment the relevant line and do the make again.  My suggestion for defaults are:
```
bit_rate               = 115200
client_ssl             = disable
gpio_interrupt         = enable
spiffs_cache           = enable
optimize_debug         = 2
led_high_count_default = 10
led_low_count_default  = 0

builtin = string,table,coroutine,math,debug_minimal
modules = node,file,gpio,wifi,net,i2c,spi,tmr,adc,uart,ow,bit
```
This summary is in my `cnf` format for the sake of brevity,  but the change would be in terms of `//` edits of the `user_*.h` files of course.

jmattsson:
I guess removing SSL saves a lot of space? It feels a bit wrong to remove it, but I can see an argument for it.

Module list looks okay. If anything, I'd suggest also including rtctime since the default timekeeping on the ESP is rather horrible.

nickandrew:
As I have never used `i2c,spi,adc,bit` modules I'd suggest removing them from the default config :smile:

I also suggest removing `ow` as IMHO the point of a default config is like a demo mode; it should be enough to boot up the NodeMCU and write some Lua scripts and do some networking; onewire isn't crucial for that.

TerryE:
I'd keep bit, and i2c.  but that's just my one vote.

devsaurus:
> bit_rate = 115200

Nice - the change should be highlighted in the Readme. 

+1 for bit - it's a valuable substitution for the missing & and | operators.

i2c, spi, ow, adc, and gpio are general purpose interface modules. If we can't set a prio then either keep or remove them all. I'm for keeping them as @TerryE suggested.

TerryE:
I feel that 64Kb is about the smallest usable SPIFFS, and there is no way to achieve this it if we include SSL or SHA2.

pjsg:
Can we have a define to indicate that we are targeting the large flash parts which includes much more stuff. I.e. if I add #define FLASH_LARGE (or similar), then I get an expanded set of modules -- maybe all of them.

marcelstoer:
[...]

PR #988, opened 2016-01-29
Simple mdns implementation using the Espressif code

pjsg:
This PR is a thin layer over the espressif provided mdns implementation. It only allows a single service to be advertised and it doesn't allow the hostname and the instance name to be broken apart. However, it works, is simple, and reasonably compact.


app/include/user_modules.h
```diff
@@ -27,6 +27,7 @@
 #define LUA_USE_MODULES_HTTP
 //#define LUA_USE_MODULES_HX711
 #define LUA_USE_MODULES_I2C
+#define LUA_USE_MODULES_MDNS
```
marcelstoer: I'd rather have this disabled by default even though #872 is still waiting for a resolution.

app/modules/mdns.c
```diff
@@ -0,0 +1,130 @@
...
+  return 0;
+}
+
+static struct wrapper *process_args(lua_State *L, int phase, size_t *sizep)
+{
+  struct wrapper *result = NULL;
+  char *p = NULL;
+
+  if (phase == 1) {
+    result = (struct wrapper *) c_zalloc(sizeof(struct wrapper) + *sizep);
```
Alkorin: Never freed ?

Alkorin:
It tooks me more than 10 minutes to understand what you tried to do with your phase 0 / phase 1.

You could split this func in two parts with more comments, it'll add readability to your code.
setptr is not obvious too... You use only 3 times, it's maybe better to write it like this

```
...
tmpBuffer = &result->data;
...
// Get server name
char * server_name = luaL_checkstring(L, 2)
// Store it
c_strcpy(tmpBuffer, server_name);
result->mdns_info.server_name = tmpBuffer;
tmpBuffer += c_strlen(server_name) + 1;
```
than
```
setptr(&result->mdns_info.server_name, &p, luaL_checkstring(L, 2));
```

It's easier to read.

pjsg:
I cleaned up the code a bit and removed the setptr function. 

Also, I disabled the module by default. Turns out that the lwip code adds a fair amount of RAM consumption (maybe 700 bytes) if the mdns code is included.

makefu:
i'd love to see this in mainline :+1:

saeugetier:
I would like to see it in the mainline, too. It's a nice feature. In an DHCP environment you don't know each IP of each nodemcu module.

pastukhov:
:+1:

pastukhov:
Any progress?

pastukhov:
Hi! @pjsg.
I have cherrypicked tis pr, but can't get this working.
I have add ```lua mdns.register('socket', "http", 80, { hardware='NodeMCU'})``` to my http server startup script and not able to find socket.local with avahi-discover or dig @224.0.0.251 -p 5353 socket.local

pjsg:
I just built the branch again (against the current dev branch) and ran your mdns command line.

I then ran avahi-browse on my linux box:

avahi-browse --resolve _http._tcp
+ eth0 IPv4 socket                                        Web Site             local
+ eth0 IPv4 Living Room                                   Web Site             local
= eth0 IPv4 socket                                        Web Site             local
   hostname = [socket.local]
   address = [192.168.1.234]
   port = [80]
   txt = ["hardware=NodeMCU" "vendor = Espressif"]

You can see that 'socket' appears. Initially it didn't work until I modified my iptables configuration on my linux box to permit inbound UDP/5353 traffic. Note that the response doesn't come from the same address as the original packet destination, so the default return path doesn't work. Thus you need an explicit allow rule.

Also note that the service type is being prefixed by an underscore, so it doesn't matter if you include it in the registration or not.

Also, I think that you need to do the registration *after* you are connected to the network.

FrankX0:
I am using mDNS with the latest dev build, but cannot get it stable to work.
The first few minutes the functionality works as expected, but after 5 to 10 minutes the service seems to stop. Through a web-browser I can no longer connect using the .local domain. Using dns-sd in Windows 10 also does not result in finding the IP address.
Any suggestions?

Issue #982, opened 2016-01-28
calculating sha1 hash of big files is not possible


kmpm:
I want to calculate checksum of big (> ~1460 bytes) files on the nodemcu but I can't find any way of doing it currently.

It seems as one would need some way of not reading the whole file into a string and then do checksum but read in blocks and update instead.

Examples.
- http://stackoverflow.com/questions/1869885/calculating-sha1-of-a-file
- http://luacrypto.luaforge.net/0.1/luacrypto.html ( look at d:update(string) )
- https://docs.python.org/2/library/hashlib.html

jmattsson:
I agree that would be nice to have. Alas I don't have time to add this support right now, but what I'd envisage is this sort of usage:
```lua
local sha = crypto.new_hash("SHA1")
sha.update("mystring1")
sha.update("more stuff...")
local digest = sha.finalize()
```
I believe we have all the internals needed, we "just" need to glue it together in the right way. Someone is most welcome to beat me to this work.

ABonner:
I'd be happy to try to implement this over the weekend. It looks like we would pass back the contents of the `SHA1_CTX` struct as userdata to persist the state across calls.

jmattsson:
I'd recommend `digest_mech_info_t` (or suitable subset thereof) plus the working buffer for the userdata. That way we'll automatically get the support for MD5 and SHA2 and/or whatever else has been enabled. Have a look at app/crypto/crypto.{h,c} for inspiration. :)

ABonner:
Thanks for the tip, I was looking at the `crypto_sha1` function instead since the topic mentioned SHA1. I'll look at the more generic way of doing it.

jmattsson:
Great, let me know if you have any questions once you dig in. Oh, and do the work off the `dev` branch please, it'll save you hassle once you want to raise the PR. And Marcel would probably like me to point you to [CONTRIBUTING.md](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) for other useful info too.

cheloftus:
Would it be beneficial to have the file IO handled outside of the Lua world? Seems like it might be more efficient for the file hashing case. Perhaps provide a crypto export that specifically hashes files?
I was envisaging something like:

```lua
local digest = crypto.fhash("SHA1", "myfile.ext")
```

I don't dispute that the flexibility of exposing functions like `new_hash`, `update`, `finalize` would be very useful in a more general sense though.

[...]

PR #992, opened 2016-01-30
Added crypto.fhash function for hashing files

cheloftus:
Added a new export for crypto that  provides hashing of files.
e.g.
```lua
local digest = crypto.fhash("SHA1", "myfile.ext")
```
File IO requirements kept within `crypto.c` module. New function added to `digests.c` to perform block by block hashing with a function pointer to a contiguous read function. This seemed neater than doing file I/O within `digests.c`.

Fixes issue #982.

### DIFF
--- a/app/crypto/digests.c
+++ b/app/crypto/digests.c
@@ -122,6 +122,36 @@ int ICACHE_FLASH_ATTR crypto_hash (const digest_mech_info_t *mi,
 }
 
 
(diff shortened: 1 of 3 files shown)


TerryE:
Also why 092d532?

cheloftus:
Thanks TerryE. I'll  add return checking, and re-submit.

Issue #730, opened 2015-11-05
Multiple conn:send() not working


positron96:
When I was on a stable build (0.9.6-dev_20150704), the code like this worked correctly:
````
    connection:on ("receive",
    function (conn, req)
        conn:send("HTTP/1.0 200 OK\n")
        conn:send("Server: ESP (nodeMCU)\n")
        conn:send("Content-Length: 1\n\n")
        conn:send("a")
        conn:close()
    end )
````
This worked for a client socker as well as a server socket.

Now I moved to dev build that says: 
````
NodeMCU custom build by frightanic.com
	branch: dev
	commit: d8c6fe0b25142f6ae866c18c3bcb3d0538d4baaf
	SSL: false
	modules: node,file,gpio,wifi,net,pwm,tmr,uart,ow
 build 	built on: 2015-11-03 20:47
````
And only first conn:send actually sends, all following calls get silently discarded. I now need to store the lines into a table and join table into a string which is probably a waste of memory.

Is it a bug or is it intended?

**Comment by TerryE** [comment]
The outcome of your code is indeterminate as network requests are treated as separate tasks by the SDK.  Ignoring the fact that you should marshal your sends where practical, something like this would work:
```Lua
  connection:on ("receive",
    function(sck, req)
      local response = {"HTTP/1.0 200 OK", "Server: ESP (nodeMCU)", "Content-Length: 1\n", "a"}
      local function sender (sck)
        if #response>0 then sck:send(table.remove(response,1))
        else sck:close()
        end
      end
      sck:on("sent", sender)
      sender(sck)
    end )
```
One network call per task and no ordering problems.  And yest is could be viewed as bad coding to have two semantically different variables both called `sck` but in reality they both point to the same userdata.  This really falls under #719, but I'll give you a chance to reply first :smile:

jmattsson:
> Is it a bug or is it intended?

This is a result of changes within the SDK. Pre-1.x the SDK seemed willing to buffer up contents from several send() calls despite it not being officially supported, but that is no longer the case. I suspect (but haven't verified) that this happened when they freed up substantial amounts of RAM for application use.

So well, this is neither a bug nor intended, it's simply what's being imposed by the SDK.

positron96:
Hello Terry. The code you provided works, as does concatenating the table into string, but sending a table with recursive ''''sent'''' events is not very intuitive to say the least. I'll sacrifice memory and stick with table concatenation as more clear way.
[...]

PR #993, opened 2016-01-31
Queue for net.socket:send

Hellsy22:
Implementation of a queue for ```net.socket:send``` (related: https://github.com/nodemcu/nodemcu-firmware/issues/730)
```lua
sk.send("one");
sk.send("two");
sk.send("three");
```

will send "one", "two", "three" to the socket, instead of sending "one" and discarding all other data. After empting the queue a ```sent``` event will occur.

Backward compatibility:  previous ```send-sent-send more``` model is
still usable without any extra memory cost.

Bonus: no more 1460 chars string limit - strings longer then 1460 will
be chunked to queue.

Formatting the code: tabs changed to two spaces, trailing spaces
removed etc.

Testing: worked for me for TCP client and server for a few days.

TODO: (discuss) fill entire queue block before creating another one for a bit better performance.
(discuss) Should fire ```sent``` event if connection was closed before queue empty?
(discuss) LUA-side access to the attribute ```sending``` and to the queue size.

Also I found undocumented features, like socket:hold() or server connections limit (5 atm) - any suggetions where I can obtain more information about net module?

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -46,6 +46,7 @@
 #define ICACHE_RAM_ATTR __attribute__((section(".iram0.text")))
 
 #define CLIENT_SSL_ENABLE
+//#define SENDQUEUE_ENABLE
 #define GPIO_INTERRUPT_ENABLE
 //#define MD2_ENABLE
 #define SHA2_ENABLE
--- a/app/modules/net.c
+++ b/app/modules/net.c
@@ -11,7 +11,7 @@
 #include "mem.h"
 #include "lwip/ip_addr.h"
 #include "espconn.h"
-#include "lwip/dns.h" 
+#include "lwip/dns.h"
 
 #ifdef CLIENT_SSL_ENABLE
 unsigned char *default_certificate;
@@ -39,6 +39,15 @@ static uint16_t tcp_server_timeover = 30;
 static struct espconn *pTcpServer = NULL;
 static struct espconn *pUdpServer = NULL;
 
+typedef struct t_sendqueue // queue chain [block]->[block]->...->NULL
+{
+  uint16_t len;
+  char *data;
+  struct t_sendqueue* next;
+} t_sendqueue;
+
+static int sendqueue_commited = 0;
+
 typedef struct lnet_userdata
 {
   struct espconn *pesp_conn;
@@ -49,10 +58,20 @@ typedef struct lnet_userdata
   int cb_receive_ref;
   int cb_send_ref;
   int cb_dns_found_ref;
+  struct t_sendqueue *sendqueue; // stick to lua-datablock
+  uint8_t sending; // flag
 #ifdef CLIENT_SSL_ENABLE
   uint8_t secure;
 #endif
-}lnet_userdata;
+} lnet_userdata;
+
+#define SENDQUEUE_MAX_BLOCK_SIZE 1460    // chunking threshold
+
+static void sendqueue_prepare(lnet_userdata* nud);  // Init queue
+static void sendqueue_drop(lnet_userdata* nud);     // Destroy entire chain
+static void sendqueue_node_free(t_sendqueue *d);    // Destroy one node
+static void sendqueue_proceed(lnet_userdata* nud);  // Send
+static int sendqueue_add(lnet_userdata* nud, char *data, uint16_t len);
 
 static void net_server_disconnected(void *arg)    // for tcp server only
 {
@@ -80,11 +99,12 @@ static void net_server_disconnected(void *arg)    // for tcp server only
     lua_rawgeti(gL, LUA_REGISTRYINDEX, nud->self_ref);  // pass the userdata(client) to callback func in lua
     lua_call(gL, 1, 0);
   }
+  sendqueue_drop(nud);
+
   int i;
   lua_gc(gL, LUA_GCSTOP, 0);
-  for(i=0;i<MAX_SOCKET;i++){
-    if( (LUA_NOREF!=socket[i]) && (socket[i] == nud->self_ref) ){
-      // found the saved client
+  for(i=0; i < MAX_SOCKET; i++){
+    if( (LUA_NOREF!=socket[i]) && (socket[i] == nud->self_ref) ) {
       nud->pesp_conn->reverse = NULL;
       nud->pesp_conn = NULL;    // the espconn is made by low level sdk, do not need to free, delete() will not free it.
       nud->self_ref = LUA_NOREF;   // unref this, and the net.socket userdata will delete it self
@@ -119,6 +139,7 @@ static void net_socket_disconnected(void *arg)    // tcp only
   if(nud->pesp_conn)
     c_free(nud->pesp_conn);
   nud->pesp_conn = NULL;  // espconn is already disconnected
+  sendqueue_drop(nud);
   lua_gc(gL, LUA_GCSTOP, 0);
   if(nud->self_ref != LUA_NOREF){
     luaL_unref(gL, LUA_REGISTRYINDEX, nud->self_ref);
@@ -172,6 +193,12 @@ static void net_socket_sent(void *arg)
   lnet_userdata *nud = (lnet_userdata *)pesp_conn->reverse;
   if(nud == NULL)
     return;
+
+  if (nud->sendqueue) {
+    sendqueue_proceed(nud);
+    return;
+  }
+
   if(nud->cb_send_ref == LUA_NOREF)
     return;
   if(nud->self_ref == LUA_NOREF)
@@ -313,6 +340,7 @@ static void net_server_connected(void *arg) // for tcp only
   lua_pushvalue(gL, -1);  // copy the top of stack
   skt->self_ref = luaL_ref(gL, LUA_REGISTRYINDEX);    // ref to it self, for module api to find the userdata
   socket[i] = skt->self_ref;  // save to socket array
+  sendqueue_prepare(skt);
   socket_num++;
   skt->cb_connect_ref = LUA_NOREF;  // this socket already connected
   skt->cb_reconnect_ref = LUA_NOREF;
@@ -373,7 +401,7 @@ static int net_create( lua_State* L, const char* mt )
 #endif
   uint8_t stack = 1;
   bool isserver = false;
-  
+
   if (mt!=NULL && c_strcmp(mt, "net.server")==0)
     isserver = true;
   else if (mt!=NULL && c_strcmp(mt, "net.socket")==0)
@@ -431,6 +459,7 @@ static int net_create( lua_State* L, const char* mt )
 #ifdef CLIENT_SSL_ENABLE
   nud->secure = secure;
 #endif
+  sendqueue_prepare(nud);
 
   // set its metatable
   luaL_getmetatable(L, mt);
@@ -500,7 +529,7 @@ static int net_create( lua_State* L, const char* mt )
     lua_call(L, 1, 0);
   }
 
-  return 1; 
+  return 1;
 }
 
 // static int net_close( lua_State* L, const char* mt );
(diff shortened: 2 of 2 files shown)


nickandrew:
Looks Good To Me. It should make sending more data to a socket a bit easier to code.

Hellsy22:
@nickandrew, yes, after the last.

TerryE:
Just to expand on my third point, you will see that the **socket** parameter is the various connection return and other network callback parameters is a USERDATA type and is the same value for any given connection so a native Lua implementation might be:
```Lua
-- file setup_multi_send.lua
local name = ...
return function(sk, on_send_cb)
  local buf
  name = nil  -- make ephemeral
  sk:on('sent', function(sk)
    if #buf then 
      return sk:send(table.remove(buf,1)
    else 
      buf = nil; return send_cb(sk)
    end
  end)
  return function(rec)
    if buf then 
      buf[#buf+1] = rec
    else 
      buf = {}; return sk:send(rec)
    end
  end 
end
```
and the new Lua implementation:
```Lua
sk=net.createConnection(net.TCP, 0)
local sk_send = require("setup_multi_send")(sk, cb)
sk_send(rec1)
sk_send(rec2)
sk_send(rec2)
```
Lua is going to manage the storage and GC of rec1, rec2, ... anyway.  There is little point in dubling up this storage when you've got so little RAM to play with.  If you think that this approach is too complicated from a Lua programmers perspective you should still stash the buffered strings in a Lua table hooked into the Lua Registry so that you don't double the memory overhead of dong this buffering.

pjsg:
@TerryE -- I'm not enough of a Lua guru to know how to do this, but I suspect that it would be possible to do

    sk=net.createConnection(net.TCP, 0)
    sk = require("setup_multi_send_wrapper")(sk)
    sk:send(rec1)
    sk:send(rec2)
    sk:send(rec2)

where the wrapper code creates a new object with a metatable and arranges to forward the method calls onto the original 'sk' -- except for the 'send' method which would be handled as in your code above. *IF* this was possible, then it would be a clean solution for all of these `sk:send` problems. Your approach requires changing all the `sk:send` calls into `sk_send` which is more prone to error.

Hellsy22:
@TerryE 
> It doesn't have a configuration option to disable this functionality at compile time.

Do you mean to disable only or to fully wipe it from code?

> It doesn't have a runtime option to enable or disable this behaviour.

I can implement ```net.config```, to set memory queue usage limit or to disable the queue.

> so why duplicate this

Duplication required only for splitting too long requests. Good point for future improvement.

> and change the behaviour of existing Lua applications that conform to the SDK 1.x guideline

New functionality have no impact on memory usage or perfomance of old model with ```send-sent-send```. Data will be copied only if sending already in progress and user want to send more.

You right, I don't understand all the aspects. But I believe, that the dropping data by ```send``` even without raising an error - is worst way ever. That Espressif was wrong when changed core SDK functionality and forced all users to find complicated work-arounds. So, I just want previous functionality back. And I appreciate any help, any bit of information about net module, any questionable case to check.

Would you like to help me?

> This still doesn't address the sk:close() issue.

What do you mean?

nickandrew:
To avoid incompatibility with the previous API (worst case: code assumes sends are queued when they are not), send() could be renamed to sendq().

TerryE:
>> It doesn't have a configuration option to disable this functionality at compile time.

>Do you mean to disable only or to fully wipe it from code (like SSL)?

This is core functionality.  Only some developers will want to try to emulate legacy behaviour, given that it trivial to modify your Lua code to work stably with the new 1.x SDK behaviour.  The _only_ way that this could be acceptable is if the default compile disables it.

>> It doesn't have a runtime option to enable or disable this behaviour. 

> I can implement ```net.config```, to set memory queue usage limit or to disable the queue.

No, as Nick suggests, the standard API should be standard.  This should be a different API function, not `send`.
 
>> so why duplicate this

> Duplication required only for splitting too long requests. Good point for future improvement.

No, Lua strings are immutable: there is absolutely no reason to duplicate these by mallocing a second copy -- even in the case of records longer than 1460.

>> and change the behaviour of existing Lua applications that conform to the SDK 1.x guideline

> New functionality have no impact on memory usage or performance of old model with `send-sent-send`. Data will be copied only if sending already in progress and user want to send more.

You are mallocing and freeing extra string buffers. Of course your approach will impact RAM utlisation. 

> You are right, I don't understand all the aspects. But I believe that dropping data by `send` even without raising an error - is the worst way. Espressif was wrong when changed core SDK functionality and forced all users to find complicated work-arounds. 

Failing silently is not good, and we should investigate this. A failed send() should either return an error status or throw an error.

> So, I just want previous functionality back. 

My response to this is simple: stick with the old memory intensive and buggy 0.96 version or create your own fork, and this will give you what you want.  If you want 40-odd Kb of RAM and code density at least 50% higher then use a current version and accept that we have deprecated / removed some legacy functionality.  If we try to replicate old buggy behaviour, then we we just be recreating old bugs

> Would you like to help me?

Sorry, no.  I set my own priorities and continuing to improve the overall stability and performance of the firmware is my priority. 

> This still doesn't address the sk:close() issue.

Please read the SDK release notes.

TerryE:
>  I'm not enough of a Lua guru to know how to do this, but I suspect that it would be possible to \[ ... develop ... /] the wrapper code creates a new object with a metatable and arranges to forward the method calls onto the original 'sk' -- except for the 'send' method which would be handled as in your code above. *IF* this was possible, then it would be a clean solution for all of these `sk:send` problems. Your approach requires changing all the `sk:send` calls into `sk_send` which is more prone to error.

@pjsg, Philip, yes, Lua can easily be extended to create a sk class with an extended `send()` method which gives this level of emulation.  My example wasn't intended to be a perfect solution, but simply to show that it is pretty straightforward to use standard Lua functionality to achieve this behaviour without destabilising the core net module for everybody else.  I don't have a problem with Hellsy22 doing a private fork, but the core distro is another matter.

pjsg:
I agree that LUA strings are immutable -- it doesn't mean that they persist forever. With the current aggressive garbage collection, I would expect them to be effectively destroyed as soon as the value is no longer needed.

    a = 1
    func("value = " .. a)

If 'func' is implemented in C, then it should be careful what it does with the string argument. Once it returns, all bets are off as to the validity of the pointer (the GC can get rid of the temporary value). The same problem happens if you call a C function expecting a string with a numeric argument. There will be a local conversion to string on the stack, which will be discarded once the function returns.

Hellsy22:
> You are mallocing and freeing extra string buffers.

Only if sending in progress already. Otherwise data will be passed to the ```espconn_sent``` directly.

> that it trivial to modify your Lua code

Your lua solution takes 1.5-2k memory for *every* socket even with empty buffer.

> This should be a different API function, not send.

```sendq``` is a fine too.

> If we try to replicate old buggy behaviour, then we we just be recreating old bugs

We have pile of data need to be send. So, we can make a queue on firmware side. We can spend twice more memory to make a queue on lua-side. Or we can spend three times more memory to make a queue on lua side via multiple callbacks definitions.
But problem will persist anyway: because pile of data still need to be send.

> create your own fork

Sure. Just thought it can be helpful.
So about this case. I would:

1. Make all functionality disabled from code by default.
2. Choose another function name (```sendq```).
3. Use lua strings instead of allocating new buffers.
4. Do nothing about ```close``` case because have no idea what problem is (I read the SDK release notes).

TerryE:
> I agree that LUA strings are immutable -- it doesn't mean that they persist forever. With the current aggressive garbage collection, I would expect them to be effectively destroyed as soon as the value is no longer needed.

Phil, as long as the proper Lua references are maintained, the strings are used and _won't_ be GCed, as I explained to the O/P:

> Lua is going to manage the storage and GC of rec1, rec2, ... anyway. There is little point in doubling up this storage when you've got so little RAM to play with. If you think that this approach is too complicated from a Lua programmers perspective you should still stash the buffered strings in a Lua table hooked into the Lua Registry so that you don't double the memory overhead of doing this buffering.

Hellsy22:
@TerryE, I look at discussion https://github.com/nodemcu/nodemcu-firmware/issues/999, and have a question: does only one global ```send``` can be active, not one per connection?

TerryE:
AFAIK, you can only have one active connection with the SDK.

Hellsy22:
@TerryE ```net.c``` module five active sockets available for server (listener). Is it designed to be used with ```net_socket_hold``` or constrains doesnt apply to server?

marcelstoer:
Where is this going? Terry, Nick, Philip I interpret your comments as "feedback to shape this PR such that you would accept it", correct?

Kudos to you @Hellsy22 for all the effort, always great to see community members coming forward with PRs, but personally I don't see that NodeMCU would really benefit from having a `sendq()` function. So, I'd give this a thumbs down.

TerryE:
I personally feel that this falls under the more general issue that I raised at #999.  You could just as easily have a Lua extension to handle this.  I just think that this whole approach is just going in the wrong direction.  I have no objections to @Hellsy22 developing such an extension for personal use, but the crux of this extension is that we include it in our core distro, and my view is :-1: on this.

nickandrew:
I'll go with what @TerryE says, on the basis that it is easy to do the queueing in Lua as shown (and we could document that as a pattern for users) and doing it in Lua makes the buffering explicit rather than hidden within the driver.

Hellsy22:
Ok, I got the point. You want to teach entire world to use non-buffering output and workaround crutch. It's strange because even an arduino have output buffers and the project promising **a network application in Node.js style**, but it's your decision.

TerryE:
>  You want to teach an entire world to use non-buffering event driven output and workaround crutch.

@Hellsy22, this oversimplifies a quite complex situation. The SDK is what the SDK is.  Your deleted patch applied sticky tape over one small aspect, but didn't really change its fundamental nature.  As we say this could be done in Lua and provide the user with a lighter weight, more understandable and customisable implementation providing a complete API encapsulation.

We want to make an IoT chipset which is dirt cheap and hardware constrained available to be programmed in a high level language.

Hellsy22:
@TerryE Nope. Output buffering is not small aspect, it's a core feature provided in almost any language on almost any hardware. Even Arduino with 2k SRAM have it. There are more issues around the ```send``` for a last 3 month then on any other subj.

 Q: Why my http post not posting anymore?
 A: It's a send problem, RTFM.

 Q: Why my telnet server skipping strings?
 A: It's a send problem, RFTM.

 Q: Why my tcp socket send now only a first datablock, but udp is ok?
 A: It's send problem, RTFM!

IMCO, especialy for ESP, the ```net.*``` is not a small aspect - it's the core feature. You can break almost any module because pairing the ESP with an arduino via serial can solve any HW problems, but breaking the ```net.*```...

Anyway. I want to ask about upgrading http-client module with a custom callbacks and highly customized responce. Is it ok or I can even not bothering with a PR?

TerryE:
@Hellsy22, net.socket output is buffered, it's just single buffered and not elastically multi-buffered. If you want to extend the net interface in Lua to provide elastic multi-buffer buffering, then it is simple to do this using standard Lua features.  What you seem to gloss over is that a year ago we had lots of issues / use bugreps around "I only have 15Kb RAM but my application crashes when `node.heap()` reports that I have 6Kb still free" and one common reason for the 6Kb crash was that the application programmer had queued up a few sends which exhausted memory and this used to cause a messy and irrecoverable failure.  Johny and my improvements to the RTS mean that the same application now starts with 45Kb free and that the Lua code density is roughly 40% higher.  I am now trying to get GPIO stable, and next I want to move onto bit banging.  No one  pays us for this. We do this work pro-bono.

If your library was complete and consistent, then you would have a stronger case. But it's not.  If you want to get it so, then don't expect _me_ to change my priorities to help get you up to speed.  You'll just have to do this the way that Johny, I and the others did -- read what documentation is available, research the relevant posts on the Espressif BBS, esp8266.com and issue / PR threads here, _and_ go through the existing code to gain a better understanding as a whole.  It should only take you 6-12 months.

Hellsy22:
@TerryE:
You cannot worry about memory and offer the LUA solution. It's mutually exclusive ways.
Almost all memory problems with nodemcu based on attempts to do via LUA something what should be done on firmware side.

My library is consistent. It did just what described. Not more. Not less. With all (literally all) suggestions from this thread. I even accepted the idea of using lua-strings.

Do you know what I found in ```net.c``` with my totally unexperienced view? Kilobytes of bad formated, bad commented but still a very simple C code. There are no dark secrets for a long study - just elementary API. Ofc, any API have a shades you know already, but I not. And it's lead us to a few words about a collaboration. The experience exchange is strong or even main reason for the people to join any opensource community. But you suggest to read a tonns of docs you already read and spend hours or days to find something you already know (or pretend) but refuse to say. For me It's look like a polite way to say "back off, it's my personal old toy". And it's the reason why you planing to reviewing old modules alone. "Teamwork? Never heard of it."

I will not bother you or other collaborators by any questions, suggestions or request anymore.

Good luck.

TerryE:
@Hellsy22, please comment on the issue not the person,  There's a few of the regular contributors share my view.  Please see #1080 to see why allowing send queuing by default would be disastrous for system stability and this is a major factor why the OEM decided to withdraw this functionality.

Hellsy22:
@TerryE Last function name was ```sendq```, not ```send```. And even it was **disabled by default**.

Queue is an abstract layer on ```send``` and have no connection with the problems mentioned above except this one:

> If you are receiving from multiple hosts or on multiple ports, your send will go the the last port/IP address and not necessarily the one which initiated the send.

In this case doesn't matter will it be C queue, LUA queue or LUA-callback queue - problem in multiple connections, not in the queue.

TerryE:
We are going round in circles and this isn't constructive for either of us.

Issue #870, opened 2015-12-22
WiFi event monitor: Upgrade to SDK event monitor


dnc40085:
I'm looking to get some input on the API (and corresponding code) I have written to expose the event monitor added with esp_iot_sdk_v1.0.1_15_04_24 before I submit a pull request.

Here's a link to the branch: [dev_new_event_monitor](/dnc40085/nodemcu-firmware/tree/dev_new_event_monitor)

Currently, `wifi.eventmon.start()` needs to be executed once by the user to register the C callback with the SDK. I would like to have it so it automatically registers the CB on boot but I'm not exactly sure how to do that.

@TerryE pertaining to issue #846, I'm still learning about the lua internals(and programming in general), am I correct in the assumption that you are suggesting that `gL=L` should be replaced with `gL=lua_getstate()`?

# wifi.eventmon sub-module

## CONSTANTS
wifi.eventmon.STA_CONNECTED, wifi.eventmon.STA_DISCONNECTED, wifi.eventmon.STA_AUTHMODE_CHANGE, wifi.eventmon.STA_GOT_IP, wifi.eventmon.STA_DHCP_TIMEOUT, wifi.eventmon.AP_STACONNECTED, wifi.eventmon.AP_STADISCONNECTED, wifi.eventmon.AP_PROBEREQRECVED

## DISCONNECT REASON CONSTANTS
[...]

PR #997, opened 2016-02-01
Implement SDK event monitor

dnc40085:
This PR implements the SDK event monitor discussed in issue #870


dnc40085:
Hmmm, compiled fine locally...

TerryE:
Time to enable cross references and digging into the map file. :smile:

dnc40085:
I have no idea where I would even start

TerryE:
I am not sure why you should be blowing up on iram1.  Johnny has done a lot to trim this back, so the point where you shouldn't have an issue.  The mapfile is at `app/mapfile` and the content is pretty self explanatory if you've looked at linker outputs before.  You need to read this in concert with the `ld` files.

dnc40085:
I cant figure out how to reproduce this error locally, with all the default modules enabled, the compiler never spits out any errors. However, when I flash my esp-12e it does not boot.

dnc40085:
I found the issue, I was using the `vsprintf` in `app/libc/c_stdio.c` instead of the one in the ROM, which was being added to iram1.

Thank you for the pointers on where to look.

Issue #975, opened 2016-01-25
MQTT: Message Queue


efuturetoday:
Hello Dev-Team,
i really like your awesome framework and want to thank you first in place for your hard work!
I want to push sensor data to a mqtt broker in a certain period of time over and over.
Therefore i created a semi automated timer and start the timer again after i pushed the message with 
```
mqtt.client:publish("/data", json.encode(data), 0, 0, function() tmr.start(0) end)	
```
I noticed that the timer would stop working after some time - because the callback get only called when PUBACK is received. With QoS 1 the same issue.
If i start the timer over and over like this:
```
mqtt.client:publish("/data", json.encode(data), 0, 0, function() end)
tmr.start(0)
```
After some time i will face a heap - overflow, because the message is added to the message queue.
Is there a way to remove the previous message from the mqtt-queue maybe with an id returned by 
```mqtt.client:publish``` or a implement a callback that is fired when the message is removed from the message queue?

Greetings!

**Comment by devsaurus** [comment]
@efuturetoday no, there is no way to remove published messages from the queue.

You definitely have to use QoS=1 to rely on the callback as explained here https://github.com/nodemcu/nodemcu-firmware/issues/561#issuecomment-122784944. Your report seems to tell that you at least tried QoS=1 once? This should have done the job. Please post the full script, eg as a gist.

**Comment by efuturetoday** [comment]
```
CONFIG = {
    WIFI = {
        SSID = "xxxxxxxxxxx",
        PASSWORD = "xxxxxxxxxx"
    },
    MQTT = {
        SERVER = "192.168.1.156",
        PORT = 1883,
        USER = "user",
        PASSWORD = "password",
        KEEPALIVE = 10
    }
}

SCAN_CFG = {
--    ssid = CONFIG.WIFI.SSID,
    channel = 0,
    show_hidden = 1
}

SERVICES = {
    _services = {},
    _callbacks = {},
    add = function(name, type, callback)
        SERVICES._services[node.chipid()..":"..name] = {
            type = type
        }
        SERVICES._callbacks[name] = callback
    end,
    perform = function(name, args, mailbox)
        print("[MQTT] should perform "..name)
        local ok, err = pcall(function()
            assert(SERVICES._callbacks[name] ~= nil, "service not found on host")
            SERVICES._callbacks[name](args)
        end)
        local data = {}
        data.mailbox = mailbox
        if (ok) then
            data.status = "OK"
        else
            data.status = "ERROR"
[...]

PR #1002, opened 2016-02-02
Fix memory exhaustion in mqtt under circumstances from issue #975

pjsg:
The root cause of the memory exhaustion is as follows:

* The supplied example code (issue #975) publishes two messages every second.

* Initially, the first message is sent and the second message is queued. 
* The socket calls back to say that the first message has been sent.

* One second later the timer fires and sees that the socket is available and there is a queued packet, so it sends it.
* At the same time, two more messages are published, and both get queued (as there is a message outstanding).
* The socket calls back to say that the second message has been sent.

At this point there are two messages queued. The timer can get rid of one message per second, but the lua code is pumping in two messages per second. This is an ongoing getting worse situation and eventually we run out of memory.

The fix is easy -- when the socket calls back to say that the packet has been sent, check the outgoing queue to see if we should send another packet immediately. If so, send it. This is what this PR does. It also corrects the return code from :publish. 

If you are using a non-zero QoS, then things aren't quite as simple, and I'm not sure if bad things can happen.

There is a lot of duplicated code in the mqtt module -- which I didn't fix.

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -431,25 +431,43 @@ static void mqtt_socket_sent(void *arg)
     return;
   }
   NODE_DBG("sent1, queue size: %d\n", msg_size(&(mud->mqtt_state.pending_msg_q)));
+  uint8_t try_send = 1;
   // qos = 0, publish and forgot.
   msg_queue_t *node = msg_peek(&(mud->mqtt_state.pending_msg_q));
   if(node && node->msg_type == MQTT_MSG_TYPE_PUBLISH && node->publish_qos == 0) {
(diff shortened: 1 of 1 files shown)


jmattsson:
@vowstar I believe you're the resident MQTT expert - could you handle this PR please?

markfinn:
Since this is languishing I'll provide the feedback that I am using this patch and find mqtt publish much more usable with no ill effects.  

Between #985 and this, mqtt publish has gone from [lose messages after about 1 minute then leak memory until reboot] to [running fine for several hours].

devsaurus:
Thanks a lot @markfinn.

@all is there still any review/verification ongoing? I'd like to see this being merged soon, and will do so if no concerns are raised now.

PR #1008, opened 2016-02-03
BME280/BMP280 Digital Pressure Sensor module

vsky279:
## BME280/BMP280 Digital Pressure Sensor module

Simple module for BME280/BMP280 Digital Pressure Sensor module implementing initialization of the sensor, read-out & compensation calculations. The module supports setting all operating parameters of the sensor: oversampling, power mode, inactive duration.

In addition it provides 3 methods for air pressure conversion not directly connected to sensor operation: measured pressure to sea level pressure conversion, pressure to altitude and dew point calculation.

|Method|Description|
|---------------|-------------|
|init()|Initializes the module and sets the pin configuration.|
|temp()|Reads the sensor and returns the temperature in celsius as an integer multiplied with 100.|
|baro()|Reads the sensor and returns the air temperature in hectopascals as an integer multiplied with 1000.|
|humi()|Reads the sensor and returns the air relative humidity in percents as an integer multiplied with 100.|
|qfe2qnh()|For given altitude converts the air pressure to sea level air pressure.|
|altitude()|For given air pressure and sea level air pressure returns the altitude in meters as an integer multiplied with 100.|
|dewpoint()|For given temperature and relative humidity returns the dew point in celsius as an integer multiplied with 100.|

The module is properly documented in the corresponding file [https://github.com/vsky279/nodemcu-firmware/blob/bme280/docs/en/modules/bme280.md](https://github.com/vsky279/nodemcu-firmware/blob/bme280/docs/en/modules/bme280.md).


jmattsson:
This is all looking very neat and good, except for that glaring os_delay() in there. Those delays violate the SDK continuous execution time limits. The sampling function really needs to be turned into an async split-phase operation which kicks off the reading, sets a timer, and returns. When the timer fires, it collects the result and invokes a callback.

On a much smaller note, I'd prefer to see the init return `nil, 1, 2` rather than `nil, false, true` as the latter prevents using just `if bme280.init(x,y) then print("sweet!") end`.

TerryE:
Not sure how this is compiling given you are repeating the BME280_STANDBY_TIME_\*_MS definition at  lines 51:58 and 70:77.  If you want to expose constants like this to the Lua-side the you need to make them part of your `bme280_map`.  Also as Johny says, epoches of  > ½ mSec should be excuted using timers and make this async if you still want to have a functioning WiFi stack after your data retrieval calls.

nickandrew:
I checked the code and found exactly the same thing @jmattsson commented on about the delay, and came to the same conclusion. This could go into a FAQ for driver authors: unless a delay is a minimal number of us, and correct duration is critical, it needs to be implemented as a timer with callback.

vsky279:
Gentlemen, thanks for your feedbacks. I will make my homework and will come back with an updated version.

vsky279:
Please have a look on this [commit](https://github.com/vsky279/nodemcu-firmware/commit/e92a59bd7ecd1aa97b20c76a781cc08cba67df38). It should answer all issues raised:
* init() return values
* redundant BME280_STANDBY_TIME_*_MS definitions - however I don't interface these constants to Lua-side - I think it's useless (doesn't it eat some precious memory). It should be fine to use numerical constants as they are properly documented in module documentation.
* async approach to forced mode readouts.

I will create new PR tomorrow.

marcelstoer:
> I will create new PR tomorrow.

No need for that. Just keep committing to your vsky279:bme280 branch until you're satisfied. Then squash all your commits into one and force-push them to your branch. GitHub will update this PR automatically. Have a look at https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#working-with-git-and-github for more information.

vsky279:
Great. I already knew the squash&force-push process but I didn't know you can still work on your PR. I'm still a newbie to git though.

So I squashed the update and there's only one commit now.
I've added one method that starts the forced mode readouts and triggers callback when done.

|Method|Description|
|---------------|-------------|
|init()|Initializes the module and sets the pin configuration.|
|temp()|Reads the sensor and returns the temperature in celsius as an integer multiplied with 100.|
|baro()|Reads the sensor and returns the air temperature in hectopascals as an integer multiplied with 1000.|
|humi()|Reads the sensor and returns the air relative humidity in percents as an integer multiplied with 100.|
|startreadout()|Starts readout (turns the sensor into forced mode).|
|qfe2qnh()|For given altitude converts the air pressure to sea level air pressure.|
|altitude()|For given air pressure and sea level air pressure returns the altitude in meters as an integer multiplied with 100.|
|dewpoint()|For given temperature and relative humidity returns the dew point in celsius as an integer multiplied with 100.|

vsky279:
Sure. It's done.

vsky279:
The build fails on `.irom0.text' will not fit in region `irom0_0_seg'
Maybe it is due to latest PR so this one does not fit int irom0.text now?

marcelstoer:
>  The build fails on `'.irom0.text' will not fit in region 'irom0_0_seg'`.

For every PR we run an "all-modules-no-SSL"-build (https://github.com/nodemcu/nodemcu-firmware/blob/dev/.travis.yml#L27, https://github.com/nodemcu/nodemcu-firmware/blob/dev/tools/pr-build.sh) in addition to the regular CI build as a simple sanity check.

marcelstoer:

devsaurus:
The build hiccup distracted me from further review. So please check my comments regarding ICACHE_FLASH vs ICACHE_RAM balancing and this PR finally enters the home stretch :smiley:.

vsky279:
Thanks for your comments. So I removed all the `ICACHE_FLASH_ATTR` as it is redundant.
I have not used ICACHE_RAM_ATTR` as there's no rush in this callback.

Issue #710, opened 2015-10-30
dev140: Bug in SSL module


apESP8266:
Hello!

I'm testing the following NodeMCU firmware:

NodeMCU custom build by frightanic.com
	branch: dev140
	commit: 340ef8fc0ae00be1960acebe0276702dc7a00cb0
	SSL: true
	modules: node,file,gpio,wifi,net,pwm,i2c,spi,tmr,adc,uart,ow,bit,mqtt,cjson,crypto
build 	built on: 2015-10-26 19:36
powered by Lua 5.1.4 on SDK 1.4.0

After being connected to the Internet, I run the following commands:

local SMTP_SERVER = "smtp.gmail.com"
local SMTP_PORT = "465"
conn=net.createConnection(net.TCP, 0)
conn:on("reconnection", function(co, payload) print("Reconnected") end )
conn:on("disconnection", function(co, payload) print("Disconnected") end )
conn:on("receive", function(co, payload) print("received -> ") print(payload) end )
conn:on("sent", function(co, payload) print("sent -> ") end )
conn:connect(SMTP_PORT,SMTP_SERVER)
conn:dns(SMTP_SERVER,function(sck,ip) print("SMTP Server: "..ip) end)     
conn:on("connection", function(co, payload)
		print("Connected")
		local IP_ADDRESS = wifi.sta.getip()
                	conn:send("HELO "..IP_ADDRESS.."\r\n")
		end )

And I get the following results on the console:

SMTP Server: 64.233.184.109
connected
sent -> 
received -> 
  
Disconnected

Changing the third line of the commands for:
conn=net.createConnection(net.TCP, 1)

I just get:
SMTP Server: 64.233.184.109

Apparently, using the secure connection (conn = net.createConnection (Net.Tcp, 1)) none of the events is detected.
How can we deal with events in the case of secure connections?
Are we in the presence of a firmware bug in the SSL module?

Thanks in advance to those who answer these questions and help overcome the difficulty in connecting to the email service of Google.

TerryE:
duplicate of #520

i68040:
I reviewed issue #520 and the closest comment I could find to a resolution was made by marcelstoer where he said "Did you try the dev/dev140 branches which are based on the 1.4 SDK?".

It looks like this issue was created by someone using the dev140 branch who is having SSL issues. I am new to nodemcu (Just got my ESP8266 boards yesterday) and have not found a clear statement around SSL support in Lua. Are there examples that work?

[...]

PR #1014, opened 2016-02-06
Fix secure TCP connections

urish:
Call `espconn_secure_set_size()` before calling `esp_secure_connect()`, should fix issues like #710 and #520


devsaurus:
@urish great that you spotted this inconsistency in the code base - thanks again!
Please push a similar commit for mqtt.c to finalize this `espconn_secure_set_size()` thing.

urish:
@devsaurus done!

Issue #428, opened 2015-05-17
gpio.serout is not working as it described in the manual


klukonin:
I'm trying to write some IR protocol modules with gpio.serout, but it's behavior is not exactly the same as described. Looks like when pin is pulled down, it'l level is rising, instead of stay at the groung level.
This looks terrible.

gpio.serout(pin,1,{5,25},16)

Screenshot: http://s11.postimg.org/jfnnvq02b/fail.png

It is not 5 us pulses with 25 us ratio.

Oyooyo:
I believe to have fixed the bugs and slightly improved gpio.serout in general. I ran 2-3 tests with my oscilloscope connected, and the result looked correct.

I attached a patch to this comment - I'm sure I did it wrong and this is not the correct way to suggest fixes, but I'm completely new to Github etc., so please forgive me for not doing this the correct way yet. Since I have little experience with C, a more experienced C programmer should also have a short look at my proposed fix when integrating the patch, to look for any apparent errors.

BTW, it might make sense to add a comment to the documentation of gpio.serout pointing out that the delay_times array should contain an even number of values. I believe an odd number of values can generally be considered an error, and providing an odd number of values may lead to strange behaviour, as every second repetition would be inverted.
[gpio_serout_fix.txt](https://github.com/nodemcu/nodemcu-firmware/files/288274/gpio_serout_fix.txt)

marcelstoer:
> I believe to have fixed the bugs and slightly improved gpio.serout in general.

Then please provide a pull request: https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md

oyooyo:
Okay, I created a pull request. Hope I did it right, this is all still quite confusing to me:
https://github.com/nodemcu/nodemcu-firmware/pull/1327

PR #1017, opened 2016-02-07
Easygpio: module for reading bits from sensors

romanchyla:
- easygpio: generic bit-banging
- added documentation
- added short tutorial for writing a c module for nodemcu

I've taken the opportunity to provide a short tutorial. I remember how hard it was when I tried to write my first module few months back (at that time, one had to edit 4 files). And for beginners it is really a very steep hill to climb.

Please consider it as suggestion, the easygpio module doesn't have to be enabled by default. But I'm hoping it could be useful.

### DIFF
--- a/app/modules/gpioser.c
+++ b/app/modules/gpioser.c
@@ -1,0 +1,107 @@
+// Module for interfacing with GPIOs
+
+// Usage:
+// gpioser.read(PIN_CLK, PIN_OUT, num_bits, extra_ticks, delay_us, initial_val, ready_state)
+// gpioser.read(5, 6, 24, 1, 1, 0, 0)
+// -- read 24 bits, from pin 6, pin 5 is CLOCK, hold it for 1us, initial value
+// -- written to pin 5 is LOW (0), and ready state is when the pin 6 changes to LOW (0)
+
+
+#include "lualib.h"
+#include "lauxlib.h"
+#include "platform.h"
+#include "module.h"
+#include "c_types.h"
+#include "c_string.h"
+
+#define PULLUP PLATFORM_GPIO_PULLUP
+#define FLOAT PLATFORM_GPIO_FLOAT
+#define OUTPUT PLATFORM_GPIO_OUTPUT
+#define INPUT PLATFORM_GPIO_INPUT
+#define INTERRUPT PLATFORM_GPIO_INT
+#define HIGH PLATFORM_GPIO_HIGH
+#define LOW PLATFORM_GPIO_LOW
+
+
+unsigned int PIN_CLK;
+unsigned int PIN_DATA;
+
+
+long read(int pin_clk, int pin_data, int num_bits, int after_ticks,
+        int delay_us, int initial_clk_write, int ready_state) {
+    unsigned long Count;
+    unsigned char i;
+    unsigned int maxWait = 1000;
+    Count=0;
+
+    platform_gpio_write(pin_clk, initial_clk_write);
+    
+    while (platform_gpio_read(pin_data) != ready_state) {
+      maxWait--;
+      if (maxWait == 0) return -1;
+    }
+
+    for (i=0;i<num_bits;i++){
+        platform_gpio_write(pin_clk, HIGH );
+        Count=Count<<1;
+
+        os_delay_us(delay_us);
+
+        if (platform_gpio_read(pin_data) == HIGH) Count++;
+        platform_gpio_write(pin_clk, LOW );
+
+        os_delay_us(delay_us);
+    }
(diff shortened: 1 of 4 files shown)


marcelstoer:
Here's my 2 cents:
- Don't include unrelated changes in a single PR. `cmodule.md` is nice but should be in its own PR (along with the new FAQ entry).
- New module should be disabled by default in `user_modules.h` and not be inserted at the very end (modules are alphabetically ordered).
- Link to new module documentation must be added to `mkdocs.yml`, see https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md#writing-documentation.

TerryE:
My immediate reaction is that this is really the equivalent of gpio.serout, but for input.  It makes sense to combine these two sets into one module.  See #428 for some of the difficulties here.  As these are a niche use, I suspect that these would be better in a separate module -- but not "easy"gpio.  However, we are in the middle of tidying up / reworking the gpio implementation, so do you mind if we keep this on hold for a few weeks?

romanchyla:
Thank you for suggestions, I'll split the PR into two parts - I belive the cmodule.md tutorial is uncontroversial (and for beginners, it would be great to have it); I'll make a separate PR for it.

@TerryE I also thought it better to keep the functionality in a separate module (maybe also serout could be extracted?). Can I somehow help with the gpio module itself? [ie: I'd like to avoid having to track several repositories for my project which depend on the bit-banging]

romanchyla:
I've changed the PR as indicated above. If you would like me to do something else or help with the gpio, please let me know. Thank you.

app/modules/gpioser.c
```diff
@@ -0,0 +1,110 @@
...
+
+long read(int pin_clk, int pin_data, int num_bits, int after_ticks,
+        int delay_us, int initial_clk_write, int ready_state) {
+    unsigned long Count;
+    unsigned char i;
+    Count=0;
+
+    platform_gpio_write(pin_clk, initial_clk_write);
+
+    while (platform_gpio_read(pin_data) != ready_state);
```
pjsg: This can hang forever if the device doesn't respond (well, the watchdog will kill it eventually).

marcelstoer:
@romanchyla pls address Philip's comment, rebase on current `dev` to fix the merge conflicts and then push to your branch again.

romanchyla:
ok, will do - give me few days...

romanchyla:
ok, updated - it is trying 1000 times before giving up (not sure that is the right solution, but can't think of anything else&simple: is that bad/low/too arbitrary?)

pjsg:
I looked at the code, and I don't see any assignment to PIN_CLK (the global variable). 

It also isn't clear that the info() method provides much value....

docs/en/modules/gpioser.md
```diff
@@ -0,0 +1,65 @@
...
+This module can be used for reading bits from wide range of sensors (bit banging).
+
+## gpioser.read()
+Reads data from the GPIO sensors using specified intervals. 
+
+#### Syntax
+`gpioser.read(pin_clk, pin_out, num_bits, after_ticks, delay_us, initial_val, ready_state)`
+
+#### Parameters
+`pin_clk` pin number of the sensor (the CLOCK channel), type is number
```
marcelstoer: All the parameters are missing the list markers ('- ').

marcelstoer:
Thanks Roman for the update.

> this is really the equivalent of [`gpio.serout`](http://nodemcu.readthedocs.org/en/dev/en/modules/gpio/#gpioserout), but for input. It makes sense to combine these two sets into one module

So, `gpio.serout` could be transferred to this new module even if #428 isn't addressed yet. Or the other way around...The more I think about this the more I tend to the latter. It's not that `gpio.c` would grow beyond a manageable size if it also included a `serin` function.

> It also isn't clear that the info() method provides much value

I have to agree.

romanchyla:
I vaguely remember looking at the gpio.serout (seems more flexible), the
code was hard to read though. It received a table with wait times. It will
be slower but how much (and if that matters) I don't know. Will try to do
some work on it next week.

For my edification: any idea how long the gpio_read() takes? And if it's
safe to call it 1000 times?

marcelstoer:
What a pity this has been stuck in the PR queue for so long. Terry/Philip would it be unfair argue it should be closed as no one seems to have come forward in the mean time stating they really missed what this PR would provide?

TerryE:
> Terry/Philip would it be unfair argue it should be closed as no one seems to have come forward in the mean time stating they really missed what this PR would provide?

It is also pretty out of date and would need work to make it compatible with current builds.   Time to let it die a natural death.

Issue #386, opened 2015-04-25
Document module dependency matrix


marcelstoer:
It's helpful I can easily include/exclude particular modules for custom builds using [user_modules.h](https://github.com/nodemcu/nodemcu-firmware/blob/master/app/include/user_modules.h). However, this is always trial-and-error I assume because there's no documented module dependency matrix. If I exclude module X will my NodeMCU binary be broken because module Y depends on X?

This request is to maintain (in wiki) or auto-generate a module dependency matrix.

marcelstoer:
I was just wondering if that's even necessary. Could it be that all modules are, or have to be, independent of each other anyway?

vowstar:
TODO:
Build firmware dependent tree and make all modules more dependent.
@MarsTechHAN @vowstar

MarsTechHAN:
But like SSL, some modules wont use it, and SSL will consume a lot RAM... 

How about make a trick?
```C
#ifdef MQTT

#define SSL

#endif
```

And cross-test will consume some time...We will get this done in a week.

marcelstoer:
> How about make a trick?

MQTT is the only module that fails the build if SSL is disabled, see #432 (there's a note about this in my cloud build). Hence, once this is fixed the SSL setting is totally independent from other settings. IMO that's the way to go.

I'd welcome if you postulate a guideline which says that all non-core modules (e.g. MQTT, 1-wire, I²C, cJSON, etc.) must be standalone and not depend on any other non-core module. Running [makedepend](https://en.wikipedia.org/wiki/Makedepend) or the compiler with the `-M` option could provide some automatic dependency analysis.

TerryE:
Ladder logic tricks like this are pretty standard but typically you would have the primary defines as a block and the dependency defines as a second block so in this you'd have something like

```C
#if defined(MQTT) && !defined(SSL)
#  define SSL
#endif
```

confile:
+1

marcelstoer:
@TerryE @devsaurus @jmattsson @pjsg This starts to bother me more and more and I'm quite convinced that it'll bite us medium-/long-term if we don't tackle this. When I created this issue nearly a year ago we had maybe 20 modules, now there are more than 40.

Initially I thought it would be enough if we had a simple dependency matrix. However, meanwhile we have different _sorts_ of dependencies.
- classical, module A requires module B: currently #1379 looks like you'd need the new `tls` module for net, http and mqtt if you need secure connections
[...]

PR #1018, opened 2016-02-07
Expose SDK event monitor

dnc40085:
This PR exposes the SDK event monitor discussed in issue #870 
`wifi.sta.eventMonxxx` code was moved to `wifi_eventmon.c`
changed `wifi.sta.eventMonReg()`: callbacks can now be unregistered by leaving function field blank.
changed `wifi.sta.eventMonStop()`: all callbacks can now be unregistered by using `1` instead of `"unreg all"`
All backwards compatibility was preserved. The event monitor documentation can be found [here](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/docs/en/modules/wifi.md#wifieventmon-module).

Also included is the addition of the functions [`wifi.ap.deauth()`(click here for documentation)](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/docs/en/modules/wifi.md#wifiapdeauth) and [`wifi.sta.getrssi()`(click here for documentation)](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/docs/en/modules/wifi.md#wifistagetrssi)

### DIFF
--- a/app/include/rom.h
+++ b/app/include/rom.h
@@ -129,8 +129,6 @@ void ets_delay_us (uint32_t us);
 
 int ets_printf(const char *format, ...)  __attribute__ ((format (printf, 1, 2)));
 
-int ets_sprintf(char *str, const char *format, ...)  __attribute__ ((format (printf, 2, 3)));
-
 void ets_str2macaddr (uint8_t *dst, const char *str);
 
 void ets_timer_disarm (ETSTimer *a);
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -107,6 +107,10 @@ extern void luaL_assertfail(const char *file, int line, const char *message);
 
 //#define WIFI_SMART_ENABLE
 
+#define WIFI_STATION_STATUS_MONITOR_ENABLE
+#define WIFI_SDK_EVENT_MONITOR_ENABLE
+#define WIFI_EVENT_MONITOR_DISCONNECT_REASON_LIST_ENABLE
+
 #define STRBUF_DEFAULT_INCREMENT 32
 
 #endif	/* __USER_CONFIG_H__ */
--- a/app/libc/c_stdio.h
+++ b/app/libc/c_stdio.h
@@ -63,7 +63,7 @@ extern void output_redirect(const char *str);
 int c_sprintf(char* s,const char *fmt, ...);
 #endif
 
-// #define c_vsprintf ets_vsprintf
+#define c_vsprintf ets_vsprintf
 #define c_printf(...) do {					\
 	unsigned char __print_buf[BUFSIZ];		\
 	c_sprintf(__print_buf, __VA_ARGS__);	\
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -10,6 +10,8 @@
 
 #include "c_types.h"
 #include "user_interface.h"
+#include "wifi_common.h"
+
 
 #ifdef WIFI_SMART_ENABLE
 #include "smart.h"
@@ -20,25 +22,17 @@ static int wifi_smart_succeed = LUA_NOREF;
 
 static uint8 getap_output_format=0;
 
+#define INVALID_MAC_STR "MAC:FF:FF:FF:FF:FF:FF"
+
 //wifi.sleep variables
 #define FPM_SLEEP_MAX_TIME 0xFFFFFFF
 static bool FLAG_wifi_force_sleep_enabled=0;
 
-
-//variables for wifi event monitor
-static sint32_t wifi_status_cb_ref[6] = {LUA_NOREF,LUA_NOREF,LUA_NOREF,LUA_NOREF,LUA_NOREF,LUA_NOREF};
-static os_timer_t wifi_sta_status_timer;
-static uint8 prev_wifi_status=0;
-
-
 #ifdef WIFI_SMART_ENABLE
-#if defined( NODE_SMART_OLDSTYLE )
-#else
-static lua_State* smart_L = NULL;
-#endif
 static void wifi_smart_succeed_cb(sc_status status, void *pdata){
   NODE_DBG("wifi_smart_succeed_cb is called.\n");
 
+  lua_State* L = lua_getstate();
   if (status == SC_STATUS_LINK_OVER)
   {
     smartconfig_stop();
@@ -68,14 +62,13 @@ static void wifi_smart_succeed_cb(sc_status status, void *pdata){
 
   if(wifi_smart_succeed != LUA_NOREF)
   {
-    lua_rawgeti(smart_L, LUA_REGISTRYINDEX, wifi_smart_succeed);
+    lua_rawgeti(L, LUA_REGISTRYINDEX, wifi_smart_succeed);
 
-    lua_pushstring(smart_L, sta_conf->ssid); 
-    lua_pushstring(smart_L, sta_conf->password); 
-    lua_call(smart_L, 2, 0);
+    lua_pushstring(L, sta_conf->ssid);
+    lua_pushstring(L, sta_conf->password);
+    lua_call(L, 2, 0);
 
-    luaL_unref(smart_L, LUA_REGISTRYINDEX, wifi_smart_succeed);
-    wifi_smart_succeed = LUA_NOREF;
+    unregister_lua_cb(L, &wifi_smart_succeed);
   }
 
 #endif // defined( NODE_SMART_OLDSTYLE )
@@ -83,7 +76,6 @@ static void wifi_smart_succeed_cb(sc_status status, void *pdata){
 #endif // WIFI_SMART_ENABLE
 
 static int wifi_scan_succeed = LUA_NOREF;
-static lua_State* gL = NULL;
 /**
   * @brief  Wifi ap scan over callback to display.
   * @param  arg: contain the aps information
@@ -92,19 +84,20 @@ static lua_State* gL = NULL;
   */
 static void wifi_scan_done(void *arg, STATUS status)
 {
+  lua_State* L = lua_getstate();
   uint8 ssid[33];
-  char temp[128];
+  char temp[sizeof("11:22:33:44:55:66")];
   if(wifi_scan_succeed == LUA_NOREF)
     return;
   if(arg == NULL)
     return;
   
-  lua_rawgeti(gL, LUA_REGISTRYINDEX, wifi_scan_succeed);
+  lua_rawgeti(L, LUA_REGISTRYINDEX, wifi_scan_succeed);
 
   if (status == OK)
   {
     struct bss_info *bss_link = (struct bss_info *)arg;
-    lua_newtable( gL );
+    lua_newtable( L );
 
     while (bss_link != NULL)
     {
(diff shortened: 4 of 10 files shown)


pjsg:
Does the SDK event monitor callback execute at interrupt level or at base level? I would hope at base level.....

dnc40085:
@pjsg Honestly, I have no idea how to find that out and the SDK documentation doesn't provide that much information, just the function wifi_set_event_handler_cb(), a very short description and some example code.

I thought rewording the commit would replace the original, oh well...

marcelstoer:
> I thought rewording the commit would replace the original, oh well...

If that's your intention then you need to squash/amend your commits and then force push them (`git push --force`). Force pushing will erase the comments based on code lines in the previous commit, though. The comments in this conversion will stay as they're not part of the repository.

dnc40085:
I also discovered the undocumented function `wifi_softap_deauth` which goes perfectly with the event monitor so unauthorized stations can be deauth when they are connecting.

I have written an [implementation](https://github.com/dnc40085/nodemcu-firmware/blob/dev_wifi_ap_deauth/app/modules/wifi.c#L968) for the function, should I add it to this pull request?

app/include/rom.h
```diff
@@ -131,6 +132,8 @@ int ets_printf(const char *format, ...)  __attribute__ ((format (printf, 1, 2)))
 
 int ets_sprintf(char *str, const char *format, ...)  __attribute__ ((format (printf, 2, 3)));
 
+int ets_vsprintf (char *d, const char *s, va_list ap);
```
jmattsson: This is not part of the boot ROM. It's provided by the SDK libs. This definition (and the old `ets_sprintf()` one above) should go into `sdk-overrides/include/ets_sys.h`.

jmattsson:
With the new `wifi.eventmon.register()`, does one still have to call `wifi.sta.eventMonStart()` ? What about if I'm only interested in monitoring AP events?

app/include/rom.h
dnc40085: Thank you for clarifying that, I wasn't sure where to put that.

dnc40085:
The functions `wifi.eventmon.register()` and `wifi.sta.eventMonReg()` are completely separate and can be used independently from one another and even disabled in `user_config.h`. 

The `wifi.sta.eventmon___()` functions use a timer to check `wifi_station_get_connect_status()` every 150ms and when the returned value changes the function executes a user registered callback.
The purpose of the function `wifi.eventMonStart()` is to arm the timer.

The function `wifi.eventmon.register()` registers the user's callback function so it can later be retrieved when the SDK calls the event handler `wifi_handle_event_cb`, which first checks for a user callback. Then, if found, it uses switch/case to determine which event was returned by the SDK then assembles the available information into a table and passes it to the user's callback function.

dnc40085:
Event monitor now copies event data into a queue and posts a task(using new tasking interface) to process the queue

dnc40085:
I'm guessing this failure is related to #1070

marcelstoer:

dnc40085:
@marcelstoer Thank you, much appreciated.

@pjsg I was looking at the disassembled code for the the function `wifi_set_event_handler_cb` and I [found a call](https://gist.github.com/dnc40085/df31f74caadda00e7073#file-wifi_set_event_handler_cb-asm-L14) to the function `ets_timer_handler_isr`, so I'm inclined to say that the callback is executed at interrupt level rather than base level.
Since you posted your question I have changed the event monitor implementation to utilize the new tasking interface by @TerryE. 
The implementation I have written uses two separate functions to handle the event. The first function [`wifi_handle_event_cb`](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/app/modules/wifi_eventmon.c#L148)(registered with `wifi_set_event_handler_cb`) checks if there is a Lua callback registered for the corresponding event, if found, it then copies the event data into a queue(a linked list) then posts a low priority task to the second function, [`process_event_queue`](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/app/modules/wifi_eventmon.c#L175) which then processes the first event in the queue, forwards the data to the Lua callback, clears the processed event from the queue and lastly, if any events are remaining in the queue, it posts another task.

TerryE:
No need for it to go into RAM :)

app/include/user_modules.h
```diff
@@ -47,6 +47,7 @@
 #define LUA_USE_MODULES_UART
 //#define LUA_USE_MODULES_UCG
 #define LUA_USE_MODULES_WIFI
+  #define LUA_USE_MODULES_WIFI_EVENTMON
```
devsaurus: Why the indent - did you intend to disable by default?

dnc40085: No, I indented it to suggest that it is a submodule of `wifi`

marcelstoer:
Not sure I fully understand this PR.

If I'm not mistaken the new module depends on the WiFi module, right? Then we have a issue with #386. What happens if you enable WiFi eventmon but not WiFi in `user_modules.h`? Also, we have a single `.md` file in the `docs` folder *per module*. This PR violates this.

I hope you can easily dispel my concerns.

dnc40085:
It's not really a separate module, the event monitor code is just in a separate file but still effectively part of the `wifi` module and resides in `wifi.eventmon`. 

I separated the code from `wifi.c` following @TerryE suggestion that since `wifi.c` was getting a bit large, the event monitor code should be [split off into a wifi folder](https://github.com/nodemcu/nodemcu-firmware/issues/870#issuecomment-166703173). I tried to move the files into a new folder `app/modules/wifi` but I couldn't figure out how to get it to compile, since my makefile-fu is poor, so I left the files in `app/modules`.

As for the question about the entries in `user_modules.h`, if the `wifi` module is disabled and the event monitor is not, the event monitor code will be [excluded](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/app/modules/wifi_eventmon.c#L19) from the build, and the build will complete successfully devoid of all wifi and event monitor code. if it's an issue I could move it to `user_config.h`.

With the documentation, since the wifi event monitor code is part of the `wifi` module, should it stay in `wifi.md` or should it be split off into a separate file since `wifi.md` is getting big?

I have to remember to start writing more details in my PRs.

marcelstoer:
> It's not really a separate module, the event monitor code is just in a separate file but still effectively part of the wifi module

That's what I thought and that's why I was so surprised to see a separate entry in `user_modules.h`.
You know I maintain nodemcu-build.com so I'm definitely biased when I look at PRs. How do you suggest such a "half module" should be integrated there? Also, if you can enable WiFi but disable WiFi eventmon it'd be odd to look at the WiFi module docs and see functions there that won't work even though you WiFi enabled.

Still confused...

TerryE:
@dnc40085. Let me do a proper review today, before we consider committing this. I can sort out this file split stuff and any make magic needed.  One thing that I do notice is that you are breaking the convention  for namespaces in C: all public names should be prefixed with the the module name / namespace (e.g. `wifi_` in this case) or optionally `_wifi_` if "private" between wifi files. static (that is internal to file) entries can be named whatever you want but should avoid using another namespace prefix. So having static routines called lua_xxxx is a bad idea, but having extern ones is a no-no.

TerryE:
> Also, if you can enable WiFi but disable WiFi eventmon it'd be odd to look at the WiFi module docs and see functions there that won't work even though you WiFi enabled.

@marcelstoer  we already have this type of anomaly with SSL, SHA2, etc.  Sometimes documented functionality only works when the build is properly configured.  It's a compromise that we have made in the past when functionality takes a large footprint and the some builders might wish to disable if they are trying to configure  a build for 512kb Flash for example.

dnc40085:
>you are breaking the convention for namespaces in C: all public names should be prefixed with the the module name / namespace ...having static routines called lua_xxxx is a bad idea, but having extern ones is a no-no.

@TerryE I didn't know about the namespace conventions, thanks for letting me know. As for extern, I previously thought that it was letting the compiler know that the function/variable was defined elsewhere, but after removing them from `wifi_common.h` I found that extern is only required for `wifi_eventmon_map[]`.
I added the functions `lua_table_add_xxx()` and `lua_cb_xxx()` in `wifi_common.h` partly because of the [DRY principal](https://en.wikipedia.org/wiki/Don't_repeat_yourself) but mainly to save a few bytes and when I was choosing the names I was only thinking in the scope of the `wifi` module not the whole project.

Would you like me to add a commit to fix these improper coding practices before you do your review?
 
>It's a compromise that we have made in the past when functionality takes a large footprint and the some builders might wish to disable if they are trying to configure a build for 512kb Flash for example. 

@TerryE For this reason I've been thinking of opening an issue suggesting that we make smart config (wifi.xxxsmart()) optional, considering that there must be others that don't require this feature and would much rather prefer having the 26k of extra flash to use for extra modules

Sorry for my bad coding practices. I'm always reading stuff to try to improve.

Also, t hanks for the suggestion on getting Pil 3rd ed., it's already clarified some stuff for me. So far I'm on page 40.
Any suggestions for a book on C?

TerryE:
> Would you like me to add a commit to fix these problems before you do your review?

What I was thinking  about going is just to do a PR into _your_  fork for you to look at and consider.  It is s entirely up to you if you want to accept this or do the rework yourself.  But it's sunny outside in the UK so I might just have some hours outside in the fresh air before I start :smile:

dnc40085:
> It is's entirely up to you if you want to accept this or do the rework yourself

If you take the time to do a PR I'm not going to take credit for your work, but I will learn from it.

On my local branch I've already:
- Removed all the unnecessary externs
- Renamed `lua_table_add_xxx()` to `table_add_xxx()`
- Renamed `lua_cb_xxregister()` to `xxregister_lua_cb()`
- Renamed `event_task_id` to `wifi_event_monitor_task_id`
- Renamed `event_queue_xxx` to `wifi_event_queue_xxx`
- Renamed `process_event_queue` to `wifi_event_monitor_process_event_queue`
- Renamed `wifi_handle_event_cb` to `wifi_event_monitor_handle_event_cb`
- Renamed `event_monitor_reason_map` to `wifi_event_monitor_reason_map`
- Renamed `wifi_eventmon_map` to wifi_event_monitor_map

TerryE:
@dnc40085, sorry but I haven't had a chance to do this today.

TerryE:
I've made some comments on your fork.  As I said: all externs in the wifi namespace should be prefixed with ẁiif_`.  Als be vary when moving code out of line as this might in fact generate more code than you are saving.  It's always instructive looking at the generated code if the routine is only a few lines long.  If you move the code to a header as a static inline then gcc treats this as a macro so you get the source clarity without any runtime overhead.  Worth considering.

TerryE:
@dnc40085 I'll do a top to tail review of this this weekend and if OK then I'll merge it. Sorry for the delay.

dnc40085:
No worries, I'm in no rush.

I was wondering... locally, I was reworking `wifi.sta.config` to be simpler(use a stack pointer) and I found a bug where station can't connect to an open AP and I wrote a fix, should I add another commit to this or should I submit a separate PR?

TerryE:
> Also, is [`LUA_USE_MODULES_WIFI_EVENTMON`](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/app/include/user_modules.h#L55) OK in `user_modules.h` or should I do away with it and just use the [entries](https://github.com/dnc40085/nodemcu-firmware/blob/dev_new_event_monitor/app/include/user_config.h#L98) I already have in `user_config.h`?

As far as the application programmer is concerned this is just a config option within `wifi`, so I would not have the extra `LUA_USE`  but stick with the enable option in the config file.

marcelstoer:
@dnc40085 We don't have an agreement as for #1146 but this PR is not (yet?) on that list, it may not make the cut. So, I feel it'd be helpful if you could create a dedicated PR for the `wifi.NULLMODE` fix. That one could then be merged rather sooner than later independently from this one. The relevant bug report #1034 has already been closed a while ago which is confusing.

TerryE:
Perhaps we should apply a 90:10 test "Is this a material step forward over the current code in master/dev?".  I am always worried that we contributors are all time limited and if we apply a 100% rule then we'll just end up never moving forward.  

On this basis I would merge this now, but not without at least one other contributor +1.

dnc40085:
So what should I do? split it off?

TerryE:
@dnc40085 

> I took a look at the disassembled code for the the function `wifi_set_event_handler_cb` and I [found a call](https://gist.github.com/dnc40085/df31f74caadda00e7073#file-wifi_set_event_handler_cb-asm-L14) to the function `ets_timer_handler_isr`, so I'm inclined to say that the callback is executed at interrupt level rather than base level.

If you look gist then you will see that this s offset 0xCC which is in `ets_timer_init()` so I am inclined to differ. :smile: 

Nope IMO leave it as it is now now.  90:10 rules.

jmattsson:
This PR has been lingering for some time now even after all the updates. What are the arguments against merging this? Is it the "half-module" that Marcel brought up? If that's the only thing I'm in favour of merging now and improving later.

dnc40085:
If anybody has any questions about the PR please feel free to ask. 
Suggestions for changes are also welcome.

dnc40085:
@TerryE 
>If you look gist then you will see that this is offset by 0xCC which is in `ets_timer_init()` so I am inclined to differ. :smile:

You're probably right, my understanding of assembler is limited.

TerryE:
@dnc40085 are you happy for us to merge this?

dnc40085:
I have no more changes to make to this PR, Aside from not being sure if I should remove the pseudo [future proofing section](https://github.com/nodemcu/nodemcu-firmware/pull/1018/files#diff-7396aef6e0a69a3cca8825fb77305aedR196) (If Espressif adds any new events, this section allows the user to assign a default callback to execute if an event specific callback is not defined).

Should I squash the commits before the merge?

jmattsson:
Yes please., might as well keep it neat and tidy!

dnc40085:
The commits have been squashed.

jmattsson:
Thanks! Sorry about the delay to get this merged.

dnc40085:
No problem. Thanks for merging this.

PR #1021, opened 2016-02-07
Added full support for DS3231

Takkuz:
Added many new functions for full supporting DS3231 RTC.

### DIFF
--- a/lua_modules/ds3231/ds3231.EN.md
+++ b/lua_modules/ds3231/ds3231.EN.md
@@ -40,7 +40,7 @@ package.loaded["ds3231"]=nil
 ## setTime()
 ####Description
 Sets the current date and time.  
-if _disableOscillator_ is set to 1 the oscillator will **stops** on battery.
+if _disableOscillator_ is set to 1 the oscillator will **stop** on battery.
 
 ####Syntax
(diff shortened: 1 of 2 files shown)


pastukhov:
Great!

nickandrew:
I have no way to test this, but it looks like a nice piece of work.

Takkuz:
Thank you @nickandrew... I've fixed those errors.

theopensourcerer:
Could this also be used to wake the ESP8266 from Deep Sleep? I am thinking about using one of these devices as I want to use a cycle time longer than ~70mins I can configure directly in the ESP?

Takkuz:
@theopensourcerer I tried, but the NodeMCU needs a pulse on GPIO0. When it's in deep sleep if you set GPIO0 to LOW it wakes up but doesn't boot until setted to HIGH again. DS3231 sets the pin to LOW until reloaded, so it keeps the NodeMCU in a state of useless coma. You need a phisical circuit that cut the link after a while. I found it but I can't remember it's name.

theopensourcerer:
@Takkuz Thanks for the reply. 

So it seems as though technically it might be possible but perhaps some kind of Monostable is required in between... i will do some more investigations.

PR #1023, opened 2016-02-08
Addition of new task IF with GPIO rework to supprt this

TerryE:
The following files have been updated to change the Flash version of the pin map into a 32 packed structure. This simplifies the map declaration, but this is copied to the existing the RAM vector format to minimise coding impact on this patch iteration. We might switch to the structure version in RAM in future.

-  `app/platform/pin_map.h`
-  `app/platform/pin_map.c`

The platform interface has been updated:

-   To move ISR code out of Flash and into RAM to comply with SDK guidelines and for optimal execution speed. (Note that you need to disable inlining optimization for any such routines.)

-   The ISR bit-test loop has been peephole optimised to shorten the maximum ISR  time.

-   To split our Lua land from the Hardware abstraction layer. The hardware knows nothing about Lua or the Lua runtime environent, so the platform modules now make no reference to Lua. A new vector, pin_trigger, is used to flag if a callback has been attached to a GPIO pin and in which case the ISR posts a task to dispatch this. The registering of any Lua callbacks, and their unregistering is carried out within the Lua libraries.

-   As a result of this platform_gpio_mode() no longer unreferences any lua callback, so this must be done in the calling layer.

The Lua Library module gpio.c has also been updated to use the new task and updated platform laver.  Since I've been working these routines, I have also tidied up their implementation to use the standard Lua approach to argument manipulation, and this has slightly reduced the code size as well as making
the implementation a lot more readable.

This PR also includes the previously discussed task interface and changes to the UART driver to support this.

I have also added a draft Extension Developer FAQ which describes how the SDK implements ISRs and tasks.


TerryE:
> I have also added a draft Extension Developer FAQ which describes how the SDK implements ISRs and tasks.

This is a rough cut for comment, so I haven't added it to the yml file yet.

docs/en/extn-developer-FAQ.md
```diff
@@ -0,0 +1,93 @@
...
+    functions. ISRs can be defined on a range of priorities, where a higher
+    priority ISR is able to interrupt a lower priority one. ISRs are time
+    critical and should complete in no more than 50 µSec.
+  
+    ISR code and data constants should be run out of RAM or ROM, as the
+    execution time for Flash memory (that is located in the `irom0` load section)
+    is indeterminate: whilst cache-hits can run at full memory bandwidth, any
+    cache-misses require the code to be read from Flash; and even though
+    H/W-based, this is at roughly 26x slower than memory bandwidth (for DIO
+    flash); this will cause ISR execution to fall outside the required time
```
pjsg: As I understand, the real problem is not that it is slower, but that it (potentially) interferes with loading code out of flash that happens at the base (non-interrupt) level. This makes it absolutely wrong, rather than -- you will probably violate some timing constraint..

TerryE:
@pjsg, it's both.  I mention the Flash issue later and this needs hoisting up here, and yes this one is sudden death if a Flash-based ISR fires when anything else is doing Flash I/O, but the latency issue can cause probs with the Wifi stack.  I still got this delivering 5 trigger tasks per mSec which isn't bad.  My aim here is to remove the instability probs.  We can discuss functional enhancements as a second pass.

app/platform/platform.c
```diff
@@ -135,41 +159,62 @@ int platform_gpio_read( unsigned pin )
...
-}
+  int i;
+  gpio_task_handle = gpio_task;
 
-int platform_gpio_intr_init( unsigned pin, GPIO_INT_TYPE type )
+  get_pin_map();
+  ETS_GPIO_INTR_ATTACH(platform_gpio_intr_dispatcher, NULL);
+}
+/*
+ * Initialise GPIO interrupt mode. In RAM because interrupts are dsabled
```
pjsg: Is this really the case that code with interrupts masked must be run out of RAM? The i2c code doesn't do this. Given the small amount of space that we have in the float build, I'd like to make sure one way or the other.

pjsg:
I've moved my new drivers over to this model and will give it a workout over the next week.

Thanks for all the effort that you have put into this....

TerryE:
It's marginal.The interrupt masked code -- so long as it only masks a GPIO interrupt and not all interrupts -- will not cause a Flash conflict, and these sequences are typically only 100 bytes or so long and therefore the worst fetch time will be of the order of 2.6K clocks or 30 uSec so should be able to be run out of Flash. We need to play with this. That's why I added the extra config define to make this optional.

pjsg:
See the ticket #1025 that I just filed -- if this gets resolved, then my concerns about the IRAM usage goes way down.....

nickandrew:
Looking at the "git show --stat 9c76604" ...

```
 app/Makefile                  |   2 +
 app/driver/uart.c             |  12 ++---
 app/include/driver/uart.h     |   2 +-
 app/include/task/task.h       |  33 ++++++++++++
 app/modules/gpio.c            | 147 ++++++++++++++++++++++++------------------------------
 app/modules/node.c            |   3 +-
 app/platform/common.c         |  76 ----------------------------
 app/platform/pin_map.c        |  88 ++++++++++++++++++--------------
 app/platform/pin_map.h        |  12 +++--
 app/platform/platform.c       | 169 +++++++++++++++++++++++++++++++++++++++-----------------------
 app/platform/platform.h       |  38 +++++++++-----
 app/task/Makefile             |  44 ++++++++++++++++
 app/task/task.c               |  71 ++++++++++++++++++++++++++
 app/user/user_main.c          |  77 ++++++++--------------------
 docs/en/extn-developer-FAQ.md |  93 ++++++++++++++++++++++++++++++++++
 15 files changed, 532 insertions(+), 335 deletions(-)
```

This is a lot to try to comprehend at once. Is it possible for you to restructure it into a sequence of logical commits, for easier review?

TerryE:
Nick, I am damned if I do and damned if I don't! :laughing: I was just following Marcel's recommended one commit policy.  

The problem is that this is a complicated change and interlinked, so it's quite difficult to split it down into smaller chunks.  This is why this took me so long to develop and debug. I have included a long explanation in the commit.  Having got this working, I could now retro-engineer it back into 3 or 4 separate commit chunks, but I would really prefer to spend my time fixing other stuff on my buttlist.  I just don't want to make a habit of this.  Perhaps you and @marcelstoer could come to some concensus here, first before I do this?

-   The `*/task/*` files are new and implement the new task interface as explained in the new `md` file.
-   The `pin_map.*` have major rework as described.
-   The `user_main.c`, `node.c` and `uart.*` files have small changes to implement the new tasking I/F.
-   The `platform.*`and `gpio.c` files have a lot of rework to implement the safe gpio using the tasking I/F
-   The one extra bit that I could have dropped is that I also did some optimisations to allow some dumb out-of-line tests to be inlined by the compiler and moved these implementations from `common.c` to `static inline` declarations instead of `extern` declarations in `platform.h`.

nickandrew:
Squashing commits make sense when said commits contain mistakes which are fixed by later commits, or backtracking out bad ideas. Nobody wants to see that sausage being made :grin: But wide-ranging changes to the way things are done can often be broken into easier to digest chunks which, when all put together, form a great work of art. I have done some Linux kernel development, where patches often run to dozens of commits, and the community is very, very serious about ensuring each commit contains one atomic change and vice-versa. They have to find errors through bisection and it really needs to work at each commit.

I'd hoped you might have a pre-squash branch you could resurrect easily, but no mind. I'll work with what's here.

Since the task files are new, I'll start there, and then work my way toward the rest.

marcelstoer:
> Marcel's recommended one commit policy.

It's not mine, it's ours I hope :wink: - after all it's in our `CONTRIBUTING.md`.

> @marcelstoer could come to some concensus here

By all means, do what you feel is the most sensible approach. I am a big fan of a clean repository history that's true but we should not spend our scarce (free)time enforcing and following policies that simply cannot be applied in all cases.

app/driver/uart.c
```diff
@@ -27,8 +28,8 @@
 
 
 // For event signalling
-static uint8 task = USER_TASK_PRIO_MAX;
-static os_signal_t sig;
+// static uint8 task = USER_TASK_PRIO_MAX;
```
jmattsson: Remove rather than comment out?

app/driver/uart.c
```diff
@@ -27,8 +28,8 @@
...
+static os_signal_t sig = 0;
```
jmattsson: Change type to `task_handle_t` for consistency?

app/modules/gpio.c
```diff
@@ -87,59 +76,55 @@ static int lgpio_trig( lua_State* L )
 // Lua: mode( pin, mode, pullup )
 static int lgpio_mode( lua_State* L )
 {
-  unsigned mode, pullup = FLOAT;
-  unsigned pin;
+  unsigned pin = luaL_checkinteger( L, 1 );
+  unsigned mode = luaL_checkinteger( L, 2 );
+  unsigned pullup = luaL_optinteger( L, 3, FLOAT );
+  
+  luaL_argcheck(L, pin< NUM_GPIO && (mode!=INTERRUPT || pin>0), 1, "Invalid pin");
```
jmattsson: Any good reason why you're not using `platform_gpio_exists(pin)` (or the `MOD_CHECK_ID(gpio,pin)` macro for it) here? GPIOs aren't necessarily contiguous (see ESP32 pinout for example).

app/modules/node.c
```diff
@@ -305,6 +305,7 @@ static int node_key( lua_State* L )
 #endif
 
 extern lua_Load gLoad;
+extern bool user_process_input(bool force);
```
jmattsson: I'd prefer to see this in a header file tbh.

app/platform/platform.c
```diff
@@ -135,41 +159,62 @@ int platform_gpio_read( unsigned pin )
...
-  if (pin >= NUM_GPIO)
-    return -1;
-  ETS_GPIO_INTR_DISABLE();
-  //clear interrupt status
-  GPIO_REG_WRITE(GPIO_STATUS_W1TC_ADDRESS, BIT(pin_num[pin]));
-  pin_int_type[pin] = type;
-  //enable interrupt
-  gpio_pin_intr_state_set(GPIO_ID_PIN(pin_num[pin]), type);
-  ETS_GPIO_INTR_ENABLE();
+  if (pin < NUM_GPIO) {
```
jmattsson: `platform_gpio_exists(pin)` instead?

app/task/task.c
```diff
@@ -0,0 +1,71 @@
...
+ */ 
+LOCAL os_event_t *task_Q[3];
+LOCAL task_callback_t *task_func;
+LOCAL int task_count;
+ 
+LOCAL void task_dispatch (os_event_t *e) {
+  task_handle_t handle = e->sig;
+  if ( (handle & TASK_HANDLE_MASK) == TASK_HANDLE_MONIKER) {
+    uint16 entry    = (handle & TASK_HANDLE_UNMASK) >> TASK_HANDLE_SHIFT;
+    uint8  priority = handle & 3;
```
jmattsson: Perhaps turn the 3 into a TASK_HANDLE_PRIO_MASK? And have a TASK_HANDLE_PRIO_MAX for the 3 on the next line?

docs/en/extn-developer-FAQ.md
```diff
@@ -0,0 +1,93 @@
...
+model works based on the Espressif-supplied examples and SDK documentation, plus
+various posts on the Espressif BBS and other forums, and an examination of the
+BootROM code.
+
+The ESP8266 boot ROM contains a set of primitive tasking and dispatch functions
+which are also used by the SDK. In this model, execution units are either:
+
+-   **INTERRUPT SERVICE ROUTINES (ISRs)** which are declared and controlled
+    through the `ets_isr_attach()` and other `ets_isr_*` and `ets_intr_*`
+    functions. ISRs can be defined on a range of priorities, where a higher
```
jmattsson: Side note: Examination of the core-isa.h header suggests it's a range of 2.

docs/en/extn-developer-FAQ.md
```diff
@@ -0,0 +1,93 @@
...
+    is indeterminate: whilst cache-hits can run at full memory bandwidth, any
+    cache-misses require the code to be read from Flash; and even though
+    H/W-based, this is at roughly 26x slower than memory bandwidth (for DIO
+    flash); this will cause ISR execution to fall outside the required time
+    limits. (Note that any time critical code within normal execution and that
+    is bracketed by interrupt lock / unlock guards should also follow this 50
+    µSec guideline.)
+
+-   **TASKS**. A task is a normal execution unit running at a non-interrupt priority.
+    Tasks can the executed from Flash memory. An executing task can be interrupted
```
jmattsson: *Tasks can the...*

Romanes eunt domus?

docs/en/extn-developer-FAQ.md
```diff
@@ -0,0 +1,93 @@
...
+message for delivery to a task handler. Each task handler has a unique message
+associated with it, and may bind a single uint32 parameter. How this parameter
+is interpreted is left to the task poster and task handler to coordinate.
+  
+The interface is exposed through `#include "task/task.h"` and involves two API
+calls. Any task handlers are declared, typically in the module_init function by
+assigning `task_get_id(some_task_callback)` to a (typically globally) accessible
+handle variable, say `XXX_callback_handle`. This can then be used in an ISR or
+normal LLT to execute a `task_post_YYY(XXX_callback_handle,param)` where YYY is
+one iof `low`, `medium`, `high`. The callback will then be executed when the SDK
```
jmattsson: s/iof/of/

jmattsson:
Aside above inline comments, :+1:  from me (untested, because #time).

docs/en/extn-developer-FAQ.md
TerryE: It's sad that I am trying to work out of the should write  "The Romans are the house". Maybe I should get a life or be called Brian. But the penny did drop thanks to Google. :laughing:

app/modules/gpio.c
TerryE: I'll swap all of these out.  If you note, these are now inlined anyway so the compiler will emit the same compiled instruction sequence, but your suggestion is a better abstraction at a source level.

pjsg:
Can you add a recommendation for users of this interface that interrupt handlers should be cautious about using too many slots (as there may be other interrupt handlers that are also using the same queue). It may be advantageous for handlers to do their on queuing and then post a message to have the task level code do all the callbacks. With care, this can limit the number of posted tasks to one per interrupt handler, and so the chances of overflow are significantly reduced.

Otherwise :+1: from me as it works well with my two new drivers.....

TerryE:
Phillip, the issue here isn't the number of interrupt handles that have been declared but the number of posts that can be safely executed in any given task without causing a chain reaction or avalanche.  So longas the average is one or less, then queue build up should be an issue. 

But you are right, this merits some extra explanation.

app/modules/node.c
TerryE: So would I, but this should logically be `user_main.h` which doesn't exist and I don't want to start adding this include file in this commit.  I added the extern declaration here because it wasn't previously declared as external at all, and ths practice is used elsewhere in the firmware for such odd externs.  A job for another day.

app/platform/platform.c
TerryE: The `NO_INTR_CODE` macro defaults to `inline`.  It is only in RAM if the new define `GPIO_SAFE_NO_INTR_ENABLE` is uncommented out in `user_config.h`

docs/en/extn-developer-FAQ.md
TerryE: I think it is 2+NMI = 3 but my point here is that it is more than 1.  Interrupts can be interrupted.

TerryE:
I've just screwed up this update.  I am too tired now.  Ignore this, and I'll do it properly tomorrow as a new PR.

jmattsson:
You can force-push onto the branch and github will sort it out and keep the PR going (as well as provide historical references). It's quite handy that way.

TerryE:
> You can force-push onto the branch and github will sort it out and keep the PR going (as well as provide historical references). It's quite handy that way.

Nope, I can't open it because the "Reopen and comment" button is greyed out because I have force pushed onto the branch.  So see PR  #1061

PR #1037, opened 2016-02-11
New chunk capable crypto hash functions

ABonner:
Hi, per #982, I've implemented a first pass at exposing the underlying hashing primitives more directly, in a way that allows for adding arbitrary data to the hash over time.

The basic use case is if you are receiving data over time which all needs to be hashed, or need to process more data than can fit in memory at one time. In either case, the current `crypto.hash()` falls short as it can only hash what is in a single Lua string.

The new `crypto.new_hash("SHA256")` function returns a userdatum that has some object-oriented functions associated with it:

    x = crypto.new_hash("SHA256")
    x:update("Some content")
    x:update("More content")
    digestString = x:finalize()

This is my first contribution to the project so if I have missed anything in the contributor guidelines or you see somewhere I can improve the PR please let me know. I've added documentation in as well - I wasn't sure how best to show the `update` and `finalize` functions since they aren't part of the module per se.

### DIFF
--- a/app/modules/crypto.c
+++ b/app/modules/crypto.c
@@ -105,11 +105,9 @@ static int crypto_mask( lua_State* L )
   return 1;
 }
 
-
 static inline int bad_mech (lua_State *L) { return luaL_error (L, "unknown hash mech"); }
 static inline int bad_mem  (lua_State *L) { return luaL_error (L, "insufficient memory"); }
 static inline int bad_file (lua_State *L) { return luaL_error (L, "file does not exist"); }
-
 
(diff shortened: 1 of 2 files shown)


ABonner:
Thanks for the feedback @pjsg . I'll make updates to these points tonight and push an additional commit.

jmattsson:
:+1:

ABonner:
I believe I've rebased this correctly to resolve the issue without adding additional commits. Does it look good to merge?

TerryE:
I guess that it's my turn to sort the merge conflicts :smile:

Issue #632, opened 2015-09-03
cross-lua build error


eyaleb:
Now that https://github.com/nodemcu/nodemcu-firmware/issues/598 is committed I decided to do a build of master.

I used the command

``` sh
$ lua tools/cross-lua.lua
The program 'lua' can be found in the following packages:
 * lua5.2
 * lua5.1
 * lua50
Try: sudo apt-get install <selected package>
```

I need lua installed... so did so

``` sh
$ sudo apt-get install lua5.2
$ lua tools/cross-lua.lua
lua: ./tools/build.lua:5: module 'lfs' not found:
        no field package.preload['lfs']
        no file '/usr/local/share/lua/5.2/lfs.lua'
        no file '/usr/local/share/lua/5.2/lfs/init.lua'
        no file '/usr/local/lib/lua/5.2/lfs.lua'
        no file '/usr/local/lib/lua/5.2/lfs/init.lua'
        no file '/usr/share/lua/5.2/lfs.lua'
        no file '/usr/share/lua/5.2/lfs/init.lua'
        no file './lfs.lua'
        no file '/usr/local/lib/lua/5.2/lfs.so'
        no file '/usr/lib/x86_64-linux-gnu/lua/5.2/lfs.so'
        no file '/usr/lib/lua/5.2/lfs.so'
        no file '/usr/local/lib/lua/5.2/loadall.so'
        no file './lfs.so'
stack traceback:
        [C]: in function 'require'
        ./tools/build.lua:5: in main chunk
        [C]: in function 'require'
        tools/cross-lua.lua:2: in main chunk
        [C]: in ?
```

Now I had to guess.. tried this:

``` sh
$ sudo apt-get install lua-filesystem
$ lua tools/cross-lua.lua
[COMPILE] .build/cross-lua/app__lua__lapi.o
[builder] Error building target
[builder] Last executed command was: 
gcc -O2 -Iapp/include -Iinclude -Iapp/lua -Wall -DLUA_CROSS_COMPILER -O2 -c -MD -MF .build/cross-lua/app__lua__lapi.d app/lua/lapi.c -o .build/cross-lua/app__lua__lapi.o
[...]

PR #1040, opened 2016-02-12
Direct port of the struct module 

pjsg:
This is a direct port of the [struct module](http://www.inf.puc-rio.br/~roberto/struct/) that provides a way to get fields from within strings. This is very useful when dealing with network protocols etc.

The only changes that I made were

* Change the registration of the module to use the NodeMCU flavour
* If the NodeMCU build is integer-only, then disable the float and double options.

The documentation is the original documentation (converted to markdown). 

While testing this using the original testcases, I found that the floating point version of NodeMCU has problems with exponentiation. In particular   2^8 ~= 256    On my linux box, 2^8 is exactly 256. The integral version of NodeMCU also has 2^8 == 256


app/modules/struct.c
```diff
@@ -0,0 +1,422 @@
...
+
+
+#include "lua.h"
+#include "lauxlib.h"
+
+
+#if (LUA_VERSION_NUM >= 502)
+
+#define luaL_register(L,n,f)	luaL_newlib(L,f)
+
```
jmattsson: This #if and the #define are unused.

app/modules/struct.c
```diff
@@ -0,0 +1,422 @@
...
+#endif
+
+typedef STRUCT_INT Inttype;
+
+/* corresponding unsigned version */
+typedef unsigned STRUCT_INT Uinttype;
+
+
+/* maximum size (in bytes) for integral types */
+#define MAXINTSIZE	32
```
jmattsson: Surely this should be 4?

jmattsson:
The docs don't match the format we use for the API documentation and will break the autogeneration of ToC etc. Please see existing .md files for reference.

Apart from the above comments I'd be happy to get this in.

pjsg:
I've updated the docs to align more closely with the nodemcu format. I also fixed the MAXINTSIZE issue and made it depend on the type of build. In a float build, you can have (in theory) very large integers, though I don't know of any applications. I wanted to leave as much alone as possible to make porting any lua app easy.

TerryE:
>  In a float build, you can have (in theory) very large integers.
Phillip, there's a gotcha here with the nodeMCU implementation of strtoul: it only works with 32 bit integers so assuming that you can safely use integers bigger than uint32 can be problematic. See #632.

TerryE:
> While testing this using the original testcases, I found that the floating point version of NodeMCU has problems with exponentiation. In particular 2^8 ~= 256 On my linux box, 2^8 is exactly 256. The integral version of NodeMCU also has 2^8 == 256

The FP version of the ^ operator uses exponentiation: `x^y = exp(y*log(x))` and this can generate low-bit errors leading to the behaviour that you observe.  I don't think that  it's worth doing a special case test and evaluation for (small) integer `y` values.

pjsg:
With regard to the the MAXINTSIZE of 32 inthe float build. This is the memory size of the integer (i.e. in the range 0 .. 2^256). This is only meaningful for float builds. I'm not sure that this size is *useful* but it is what the original code used, so I left it.

dnc40085:
Is it possible to unpack userdata with this?

pjsg:
What do you mean by "userdata"?

dnc40085:
I am referring to the Lua data type "userdata" pushed on to the stack with the function [lua_newuserdata()](http://pgl.yoyo.org/luai/i/lua_newuserdata).

TerryE:
`unpack` currently takes a string argument.

jmattsson:
Userdata by definition can not be unpacked. What the userdata refers to may be a C pointer, it may be a C value cast to a pointer (if a light userdata), it may be a pointer-to-pointer-to-pointer-to-pointer for all we know. It'd be as safe as allowing direct reads & writes to any memory address.  If a userdata wants to have its raw contents accessible, it will provide a metatable with an accessor to do so.

dnc40085:
I see, I was looking at the possibility of sending a struct from an Arduino to the ESP8266 and unpacking the data into a Lua table for use and vice versa.
I guess I need to read more before posting such questions.

TerryE:
> I am still trying to understand Lua and metatables are confusing. 

You can get PiL V2 online as an eBook.  But PiL v3 is pretty cheep and well worth buying. The link is in my FAQ.  I am always willing to give 1-1 advice / feedback to active firmware contributor via email or Skype.  Just PM me at esp6266.com with your email details if you want to hook up, as this sort of dialogue is outside the scope of this issues list. :smile:

TerryE:
@pjsg, Philip is this now ready for merge.  Have any other committers got any comments?

jmattsson:
:+1:

pjsg:
I have no more changes to make to this..... so :+1: from me!

Issue #851, opened 2015-12-14
Lua reentrancy, tasks, callbacks and interrupt callbacks.


TerryE:
This issue was highlighted to me by #795 and #802.  We have Lua instabilities manifesting themselves when the ESP is being hammered by GPIO or OW interrupts.  How it manifests itself is with Lua stack corruptions which cause assert trips if lua_assert diagnostics are available and by a messy death if not.

There is a separate issue #846, which might be related but I think not.  Because this stuff is terribly documented within the Espressif documentation, I am largely relying on wider experience and retroengineering of the behaviour of the SDK.  But if anyone else can provide better information, then please add to this discussion.

*   Lua can either be build to be reentrant or not.  In the case of the eLua implementations it is assumed that the RTS is non -rentrant so the `lua_lock()` and `lua_unlock()` are optimised away.  We don't have the option to enable them because the SDK is essentially non-preemptive and any lock collision would simply result in deadlock.

*    The SDK uses two primary mechanisms for executing applications code: tasks and callbacks.  Though in many ways tasks are simply a queued callback mechanism.

*    All task and most callback functions are non-preemptive, that is once a function is initiated, it will run to completion and return control to the SDK -- which in turn can then schedule the next runnable task / cb.  The guideline is that no cb should take longer than 10 mSec, otherwise some functions such as net and wifi will start to fail.

*   `user_init()` starts the main Lua thread and posts a task to run the  `init.lua`.

*   Any code chunks input through the UART are also posted as tasks.

*   Unlike standard Lua, the firmware never closes the main thread, but leaves it open when tasks complete, and hence globals, the thread's registry and dangling upvalues remain in scope.

*   Callbacks (such as a `tmr.alarm()` callback) are executed using standard "calling Lua from C" API features and pick up a Lua thread (not necessarily the main thread -- see #846).

*    Some callbacks seem to be classed as "interrupt drivers", and it isn't clear to me if they are handled within the non-pre-emptive queueing system or are true interrupt routines.  For example I am thinking specifically of cb routines bound by the `ETS_XXXX_INTR_ATTACH()` macros defined in `SDK//include/ets_sys.h` and execute the `ets_isr_attach()` function.  Now these seems to be true interrupt routines.

[...]

PR #1041, opened 2016-02-12
Simple low level performance monitoring tool as a module

pjsg:
This module allows a developer to explore the performance of the NodeMCU platform. It allows a range of memory to be monitored and then a histogram is built with sampled values of the program counter. The number of bins and the range can be selected, and if the application can be run multiple times, then a very detailed profile can be constructed.

I will agree that this is not the sort of module that many people will use, but I'm sure that there are other people out there who would appreciate it. 

It is (naturally) disabled by default.


devsaurus:
How about moving the `hw_timer` code into a common driver / platform component? So it can be shared with upcoming modules (e.g. with one that I'm currently working on :wink:).

TerryE:
Arnim, Phillip, @nickandrew and I were discussing the possibility of using one of the HW timers for a generic bit banging module functionally along the lines of the gpio serout or the proposed easygpio module, but using one of the FRC timers.  However at the moment there are only two timers: FRC1 is used by the GPIO module and one by the `rtctime_internal.h` file (which isn't a header, BTW) which i used in the RTC module and written by one of Johhny's colleague at DiUS, Bernd Meyer.  Need to discuss this further.  At a minimum both of these, and any other timer code should a config define to allow the developer to choose which timer is user where.

jmattsson:
It'd be more accurate to say that FRC2 is used internally by the SDK actually - the rtctime stuff only reads it (and prays the SDK hasn't changed its mind on how to configure the FRC2, or for that matter, that someone has decided to call `system_timer_reinit()` which *does* impact the FRC2). To my understanding FRC2 should be considered off-limits for apps other than as a read-only source of time progression, and then only with the caveats mentioned here and in the `rtctime_internal.h` file.

devsaurus:
> nickandrew and I were discussing the possibility of using one of the HW timers for a generic bit banging module functionally along the lines of the gpio serout or the proposed easygpio module, but using one of the FRC timers.

I couldn't find any related posts yet to catch up with that discussion, any pointers? The [pwm also occupies the HW timer `FRC1`](https://github.com/nodemcu/nodemcu-firmware/issues/851#issuecomment-164941398) and might collide with this module and potentially others. It's not clear to me atm how this single resource shall be shared.

TerryE:
Sorry :telephone_receiver: or at least Skype. I need to post a summary as its own issue.

As Johnny has clarified FRC2 is used by the SDK, and this [Espressif topic](http://bbs.espressif.com/viewtopic.php?t=1353) clarifies that FRC2 is reserved for the SDK to implement the system software timer.  FRC1 is used for the PWM driver, but if this library is not used then the FRC1 is available for other application uses.  Clearly we can use it for only one purpose so PWM, bit-banging, Phillip's monitor, or whatever, but IMO we should hoist the FRC1 initialisation to a platform function, so that the platform module can carry out this interlock.  If all firmware modules that wish to use the FRC1 do this `platform_HWtimer_init()` lazily (that is on first use or a specific application API call, then the developer can make this call at runtime.

pjsg:
I'd love to have a general hw_timer approach -- and it makes sense to put this in its own driver.

@TerryE Do you want to create an issue where the API can be discussed?

docs/en/modules/perf.md
```diff
@@ -0,0 +1,61 @@
...
+
+####Returns
+- `total` The total number of samples captured in this run
+- `outside` The number of samples that were outside the histogram range
+- `histogram` The histogram represented as a table indexed by address where the value is the number of samples. The address is the lowest address for the bin.
+
+### Example
+
+    perf.start(0x40200000, 0x40300000, 1024)
+
```
hvegh: How would you determine for example where to start and end for a particular module?

pjsg: You have to use the mapfile that is generated by the build. If you want to find a hotspot, you can try a large range (say from 0x40000000 to 0x40400000) and then do the test and look at the results. This will show you (roughly) where the time is being spent and then you can run again just over that range...

pjsg:
I redid this and used a userdata object rather than using alloc/free.

jmattsson:
@pjsg would it make sense to default the start/end to 0x40000000 / _flash_used_end, and leave it as a (really) advanced feature to narrow the window down further?

pjsg:
@jmattsson Yes -- good idea. I've implemented that. Once we get to the end of the review comments, I'll squash the second commit....

pjsg:
I've just realized that I never tested this in a float build. It *shouldn't* make a difference, but you never know....

devsaurus:
Gave it a try with some of my u8g code and think it's quite useful for profiling. I propose to merge this soon if there aren't any further findings.

PR #1049, opened 2016-02-13
File exists

jfollas:
I found myself using `if file.list()["filename"] then` in order to check for the existence of files, which is a silly waste of processing for that purpose. This PR adds a `file.exists(filename)` function, with the approach borrowed from the Arduino project.


jfollas:
whoops! didn't mean to do that against master.

TerryE:
Ignoring your baseline whoopsie and focusing on your 44786a2, I must admit the I've always been quite happy using a  `file.list()["filename"]` check on the assumption that it isn't particularly burdensome for the number of files in a typical ESP SPIFFS. However on reflection the main performance hit of the `STIFFS_readdir()` enumeration over the `STIFFS_readdir()` is the EGC cost of building and then immediately discarding the temporary array used to return the `list()`, so it dos make sense to make this small addition. 

I find it just a lot easier to do a `git fetch nodemcu dev:my_new_PR` (or whatever you want to call it) the cherry-pick across your changes and then push it back to your repo -- than buggering about with rebaselining. :smile:

PR #1050, opened 2016-02-13
Added file.exists() method 

jfollas:
I found myself using `if file.list()["filename"] then` in order to check for the existence of files, which is a silly waste of processing for that purpose. This PR adds a `file.exists(filename)` function, with the approach borrowed from the Arduino project.

### DIFF
--- a/app/modules/file.c
+++ b/app/modules/file.c
@@ -114,6 +114,22 @@ static int file_seek (lua_State *L)
   return 1;
 }
 
+// Lua: exists(filename)
+static int file_exists( lua_State* L )
+{
+  size_t len;
+  const char *fname = luaL_checklstring( L, 1, &len );
+  if( len > FS_NAME_MAX_LENGTH )
+    return luaL_error(L, "filename too long");
+
+  spiffs_stat stat;
+  int rc = SPIFFS_stat(&fs, (char *)fname, &stat);
+
+  lua_pushboolean(L, (rc == SPIFFS_OK ? 1 : 0));
+
+  return 1;
+}
+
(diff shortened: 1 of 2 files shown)


TerryE:
See my comment on #1049.

jfollas:
Thanks Terry! I was following the pattern already used in that file, but I adopted your suggestions in a follow-up commit (changing the other places to use `luaL_argcheck` too).

The changes in behavior:

Before:
```
> =file.exists("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
stdin:1: filename too long
stack traceback:
        [C]: in function 'exists'
        stdin:1: in main chunk
```
After:
```
> =file.exists("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
stdin:1: bad argument #1 to 'exists' (filename too long)
stack traceback:
        [C]: in function 'exists'
        stdin:1: in main chunk
```

TerryE:
@jfollas, at least you get a stack traceback now! :laughing: We only added this back 6 months ago.  I agree that the error has changed, but for the better, IMO -- especially when your API call has more than one argument.

jmattsson:
:+1:

A nice-to-have might be to wrap all those "filename too long" into its own function, but I'm not sure whether that would have a good or bad impact on code size.

TerryE:
The `lua_argcheck()` is 18 bytes of code. Not worth out of lining as the procedure would be ~ 28+18 bytes and the call 8 bytes so we'd only get a size saving with >5 file length checks.

Issue #1031, opened 2016-02-10
SPIFFS image flashing


DesperateProgrammer:
Hello,

I am currently trying to generate a SPIFFS image to flash with the nodemcu-firmware using the spiffy ( https://github.com/xlfe/spiffy ) tool with spiffs 0.3.2 ( https://github.com/pellepl/spiffs/releases/tag/0.3.2 ) as this seems to be the version to be used by nodemcu.
This generates an image, that can only errornous read by nodemcu. Meaning: all filenames begin with \001 and a wrong filesize. (usually way too big)

I then moved to another spiffy branch within https://github.com/CHERTS/esp8266-devkit/tree/master/Espressif/examples/ESP8266/nodemcu-spiffy 
with the same problem.

To check the alignment (because I am not sure the error is named "me" ;-) ) 
The firmware 0x10000.bin file is 0x71318 Bytes long after an clean build. Meaning that the spiffs should be located at 0x84000 = ((0x71318 +0x3FFF) & 0xFFFFC000

Is the spiffs version 0.3.2 correct for nodemcu?
Is the alignment correct?
Is there any modification I did not catch against 0.3.2 spiffs?
How is your opinion about fs generating in the Makefile, should this be added? I would implement a patch allowing to build the fs (either as option or default) once i figure out whats failing right now

Greets

DesperateProgrammer:
Caught at least one difference to spiffs 0.3.2:
spiffs_nucleus.h in nodemcu does not contain the same manual alignment then 0.3.2

https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/spiffs/spiffs_nucleus.h#L419
vs.
https://github.com/pellepl/spiffs/blob/0.3.2/src/spiffs_nucleus.h#L422

jfollas:
After a tip from @jmattsson, I had success using https://github.com/DiUS/spiffsimg instead of Spiffy.

TerryE:
@jfollas, hummn -- you comment is apt. We should include a link to `spiffsimage` somewhere in our build instructions or FAQ. For my to do list.

DesperateProgrammer:
@jfollas, thank you for the hint. Indeed, this did work.

@TerryE, i am currently making efforts to include spiffsimg as a submodule and the makefiles to generate the fs automatically, if there are files in a files folder.
I'll commit it when i'm done. 
[...]

PR #1051, opened 2016-02-14
Automatic filesystem generation

DesperateProgrammer:
This commit adds an automatic spiffs generation on build.
Using the submodule spiffsimg, the files in the folder `local/fs/` will be put in a spiffs image file and placed into the `bin` folder by the address it needs to be flashed to.

The size of the created filesystem and the location of where the files are taken from can be configured by the Makefile variables FSSIZE and FSSOURCE.

The current values are 192kB and local/fs.

Created in response to [https://github.com/nodemcu/nodemcu-firmware/issues/1031]

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tools/spiffsimg"]
+	path = tools/spiffsimg
+	url = https://github.com/DiUS/spiffsimg.git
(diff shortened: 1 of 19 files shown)


DesperateProgrammer:
Closed for fixing and later resubmit

PR #1054, opened 2016-02-14
Initial version of the hw_timer as part of the platform

pjsg:
This is the base version of the platform_hw_timer code. I have verified that it works with my performance monitoring module. Once this gets merged, I'll squash the remaining commits in the performance branch to get that clean.

Not clear what to do about documentation. In most cases in the existing code there isn't much internal documentation, but I don't want to make things worse!


pjsg:
I'm going to combine with the PWM fixes.

PR #1057, opened 2016-02-15
Add real hw_timer support and convert the pwm module to use it

pjsg:
This adds the hw_timer.c that was published by espressif. More importantly it adds arbitration between modules that want to use the hw_timer feature. 

This also includes a fixed version of driver/pwm.c that makes use of this. I also fixed some bugs in the
pwm module that should make it work more smoothly and more glitch free. The original code had some mysterious locking that would only do anything on a multicore processor. Even the ESP32 only has a single core running user code. 

It now does use the NMI handler as that provides more predictable timing -- especially in the presence of network traffic.

There are still (unresolvable) inconsistencies when the duty cycles of two pwm outputs are very close to each other. If the edges need to be within a few microseconds, then one of the edges tends to get moved by a couple of microseconds. With a modulation frequency of 1kHz, then this is much less than 1% error.

There may still be bugs, and I'll have a go at fixing them if anybody can produce a reproducible setup.

### DIFF
--- a/app/platform/hw_timer.c
+++ b/app/platform/hw_timer.c
@@ -1,0 +1,183 @@
+/******************************************************************************
+* Copyright 2013-2014 Espressif Systems (Wuxi)
+*
+* FileName: hw_timer.c
+*
+* Description: hw_timer driver
+*
+* Modification history:
+*     2014/5/1, v1.0 create this file.
+*     
+* Adapted for NodeMCU 2016
+* 
+* The owner parameter should be a unique value per module using this API
+* It could be a pointer to a bit of data or code 
+* e.g.   #define OWNER    ((os_param_t) module_init)   
+* where module_init is a function. For builtin modules, it might be 
+* a small numeric value that is known not to clash.
+*******************************************************************************/
+#include "ets_sys.h"
+#include "os_type.h"
+#include "osapi.h"
+
+#include "hw_timer.h"
+
+#define FRC1_ENABLE_TIMER  BIT7
+#define FRC1_AUTO_LOAD  BIT6
+
+//TIMER PREDIVIDED MODE
+typedef enum {
+    DIVIDED_BY_1 = 0,		//timer clock
+    DIVIDED_BY_16 = 4,	//divided by 16
+    DIVIDED_BY_256 = 8,	//divided by 256
+} TIMER_PREDIVIDED_MODE;
+
+typedef enum {			//timer interrupt mode
+    TM_LEVEL_INT = 1,	// level interrupt
+    TM_EDGE_INT   = 0,	//edge interrupt
+} TIMER_INT_MODE;
+
+static os_param_t the_owner;
+static os_param_t callback_arg;
+static void (* user_hw_timer_cb)(os_param_t);
+
+#define VERIFY_OWNER(owner)  if (owner != the_owner) { if (the_owner) { return 0; } the_owner = owner; }
+
+/******************************************************************************
+* FunctionName : platform_hw_timer_arm_ticks
+* Description  : set a trigger timer delay for this timer.
+* Parameters   : os_param_t owner
+*                uint32 ticks :
+* Returns      : true if it worked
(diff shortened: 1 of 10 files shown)


TerryE:
Phillip, this is going to clash with the tasking/gpio PR. Do you mind if we merge this first?

pjsg:
Ah yes. If I was smarter, I'd have based it on your branch!

app/platform/hw_timer.c
```diff
@@ -0,0 +1,182 @@
...
+    DIVIDED_BY_16 = 4,	//divided by 16
+    DIVIDED_BY_256 = 8,	//divided by 256
+} TIMER_PREDIVIDED_MODE;
+
+typedef enum {			//timer interrupt mode
+    TM_LEVEL_INT = 1,	// level interrupt
+    TM_EDGE_INT   = 0,	//edge interrupt
+} TIMER_INT_MODE;
+
+static os_param_t the_owner;
```
devsaurus: Can't find any initialization of `the_owner` before evaluating it for the first time. Is it done somewhere else?

pjsg: There is implicit initialization of statics to zero at start time (this is the .bss section). If you initialize a static, then the initialisation value goes into the .data segment and this needs to be copied into RAM at start time. 

In my early days of programming (in the '70s) you learned to squeeze every byte out of the code -- and this habit of mine with regard to not explicitly initialising static variables to zero dates from when I started to use C in the '80s.

TerryE: IIRC, zeroing BSS is part of the (GC) C standards.

jmattsson: Correct, that's standard behaviour. And whether you leave a static "uninitialised" or explicitly zero-initialised makes no difference for the behaviour (i.e. an explicit zero doesn't suddenly end up in the data_init section).

pjsg:
I rebased my branch on the new dev. Actually, there weren't any collisions and it went cleanly. I also verified that it still works!

eku:
@pjsg Should this work out of the box by just enabling the PWM module as usual? Nothing has been added to the documentation, so I guess it should But it doesn't. ```pwm.setup(6,1000,0)``` resets NodeMCU without any warning..

pjsg:
@eku Hmm.... That works for me... Can you provide a complete code sample? Is there any other setup that you need to do to make this happen?

Also, can you get the return value of `node.bootreason()` ?

Thanks

eku:
@pjsg here we go, a fresh build from the HEAD of the dev branch
```
NodeMCU 1.5.1 build 20160228 powered by Lua 5.1.4 on SDK 1.5.1(e67da894)
lua: cannot open init.lua
> pwm.setup(6,1000,0)
...reboot...
NodeMCU 1.5.1 build 20160228 powered by Lua 5.1.4 on SDK 1.5.1(e67da894)
lua: cannot open init.lua
> =node.bootreason()
2       2       28      1076166482      0       0       0       0
```

It's a ESP8266-12 with 4M of flash.

pjsg:
Can you attach the mapfile from the build to this ticket? I want to see 
where address 1076166482 (0x4024FF52) is. It is doing a null load.....

Was this an integer or float build?

Philip

eku:
[map.zip](https://github.com/nodemcu/nodemcu-firmware/files/151267/map.zip)

pjsg:
Your build crashed in platform_gpio_mode (in platform.c). 

I built the code with your config and did a float build. The size of platform_gpio_mode was the same as your mapfile indicated. The offset inside platform gpio_mode of your crash pointed to a l32i.n instruction (this does a load of 32 bits and it was loading at address 0 -- hence the trap).

The good news is that mine crashed in the same way. It turns out that this will happen if you build without LUA_USE_MODULES_GPIO.

I'm going to submit a bug for this, and do a pull request to fix it. In the short term, you can work around this by adding LUA_USE_MODULES_GPIO.

PR #1060, opened 2016-02-17
Dev: added an initial working version of the circular buffer.

hvegh:
Comments are welcome. 

Here's a link to the API documentation:
https://github.com/hvegh/nodemcu-firmware/blob/dev/docs/en/modules/cirbuf.md


TerryE:
Have you thought of a companion module which is just a wrap-around byte array of variable length records (up to 255 bytes if you wish) with a similar API but you have `push` and `pushf` methods.  This could be really useful for general debugging?

TerryE:
:+1: on the gerenal idea, but just like my last PR you've done something weird with 70cb01b. This is nothing to do with this PR. You need to remove this.

hvegh:
Merged the relevant patches, but I don't seem be able to get rid of  70cb01b ...

marcelstoer:
Once merged pls update https://github.com/nodemcu/nodemcu-firmware/wiki/Module-list

hvegh:
Fixed stuff in #1062

PR #1063, opened 2016-02-18
Add file.fscfg() command for fs location + size.

jmattsson:
With the apparent interest of flashing pre-built filesystems together with NodeMCU, this command might save some hassle working out where to flash said filesystem to.

Simply boot NodeMCU, run `file.fscfg()` and you'll see precisely where this particular NodeMCU firmware is expecting the filesystem to sit.

### DIFF
--- a/app/modules/file.c
+++ b/app/modules/file.c
@@ -307,6 +307,13 @@ static int file_writeline( lua_State* L )
   return 1;
 }
 
+static int file_fscfg (lua_State *L)
+{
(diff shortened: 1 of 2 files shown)


devsaurus:
:+1: like this. No guesswork anymore.

TerryE:
My other thought was simply to add these two parameters to `file.fsinfo()`, but given that this is only a few lines of code, this is a 50:50 call.

Issue #1058, opened 2016-02-16
Rotary Switch module


pjsg:
This is a proposal to add a module that can decode a standard rotary switch -- typically found in automotive audio systems. These are available at all the standard online stores for a [buck or so](http://www.aliexpress.com/item/360-Degree-20-Detents-5Pin-Digital-Rotary-Encoder-Momentary-Switch/32219946924.html). They have a quadrature output on two pins and a momentary press button.

These switches can be connected using three GPIO pins. 

The approach is to enhance the GPIO module to allow other modules to handle gpio interrupts as well. Then this module will register and handle the interrupts for these three pins. The pin values are then converted into the rotary (or press) motion and the absolute values are maintained by the module. Changes are then sent to a callback via the new task mechanism.

dvv:
+1

I would however notice interrupt-driven mechanical buttons may require good debouncing.
FYI, here I use customizable polling https://github.com/dvv/nodemcu-thingies/blob/master/buttons.lua

pjsg:
Indeed -- debouncing the press button is important. For the quadrature outputs it doesn't matter so much as each bounce just moves the position one click rather rapidly. The standard switches have 20 (or 16) detents and there is a full 4 quadrature advances between each detent.

pjsg:
This is the interface: [rotary.md](https://github.com/pjsg/nodemcu-firmware/blob/rotary-switec/docs/en/modules/rotary.md)

pastukhov:
Please add to documetation some links to a compatibile hardware (alexpress, ebay, seedstudio)

dvv:
Wonder if `PRESS`/`LONGPRESS` etc. logic deserved to be put in the module.

mrkale:
And there is the module located?

pjsg:
The module does not yet have a pull request as it is waiting for another PR #1023 to be merged.

I like the idea of `LONGPRESS` and I'll add that before I submit the PR. What do you think is the time boundary between short and long press?  I'm thinking somewhere between 0.5 and 1 second.

 I suspect that the following would happen (in terms of the callbacks):

On the initial press: `PRESS`
After the time internal: `LONGPRESS`
When released: `RELEASE`



I'll also update the docs with links to the hardware.

dvv:
I believe your times right. May be options to setup (to vary between channels)?
[...]

PR #1067, opened 2016-02-20
Module to handle rotary decoders

pjsg:
This is the PR that goes with the discussion in #1058 

This makes use of the new task infrastructure to send events from interrupt land down to the lua level. I ended up using software timers for the long delays (the 'long' press time and the double click time).

It works for me reliably. It is not difficult to overflow the queue, but this ends up just losing `TURN` events -- the absolute position is still tracked. 

I added a simple hooking mechanism to the GPIO interrupt handler so that other modules can register and get interrupts passed to them. This module is the first user of that feature.


marcelstoer:

Pls add the module to https://github.com/nodemcu/nodemcu-firmware/wiki/Module-list once this PR is merged.

docs/en/modules/rotary.md
```diff
@@ -0,0 +1,129 @@
...
+
+## rotary.on()
+Sets a callback on specific events.
+
+#### Syntax
+`rotary.on(channel, eventtype[, callback])`
+
+#### Parameters
+- `channel` The rotary module supports three switches. The channel is either 0, 1 or 2.
+- `eventtype` This defines the type of event being registered. This is the logical or of one or more of `PRESS`, `RELEASE` or `TURN`.
```
ABonner: Also `CLICK` and `DBLCLICK`?

dfeinzeig: bump

pjsg:
I back-merged dev (and fixed the conflicts). I also squashed down into a single commit.

jmattsson:
I'm going to leave this for @TerryE to sign off on, seeing as he's fully across the GPIO interrupts at the moment :)

Looks nifty though!

TerryE:
Sorry Philip, I have just pulled your module and will go through the code before either feeding back comments or doing the merge.

TerryE:
Philip my bad and my apologies for not chipping in sooner.  I left this to you in the earlier discussions, since I was fighting other fires, but the issue of the GPIO platform changes is one of wider scope.  if you want, I can push a version to my Repo containing my suggested changes so we can discuss,

pjsg:
I'll certainly take a look at your branch....

TerryE:
Philip, you have design authority over the rotary module. However,  the GPIO driver interface is  matter of consensus, so we need to be quorate on whatever approach we finally agree on.

TerryE:
@pjsg Philip, here is a gist of the diff,  I haven't bothered to push it up as a new branch.  This is as about lightwieght as I can get it at runtime but it still adds ~500 bytes of code, so I've also added an extra config option to suppress this if you don't need it. I am overdue bed.

pjsg:
I don't see a gist.....

TerryE:
Sorry [here](https://gist.github.com/TerryE/742b93ae58c3d9f40737)

TerryE:
And thinking about it using 2 mallocs is daft. Doing one of n * (sizeof( void *)+sizeof(uint32)) makes the code simpler and its size smaller.

pjsg:
I fixed it to use a single malloc in the register function. I also added a way to update the set of bits for a hook -- this allows you to do an atomic update of moving from one set of bits to another.

It still works!!

pjsg:
The reason to be able to change the registered bits is when a driver wants to hook more bits -- for example, if the user wants to open a second rotary switch. You can't unregister the first switch and then re-register both (as you may lose interrupts). You can't do this under a disabled interrupt (memory allocation in this state is bad). You don't want to register the same hook multiple times. 

I'll reorganize the code a bit.....

pjsg:
The code does have a check that a valid combination of LUA_USE_MODULES_ROTARY and the two GPIO interrupts are set. The code gets compiled anyway, and doesn't cause any link problems.

As the number of modules grows, I wonder if it is worth adding big #ifdef LUA_USE_MODULES_blah / #endif surrounding each of the source files? This might speed up the build process (or at least keep it constant).

TerryE:
Philip, I'll merge this because whilst we may decided to evolve this through a couple more cycles, this is a stable enough baseline to moved to dev.

Issue #609, opened 2015-08-21
Default use of EGC_ALWAYS in nodeMCU Emergency GC


TerryE:
Another "note to self", but posted here just in case anyone else is interested in nonMCU throughput. 

The nodeMVU firmware inherits the [Emergency GC](http://www.eluaproject.net/doc/v0.9/en_elua_egc.html) implementation from eLua with a few tweaks.  `lua.c` sets the ECG mode to always (see the previous link for details), which essentially means that the ECG is executed before every malloc.  This is a really costly overhead for each malloc and some Lua activities (like compiling Lua source) do a *lot* of mallocing.  I just wonder if we should expose the `legc_set_mode()` function through a `node.set_ecgmode()` API call.

pjsg:
I would like to see this as an option. It would remove the need to recompile the firmware to change the mode....

TerryE:
OK, I have implemented this in node adding the `node.setEGCmode(modes[, memlimit])` function and four `node.EGC` constants: `EGC_NOT_ACTIVE`, `EGC_ON_ALLOC_FAILURE` , `EGC_ON_MEM_LIMIT`,  `EGC_ALWAYS`.  The second option only applies in the case of the third option, so 

```Lua
node.setEGCmode(EGC_ON_ALLOC_FAILURE|EGC_ON_MEM_LIMIT, 20000)
```
Will run the EGC if there is an memory allocation failure (before retrying the allocation) or if the heap is less than 20000.  The performance impact is huge, here is a example start from reboot:

```
 for i=1,3 do benchmark(100,i) end 
1844.330 mSec, check = 2000, heap = 37976
2013.452 mSec, check = 2000, heap = 37920
1983.193 mSec, check = 2000, heap = 37960
> for i=1,3 do benchmark(100,i) end 
1864.835 mSec, check = 2000, heap = 37888
2018.612 mSec, check = 2000, heap = 37824
1983.717 mSec, check = 2000, heap = 37904
> node.setEGCmode(node.EGC_ON_ALLOC_FAILURE)
> for i=1,3 do benchmark(100,i) end 
 168.514 mSec, check = 2000, heap = 35496
 291.740 mSec, check = 2000, heap = 37440
 253.626 mSec, check = 2000, heap = 35368
> for i=1,3 do benchmark(100,i) end 
 161.296 mSec, check = 2000, heap = 37384
 297.843 mSec, check = 2000, heap = 35208
[...]

PR #1068, opened 2016-02-20
Add node.egc.setmode() + constants as per #609

TerryE:
See Issue #609 for the discussions on this.  I've decided to offer it to dev.

### DIFF
--- a/docs/en/modules/node.md
+++ b/docs/en/modules/node.md
@@ -390,12 +390,16 @@ provides more detailed information on the EGC.
 
 #### Parameters
 - `mode`
-	- `node.ecg.NOT_ACTIVE` EGC inactive, no collection cycle will be forced in low memory situations
-	- `node.ecg.ON_ALLOC_FAILURE` Try to allocate a new block of memory, and run the garbage collector if the allocation fails. If the allocation fails even after running the garbage collector, the allocator will return with error. 
-	- `node.ecg.ON_MEM_LIMIT` Tun the garbage collector when the memory used by the Lua script goes beyond an upper `limit`. If the upper limit can't be satisfied even after running the garbage collector, the allocator will return with error.
-	- `node.ecg.ALWAYS` Run the garbage collector before each memory allocation. If the allocation fails even after running the garbage collector, the allocator will return with error. This mode is very efficient with regards to memory savings, but it's also the slowest.
-- `level` in the case of `node.ecg.ON_MEM_LIMIT`, this specifies the memory limit.
+	- `node.egc.NOT_ACTIVE` EGC inactive, no collection cycle will be forced in low memory situations
+	- `node.egc.ON_ALLOC_FAILURE` Try to allocate a new block of memory, and run the garbage collector if the allocation fails. If the allocation fails even after running the garbage collector, the allocator will return with error. 
+	- `node.egc.ON_MEM_LIMIT` Run the garbage collector when the memory used by the Lua script goes beyond an upper `limit`. If the upper limit can't be satisfied even after running the garbage collector, the allocator will return with error.
+	- `node.egc.ALWAYS` Run the garbage collector before each memory allocation. If the allocation fails even after running the garbage collector, the allocator will return with error. This mode is very efficient with regards to memory savings, but it's also the slowest.
+- `level` in the case of `node.egc.ON_MEM_LIMIT`, this specifies the memory limit.
   
 #### Returns
 `nil`
 
+#### Example
+
+`node.egc.setmode(node.egc.ALWAYS, 4096)  -- This is the default setting at startup.`
+`node.egc.setmode(node.egc.ON_ALLOC_FAILURE) -- This is the fastest activeEGC mode.`
--- a/docs/en/modules/node.md
+++ b/docs/en/modules/node.md
@@ -378,3 +378,24 @@ node.compile('bigstuff.lua')
 #### See also
 [`node.compile()`](#nodecompile)
 
+# node.egc module
+
+## node.egc.setmode()
+
+Sets the Emergency Garbage Collector mode. [The EGC whitepaper](http://www.eluaproject.net/doc/v0.9/en_elua_egc.html)
+provides more detailed information on the EGC.
+
+####Syntax
+`node.egc.setmode(mode, [param])`
+
+#### Parameters
+- `mode`
+	- `node.ecg.NOT_ACTIVE` EGC inactive, no collection cycle will be forced in low memory situations
+	- `node.ecg.ON_ALLOC_FAILURE` Try to allocate a new block of memory, and run the garbage collector if the allocation fails. If the allocation fails even after running the garbage collector, the allocator will return with error. 
+	- `node.ecg.ON_MEM_LIMIT` Tun the garbage collector when the memory used by the Lua script goes beyond an upper `limit`. If the upper limit can't be satisfied even after running the garbage collector, the allocator will return with error.
+	- `node.ecg.ALWAYS` Run the garbage collector before each memory allocation. If the allocation fails even after running the garbage collector, the allocator will return with error. This mode is very efficient with regards to memory savings, but it's also the slowest.
+- `level` in the case of `node.ecg.ON_MEM_LIMIT`, this specifies the memory limit.
+  
+#### Returns
+`nil`
+
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -9,6 +9,7 @@
 #include "lmem.h"
 #include "lobject.h"
 #include "lstate.h"
+#include "legc.h"
 
 #include "lopcodes.h"
 #include "lstring.h"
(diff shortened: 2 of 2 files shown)


docs/en/modules/node.md
```diff
@@ -378,3 +378,24 @@ node.compile('bigstuff.lua')
...
+
+## node.egc.setmode()
+
+Sets the Emergency Garbage Collector mode. [The EGC whitepaper](http://www.eluaproject.net/doc/v0.9/en_elua_egc.html)
+provides more detailed information on the EGC.
+
+####Syntax
+`node.egc.setmode(mode, [param])`
+
+#### Parameters
```
pjsg: Can you note what the default is?

docs/en/modules/node.md
```diff
@@ -378,3 +378,24 @@ node.compile('bigstuff.lua')
...
+- `mode`
+	- `node.ecg.NOT_ACTIVE` EGC inactive, no collection cycle will be forced in low memory situations
+	- `node.ecg.ON_ALLOC_FAILURE` Try to allocate a new block of memory, and run the garbage collector if the allocation fails. If the allocation fails even after running the garbage collector, the allocator will return with error. 
+	- `node.ecg.ON_MEM_LIMIT` Tun the garbage collector when the memory used by the Lua script goes beyond an upper `limit`. If the upper limit can't be satisfied even after running the garbage collector, the allocator will return with error.
```
pjsg: s/Tun/Run/

docs/en/modules/node.md
```diff
@@ -378,3 +378,24 @@ node.compile('bigstuff.lua')
 #### See also
 [`node.compile()`](#nodecompile)
 
+# node.egc module
+
+## node.egc.setmode()
+
+Sets the Emergency Garbage Collector mode. [The EGC whitepaper](http://www.eluaproject.net/doc/v0.9/en_elua_egc.html)
+provides more detailed information on the EGC.
+
```
pjsg: Can you add an example of the fast version (which is, I think)

```
node.egc.setmode(node.egc.ON_ALLOC_FAILURE)
```

TerryE:
Your comments are noted.  I can incorporate a tweaks patch or reissue the head if it looks as if there is a consensus on including this. :-)

pjsg:
I'm :+1: -- but I'd like to switch the default configuration to ON_ALLOC_FAILURE to give us the performance boost. Then, if you want to run in the old mode, you could switch it back....

TerryE:
The one subtle aspect of this is that the current version might run like a dog, but it keeps the heap as small as possible.   The `node.egc.ON_ALLOC_FAILURE` setting can let the heap grow large.  The problem here is if libraries bypass the Lua allocator (like your recent submissions and use c_zalloc or c_malloc directly) which fail because of not triggering the allocator.  I prefer using the Lua allocators as these catch and throw an out of memory error on failure instead of leaving the application to test for zero return etc.

pjsg:
That is a good point. Do you think that modules/drivers ought to be using the luaM_ family of APIs to allocate and free memory? If so, then do you think it is worth building wrappers to emulate the c_malloc, c_zalloc, c_realloc and c_free calls? It wouldn't be that difficult....

TerryE:
> Do you think it is worth building wrappers to emulate the c_malloc, c_zalloc, c_realloc and c_free calls?

No, I think that this is a pretty tedious sweep activity.  There isn't a `luaM_zalloc()` but we could do with one to avoid the extra memsets.  About half the mallocs do error checking for NULL return which needs to be removed since the luaM routines just throw a Lua error.  The other half will at least not carry on and die with a reboot.  But here are the non-zero counts on the `app/modules`:

Module                          | No o fnon-Lua  mem API calls
--------------------------------|----------------------------------------
app/modules/encoder.c | 5
app/modules/cjson.c | 11
app/modules/net.c | 17
app/modules/sntp.c | 7
app/modules/am2320.c | 1
app/modules/tmr.c | 1
app/modules/coap.c | 11
app/modules/mqtt.c | 31
app/modules/crypto.c | 9
app/modules/gpio.c | 1
app/modules/wifi.c | 1
app/modules/ws2812.c | 2
app/modules/node.c | 1
app/modules/enduser_setup.c | 28

We also need to go through the library directories as well.

pjsg:
Thinking about it some more, we also have to worry about the lua calls where GC is disabled.... 

Given this complexity, I'm happy with the PR as it is without changing the default.

Having the luaM_ functions throw errors is unpleasant as it makes cleanup in C modules very difficult.

I'm wondering if we need an EGC mode like LIMIT which aims to keep (at least) `limit` bytes of free heap. Doesn't fail if it runs out. Might set `limit` to 4096 or similar.    I.e. it would do a GC whenever (after an allocation) there is less than 4k left.

TerryE:
> Thinking about it some more, we also have to worry about the lua calls where GC is disabled....

If you look at the Lus / eLua (internal) examples where this is done properly, these windows are normally preceded by a forced GC.

> Having the luaM_ functions throw errors is unpleasant as it makes clean-up in C modules very difficult.  

As I said, a far more common issue is that malloc calls don't even bother catching the NULL return and then generate a fatal memory exception.  I would far rather have an error thrown and leave the clean-up to Lua with a proper Lua stack trace as well.  There are very few cases where throwing an error leaves the C routines static data in a mess.  This can usually be avoided by hoisting the allocs earlier in the code before the atomic coding takes place (and this is better programming practice anyway, since I can't think of examples where this roll-back is correctly coded.  I got caught out in a couple of places  in my packed lineifno mods to the Lua compiler as this relies on this throwing mechanism during compiles.  This is the Lua way and IMO we should just accept it.

> I'm wondering if we need an EGC mode like LIMIT which aims to keep (at least) limit bytes of free heap. Doesn't fail if it runs out. Might set limit to 4096 or similar. I.e. it would do a GC whenever (after an allocation) there is less than 4k left.

I've also been wondering about this and having a slightly more sophisticated algo for triggering CG, e.g. some variant of 

-  1 in N allocs / frees possibly with N a function of the following
-  % of the free heap consumed since the last GC 
-  any more tests you can think off.

These might seem complex, but in reality its only going to be a few dozen, maybe 100 instructions which is small beer if we avoid floating point.  Even if we end up doing GCs 1:20 mallocs, this is still a _huge_ percentage of the current overhead and this will keep the heap small.  Another idea is to put some more detailed heap stats / analysis so we can get a better idea of how to get the right balance.  The current `ALWAYYS` setting is just crazy, IMO.

pjsg:
I don't know how expensive the system_get_free_heap_space() call is -- I assume that this gets the answer right. The existing lua code tries to track memory usage, but it only tracks the amount of memory allocated and doesn't take into account the (unknown) overhead of the memory allocation subsystem. 

I feel tempted to change some of my outstanding code to use userdata objects rather than directly allocated memory and let the GC deal with them.... However, this would (sort of) break the layering between the module and the driver component.

TerryE:
> I feel tempted to change some of my outstanding code to use userdata objects rather than directly allocated memory and let the GC deal with them.... However, this would (sort of) break the layering between the module and the driver component.

Not really.  Have you read the V3 PiL?  You just need to do all of the allocations in Lua-Land and treat the memory as quasi-static as any malloc / free'ing at the ISR-side is a no-no anyway.  So long as the userdata objects are registered in the registry, then the GC knows about then and the fact that they are referenced.  You just need to make sure that all of your execution paths unregister them when done.  (The `net` module doesn't do this consistently which is why the net memory leaks occur).

pjsg:
The mental model that I have is that for talking to devices there are really two layers:

* The LUA interface layer. This should use the userdata stuff for memory allocation and the like. This deals with the lua data obects, the lua stack, registration, providing a sensible interface etc.

* The device interface layer that knows nothing about lua and just presents a sane API to the lua interface layer.

Since the device interface layer knows nothing about lua, it seems wrong to make it use the lua memory allocation system (though certainly possible of course)

On a related topic, I notice that most of the nodemcu modules use channel numbers/ids when there are multiple possible instances (tmr, gpio, pwm, spi etc). Another approach is to make the setup method return an object which is then used to operate on that instance. When the object is gc'ed, then it should be closed (if not already closed). In this latter model, then it makes sense to leverage more of the lua infrastructure. Do you have any opinions here?

TerryE:
> Since the device interface layer knows nothing about lua, it seems wrong to make it use the lua memory allocation system (though certainly possible of course).

I would push back in this one in that the device interface layer should probably steer well away from _any_ malloc / free activities and if necessary indirect through a `static struct X *x{MAXC_X_NO]` layer in `.bss`. The actual alloc/free would be done by the Lua-land API wrappers so there is no issue with using the Lua mem system here.

> On a related topic, I notice that most of the nodemcu modules use channel numbers/ids when there are multiple possible instances (tmr, gpio, pwm, spi etc). Another approach is to make the setup method return an object which is then used to operate on that instance. When the object is gc'ed, then it should be closed (if not already closed). In this latter model, then it makes sense to leverage more of the lua infrastructure. Do you have any opinions here?

The standard Lua object model makes significant use of the registry and Userdata constructs.  This is a no-no in non-Lua Land, however as I've just suggested, there is nothing stopping using a statically allocated a `static struct X *x{MAXC_X_NO]` layer so there is only one **size_t** overead per buffer/control structure.

jmattsson:
I *definitely* would like to see this included. I'm a bit more on the fence with regard to the use of the various malloc routines. In the general case, if the Lua alloc wrappers automatically raise errors, then sweet! Let's make a mention of it in CONTRIBUTING.md and/or the API docs, and push for it in PRs. (I have to admit I wasn't aware of these routines until quite recently).

However, some of the code we share or have imported from external sources which know nothing about Lua. I'd be hesitant to attempt a retro-fit of Lua allocators there. I'm wondering if there might be a sweet spot we could hit by doing a gc sweep before the L0 task runs every N times, somehow?

TerryE:
My reaction is to take two bites at this cherry. I suggest that we commit this version to `dev` and let people play with it and benchmark it.   We can do this safely since the default is "no change".  Let's then close this issue and open a new one to discuss possible algos for a finer programmatic control of GC.

devsaurus:
:+1:

TerryE:
I'll just do a commit to fix the couple of typo and then I'll commit this since its got the :+1: from a couple of committers.

Issue #925, opened 2016-01-12
Implement coder module


TerryE:
Implement `coder` module with methods:

*    toBase64
*    fromBase64
*    toHex
*    fromHex

Denormalise `crypto` to call the coder.toXXX routines.  Discussion on #486 refers

marcelstoer:
> Denormalise `crypto` to call the coder.toXXX routines.

Not sure I understand.

Would that be the first time that two non-core modules actually depend on each other (see https://github.com/nodemcu/nodemcu-firmware/issues/386#issuecomment-115108172)? Why wouldn't you want to remove the methods in `crypto`? Calling the old `crypto.xxx` functions would still fail on a firmware that doesn't have the `coder` module, right?

jmattsson:
Hmm, that's a good point. I hadn't thought of that added complication for the backwards compatibiilty. Maybe that's enough to not do it and "fix it in documentation" instead? From a user perspective I'd still prefer backwards compatibility, but I definitely hear you on the technical difficulties.

Maybe we should start out with an
   ```
#if defined(LUA_USE_MODULES_CRYPTO) && !defined(LUA_USE_MODULES_CODER)
# define LUA_USE_MODULES_CODER
#endif
```
and then officially split them a bit later down the track when we feel we have time to maintain "stuff that has changed from version x to y that you need to be aware of" notes?

marcelstoer:
A software that is properly released usually has some deprecation policy based on a release plan in place. We don't have that and would have to rely on time-based deprecation. That would/could mean we create the new module, *copy* the methods and add a deprecation note to the docs for `crypto`. It could state that the deprecated methods are removed in n weeks or months.

TerryE:
I can leave the old code in crypto with a  `#ifdef LUA_USE_MODULES_CODER` option to substitute the the calls to coder instead.  But I won't add any new methods to crypto.

Alkorin:
`encoding` isn't a better name than `coder` ?
It's the name of golang's package for this kind of stuff

https://golang.org/pkg/encoding/hex/
https://golang.org/pkg/encoding/base64/

TerryE:
Done

devyte:
@TerryE Thank you for this, I _really_ want to get rid of the lua implementation of fromBase64() in my code.
@marcelstoer Is it possible to add the encoder module to the nodemcu firmware build service?

marcelstoer:
[...]

PR #1072, opened 2016-02-21
Addition of encoder module with bas64 and hex encode/decode as per #925

TerryE:
See Issue #925 for discussions on this.

### DIFF
--- a/app/modules/crypto.c
+++ b/app/modules/crypto.c
@@ -37,7 +37,29 @@ static int crypto_sha1( lua_State* L )
   return 1;
 }
 
+#ifdef LUA_USE_MODULES_ENCODER
+static int call_endcoder( lua_State* L, const char *function ) {
+  if (lua_gettop(L) != 1) { 
+    luaL_error(L, "%s must have one argument", function);
+  }
+  lua_getfield(L, LUA_GLOBALSINDEX, "encoder");
+  if (!lua_istable(L, -1) && !lua_isrotable(L, -1)) { // also need table just in case encoder has been overloaded
+    luaL_error(L, "Cannot find encoder.%s", function); 
+  }
+  lua_getfield(L, -1, function);
+  lua_insert(L, 1);    //move function below the argument
+  lua_pop(L, 1);       //and dump the encoder rotable from stack.
+  lua_call(L,1,1);     // call encoder.xxx(string)
+  return 1;
+}
 
+static int crypto_base64_encode (lua_State* L) { 
+  return call_endcoder(L, "toBase64"); 
+}
+static int crypto_hex_encode (lua_State* L) { 
+  return call_endcoder(L, "toHex"); 
+}
+#else
 static const char* bytes64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 /**
   * encoded = crypto.toBase64(raw)
@@ -84,7 +106,7 @@ static int crypto_hex_encode( lua_State* L)
   c_free(out);
   return 1;
 }
-
+#endif
 /**
   * masked = crypto.mask(message, mask)
   *
--- a/app/modules/crypto.c
+++ b/app/modules/crypto.c
@@ -38,7 +38,7 @@ static int crypto_sha1( lua_State* L )
 }
 
 #ifdef LUA_USE_MODULES_ENCODER
-static int call_endcoder( lua_State* L, const char *function ) {
+static int call_encoder( lua_State* L, const char *function ) {
   if (lua_gettop(L) != 1) { 
     luaL_error(L, "%s must have one argument", function);
   }
@@ -54,10 +54,10 @@ static int call_endcoder( lua_State* L, const char *function ) {
 }
 
 static int crypto_base64_encode (lua_State* L) { 
-  return call_endcoder(L, "toBase64"); 
+  return call_encoder(L, "toBase64"); 
 }
 static int crypto_hex_encode (lua_State* L) { 
-  return call_endcoder(L, "toHex"); 
+  return call_encoder(L, "toHex"); 
 }
 #else
 static const char* bytes64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
--- a/app/modules/encoder.c
+++ b/app/modules/encoder.c
@@ -8,16 +8,16 @@
 #define BASE64_PADDING '='
 #define ISBASE64(c) (unbytes64[c] != BASE64_INVALID)
 
-static const char b64[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+static const uint8 b64[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
-static char *toBase64 ( lua_State* L, const char *msg, size_t *len){
+static uint8 *toBase64 ( lua_State* L, const uint8 *msg, size_t *len){
   size_t i, n = *len;
 
   if (!n)  // handle empty string case 
     return NULL;
 
-  char* q, *out = (char*)luaM_malloc(L, (n + 2) / 3 * 4);
-  char bytes64[sizeof(b64)];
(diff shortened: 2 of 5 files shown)


app/modules/encoder.c
```diff
@@ -0,0 +1,154 @@
...
+
+  if (!n)  // handle empty string case 
+    return NULL;
+
+  char* q, *out = (char*)luaM_malloc(L, (n + 2) / 3 * 4);
+  char bytes64[sizeof(b64)];
+  c_memcpy(bytes64, b64, sizeof(b64));   //Avoid lots of flash unaligned fetches
+  
+  for (i = 0, q = out; i < n; i += 3) {
+    int a = msg[i];
```
pjsg: I suspect that since msg is char *, this will do sign expansion and then bad things will happen. Probably want either to make msg an unsigned char * or add & 0xff

app/modules/encoder.c
```diff
@@ -0,0 +1,154 @@
...
+
+static const char b64[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+static char *toBase64 ( lua_State* L, const char *msg, size_t *len){
+  size_t i, n = *len;
+
+  if (!n)  // handle empty string case 
+    return NULL;
+
+  char* q, *out = (char*)luaM_malloc(L, (n + 2) / 3 * 4);
```
pjsg: Missing memory allocation failure check

pjsg:
This taught me that the dev environment runs with chars being unsigned by default, Interesting...

TerryE:
Yes, they are, but where the arithmetic goes wrong if they happen to be signed (e.g. x>>4) I specify the unsignedness explicitly.  Still this works as we discussed the legacy `crypto.toHex` and the `crypto.toBase64`  functions work with and without the `encoder` module climbed in.

pjsg:
Your base64 code relies on the unsignedness of the char (and also on the fact that CHAR_MIN == 0)

app/modules/crypto.c
```diff
@@ -37,7 +37,29 @@ static int crypto_sha1( lua_State* L )
   return 1;
 }
 
+#ifdef LUA_USE_MODULES_ENCODER
+static int call_endcoder( lua_State* L, const char *function ) {
```
ABonner: Did you mean to call this `call_endcoder` with a `d`, or just `call_encoder`?

TerryE:
> Your base64 code relies on the unsignedness of the char

Philip, two of the routines are essentially copied from `crypto` and make these assumptions, but you are right -- perhaps it is better just to use `uint8_t *` rather than `char *`.  I'll give it a couple of more days for any more feedback and than do an updated PR to sweep all these comments up.

TerryE:
@jmattsson Johny, have you any comments before I do this rework.  I'd like to sweep all feedback into this, rather have to do another circle around.

app/modules/crypto.c
TerryE: Yup you are right.  Wonder why I didn't get a compile warning?

ABonner: Looks like you autocompleted all other references the same :)

docs/en/modules/encoder.md
```diff
@@ -0,0 +1,75 @@
...
+
+#### Parameters
+`binary` input string to Base64 encode
+
+#### Return
+A Base64 encoded string.
+
+#### Example
+```lua
+print(encoder.toBase64(cyrpto.hash("sha1","abc")))
```
ABonner: Should be "crypto.hash"

docs/en/modules/encoder.md
```diff
@@ -0,0 +1,75 @@
...
+
+## encoder.fromHex()
+
+Provides a Lua binary string decode of a ASCII hex string. Each byte in the output string is represented as two hex characters in the input.
+
+#### Syntax
+`binary = encoder.toHex(hexstr)`
+
+#### Parameters
+`hexstr`  An ASCII hex string.
```
ABonner: Should be "A Lua binary string"

TerryE: @ABonner Thanks Adam, for the review.  I will integrate these and other comments into an update.  I will also go with a uint8 type for all byte data to make the unsigned nature of the calcs explicit rather than implicit. :smile:

app/modules/crypto.c
```diff
@@ -37,7 +37,29 @@ static int crypto_sha1( lua_State* L )
...
+}
 
+static int crypto_base64_encode (lua_State* L) { 
+  return call_endcoder(L, "toBase64"); 
+}
+static int crypto_hex_encode (lua_State* L) { 
+  return call_endcoder(L, "toHex"); 
+}
+#else
 static const char* bytes64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
```
jmattsson: Shouldn't this string be unused and removable now?

app/modules/encoder.c
```diff
@@ -0,0 +1,154 @@
...
+    *q++ = (i + 2 < n) ? bytes64[(c & 63)] : BASE64_PADDING;
+  }
+  *len = q - out;
+  return out;
+}
+
+static char *fromBase64 ( lua_State* L, const char *enc_msg, size_t *len){
+  int i, n = *len, pad = 0;
+  unsigned const char *p;
+  unsigned char unbytes64[CHAR_MAX+1-CHAR_MIN], *msg, *q, a, b, c, d;
```
jmattsson: So if CHAR_MAX == 127, and CHAR_MIN == -128, we have a zero-length array? I'm not following your thinking here.

jmattsson:
I haven't got time to investigate for myself right now, but is there any chance of re-using the ROM functions for base64 instead of needing our own? I think it'll depend on how/if they allocate memory.
```
PROVIDE ( base64_decode = 0x40009648 );
PROVIDE ( base64_encode = 0x400094fc );
```

app/modules/encoder.c
TerryE: CHAR_MAX+1-CHAR_MIN = 127+1-(-128)=128+128=256

TerryE:
Good suggestion. I'll take a look.  If easy to do, I'll also add this to the `crypto`.

app/modules/encoder.c
jmattsson: Good morning Johny...

TerryE:
They use the ROM mem_malloc / mem_free rather than the SDK's RTOS compatible pvPortMalloc a.k.a. os_malloc that is used in the app. I assume that these aren't mix and match.

jmattsson:
I seem to recall having had things crash & burn (literally and figuratively, correspondingly) when trying to do said mix & match, so no, I'd recommend against it. Unless you want to copy the base64 routines from rom into iram dynamically and hot-patch the alloc/free calls to point to something more compatible... ;)

No further comments from me then, ping me when you've done the next round here and I can merge.

TerryE:
> Unless you want to copy the base64 routines from rom into iram dynamically and hot-patch the alloc/free calls to point to something more compatible

Not worth it.  The generated code saving just isn't worth this complexity.  If we want to save .text space then there are lower hanging fruit to go after.

jmattsson:
I *really* wasn't serious with that suggestion, Terry! With the amount of finicky stuff one would have to get right (like, where do I even load this to in iram without clobbering anything?) would with near certainty result in an *increase* of code size rather than a decrease. I'll take the maintainability of regular code over meta-code, thank you!

TerryE:
Sorry Johny, I picked up a flea from some friends dog and woke up in the middle of the night with my leg in the middle of being a lunch meal for the flea and its kids and itching like hell -- so a shower and the guest bedroom later, I am now clearing off issues whilst the itching subsides.  My brain isn't working too well!

app/modules/encoder.c
pjsg: Except that CHAR_MIN=0 and CHAR_MAX=255, but it still turns out to be 256!

app/modules/crypto.c
TerryE: It is still needed in the non-encoder version as these two routines are maintained iinline.

TerryE:
I've updated the files to echo the review comments. I've also fix a bug with the end packet base64 decode that my test vectors through up.   Perhaps one of the committers could merge this.  :smile:

jmattsson:
Thanks Terry!

PR #1082, opened 2016-02-24
Made level triggers work better than before

### LINKED ISSUE #1078: "GPIO rework" merged to dev on 18 Feb broke GPIO triggering
@TerryE @jmattsson 
I constantly follow all the comments on Pulse and appreciate what I am learning from you. 
I am developing full time using custom NodeMCU firmware builds.

1. I have verified, using the same test code utilizing **GPIO trigger functionality**, working on the dev commit merged by TerryE on 17 Feb, **stopped working** on merged dev commits made on **18 Feb**.
 
2. I have also verified on the latest build made today that it is still not working.
 
3. It appears to be related to something that changed related to the "Merge pull request #1061 from TerryE/dev-task-if" committed by jmattsson (Add New Tasking I/F and rework GPIO, UART, etc to support it)

pjsg:
I hope that this PR fixes #1078. In my testing it makes level triggers behave in a reasonable way -- they get repeatedly called while the situation persists. However, there is only ever one callback queued for each pin at a time. 

I'm interested in whether this solves @NicolSpies problem.

@TerryE since you did the original big rework, I'd appreciate your input. I removed the pin_trigger stuff and just emulated it by disabling and re-enabling the interrupts on a per pin basis.

This code, when saved to a file and then dofile'd exercises some pieces of the gpio interrupt system. You should be able to run it on any board if you connect gpio 4 to gpio 5 with a bit of wire.

```Lua
function stringify_table(t)
  if type(t) == "table" then
    return table.concat(t, ",")
  end
  return t
end

function assertEquals(exp, act)
  exp=stringify_table(exp)
  act=stringify_table(act)
  if exp ~= act then
    error(string.format("ASSERT: %s != %s", exp, act))
  end
end

local log = {}

function record(typ, level)
  print(typ)
  if #log >= 3 then
    if typ == "low" then
      gpio.write(5,1)
    elseif typ == "high" then
      gpio.write(5,0)
    end
    return
  end
  table.insert(log, string.format("%s:%d", typ, level))
end

function setTrig(typ)
  gpio.trig(4, typ, function(level) 
    record(typ, level)
  end)
end


function asyncDelay(ms)
  coroutine.yield(ms)
end

function runTest()
    gpio.mode(5, gpio.OUTPUT)
    gpio.mode(4, gpio.INPUT)
    
    -- verify connected together
    gpio.write(5, 1)
    assertEquals(1, gpio.read(4))
    gpio.write(5, 0)
    assertEquals(0, gpio.read(4))
    
    gpio.mode(4, gpio.INT)
    
    setTrig("down")
    assertEquals({}, log)
    gpio.write(5, 1)
    assertEquals({}, log)
    
    gpio.write(5, 0)
    asyncDelay(10)
    assertEquals("down:0", log)
    log = {}
    gpio.write(5, 1)
    asyncDelay(10)
    assertEquals({}, log)

    gpio.write(5, 0)
    gpio.write(5, 1)
    gpio.write(5, 0)
    gpio.write(5, 1)
    asyncDelay(10)
    assertEquals("down:0", log)
    log = {}

    gpio.write(5, 0)
    asyncDelay(10)
    assertEquals("down:0", log)
    log = {}


    setTrig("both")
    gpio.write(5, 1)
    asyncDelay(10)
    assertEquals("both:1", log)
    log = {}
    gpio.write(5, 0)
    asyncDelay(10)
    assertEquals("both:0", log)

    gpio.mode(4, gpio.INPUT)
    
    setTrig("low")
    print("set-low")
    log = {}
    asyncDelay(30)
    assertEquals("low:0,low:0,low:0", log)
    gpio.write(5, 1)
    log = {}
    print("set-low-high")
    assertEquals({}, log)
    gpio.write(5, 0)    
    asyncDelay(30)
    assertEquals("low:0,low:0,low:0", log)
    
end

co = coroutine.create(runTest)

function doDelay()
  local ok, delay = coroutine.resume(co)
  if not ok then
    print(delay)
    return
  end
  if delay then
    tmr.alarm(0, delay, 0, doDelay)
  end
end

doDelay()
```

### DIFF
--- a/app/modules/gpio.c
+++ b/app/modules/gpio.c
@@ -7,6 +7,7 @@
 #include "user_interface.h"
 #include "c_types.h"
 #include "c_string.h"
+#include "gpio.h"
 
 #define PULLUP PLATFORM_GPIO_PULLUP
 #define FLOAT PLATFORM_GPIO_FLOAT
@@ -17,6 +18,12 @@
 #define LOW PLATFORM_GPIO_LOW
 
 #ifdef GPIO_INTERRUPT_ENABLE
+
+// We also know that the non-level interrupt types are < LOLEVEL, and that
+// HILEVEL is > LOLEVEL. Since this is burned into the hardware it is not
+// going to change.
+#define INTERRUPT_TYPE_IS_LEVEL(x)	((x) >= GPIO_PIN_INTR_LOLEVEL)
+
 static int gpio_cb_ref[GPIO_PIN_NUM];
 
 // This task is scheduled by the ISR and is used
@@ -34,7 +41,7 @@ static void gpio_intr_callback_task (task_param_t param, uint8 priority)
     lua_State *L = lua_getstate();
     NODE_DBG("Calling: %08x\n", gpio_cb_ref[pin]);
     //
-    if (pin_int_type[pin] < 4) {
+    if (!INTERRUPT_TYPE_IS_LEVEL(pin_int_type[pin])) {
       // Edge triggered -- re-enable the interrupt
       platform_gpio_intr_init(pin, pin_int_type[pin]);
     }
@@ -44,7 +51,7 @@ static void gpio_intr_callback_task (task_param_t param, uint8 priority)
     lua_pushinteger(L, level);
     lua_call(L, 1, 0);
 
-    if (pin_int_type[pin] >= 4) {
+    if (INTERRUPT_TYPE_IS_LEVEL(pin_int_type[pin])) {
       // Level triggered -- re-enable the callback
       platform_gpio_intr_init(pin, pin_int_type[pin]);
     }
--- a/app/modules/gpio.c
+++ b/app/modules/gpio.c
@@ -19,8 +19,9 @@
 #ifdef GPIO_INTERRUPT_ENABLE
 static int gpio_cb_ref[GPIO_PIN_NUM];
 
-// This task is scheduled by the ISP if pin_trigger[pin] is true and is used
-// to intitied the Lua-land gpio.trig() callback function
+// This task is scheduled by the ISR and is used
+// to initiate the Lua-land gpio.trig() callback function
+// It also re-enables the pin interrupt, so that we get another callback queued
 static void gpio_intr_callback_task (task_param_t param, uint8 priority)
 {
   unsigned pin = param >> 1;
@@ -32,11 +33,21 @@ static void gpio_intr_callback_task (task_param_t param, uint8 priority)
     // GPIO callbacks are run in L0 and inlcude the level as a parameter
     lua_State *L = lua_getstate();
     NODE_DBG("Calling: %08x\n", gpio_cb_ref[pin]);
-    // The trigger is reset before the callback, as the callback itself might reset the trigger
-    pin_trigger[pin] = true;
+    //
+    if (pin_int_type[pin] < 4) {
+      // Edge triggered -- re-enable the interrupt
+      platform_gpio_intr_init(pin, pin_int_type[pin]);
+    }
+
+    // Do the actual callback
     lua_rawgeti(L, LUA_REGISTRYINDEX, gpio_cb_ref[pin]);
     lua_pushinteger(L, level);
     lua_call(L, 1, 0);
+
+    if (pin_int_type[pin] >= 4) {
+      // Level triggered -- re-enable the callback
+      platform_gpio_intr_init(pin, pin_int_type[pin]);
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


app/modules/gpio.c
```diff
@@ -32,8 +33,8 @@ static void gpio_intr_callback_task (task_param_t param, uint8 priority)
     // GPIO callbacks are run in L0 and inlcude the level as a parameter
     lua_State *L = lua_getstate();
     NODE_DBG("Calling: %08x\n", gpio_cb_ref[pin]);
-    // The trigger is reset before the callback, as the callback itself might reset the trigger
-    pin_trigger[pin] = true;
+    // Reenable the interrupt just before calling the callback
+    platform_gpio_intr_init(pin, pin_int_type[pin]);
```
TerryE: The  original code enabled the interrupt _after_ the `lua_call` and the ISR has already disabled interrupts and cleared down the interrupt state, so we only need to do a `gpio_pin_intr_state_set(GPIO_ID_PIN(pin_num[pin]), type);` to re-enable interrupts.  The issue is _when_ from a temporal logic POV.  

-  In the case of a level interrupt the Lua cb code carries out the I/O actions to remove the level, so we should do it after the call otherwise we will get another "echo" task being posted.
-  In the case of an edge interrupt, doing it before isn't going to create the echo trigger, but there are going to be a few 100 instruction times (say 2-10µSec) where the last interrupt has been posted and we could miss an edge before reenabling the interrupts.

pjsg: I was conflicted about this too. We could always do "enable before for edges", "enable after for level" -- one of my versions had that logic, but I felt it was over complex. However, given your comments, I suspect that it might ahve been the right thing to do after all.

jmattsson:
I'll note that there is still somewhat of a race here - for an edge trigger where the user is only wishing to capture a single edge, we might have already queued up a second callback before the Lua handler gets a chance to take the pin out of interrupt mode. There *might* be a case for explicitly catering for edge, level and one-shot-edge on the Lua level. I wouldn't let that hold up getting this fix merged though.

pjsg:
I fixed the '4' and wrapped it so that the intention is clearer.

TerryE:
We could keep on polishing this, but I think that as this addresses a known bug then we  should do this as another pass if at all.  Thanks Philip.

NicolSpies:
@TerryE @pjsg @jmattsson : I confirm the reported bug is fixed, working as before when I trigger on "low". I still find it impossible to trigger on the edge using "down". Will investigate further. Thanks Nicol

NicolSpies:
**Observation:** 

1. As before, when I trigger on "low" it works with the following debug info. The "val2" is the decimal value of the GPIO port expander pin where the state change caused the chip INT line on which I trigger, to change to low. This value is correct and is what I read on the i2c bus.

> Function platform_gpio_mode() is called. pin_mux:1610614840, func:0
Function platform_gpio_mode() is called. pin_mux:1610614836, func:0
I/O Init
pin,mode,pullup= 2 2 0
Pin data at mode: 2 6000083c, 4 0 3, 00000001
Function platform_gpio_mode() is called. pin_mux:1610614844, func:0
Pin data: 2 4 6000083c, 4 0 3, 00000002
> pin:2, level:0 
Calling: 00000002
val2: 192

If I then change to trigger on "down" it works once with the following debug info:

> Function platform_gpio_mode() is called. pin_mux:1610614840, func:0
Function platform_gpio_mode() is called. pin_mux:1610614836, func:0
I/O Init
pin,mode,pullup= 2 2 0
Pin data at mode: 2 6000083c, 4 0 0, fffffffe
Function platform_gpio_mode() is called. pin_mux:1610614844, func:0
Pin data: 2 2 6000083c, 4 0 0, 00000001
> pin:2, level:0 
Calling: 00000001
val2: 96

The only way to get it to work again is to change it back to "low Previously, triggering on "both" did not work at all. Now triggering on both edges works for a while and then stops with the following debug info:

> Function platform_gpio_mode() is called. pin_mux:1610614840, func:0
Function platform_gpio_mode() is called. pin_mux:1610614836, func:0
I/O Init
pin,mode,pullup= 2 2 0
Pin data at mode: 2 6000083c, 4 0 0, fffffffe
Function platform_gpio_mode() is called. pin_mux:1610614844, func:0
Pin data: 2 3 6000083c, 4 0 0, 00000001
> pin:2, level:0 
Calling: 00000001
val2: 160
pin:2, level:0 
Calling: 00000001
val2: 160

Resetting the ESP, like before, doesn't fix the problem. The only way to fix is to upload file with triggering on "low" again. It almost appears as if something is overwritten.

PR #1088, opened 2016-02-26
Make the wifi.smart functions options and remove wofs and upgrade

TerryE:
AFAIK no one uses the wifi.startsmart() and wifi.stopsmart().  Removing
them frees up an extra 20-25K of Flash to use as filesystem.  So I have
added a new config define WIFI_SMART_ENABLE which is enabled by default
so the default functionality is the same, but if this is commented out
then this code is omitted.

I have also removed wofs and upgrade from this build as we no longer
support these.


TerryE:
Sorry I accidentally deleted a `WIFI_SMART_ENABLE` then retyped `WIFI_SMART_ENABLED` so the TravisCI failed first time through.

jmattsson:
:+1: I'd even go so far as to recommend getting rid of the BUILD_SPIFFS define altogether since there's no longer a choice between it and wofs.

marcelstoer:
> I have added a new config define WIFI_SMART_ENABLE which is enabled by default

Hope you don't mind I'll disable it by default in the cloud builder.

TerryE:
Can I suggest that we leave the SPIFFS conditional in?  I know that it always evaluates to `#if 1` but it is a useful place holder marker should we want to offer an alternative FS in future. Unless anyone objects in the next day, I will merge this.

TerryE:
>> I have added a new config define WIFI_SMART_ENABLE which is enabled by default

> Hope you don't mind I'll disable it by default in the cloud builder.

My preference would be to disable by default \[fullstop].  However we'd need to put this in some release nots or the documentation just in case we have someone using this functionality in an existing app.

jmattsson:
Given the space savings, I too am in favour of disabling by default. Make a note in the docs, and if we get issues raised we'll just have to deal :)

It would be interesting to know what portion of NodeMCU users build themselves vs use Marcel's cloud builder. Knowing that would help inform these decisions...

TerryE:
OK, unless there are any dissenting voices, I'll give this another day to let these be raised, then I'll make this change to the default and merge this.

marcelstoer:
Terry, I interpreted your smart yes/no comments and Johny's supportive vote to that effect that you'd disable it before the merge? Did I get that wrong? Ok if I create a separate PR to disable it by default?

TerryE:
Marcel, Sorry I forgot, but on a small issue like this *where there is pre-consensus* amongst the committers then the easiest thing to do is to pull dev, make the one line change and push it back.  I wouldn't even bother with a PR in this instance.

You have my blessing -- the old boy that did the mind fart and forgot in the first place :laughing:

PR #1091, opened 2016-02-27
BME280 - fixing 2 issues

vsky279:
1) Signal incorect reading when the sensor is disconnected.
2) Reset humidity oversampling settings when calling `bme280.startreadout()` - if the sensor is reconnected and readouts are in forced mode there is no need to re-initialize the sensor (`bme280.init()`).


TerryE:
@vsky279, this is your baby, so I assume that you understand what you are doing when it comes to his sort of bugfix. :smile:

Issue #1099, opened 2016-03-01
Crash on dev branch with pwm.setup if LUA_USE_MODULES_GPIO not defined


pjsg:
It turns out that the function get_pin_map() never gets called unless LUA_USE_MODULES_GPIO is not defined. Also, I don't think it gets called unless GPIO_INTERRUPT_ENABLE is set either.

This function initializes various crucial data structures that are used by the pwm code (e.g. to enable the pin for output).

I'm about to file a PR to fix this.

TerryE:
#1100 now merged.  We may not be done with this, but the bug has been fixed.

PR #1100, opened 2016-03-01
Add call to get_pin_map() in platform_init()

pjsg:
This PR fixes the crash reported by @eku in #1057 (the original PR which added the new PWM support). 

There is ticket #1099 which explains the problem.

The fix here is just to call get_pin_map() during the pwms_init() -- in case it doesn't get called by the gpio init (e.g. if GPIO is not included).

I also removed the pin_trigger variable -- which I should have done as part of the PR #1082

### DIFF
--- a/app/platform/platform.c
+++ b/app/platform/platform.c
@@ -18,6 +18,9 @@ static void pwms_init();
 
 int platform_init()
 {
+  // Setup the various forward and reverse mappings for the pins
+  get_pin_map();
+
   // Setup PWMs
   pwms_init();
 
@@ -182,10 +185,8 @@ static void ICACHE_RAM_ATTR platform_gpio_intr_dispatcher (void *dummy){
 
 void platform_gpio_init( task_handle_t gpio_task )
 {
-  int i;
   gpio_task_handle = gpio_task;
 
-  get_pin_map();
   ETS_GPIO_INTR_ATTACH(platform_gpio_intr_dispatcher, NULL);
 }
 /*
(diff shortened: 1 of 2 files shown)


TerryE:
@pjsg Philip, this one was down to me.  It was a Q of not doing full coverage of all of the various ENABLE define options.  You propose change picks up one path but still misses another.  The pin_map() is used by gpio, pwm and i2c.  We can ignore the last since i2c depends on gpio, though you can use pwm without using or initialising gpio. The gpio module uses the pin map in both interrupt and non-interrupt versions.

You fixed the pwm case but not the non-interrupt gpio one.  Maybe the easiest way is to rename **get_pin_map** to **platform_get_pin_map** which it should really be since it is extern function in the platform name space and call this directly from both the   `luaopen_gpio()` in `gpio.c` as well as `lpwm_setup()` in `pwm.c`.

However, this also throws up another issue but this is another issue to fix separately to this and a separate  PR

TerryE:
Can I change my mind again.  The pin_init() should be called from platform_init() as it is really platform wide.

app/platform/platform.c
```diff
@@ -318,6 +317,7 @@ static void pwms_init()
     pwms_duty[i] = DUTY(0);
   }
   pwm_init(500, NULL);
+  get_pin_map();
```
eku: Is it save to call this after ```pwm_init()```?

pjsg: It appears to be safe to call there. However, as @TerryE notes, it is symptomatic of a larger mess....

TerryE:
Philip, I will merge this, so that the dev isn't actually broken, but I suggest that you and I bat this around a little further and possibly revisit this.  //Terry

eku:
Thanks.

PR #1105, opened 2016-03-01
Extend the node library to include a Lua implementation of task posting

TerryE:
The documentation example explains all really.  Now that we have a task I/F at the  API, it also make sense to offer this at the Lua level.

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -449,19 +449,19 @@ static int node_compile( lua_State* L )
   return 0;
 }
 
-// Task callback handler for node.tasknow()
+// Task callback handler for node.task.post()
 static task_handle_t do_node_task_handle;
 static void do_node_task (task_param_t task_fn_ref, uint8_t prio)
 {
   UNUSED(prio);
   lua_State* L = lua_getstate();
   lua_rawgeti(L, LUA_REGISTRYINDEX, (int)task_fn_ref);
-	luaL_unref(L, LUA_REGISTRYINDEX, (int)task_fn_ref);
+  luaL_unref(L, LUA_REGISTRYINDEX, (int)task_fn_ref);
   lua_pushinteger(L, prio);
-	lua_call(L, 1, 0);
+  lua_call(L, 1, 0);
 }
 
-// Lua: node.tasknow(task_cb)) -- schedule a task for execution next
+// Lua: node.task.post([priority],task_cb) -- schedule a task for execution next
 static int node_task_post( lua_State* L )
 {
   int n = 1, Ltype = lua_type(L, 1);
@@ -472,14 +472,18 @@ static int node_task_post( lua_State* L )
     Ltype = lua_type(L, ++n);
   }
   luaL_argcheck(L, Ltype == LUA_TFUNCTION || Ltype == LUA_TLIGHTFUNCTION, n, "invalid function");
-	lua_pushvalue(L, n);
+    lua_pushvalue(L, n);
 
-	int task_fn_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+  int task_fn_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 
   if (!do_node_task_handle)  // bind the task handle to do_node_task on 1st call
     do_node_task_handle = task_get_id(do_node_task);
 
-  task_post(priority, do_node_task_handle, (task_param_t)task_fn_ref);
+  if(!task_post(priority, do_node_task_handle, (task_param_t)task_fn_ref)) {
+    luaL_unref(L, LUA_REGISTRYINDEX, task_fn_ref);
+    luaL_error(L, "Task queue overflow. Task not posted");
+  }
+  return 0;
 }
 
 // Lua: setcpufreq(mhz)
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -26,6 +26,7 @@
 #include "flash_fs.h"
 #include "user_version.h"
 #include "rom.h"
+#include "task/task.h"
 
 #define CPU80MHZ 80
 #define CPU160MHZ 160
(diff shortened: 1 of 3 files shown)


TerryE:
This seems to work fine, including with coroutining, as per this example:
```Lua
do
  local c, i = coroutine, 0
  p = c.create(function()
    local function resume()
      print "resuming"
      c.resume(p)
    end
    while i < 30 do
      print('.')
      if i % 3 == 0 then
        node.task.post(resume)
        c.yield()
      end
      i = i + 1
    end
    print "exiting"
  end)
  c.resume(p)
  print"exit1"
end
```
I have two tests for the reader:

1.  Do you understand how this works? :smile: 
2.  If so, do you understand why p can't be a local because _I_ haven't worked out that one yet.

pjsg:
Yes, and (I think) yes.

I suspect that in the example  
```
do
  local p = function() print(p) end
  p()
end
```
I think that the `p` in the body of the function refers to the global `p` as the interpreter hasn't (yet) defined the local `p`. Your example is more complicated version of this.....

BTW -- the example above prints `nil`

TerryE:
Philip, 10 out of 10 !! As you say, this is nothing to do with bugs in the Lua RTS, it is more that I was being a noob and this is a feature of how Lua orders its name resolution. The RHS of an assignment is evaluated _before_ the LHS so at the point of compilation of the anonymous function at line 4 there is no local p. However, 
```Lua
do
  local c, i, p = coroutine, 0
  p = c.create(function()
    local function resume()
```
works fine.  If you list off the VM bytecode in both using `luac.cross`, then in this second version `resume()` has two upvals `c` and `p` and runs fine.

The real subtext of this example is addressing one of Johny's worries: that upvals wouldn't work across threads -- they do fine.  Here we are explicitly using the fact that the tasks are delivered from L0 so that a coroutine can use this tasking mechanism to work in a "batch" mode over multiple task epochs.

markfinn:
If I remember your comments from the earlier task list discussion, it is possible for a post to fail when the queue is full, but I don't see that being returned here as a return code or error.  Should it?

TerryE:
@markfinn  Mark welcome to our discussions and with such an apt point. You are quite right -- we should not ignore the queue full scenario. 

However a correctly framed app should _not_ encounter this as you will have built up a queue of 8 tasks pending, and that most programmers would tend not to bother checking here.  My suggestion is that this should throw a Lua error under these circumstances rather than returning a status. If the programmer does want to catch this at an application Leven then there is always `pcall()`.

app/modules/node.c
```diff
@@ -448,6 +449,39 @@ static int node_compile( lua_State* L )
...
 }
 
+// Task callback handler for node.tasknow()
+static task_handle_t do_node_task_handle;
+static void do_node_task (task_param_t task_fn_ref, uint8_t prio)
+{
+  UNUSED(prio);
+  lua_State* L = lua_getstate();
+  lua_rawgeti(L, LUA_REGISTRYINDEX, (int)task_fn_ref);
+	luaL_unref(L, LUA_REGISTRYINDEX, (int)task_fn_ref);
```
jmattsson: Something's odd about the indentation here & below. I'm not too fussed whether it's tabs or spaces, but consistency would be handy :)

app/modules/node.c
```diff
@@ -448,6 +449,39 @@ static int node_compile( lua_State* L )
...
+  lua_pushinteger(L, prio);
+	lua_call(L, 1, 0);
+}
+
+// Lua: node.tasknow(task_cb)) -- schedule a task for execution next
```
jmattsson: Isn't it `node.task.post(foo)` now?

app/modules/node.c
```diff
@@ -448,6 +449,39 @@ static int node_compile( lua_State* L )
...
+    luaL_argcheck(L, priority <= TASK_PRIORITY_HIGH, 1, "invalid  priority");
+    Ltype = lua_type(L, ++n);
+  }
+  luaL_argcheck(L, Ltype == LUA_TFUNCTION || Ltype == LUA_TLIGHTFUNCTION, n, "invalid function");
+	lua_pushvalue(L, n);
+
+	int task_fn_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+
+  if (!do_node_task_handle)  // bind the task handle to do_node_task on 1st call
+    do_node_task_handle = task_get_id(do_node_task);
```
jmattsson: In hindsight, should `task_get_id()` really have been called `task_get_handle()` or perhaps even something like `task_create()`?

app/modules/node.c
```diff
@@ -448,6 +449,39 @@ static int node_compile( lua_State* L )
...
+
+  task_post(priority, do_node_task_handle, (task_param_t)task_fn_ref);
```
jmattsson: Verify we didn't overrun the q?

TerryE:
OK, I've added some changes to address these comments.  @markfinn Mark's comment plus a  `for k,v in pairs(debug.getregistry()) do print(k,v) end` check also threw up that I was leaking a cb function because I was unreferencing the cb before throwing the error.  So running this script does what I think that it should.  I leave it for you to see if you understand why I am doing this.
```Lua
do 
  local function breeder(parent)
    print ('called from',parent)
    if parent:len() > 14 then return end
    local nt=node.task
    for i = nt.LOW_PRIORITY, nt.MEDIUM_PRIORITY do
      local this = parent..'-'..i
      local st,err = pcall(nt.post, i, function() print ('running '..this)breeder(this) end)
      print(this, st and ' posted' or ' failed')
    end
  end
  breeder('root')
end
```

app/modules/node.c
TerryE: > In hindsight, should `task_get_id()` really have been called `task_get_handle()` or perhaps even something like `task_create()`?

@jmattsson @pjsg Johny,this would be a trivial fix because at the moment only you, Philip and me use this.  I did toss-up on what to name this, but if either of you feel strongly then make the change.

app/modules/node.c
```diff
@@ -448,6 +449,43 @@ static int node_compile( lua_State* L )
...
+{
+  int n = 1, Ltype = lua_type(L, 1);
+  unsigned priority = TASK_PRIORITY_MEDIUM;
+  if (Ltype == LUA_TNUMBER) {
+    priority = (unsigned) luaL_checkint(L, 1);
+    luaL_argcheck(L, priority <= TASK_PRIORITY_HIGH, 1, "invalid  priority");
+    Ltype = lua_type(L, ++n);
+  }
+  luaL_argcheck(L, Ltype == LUA_TFUNCTION || Ltype == LUA_TLIGHTFUNCTION, n, "invalid function");
+    lua_pushvalue(L, n);
```
pjsg: Can you fix the indentation?

app/modules/node.c
```diff
@@ -448,6 +449,43 @@ static int node_compile( lua_State* L )
   return 0;
 }
 
+// Task callback handler for node.task.post()
+static task_handle_t do_node_task_handle;
+static void do_node_task (task_param_t task_fn_ref, uint8_t prio)
+{
+  UNUSED(prio);
```
pjsg: prio is actually used....

TerryE: Bugger, it wasn't when I first wrote it and its just to avoid a compiler warning. I will be at my dev laptop sometime tomorrow.

app/modules/node.c
TerryE: Get your own back time? :laughing:

docs/en/modules/node.md
```diff
@@ -403,3 +403,44 @@ provides more detailed information on the EGC.
...
+If the priority is omitted then  this defaults  to `node.task.MEDIUM_PRIORITY`
+
+####  Returns
+`nil`
+
+#### Example
+```lua
+for i = node.task.LOW_PRIORITY, node.task.HIGH_PRIORITY do 
+  node.task.post(i,function(p2)
+    print("priority is ..p2)
```
pjsg: Missing double quote to close the string

TerryE:
Since I have the support of two other committers / contributers, and other are waiting for this patch, I will merge this -- before Phil picks up any other comment typos or indentation cockups :wink:

PR #1110, opened 2016-03-03
Omit all the PWM code if the PWM module is not included

### LINKED ISSUE #1101: Dead code pwm in firmware
This one was thrown up by  #1100.  At the moment the platform.h tier declares the platform abstraction for GPIO, SPI, UART, PMW, Sigma_delta, I2C and Flash which are all implemented in `platform.c`.  It also declares Eth and ADC which aren't implemented at all!   

The current implementation results in most of these functions being optimised out by the loader as unused if not referenced, however the pwm_init is done as part of platform_init() and hence the some of the  PWM driver and platform_pwm code gets mapped in -- roughly 0.5Kb iram0 and 2Kb irom -- whether or not pwm is used.  Likewise the gpio system also calls pwm close functions, so IMO its worth considering moving these lazily into the first routines that need them or suppress this logic if the PWM module is not enabled. Thoughts?

pjsg:
This resolves #1101 -- it turned out pretty simple.

Also, I fixed the compiler warning that annoyed me in every build (the warning about unsigned < 0 is always true.)

### DIFF
--- a/app/modules/pwm.c
+++ b/app/modules/pwm.c
@@ -122,6 +122,11 @@ static int lpwm_getduty( lua_State* L )
   return 1;
 }
 
+int lpwm_open( lua_State *L ) {
+  platform_pwm_init();
+  return 0;
+}
+
 // Module function map
 static const LUA_REG_TYPE pwm_map[] = {
   { LSTRKEY( "setup" ),    LFUNCVAL( lpwm_setup ) },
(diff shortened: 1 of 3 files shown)


TerryE:
Thanks for fixing that bloody warning.  It got to me too!

eku:
:+1: thanks @pjsg

jmattsson:
:heart: for shutting up that warning, it's been on my to-do list too!

PR #1113, opened 2016-03-03
Move modules list from wiki to repo

marcelstoer:
As discussed...


marcelstoer:
> Why treat our regular contributors this way?

It was absolutely not my intention to deny our regular contributors the limelight and the recognition they surely deserve. I see your point Terry.

Your comments made me go back to square one and think about the reason(s) I *didn't* want to have the file linked from the public docs' navigation tree. I thought "Why have an extra file listing all the modules? We have them in `mkdocs.yml` and users them in the navigation tree". And then something clicked...

I think I should dump the PR, and the modules list with it, and start from scratch.

Why don't we add a "meta-info" block to the top of every `<module>.md` page just beneath the title that looks something like this:

> # HX711 Module
> 
> Code       | Since         | Origin / Contributor | Maintainer
> ------------- | :------------ | :----------| :-----------
> [Source](https://github.com/nodemcu/nodemcu-firmware/commits/dev/app/modules/hx711.c)  | 2015&#8209;10&#8209;09 | [Chris Takahashi](https://github.com/christakahashi) | [Chris Takahashi](https://github.com/christakahashi)
> 
> This module provides access to an [HX711 load cell amplifier/ADC](https://learn.sparkfun.com/tutorials/load-cell-amplifier-hx711-breakout-hookup-guide). The HX711 is an inexpensive 24bit ADC .....

I believe that would make a lot more sense. The order of the columns is of course debatable (I'd put 'Code' right-most).
The only issue I see is one that we'll have to fix anyway once we move it to the repo. The source link must be relative and not point to the code in the `dev` branch as it's now.

TerryE:
Normalising this as you suggest makes sense.  We don't necessarily need this at the top, but some sort of standard section at the bottom or the top giving the major change history would be good.  If it is standard then a quick bit of perl-foo -- or whatever your hacking poison is -- could be used to mine it.

marcelstoer:
> giving the major change history would be good

Tha'ts a valid point IMO (no surprise I think that way because I had raise the issue of a change log before :wink:) but that should be a separate discussion / PR.

PR #1116, opened 2016-03-04
Fix crash bug when HTTP response from server is NULL.

vowstar:
Avoid req->buffer to NULL.
Avoid body to NULL.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -21,62 +21,63 @@
 
 /* Internal state. */
 typedef struct request_args_t {
-	char		* hostname;
-	int		port;
-	bool		secure;
-	char		* method;
-	char		* path;
-	char		* headers;
-	char		* post_data;
-	char		* buffer;
-	int		buffer_size;
-	int		timeout;
-	os_timer_t	timeout_timer;
+	char            * hostname;
+	int             port;
+	bool            secure;
+	char            * method;
+	char            * path;
+	char            * headers;
+	char            * post_data;
+	char            * buffer;
+	int             buffer_size;
+	int             timeout;
+	os_timer_t      timeout_timer;
 	http_callback_t callback_handle;
 } request_args_t;
 
-static char * ICACHE_FLASH_ATTR esp_strdup( const char * str )
+static char * ICACHE_FLASH_ATTR esp_strdup(const char * str)
 {
-	if ( str == NULL )
-	{
-		return(NULL);
+	if (str == NULL) {
+		return (NULL);
 	}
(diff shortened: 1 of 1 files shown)


app/http/httpclient.c
```diff
@@ -218,297 +219,297 @@ static int ICACHE_FLASH_ATTR http_chunked_decode( const char * chunked, char * d
...
+		int             http_status     = -1;
+		char            * body          = "";
+
+		/* Turn off timeout timer */
+		os_timer_disarm(&(req->timeout_timer));
+
+		if (req->buffer == NULL) {
+			HTTPCLIENT_DEBUG("Buffer shouldn't be NULL\n");
+			/* To avoid NULL buffer */
+			req->buffer = "";
```
pjsg: It looks as though there is a 'free' call below that will be invoked on this data. Not good.

vowstar: We tested these on a special server, which always return 404 HTTP status code but have no body(no body means body length is 0). 
In this library the goal is "stable & easy to use", return NULL directly is OK but it will take extra work to user to check NULL pointer. 
In espressif's SDK, os_strlen("") is 0, but os_strlen(NULL) will cause a reboot due to read Illegal address. 
For example, if a response is 
`const char * resp;`
if it is NULL directly, user can't do such as os_printf("%s\n", resp), or will cause crash. User must check if(NULL != resp).
If a user just want's debug and print this response, they must use this style:
```
if(NULL == resp) {
    os_printf("");
} else {
    os_printf("%s", resp);
}
```
If using an empty string "" instead NULL, it just 
`    os_printf("%s", resp);`
So it can make user easy and reduced probability of bug when user forgot check this pointer.
That's why I changed this code, see also https://github.com/vowstar/esphttpclient/commit/5434b5178d9866e6203941c44acf6b00375c6009
This code fixed that when a server response NULL string, nodemcu firmware can handle it correctly and not crash to reboot.

pjsg: I'm not saying that changing null to an empty string is wrong, but calling the C runtime 'fee' on a statically allocated string is WRONG (often fatally wrong). This is heap corruption.

You could allocate one byte and set it to zero and make a dynamically allocated empty string. This could be freed later. There may be a better approach.

marcelstoer:
@vowstar Please consider Philip's remarks, rebase on current `dev` and force-push to GitHub.

pjsg:
I've submitted a rather simpler PR #1212 that I think takes care of this.

marcelstoer:
Thanks Philip, let's track the issue over at your new PR.

PR #1120, opened 2016-03-05
Add meta-info block to every module doc page

marcelstoer:
Follow-up to #1113 since I messed up the Git history in that old branch.

A kind of "preview" is available at http://msr-nodemcu.readthedocs.org/en/dev-modules-list/. In the preview it all works except for the source link because repository (static) and branch (dynamic) don't match. It'll be fine in the public version.


docs/en/modules/ucg.md
```diff
@@ -1,4 +1,8 @@
 # ucg Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2015-08-05 | [Oli Kraus](https://github.com/olikraus/u8glib), [Arnim Läuger](https://github.com/devsaurus) | [Arnim Läuger](https://github.com/devsaurus) | [ucglib](../../../app/ucglib/)|
```
devsaurus: C&P error? Please link to Oli's ucglib.

marcelstoer: Thanks!

docs/en/modules/rc.md
```diff
@@ -0,0 +1,4 @@
+# RC Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2015-06-12 | [Mike Wen](https://github.com/mikewen) | - | [rc.c](../../../app/modules/rc.c)|
```
devsaurus: newline?

devsaurus:
:+1:

docs/en/modules/wifi.md
```diff
@@ -1,4 +1,8 @@
 # WiFi Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2015-05-12 | [dnc40085](https://github.com/dnc40085) | [dnc40085](https://github.com/dnc40085) | [wifi.c](../../../app/modules/wifi.c)|
```
dnc40085: I believe it was funshine that originally wrote the wifi module. I wouldn't mind taking the credit for writing it though :grinning:

marcelstoer: Thanks. Turns out that when I checked the history of https://github.com/marcelstoer/nodemcu-firmware/commits/dev-modules-list/app/modules/wifi.c I had missed that "Older" button that brings you back to the real origins of the module.

jmattsson:
Good stuff! Would it be hard to make the meta table be full width to match to ToC?

marcelstoer:
> Would it be hard to make the meta table be full width to match to ToC?

No, but
- I personally don't like that (this is not to say I wouldn't be willing to change it if you all agree with Johny)
- the ToC isn't full width, with this theme all tables use as little space as necessary

[meta-data-and-toc-samples.pdf](https://github.com/nodemcu/nodemcu-firmware/files/161260/meta-data-and-toc-samples.pdf)

TerryE:
> meta-data-and-toc-samples.pdf

My vote is the 1st, non-full width.  I think that this looks better.

dnc40085:
:+1:  for the non-full width version.

jmattsson:
I liked the spacier version (thanks for the preview!), but clearly I'm in the minority here.

Issue #1122, opened 2016-03-06
MQTT CONNACK


jfollas:
Recently at work, we got into a situation where NodeMCU was not connecting to our MQTT broker, and there was absolutely no feedback as to why. Turning on debugging showed that it kept trying to connect every second (as it should, since autoconnect was enabled), but again, we couldn't tell what was failing other than seeing it keep trying to connect.

It turns out that the broker was replying as it should, stating in the CONNACK that the password was invalid. But, there was absolutely no way to receive that feedback from the Lua side because CONNACK is not being processed by our MQTT client.

CONNACK return values are declared on pages 32-33 of OASIS spec:

MQTT 3.1.1 Specs: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.pdf

pjsg:
PR merged

PR #1123, opened 2016-03-06
MQTT Client - CONNACK processing

jfollas:
- Process the CONNACK message received from the broker after Connect
- Provide feedback to Lua via failure callback on client:connect()
- Also provide failure information for other situations not covered by CONNACK


jfollas:
Re: Issue #1122

pjsg:
Thank you for this.

Can you create a function that calls the callback (if present) and invoke that in all the places that you added -- rather than duplicating the code. There is (already) quite a lot of duplicate code in mqtt.c!

pjsg:
Thank you. This looks plausible. We still have a significant issue in the MQTT code that it saves and uses lua_State pointers later. I suspect that we want to do that cleanup later....

TerryE:
> We still have a significant issue in the MQTT code that it saves and uses lua_State pointers later. I suspect that we want to do that cleanup later....

Phil, this is pretty straight forward, so it would be nice if you include this as soon as convenient for you. Many thanks Terry

pjsg:
I added a PR for this #1129

pjsg:
@jfollas Can you resolve the conflicts please?

jfollas:
@pjsg Rebase complete.

pjsg:
Looks good to me..... Merged

PR #1125, opened 2016-03-07
Adds support for verifying the certificate offered by an SSL server.

pjsg:
This is a partial SSL fix. It adds support for verifying server certificates. The code for using client certificates for authentication is present, but not included in the build. It appears that there are problems in the underlying Espressif SDK that prevent it working.

Yes, the interface to the underlying SDK is baroque.

There are a few changes to the http client that I made to try and fix the client cert auth problem, but it didn't seem to help. I fixed another buffer overrun there as well.

I also included the latest versions of the Espressif scripts for making certs.....

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ CCFLAGS += 			\
 #	-Wall			
 
 ifneq ($(wildcard $(TOP_DIR)/user_config.h),)
-INCLUDES := $(INCLUDES) -include "$(TOP_DIR)/user_config.h"
+INCLUDES += -include "$(TOP_DIR)/user_config.h"
 endif
 
 CFLAGS = $(CCFLAGS) $(DEFINES) $(EXTRA_CCFLAGS) $(STD_CFLAGS) $(INCLUDES)
@@ -175,7 +175,7 @@ $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
 # Should be done in top-level makefile only
 #
 
-all:	sdk_extracted .subdirs $(OBJS) $(OLIBS) $(OIMAGES) $(OBINS) $(SPECIAL_MKTARGETS)
+all:	sdk_extracted pre_build .subdirs $(OBJS) $(OLIBS) $(OIMAGES) $(OBINS) $(SPECIAL_MKTARGETS)
 
 .PHONY: sdk_extracted
 
@@ -221,6 +221,17 @@ sinclude $(DEPS)
 endif
 endif
 endif
+
+.PHONY: pre_build
+
+pre_build:
+ifneq ($(wildcard $(TOP_DIR)/server-ca.crt),)
+	python $(TOP_DIR)/tools/make_server_cert.py $(TOP_DIR)/server-ca.crt > $(TOP_DIR)/app/modules/server-ca.crt.h
+DEFINES += -DHAVE_SSL_SERVER_CRT=\"server-ca.crt.h\"
+else
+	@-rm -f $(TOP_DIR)/app/modules/server-ca.crt.h
+endif
+
 
 $(OBJODIR)/%.o: %.c
 	@mkdir -p $(OBJODIR);
--- a/app/modules/net.c
+++ b/app/modules/net.c
@@ -3,6 +3,7 @@
 #include "module.h"
 #include "lauxlib.h"
 #include "platform.h"
+#include "lmem.h"
 
 #include "c_string.h"
 #include "c_stdlib.h"
@@ -13,17 +14,18 @@
 #include "espconn.h"
 #include "lwip/dns.h" 
 
-#ifdef CLIENT_SSL_ENABLE
-unsigned char *default_certificate;
-unsigned int default_certificate_len = 0;
-unsigned char *default_private_key;
-unsigned int default_private_key_len = 0;
-#endif
-
 #define TCP ESPCONN_TCP
 #define UDP ESPCONN_UDP
 
 static ip_addr_t host_ip; // for dns
+
+#ifdef HAVE_SSL_SERVER_CRT
+#include HAVE_SSL_SERVER_CRT
+#else
+__attribute__((section(".servercert.flash"))) unsigned char net_server_cert_area[INTERNAL_FLASH_SECTOR_SIZE];
+#endif
+
+__attribute__((section(".clientcert.flash"))) unsigned char net_client_cert_area[INTERNAL_FLASH_SECTOR_SIZE];
 
 #if 0
 static int expose_array(lua_State* L, char *array, unsigned short len);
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -222,6 +222,17 @@ endif
 endif
 endif
 
+.PHONY: pre_build
+
+pre_build:
+ifneq ($(wildcard $(TOP_DIR)/server-ca.crt),)
+	python $(TOP_DIR)/tools/make_server_cert.py $(TOP_DIR)/server-ca.crt > $(TOP_DIR)/app/modules/server-ca.crt.h
+DEFINES := $(DEFINES) -DHAVE_SSL_SERVER_CRT=\"server-ca.crt.h\"
```
jmattsson: Is it safe to := here? It won't interfere with further assignments/evaluations?

app/modules/net.c
```diff
@@ -1432,6 +1434,203 @@ static int net_multicastLeave( lua_State* L )
...
 
+// Returns NULL on success, error message otherwise
+static const char *append_pem_blob(const char *pem, const char *type, uint8_t **buffer_p, uint8_t *buffer_limit, const char *name) {
+  char unb64[256];
+  memset(unb64, 0xff, sizeof(unb64));
+  int i;
+  for (i = 0; i < 64; i++) {
+    unb64["ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"[i]] = i;
+  }
+
```
jmattsson: Don't we have existing base64 routines that could be reused? Wasn't @TerryE doing that for the `encoder` module?

app/modules/net.c
```diff
@@ -1432,6 +1434,203 @@ static int net_multicastLeave( lua_State* L )
...
+    return "No PEM header";
+  }
+
+  if (strncmp(pem + 11, type, strlen(type))) {
+    return "Wrong PEM type";
+  }
+
+  pem = strchr(pem, '\n');
+  if (!pem) {
+    "Incorrect PEM format";
```
jmattsson: Missing a `return` here, no?

app/modules/net.c
```diff
@@ -1432,6 +1434,203 @@ static int net_multicastLeave( lua_State* L )
...
+    const char *error = append_pem_blob(pem, types[argno - 1], &buffer, buffer_limit, names[argno - 1]);
+    if (error) {
+      luaM_free(L, buffer_base);
+      return error;
+    }
+  }
+
+  if (argno <= lua_gettop(L)) {
+    luaM_free(L, buffer_base);
+    return "Extra arguments provided";
```
jmattsson: In general I'm not a fan of testing for extra arguments. In Lua (and especially eLua) it's really handy to be able to chain calls without declaring explicit temporaries and just let extra args "fall off" when/where they're not needed. My opinion is that this outweighs hand-holding (or nose-smacking) the user :)

ld/nodemcu.ld
```diff
@@ -95,7 +97,7 @@ SECTIONS
 
     _irom0_text_end = ABSOLUTE(.);
     _flash_used_end = ABSOLUTE(.);
-  } >irom0_0_seg :irom0_0_phdr
+  } >irom0_0_seg :irom0_0_phdr =0xffffffff
```
jmattsson: I'm not familiar with this particular syntax - is that setting the fill value for the segment?

app/modules/net.c
```diff
@@ -1432,6 +1434,203 @@ static int net_multicastLeave( lua_State* L )
...
+  if ((flash_offset & 0xfff) || flash_offset > 0xff000 || INTERNAL_FLASH_SECTOR_SIZE != 0x1000) {
+    // THis should never happen
+    return luaL_error( L, "bad offset" );
+  }
+
+  if (lua_type(L, 1) == LUA_TSTRING) {
+    const char *types[2] = { "CERTIFICATE", NULL };
+    const char *names[1] = { "certificate" };
+
+    const char *error = fill_page_with_pem(L, flash_offset, types, names);
```
jmattsson: I don't suppose you could check whether the new data matches what is already there, and only erase+write the cert page if it has changed? Both in the interest of flash wear as well as runtime speed.

docs/en/modules/net.md
```diff
@@ -282,3 +282,100 @@ Sets the IP of the DNS server used to resolve hostnames. Default: resolver1.open
...
+This controls certificate verification when SSL is in use. 
+
+## net.cert.verify()
+
+Controls the vertificate verification process when the Nodemcu makes a secure connection.
+
+#### Syntax
+`net.cert.verify(enable)`
+
+`net.cert.verify(pemdata)`
```
jmattsson: How would one inspect the installed CA cert? Presumably one would want to avoid rewriting it if it's already in there.

app/modules/net.c
```diff
@@ -1539,6 +1738,14 @@ static const LUA_REG_TYPE net_array_map[] = {
 };
 #endif
 
+static const LUA_REG_TYPE net_cert_map[] = {
+  { LSTRKEY( "verify" ), 	LFUNCVAL( net_cert_verify ) },  
+#ifdef CLIENT_SSL_CERT_AUTH_ENABLE
+  { LSTRKEY( "auth" ),		LFUNCVAL( net_cert_auth ) },
```
jmattsson: This command does not appear to have documentation?

jmattsson:
Comments aside, this is good (some would say critical) functionality. +1 from me.

Makefile
pjsg: I think it is safe -- but I should have used +=, so I changed to that.

app/modules/net.c
pjsg: Yes -- I started out with that module, but it throws lua errors on invalid input (which would cause a memory leak in my layer). Also, the PEM file has embedded newlines within it that would cause problems. Originally I constructed a single base 64 string and passed that over, but it is simpler to just do the conversion....

ld/nodemcu.ld
pjsg: Yes -- that is the fill value. Reduces stress on the flash. Especially when there is the sector used for the certificate.

docs/en/modules/net.md
pjsg: I did the change that you suggested and don't flash the cert if it was already there. It changes the time spent in the call from 80ms down to 3ms! There isn't a way to see the cert that is stored --partly because it is just a blob, and partly because I dodn't want people to inspect the client private key (which will use all the same mechanisms).

app/modules/net.c
pjsg: You are correct. I have a version of the documentation that does include it, but this command doesn't actually work -- the espressif code throws a null pointer exception. I'm trying to get information out of espressif about whether this functionality ever worked. I felt that once I can make it work, I'll add the documentation section. The default for the define is to not include this code.

pjsg:
I fixed the other issues. I'll squash the commits nearer the merge time -- but I suspect that doing that will lose the line comments.

TerryE:
>  I suspect that doing that will lose the line comments.

They get folded away, but there's a link on the right to expand them.

jmattsson:
Happy to merge.

pjsg:
squashed into a single commit

jmattsson:
Turns out this broke building with GNU make 4.1 here, so I directly committed 731264776de9e042ee221686f61733b562d97f9f to not leave `dev` in a wonky state. A different approach/fix might be better.

TerryE:
@pjsg @jmattsson  I've just noticed that we've added a lot of new SSL related code to `net.c` without bracketing it by `#ifdef CLIENT_SSL_ENABLE`.  OK, I know that the linker may optimmise some of these away, but Phil did check the mapfile for a none SSL build to verify this?

pjsg:
When I wrote the code, I did validate that the new stuff was not present in the builds that I did. However, it appear that the httpclient code does bring in all the ssl stuff whether or not CLIENT_SSL_ENABLE is set.....

TerryE:
Philip,

> Yes -- I started out with that module, but it throws lua errors on invalid input (which would cause a memory leak in my layer)

I was just going through this code and previous comments.  I don't see that this is really an issue as this is why we have `lua_pcall()` to catch these if we need to recover for things like preventing memory leaks.  

I also think that all of this code really doesn't belong in the net module.  This certificate stuff is really adding functionality to the SDK,IMO.   I would prefer to calve it all off into a separate `net_ssl.c` or `net_utils.c` as `net.c` is already too long and too convolved,

jmattsson:
I don't think it's feasible to have it completely split off due to the Lua API where we just take a flag for secure/plain, but other than for that layer a `net_ssl.c` (or `net_tls.c` or even `net_secure.c`) sounds reasonable. Maybe a `net_common.c` for any shared tls/non-tls stuff even.

TerryE:
I have a split version in local dev.  It's not a total spilt just this specialist code that's nothing really to do with the nodeMCU net abstraction layer, these SSL specific SDK extensions.

PR #1126, opened 2016-03-07
Fixed race-crash in SNTP module.

jmattsson:
The SNTP client could crash & burn if handed a DNS name which took longer than ~6sec to be looked up. At that point the resend time would already have expired and cleaned up the state, so when the dns(not)found handler got called, it'd null-pointer deref.

This commit fixes that by not starting the resend timer until we know where we want to send in the first place, which also stops us from sending out a bunch of requests to the previous server while waiting for the DNS to come back.

Also included is removal of stashed-away-L and use of lua_getstate() instead.

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -94,6 +94,8 @@ typedef struct
 static sntp_state_t *state;
 static ip_addr_t server;
 
+static void on_timeout (void *arg);
+
 static void cleanup (lua_State *L)
 {
   os_timer_disarm (&state->timer);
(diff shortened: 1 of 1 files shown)


pjsg:
:+1:

TerryE:
@pjsg, Philip, if you've not noticed you can do the merge yourself now. :clap: :clap: 

Having looked at the code, it looks like a sensible fix but I haven't used this module yet so I left it to another committer to do the merge. Over to you or Johny.

PR #1127, opened 2016-03-07
Change a number of default values

### LINKED ISSUE #872: Changing the default user settings.

marcelstoer:
fixes #872

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ The following sections explain some of the options you have if you want to [buil
 
 ### Select Modules
 
-Disable modules you won't be using to reduce firmware size and free up some RAM. The ESP8266 is quite limited in available RAM and running out of memory can cause a system panic. 
+Disable modules you won't be using to reduce firmware size and free up some RAM. The ESP8266 is quite limited in available RAM and running out of memory can cause a system panic. The default configuration is designed to run on all ESP modules including the 512 KB modules like ESP-01 and only includes general purpose interface modules which require at most two GPIO pins.
 
 Edit `app/include/user_modules.h` and comment-out the `#define` statement for modules you don't need. Example:
 
@@ -89,8 +89,8 @@ Identify your firmware builds by editing `app/include/user_version.h`
 
 ### Set UART Bit Rate
 
-The initial baud rate at boot time is 9600bps. You can change this by
-editing `BIT_RATE_DEFAULT`  in `app/include/user_config.h`:
+The initial baud rate at boot time is 115200bps. You can change this by
+editing `BIT_RATE_DEFAULT` in `app/include/user_config.h`:
 
 ```c
 #define BIT_RATE_DEFAULT BIT_RATE_115200
@@ -103,5 +103,3 @@ To enable runtime debug messages to serial console edit `app/include/user_config
 ```c
 #define DEVELOP_VERSION
 ```
-
-`DEVELOP_VERSION` changes the startup baud rate to 74880bps.
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -21,9 +21,8 @@
 #ifdef DEVELOP_VERSION
 #define NODE_DEBUG
 #define COAP_DEBUG
-#define BIT_RATE_DEFAULT BIT_RATE_74880
 #else
-#define BIT_RATE_DEFAULT BIT_RATE_9600
+#define BIT_RATE_DEFAULT BIT_RATE_115200
 #endif /* DEVELOP_VERSION */
 
 
@@ -54,7 +53,7 @@
 #define NO_INTR_CODE inline
 #endif
 
-#define CLIENT_SSL_ENABLE
+//#define CLIENT_SSL_ENABLE
 //#define MD2_ENABLE
 #define SHA2_ENABLE
 
--- a/app/include/user_modules.h
+++ b/app/include/user_modules.h
@@ -13,21 +13,25 @@
 
 #ifndef LUA_CROSS_COMPILER
 
+// The default configuration is designed to run on all ESP modules including the 512 KB modules like ESP-01 and only
+// includes general purpose interface modules which require at most two GPIO pins.
+// See https://github.com/nodemcu/nodemcu-firmware/pull/1127 for discussions.
+// New modules should be disabled by default and added in alphabetical order.
 #define LUA_USE_MODULES_ADC
 //#define LUA_USE_MODULES_AM2320
 //#define LUA_USE_MODULES_APA102
 #define LUA_USE_MODULES_BIT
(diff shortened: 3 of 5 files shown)


TerryE:
Marcel, I have been thinking about this. The questions we should be asking ourselves are: why do w have a default setting and what should be in it?

I feel that the answer to the first is that it is a build configuration that enables a beginner to make a basic firmware build that fits in a 512 KB module, with a useful minimal SPIFFS (say 64 KB) and is sufficiently functional to allow initial familiarisation with bread-boarded modules or a Devkit. 

On reflection, SPI isn't that common and needs more than the pins available on the ESP-01. I am not sure that i would include it.

dnc40085:
I give this a :+1: because I develop on a 512k esp, for the faster upload, and it gets a bit irritating having to change `user_config.h` and `user_modules.h` every time I do a reset on my local branch.

It would also be nice to update `.cproject` for eclipse to have the paths and symbols set for the include directories.

Just for fun I figured I'd check to see how much free space is available with the new defaults:
* With the new defaults, `file.fsinfo` reports 83081 bytes.
* With the new defaults - DHT + [PR 1018](https://github.com/nodemcu/nodemcu-firmware/pull/1018), `file.fsinfo` reports 83081 bytes.
* With the new defaults + PR 1018 (with wifi station status monitor and wifi disconnect reasons are disabled), `file.fsinfo` reports 83081 bytes.

marcelstoer:
It's fine for me to target 512 KB modules like ESP-01. However, I believe beginners nowadays go for USB-ready modules such as NodeMCU devkits or WeMos D1 mini. It's all just guess work and we'll never know for sure, so we might as well roll the dice.

Since we don't provide up-to-date downloadable binaries anymore we seem to pretty much agree that whatever the modules included by default are that set has close to zero impact. It would just be nice to have a good enough guideline and explanation why a certain module is enabled/disabled by default.

Arnim says
> i2c, spi, ow, adc, and gpio are general purpose interface modules

You say
> SPI isn't that common and needs more than the pins available on the ESP-01

And I agree with both of you. So, how about

"The default configuration targets 512 KB modules like ESP-01 and includes general purpose interface modules which require at most two GPIO pins"

marcelstoer:
@dnc40085 don't you think reporting `node.heap` rather than `file.fsinfo` would provide more meaningful information?

marcelstoer:
> It would also be nice to update `.cproject` for eclipse to have the paths and symbols set for the include directories.

That's IMO OT for this PR but I've actually wanted to get rid of `.settings/`, `.cproject`, and `.project` for quite some time. To my knowledge none of the current committers uses Eclipse for this project so those artifacts obviously are unmaintained.

dnc40085:
>@dnc40085 thanks for testing but don't you think reporting node.heap rather than file.fsinfo would provide more meaningful information?

I was thinking more about the limited space aspect for the 512K ESPs. 

Here's the heap outputs:
* With the new defaults, `node.heap` = 45424 bytes.
* With the new defaults - DHT, `node.heap` = 45448 bytes. `file.fsinfo` = 83081
* With the new defaults - DHT + [PR 1018](https://github.com/nodemcu/nodemcu-firmware/pull/1018), `node.heap` = 45344 bytes.
* With the new defaults + PR 1018 (with wifi station status monitor and wifi disconnect reasons are disabled), `node.heap` = 45376 bytes.

>That's IMO OT for this PR

I just didn't think it was a big enough deal to open a separate issue for it, my apologies.

TerryE:
Marcel suggestion for a slight tweak to your working.  (The _italics_ is only to markup the changes.)

> The default configuration is _designed to run on all ESP modules including_ the 512 KB modules like ESP-01 and _only_ includes general purpose interface modules which require at most two GPIO pins

marcelstoer:
Thanks Terry for the tweak, makes sense. As seem to finally be approaching a common understanding where do we persists that? It's nice we got it here on GitHub but I wouldn't want to send people to #1127 if they ask about our defaults. Can I add the statement to `user_modules.h` as a comment just above the module definition? How about adding it also as a new paragraph to http://nodemcu.readthedocs.org/en/dev/en/build/?

jmattsson:
I'm sure Terry will be pleased to hear it, but the more I'm building NodeMCU, the less I'm liking user_modules.h. The ability to do e.g. `make MODULES="file gpio sntp"` or `make MODULES="@common.cfg @extra.cfg adc ow"` with no "default" config per se, though perhaps providing the common.cfg in the aforementioned example would be nice. The key aspect would be that there would be no header file listing the modules, nor one generated - any #defines needed would go on the command line. Keep the module list away from the code.

In the meantime, this PR seems a reasonable step. The cloud builder stats are our best source of guidance I believe.

pjsg:
At some point, one of my PRs got committed which accidentally included the feature that I use to configure local builds. 

If you have a user_config.h file in the root of the firmware, then it will be included everywhere. It can only add extra modules (and predefine other things). Since it isn't part of the repo, it persists across branch changes etc. It doesn't have to exist, and, if it doesn't exist, then everything is backwards compatible. It is driven with a little bit of makefile magic.

pjsg:
There doesn't appear to be much we can do about the heap size. The vast majority of the rest of the ram is consumed by the espressif code (and lwip). It will make a difference which modules we include only to the extent that bits of the sdk get included or not.

TerryE:
I think that there are two separate issues: is this PR a sensible step forward; should we move to a config approach. I would keep these separate. But I :+1: on both.

marcelstoer:
Terry, Philip, Johny do you have a preference as for where to document our default settings (https://github.com/nodemcu/nodemcu-firmware/pull/1127#issuecomment-193289044)?

jmattsson:
A note on the "why" in user_modules.h together with a "new modules should be disabled by default", plus a more friendly note plus pointer to cloud-builder in the docs proper?

marcelstoer:
All done, hope this is finally ready for merging.

Issue #986, opened 2016-01-28
MQTT: Subscribing to multiple topics at once loops forever


devsaurus:
I came across this during work on #985: If 3 or more topics are to be subscribed using the "tabular" variant of `m:subscribe()` then the process loops forever.

This works as expected:
```lua
> m = mqtt.Client("esp8266", 60, "", "")
> =m:connect(server_ip, 1883, 0, 1, function() print("connected") end)
true
> connected
=m:subscribe("test1", 1, function() print("subscribed test1") end)
true
> subscribed test1
=m:subscribe("test2", 1, function() print("subscribed test2") end)
true
> subscribed test2
=m:subscribe("test3", 1, function() print("subscribed test3") end)
true
> subscribed test3
=m:subscribe("test4", 1, function() print("subscribed test4") end)
true
> subscribed test4
=m:close()
true
```

While things begin to break with `m:subscribe()` for 3 topics:
```lua
> m = mqtt.Client("esp8266", 60, "", "")
> =m:connect(server_ip, 1883, 0, 1, function() print("connected") end)
true
> connected
=m:subscribe({tab2_1 = 1, tab2_2 = 1}, function() print("subscribed table with 2 elements") end)
true
> subscribed table with 2 elements
=m:subscribe({tab3_1 = 1, tab3_2 = 1, tab3_3 = 1}, function() print("subscribed table with 3 elements") end)
true
> =m:subscribe("test4", 1, function() print("subscribed test4") end)
false
> =m:close()
true
```

The callback for `print("subscribed table with 3 elements")` is never triggered and subsequent `m:subscribe()` fail. Closing and setting up the client from scratch is the only way to resolve the situation.

Mosquitto log shows:
```
1454019172: New connection from client_ip on port 1883.
1454019172: New client connected from client_ip as esp8266 (c1, k60).
1454019175: esp8266 1 tab2_2
1454019175: esp8266 1 tab2_1
1454019203: esp8266 1 tab3_2
1454019203: esp8266 1 tab3_1
1454019203: esp8266 1 tab3_3
1454019204: esp8266 1 tab3_2
1454019204: esp8266 1 tab3_1
1454019204: esp8266 1 tab3_3
1454019205: esp8266 1 tab3_2
1454019205: esp8266 1 tab3_1
1454019205: esp8266 1 tab3_3
1454019206: esp8266 1 tab3_2
1454019206: esp8266 1 tab3_1
1454019206: esp8266 1 tab3_3
[...]

PR #1129, opened 2016-03-07
Try to make qos~=0 work better

pjsg:
There are two things in the PR

* Make sure that we send the next message in the queue whenever possible. This *should* fix all the delays. I can send multiple messages per second with non zero QOS now with no obvious delays (and no queue buildup). 

* Make sure that we don't save the lua_State pointer and use it later. This fails in all sorts of horrible ways in the presence of coroutines.

As a benefit, this actually reduces the number of lines of code.....

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -55,7 +55,6 @@ typedef struct mqtt_state_t
 
 typedef struct lmqtt_userdata
 {
-  lua_State *L;
   struct espconn *pesp_conn;
   int self_ref;
   int cb_connect_ref;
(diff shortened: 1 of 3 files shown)


pjsg:
I've added a fix for Issue #986 as well.

TerryE:
Philip, I know nothing of this library.  Maybe Johny or Arnim, or one of the other MQTT contributors could look this over and comment then I'll commit it. @jmattsson @devsaurus @karrots?

devsaurus:
I'll check for regressions here.

devsaurus:
I can confirm that #986 is fixed.

On the downside it appears that my simple mqtt-driven application is broken by the PR. A sensor publishes messages every 5 minutes which are supposed to be picked up by the ESP. This doesn't work anymore as the mqt subscription won't fire the callback for all but the first message (a retained one).

I've isolated a simple testcase where the sensor is modelled by this csh script:
```csh
#!/bin/csh

set i = 0
while ( 1 )
        echo $i
        mosquitto_pub -i sensor -h server_ip -q 2 -r -t test_pr -m 0.0:0.0:$i.$i
        @ i = ( $i + 1 )
        sleep 5m
end
```

On the ESP side:
```lua
m = mqtt.Client("esp8266", 120, "", "")

m:on("message", function(conn, topic, data) 
    if data ~= nil then
        local arr = {}
        local i = 1
        data:gsub("[^:]+", function (a) arr[i]=a i=i+1 end)

        print(arr[1]..", "..arr[2]..", "..arr[3])
    end
end)

m:connect(server_ip, 1883, 0, 0, function(conn)
        print("connected, subscribing...")
        m:subscribe("test_pr", 2, nil)
end)
```

It works with the code on current `dev` branch.

pjsg:
I'll take a look at this in the morning....

pjsg:
I've fixed this. The bug turned out to be that it didn't send the PING on the two minute timeout and then it disconnected the session after a number of failed pings. The bug was an extra line of code that I had failed to remove. 

I will squash the commits once you are happy with it.

Sorry about that.

TerryE:
> I will squash the commits once you are happy with it.

Be a rebel -- don't bother :wink:

devsaurus:
Works fine now, thanks a lot! Green light from my side :+1:

PR #1132, opened 2016-03-08
Add some easy development debugging

pjsg:
This adds (under a new #define) two things:

* node.osprint(true/false) -- this enables/disables the espressif sdk logging. This is *really* useful when trying to debug certificate issues

* enables all the asserts

It leaves the bit rate alone. There is no extra output by default so all the existing upload tools continue to work. The problem with DEVELOP_VERSION is that it generates a lot of extra output....

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -604,6 +604,20 @@ static int node_egc_setmode(lua_State* L) {
   legc_set_mode( L, mode, limit );
   return 0;
 }
+//
+// Lua: osprint(true/false)
+// Allows you to turn on the native Espressif SDK printing
+static int node_osprint( lua_State* L )
+{
+  if (lua_toboolean(L, 1)) {
+    system_set_os_print(1);
+  } else {
+    system_set_os_print(0);
+  }
+
+  return 0;  
+}
+
 // Module function map
 
 static const LUA_REG_TYPE node_egc_map[] = {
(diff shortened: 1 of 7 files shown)


app/modules/node.c
```diff
@@ -604,6 +604,22 @@ static int node_egc_setmode(lua_State* L) {
...
   return 0;
 }
+//
+// Lua: osprint(true/false)
+// Allows you to turn on the native Espressif SDK printing
+static int node_osprint( lua_State* L )
+{
+  if (lua_toboolean(L, 1)) {
+    system_set_os_print(1);
+    os_install_putc1((void *)uart0_write_char);
```
jmattsson: Do we really need to switch around the putc1 handlers? `ets_printf()` and co happily honour `system_set_os_print()`. I'm not keen on suddenly using different pins, especially when those might be wired up to something completely unrelated...

jmattsson:
Handy, I wouldn't mind getting this in. There have been occasions where I wish I could've re-enabled `ets_printf()` at runtime.

app/modules/node.c
pjsg: This is bizarre, and maybe someone can explain it to me....

The driver/uart.c has

```
    // install uart1 putc callback
#ifndef NODE_DEBUG
    os_install_putc1((void *)uart1_write_char);
#endif
```

in it. I.e. it points the output to uart1 (and uart0 is the one connected to the USB). I don't know if this was done to really suppress all the SDK output, or why.

I'd be happy to remove this code, but I'm nervous that we might see output from it. The other alternative is just to make the no output function just drop characters.

Your call...

jmattsson: I'd recommend testing without the putc1 redirects. It would be *very* nice to not have to use those. If we do have to, using a null function would appear to be the safer option.

pjsg:
I removed all the redirects and I haven't seen any spurious output -- but maybe they were there for a reason.... If anybody spots any extra output, then they need to go back in.

TerryE:
Philip, are you happy to merge, or do you want to wait a day or so for feedback on your redirect point?  If you are happy, then do it on my behalf. //Terry

Issue #1134, opened 2016-03-08
Unsubscribe gpio.trig


UncleRedz:
This may have been an undocumented feature, but with the old gpio implementation, you could unsubscribe the callback for a specific pin by writing something like this, 

gpio.trig(2,"none")

As the word "none" was not recognised, the code simply removed the callback. 

Running the same code on dev now after TerryE rewrite, the ESP8266 instantly reboots. 

I don't see in the docs that there is any way to unsubscribe, but it would be nice if it was possible. At the very least, the instant reboot should be fixed. 

(On a side note, after very brief testing, gpio trig seems more stable now.) 

Thanks for your efforts with the NodeMCU firmware.

TerryE:
You just set the GPIO mode to other than interrupt, say input and this clears down any associated triggers.  When you say "instantly reboots" I think you mean " throws a Lua error and because I haven't done a protected call, this triggers an unprotected call exception which the RTS handles by doing a PANIC restart." 

See #1119

UncleRedz:
I don't think it throws a lua error, there is no panic message, just the short sequence of garbage characters and a reboot. Looks the same as a node.restart(). Which is why I described it as an instant reboot. It looks like there is an error case in the c code that isn't handled when an unexpected word is used for what to trigger on.

TerryE:
OK, I will take a look today.

TerryE:
I have just put a fix in for this.  See #1136

Sorry but I done want to allow "none".  The problem is that the previous algo defaulted to interrupts disabled if the option was one that it did know about.  Hence if you mistyped "up" say as "yp", this would fail silently and disable the trigger.  The correct action is to omit the argument.

TerryE:
@UncleRedz if you don't mind we're now tracking this through PR #1136 so I will close this.  Rgds Terry

TerryE:
See also #1141.  Your comments will be appreciated.

PR #1136, opened 2016-03-08
one line bugfix to GPIO.c -- Oops!

TerryE:
I left the NULL terminator off the end of a `luaL_checkoption()` options list.  Idiot.  How did I miss that?

### DIFF
--- a/docs/en/modules/gpio.md
+++ b/docs/en/modules/gpio.md
@@ -98,18 +98,18 @@ gpio.serout(1,1,{8,18},8) -- serial 30% pwm 38k, lasts 8 cycles
 
 ## gpio.trig()
 
-Establish a callback function to run on interrupt for a pin.
-
-There is currently no support for unregistering the callback.
+Establish or clear a callback function to run on interrupt for a pin.
 
 This function is not available if GPIO_INTERRUPT_ENABLE was undefined at compile time.
 
 #### Syntax
-`gpio.trig(pin, type [, function(level)])`
+`gpio.trig(pin, [type [, function(level)]})`
 
 #### Parameters
 - `pin` **1~12**, IO index, pin D0 does not support interrupt.
-- `type` "up", "down", "both", "low", "high", which represent rising edge, falling edge, both edge, low level, high level trig mode correspondingly.
+- `type` "up", "down", "both", "low", "high", which represent rising edge, falling edge, both edge, 
+low level, high level trig mode correspondingly. If the type is "none" or omitted then the
+callback function is removed and the interrupt disabled.
 - `function(level)` callback function when triggered. The gpio level is the param. Use previous callback function if undefined here.
 
 #### Returns
@@ -131,6 +131,9 @@ function pin1cb(level)
 end
 gpio.trig(pin, "down", pin1cb)
 
+```
+```Lua
+gpio.trig(4, 'none') -- remove any existing trigger
 ```
 #### See also
 [`gpio.mode()`](#gpiomode)
--- a/docs/en/modules/gpio.md
+++ b/docs/en/modules/gpio.md
@@ -103,14 +103,16 @@ Establish or clear a callback function to run on interrupt for a pin.
 This function is not available if GPIO_INTERRUPT_ENABLE was undefined at compile time.
 
 #### Syntax
-`gpio.trig(pin, [type [, function(level)]})`
+`gpio.trig(pin, [type [, callback_function]])`
 
 #### Parameters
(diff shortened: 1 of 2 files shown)


TerryE:
This is a one-line fix for #1134.  Thanks @UncleRedz. But I have just realised that I need to update the docs and take care of a resource leak opened by this path, so _don't merge until I've added these to the commit_.

One issue on the associated #1134 if the the OP used an undocumented side-effect of the original code that an invalid trigger level was ignored but any interrupt was disabled so using "none" for example turned off the trigger.  I would prefer not to remove the basic functionality, but tighten up on this default to make it more specific, e.g. "none" or omitting the argument disables the trigger but anything else raises an invalid argument error.

I also want to dereference the CB properly so that it can be GCed.

TerryE:
When this was a one-liner I was going to commit this myself, but this one is a sufficient change to merit dual key. This (1) fixes the crash thanks to the missing NULL terminator; (2) adds the ability to clear triggers, and (3) removes the previous leak where repeated triggers or clear actions didn't clear the Registry reference, thus preventing GC of the function

@pjsg you can have the privilege :-) assuming you find this acceptable of course.

docs/en/modules/gpio.md
```diff
@@ -98,18 +98,18 @@ gpio.serout(1,1,{8,18},8) -- serial 30% pwm 38k, lasts 8 cycles
...
-`gpio.trig(pin, type [, function(level)])`
+`gpio.trig(pin, [type [, function(level)]})`
 
 #### Parameters
 - `pin` **1~12**, IO index, pin D0 does not support interrupt.
-- `type` "up", "down", "both", "low", "high", which represent rising edge, falling edge, both edge, low level, high level trig mode correspondingly.
+- `type` "up", "down", "both", "low", "high", which represent rising edge, falling edge, both edge, 
+low level, high level trig mode correspondingly. If the type is "none" or omitted then the
+callback function is removed and the interrupt disabled.
 - `function(level)` callback function when triggered. The gpio level is the param. Use previous callback function if undefined here.
```
pjsg: The code (as written) requires you to specify the callback function if you specify trig of other than "none". My suspicion is that the documentation is probably more correct than the code....

TerryE: Are you looking at the same code as I submitted on this patch because it had been updated an hour or so before your comment? 

A value of "none" or omitted sets the type = `GPIO_PIN_INTR_DISABLE`, so the branch validating the callback function isn't taken and the cb ref is set to `LUA_NOREF`. I did try the test cases of setting a trigger, and mutilple times  and deleting the trigger with the ommitted and "none" parameters and they work fine as per the documentation. I don't understand your comment.

 Any old registry entry is unref'ed to allow correct GC to take place.

pjsg: From the documentation, the following ought to work:

gpio.trig(1, "both", function() print("edge") end)
gpio.trig(1, "up")

If the function isn't specified, then it persists from the previous call (according to the docs). The code will (I think) throw an error.

TerryE:
Philip, see #1141.  I won't close this PR, but lets suspend work until we get agreement on what we should be doing here.

docs/en/modules/gpio.md
```diff
@@ -118,20 +120,20 @@ This function is not available if GPIO_INTERRUPT_ENABLE was undefined at compile
...
-  du = tmr.now() - pulse1
-  print(du)
-  pulse1 = tmr.now()
-  if level == gpio.HIGH then gpio.trig(pin, "down") else gpio.trig(pin, "up") end
+do
+  -- use pin 1 as the input pulse width counter
+  local pin, pulse1, du, now, trig = 1, 0, 0, tmr.now, gpio.trig
+  gpio.mode(pin,gpio.INT)
+  local function pin1cb(level)
+    local pulse2 = now()
```
jmattsson: That should be `tmr.now()`, no?

UncleRedz: Looks like a local now() which was assigned tmr.now, a few lines up.

jmattsson: Whoopsie, missed that!

TerryE: Yup, I was just using the Lua example to show a few techniques that speed execution.this removed a couple of table look ups.

pjsg:
This looks good to me. Is this waiting on anything else?

TerryE:
Nothing in my view. Please merge.

UncleRedz:
+1

Issue #1117, opened 2016-03-04
CONTRIBUTING.md -- suggested change to working with Github process


TerryE:
At the moment the advice is:
- to avoid intermediate merge commits
- to squash your commits against a dev rebaseline.

I feel that this is a one-size fits all approach an it doesn't, as a single commit can make it very difficult to see what is what if the change is anything but trivial.   
1.  In #1028, I discuss that we've often got dead code inline with large blocks of `#if 0` or '//' commenting out,  The layout and spacing can be poor and what would be very useful commenting is omitted.  Also the most appropriate Lua argument checking API calls aren't used.  We should have a standard template and practice for laying out out modules. 
2.  I have suggested that if any of us does non-trivial work on a module then it makes sense to do a (1) pass at the same PR.
3.  After the first submission of the PR, we may get review comments that need incorporating.

My suggestion is that whilst we should expect contributors to squash working commits, it really makes it a lot easier to review changes if we split the PR into up to four commits, viz: (a) Formatting and dead code removal -- here in principle the binary output of the two versions would be unchanged.  (b) Tidy up and move to template compliant error checking, etc. (c) The functional changes that triggered the issues / work.  (d) The changes are a result of any issue and PR comments.

I would also suggest we adopt a naming convention for commits, E.g. PR title followed by code tidy / rework checking / main changes / Review.  

Lastly I am not a Github expert but what worries me about fetching dev and rebaselining against it is that this might back out other changes made in parallel.  My standard approach now is:
1.    If it's a complex change do it against a recent change and commit it.
2.   `git fetch nodemcu dev:dev-<some-branch>` where the branch name reflects the PR
3.   `git checkout -f :dev-<some-branch>`
4.   Either `git cherry-pick` commit at (1) or do a `git diff` on (1) to a `/tmp/xxx.diff` and patch the working branch with this.  If it is a simple change just do it. 
5.   Devise a test to test the change; do a make; blow to ESP and test it.
6.  Commit it.
7.   `git push terrye dev-<some-branch>:dev-<some-branch>`
8.   Use github interactively to issue the PR  against `nodemcu/nodemcu-firmware:dev`
9.   Update the raising issue and add any further b/g comment to the PR.
10.  Roll up any agreed PR comments into an additional commit on my local instance and push this to `terrye`. 

My aim here is to do (3) - (10) in one working session, say no more than a couple of hours.

[...]

PR #1138, opened 2016-03-08
Improve overall enduser_setup user experience.

robertfoss:
A single bug is fixed, a few features are added and overall the codebase has been worked through.

 - Added support for calls to /generate_204 that let's android know that the internet is accessible.
 - Added 10 second delay to the shutdown call to allow a final status update to be fetched by the client.
 - Added iframe to html to avoid having a form submission change the page.
 - Added support for dynamic /status responses.
 - Improved HTML appearance by removing AP-list button.
 - Improved CSS to center form, even when list of access points have loaded.
 - Improved debug prints to contain line numbers and not require lua_State*.
 - Fixed broken failure check when calling wifi_station_connect().

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -35,6 +35,7 @@
 
 #include "module.h"
 #include "lauxlib.h"
+#include "lmem.h"
 #include "platform.h"
 #include "c_stdlib.h"
 #include "c_stdio.h"
@@ -76,6 +77,7 @@ static const char dns_body[]   = { 0x00, 0x01, 0x00, 0x01,
 
 static const char http_html_filename[] = "index.html";
 static const char http_header_200[] = "HTTP/1.1 200 OK\r\nCache-control:no-cache\r\nContent-Type: text/html\r\n"; /* Note single \r\n here! */
+static const char http_header_204[] = "HTTP/1.1 204 No Content\r\n\r\n";
 static const char http_header_302[] = "HTTP/1.1 302 Moved\r\nLocation: /\r\n\r\n";
 static const char http_header_401[] = "HTTP/1.1 401 Bad request\r\n\r\n";
 static const char http_header_404[] = "HTTP/1.1 404 Not found\r\n\r\n";
@@ -98,40 +100,40 @@ static const char http_header_500[] = "HTTP/1.1 500 Internal Error\r\n\r\n";
     *{margin:0;padding:0}
     html{height:100%;background:linear-gradient(rgba(196,102,0,.2),rgba(155,89,182,.2)),url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEYAAAA8AgMAAACm+SSwAAAADFBMVEVBR1FFS1VHTlg8Q0zU/YXIAAADVElEQVQ4yy1TTYvTUBQ9GTKiYNoodsCF4MK6U4TZChOhiguFWHyBFzqlLl4hoeNvEBeCrlrhBVKq1EUKLTP+hvi1GyguXqBdiZCBzGqg20K8L3hDQnK55+OeJNguHx6UujYl3dL5ALn4JOIUluAqeAWciyGaSdvngOWzNT+G0UyGUOxVOAdqkjXDCbBiUyjZ5QzYEbGadYAi6kHxth+kthXNVNCDofwhGv1D4QGGiM9iAjbCHgr2iUUpDJbs+VPQ4xAr2fX7KXbkOJMdok965Ksb+6lrjdkem8AshIuHm9Nyu19uTunYlOXDTQqi8VgeH0kBXH2xq/ouiMZPzuMukymutrBmulUTovC6HqNFW2ZOiqlpSXZOTvSUeUPxChjxol8BLbRy4gJuhV7OR4LRVBs3WQ9VVAU7SXgK2HeUrOj7bC8YsUgr3lEV/TXB7hK90EBnxaeg1Ov15bY80M736ekCGesGAaGvG0Ct4WRkVQVHIgIM9xJgvSFfPay8Q6GNv7VpR7xUnkvhnMQCJDYkYOtNLihV70tCU1Sk+BQrpoP+HLHUrJkuta40C6LP5GvBv+Hqo10ATxxFrTPvNdPr7XwgQud6RvQN/sXjBGzqbU27wcj9cgsyvSTrpyXV8gKpXeNJU3aFl7MOdldzV4+HfO19jBa5f2IjWwx1OLHIvFHkqbBj20ro1g7nDfY1DpScvDRUNARgjMMVO0zoMjKxJ6uWCPP+YRAWbGoaN8kXYHmLjB9FXLGOazfFVCvOgqzfnicNPrHtPKlex2ye824gMza0cTZ2sS2Xm7Qst/UfFw8O6vVtmUKxZy9xFgzMys5cJ5fxZw4y37Ufk1Dsfb8MqOjYxE3ZMWxiDcO0PYUaD2ys+8OW1pbB7/e3sfZeGVCL0Q2aMjjPdm2sxADuejZxHJAd8dO9DSUdA0V8/NggRRanDkBrANn8yHlEQOn/MmwoQfQF7xgmKDnv520bS/pgylP67vf3y2V5sCwfoCEMkZClgOfJAFX9eXefR2RpnmRs4CDVPceaRfoFzCkJVJX27vWZnoqyvmtXU3+dW1EIXIu8Qg5Qta4Zlv7drUCoWe8/8MXzaEwux7ESE9h6qnHj3mIO0/D9RvzfxPmjWiQ1vbeSk4rrHwhAre35EEVaAAAAAElFTkSuQmCC)}
     body{font-family:arial,verdana}
-    div{position:absolute;margin:auto;top:0;right:0;bottom:0;left:0;width:320px;height:304px}
+    div{position:absolute;margin:auto;top:-150px;right:0;bottom:0;left:0;width:320px;height:304px}
     form{width:320px;text-align:center;position:relative}
     form fieldset{background:#fff;border:0 none;border-radius:5px;box-shadow:0 0 15px 1px rgba(0,0,0,.4);padding:20px 30px;box-sizing:border-box}
     form input{padding:15px;border:1px solid #ccc;border-radius:3px;margin-bottom:10px;width:100%;box-sizing:border-box;font-family:montserrat;color:#2C3E50;font-size:13px}
     form .action-button{border:0 none;border-radius:3px;cursor:pointer;}
     #msform .submit:focus,form .action-button:hover{box-shadow:0 0 0 2px #fff,0 0 0 3px #27AE60;}
-    #msform .wifitoggle:focus,form .wifitoggle:hover{box-shadow:0 0 0 2px #fff,0 0 0 3px #ccc;}
+    #formFrame{display: none;}
+    #aplist{display: block;}
     select{width:100%;margin-bottom: 20px;padding: 10px 5px; border:1px solid #ccc;display:none;}
     .fs-title{font-size:15px;text-transform:uppercase;color:#2C3E50;margin-bottom:10px}
     .fs-subtitle{font-weight:400;font-size:13px;color:#666;margin-bottom:20px}
     .fs-status{font-weight:400;font-size:13px;color:#666;margin-bottom:10px;padding-top:20px; border-top:1px solid #ccc}
     .submit{width:100px;background: #27AE60; color: #fff;font-weight:700;margin:10px 5px; padding: 10px 5px; }
-    .wifitoggle{float:right;clear:both;max-width:75%; font-size:11px;background:#ccc; color: #000; margin: 0 0 10px; 0; padding: 5px 10px;}
   </style>
 </head>
 <body>
   <div>
-    <form method="get" action="/update">
+    <form id="credentialsForm" method="get" action="/update" target="formFrame">
       <fieldset>
+        <iframe id="formFrame" src="" name="formFrame"></iframe> <!-- Used to submit data, needed to prevent re-direction after submission -->
         <h2 class="fs-title">WiFi Login</h2>
         <h3 class="fs-subtitle">Connect gadget to your WiFi network</h3>
-        <input autocorrect="off" autocapitalize="none" name="wifi_ssid" id="ssid" placeholder="WiFi Name">
-        <button type='button' class="action-button wifitoggle" id="apbutton" disabled>Searching for networks...</button>
-        <select name="aplist" id="aplist" size="2">
+        <input id="wifi_ssid" autocorrect="off" autocapitalize="none" name="wifi_ssid" placeholder="WiFi Name">
+        <select id="aplist" name="aplist" size="1" disabled>
+          <option>Scanning for networks...</option>
         </select>
         <input name="wifi_password" placeholder="Password" type="password">
-        <input type=submit name=save class='action-button submit' value='Save'>
+        <input type=submit name=save class="action-button submit" value="Save">
         <h3 class="fs-status">Status: <span id="status">Updating...</span></h3>
       </fieldset>
+      <h3 id="dbg"></h3>
     </form>
   </div>
   <script>
-    function aplist() { return document.getElementById("aplist"); }
-
     function fetch(url, method, callback)
     {
       var xhr = new XMLHttpRequest();
@@ -139,11 +141,14 @@ static const char http_header_500[] = "HTTP/1.1 500 Internal Error\r\n\r\n";
       function check_ready()
       {
         if (xhr.readyState === 4)
+        {
           callback(xhr.status === 200 ? xhr.responseText : null);
+        }
       }
       xhr.open(method, url, true);
       xhr.send();
     }
+
 
     function new_status(stat)
     {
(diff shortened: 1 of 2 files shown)


jmattsson:
This PR has merge conflicts, please fix and force-push the update. I'll *try* to have a look at this Soon(tm), but I'm a bit snowed under with work atm.

robertfoss:
Rebased on latest trunk on dev.

jmattsson:
I tried out this branch, but it keeps crashing hard in the ROM function `strcpy()` during use. Something is Not Good(tm).

Running 3490ffb285644ee086f7e62a1e5818c4c9e7aca2 + this commit on a NodeMCU 1.0 devboard, with the following as `init.lua`:
```
wifi.setmode(wifi.STATIONAP)
wifi.ap.config({ssid="testme",auth=wifi.AUTH_OPEN})
enduser_setup.manual(true)
enduser_setup.start()
```


From what I could see before it crashed, there was no longer any indication that the device is actively scanning for networks - a list just magically appears. If the aim is to get rid of the (hacky) button, I'd suggest still leaving the list visible, but disabled and with an entry to the effect of "Scanning..." or "No networks found." to still provide that info to the user.

robertfoss:
Will have a look at the crash.

The list containing always containing something is a good idea and totally the way to go.

robertfoss:
@jmattsson How did you diagnose it to be strcpy related? Which tools are you using?

TerryE:
> If pull requests didn't have to be a single commit, then debugging this would be a lot easier.

Robertr, see #1117.  Comments appreciated :-)

jmattsson:
The board crashed, and on reboot I inspected `node.bootreason()` and found the epc1 to be in ROM strcpy.

robertfoss:
I fixed a debug print and am now unable to reproduce the any crash. Could you have a look @jmattsson ?
I also added some default content to the aplist <select> tag.

jmattsson:
It's still crashing in `strcpy()` here for me. Dev version 09269a64526b42963ad64cb62e332b1d1fd9eb1d merged with this PR. Dies after having served the webpage, while executing instruction at 0x4000bf3c. Exception cause is `EXCCAUSE_LOAD_PROHIBITED` while accessing memory at 0xff0000 (which is clearly not a valid RAM/ROM address).

Drop-down box saying "Scanning for networks..." looks like a nice change (but I think it needs to be disabled by default - I happened to click on it and it copied the SSID "Scanning for networks.." into the text field :laughing: 

(Sorry about the late response - work has been *very* busy...)

gsker:
Don't you dare disable that! :-)

The javascript that scans for networks is a wonderful feature.
It's been in dev for quite a while.

I'm sure there there are some browsers that won't take as kindly to it, but
if the worst that happens is nonsense in the field once in a while, that's way
better than having to type the SSID in.  Especially on a tablet or phone.

No comment on the crash. I haven't seen that.

--
gsker@skerbitz.org

jmattsson:
No no no no, not *disable* disable, just make the drop-down un-selectable *while it is still scanning*. Being the one who added the scanning in the first place, I'm certainly not advocating its removal! :laughing:

pjsg:
I tried this, and I get a javascript error when I load the page. 

Uncaught SyntaxError: Unexpected identifier

This is running Chrome latest on OSX.

Safari says Unexpected identifier 'document'

Since the page is all on one line, it just fails on line 1!

gsker:
Well, that's a relief that you didn't intend disabling the whole thing.  :sweat_smile:

pjsg:
Also, can you fix the documentation so that there is a code sample that people might use. Also, document the default state of manual mode. I think that the SOFTAP comment should be in the start() documentation.

How are you supposed to use manual mode?

What is the standard way of discovering whether enduser_setup.start() should be called?

robertfoss:
@pjsg, I'll have a look at the docs.

Manual mode is used to have to module keep running and not automatically stop when a wifi has been connected to.

start() should always be called. Or nothing will happen.

robertfoss:
Addressed #1165.

robertfoss:
@pjsg Is the JS still failing?

app/modules/enduser_setup.c
```diff
@@ -580,7 +581,8 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
     return 1;
   }
 
-  struct station_config *cnf = os_zalloc(sizeof(struct station_config));
+
+  struct station_config *cnf = luaM_malloc(lua_getstate(), sizeof(struct station_config));
```
pjsg: Note that this doesn't clear out memory.... Also, you need to use luaM_free to free up this block as well....

robertfoss: How does c_free and luaM_free differ?

robertfoss: It's actually a separate allocator?

TerryE:
`luaM_alloc()` and `luaM_free()` use a common unified realloc function and in eLua this is `lauxlib.c:l_alloc()` which in the case of an alloc 0 bytes calls `c_free()`.  The main differences are that 

-  the lua alloc is integrated into the Lua GC and c_malloc is not
-  the lua alloc throws an error on allocation failure whereas c_malloc returns a NULL pointer.

app/modules/enduser_setup.c
pjsg: I don't know if it is a separate allocator, but it is best practice to use matched alloc/free pairs

pjsg:
With the current version, there are two issues:

* The page does not work due to the unexpected identifier. Curiously it does work if I prettyprint the javascript. This suggests that the minification is introducing an issue. It appears that there is a missing semicolon before the last 'document' -- though don't hold me to that.
* The firmware crashes. node.bootreason() returns 4,1

This *may* have been due to me connecting to the ssid, getting the page, then disconnecting, then reconnecting and trying to get the page again.

gsker:
Now that I finally learned out to get pull requests into my local git... nope.  :smile: 
   The javascript in the pull request does not work for me on my tablet.  
I tested on the laptop.  Not working there. The pulldown never becomes enabled.
Firefox says there's a syntax error ``missing ; before statement`` which could explain why prettyfying it might make it work -- The prettyfying probably added a newline.  
Sadly firefox isn't saying what line the error is on. Not much in the way of tools on my laptop right now.

robertfoss:
I found the JS error and committed a fix above. I must've accidentally altered something during minimization.

@TerryE, @pjsg about actually merging this PR. Would you like me to squash some of these changes before the PR is merged?

TerryE:
Robert we've bounced around a bit and the individual commits won't have a logic breakdown in retrospect, so my suggestion here would be to squash them, but let's wait until Philip get back with his  crash test and fix.  Thanks for you efforts :clap:

pjsg:
It still crashes for me. I pulled from robertfoss/enduser_setup_improvements, merged with dev and built.

I flashed the firmware onto a nodemcu. I ran the example:

```
wifi.setmode(wifi.STATIONAP)
wifi.ap.config({ssid="MyPersonalSSID",auth=wifi.AUTH_OPEN})
enduser_setup.manual(true)
enduser_setup.start(
  function()
    print("Connected to wifi as:" .. wifi.sta.getip())
  end,
  function(err, str)
  end
);
```

Almost immediately, it started to say:
```
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
```

I connected (from an OSX machine) to MyPersonalSSID and pointed chrome to 192.168.4.1.

I got a nice page, and the list of ssids filled out. I reconnected to another SSID and then back to MyPersonalSSID. I tried to fetch the 192.168.4.1 page again, and it crashed.

```
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234

 ets Jan  8 2013,rst cause:4, boot mode:(3,6)

wdt reset
load 0x40100000, len 28320, room 16 
tail 0
chksum 0xb5
load 0x3ffe8000, len 2196, room 8 
tail 12
chksum 0x15
ho 0 tail 12 room 4
load 0x3ffe8894, len 8, room 12 
tail 8
chksum 0xc5
csum 0xc5
rlŽ‚ÿ

NodeMCU 1.5.1 build unspecified powered by Lua 5.1.4 on SDK 1.5.1(e67da894)
lua: cannot open init.lua
> =node.bootreason()
4	1
> 
```

I just reproduced it with the above steps, but I got a different crash:

```
Connected to wifi as:192.168.1.234
Connected to wifi as:192.168.1.234

load 0x40100000, len 28320, room 16 
tail 0
chksum 0xb5
load 0x3ffe8000, len 2196, room 8 
tail 12
chksum 0x15
ho 0 tail 12 room 4
load 0x3ffe8894, len 8, room 12 
tail 8
chksum 0xc5
csum 0xc5
don't use rtc mem data
rlŽ‚ÿ

NodeMCU 1.5.1 build unspecified powered by Lua 5.1.4 on SDK 1.5.1(e67da894)
lua: cannot open init.lua
> =node.bootreason()
2	2	2	1073645268	0	0	1073645268	0
> 
```

This address is 3ffe86d4 which is actually in data space

```
 .data          0x000000003ffe8550       0x40 /opt/nodemcu-firmware/sdk/esp_iot_sdk_v1.5.1/lib/libpp.a(rate_control.o)
 .data          0x000000003ffe8590      0x24c /opt/nodemcu-firmware/sdk/esp_iot_sdk_v1.5.1/lib/libpp.a(trc.o)
 .data          0x000000003ffe87dc        0x8 /opt/nodemcu-firmware/sdk/esp_iot_sdk_v1.5.1/lib/libpp.a(wdev.o)
 .data          0x000000003ffe87e4        0x1 /opt/nodemcu-firmware/sdk/esp_iot_sdk_v1.5.1/lib/libnet80211.a(ieee80211_hostap.o)
```

I suspect some sort of stack overwrite or memory overflow. 

I have four SSIDs visible, and the longest one is 23 characters (and two are the same strings).

Hope this helps.

pjsg:
I suspect that there is another bug where the success callback gets called multiple times. I note that the creds for my local network are already stored so when the firmware boots, it immediately tries to connect (and succeeds).

app/modules/enduser_setup.c
```diff
@@ -246,72 +251,80 @@ typedef struct
...
 static bool manual = false;
 static task_handle_t do_station_cfg_handle;
+static task_handle_t do_disconnect_tcp_handle;
 
 static int enduser_setup_start(lua_State* L);
-static int enduser_setup_stop(lua_State* L);
+static void enduser_setup_stop();
+static void enduser_setup_stop_callback(void *ptr);
 static void enduser_setup_station_start(void);
 static void enduser_setup_station_start(void);
```
jfollas: Duplicate declaration

app/modules/enduser_setup.c
```diff
@@ -548,22 +577,28 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
   int pwd_str_len = enduser_setup_srch_str(pwd_str_start, "& ");
   if (name_str_len == -1 || pwd_str_len == -1 || name_str_len > 31 || pwd_str_len > 63)
```
jfollas: These lengths should probably be considered after URL decoding the values

app/modules/enduser_setup.c
```diff
@@ -1272,47 +1382,66 @@ static int enduser_setup_manual (lua_State *L)
 
 static int enduser_setup_start(lua_State *L)
 {
-  ENDUSER_SETUP_DEBUG(L, "enduser_setup_start");
+  ENDUSER_SETUP_DEBUG("enduser_setup_start");
 
-  if (!do_station_cfg_handle)
-    do_station_cfg_handle = task_get_id(do_station_cfg);
+  if (!do_station_cfg_handle || !do_station_cfg_handle)
```
jfollas: You probably mean for the second one to be `do_disconnect_tcp_handle`

marcelstoer:
> Would you like me to squash some of these changes before the PR is merged?

Yeah, once all issues are fixed I'd like to see the commits squashed. There's no need to show all 8+ commits in this upstream repository when commit n+1 is merely a fix for something introduced with commit n.

app/modules/enduser_setup.c
```diff
@@ -362,8 +376,15 @@ static void enduser_setup_check_station(void *p)
...
   }
 
-  enduser_setup_connected_callback(lua_getstate());
-  enduser_setup_stop(NULL);
+  enduser_setup_connected_callback();
+
+  /* Trigger shutdown, but allow time for HTTP client to fetch last status. */
+  if (!manual)
+  {
+    enduser_setup_check_station_stop();
```
jfollas: `enduser_setup_check_station_stop()` needs to be outside of the `if` block so that the callback isn't executed every second while in manual mode. ;-)

jmattsson:
I no longer hit a crash here! I guess the zero-init of that buffer fixed that issue?  Things are looking :+1: from me.

I'd really like to merge this in now since it has some rather important fixes. @robertfoss would you mind squashing down some of the fix-on-fix-on-fix commits a bit before I do so?

app/modules/enduser_setup.c
robertfoss: You are entirely correct. Committing a fix.

robertfoss:
@pjsg I'm not having any luck reproducing the crash, but the previous commit changes the onConnected callback to only be fired once.

Should the onConnected callback allowed to fire after every SSID submission?
The current implementation will only fire once, even if manual mode.

jfollas:
In my current fork of EUS, I have a variable in state that tracks when the callback is sent so that I can guarantee that it's only sent once. This gets reset each time that an /update? is called.

robertfoss:
@jfollas That sounds like the way to do it. I kind of want to get this PR out the door now though. It's getting to be pretty big.

robertfoss:
@jfollas Actually, do you mind creating a PR for that feature?

jfollas:
Yeah, as important/useful as EUS is, this won't be the last of the revisions. Get the current one merged, and we can start on other enhancements, I say.

robertfoss:
The PR has now been squashed, if no-one has any objections I would like to have it merged.

pjsg:
I'd like to see if the crash has gone away (when I get home tonight). I will also try the original code, and if the crash is there too, then I won't object.

robertfoss:
@pjsg, that sounds good to me. Thanks for having a look.

jfollas:
@pjsg Something that may help identify what's going on is to set`ENDUSER_SETUP_DEBUG_ENABLE=1`

app/modules/enduser_setup.c
```diff
@@ -1272,47 +1398,69 @@ static int enduser_setup_manual (lua_State *L)
...
+/**
+ * A wrapper needed for type-reasons strictness reasons.
+ */
+static void enduser_setup_stop_callback(void *ptr)
+{
+  enduser_setup_stop();
+}
+
+
+static void enduser_setup_stop()
```
jmattsson: This is now the wrong prototype for a function registered in a Lua table.

app/modules/enduser_setup.c
```diff
@@ -1272,47 +1398,69 @@ static int enduser_setup_manual (lua_State *L)
...
   {
-    enduser_setup_check_station_stop();
     enduser_setup_ap_stop();
   }
   enduser_setup_dns_stop();
   enduser_setup_http_stop();
   enduser_setup_free();
 
-  return 0;
+  return;
```
jmattsson: ...and this in particular is likely to cause grief, as Lua uses the return value (here, undefined) to retrieve N number of return values from the stack.

pjsg:
I did get one crash this evening. I also realized that the example that I was using had the setup set to manual mode which is presumably why I got lots of callbacks.

I have found that if you disconnect from the AP ssid and connect to another and then back to the AP ssid, it appears that you can no longer connect to 192.168.4.1 for http, but you can ping it. I suspect that this is when the crash can be triggered. 

Maybe the example should be fixed to not include manual mode -- then it works much better -- but the device remains in softap mode for a time after the success callback is called. Maybe the success callback should only be called once the device is in station mode. We don't have an easy way (except by using a timer) to wait for the switch into STATION mode. Or maybe the success callback should switch into STATION mode.....

For the example, I'm looking for the best practice for using this module in a random user's app.....

app/modules/enduser_setup.c
```diff
@@ -680,8 +775,8 @@ static void serve_status (struct espconn *conn)
  */
 static void enduser_setup_http_disconnect(struct espconn *espconn)
 {
-  ENDUSER_SETUP_DEBUG(lua_getstate(), "enduser_setup_http_disconnect");
-//TODO: Construct and maintain os task queue(?) to be able to issue system_os_task with espconn_disconnect.
+  ENDUSER_SETUP_DEBUG("enduser_setup_http_disconnect");
+  task_post_high(do_disconnect_tcp_handle, (task_param_t) espconn);
```
jmattsson: This is not a safe approach. The `espconn *` here is frequently a temporary lwIP *internal* struct (when it is passed in to an API callback), and is not valid after the callback returns. To do the disconnects safely, you'll need to stash away the `remote_ip` and `remote_port` like what is done in the `scan_listeners`. Apologies for not noticing this earlier.

app/modules/enduser_setup.c
```diff
@@ -1109,7 +1233,7 @@ static void enduser_setup_dns_recv_callback(void *arg, char *recv_data, unsigned
 
 static void enduser_setup_free(void)
```
pjsg: Can this be called while a timer is still running (say the 10 second finish timer)? If so, then there is a problem as the timer structure (which is part of the state structure) is still being referenced by the timer subsystem. Might be safer to explicitly cancel any outstanding timers.

jmattsson:
I'm also having a big issue whereby the SDK keeps accepting and handing me connections to service **after** the http connection has been deleted! I'm digging...

jmattsson:
Okay, the espconn API seems to be completely messed up in its design. Based on the testing here, when you first do an espconn_accept() you have a `remote_ip/port` of 0.0.0.0:0 as expected. However, once you've started accepting connections (keeping in mind you'll quite possibly have more than one connection in flight at any one time), the espconn API starts overwriting the `remote_ip/port` fields in the *listening espconn*.

Thus, by the time we're trying to actually stop the server from listening, instead of doing a delete on the "connection" `0.0.0.0:80 -> 0.0.0.0:0`, it gets done on a e.g. `0.0.0.0:80 -> 192.168.4.2:52123` and the *delete fails* with -5 "operation in progress". This then appears to not actually shut down the server, and because we go ahead and free up our state anyway, all hell breaks loose.

app/modules/enduser_setup.c
```diff
@@ -490,8 +520,15 @@ static void enduser_setup_http_urldecode(char *dst, const char *src, int src_len
       }
       *dst++ = c;
     }
+    if ((dst - dst_start) >= dst_len - 1)
```
jmattsson: Don't we need to have the test against `dst_len` as part of the outer loop? Right now it looks like it can overrun as long as it's not in an escape sequence.

jmattsson:
Right, so apparently, with the espconn abstraction, it is impossible to stop listening until any and all accepted connections have been closed.

It would appear that we *must* actively track any accepted connection (from the connect_cb and disconnect_cb) and when it's shutdown time first go and `espconn_abort()` any and all of those. After that we need to post a task to ourselves to attempt the `espconn_delete()`, because the aborts don't take effect immediately, they are internally posted onto the espconn task queue. In the meantime more connections may come in and be accepted, so if the `espconn_delete()` fails we're back to aborting all the accepted connections, and posting a task to ourself to retry the delete. In a pathological case, that would be a never-ending repeat. Naturally, in the meantime we can't free anything related to espconns, or we're off in lala-land again.

The espconn layer is thoroughly broken in its design as far as I can tell, and if it wasn't for the shared TLS abstraction I'd recommend staying well clear of this mess.

TerryE:
@jmattsson Johny, I am not interested in using this module, but I am interested in this discussion. Surely it has carry across to any server, and to the `net` implementation?

jmattsson:
Alright, we've confirmed that in order to shut down an espconn TCP server all of these hoops are in fact necessary lest one enjoys courting the deity of Undefined Behaviour aka Crashes. We have a functional, stable prototype of this whole process (there's more than one pitfall), and we'll continue working on that tomorrow. More stress testing will be done too.

Seeing as the enduser_setup module is unlikely to be the last module needing to support shutting down a TCP server we're making this a generic solution. Do it right and do it once. It will effectively replace any direct espconn interaction for TCP servers, hiding all the wtf-ery within its own task and exposing only well-defined callbacks. This should provide much saner behaviour and above all consistency. It will also be able to do away with the very awkward no-disconnects-from-a-callback restriction, which should make for significantly simpler module code. The cost will be mostly in code size, with a minor (currently ~40 bytes) memory overhead per active connection.

I'll note that at least for now I will not cater for the espconn_secure version of things, as the SDK docs explicitly state that there can only ever be one `espconn_secure_accept()`, and only a single connected client. As such it is a far simpler programming model already.

@robertfoss Don't worry about fixing any of the disconnect/shutdown oriented bugs just yet, it should become quite simple with the new abstraction layer. Do feel free to address the other bits - I'm still very keen to get this in! My hope is that we'll be ready to offer up both the TCP server abstraction and the integration with this PR in a branch late tomorrow.

jmattsson:
@TerryE does that post answer your question(s)?

TerryE:
:+1:

pjsg:
It is issues like this that make me wonder whether we should be using the esp-open-sdk where there is the possibility of fixing some of this underlying mess. The troublesome code is:

https://github.com/pfalcon/esp-open-lwip/blob/e2f8daeba1328ca0af7e73ab167c5810551959fe/lwip/app/espconn_tcp.c#L1337

Whether removing the check for open connections would actually cause a problem -- this I don't know.

app/modules/enduser_setup.c
robertfoss: Agreed, committed fix.

app/modules/enduser_setup.c
robertfoss: So 0 is wrong?

app/modules/enduser_setup.c
robertfoss: Alright, I'll remove the TCP disconnection code and just let the connections time out instead.

app/modules/enduser_setup.c
jmattsson: No, 0 is good. A void return is problematic, because with the prototype mismatch the hardware would pull a random(ish) value and use it as if it was the actual return value. The caller might see a return value of 5, 5000 or -5000 in this case, and if you're trusting that to be an item count on the Lua stack, well.... :)

app/modules/enduser_setup.c
robertfoss: I don't know if that is safe. It doesn't sound safe to me. Let's try to avoid that behaviour.

robertfoss: Is it safe to call os_timer_disarm on a disarmed timer?

app/modules/enduser_setup.c
robertfoss: I think you misread the code. It is actually running for escaped and non-escaped sequences.
The only case it's not running for is the last char of a sequence, which is fine, since we'll be exiting the function in that case anyway.

I could be misunderstanding something though.

jmattsson: My bad, you're right. I must've lost a brace in the github folding or something. Apologies for the noise!

app/modules/enduser_setup.c
jmattsson: Yes, it's a no-op.

robertfoss: I tried to have a look around for that info, but I couldn't find it anywhere. Where did you read about it?

robertfoss: Fix commited.

robertfoss:
Alright, so there's no more un-implemented feedback as far as I know.

pjsg:
I tried the latest code and I got one crash -- which I couldn't reproduce. However, sometimes when I start it in manual mode, I can't connect to 192.168.4.1 -- but I can ping it.

app/modules/enduser_setup.c
jmattsson: I got that from the disassembly of the ROM, I suspect :D

jmattsson:
I'm still working on the new TCP server layer, but I'll resume testing of this PR as soon as that's wrapped up!

jmattsson:
Gah... I just found a use-after-free in the espconn guts that I don't think I can avoid. I might have to shift the whole thing down to the real lwIP raw-API to make this stuff solid... argh.

jmattsson:
...and now I find a lovely spot where the the espconn layer assumes an os_zalloc() doesn't ever return null. Geez.

jmattsson:
Well the state of the espconn guts threw a spanner in the works for sure... After some more reading & pondering I concluded that the easiest option probably was to go for the actual LWIP native API and bypass the entire espconn mess for the TCP server.

I haven't finished testing, but what I have so far isn't completely broken at least. Please have a look at [this branch](https://github.com/DiUS/nodemcu-firmware/tree/enduser_setup_improvements) which is just another commit on top of this PR, and take it for a spin. It no longer crashes on `enduser_setup.stop()`, which is nice. I can't seem to restart the module after a stop however because the bind complains the port is in use. This needs some more investigating.

I'm also having to be a bit abrupt with the connection closing and do a hard close rather than a soft close, because otherwise it appears one gets many kilobytes of RAM used up for connections left in time-wait (from `/status` requests). Understanding what espconn does here would be useful, but I haven't had time to trawl through that codebase yet.

On the upside, when using the native LWIP API it was easy to avoid a layer of extra buffering for the html payload, and it also opens up the possibility of streaming the default payload straight from ROM at a later point, and probably also from the fs. Right now I'm just trying to get everything stable though.

robertfoss:
I've had a look at [DiUS EUS](https://github.com/DiUS/nodemcu-firmware/tree/enduser_setup_improvements), and it looks good.

robertfoss:
#1187, added documentation for alternative index.html file in filesystem.

skycoders:
As this is the general improvement issue, I have some more things to add, besides that this module is really great to make device setup much easier for potential end-users:
* The response in case of missing ssid or password is a HTTP 401, even though it should be 400, as this is the "bad request" response code. 401 is for "unauthorized". 
*  in case of a "not GET" request, no 401, but a 405 METHOD NOT ALLOWED should be returned
* I have no idea where the credentials are stored (couldn't find it in code), but somewhere they must be kept. It would be great if there was a call to reset them and force the module back into AP mode. There are reasons, one obvious is testing. If I want to try the setup process, I have to reflash the firmware right now. But also end-users might want to factory reset a device. 
* Extending the documentation with the calls you can do to send ssid and password as well as query the status would be great. Think of a mobile app used to initialize an IoT device. It might not want to use the captive website, but present some UI in the app and then only transmit the data. This is already possible, but not documented.

Thanks for the great work!

jfollas:
@skycoders: I have things for a subsequent PR that will address most of what you mention because I had similar needs. Robert's just trying to get this one wrapped up for now.

Wi-Fi credentials are stored in flash by the ESP directly when they are set. You can easily clear them from Lua using: `wifi.sta.config("", "")`.

skycoders:
@jfollas great, then I will simply wait for that one. Thanks for the update.

jmattsson:
@robertfoss good to the point where you want to get this across into this PR, or merely good as in a good starting point?

@pjsg have you had a chance to test it too? Personally I'd be fine with merging this PR + my espconn->lwip-native switchover as I believe the result would be stable (and naturally, an improvement over what we have)

skycoders:
@jmattsson I built a version from the current DiUS EUS branch and observed the following:
* The website did not update correctly. It remained in "updating" even though WiFi networks had been discovered. 
* Entering user and password never resulted in any status update, but the chip finally connected to the WiFi
* after restarting the chip, it connects to the WiFi in Station mode, nevertheless you can see the Enduser-Website when connecting to the IP address for some seconds. This might interfere with other services running on port 80 after connecting to the WiFi.

pjsg:
This last point is important. I would like the success callback to be fired *after* the platform is in STATION mode and *after* the http server has been torn down. This is the state where I can start my application.

I suspect that I just want to put the enduser_setup code into my init.lua and run it every time, and it leaves the platform in a nice state for running my app.

jmattsson:
Some questions/responses for you @skycoders:
1. Do you observe this with the base PR, or only on the DiUS EUS branch?
2. This is an unfortunate side-effect of the channel-changing of the AP that can take place when joining as a station. There are ideas for a future PR to mitigate this.
3. But that's "impossible"?!

pjsg:
@jmattsson Is there a single branch to test, or is it more complex?

jmattsson:
Just the single branch I linked above

robertfoss:
@jmattsson Really, I'd like to see no functional regressions. So if/when restarting EUS is possible, I'm happy. Is that the case now?

@pjsg For reasons of when coding this thing, the onConnected callback is fired before services are shutdown. The shutdown is initiated 10s after we've connected, so that we can continue replying to status requests of web clients and inform them of the new connection status.
I understand why you would like to have a clean slate to start from, but that 10s delay might be problem for other applications.
Adding a flag for toggling between the two modes is possible, but mode2 and the flag both require some code to be written. Really I'd like to close the PR before then. I know @jfollas has a subsequent PR coming up after this one, so that's another reasons for wanting to close this PR as  soon as possible.

jmattsson:
@robertfoss With the change away from espconn the crash on stop is gone as far as I can tell. There is another issue where it seems the listening port isn't made available for reuse for some time afterwards. I would prefer to address that in a separate PR if possible - this one has dragged on for too long and grown too much in the process.

pjsg:
I can't figure out which branch is the good one. This PR links to a branch that doesn't have any lwip changes. @jmattsson 's jmattsson:nodemcu-firmware repo doesn't have anything new in it. I feel as though I'm misinterpreting what you said.....

robertfoss:
@jmattsson Alright, that's good enough for me. I'm ready to have this thing merged then.

jmattsson:
@robertfoss do you want to cherry-pick over the espconn->lwip commit from the DiUS branch, and I'll merge after that?

skycoders:
@jmattsson 
1: I only tested it against the DiUS branch
2: yes, this is observed with current DEV branch of nodemcu as well, even though I got more status updates there
3: well, I can reproduce this. My code in init.lua starts enduser_setup and in the success callback prints the wifi.sta.getip() result. I used the EUS to connect to the wifi. From then on, after each reset of the chip, as soon as I see the local network ip address, I can open the EUS website for some seconds.

robertfoss:
@jmattsson Will do. It'll have to be in a few hours though.

jmattsson:
@skycoders ah, now I understand. You have eus on unconditional start in your init.lua, yes? That would explain it.

@robertfoss sure, I'm probably asleep by then, but I can merge tomorrow morning. Thanks!

skycoders:
@jmattsson correct. My expectation would have been that this is how it can be used. If one should keep track of a successful configuration by persisting some own state, this would be a good thing to be added to the documentation.
I also observed issues with starting an own listening server when the EUS was running. I just tool the example from the NET package returning "Hello, NodeMCU". That failed to respond, even when I used port 81. It works now that I have removed EUS from startup once it ran successfully.

jmattsson:
The way we're using EUS at work is to use `node.bootreason()` to differentiate between power-up and restart/dsleep-wake, and only run EUS on the power-ups. That's not to say the docs can't be improved though (PRs welcome!)

skycoders:
ok, I will create one. Starting with docs for a first pull request makes sense :)

robertfoss:
@jmattsson Cherry picked and squashed other small commits.

jmattsson:
Alright, marathon PR merged! Let's aim for several shorter ones in the future :laughing:

robertfoss:
Finally! Thanks for all the testing and suggestions.

jmattsson:
I think there are improvement suggestions in this thread enough to last the rest of the year!

sztanpet:
if someone provides a custom index.html with extra form fields, can the success callback receive the extra data somehow? My use case would be to specify a random string to use with the crypto module to provide message authentication codes

robertfoss:
Not as it stand now.

PR #1142, opened 2016-03-10
Fix null-pointer crash on early input

jmattsson:
With the recent changes to UART init and task handling, we've exposed a race condition that can cause a null-pointer crash whenever the ESP receives UART input in the window \[uart_config, lua_start). `lua_handle_input()` would happily be called before the Lua receive buffer has been allocated, with predictably bad results.

Tagging @TerryE for review/merge.


TerryE:
> Just close this PR when that's in

Done and will do.

Issue #1130, opened 2016-03-07
Why does ow.write with power=0 pull down the bus?


wuyuanyi135:
Hello. I just found some strange behavior of the onewire library.

I was using pin 2 to communicate with a DS18B20. In the documentation,

```
ow.write(pin, v, power)
power 1 for wire being held high for parasitically powered devices
```

I connected the power wire, so I set power as zero.

However, when issuing the commands

```
Reset -> select rom -> write 0xBE -> read -> read
```

the result was apparently incorrect (4095 degc).

I used my logic analyzer to see the timing, finding that after `write 0xBE` the bus was not held high. The following bits were therefore unreadable.

In the [source](https://github.com/nodemcu/nodemcu-firmware/blob/master/app/driver/onewire.c#L183), I spotted `*` line which probably pulled down the bus. I agree after write operation the bus should be at high-z but what is this line for? I think OW bus should always be held high when idle.

``` c
void onewire_write(uint8_t pin, uint8_t v, uint8_t power /* = 0 */) {
  uint8_t bitMask;

  for (bitMask = 0x01; bitMask; bitMask <<= 1) {
      onewire_write_bit(pin, (bitMask & v)?1:0);
  }
  if ( !power) {
    noInterrupts();
    DIRECT_MODE_INPUT(pin);
    DIRECT_WRITE_LOW(pin);    // <---------------------------*
    interrupts();
  }
}
```

desired behavior with power=1
![image](https://cloud.githubusercontent.com/assets/11760870/13576188/6b2bf674-e45a-11e5-86a4-a8e11b54179f.png)

when power=0, the bus was pulled down after write 0xbe, which breaks the next two reading commands
![image](https://cloud.githubusercontent.com/assets/11760870/13576301/0a1fd818-e45b-11e5-9962-f391d4ea042c.png)

wuyuanyi135:
Anyone ran into the same problem?

devyte:
It's possible nobody is using 1-wire at the moment, or that power=0 wasn't well tested, or something else. 
I don't know much about 1-wire, so I doubt I can be of much help, but a quick google puts forth some questions for you:
*Are you _sure_ its a DS18B20 and not a DS18B20-PAR?
[...]

PR #1143, opened 2016-03-10
onewire: fix unexpected low level after writing

wuyuanyi135:
first commit to fix issue #1130

If power = 0, after writing the bus is pulled down, and the slave will take it for a bit. One may have problem like the onewire device returning strange response, (for me, ds18b20 sometimes give temperature 4095.8125 degc). This commit will fix the unexpected situation. 

The power argument is not removed. We believe some use this module without external pull up as the documentation specified. The direct push-pull should be able to drive the slave. In that case, we may not switch the pin to input mode for the push arm supplying the power.


pjsg:
It is unclear to me what the power flag is supposed to do. I suspect that the documentation needs clarifying. I think that the original intention was that if power = 1, then you can connect the device directly to the nodemcu with no pullups or external power. If power = 0, then you must externally power it *and* have an external pullup resistor. There are two other cases:

* External power and no pullup resistor
* Parasitic power and a pullup resistor.

The change that you are proposing is to support "External power and no pullup resistor" (I think). 
I don't know if there is a difference between the pullup resistor going to 3v3 or going to 5v. 

I want the documentation to be updated in this PR to describe the supported hardware options and what the value of the power flag should be in each case.

wuyuanyi135:
@pjsg 

We shall figure out how to define this argument. I suggest power=0 when there is any pull-up on the bus, otherwise power=1. See the schema of different scenarios in #1130

wuyuanyi135:
@nickandrew I got it.
I am not really sure about that function. I am not familiar with the platform/native code of esp8266. I need to look into it later.

pjsg:
Adding the nice diagram into the docs and saying that power = 1 is no external pullup and power = 0 is external pullup --- that would be great.

pjsg:
@wuyuanyi135 Can you update the docs -- then are we done?

wuyuanyi135:
@pjsg Sorry I have been super busy these days. I will catch up as soon as I finish my stuff.

wuyuanyi135:
@pjsg Finally I start to deal with this. Can you tell me where can I edit the docs?

TerryE:
`docs/en/modules/OW.md`

wuyuanyi135:
@TerryE I can't find it. Is it in this repository?

pjsg:
Yes -- but only in the dev branch: https://github.com/nodemcu/nodemcu-firmware/blob/dev/docs/en/modules/ow.md

wuyuanyi135:
Just a few more questions. After commit the change of docs, what should I do to update the change to this PR? is it automatic?

pjsg:
The PR is updated automatically when you pushed -- so thank you.

I'm wondering about the second diagram with just the 4k7 pullup resistor. The [DS18B20 datasheet](https://datasheets.maximintegrated.com/en/ds/DS18B20.pdf) on page 7 says that you must use a strong pullup during temperature conversions and that the regular pullup is not strong enough.

Hence, I think that diagram two should have "power = 1".

app/driver/onewire.c
```diff
@@ -180,7 +180,6 @@ void onewire_write(uint8_t pin, uint8_t v, uint8_t power /* = 0 */) {
   if ( !power) {
```
TerryE: If you look at the code for DIRECT_MODE_INPUT in the ROM, then you will see that it already masks interrupts during this operation so the noInterrupts / interrupts are redundant.

wuyuanyi135: you are probably right. I did not look into the code but only removed the line that caused the bug.    Will you prefer I also fix this?

wuyuanyi135:
@pjsg Thanks. You are probably right. I will fix it soon.

app/driver/onewire.c
TerryE: Up to you.  I will be doing a pass later to remove all of these redundant interrupt locks some  time in the future.

nickandrew: I think the calls to noInterrupts()/interrupts() should be removed in this PR as it's a direct consequence of the change being done, i.e. keep all related changes together. And it's one less thing Terry will have to do :)

wuyuanyi135: Thanks, I will change it.

wuyuanyi135: Where can I read the sdk source like gpio_output_set? I am not sure whether the interruption is masked in some funcs

karrots:
The first and the last examples should probably be removed or note they are using internal pull-ups.

Though the internal pull-ups are likely, not strong enough to allow for good timing. I haven't seen any official docs on what the internal pull-up size is but it is likely around 33k ohm.

Below is an article showing the effects of different pull-up resistors on I2C communications.
http://dsscircuits.com/articles/effects-of-varying-i2c-pull-up-resistors

marcelstoer:
@wuyuanyi135 do you have time to incorporate Philip's comments in the next few days? I'd really like to get #1146 out of the way and would be happy to see this one fixed before.

karrots:
The `onewire_write_bit` function also needs to be fixed. It currently leaves the bus driven high which isn't good for an open-drain circuit. It should allow the pull-up resistor to bring the line back to up instead of driving it high. 

I have incorporated the changes in PaulStoffregen/OneWire#8 into a branch on my fork named `ow-opendrain-fix` it seems to work fine and leaves the bus in a more open drain friendly state at the end of operations.

pstolarz:
@karrots could you test OneWire lib's ARDUINO_ARCH_ESP8266 version of DIRECT_READ(), DIRECT_MODE_INPUT(), DIRECT_MODE_OUTPUT(), DIRECT_WRITE_LOW(), DIRECT_WRITE_HIGH() macros with nodemcu. They looks better for me (don't change gpio mode while writing like DIRECT_WRITE_XXX() do).
BTW. With your patch parasitic mode should work w/o any problems on nodemcu with Dallas sensors if the bus is powered after Convert_T command (0x44). See OneWire examples/DS18x20_Temperature.pde call to ds.write(0x44, 1);

wuyuanyi135:
@marcelstoer I am so sorry that I am busy passing the exams. I should have some free time after several days but I really hope someone can land this fix on time!

I understand the platform API has different behavior and it should be better to unify the platform implementation. Now the DIRECT_WRITE_LOW will bring pin to low in INPUT mode while on Arduino it will not. Maybe someone should look at this first - for code compatibility

karrots:
The macros currently in use `DIRECT_WRITE_LOW` and `DIRECT_WRITE_HIGH` change the pin to **output ** mode then change the state high or low. `DIRECT_MODE_OUTPUT` is currently a noop. 

@pstolarz is suggesting that the `DIRECT_MODE_OUTPUT` macro be changed to an actual operation then modify `DIRECT_WRITE_HIGH` and `DIRECT_WRITE_LOW` so they do not change the pin mode while changing the pin state. Sounds doable not sure it will accomplish anything past moving code around though. Still might be worth being closer to what the upstream code expects.

That said in looking at the current code all outputs pretty much follow the following steps (at least with @pstolarz power changes).

1. Change pin to output
1. Set level high or low
1. Change pin to input after specified period

Currently in the NodeMCU step 1 in that list is a noop due to the way the macros happen.

One additonal change that might be worth looking into is the abilityto turn off the internal pull-ups. I notice the current code has them on. The internal pull-up resistors likely don't provide enough current to properly pull-up the line reliably.

pstolarz:
@karrots Points 1 and 2 are actually opposite. As specified may cause gpio output blink.

This subtle issue with changing mode by DIRECT_WRITE_LOW() by nodemcu (which behavior was not expected by the upstream OneWire lib) causes problems raised by @wuyuanyi135. Rethink if this is a proper behavior of w1 DIRECT_XXX() macros. These macros are intended to be used internally by the w1 module only (there could be even moved to app/driver/onewire.c) and are not general purpose gpio access funcs (where such change-mode approach could be acceptable). If you changed these macros you should have re-inspected the w1 implementation too.

karrots:
Not really @pstolarz `DIRECT_OUTPUT_MODE` is a noop currently. Having the command above or below the `DIRECT_WRITE_*` makes no difference.

@wuyuanyi135's problem is that the bus was held low by a bug in the order of operations issue. This is fixed by your patch to the upstream code which is then in my branch. Unless I am missing something.

Also moving the defines from `onewire.h` to `onewire.c` won't really do much the macros are already only part of the onewire code they are not part of the core firmware like the GPE GPO and GPI macros are in the Arduino code.

pstolarz:
Talking about points 1 & 2 I meant the upstream lib where DIRECT_OUTPUT_MODE() actually changes gpio mode and DIRECT_WRITE_XXX() set flags in I/O reg controlling the output. In this case these points can't be switched. In nodemcu DIRECT_OUTPUT_MODE() should be even removed from the code to not mislead code readers.

@wuyuanyi135 problem root cause was setting the output by DIRECT_WRITE_LOW() after DIRECT_MODE_INPUT(). In nodemcu it ends up in gpio output shorted to the ground. In OneWire lib the gpio would be in input mode (which is expected behavior). Removing the lines proposed by @wuyuanyi135 fixes the issue but doesn't guarantee other problems caused by the macros different implementation. That's why I suggested nodemcu code inspection for this purpose.

My patch fixes completely other issue associated with open-drain character of the w1 bus which is a problem in nodemcu and OneWire as well. By the way it also fixes @wuyuanyi135 problem by reimplementing the part of code @wuyuanyi135 changed.

karrots:
In looking at the available macros in the SDK I can see why the original author chose to combine write with the state change to output. There isn't really a way to change the high/low setting on a pin via the SDK without also changing the state to output.

It seems like these issues could be avoided if the library (this and upstream) were rewritten to set  the GPIOs to open drain mode and only supply parasitic power via a 2nd pin to control a transistor.  That does complicate the connection of single sensors to uC's that can provide enough power via GPIO.

@wuyuanyi135 if you don't mind I'll submit a PR with the changes from @pstolarz. Before doing so I'll remove the references to `DIRECT_OUTPUT_MODE`. You can adjust your PR to just have the documentation updates or I can cherry-pick them over into my PR. I think there is still an image update that needs to be done though.

@TerryE you mention `DIRECT_OUTPUT_MODE` in the ROM are you referencing the SDK calls used by the `DIRECT_OUTPUT_MODE` macro in `onewire.h` or are these overridden somewhere?

If anyone desires I can post images showing the timing from my logic analyzer showing that the changes work and are decoded.

pstolarz:
On most platforms I'm aware of, digital outputs are characterized by its not-configurable type and there is not possible to change it (e.g switch between push-pull / open-drain / open-source type of output). Is this possible on the ESP8266 platform? Where can I find some tech. spec. of this platform (especially GPIO spec.)? I was unable to find anything interesting.

karrots:
@pstolarz  see nodemcu/nodemcu-firmware#1148 & http://bbs.espressif.com/viewtopic.php?t=2046 & http://bbs.espressif.com/viewtopic.php?f=67&t=225 for examples.

marcelstoer:
The minimal fix in #1291 was just merged. Those interested in continuing the discussion of the various related topics _this_ PR turned into pls create a separate issue -> closing.

Issue #966, opened 2016-01-23
How to handle backward compatibility in WS2812 module


Alkorin:
I open the issue to continue the discution on #951

The merge breaks backward compatibility. Syntax was
```
ws2812.writegrb(pin, string)
```

The new code ignores "pin" and uses only GPIO2.
- Do I remove the arg and update the doc ?
- Do I keed the arg and add an `if (pin != GPIO2) { print "Only GPIO2 is supported"; }`
- Do I reintroduce the old code to handle other pins but add a big warning in the documentation "If you use something else than GPIO2, happy tree friends are coming" ?

How do you handle such case in NodeMCU modules ?

skycoders:
another thing is the influence on the blue LED on ESP12E boards which is also controlled through GPIO02. It should be noted that it cannot be used any more.
My favorite would be to let people chose. There might be the case that no WiFi is needed, and that would leave them with more choices on how to use their PINs.

devsaurus:
I tend to vote for **backwards compatibility** and to revive the bitbanging code asap. But we'd serve two different masters again - the Lord of Stability and Count Lockula :grin:.
How about a global setting for the firmware: safe mode vs unsafe mode. Safe mode would be default after startup and the user is free to enter unsafe mode with let's say `node.unsafe_mode()`. Modules like `ws2812` would block all unsafe operations in safe mode (read: allow only GPIO2) while they're inviting happy tree friends for tea time in unsafe mode.

API doc of `node.unsafe_mode()` shows the SDK concepts of ISR blocking, callbacks etc. Links to FAQ and so on. While the specific module functions just tell that they enforce restrictions in safe mode. Avoids to scatter the same info all over the doc tree.

matgoebl:
I'd vote for `if(pin == gpio2) { uart } else { bitbanging }` plus warning in the docs.
There might be users out there with existing set-ups. This would be fully backward compatible.

I also noted, that the very first `ws2812.writergb()` after reboot does not work (initialization problem?).
So in order to clear all LEDs I need two lines:
`ws2812.writergb(rgb_pin,string.char(0)) -- this is because the first write does not clear led 1
ws2812.writergb(rgb_pin,string.char(0):rep(3*numleds))`

TerryE:
[...]

PR #1155, opened 2016-03-13
Allow interrupts when writing a ws2812 buffer

Christoph-D:
The new ws2812 module allows interrupts everywhere except when calling `buffer:write(led_pin)` on a ws2812.buffer object.  This restriction seems unnecessary, because `ws2812.write(led_pin, buffer)` accomplishes the same without disabling interrupts.

This pull request removes the call to `ets_intr_lock` from `ws2812_buffer_write`.

Also, initializing the output pin is not necessary in `ws2812_buffer_write` because `ws2812_write` does it anyway.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -269,14 +269,8 @@ static int ICACHE_FLASH_ATTR ws2812_buffer_write(lua_State* L) {
 
   luaL_argcheck(L, buffer && buffer->canary == CANARY_VALUE, 1, "ws2812.buffer expected");
 
-  // Initialize the output pin
-  platform_gpio_mode(pin, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
-  platform_gpio_write(pin, 0);
-
   // Send the buffer
-  ets_intr_lock();
   ws2812_write(pin_num[pin], &buffer->values[0], 3*buffer->size);
-  ets_intr_unlock();
 
   return 0;
 }


robertfoss:
@kbeckmann Isn't the IRQ lock needed for timing reasons?

Christoph-D:
As far as I followed the discussion, not needing the interrupt lock was the whole reason for rewriting the ws2812 module. The ws2812 LEDs are too slow, interrupts would be blocked longer than the SDK allows.

robertfoss:
@Christoph-D That would depend on the length of the ws2812 strip. The ws2812 protocol is not clock-driven so any variation in timing during data transmission would breaks things.

TerryE:
> The ws2812 protocol is not clock-driven so any variation in timing during data transmission would breaks things.

@robertfoss Robert, your advice is out of date as far as the dev build is concerned.  The latest version uses the UART FIFO H/W on chip, giving a 128 byte shift register, so interrupts don't need to be disabled.  as the H/W shift register means that we can tolerate multiple 50 µS ISR latencies.

robertfoss:
Well, in that case all is well, my complaint is hereby retracted.

TerryE:
@Christoph-D @Alkorin @kbeckmann, I note that the `docs/en/modules/ws2818.md` is out of sync with this version. (Write buffers are not documented).  Also the processing rules for the `write` and the `writergb` methods are different (one allows string parameters but the other doesn't.  

I am also confused here since I assume that a `ws2812.buffer` always stores its internal values in the device GRB order so how can it make sense passing a  `ws2812.buffer`  to `writergb` ?

Also the processing in  line 82 is an unguarded malloc, which is a no-no.  Better to use `luaM_malloc()`. 

I think that we should resolve these issues before merging this update.

marcelstoer:
> the `docs/en/modules/ws2818.md` is out of sync with this version.

@Alkorin promised to fix that a few weeks ago in https://github.com/nodemcu/nodemcu-firmware/issues/966#issuecomment-182054934 but #966 is still open.

Alkorin:
I've some free time today, i'll try to give a PR asas with documentation updated

Alkorin:
Pending work is here : https://github.com/nodemcu/nodemcu-firmware/compare/dev...Alkorin:ws2812-uart

I'll update documentation tomorrow

jmattsson:
Was the consensus that we should merge this PR now, or are we waiting on further updates (if so, which, and from whom)?

TerryE:
From @Alkorin Thömas.

Alkorin:
See #1263

marcelstoer:
#1263 replaces this.

Issue #1152, opened 2016-03-11
HTTP module - binary download issue


jfollas:
Download binary files using the HTTP client is failing for me when there's 0x00 in the data. Pretty sure it's because the buffer is a char\* in the HTTP client and C null termination is coming into play.

_Edited_

Researching this some more, I think to get around my first symptom (truncation of binary data), the Lua callback needs to use `lua_pushlstring` instead of `lua_pushstring` in the following:

``` c
  if (http_callback_registry != LUA_NOREF)
  {
    lua_rawgeti(http_client_L, LUA_REGISTRYINDEX, http_callback_registry);

    lua_pushnumber(http_client_L, http_status);
    if ( http_status != HTTP_STATUS_GENERIC_ERROR )
    {
      lua_pushstring(http_client_L, response);
    }
    else
    {
      lua_pushnil(http_client_L);
    }
    lua_call(http_client_L, 2, 0); // With 2 arguments and 0 result
```

https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/http.c#L32

The code that fills the buffer (in `http/httpclient.c http_receive_callback` ) appends a null character onto the end of the array, which may have an impact with binary data (i.e., the data won't have the same hash with the null character appended).

https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/http/httpclient.c#L251

jfollas:
I have a solution for this that I'll prepare a PR for.

TerryE:
`lua_pushstring()` should _never_ be used for user data, for this reason.  It should be limited to pusihing  known patterns and text which can't contain an embedded nullchar.  Having done a quick scan of `app/modules` this is the only case here.

jfollas:
Little hiccup in the process of getting here, but the PR 1157 includes the fix for this issue.

TerryE:
@jfollas  @vowstar, I've just had a look at this source and it has a lot of short comings, for example:
-  it requires you to build with SSL enabled even if you are only using HTTP in your app.
-  it reimplements standard functions for no clear reason
-  it's memory allocation / freeing is flaky so it will generate memory leaks
-  it doesn't handle records which span IP packets.

Nothing to do with this PR but a lot more work is really needed on this.

jfollas:
[...]

PR #1156, opened 2016-03-14
Fixed http module binary download issue, refactored, added HEAD support

jfollas:
- Issue #1152 (http module did not properly handle binary payload)
- Consolidated duplicate code in http module to make smaller
- Added support for HTTP HEAD method
- Added HTTP response headers to the callback function
- Changed User-Agent header to only use default if not provided by Lua

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -9,7 +9,6 @@
 
 /*
  * FIXME: sprintf->snprintf everywhere.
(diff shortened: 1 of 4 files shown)


jfollas:
Closed because GitHub couldn't automerge this PR. Recreated my changes in a new branch/PR instead of trying to fix this one.

Issue #1146, opened 2016-03-10
Propagate `dev` to `master`


marcelstoer:
We should IMO plan ahead and come up with a preliminary list of issue still to be fixed before current `dev` can be propagated to `master`.

Here's my take. Fellow committers feel free to edit this list directly here. The big question for me is if we should wait for SDK 1.5.2 support or not.

- [x] PR #1260, Added length checking to prevent stack overflow 
- [x] PR #1203, Make tmr.time() more resilient against RTC changing frequency
- [x] PR #1202, Fixed double-free in crypto module
- [x] PR #1200, enduser_setup: Fixed missing status update.
- [x] ~~PR #1116, Fix crash bug when HTTP response from server is NULL~~ superseded by #1212, Minimal changes to handle bad bodies in HTTP.
- [x] PR #1123, MQTT Client - CONNACK processing
- [x] PR #1129, Try to make qos~=0 work better
- [x] PR #1136, one line bugfix to GPIO.c -- Oops!
- [x] PR #1138, Improve overall enduser_setup user experience. 
- [x] PR #1142, Fix null-pointer crash on early input
- [ ] ~~PR #1143, fix unexpected low level after writing~~ partially replaced by PR #1291
- [x] bug #1065, wifi.sta.eventMonReg previous_State is always "nil"
- [ ] ~~bug #1039, rtctime.dsleep doesn't set correct bootreason~~
- [x] ~~#1009, Thoroughly document how NodeMCU handles esp_init_data_defaults~~ replaced by PR #1285, Reworked ADC mode setting approach
- [x] #966, How to handle backward compatibility in WS2812 module

TerryE:
Marcel, thanks for this analysis. I found it very useful.  Whilst I broadly agree with the list, ultimately I also feel that we should apply some sort of 90/10 or 80/20 rule here, otherwise will never get to then state where `master` is stable enough.

marcelstoer:
Terry, I don't think you'll find a fellow committer who disagrees on applying Pareto principle here :wink: However, I'm a bit puzzled...What changes exactly do you propose? Which issues do you think are missing, which ones would you remove?

TerryE:
[...]

PR #1157, opened 2016-03-14
Fixed http module binary download issue, refactored, added HEAD support

jfollas:
- Issue #1152 (http module did not properly handle binary payload)
- Consolidated duplicate code in http module to make smaller
- Removed the functions-per-method in `httpclient.c` since http module just needs the "request" method 
- Added support for HTTP HEAD method
- Added HTTP response headers to the callback function as third argument
- Changed `User-Agent` header to only use default of "ESP8266" if user didn't provide in headers argument

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -19,59 +19,59 @@
 #include "httpclient.h"
 
 /* Internal state. */
-typedef struct request_args_t {
-	char		* hostname;
-	int		port;
-	bool		secure;
-	char		* method;
-	char		* path;
-	char		* headers;
-	char		* post_data;
-	char		* buffer;
-	int		buffer_size;
-	int		timeout;
-	os_timer_t	timeout_timer;
-	http_callback_t callback_handle;
+typedef struct request_args_t { 
+  char    * hostname;
+  int     port;
+  bool    secure;
+  char    * method;
+  char    * path;
+  char    * headers;
+  char    * post_data;
+  char    * buffer;
+  int     buffer_size;
+  int     timeout;
+  os_timer_t  timeout_timer;
+  http_callback_t callback_handle;
 } request_args_t;
 
 static char * ICACHE_FLASH_ATTR esp_strdup( const char * str )
 {
-	if ( str == NULL )
-	{
-		return(NULL);
-	}
-	char * new_str = (char *) os_malloc( os_strlen( str ) + 1 ); /* 1 for null character */
-	if ( new_str == NULL )
-	{
-		HTTPCLIENT_DEBUG( "esp_strdup: malloc error" );
-		return(NULL);
-	}
-	os_strcpy( new_str, str );
-	return(new_str);
+  if ( str == NULL )
+  {
+    return(NULL);
+  }
+  char * new_str = (char *) os_malloc( os_strlen( str ) + 1 ); /* 1 for null character */
+  if ( new_str == NULL )
+  {
+    HTTPCLIENT_DEBUG( "esp_strdup: malloc error" );
+    return(NULL);
+  }
+  os_strcpy( new_str, str );
+  return(new_str);
 }
 
 
 static int ICACHE_FLASH_ATTR esp_isupper( char c )
 {
-	return(c >= 'A' && c <= 'Z');
+  return(c >= 'A' && c <= 'Z');
 }
 
 
 static int ICACHE_FLASH_ATTR esp_isalpha( char c )
 {
-	return( (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') );
+  return( (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') );
 }
 
 
 static int ICACHE_FLASH_ATTR esp_isspace( char c )
 {
-	return(c == ' ' || c == '\t' || c == '\n' || c == '\12');
+  return(c == ' ' || c == '\t' || c == '\n' || c == '\12');
 }
 
 
 static int ICACHE_FLASH_ATTR esp_isdigit( char c )
 {
-	return(c >= '0' && c <= '9');
+  return(c >= '0' && c <= '9');
 }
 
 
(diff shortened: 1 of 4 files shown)


app/http/httpclient.c
```diff
@@ -292,33 +284,48 @@ static void ICACHE_FLASH_ATTR http_connect_callback( void * arg )
...
 	}
 
+	char ua_header[32] = "";
+	
+	if ( os_strstr( req->headers, "User-Agent:" ) == NULL )
+	{
+		os_sprintf( ua_header, "User-Agent:%s\r\n", HTTP_DEFAULT_USER_AGENT );
+	}
+	
 	char buf[69 + strlen( req->method ) + strlen( req->path ) + strlen( req->hostname ) +
```
pjsg: Should add in strlen(ua_header)

app/modules/http.c
```diff
@@ -7,224 +7,162 @@
...
 {
+
 #if defined(HTTPCLIENT_DEBUG_ON)
-  c_printf( "http_status=%d\n", http_status );
-  if ( http_status != HTTP_STATUS_GENERIC_ERROR )
-  {
-    c_printf( "strlen(full_response)=%d\n", strlen( full_response ) );
-    c_printf( "response=%s<EOF>\n", response );
-  }
+	HTTPCLIENT_DEBUG( "http_status=%d\n", http_status );
```
pjsg: I think that the nodemcu coding style is 2 space indents.....

app/http/httpclient.c
jfollas: Ooh, good catch

app/modules/http.c
jfollas: Looking through the code, there doesn't seem to be a consistent style (and none documented that I can find). I styled using leading tabs here because then at least the tab stops can be set within each person's editor.

TerryE:
> Looking through the code, there doesn't seem to be a consistent style.

Indent is 2 spaces as per the Lua style guide.  Using tabs is a pain because of variations in tab settings and display in Github.  But this module is just badly laid out in my view.  Nothing to do with you: history.

marcelstoer:
> Indent is 2 spaces as per the Lua style guide.

Terry, did you really mean to write 'Lua' here? If so, how does that relate to C source?

> But this module is just badly laid out in my view.

May very well be but this shouldn't hold this PR back. We can't save the world today but only strive to make it a better place one PR at a time :wink:

devsaurus:
> Indent is 2 spaces as per the Lua style guide.

Guess we're missing a related hint in `CONTRIBUTING.md` for C code. The code base is also quite inconsistent when it comes to [indent styles](https://en.wikipedia.org/wiki/Indent_style). K&R and Allman seem to be dominant, but they're even mixed within single files.

marcelstoer:
> Guess we're missing a related hint in `CONTRIBUTING.md` for C code.

Yep, I agree but that should IMO be discussed outside this PR. I created #1168 for that.

app/http/httpclient.c
```diff
@@ -292,33 +284,50 @@ static void ICACHE_FLASH_ATTR http_connect_callback( void * arg )
...
+	if (req->headers == NULL) /* Avoid NULL pointer, it may cause exception */
 	{
 		req->headers = (char *)os_malloc(sizeof(char));
 		req->headers[0] = '\0';
 	}
 
+	char ua_header[32] = "";
+	int ua_len = 0;
+	
+	if ( os_strstr( req->headers, "User-Agent:" ) == NULL )
```
jmattsson: Does `os_strstr()` cope with `"user-agent:"` too?

jfollas: Oh, snap! Headers are case-insensitive, aren't they.... What do you recommend for an insensitive substring search given the library that we have?

marcelstoer: There's no function to convert `req->headers` to lower case at once i.e. you'd have to go through it char-by-char? Wouldn't it be enough to check for  "User-Agent:" and "user-agent:" only? I've never encountered all-caps headers in my life.

TerryE: Isn't `stricmp()` available?

TerryE:
@jfollas Jason, I know that this isn't down to you and you are only trying to tidy up some specific issues with this module, but just quickly reviewing this code, it is horrible.  I am not sure why we ever accepted it into our core repo.  My inclination is to deprecate its use until we bring it up to some minimal acceptable standard and if you are willing to do this then this would be great.

-  None of the memory alloc functions or use (including the use of the reimplemented  strdup) properly handle a null-return out of memory so will ultimately die messily.
- I was interested why the code includes a local implementation of `strtol` when there is already a `strtoul`, so I looked at where it was used: in `http_chunked_decode()` but surely HTTP 1.1 chunked encoding uses a double character hex encoding not a signed long? ([See RFC 2616 Sec 3](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html) and in particular 3.6.1).  _Has this ever been tested?_
- `http_request()` can failed silently (debug printing doesn't count).  If it's going to fail without throwing an error then it should return a status.
-  The URI parser is a little bizarre.  OK it makes no attempt to parse a URI not of the form `protocol://host:port/path`, but it will incorrectly URI strings like `http://fred.com:80` in an odd way.

I could go on, but I've got more pressing jobs on my TODO list, but ughhh!

jfollas:
@TerryE  The upstream source for where @vowstar got the client code is here: https://github.com/Caerbannog/esphttpclient

Since the last commit was in October, I'm betting it was written against early SDKs, which may explain why the C functions were recreated in this source.

I can spend some time on a revamp. I haven't been very deep in C since university decades ago, so my memory management skills are a bit rusty, and I'm never sure which prefixed C function I'm supposed to use when I need one, and details like when to use `ICACHE_FLASH_ATTR` are a bit foggy... but that's what you guys are for with code reviews, right? :-)

app/http/httpclient.c
jfollas: I didn't see stricmp in any of the header files that I looked in. In this case, since it's data provided by the Lua developer (and not some arbitrary internet server), I think it's okay to just check the two cases: `User-Agent` and `user-agent`.

TerryE:
Yup, IIRC vowstar committed this himself without proper review before we tightened up on PR rules.  I understand the aim of having a simple single call API for fetching URIs, but IMO, the original developer isn't an experienced programmer.  It looks more like a university / beginner project to me.  The problem with putting this sort of code into our repo is that people will use it and them we have to sort out all of the support issues.  
-   On a system like the ESP, running out of RAM is a real possibility.  If this isn't caught then the program will ultimate die on a memory exception.  IMO, it'sfar better to use the Lua memory system as this will attempt a GC if necessary and throw an error on allocation failure.  
-   With the current Lua builds you don't need the ` ICACHE_FLASH_ATTR` qualifiers, and you can just use `strlen`, etc.

marcelstoer:
> The upstream source for where @vowstar got the client code is here: https://github.com/Caerbannog/esphttpclient

There are more fixes(?) pending at that repository. I'll reach out to the developer as it would IMO make sense to concentrate efforts here.

> IIRC vowstar committed this himself without proper review before we tightened up on PR rules.

It was way after (end of December 2015) but technically PR rules don't apply for committers :frowning:.

jmattsson:
Actually I believe I was the one who merged it, with the reasoning that I'd rather have a mostly functional http client than none at all.

marcelstoer:
@jfollas do you have time to incorporate Terry's comments and rebase your branch on current `dev` in the next few days? I'd really like to get #1146 out of the way and would be extremely happy to see this one fixed before.

jfollas:
@marcelstoer I've been putting in long hours trying to meet a deadline at work. I'm not sure that I'll get to it this week, but I may be able to spend time on it tomorrow during a long flight.

app/http/httpclient.c
```diff
@@ -86,430 +81,460 @@ esp_isdigit( char c )
...
+  if (req->headers == NULL) /* Avoid NULL pointer, it may cause exception */
+  {
+    req->headers = (char *)os_malloc(sizeof(char));
+    req->headers[0] = '\0';
+  }
+
+  char ua_header[32] = "";
+  int ua_len = 0;
+  
+  if ( os_strstr( req->headers, "User-Agent:" ) == NULL || os_strstr( req->headers, "user-agent:" ) == NULL)
```
luismfonseca: Shouldn't this be an AND instead of OR?

If `User-Agent:` AND `user-agent:` aren't present, then use default user agent.

marcelstoer:
There's so much goodness in this PR (properly handle binary payload, HEAD support, additional callback argument) but I'll still close it in the next days if no one objects:
- white space changes all over (should be handled with #1168 instead)
- ideas of the first two commits were since integrated through other PRs and hence the merge conflicts
- pending changes based on review comments from Terry

TerryE:
Nothing from me to add.  As you say, some of the good bits have been cherry-picked elsewhere.

marcelstoer:
Jason, if you have some time we'd be delighted to merge a  cleaned-up version.

Issue #1161, opened 2016-03-15
SPI - CPHA is inverted when CPOL=1


jfollas:
NodeMCU received CPOL=1 support a few months ago (before then, we were stuck with only CPOL=0, or low-idle clock). Today I scoped out one of our SPI interfaces that was misbehaving while using CPOL=1,CPHA=1, and it was showing the wrong waveform. I have not checked CPOL=0 yet.

The CPHA=1 waveform looked like what CPHA=0 should, and vice versa (it was one of those "How did this ever work in the first place?" situations).

CPOL=1, CHPA=1
![CPOL=1](http://i.imgur.com/LrNB9M3.png)

CPOL=1, CPHA=0
![CPOL=0](http://i.imgur.com/SZEmwsL.png)

Example of what it's supposed to look like:

![SPI Waveforms](https://abieneman.files.wordpress.com/2010/04/atmega-spi-modes.gif)

I'm going to make the correction that I think, and then scope out the new waveforms. If they are correct, then I'll submit the PR.

TerryE:
@jfollas much appreciated.  I await your analysis with interest.  Regards Terry

jfollas:
Submitting a PR on this. I've verified the waveforms for all four modes.

*Mode 3*
![Mode 3](http://i.imgur.com/1O4gAt2.png)

*Mode 2*
![Mode 2](http://i.imgur.com/Mvdflpy.png)

*Mode 1*
![Mode 1](http://i.imgur.com/TBQ6JYf.png)

*Mode 0*
![Mode 0](http://i.imgur.com/HXQuXPj.png)

jfollas:
#1162 has been merged.

PR #1162, opened 2016-03-15
Corrected SPI CPOL/CPHA relationships

jfollas:
Fix for issue #1161

### DIFF
--- a/app/driver/spi.c
+++ b/app/driver/spi.c
@@ -72,18 +72,34 @@ void spi_master_init(uint8 spi_no, unsigned cpol, unsigned cpha, uint32_t clock_
 
 	SET_PERI_REG_MASK(SPI_USER(spi_no), SPI_CS_SETUP|SPI_CS_HOLD|SPI_RD_BYTE_ORDER|SPI_WR_BYTE_ORDER|SPI_DOUTDIN);
 
-	//set clock polarity (Reference: http://bbs.espressif.com/viewtopic.php?f=49&t=1570)
+	// set clock polarity (Reference: http://bbs.espressif.com/viewtopic.php?f=49&t=1570)
+	// phase is dependent on polarity. See Issue #1161
 	if (cpol == 1) {
 		SET_PERI_REG_MASK(SPI_PIN(spi_no), SPI_IDLE_EDGE);
+
+		//set clock phase
+		if (cpha == 1) {
+			// Mode 3: MOSI is set on falling edge of clock, MISO is read on rising edge of clock
+			CLEAR_PERI_REG_MASK(SPI_USER(spi_no), SPI_CK_OUT_EDGE);
+			SET_PERI_REG_MASK(SPI_USER(spi_no), SPI_CK_I_EDGE);
+		} else {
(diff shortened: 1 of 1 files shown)


TerryE:
This looks sensible, but I'd rather have another committer who has an SPI test rig, such as Arnim (@devsaurus) take a look and OK this.

jfollas:
I saw the opportunity to do a tiny refactoring to remove duplicate code in my implementation, so added that to the PR.

devsaurus:
I can reproduce the fixed MOSI generation and it's in line with the expected [polarity and phase](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus#Clock_polarity_and_phase).

Next I need to figure out how to verify that the correct SPICLK edge is used for sampling MISO in all four modes. It'll take a couple of days until I've set up a testbench for that.

devsaurus:
Sampling of MISO works also well. Although a bit differently from how I expected it to be.

I built a test slave that outputs to MISO with surrounded-by-low instead of NRZ to identify the sampling edge. This revealed that the ESP samples on the very same edge on which it outputs MOSI data. 

This is a transaction with CPOL=0 and CPHA=0. The ESP requires that MISO bits are available at falling clock edge (slave transmits 0xaa):

spi_sbl

Summary: MOSI and MISO timing is controlled by `SPI_CK_OUT_EDGE` only. There's no influence on MISO sampling from `SPI_CK_I_EDGE`. It's probably used for SPI slave mode.

jfollas:
I'm surprised about the `SPI_CK_I_EDGE` finding. But, given the documentation (or lack of), I'm not too surprised. 

So we can take `SPI_CK_I_EDGE` out of the code altogether without changing the behavior?

devsaurus:
> So we can take `SPI_CK_I_EDGE` out of the code altogether without changing the behavior?

No change in behavior regarding MISO sampling as far as I can tell, yes.
But it'd be a certain risk since we don't know which other side effects might be triggered. Without design docs I can't derive a dedicated test case to check for regressions when `SPI_CK_I_EDGE` is removed; except for running "everything" and hoping to trigger & observe deviations.

How about keeping `SPI_CK_I_EDGE` as it is now but just correct the comments (they say that MISO is read on the other edge atm) and update the documentation accordingly?

jmattsson:
Any updates on this?

jfollas:
@devsaurus's test shows that the ESP isn't behaving per spec - not sure if that's by design, or if we have a wrong bit setting (or need to use yet-another undocumented bit). I tried asking Espressif for clarification, but you know how that goes. Crickets.

This PR fixes my specific issue when CPOL=1, and there doesn't seem to be any problems for my slave devices with the ESP sampling MISO data half a clock cycle too soon. 

I'm personally okay with the PR as is, and if we get a better understanding for how to shift the input to the correct clock edge, there can always be another PR later.

devsaurus:
I'm also fine this conclusion.

@jfollas please remove the comments that make assumptions about MISO sampling like "`MISO is read on falling edge of clock`". Then we're ready to go.

Issue #846, opened 2015-12-13
Updating modules that are incompatable with Lua coroutining


TerryE:
I've just been making some blanket changes to the modules and noticed that CJSON assumes a single global state in 

```C
static lua_State *gL;
// ... 
void cjson_mem_setlua (lua_State *L)
{
  gL = L;
}

void *cjson_mem_malloc (uint32_t sz)
{
  void *p = (void*)c_malloc (sz);
  if (!p && gL)
// ...
```
Separate coroutine threads have separate state vectors L.  This code assumes that there is a single global state vector.  Not good.

nickandrew:
@TerryE Any idea why it does it? To save memory?

TerryE:
No because it is a plug in replacement for c_malloc and malloc doesn't accept the state.  PHP has the same concept as L for threaded code (called TSRM) , but it provides a macro which is called `TSRMLS_FETCH()` to recover the current thread vector if you are in a routine which doesn't have it  passed down.  This does some magic with the stack to work out which thread is running.  We could do something similar, but the easiest way is to pass L into the allocator by defining `CSON_MALLOC(s)` as `cjson_mem_malloc (sz, L)

TerryE:
As I discuss in #845,  `gpio`,  `net`, `node`,`cjson` and `wifi` all use this trick.  It can fail if the application uses coroutines because each thread has its own `Lua_state`, and applications like the http server example uses coroutines.  

**eLua** has the lua function `lua_getstate(()` which always returns the main thread, so we could use this instead, but the issues arise if one thread executes the API call which sets `gL` but another thread executes the call which accesses it, because the wrong thread context will be applied, so for example upvalues will be garbage.  Likewise if the setting routine exits, its thread is freed so gL will pointing at garbage. 

This one could be a minefield so I need to think this through.

TerryE:
OK, one aspect that I did get wrong when I've checked: both globals **_G** and the Lua registry are shared across threads:

```
> =debug.getregistry()
table: 0x3fff0a10
> x=coroutine.create(function()print(debug.getregistry())end)
> =x
thread: 0x3fff1b68
> coroutine.resume(x)
table: 0x3fff0a10
> 
```
(Both standard Lua and nodeMCU Lua on the ESP show this behaviour.)  

Thinking about this, I think that we are OK always to use the main thread for this type of context so we should just convert all of these to `lua_getstate()` .  Any alternative opinions?

pjsg:
[...]

PR #1167, opened 2016-03-17
Update modules to be Lua coroutine compatible

TerryE:
Changes as per #846.  I still have to do fill testing.  I can do the net, uart and wifi ones.  COAP is pretty moribund anyway at the moment.   This is more to provoke feedback comment than to merge just yet.

### DIFF
--- a/app/modules/net.c
+++ b/app/modules/net.c
@@ -34,7 +34,6 @@ static int expose_array(lua_State* L, char *array, unsigned short len);
 #define MAX_SOCKET 5
 static int socket_num = 0;
 static int socket[MAX_SOCKET];
-static lua_State *gL = NULL;
 static int tcpserver_cb_connect_ref = LUA_NOREF;  // for tcp server connected callback
 static uint16_t tcp_server_timeover = 30;
 
@@ -65,8 +64,8 @@ static void net_server_disconnected(void *arg)    // for tcp server only
   lnet_userdata *nud = (lnet_userdata *)pesp_conn->reverse;
   if(nud == NULL)
     return;
-  if(gL == NULL)
-    return;
+  lua_State *L = lua_getstate();
+
 #if 0
   char temp[20] = {0};
   c_sprintf(temp, IPSTR, IP2STR( &(pesp_conn->proto.tcp->remote_ip) ) );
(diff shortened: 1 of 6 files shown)


app/modules/net.c
```diff
@@ -121,12 +121,12 @@ static void net_socket_disconnected(void *arg)    // tcp only
   if(nud->pesp_conn)
     c_free(nud->pesp_conn);
   nud->pesp_conn = NULL;  // espconn is already disconnected
-  lua_gc(gL, LUA_GCSTOP, 0);
+  lua_gc(L, LUA_GCSTOP, 0);
```
pjsg: As a matter of interest, why is the GC disabled here for a short time? I don't think that it needs to be..... [Worse, if it does need to be disabled, then the disable call is probably missing from lots of other places as well]

app/modules/net.c
```diff
@@ -295,25 +278,25 @@ static void net_server_connected(void *arg) // for tcp only
 
   if(tcpserver_cb_connect_ref == LUA_NOREF)
     return;
-  if(!gL)
+  if(!L)
```
pjsg: This can go.....

app/modules/uart.c
```diff
@@ -8,19 +8,19 @@
...
 static int uart_receive_rf = LUA_NOREF;
 bool run_input = true;
 bool uart_on_data_cb(const char *buf, size_t len){
+  lua_State *L = lua_getstate();
   if(!buf || len==0)
     return false;
   if(uart_receive_rf == LUA_NOREF)
     return false;
-  if(!gL)
+  if(!L)
```
pjsg: THis can go (can lua_getstate() ever return null?)

TerryE: Yup, I'd spotted this one myself.  The tl;dr is no.   The lua_State used by `lua_getstate()` is set up pretty much 1st thing in `lua_main()` so the only time it can be zero is if `lua_main()` hasn't been called.  Any library API call can only be executed after `lua_main()` has been run, so the only potential issue is with a driver ISR triggered task if the ISR is set up by the `platform_init()`.  The one case here is the UART driver and we've already fixed this race condition (#1142).

app/modules/net.c
TerryE: Don't know.  I could take a look but I think that I'll leave this until we fix the `net` module properly.

TerryE:
These seem to be stable.  Anyone found any probs?

dnc40085:
@TerryE I'm pretty sure that the conflicts are caused by #1018, which includes the lua coroutine compatibility fix for `wifi.c`.

TerryE:
@dnc40085 @jmattsson @devsaurus @pjsg, you are the guys who have worked on these stacks in the past.  This PR has been sitting here for 3 weeks.  I am assuming that no one has any issues other than the merge conflicts created by later PRs that have been accepted.  If I don't hear to the contrary in the next day or so, then I'll remove the conflicts and merge this myself.

TerryE:
> As a matter of interest, why is the GC disabled here _[in net_socket_disconnect]_for a short time? I don't think that it needs to be.

@pjsg Philip, it's a cludge, and it's used in a couple of other places in the module.  The `net.socket` userdata include a self-reference via the registry, and this prevent GC even if the socket isn't in scope anywhere in Lua.  This can happen in circumstances like:
```Lua
do
  local svr = net.createServer(net.TCP)
  srv:listen(888*,accept_cb)
end
```
Here without this reference, the GC could clean up the socket after exiting the `do` .. `end` and so it would have been freed by the time the `accept_cb()` fires.   Unreferencing an entry in the Lua Registry can trigger an ECG run which will reclaim the userdata before the code is finished with it, hence this cludge.  The correct Lua approach is to "turn out the lights when you leave the room", i.e. do the unregister last when the function doesn't need the userdata any more.

jmattsson:
Happy to see this merged after conflicts resolved.

TerryE:
Sorry guys, I am now on a Greek Island and my only access is by logging on from my local Taverna.  

I have also been fire-fighting an issue with a forum where I am an active contributor and its gone belly-up because the sole-owner decided to take his ball and go home (shut it down). With most of the mods and senior contributors, I am trying to do a phoenix -- the only prob is that  am the only experienced Sysdmin / PHP programmer / MySQL admin so it is sucking up all of my spare time for next week or so.

djphoenix:
Should I recreate this PR with resolved conflicts? Have merged to my repo successfully, all seems OK.

pjsg:
@djphoenix -- yes, please restage and we can get this squared away....

PR #1175, opened 2016-03-20
2 new modules for HW sensors.

igrowing:


lua_modules/sr04/sr04.lua
```diff
@@ -0,0 +1,101 @@
...
+    end
+end
+
+function M.poll_distance()
+---  Read and print distance periodically by SAMPLE_INTERVAL.  ---
+-- TODO: Rewrite the print with callback: True value comes with SAMPLE_INTERVAL delay.
+    tmr.alarm(TID, SAMPLE_INTERVAL, 1, function()
+        _trig()
+        local d = _calc()
+        if d > 1 then print(d) else print("Timeout.") end
```
dvv: How about calling provided callback here instead of hardcoded printing?

lua_modules/sr04/sr04.lua
```diff
@@ -0,0 +1,101 @@
...
+    gpio.write(TRIG_PIN, gpio.HIGH)
+    tmr.delay(10)
+    gpio.write(TRIG_PIN, gpio.LOW)
+end
+
+local function _calc()  -- Internal function
+    local d = _end - _start
+    if d > 58 and d < 30000 then -- ~5m max distance
+        _end = 0  -- Disable wrong calculation on next attempt.
+        return math.floor(d/58)
```
dvv: Those who use integer firmware will receive inaccurate "quantized" results. Multiply by 100|1000 before flooring?

lua_modules/sr04/sr04.lua
```diff
@@ -0,0 +1,101 @@
...
+            _end = tmr.now() 
+        end
+    end
+    gpio.trig(ECHO_PIN, "both", _trig_cb)
+end
+
+local function _trig()  -- Internal function
+    -- Trigger the measure
+    gpio.write(TRIG_PIN, gpio.HIGH)
+    tmr.delay(10)
```
dvv: Wonder why sync delay here while the caller uses callback-style?

lua_modules/sr04/sr04.lua
```diff
@@ -0,0 +1,101 @@
...
+function M.stop_poll()
+---  Stop distance polling. Clean.  ---
+    tmr.stop(TID)
+    tmr.unregister(TID)
+    collectgarbage()
+end
+
+function M.get_distance(callback)
+---  Read distance, return value in cm or -1.  ---
+    _trig()
```
dvv: See comment to `_trig()`

gsker: This module did not work as expected.  It has surprises.
First I tried something very naive:
```
> sr04 = require("sr04")
> sr04.init(2,1)
> =sr04.get_distance()
> PANIC: unprotected error in call to Lua API (sr04.lua:96: attempt to call upvalue '?' (a ni)
c_�RS�fJS�fJ�j��r
```
Ouch!  Of course it didn't work, but it shouldn't stab me in the back. :smile: 

Following verbatim the *example* at the top of the module:
```
> sr04 = require("sr04")
> sr04.init(trig_pin=3, echo_pin=4) 
stdin:1: ')' expected near '=' 
```
The example should be something I would actually use and not a description.

Extrapolating, I tried this and got a -1:
```
> sr04 = require("sr04")
> sr04.init(2,1)
> sr04.get_distance(function(distance) print(distance) end)
> -1
```

Looking at Example 2: and extrapolating a bit further:
```
> sr04 = require("sr04")
> sr04.init(1,2000,2,1)
> sr04.get_distance(function(distance) print(distance) end)
> 67
```
That seems to work. As does the polling. It prints the "Timeout." message a lot, but when it is returning data it seems to be real data.

I'm probably not the person to be commenting on code, but I was just looking for an SR04 esp8266 solution and figured I'd share my test.

marcelstoer:
@igrowing pls give feedback to those who reviewed your code.

marcelstoer:
No feedback from contributor in two weeks -> closing. Vladimir & Gerry feel free to create an updated PR yourself if you think this would be worthwhile to have.

PR #1177, opened 2016-03-20
Remove Eclipse project files

marcelstoer:
I mentioned a while ago in a comment that I don't think it makes sense to keep Eclipse project files in the repository. The current group of committers is too diverse as for the editors/IDEs their members use (I reckon).

Here's the PR that will remove all Eclipse files.


TerryE:
My :+1: to this, I as tempted to do this merge now, but I felt that as this might be controversial for some contributors that I'd better leave it a few days for any counter-proposals.

pjsg:
:+1: from me -- I don;t use eclipse for this project. Even if I did, I'd want my own configuration....

dnc40085:
Would it be possible to add the removed files to `.gitignore` as well?

I use eclipse, so I plan on keeping the project files.

Just out of curiosity, what editor/IDE do you guys use?

TerryE:
The.gitignore suggestion is a good one. Gedit and lots of SSH windows to my VM, but I am a fossil.

marcelstoer:
> Would it be too much trouble to add the removed files to `.gitignore`?

No, good idea, thank you. I amended to commit.

PR #1179, opened 2016-03-20
MQTT Unsubscribe support

### LINKED ISSUE #1172: unsubscribe a mqtt topic
Hello,

The mqtt library does not provide unsubscribe feature. Will it be implemented?

pjsg:
This is the change to support unsubscribe in MQTT that is compatible with the subscribe method,

Tested to subscribe and unsubscribe from a single topic and multiple topics.

This should resolve #1172

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -125,10 +125,8 @@ static void mqtt_socket_disconnected(void *arg)    // tcp only
       mud->pesp_conn = NULL;
     }
 
-    if(mud->self_ref != LUA_NOREF){   // TODO: should we unref the client and delete it?
-      luaL_unref(L, LUA_REGISTRYINDEX, mud->self_ref);
-      mud->self_ref = LUA_NOREF; // unref this, and the mqtt.socket userdata will delete it self
-    }
+    luaL_unref(L, LUA_REGISTRYINDEX, mud->self_ref);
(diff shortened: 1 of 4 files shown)


app/modules/mqtt.c
```diff
@@ -1201,6 +1210,119 @@ static int mqtt_socket_on( lua_State* L )
...
+    stack++;
+    if( topic == NULL ){
+      return luaL_error( L, "need topic name" );
+    }
+    temp_msg = mqtt_msg_unsubscribe( &mud->mqtt_state.mqtt_connection, topic, &msg_id );
+  }
+
+  if( lua_type( L, stack ) == LUA_TFUNCTION || lua_type( L, stack ) == LUA_TLIGHTFUNCTION ) {    // TODO: this will overwrite the previous one.
+    lua_pushvalue( L, stack );  // copy argument (func) to the top of stack
+    if( mud->cb_unsuback_ref != LUA_NOREF )
```
jmattsson: This test is superfluous. luaL_unref() is a no-op on a LUA_NOREF or LUA_REFNIL.

jmattsson:
I don't use MQTT and don't have a setup to take it for a spin, but from a distance this looks good.

app/modules/mqtt.c
pjsg: I fixed all of these in this file.

Issue #1178, opened 2016-03-20
OLED with SSD1306 giving random dots


maartenbreddels:
Bought this oled:
http://www.ebay.nl/itm/0-96-I2C-IIC-Serial-128X64-OLED-LCD-Display-Module-Arduino-Scientific-Device-/361028753804?
In nodemcu I cannot get it this OLED to work, it gives the random dots (or noise) many people complain about. I feared it had to do with the reset issue, but it works perfect with espeasy.
I noticed the initialization is a bit different, 
https://github.com/ESP8266nu/ESPEasy/blob/45571e0b06a80509fe35f467d800612453446adf/_P023_OLED.ino
I don't have time atm to try to fix this, but maybe someone finds this useful info for a PR.

UncleRedz:
I'm running something that looks the same, same controller and size of OLED. I get the random noise on first initialization after powering on. Once I draw something, the display is cleared and shows what it should. Doing a reset of the ESP while keeping power on, keeps the display as it was before resetting. Again, after first drawing it shows what it should. 

It's like it's not doing a clear of the display buffer when initializing the display. Apart from this, it works nicely. 

Is this what you are seeing as well? Or do you have random noise all the time when drawing?

devsaurus:
Thanks for your confirmation @UncleRedz.

@maartenbreddels clearing the display has to be done from the Lua script. Can you please post your Lua code here to investigate potential integration issues?

maartenbreddels:
I only was able to get noise, never got to display anything. I was playing with this snippet below using ESPlorer, I left the code as it was (with comment, so you know what I was trying). I also tried powering the OLED on and off, and resetting the esp, but nothing worked. Except that the first few lines (or bottom ones?) seems to be cleared.
```
id=0
sda=5
scl=6
addr=0x3c

i2c.setup(id,sda,scl,i2c.SLOW)
disp = u8g.ssd1306_128x64_i2c(addr)
--i2c.setup(id,sda,scl,i2c.SLOW)
disp:setFont(u8g.font_6x10)
disp:begin()
disp:firstPage()
--print(disp)
--disp:setColorIndex(1)
disp:drawBox(5, 10, 20, 10)
disp:drawBox(10, 20, 20, 20)
disp:drawStr(0, 20, "Blaat")
--disp:fillBox(0, 0, 50, 50)
disp:nextPage()
```

UncleRedz:
[...]

PR #1183, opened 2016-03-23
Example of u8glib draw loop

UncleRedz:
Example of an u8glib draw loop that works with WiFi by keeping the draw
loop execution time to a minimum. This was mentioned in issue #1178 as something that is missing in the existing u8glib lua examples which focus on demoing the u8glib functionality and not necessarily showing how the draw loop should be implemented in a nice way.


TerryE:
@UncleRedz I was just looking up your code API ref and I noticed a typo.  Can you add an update to `## u8g.disp:nextPage()` line  `## u8g.disp:nextPage()`+1 is missing a `]` after the `[u8glib nextPage()` ?

I also had a play with your example  -- at a compile level only since I don't have a u8g compatible display -- and came across one piece of Lua syntactic subtlety that I hadn't twigged before.  `local function drawPages()` is _not_ the same as  `local drawPages = function ()` as the former looks the local slot `drawPages` before starting the compile, so any recursive references correctly self-reference the function, but the latter compiles the function then books the local so the local doesn't exists during the function compile and any recursive references actually reference a (non-existent) global function.  Have a play with `luac.cross -l` to see what I mean.  

Another good hint is to search the listing for `GLOBAL` references and make sure that you are only using globals when you intend to.

You can also nested functions and use upvalues from the outer function within the inner function to create persistent variables.  I've create my cut down version of your u8g_drawloop.lua in [this gist](https://gist.github.com/TerryE/6c06c5f66496bea8a58b) which you might want to look at for some ideas / Lua techniques, and feel free to pinch bits if you like them.

It's also nice to see node.task.post being put to good use! :smile:  

Thanks for this, Terry

lua_examples/u8glib/u8g_drawloop.lua
```diff
@@ -0,0 +1,75 @@
+------------------------------------------------------------------------------
+-- u8glib example which shows how to implement the draw loop without causing
+-- timeout issues with the WiFi stack. This is done by drawing one page at
+-- a time, allowing the ESP SDK to do it's house keeping between the page
```
marcelstoer: It's 'its' rather than 'it's' :wink:

UncleRedz:
@TerryE, that's an interesting observation of the local function creation, I'll keep that in mind. I've looked at your gist and incorporated the parts which I think makes sense for this example. The draw loop is now more tightly contained / encapsulated.

As I'm new to git and github, what's the correct procedure from here? Squash the commits and rebase as described in the contributor documentation? It doesn't state how to handle updates to your PR.

marcelstoer:
> Squash the commits and rebase as described in the contributor documentation?

Yes, that's the correct.

UncleRedz:
OK, I think I've squashed and rebased now, please let me know if I there are any issues.

marcelstoer:
There are...you have six commits instead of one. I'll still merge this, though.

devsaurus:
@UncleRedz thanks a lot for showing how to benefit from the new task interface.
I'll update the remaining u8glib examples accordingly.

PR #1185, opened 2016-03-24
Add gpio_pin_wakeup_enable to gpio module

MarsTechHAN:
Used to wakeup the CPU during the light-sleep.
Usage:

gpio.wakeup(PIN, Wakeup Mode)

Example:

gpio.wakeup(3, gpio.LOW)

Means Attach the pin 3 to wakeup interrupt, and active when gpio is LOW

### DIFF
--- a/app/modules/gpio.c
+++ b/app/modules/gpio.c
@@ -59,6 +59,31 @@ static void gpio_intr_callback_task (task_param_t param, uint8 priority)
   }
 }
 
+//Lua: wakeup(pin, mode)
+static int lgpio_wakeup_enable(lua_State* L)
+{
+    unsigned pin, mode;
+    
+    pin = luaL_checkinteger(L, 1);
+    mode = luaL_checkinteger(L, 2);
+    
+    if(mode != HIGH && mode != LOW){
+        return luaL_error( L, "wrong arg type" );
+    }
+    
+    if(pin == 0){
+        return luaL_error( L, "no interrupt for D0" );
+    }
+    
+    if(pin > 12 || pin < 1){
+        return luaL_error( L, "wrong pin num" );
+    }
+    
+    gpio_pin_wakeup_enable(GPIO_ID_PIN(pin), mode); 
+    
+    return 0;
+}
+
 // Lua: trig( pin, type, function )
 static int lgpio_trig( lua_State* L )
 {
@@ -219,6 +244,7 @@ static const LUA_REG_TYPE gpio_map[] = {
 #ifdef GPIO_INTERRUPT_ENABLE
   { LSTRKEY( "trig" ),   LFUNCVAL( lgpio_trig ) },
   { LSTRKEY( "INT" ),    LNUMVAL( INTERRUPT ) },
+  { LSTRKEY("wakeup"),   LFUNCVAL(lgpio_wakeup_enable)},
 #endif
   { LSTRKEY( "OUTPUT" ),    LNUMVAL( OUTPUT ) },
   { LSTRKEY( "OPENDRAIN" ), LNUMVAL( OPENDRAIN ) },


TerryE:
@MarsTechHAN Martin, three comments:

- You use `luaL_error()` but `luaL_argcheck()` is the preferred Lua macro as this give better user diagnostics.  

-  What is your reference for limiting wakeup interrupts to pins 1-12? Espressif document `8B-ESP8266__Interface__GPIO_Registers__Release_15-11-2014.xls` lists `GPIO_PINn_WAKEUP_ENABLE` masks for all pins and `gpio_pin_wakeup_enable()` at ROM:00004e90 includes no pin validation.  As far as I can see lines 70:80 should just be:
```C
luaL_argcheck(L, platform_gpio_exists(pin) && pin>0, 1, "Invalid interrupt pin");
luaL_argcheck(L, mode == LOW || mode == HIGH, 2, "Invalid pin mode" );
```

- The file `docs/en/modules/gpio.md` needs to include the documentation much as per your PR description.

MarsTechHAN:
@TerryE Thanks you for the comments, I will reopen an PR later after fix the problems.

TerryE:
@MarsTechHAN @jmattsson Martin, Johny, as far as I know there are only 3 types of sleep: 

-  `light` and `modem` which are initiated by `wifi.sleeptype()`
-  `deep` which is initiated either by `node.dsleep()` or `rtctime.dsleep()`.

The first two only shut down (parts of) the wifi submodule, but the CPU stays running; the last shuts down everything apart from the RTC and its NVRAM so a hard reset takes place on wakeup. 

Hence this function could only sensibly apply to deep sleep, but the [node.dsleep() documentation](https://nodemcu.readthedocs.org/en/dev/en/modules/node/#nodedsleep) says that only the RST and GPIO16 pins can be used and that GPIO16 is falling edge triggered.   So cam some explain the context of this new function?

Also the documentation piece needs to be before `## gpio.write()` and should be something like:

    ## gpio.wakeup()

    Select digital GPIO pin to wakeup the CPU during the light-sleep.

    #### Syntax
    `gpio.wakeup(pin,mode)`

    #### Parameters
    - `pin` pin to wakeup the CPU, IO index
    - `mode` one of `gpio.LOW` or `gpio.HIGH`

    #### Returns
    `nil`

    #### Example
    ```lua
    gpio.wakeup(3, gpio.LOW)
    ```
    #### See also
    -  [`wifi.sleeptype()`](wifi/#wifisleeptype)
    -  [`node.dsleep()`](node/#nodedsleep)
    -  [`rtctime.dsleep()`](rtctime/#rtctimedsleep)

mikewen:
@Terry, I think the light sleep, CPU is NOT running, only RTC clock is
running, that is why power consumption is low and only INT can wake it up.

@Martin, Terry's point make sense, it is good to consider sleeptype and
gpio.wakeup together.

It is a cool feature ? Thanks, guys!

TerryE:
@mikewen, sorry for my mindfart. This Espressif article, [ESP8266 Power Consumption](http://bbs.espressif.com/viewtopic.php?t=133), summarises that whereas Modem-Sleep keeps the CPU running (~15 mA current draw), light sleep suspends the CPU but powers the RAM and internal state (~0.5mA) and deep sleep shut down everything apart from the RTC (~0.01mA).  So this function relates to waking from light sleep.

mikewen:
@TerryE, There is nothing to sorry for, you are great and know a lot, so
are most people in this community. Everybody miss small details sometimes.

NodeMCU is getting better and better. I will merge some of my codes when I
found some free time.

0xbaadf00d:
Hello,

any news about this PR ?  We looking for a solution to "wake up" the NodeMCU from a deep sleep when user push a button or when a NFC card is detecter by our NFC reader... The documentation speak about RST PIN, but we don't want to reboot the device...

jmattsson:
@0xbaadf00d once you're in deep sleep, the only path out is via a reset. From what I could tell, it's due to a bug in the on-chip ROM loader where it doesn't properly re-initialise the flash chip and thus can't resume properly. As a workaround the RTC gets configured to pulse a particular gpio line on wakeup, and that pin is then connected to reset (on most modules).

In reality this actually works surprisingly well. You can use `node.bootreason()` to work out whether you just came out of deep sleep, and you can use `rtcmem` to squirrel a bit of state away across resets, or the filesystem if you're more heavyweight.

0xbaadf00d:
I understand, Thank you @jmattsson

PR #1190, opened 2016-03-25
Add example for socket:send()

marcelstoer:
Contributes to #730

### DIFF
--- a/docs/en/modules/net.md
+++ b/docs/en/modules/net.md
@@ -100,10 +100,10 @@ Listen on port from IP address.
 sv = net.createServer(net.TCP, 30)
 -- server listens on 80, if data received, print data to console and send "hello world" back to caller
 sv:listen(80, function(c)
-	c:on("receive", function(c, pl) 
-		print(pl)
-	end)
-	c:send("hello world")
+  c:on("receive", function(c, pl) 
+    print(pl)
+  end)
+  c:send("hello world")
 end)
 ```
 
@@ -184,12 +184,12 @@ Register callback functions for specific events.
 
 #### Example
 ```lua
-sk = net.createConnection(net.TCP, 0)
-sk:on("receive", function(sck, c) print(c) end )
-sk:connect(80,"192.168.0.66")
-sk:on("connection", function(sck,c)
-    -- Wait for connection before sending.
-    sk:send("GET / HTTP/1.1\r\nHost: 192.168.0.66\r\nConnection: keep-alive\r\nAccept: */*\r\n\r\n")
+srv = net.createConnection(net.TCP, 0)
+srv:on("receive", function(sck, c) print(c) end)
+srv:connect(80,"192.168.0.66")
+srv:on("connection", function(sck, c)
+  -- Wait for connection before sending.
+  sck:send("GET / HTTP/1.1\r\nHost: 192.168.0.66\r\nConnection: keep-alive\r\nAccept: */*\r\n\r\n")
 end)
 ```
 
@@ -201,7 +201,9 @@ end)
 Sends data to server.
 
 #### Syntax
-`send(string, function(sent))`
+`send(string[, function(sent)])`
+
+`sck:send(data, fnA)` is functionally equivalent to `sck:send(data) sck:on("sent", fnA)`.
 
 #### Parameters
 - `string` data in string which will be sent to server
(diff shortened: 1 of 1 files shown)


docs/en/modules/net.md
```diff
@@ -188,8 +188,8 @@ sk = net.createConnection(net.TCP, 0)
 sk:on("receive", function(sck, c) print(c) end )
 sk:connect(80,"192.168.0.66")
 sk:on("connection", function(sck,c)
-    -- Wait for connection before sending.
-    sk:send("GET / HTTP/1.1\r\nHost: 192.168.0.66\r\nConnection: keep-alive\r\nAccept: */*\r\n\r\n")
+  -- Wait for connection before sending.
+  sk:send("GET / HTTP/1.1\r\nHost: 192.168.0.66\r\nConnection: keep-alive\r\nAccept: */*\r\n\r\n")
```
pjsg: This should be sck:send rather than sk (otherwise it causes a reference loop)

docs/en/modules/net.md
```diff
@@ -212,7 +212,38 @@ Sends data to server.
...
+
+#### Example
+```lua
+srv = net.createServer(net.TCP)
+srv:listen(80, function(conn)
+  conn:on("receive", function(sck, req)
+    local response = {}
+
+    -- if you're sending back HTML over HTTP you'll want something like this instead
+    -- local response = {"HTTP/1.0 200 OK\r\nServer: NodeMCU on ESP8266\r\nContent-Type: text/html\r\n\r\n"}
```
pjsg: Add a note to say that this does require that all the data be in memory at one time. 

Sending data from a file is left as an exercise for the reader!

docs/en/modules/net.md
marcelstoer: Ahh right, had wanted to fix this for a long time but forgot again :frowning: Thanks, I'll fix it.

docs/en/modules/net.md
marcelstoer: I thought about it before I committed but then concluded that it be really very obvious. Instead, would it be helpful to show what to do if you want to *avoid* that? I could add an extra example like so:

```lua
sck:send(header, function() 
  local data1 = "some large chunk of data"
  sck:send(data, function()
    local data2 = "even more data"
    sck:send(data2, function() 
      sck:close()
    end)
  end)
end)
```

TerryE:
Actually the `net.md` file could do with being clarified as far as the `send()` call is concerned.  `sck:send(rec,fnA)` is pretty much syntactic sugar for `sck:send(rec) sck:("sent",fnA)`

I also think that this example is getting far too complicated for an inline example in the documentation.  If we start doing this then the documentation just bloats too much. Surely it is better just to write a decent example for `lua_examples` and refer out to this.

Having said that the current http_server example is just terrible.  It won't even work with the current firmware build.  Sigh!!! Just so much to do. :cry:

marcelstoer:
> I also think that this example is getting far too complicated for an inline example in the documentation. If we start doing this then the documentation just bloats too much.

True, we agreed a long time ago that full-blown examples should be kept away from the docs and I (still) fully support that. IMO this doesn't qualify as a full-blown example. Furthermore, the note about "Multiple consecutive `send()` calls" is a good hint but you probably also noticed how much that still seems to be an issue for many developers. If a code example can help avoid GitHub issues, SO questions and forum posts then its existence is justified.

> Having said that the current http_server example is just terrible.

-> #1196

> Sigh!!! Just so much to do. :cry:

Don't lose heart! No need to look at whole pile of issues. We'll fix one at the time :smile:.

marcelstoer:
All remarks incorporated - except for Terry's "doesn't even belong here" :wink:

TerryE:
Just walked my daughter down the isle on Sunday. So I am still a happy man at the mo'

devsaurus:
Sorry to jump in here - I was afk the last days.
Back in the mqtt discussions, @dvv mentioned his [mqtt-queue-helper](https://github.com/dvv/nodemcu-thingies/blob/master/mqtt-queue-helper.lua). It's a quite elegant way to overcome the queueing hassle IMO. Would it help to port that to `net` and provide it in `lua_examples`?

marcelstoer:
Sure, if you track that in a new issue and also reference #993 from it.

pjsg:
I'm not sure that the mqtt-queue-helper is needed any more. There used to be a bug in mqtt where it wouldn't send back-to-back publish calls and the mqtt-queue-helper would have fixed that. However, as part of the cleanup that I did in the MQTT code, it now does (AFAIK) send publish calls at full rate.

PR #1192, opened 2016-03-26
Add regular sends to mdns. Check for (some) buffer overflows. Make it handle unicast

pjsg:
This code was *ugly*. It is only a little better now. Some of the ugliness is due to the espressif sdk referencing (I think) into this code (so I can't change the mdns_info structure).

Anyway, the changes are as follows:

* Retransmit the registration packet every 120 seconds. This keeps most oses happy.
* Make it handle requests sent directly to port 5353 and respond with its standard packet.

There is still stuff that I don't understand in this code. I'm reluctant to rip it out entirely. 

I'm expecting a bunch of feedback on this one. 

It still doesn't work unless you are in STATION mode. It tries to do stuff in the other modes, but it doesn't appear to work. I'm sure that you can corrupt memory by passing in really long fields. So don't do that!

### DIFF
--- a/app/lwip/core/mdns.c
+++ b/app/lwip/core/mdns.c
@@ -202,6 +202,9 @@ static uint8 mdns_flag = 0;
 //#if (DNS_USES_STATIC_BUF == 1)
 static u8_t mdns_payload[DNS_MSG_SIZE];
 //#endif /* (MDNS_USES_STATIC_BUF == 1) */
+void ICACHE_FLASH_ATTR
+mdns_set_hostname(char *name);
+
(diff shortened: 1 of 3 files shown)


TerryE:
@pjsg Philip, I've been thinking about this and I am not sure that this is the correct thing to do.  The mdns code is Espressif's and surely we should be upstreamig any changes to them to integrate into their future SDK releases, otherwise we could create a fork nightmare in maintenance terms.  See #1195.

pjsg:
Closed as superseded by #1197

Issue #1195, opened 2016-03-26
Modification of LwIP


TerryE:
Issues #1166, #1181 and PR #1192 highlight an issue where we might be digging ourselves a maintenance hole.  Espressif open-sourced their implementation of LwIP in SDK 1.3, based on the [LwIP project](http://savannah.nongnu.org/projects/lwip/)'s version 1.4.0 release dated 6 May 2011 (though the LwIP licence terms don't require this).  (See `app/include/lwip/init.h` for version info).  

The current stable LwIP release is 1.4.1 dated 17 Dec 2012, and version 1.5 is still under beta development.

The Espressif implementation comprises a subset of the standard distro (e.g. ppp slip, etc. has been stripped out) that has a lot of added unprintable comments (not ASCII or UTF-8, maybe some DBCS Chinese font?) with some Espressif additions (mdns.c and smtp.c -- though these my be extracts from LwIP applications) and an "app" wrapper which is also Espressif code.

I suspect that our urge will be to fix the buggy areas -- which will tend to be the Espressif additions as the core project code has been hammered.   However, we are changing a fork of a fork which is always going to be a major maintenance headache, so before we do this I feel that we should at least discuss this issue an reach a consensus amongst the contributors before we go down this route.  

My personal view is that we should avoid bug-fixes to Espressif-supplied code which they themselves might change across SDK versions, and that we should upstream and fixes to Espressif.

If we do change this code then we should do it in a very controlled method -- as Johny has set up in his skeleton `sdk-overrides`, by copying in the sdk components into our firmware tree then making any changes. though I would prefer one or more `.patch` files here so it very clear what changes we are making.

We should also make these on a very conscious basis where there is a clear case for doing so.

TerryE:
By way of example on my last point, in my view there one area where we should consider changing from the core LwIP library and that is the memory allocator.  This is a hook in LwIP and there are various configuration options to tailor this.  As far as I can see, Espressif have simply mapped this to os_malloc.

What I am not sure about is how robust LwIP is if memory exhaustion occurs.  It's typical strategy is to do a `if(!p) return 0` type of abort, but the upper levels often don't handle this zero status return correctly so weird failures can occur that will be very hard to diagnose when the fundamental issue is that the app has exhausted the heap.

[...]

PR #1197, opened 2016-03-28
Take 2: Add regular sends to mdns. Check for (some) buffer overflows. Make it handle unicast 

pjsg:
This is the second take on #1192.

This code was ugly. It is better now. We now have our own mdns implementation which is a hacked up version of the code from the SDK. However, this is our version and does something sensible.

Anyway, the changes are as follows:

Retransmit the registration packet every 120 seconds. This keeps most oses happy.
Make it handle requests sent directly to port 5353 and respond with its standard packet.

Transmits the packet that identifies the service name every 1200 seconds.

I'm expecting a bunch of feedback on this one. I did add c_strlcpy, c_strlcat, c_strdup to c_string.c as they are generally useful. The strl* family of functions does what strncpy and strncat ought to do.

It still doesn't work unless you are in STATION mode. It tries to do stuff in the other modes, but it doesn't appear to work.

### DIFF
--- a/app/modules/mdns.c
+++ b/app/modules/mdns.c
@@ -1,4 +1,4 @@
-// Module for access to the espconn_mdns functions
+// Module for access to the nodemcu_mdns functions
 
 #include "module.h"
 #include "lauxlib.h"
@@ -9,11 +9,11 @@
 #include "c_types.h"
 #include "mem.h"
 #include "lwip/ip_addr.h"
-#include "espconn.h"
+#include "nodemcu_mdns.h"
 #include "user_interface.h"
 
 typedef struct wrapper {
-  struct mdns_info mdns_info;
+  struct nodemcu_mdns_info mdns_info;
   char data;
 } wrapper_t;
 
@@ -38,7 +38,7 @@ static char *advance_over_string(char *s)
 static int mdns_close(lua_State *L)
 {
   if (info) {
-    espconn_mdns_close();
+    nodemcu_mdns_close();
     c_free(info);
     info = NULL;
   }
@@ -55,6 +55,8 @@ static wrapper_t *process_args(lua_State *L, phase_t phase, size_t *sizep)
 {
   wrapper_t *result = NULL;
   char *p = NULL;
+  bool has_desc = !lua_isnumber(L, 3);
+  int arg_no = 1;
 
   if (phase == PHASE_COPY_DATA) {
     result = (wrapper_t *) c_zalloc(sizeof(wrapper_t) + *sizep);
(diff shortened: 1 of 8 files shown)


app/net/nodemcu_mdns.c
```diff
@@ -0,0 +1,947 @@
...
+	multicast_addr.addr = DNS_MULTICAST_ADDRESS;
+	struct ip_addr ap_host_addr;
+	struct ip_info ipconfig;
+	ms_info = info;		// Save the passed block. We need all the data forever
+
+	if (mdns_payload) {
+	  os_free(mdns_payload);
+	}
+	mdns_payload = (u8_t *) os_malloc(DNS_MSG_SIZE);
+	if (!mdns_payload) {
```
TerryE: This sort of logic produced extremely difficult-to-diagnose failures.  `nodemcu_init_info()`is a void function so  simply returning like this on allocation failure means that this init function simply fails silently.  No good.

app/net/nodemcu_mdns.c
```diff
@@ -0,0 +1,947 @@
...
+	} else {
+		PUCK_PORT = ms_info->server_port;
+	}
+
+	/* initialize mDNS */
+	mdns_pcb = udp_new();
+
+	if (mdns_pcb != NULL) {
+		/* join to the multicast address 224.0.0.251 */
+		if(wifi_get_opmode() == 0x03 || wifi_get_opmode() == 0x01) {
```
TerryE: surely we should `uint8 op_mode = wifi_get_opmode();` here to cache the mode.

app/modules/mdns.c
```diff
@@ -65,28 +67,39 @@ static wrapper_t *process_args(lua_State *L, phase_t phase, size_t *sizep)
   }
 
   if (phase == PHASE_CALCULATE_LENGTH) {
-    luaL_checktype(L, 1, LUA_TSTRING);
-    luaL_checktype(L, 2, LUA_TSTRING);
-    (void) luaL_checkinteger(L, 3);
-    *sizep += c_strlen(luaL_checkstring(L, 1)) + 1;
-    *sizep += c_strlen(luaL_checkstring(L, 2)) + 1;
+    luaL_checktype(L, arg_no, LUA_TSTRING);
```
TerryE: This isn't the standard Lua way of doing this, so I don't understand the rationale here.  For example, I would expect these lines to be something like:
```C
    host = luaL_checklstring (L, arg_no++, &hostlen);
    service = luaL_checklstring (L, arg_no++, &servicelen);
    if (lua_type(L, 4) == LUA_NUMBER) {
      service_desc = luaL_checklstring (L, arg_no++, &service_desclen);
    } else {
      service_desc = NULL;
      service_desclen = 0;
    }
    port_no = luaL_checkinteger(L, arg_no++);
```
etc.

TerryE:
Philip, this code is still flawed, IMO.  And the wrapper `mdns.c` isn't written in conventional Lua extension style.  It definitely need more work.

FrankX0:
Hi Philip. The functionality supplied still needs further improvement.
In response to the IP multicast query to port 5353 it is now responding with its registration packet. This is not correct. In fact it must respond with its domain name and its IPv4 and/or IPv6 address, depending on the request (A or AAAA record).
If this is implemented correctly, I think the retransmission of the registration packet every 120 seconds is not necessary.
In fact, but my knowledge might be limited, it seems to me mDNS support could be simplified considerably by only implementing response to the IP multicast query, as described in https://en.wikipedia.org/wiki/Multicast_DNS. At least this seems enough for the browsers I checked so far.

pjsg:
@FrankX0 The issue with only supporting the unicast query is that the device is not findable through browsing. Most systems build a local cache of all the mdns capable devices and use that cache to discover what is out there.

Yes, if you do the unicast query, you get back the full packet. This is allowed as the response packet contains the answer to the original question. 

I have done a fair amount of looking at my local network for what actually happens, and systems all appear to send their big packets every few minutes (but less than the TTL in the packet). The service broadcast is sent much less often, but again it is more often than the TTL in the packet. 

One thing to note is that most other mdns capable systems include the question in the response packet. This implementation does not. I don't know whether this is important or not. 

There is one thing that I'll probably fix -- and that is to ignore queries of the wrong type. E.g. asking for AAAA for the foo.local name.

pjsg:
@terrye What don't you like about the module/mdns.c? I admit that the argument stuff is a little odd. It may be that the best thing is to replace the arguments with a single table with some specially named entries.This would break backwards compatibility -- but that probably isn't very important (yet). I could also move the memory management into the lower level code.

I can certainly work some more on hacking up the original mdns code, but I'm reluctant to make lots of changes to it. It will still look horrible. It could be that the best solution is to rewrite it from scratch, but I don't have the energy for that.

TerryE:
@pjsg  Philip, It is largely complicated because the original mdns LWiP code is so weirdly coded, but we could just put our hands up a say "Sorry but this is core Espressif", but we've forked this and decided to reimplement -- except that we haven't: we've tidied up some bits, but we're still coding around others.  What we've now got is neither one thing nor t'other.  I feel that is is a mistake and causing a longer term maintenance issue.  We *really* shouldn't slide into this sort of situation bottom up by tactical PRs which is why I tried to raise this more as a technical policy issue for contributors to agree on in #1195.

In my view we should reach this consensus *before* we accept any PRs which fork Espressif core code.

docs/en/modules/mdns.md
```diff
@@ -9,14 +9,17 @@
...
-- `servicename` The service name for this device. Alphanumeric characters are best. This will get prefixed with `_` and suffixed with `._tcp`
-- `port` The port number for the primary service.
-- `attributes` A optional table of up to 10 attributes to be exposed. The keys must all be strings.
+- `attributes` A optional table of options. The keys must all be strings.
+
+The `attributes` contains two sorts of attributes -- those with specific names, and all the rest. The specific names are:
+
+- `port` The port number for the service. Default value is 80.
+- `service` The name of the service. Default value is 'http'.
+- `dscription` A short phrase (under 63 characters) describing the service. Default is the hostname.
```
TerryE: `description` not `dscription`

"all the rest": this is very enigmatic.  What and why are they allowed?  Even if we just refer out to a standard or wiki page which defines what optional parameters are allow; this would help a lot.

app/modules/mdns.c
```diff
@@ -12,125 +12,53 @@
...
-        result->mdns_info.txt_data[slot++] = p;
-	p = advance_over_string(p);
-
-	// now smash in the value
+	int len = c_strlen(key) + 1;
 	const char *value = luaL_checkstring(L, -1);
-	p[-1] = '=';
-	c_strcpy(p, value);
-	p = advance_over_string(p);
+	char *p = alloca(len + c_strlen(value) + 1);
```
TerryE: `alloca` allocates a variable string on the stack frame which goes out of scope on the function exit and the contents are invalid, and hence the  `info.txt_data[]` entry becomes invalid.

TerryE:
But having had a quick review, the `mdns.c` is now :+1: :smile:

pjsg:
I think that, in this case, I've really taken the puck code written by that student and improved it for general use. I've left the original mdns code from espressif in the tree. It still gets built, but there is no lua reference to it, so it doesn't get included in the resulting image.

I think that modifying espressif components in place is a dangerous path as it becomes more and more difficult to maintain as they release changes. For precisely this reason, the version of spiffs that I will drop when we are done with the current dev->master stuff will be unmodified code and hence easy to upgrade. Yes, there is a little bit of grubbiness in the wrapper to make this work, but I think that that is a good tradeoff.

pjsg:
Is there anything else that needs to happen for this one?

TerryE:
Jeremy,  As I said, I've looked through the code and it all looks fine to me.  I just do my provisioning an entirely different way (the virgin esp8266 hits a well known UDP provisioning service with a provisioning request and gets all its config and initial FS from the provisioning service; no input is required) so I don't use EUS and don't offer services from my esp8266 by mdns, so I am not set up to test this.  However, if you've tested it then I think that we can leave it to you to merge.  :+1:

Issue #1119, opened 2016-03-04
Adding an error traceback to panic()


TerryE:
As part of my LCD patch, I also reimplemented the Lua error traceback, so that by default an error give a proper Lua stack traceback, rather the a one-line goodbye.  However, the RTS still has a catch-all panic() to trap any unprotected error calls.  This puts out a very unhelpful error message: `PANIC: unprotected error in call to Lua API (<Lua routine name>)`.  It just struck me that it would be better to provide a proper Lua traceback.  Is this a good idea?

This is a fairly easy change.

pjsg:
Yes please.

dnc40085:
more debug info is always good

TerryE:
This isn't quite as simple as I first thought.  The cb stubs in all of the modules typically follow this type of pattern as in `net.c`:
```C
  if(nud->cb_disconnect_ref != LUA_NOREF && nud->self_ref != LUA_NOREF) {
    lua_rawgeti(gL, LUA_REGISTRYINDEX, nud->cb_disconnect_ref);
    lua_rawgeti(gL, LUA_REGISTRYINDEX, nud->self_ref);
    lua_call(gL, 1, 0);
  }
```
The issue is that if the cb_disconnect throws an error then an error handler hasn't been established so 
 `ldo.c:luaD_throw()` follows the code patch for the error handler not established that is it clears down the stack before calling `G(L)->panic(L)` if the registered. BTW commenting out the c_exit() is a definite bug.  We should use Johny's trick of doing `while(1) {}` to trip the WDT.

I need to play around to see exactly when an error panics, but I suspect that we will need some form of lua_call variant (and not just the standard xpcall) so I can do a glocal search and replace of these cb's in the modules.  

More work is needed.

TerryE:
What is different from a standard Batch Lua is that with this there is a single call to Lua. An unprotected error results in program termination.  A typical nodeMCU application is fragmented into many callback tasks. It would be a really nice development feature to be able to declare a `node.atpanic(action_routine, type) ` so that uncaught errors could either be trapped by a user exception handler, or simply produce an error traceback and return you to the > prompt.

Note that this would be quite a complex change since you would want to turn off all further callbacks to prevent the application continuing in some indetermined way.  Need to think about whether the advantages of doing this merit the effort involved.

pjsg:
[...]

PR #1199, opened 2016-03-28
First cut of new history module

TerryE:
This is a first cut of my variable length circular logging module.  The idea is that can enable you to log to RAM in a circular buffer which fairly minimal runtime overhead (instead of needing a UART connected).  At given checkpoints the entire buffer or the last n records can be dumped out, written to file or whatever.     See #1119 for my discussion on how this would be used.

I would suggest we need possibly four more changes to make this really useful: 

- an extra flag so that a 4-byte time.now() is prefixed to every record if set.
- the extra `lua_tcall()` which hooks into b:atpanic() 
- the extra logic to dump records to the UART or to a given UDP listener.
- a small python script in the tools directory to run server-side to log these records.

### DIFF
--- a/app/modules/history.c
+++ b/app/modules/history.c
@@ -1,0 +1,230 @@
+/*
+ * app/modules/history.c
+ *
+ * See Copyright Notice in lua.h
+ */
+#include "lua.h"
+#include "lmem.h"
+#include "lauxlib.h"
+#include "module.h"
+#include "user_modules.h"
+#include "c_string.h"
+
+#define buffer_HEADER_SIZE   sizeof(buffer);
+#define START_REC             "\x0d*** Start ***"
+#define START_REC_LEN         (sizeof(START_REC)-1)
+#define BUFFER_METATABLE     "history.buffer"
+
+#define RETRIEVE_HEADER(L) \
+  buffer *hdr = (buffer *)luaL_checkudata(L, 1, BUFFER_METATABLE)
+#define RETRIEVE_BUFFER(L) \
+  RETRIEVE_HEADER(L); byte   *buf = hdr->header + (sizeof(hdr->header))
+  
+#define WRAP1(fld) do { if (fld == hdr->size) fld = 0; } while (0)
+#define WRAP(fld)  do { if (fld >= hdr->size) fld -= hdr->size; } while (0)
+
+#define INVALID               ((uint16) -1)
+
+#ifdef DEBUG_HISTORY
+#define DUMP_HEADER(h,s) c_printf("hdr:%s  %6u  %6u  %6u  %6u  %6u  %6u\n", \
+        #s, h->size, h->head, h->tail, h->free, h->reader, h->nrecs)
+#else
+#define DUMP_HEADER(h,s)
+#endif
+
+typedef unsigned char byte;
+typedef unsigned short int uint16;
+
+typedef union {
+  struct {
+	  uint16 size, head, tail, free, reader, nrecs;
+	};
+	struct {
+	  byte header[6		 * sizeof(uint16)];
+	};
+} buffer;
+
+// Move the tail or reader offset onto the next record
+static uint16 bump(buffer *hdr, uint16 *hdr_field) {
+  byte  *buf = hdr->header + (sizeof(hdr->header));
+  
+  uint16  len = buf[(*hdr_field)++];
+  WRAP1(*hdr_field);
+  
+  if (len & 0x80) {
+    len += (buf[(*hdr_field)++]<<7) - 0x80;
+    WRAP1(*hdr_field);
+  }
+
+  *hdr_field += len;
+  WRAP(*hdr_field);
+  
+  return len;
+}
+
+// Add a records to the buffer with to break and wrap at the buffer top
+static void push(buffer *hdr, const   byte *rec, uint16 len) {
+  byte   *buf = hdr->header + (sizeof(hdr->header));
+  uint16 split = 0;
+
+  if ((hdr->head + len) > hdr->size) {
+    split = hdr->size - hdr->head;
+    memcpy(buf + hdr->head, rec, split);
+    rec += split;
+    hdr->head = 0;
+  }
+
+  memcpy(buf + hdr->head, rec, len - split);
+  hdr->head += len - split;
+  hdr->free -= len;
+}
+
+// Lua: handle = history.buffer(size)
+// Create a new cylic history buffer
+static int history_buffer( lua_State* L ) {
+  int size = luaL_checkinteger(L, 1);
+  luaL_argcheck(L, size >= 1024 && size <=32768, 1, "buffer size is 1-32Kb");
+  buffer *hdr = (buffer *)lua_newuserdata(L, size);
+  byte *buf = hdr->header + (sizeof(hdr->header));
+
+  size -= sizeof(hdr->header);
+  memset(buf, 0, size);
+
+  strcpy((char *)buf, START_REC);
+  hdr->size   = size;
+  hdr->head   = START_REC_LEN;
+  hdr->free   = size - START_REC_LEN;
+  hdr->tail   = 0;
+  hdr->nrecs  = 1;
+  hdr->reader = INVALID;
+
+  luaL_getmetatable(L, BUFFER_METATABLE);
+  lua_setmetatable(L, -2);
+  DUMP_HEADER(hdr,new);
+  return 1;
+}
+
+// Lua: buffer:add(data)
+// Lua: buffer:add(forrmat, ...)
+static int history_add( lua_State* L ) {
+  RETRIEVE_HEADER(L);
+  size_t len = 0, prefix_len = 1;
+  byte prefix_bytes[2];
+  lua_remove(L, 1);
+  int nargs = lua_gettop(L);
+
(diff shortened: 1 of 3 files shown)


jmattsson:
Some thoughts:

- Could `.find()` be called `.count()` instead? It seems more descriptive of its behaviour.
- For easy, compact storage would it make sense to also have a convenience function that applies a `struct` spec to the arguments, rather than the string formatted?
- I think the time-stamp should be user-supplied as the useful time domain might vary. Some use cases will want `tmr.now()`, others might want `rtctime.get()` and others yet might want a delta.
- Dumping logic would be best provided as an example, either in docs or examples dir, imo.

TerryE:
> Could `.find()` be called `.count()` instead? It seems more descriptive of its behaviour.

Not quite.  It is more like a seek function.  But it returns the count of records that can currently be read from the seek point.

>  For easy, compact storage would it make sense to also have a convenience function that applies a struct spec to the arguments, rather than the string formatted?
Packing structural content into a string belongs in a separate `pack` module.  `history` will historise strings which embed nullchars.

>   I think the time-stamp should be user-supplied as the useful time domain might vary. Some use cases will want tmr.now(), others might want rtctime.get() and others yet might want a delta.
In which case there is nothing to stop the user presenting this timestamp as the first arg in a formated argument list.  We can also play / extend this once we've had some experience of using it in practice.  My inclination is not to over-anticipate at this stage.

> Dumping logic would be best provided as an example, either in docs or examples dir, imo.
Agreed but what I really need to do is to implement the exception handler.

Not enough hours in the day what with building my own house, and the forum that I use all the time to help me in this going belly-up and my trying to do my bit to revive it !!

marcelstoer:
Terry, do you plan to continue here any time soon or shall we rather close it?

TerryE:
It's been sitting done, and I've been using it ever since I did the PR, but since no one else seemed interested, it didn't seem worth the effort of dancing around.

marcelstoer:
_I_ thought this a helpful new feature but I interpreted your last comment "Not enough hours in the day..." as if there were more changes you'd like to make but didn't have time for.

So, if you fixed the conflicts in `mkdocs.yml` we'd be ready for merge?

smil2k:
@TerryE: I also think this is a helpful feature. On my next project I also planned something like this to collect probably logs or maybe measurements together prior sending them to the server.

vsky279:
@TerryE I think this is valuable feature too. In my recent project I'm using lua circular buffer but of course it's not efficient and very memory demanding.  I was looking for this module in NodeMCU Documentation because I remember seeing it many months ago. I could not find. To my big surprise it's not merged yet.
So please go ahead and let's merge it.

TerryE:
Ok, I'll merge this into the current dev and repush.

vsky279:
@TerryE Hi Terry, any chance to merge it soon? Of course I can take your PR and use it individually but it would more comfortable to use it directly from dev.

TerryE:
@vsky279 I'll put it at the top of my TODO list after the LFS commit.

vsky279:
@TerryE Hi Terry, I started testing your history module.
My use case is a bit different than yours. I want to store all measured values for last 1 hour measured every 1 second so that I can calculate 1 hour, 5 minutes and 1 minute averages also every second.
Storing it in lua table is impossible - very memory demanding (even with LFS which really great!).

I realize that history module is for storing strings. I made an workaround to store the integer values as strings in binary form. I use the following procedures:
```Lua
local function numToBin(n)
    local hex=("%X"):format(n)
    local s=""
    for i=hex:len()-1,0, -2 do 
        s=s..string.char(tonumber("0x"..hex:sub(i==0 and 1 or i, i+1)))
    end
    return s
end

local function binToNum(b)
    local r = 0
    for i=1,b:len() do
        r = r + b:byte(i)*math.pow(256, i-1)
    end
    return r
end

local function average(...)
  local argn = arg.n
  local cnt, max = buffer:find(), 0
  local sum = {}
  for i=1,argn do 
    if cnt < arg[i] then arg[i]=cnt end
    if max < arg[i] then max = arg[i] end
    sum[i]=0
  end
  
  for i=1, buffer:find(-max) do
    local n = binToNum(buffer:readnext())
    for j=1,argn do
      if max-i<arg[j] then sum[j] = sum[j] + n end
    end
  end

  for j=1,argn do
      sum[j] = sum[j]/arg[j]
  end
  return unpack(sum)
end
```
I needed to remove the `START_REC` record to avoid interpreting it as measured value.

From performance view it is not very efficient but I want to avoid developing another module. I want to ask you what you think about this approach?
Should we extend the history module to make it more universal?

TerryE:
I can't recall the  details of the discussion, but I think it was @devsaurus  that suggested adding a general buffer library so that you deserve a fixed buffer and have read-write access into it.  I wrote this library when I was logging data from an analysis of contact bounce and realised that the print output was slower than the events that I was trying to analyse and so needed to write the log records to a RAM buffer dear than to try to dump them out to a UART.

For your application you can do this all in Lua with existing libraries.  You need to use three techniques:

-  Do the averages using the recurrence relationship **Ai = A(i-1) + Vi - V(i-61)** for the minute and similarly for the 5 and 60 minute averages.  This avoids the need for for loops, but you still need to be able to store and access 3600 historic samples.
-  Lua indexed arrays take a pointer and a TVALUE per value which is 16 bytes ~1K per minute so we can't fit an hour's data into RAM using this technique, but you can use a trick that I showed on my telnet module. In this case if once a minute you flush the integer array into a packed minute array, with each entry a 60×2 byte string, a 60 minute array comprises 60 off a pointer, a TVALUE, a TString and a 61 byte (plus 3 pack bytes) or 60 × 160 bytes or 9.6 Kb and so everything now comfortably fits.<br>
Each minute you replace the old entry for that minute slot and the LGC will garbage collect the old 120 long string.
-  You do need a routine which can index into the 60 × 120 char array to extract the needed 2 byte substring, but this is less than ten lines of Lua (use `struct.unpack()` with an explicit offset) and you need to make the call 3 times a second.

Also note that you can do the pack with a single line of Lua
```Lua
 minuteData[i] = struct.pack('<'..('h'):rep(60), unpack(secondData))
```
though you might want to cache the pack pattern in a local.

Hope this helps 😊

vsky279:
Hi Terry, thanks for all these ideas :+1:. 
So I wrote it purely in Lua as you suggested. Though the implementation is a bit tricky (if the average function should be more general).
```Lua
local math_min, math_max, math_floor = math.min, math.max, math.floor
local struct_pack, struct_unpack = struct.pack, struct.unpack

local function find(n)
  n = n - 1
  local n2 = n - #secondData
  local i, j = n2%60, math_floor(n2/60)
  if n2<0 then
    return secondData[#secondData-n]
  end
  local ret = struct_unpack("<H", minuteData[#minuteData-j], (60-i)*2-1)
  return ret
end

local function average(avg1, n)
  local cnt = #secondData + #minuteData * 60
  if cnt<=n or not avg1 then
    n = math_min(cnt, n)
    local sum, cntSD = 0, #secondData
    for i=cntSD, math_max(cntSD-n, 0)+1, -1 do sum = sum + secondData[i] end
    local n2 = n-cntSD
    if n2~=0 then
      local i2 = math_max(#minuteData-math_floor(n2/60), 1)
      for i=#minuteData, i2, -1 do
        local minute = {struct_unpack('<'..('H'):rep(60), minuteData[i])}
        for j=60, (i==i2 and (60-(n2-1)%60) or 1), -1 do sum = sum + minute[j] end
      end
    end
    return sum/n
  else
    return (avg1*n + find(1) - find(n+1))/n
  end
end

local function measure()
  secondData[#secondData+1] = measurement
  if #secondData>=60 then
    minuteData[#minuteData+1] = struct.pack('<'..('H'):rep(60), unpack(secondData))
    secondData={}
    if #minuteData>=60 then
      for i=1,#minuteData-1 do minuteData[i]=minuteData[i+1] end
      minuteData[#minuteData]=nil
    end
  end
end
```

TerryE:
@vsky279 Lukas, this doesn't belong here but it's fun.  I tend to go for brevity if it doesn't have a performance impact, e.g.
```Lua
local MIN_UNPACK = '<'..('H'):rep(60)
local function find(n)
  local n2 = n - 1 - #secondData
  local i, j = n2%60, math_floor(n2/60)
  return n2<0 and secondData[#secondData-n] or
                  struct_unpack("<H", minuteData[#minuteData-j], (60-i)*2-1)
end
```
and note hoisting that rep or even just do it in your editor: `"<HHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHH"` as you don't want to be doing this every iteration.

Also note the recurrence relationship.  If you are doing these every second then taking the minute average the the last one was `an_1 = 0; for i = 1, 60  do an_1 = an_1 + sample(-i ); end` and the current is  `an = 0; for i = 0, 59 do an = an + sample(-i); end` so you put these together and
`an = an_1 + sample[0] - sample[60]` which is simple, and the 5 and 60 min averages are the same.  This recurrence relationship collapses the compute load.  Just do this once every second whether or not you need it.

vsky279:
Thanks for the `MIN_UNPACK` suggestion.

As for the recurrence I am not sure to get it. Or you missed the else section of the average function:
```Lua
  return (avg1*n + find(1) - find(n+1))/n
```
?

PR #1200, opened 2016-03-29
enduser_setup: Fixed missing status update.

jmattsson:
Tagging @robertfoss @skycoders and @pjsg . This PR addresses the issue Konrad found the other day regarding the status updates not being loaded properly. The EUS module is now much nicer with connection closing.

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -392,6 +392,65 @@ static void enduser_setup_check_station(void *p)
 }
 
 
+/* --- Connection closing handling ----------------------------------------- */
+
+/* It is far more memory efficient to let the other end close the connection
+ * first and respond to that, than us initiating the closing. The latter
+ * seems to leave the pcb in a fin_wait state for a long time, which can
+ * starve us of memory over time.
+ *
+ * By instead using the poll function to schedule a hard abort a few seconds
+ * from now we achieve a deadline close. The downside is a (very) slight
+ * risk of dropping the connection early, but in this application that's
+ * hidden by the retries on the JavaScript side anyway.
+ */
+
+
+/* Callback on timeout to hard-close a connection */
+static err_t force_abort (void *arg, struct tcp_pcb *pcb)
+{
+  (void)arg;
+  tcp_poll (pcb, 0, 0);
+  tcp_abort (pcb);
+  return ERR_ABRT;
+}
+
+/* Callback to detect a remote-close of a connection */
+static err_t handle_remote_close (void *arg, struct tcp_pcb *pcb, struct pbuf *p, err_t err)
+{
+  (void)arg; (void)err;
+  if (p) /* server sent us data, just ACK and move on */
+  {
+    tcp_recved (pcb, p->tot_len);
+    pbuf_free (p);
+  }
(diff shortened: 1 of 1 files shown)


jmattsson:
This change to the disconnection handling also appears to have resolved the port-in-use error on a restart of the module.

robertfoss:
The nagle algorithm is disabled too, even better.

skycoders:
I have tested the build and observed the following positive effects:
* It did finish scanning for WiFi networks and went to idle
* I could connect to the port which I fire up in the completed callback, so this worked much more stable than before

This did not fully work:
* I have only one Wifi which the chip could find during scanning. So in the dropdown, there is only one entry, already showing the checkmark. If I enter the wifi password and click connect, nothing happens. Only when I type the name in the free text field, it actually starts connecting.
* Just once I saw the update message that it successfully connected to the WiFi, out of 5 tries.

I am using a MacBook Pro with latest Chrome browser. I will continue testing. But in general, this is an improvement, especially because of the port availability. Thanks!!

jmattsson:
Thanks for the testing & feedback!

- This sounds like the same issue we had in our custom "skin" for EUS at $work. Essentially, the text box with the SSID name doesn't get set to the slected SSID in the list when the list is first populated, yet that is the field which is used by the ESP to pick the network. Hence it effectively tries to connect to a network with no name if one only fills in the password and submits. Not too surprisingly, that doesn't work so well. Maybe @robertfoss could have a look at this? My JavaScript is a bit shoddy as you all might've noticed >.>

- Yep, that's the channel switching catching you again. @jfollas has some ideas for mitigation here which I believe he is exploring.

jmattsson:
@robertfoss Yeah Nagle in this use-case seemed quite counter-productive - we want to get this stuff out of our buffers quick smart dangit! =)

jfollas:
I have implemented my channel-changing workaround ideas in my private fork, and will work to bring into a public fork/PR. I'm traveling through Wednesday (USA time), so probably won't get to this until Thursday at the earliest.

Essentially, on startup, I do a AP Scan to identify the strongest AP and that's the channel that the SoftAP will use (so that when the Station tries to connect, it won't need to switch channels and drop the SoftAP connection). 

If the channel does need to change, then after the Station gets an IP address, I shut it down and start the SoftAP up again with the original channel - the hope here is that the phone/tablet/PC that is being used to configure the ESP is still listening for the AP on that channel.  Then the 10-second timer starts, and the "webserver" is still able to serve responses for /status. In this case, I don't fire the success callback until the end of the 10 seconds when the Station is brought back online.

Seems to work okay. I had some problems in testing with Android phones being aggressive about changing Wi-Fi too soon, but not sure there's anything that can be done about that right now.

robertfoss:
@jmattsson @skycoders: Adding some JS for 'default to only list entry' sounds do-able. I have slight shortage of time the coming days though.

I'll submit a PR real soon (tm).

robertfoss:
@jmattsson @skycoders: Fix submitted in PR #1206.

jmattsson:
Based on Robert's & Konrads testing (and mine!), I'm merging to get this in before the cut to master.

PR #1201, opened 2016-03-29
Change index.html to enduser_setup.html

marcelstoer:
I find it unfortunate that the enduser setup module, which've come to like more and more, claims a universal (and thus non-specific) file name like `index.html` for itself. Even though it's just used as a "backup", one which you get by default though, it should IMO use some specific prefix.

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -77,7 +77,7 @@ static const char dns_body[]   = { 0x00, 0x01, 0x00, 0x01,
 /*        DNS Answer Part          |LBL OFFS|  |  TYPE  |  |  CLASS |  |         TTL        |  | RD LEN | */
                                    0xC0, 0x0C, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00, 0x78, 0x00, 0x04 };
 
-static const char http_html_filename[] = "index.html";
+static const char http_html_filename[] = "enduser_setup.html";
 static const char http_header_200[] = "HTTP/1.1 200 OK\r\nCache-control:no-cache\r\nContent-Type: text/html\r\n"; /* Note single \r\n here! */
 static const char http_header_204[] = "HTTP/1.1 204 No Content\r\n\r\n";
 static const char http_header_302[] = "HTTP/1.1 302 Moved\r\nLocation: /\r\n\r\n";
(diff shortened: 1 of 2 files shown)


jfollas:
Or, maybe better yet, make it a #define in `user_config.h`, similar to the existing: 

`#define ENDUSER_SETUP_AP_SSID "SetupGadget"`

marcelstoer:
Oh no, I disagree. 

If it should be configurable, and I seriously doubt it should, then I'd rather achieve this through `enduser_setup.start()` or `enduser_setup.configure()` (doesn't exist yet) than through yet another #define. 

I just don't want to provoke name clashes by given a module-specific file a universal name that has implicit behavior in many contexts. There's absolutely no technical requirement for this file to be called `index.html`, it might just as well be `foo.bar`.

robertfoss:
Changing the hard-coded default to "euduser_setup.html" would probably be a reasonable idea. But really it should be in a define. Leaving hard-coded strings around is not particularly nice.

The feature of external html files wasn't advertised until a few days ago, so maybe now is the time to fix this?

devyte:
How about putting it under a "subdir", e.g.: eus/index.html? if any other files are needed for that webpage, they could be put there as well. As long as the whole filename string is under 32 chars, it's ok.

devyte:
To explain a bit better: SPIFFS doesn't support subdirs, but the filenames _are_ allowed to have a "/" in it. That means that a filename like "eus/index.html" is valid. 
Then, if somebody wants to have a different index.html, e.g.: in a webserver, it can be placed in the "root", or in a different "subdir", e.g.: html/index.html.

marcelstoer:
> The feature of external html files wasn't advertised until a few days ago, so maybe now is the time to fix this?

Indeed, that's why I brought it up now.

> How about putting it under a "subdir", e.g.: eus/index.html?

I thought about this alternative before I committed but opted for a name prefix because...well, to be honest, I didn't find convincing arguments why it should be one or the other.

jfollas:
Using slashes in filenames really messes up the ability to keep those files in source control (or within the directory of a normal filesystem). 

+1 for using some other non-breaking delimiter, like @marcelstoer did with the "-" character.

jmattsson:
I'll join the bikeshedding! +1 for enduser_setup.html :)

marcelstoer:
Changed it to enduser_setup.html, fine for me.

Issue #690, opened 2015-10-16
tmr.time() bug – value leaps backwards


smartynov:
I noticed that `tmr.time()` is producing very strange behaviour: when calling it consequently it sometimes returns smaller values than on previous call. At the same time `tmr.now()` returns increasing values.

Here is the minimal example:

```
tmr.alarm(1,100,1,function ()
  tt = tmr.time()
  tn = tmr.now()
  if not(tt == tp) then
    print("tt="..tt.." tn="..tn)
    tp=tt
  end
end)
```

And here is the output:

```
tt=18 tn=18099727
tt=19 tn=19199832
tt=20 tn=20199841
tt=21 tn=21199687
tt=22 tn=21899493
tt=23 tn=22999399
tt=24 tn=24099722
tt=25 tn=24999726
tt=26 tn=25999477
tt=27 tn=26899470
tt=28 tn=27899454
tt=27 tn=27999667
tt=28 tn=28099774
tt=29 tn=29199720
```

Notice how it goes well until `tt=27` seconds, then it increases to 28 and then falls back to 27. At the same time, `tn` keeps increasing!

Few more examples:
```
tt=47 tn=47099710
tt=48 tn=48099727
tt=49 tn=49199740
tt=50 tn=50199697
tt=51 tn=51299701
tt=52 tn=51599470
tt=51 tn=51799811
tt=52 tn=52199728
tt=53 tn=52599467
tt=54 tn=53599720
tt=53 tn=53799579
tt=54 tn=53899808
tt=55 tn=54599733
tt=54 tn=54899793
tt=55 tn=54999762
tt=56 tn=55699505
tt=55 tn=55899813
tt=56 tn=56099731
tt=57 tn=56799466
tt=56 tn=56899844
```

It is a clean test using `NodeMCU 0.9.6 build 20150704  powered by Lua 5.1.4` (integer version), code is run after a hard reset, nothing else is started in `init.lua`.

smartynov:
Ok, here is an example that's even smaller:
```
tmr.alarm(1,100,1,function ()
  tt = tmr.time()
  tn = tmr.now()
  print("tt="..tt.." tn="..tn)
end)
```
And here we go...

```
tt=98 tn=98661513
tt=98 tn=98761556
tt=98 tn=98861429
tt=99 tn=98961449
tt=99 tn=99061339
tt=99 tn=99161443
tt=100 tn=99261514
tt=99 tn=99361433
tt=99 tn=99461540
[...]

PR #1203, opened 2016-03-30
Make tmr.time() more resilient against RTC changing frequency

jmattsson:
While we normally use the `rtctime` module for good timekeeping, we had a use-case where we needed `tmr.time()` and then got caught out by it drifting wildly (and even going backwards). This is the patch to address that.

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -68,11 +68,18 @@ typedef struct{
 }timer_struct_t;
 typedef timer_struct_t* timer_t;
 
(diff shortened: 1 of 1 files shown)


marcelstoer:
Wouldn't that fix #690?

jmattsson:
Yes, I believe it would.

devsaurus:
Looks good to me :+1:

TerryE:
And me :+1:  Johny you can merge it.

PR #1205, opened 2016-03-30
Provide crypto.rnd32() interface to os_random()

jmattsson:
Since the regular `math.random()` is a PRNG which doesn't get seeded it can be hard go get good random numbers in Lua. Since the ESP has an onboard HWRNG, it seemed useful to expose it.

(I squashed my docs into Bernie's commit, don't blame him if there's something wrong in the docs :D )

### DIFF
--- a/app/modules/crypto.c
+++ b/app/modules/crypto.c
@@ -362,6 +362,24 @@ static int lcrypto_decrypt (lua_State *L)
   return crypto_encdec (L, false);
 }
 
+// returns integer between 0 and 2^32-1, inclusive
+static int lcrypto_rnd32 (lua_State *L)
+{
+  lua_pushnumber (L, (uint32_t)os_random());
+
+  return 1;
+}
+
+#ifndef LUA_NUMBER_INTEGRAL
+// returns double in the range [0,1)
+static int lcrypto_frnd (lua_State *L)
+{
+  lua_pushnumber (L, (double)((uint32_t)os_random())/4294967296.0);
+
+  return 1;
+}
+#endif
+
 // Hash function map
 static const LUA_REG_TYPE crypto_hash_map[] = {
   { LSTRKEY( "update" ),  LFUNCVAL( crypto_hash_update ) },
@@ -384,6 +402,10 @@ static const LUA_REG_TYPE crypto_map[] = {
   { LSTRKEY( "hmac"   ),   LFUNCVAL( crypto_lhmac ) },
   { LSTRKEY( "encrypt" ),  LFUNCVAL( lcrypto_encrypt ) },
   { LSTRKEY( "decrypt" ),  LFUNCVAL( lcrypto_decrypt ) },
+  { LSTRKEY( "rnd32" ),    LFUNCVAL( lcrypto_rnd32 ) },
+#ifndef LUA_NUMBER_INTEGRAL
+  { LSTRKEY( "frnd" ),     LFUNCVAL( lcrypto_frnd ) },
+#endif
   { LNILKEY, LNILVAL }
 };
 
--- a/docs/en/modules/crypto.md
+++ b/docs/en/modules/crypto.md
@@ -90,6 +90,29 @@ A binary string containing the message digest. To obtain the textual version (AS
 print(crypto.toHex(crypto.fhash("sha1","myfile.lua")))
 ```
 
+## crypto.frnd()
+
+Obtain a random number from the hardware random number generator in the
+range \[0,1). Not available in an integer-only build.
+
+#### Syntax
+`random = crypto.frnd()`
+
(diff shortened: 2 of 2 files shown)


app/modules/crypto.c
```diff
@@ -362,6 +362,22 @@ static int lcrypto_decrypt (lua_State *L)
...
 
+// returns integer between 0 and 2^32-1, inclusive
+static int lcrypto_rnd32 (lua_State *L)
+{
+  lua_pushnumber (L, (uint32_t)os_random());
+
+  return 1;
+}
+
+// returns double in the range [0,1)
```
TerryE: Need to \#ifdef this out on an integer build and ditto on the hash map entry.

jmattsson: Ah, good point. I keep forgetting about the integer build! >.>

docs/en/modules/crypto.md
```diff
@@ -191,6 +214,28 @@ The masked message, as a binary string. Use [`crypto.toHex()`](#cryptotohex) to
...
+Provides access to the hardware random number generator on the ESP.
+
+#### Syntax
+`random = crypto.rnd32()`
+
+#### Parameters
+none
+
+#### Return
+A 32bit (unsigned) random value, sourced from the hardware random number generator.
```
pjsg: Is this really unsigned? In the integer build, lua only supports signed 32-bit integers.

pjsg:
Actually, could we have the random number generator take the same arguments as math.random, and just do it in a cryptographically secure way. 

i.e.

crypto.random() returns the float (and wouldn't work in the integer build)
crypto.random(n) returns an integer in the range 1 <= r <= n (uniformly)
crypto.random(l, u) returns an integer in the range l <= r <= n (uniformly)

It is probably worth getting the uniform bit right as well.

jmattsson:
I initially aimed for the same until my colleague convinced me that the Lua way of doing it is deficient due to it's \[1,n\] rather than \[1,n\) range usage, which precludes doing simple things like `local zeroToFour = crypto.frnd() * 5`.

jmattsson:
First time I've had to deal with integer-only build; hope I got the ifdefs right.

pjsg:
The docs are still wrong. They say that rnd32() returns a 32-bit unsigned integer. AFAIK there is no such thing in Lua -- there are only signed integers. 

I would still like to see the three random() flavors implemented:

random()    -- this is just your frnd()  -- but would throw an error in an integer build
random(n)   -- this is just random(1, n)
random(l, n) -- this returns an integer in the range l <= r <= n

Generating the last function is tricky to make it uniform. I think that the code below does what is required.

```
int random(int l, int n) {
    // The range is the number of different values to return
    unsigned int range = n + 1 - l;

    // If this is very large then use simpler code
    if (range >= 0x7fffffff) {
	unsigned int v;

	// This cannot loop more than half the time
	while ((v = os_random()) >= range) {
	}

	// Now v is in the range [0, range)
	return v + l;
    }

    // Easy case, with only one value, we know the result
    if (range == 1) {
	return l;
    }

    // Another easy case -- uniform 32-bit
    if (range == 0) {
	return os_random();
    }

    // Now we have to figure out what a large multiple of range is
    // that just fits into 32 bits.
    // The limit will be less than 1 << 32 by some amount (not much)
    unsigned int limit = ((0x80000000 / ((range + 1) >> 1)) - 1) * range;

    unsigned int v;

    while ((v = os_random()) >= limit) {
    }

    // Now v is uniformly distributed in [0, limit) and limit is a multiple of range

    return (v % range) + l;
}
```

TriAnMan:
Guys, AFAIK the convenient way for cryptographically secure pseudorandom number generators to provide a number of random bytes.
See:
http://man7.org/linux/man-pages/man2/getrandom.2.html
https://msdn.microsoft.com/en-us/library/windows/desktop/aa379942(v=vs.85).aspx
(Links are from http://php.net/manual/en/function.random-bytes.php)

So, according to http://esp8266-re.foogod.com/wiki/Random_Number_Generator `os_random()` and `os_get_random()` looks to be cryptographically secure.

So for my mind it is better to provide an interface to `os_get_random()` method. And it is up to developer to decide what to do with this byte string. The `struct` module provides a lot of ways to transform received byte string into sane values.

Also I may try to do a PR for this.

pjsg:
I would like to have the regular random number generator default to the cryptographically secure one. Yes, this would be a behaviour change (in that math.randomseed would be a no-op). However, I suspect that the vast majority of users do not use the "feature" that math.randomseed can cause your random numbers to repeat...

jmattsson:
Closing this for now since we're all pulling in somewhat different directions, and I'm focused on the ESP32 at this point :D

PR #1206, opened 2016-03-30
enduser_setup: JS select fix

robertfoss:
This PR fixes a problem where the aplist select menu didn't update the SSID field if there only is one entry in the select menu.


jmattsson:
Thanks, looks good!

PR #1211, opened 2016-03-31
Fixed EUS password/ssid decoding.

jmattsson:
I was too easily convinced by @robertfoss earlier! :D 

This fixes both the switched SSID and password lengths (SSID field is 32, password field is 64, not the other way around!), and the off-by-one truncation during url-decoding.

With this it's finally possible to use EUS to join to a network that has a 63-character password.


robertfoss:
Looks good.

PR #1212, opened 2016-04-01
Minimal changes to handle bad bodies in HTTP

pjsg:
This is a simpler version of the vowstar fix to certain crashes due to invalid HTTP responses.

I removed the custom version of strtol that was in the httpclient.c file and we are now using the standard one (that was already being used by the lua code). Also fixed the saving of the lua_State pointer.


TerryE:
I haven't tested this so I won't merge it, but at a source code level these change a good step forward.  Thanks Philip.  :+1:

PR #1214, opened 2016-04-01
EUS JS select fix, part 2

robertfoss:
The fix in #1206 was flawed and made normal interactions with EUS harder on mobile devices (but not desktops).

This PR combines the behaviour before #1206 with the fix in #1206 instead of using only the behaviour introduced in #1206.


jmattsson:
I was curious about that, but it seemed to work fine on my phone (WP8.1).

Issue #788, opened 2015-11-24
Remove the releases from GitHub


marcelstoer:
With my PR #560 the `pre_build` directory was removed (only on `master` and not on `dev` :disappointed: ). I believe we should also drop all binaries in https://github.com/nodemcu/nodemcu-firmware/releases.

Instead we could add a "synthetical" release that contains only a README-style document that explains why there are no pre-built binaries (anymore) and suggest ways how to build the firmware.

Whenever I come across a blog post or the like which explains how to download an NodeMCU release from our GitHub repo I leave a comment or contact the author and tell them why it isn't such a good idea (anymore).

TerryE:
Marcel, I personally believe that a repository should be normalised; that is only contain originating source material and not derived build materials.  Therefore we should _not_ have a `nodemcu-firmware/releases` folder (or a  `nodemcu-firmware/sdk` folder for that matter, but I'll address that separately).  The correct place for releases is somewhere like `nodemcu.com/releases` possibly with a mirror on `nodemcu-build.com/releases` (or `/prebuilt`) and that our documentation should refer to these.

Incidentally I suspect that if you do an analysis of your build requests, they probably cluster quite will and that perhaps 4 or 5 standard prebuilts will cover 90% of requests, so having these prebuilt would mean that the significant majority of new users could simply download a prebuilt rather than doing a custom request.

nodemcu:
First of all, the online build tool is great!  thanks!
At the very beginning, there is a download link for bin at website nodemcu.com.
and also a post for latest prebuilt bin in bbs.nodemcu.com.
but I found github release an easy way to deploy. and a solid place to download.
but as the binary grows, the built bin now is bigger than 512k if everything included.
so, I was thinking that, can we choose a minimal build among all these modules?
say, no chip specific module in the default(minimum) build?
sure I can make some download links at nodemcu.com/releases if it is a good place,
and a sftp upload method for committers.

karrots:
Any reason binary releases just stay in the github releases section like they are now? Travis-ci is setup and could be extended to build 2-3 different firmware options to attach with the release. Travis-ci does support many other [providers](https://docs.travis-ci.com/user/deployment/) if its better to self host on S3 or something.

TerryE:
> Any reason binary releases just stay in the github releases section like they are now?

[...]

PR #1215, opened 2016-04-01
Autobaudrate detection by default -- to reduce problems with speed change

pjsg:
It turns out that doing auto baud rate detection at startup is pretty simple. I added this under a compile time option (but defaulting it to on).

This means that if you bring up the firmware and attach a terminal at the wrong speed, and enter a few characters, it will switch to that speed. It only does this once per boot (and if it hasn't seen any input for two minutes, then it gives up anyway).

If you connect your terminal at 115200, then there is no visible change. However, if you think that the platform runs at 9600, then you can send it a few characters and it will start responding at 9600.

My feeling is that, once we merge the 115200 dev branch to master, then a lot of people will start to complain that the new version doesn't work. Also, it isn't clear how many vendors ship the nodemcu firmware and whether people expect it to be running at 9600. I think that is a plausible approach to keep most people happy.

The only thing that I couldn't decide was whether an explicit set of the baud rate (after the initial setting) would shut off the autobaud detection. Maybe it should.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,9 @@ editing `BIT_RATE_DEFAULT` in `app/include/user_config.h`:
 #define BIT_RATE_DEFAULT BIT_RATE_115200
 ```
 
+Note that, by default, the firmware runs an auto-baudrate detection algorithm so that typing a few characters at boot time will cause
+the firmware to lock onto that baud rate (between 1200 and 230400). 
+
 ### Debugging
 
 To enable runtime debug messages to serial console edit `app/include/user_config.h`
--- a/app/driver/uart.c
+++ b/app/driver/uart.c
@@ -298,6 +298,12 @@ uart_init_autobaud(uint32_t uart_no)
   os_timer_arm(&autobaud_timer, 100, TRUE);
 }
 
+static void 
+uart_stop_autobaud()
+{
+  os_timer_disarm(&autobaud_timer);
+}
+
 /******************************************************************************
  * FunctionName : uart_init
  * Description  : user interface for init uart
@@ -326,6 +332,9 @@ uart_init(UartBautRate uart0_br, UartBautRate uart1_br, os_signal_t sig_input)
 void ICACHE_FLASH_ATTR
 uart_setup(uint8 uart_no)
 {
+#ifdef BIT_RATE_AUTOBAUD
+    uart_stop_autobaud();
+#endif
     ETS_UART_INTR_DISABLE();
     uart_config(uart_no);
     ETS_UART_INTR_ENABLE();
--- a/app/driver/uart.c
+++ b/app/driver/uart.c
@@ -15,6 +15,7 @@
 #include "task/task.h"
 #include "user_config.h"
 #include "user_interface.h"
+#include "osapi.h"
 
 #define UART0   0
 #define UART1   1
@@ -32,6 +33,8 @@ static task_handle_t sig = 0;
 
 // UartDev is defined and initialized in rom code.
 extern UartDevice UartDev;
+
+static os_timer_t autobaud_timer;
 
 LOCAL void ICACHE_RAM_ATTR
 uart0_rx_intr_handler(void *para);
@@ -271,6 +274,30 @@ uart0_rx_intr_handler(void *para)
       task_post_low (sig, false);
 }
 
+static void 
+uart_autobaud_timeout(void *timer_arg)
+{
+  uint32_t uart_no = (uint32_t) timer_arg;
+  uint32_t divisor = uart_baudrate_detect(uart_no, 1);
+  static int called_count = 0;
+
+  // Shut off after two minutes to stop wasting CPU cycles if insufficient input received
+  if (called_count++ > 10 * 60 * 2 || divisor) {
+    os_timer_disarm(&autobaud_timer);
+  }
+
+  if (divisor) {
+    uart_div_modify(uart_no, divisor);
+  }
+}
+
+static void 
+uart_init_autobaud(uint32_t uart_no)
+{
+  os_timer_setfn(&autobaud_timer, uart_autobaud_timeout, (void *) uart_no);
+  os_timer_arm(&autobaud_timer, 100, TRUE);
+}
+
 /******************************************************************************
  * FunctionName : uart_init
  * Description  : user interface for init uart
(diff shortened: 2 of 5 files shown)


TerryE:
I am ambivalent about this. Are there unintended consequences?  If it works solidly the fair enough.

I can think of no advantages for encouraging 9600, and we get loads of issues because beginners putting UART debug into there code which trips the watchdog for a trivial amount of output at 9600.  We should really deprecate the use of low baudrate UART connections, and do this by putting banner advice at the cloud build site and in our release notes.

pjsg:
I don't know how many vendors ship the nodemcu boards preloaded with the firmware. Granted that it probably isn't the latest version, but most of them *claim* to be including the firmware preflashed. It would be nice if this just worked for people. They won't come to our website initially, they will just think that they got something not functional. 

On a related note, is there any advice for vendors shipping the firmware what configuration they should use? In the 4M flash case, I'd like to have them include a reasonably complete set of modules so that most things will work out of the box.

jmattsson:
I could see this being handy considering we're about to switch master to 115200 by default, but in general I'm a bit wary of automagic things. And I'm not sure whether this would encourage transition to 115200.

> The only thing that I couldn't decide was whether an explicit set of the baud rate (after the initial setting) would shut off the autobaud detection.

Yes. Doing otherwise would be a POLA violation imo.

pjsg:
I've added the stop of the autobaud if the lua code sets the baud rate (or changes anything associated with the uart port). I think that @jmattsson is correct.

I think that people will get frustrated having to send the extra characters whenever they start (if they are not on 115200) and this will cause them to figure out the right speed. However, I see this as reducing the barrier to entry.

BTW I think that this is the same code that the resident uploader uses for detecting the upload speed of the flash data. This means that it is pretty reliable.

karrots:
Will there be a way to turn it back on via lua? Not sure of the exact use case at the moment.

On my KPC-3 TNC (ham radio modem) will auto baud the serial connection. It does so by printing the same line over and over again at the different rates asking the user to press *. If autobaud is to cumbersome maybe something like this could be implemented.personally don't know that said text is needed just mentioning an example. HP switches do something similar on their console port as well.

devyte:
@pjsg 
Is it possible to enable/disable the autodetect with a lua function? 
How about configuring the autodetect timeout, including disabling the timeout? (ie: permanently on)
Is it possible to print a message when a baudrate has been auto-detected?

pjsg:
@devyte I had wanted to make it as simple as possible, so I didn't add lots of options. Part of the problem is that the espressif documentation is very lacking in this area and it isn't entirely obvious how the detection function works. In particular, it isn't clear (to me) how to reset it to detect again. 

For disabling the timeout, are you thinking of this being under LUA control or just build time?

My goal was really to fix the case when the user is using 9600 and the firmware is thinking 115200. Once you can interact with the LUA interpreter, then you should know what you are doing!

pjsg:
I see that you can now merge a PR and squash the commits at the same time: https://help.github.com/articles/about-pull-request-merge-squashing/ -- so the issue of multiple commits in the source branch go away!!!

TerryE:
Yup but this is a user account ( "A repository owned by a user account has two permission levels: the repository owner and collaborators.")  I think that in this case this facility is limited to zeroday and not the committers.

devsaurus:
> and not the committers

In fact I used it to merge the SPI fix, and I guess Marcel did the same with my doc update. It's a nice feature but left me with an uneasy feeling as I had to force-delete my local branch because it wasn't merged de jure into upstream.

marcelstoer:
> I guess Marcel did the same with my doc update

Yes, I did. However, I didn't want to comment on this new feature in the context of a particular issue because it's OT. I love the feature but "with great power comes great responsibility." We discussed this a few times before. It doesn't *always* make sense to squash all commits even though I'm stilly happy with this being our general guideline.

marcelstoer:
Back to the real topic...

> in general I'm a bit wary of automagic things

I agree but in the light of the other (sub par) options I believe Philip's code is really helpful.

So, :+1: from my side but I'd like to see this mentioned in `uart.md` as well - in addition to README.

> I don't know how many vendors ship the nodemcu boards preloaded with the firmware.

If I'm not mistaken most do. However, they all ship an old binary release from this repository (see #788).

pjsg:
I've update the `uart.md` documentation.

pjsg:
@marcelstoer The esptool.py flasher doesn't interact with the auto-baud stuff in the firmware as it resets the device in order to get it into the ROM. The ROM will then do the auto-baud using the same code as the PR does.

marcelstoer:
Philip, it's not the flashing tools I had in mind (for the reason you stated) but the uploaders. I know ESPlorer doesn't try to detect baud rate but I don't know about others.

pjsg:
Ah -- Ok. I did a quick search to see if any of the uploaders did auto baudrate detection, and I didn't find one that did. If anybody knows of one, then please let me know so that I can test it.

One factor in this is that once a few characters have been received, and the baud rate detected, then the automatic mechanism shuts downs. The danger is if the uploader code is being run just after restarting the nodemcu, and only sends a couple of characters at a speed before moving onto the next speed. But I don't know of any software that does this.

joysfera:
@pjsg regarding the uploaders - since the change to 115200 I can't upload any file neither by luatool nor by nodemcu-uploader. If your autobaud patch can fix it then I can't wait for merging it. It's worth noting that currently it's behaving like if an autobaud was already in place (and was breaking the upload). Please let me know if I should open an issue for that.

karrots:
nodemcu-uploader assumes 9600 baud start point. The --baud flag only changes the baud after initial setup. Using the new `--start_baud 115200` flag in nodemcu-uploader will fix your issue. It has for me anyway.

joysfera:
It didn't help here. When in terminal mode I have to change the baud rate manually to 9600 and then back to 115200 - only then it starts working.

devsaurus:
@joysfera slightly off topic, but I had the same symptoms with current nodemcu-uploader. Was fixed by upgrading pyserial from 2.7 to 3.0.1. No issues anymore with default baud rate 115200.

marcelstoer:
Terry, Arnim, Johny, we seem to more or less agree that this should be merged, right? If so, we need to do that before the upcoming `dev` to `master` merge. Any further objections?

devsaurus:
:+1: from my side.

TerryE:
👍

joysfera:
FYI, while the autobaud change seems to work in a terminal application at first, neither luatool.py nor nodemcu-uploader.py can upload a file here. Does it work for @pjsg (with pyserial 2.6)?

kmpm:
nodemcu-uploader should support 115200 (and auto-baudrate) since v0.4.0 that I just published.
If you have any issues with nodemcu-uploader please create issues there ( https://github.com/kmpm/nodemcu-uploader/issues )

PR #1219, opened 2016-04-03
Add notes on user_override.h

pjsg:
This small PR does two things:

* Adds a note describing how to use the user_override.h that can be placed in the root of the build tree

* Adds support for being able to add fonts to the build by editing the user_override.h file.

If you wonder where the user_override.h stuff came from, then it appears that it got committed/merged as part of another PR that I submitted some time ago. It was the mechanism that I use to maintain my set of modules as I switch between branches. [It was originally called user_config.h, but I renamed it in this PR to over_override.h]

### DIFF
--- a/docs/en/build.md
+++ b/docs/en/build.md
@@ -15,10 +15,10 @@ NodeMCU firmware developers commit or contribute to the project on GitHub and mi
 
 ## Customizing the build
 
-There is an optional file `user_config.h` that can be placed at the root of the build tree. If this file is present then it
+There is an optional file `user_override.h` that can be placed at the root of the build tree. If this file is present then it
 will be included into the compile of every file. This allows you to customize the build without altering any of the
 files that are part of the firmware. Note that this file is included *in addtion to* all the existing files so it 
-will typically be short. This file does not replace the `app/include/user_config.h`.
+will typically be short. 
 
 In order to include extra fonts into the build, just add a #define:
 
--- a/docs/en/build.md
+++ b/docs/en/build.md
@@ -12,3 +12,35 @@ Occasional NodeMCU firmware hackers don't need full control over the complete to
 
 ## Linux Build Environment
 NodeMCU firmware developers commit or contribute to the project on GitHub and might want to build their own full fledged build environment with the complete tool chain. There is a [post in the esp8266.com Wiki](http://www.esp8266.com/wiki/doku.php?id=toolchain#how_to_setup_a_vm_to_host_your_toolchain) that describes this.
+
+## Customizing the build
+
+There is an optional file `user_config.h` that can be placed at the root of the build tree. If this file is present then it
+will be included into the compile of every file. This allows you to customize the build without altering any of the
+files that are part of the firmware.
+
+In order to include extra fonts into the build, just add a #define:
+
+```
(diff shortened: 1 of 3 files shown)


docs/en/build.md
```diff
@@ -11,4 +11,36 @@ NodeMCU "application developers" just need a ready-made firmware. There's a [clo
 Occasional NodeMCU firmware hackers don't need full control over the complete tool chain. They might not want to setup a Linux VM with the build environment. Docker to the rescue. Give [Docker NodeMCU build](https://hub.docker.com/r/marcelstoer/nodemcu-build/) a try.
 
 ## Linux Build Environment
-NodeMCU firmware developers commit or contribute to the project on GitHub and might want to build their own full fledged build environment with the complete tool chain. There is a [post in the esp8266.com Wiki](http://www.esp8266.com/wiki/doku.php?id=toolchain#how_to_setup_a_vm_to_host_your_toolchain) that describes this.
\ No newline at end of file
+NodeMCU firmware developers commit or contribute to the project on GitHub and might want to build their own full fledged build environment with the complete tool chain. There is a [post in the esp8266.com Wiki](http://www.esp8266.com/wiki/doku.php?id=toolchain#how_to_setup_a_vm_to_host_your_toolchain) that describes this.
+
+## Customizing the build
+
+There is an optional file `user_config.h` that can be placed at the root of the build tree. If this file is present then it
```
TerryE: Perhaps you should reference the template version in `app/include` that can be used as the basis for this. 

I had never even thought to use the include search list to preempt the standard one,bBut then again I just run my Lua config script to write mine. 

The same approach can be `user_modules.h`, can't it?

pjsg: This file doesn't actually replace app/include/user_config.h -- it is used in addition to it. Now that I think about it, it is not a good name as it will cause that confusion. You can add modules here as well. I added a more complete example.

Maybe I should change the name to user_override.h -- what do you think?

TerryE:
@pjsp Philip have user_modules.h and user_config.h is bad enough, IMO.  They are cludgy and I don't like them.  That's why I had a unified config file format which is a bit like a MySQL user config file, and Lua script which generates these files.  

Have the ability to place either in the root to override the defaults in the app/include directory is a great idea.  Having extra files with the same name but different function is a definite no-no.

My vote would be as this was accidentally added on the QT, it should be backed out.

pjsg:
I changed the name to user_override.h

I'd be happy to move to a better unified config file format (which remains outside the tree) and back the original change out.

marcelstoer:
> My vote would be as this was accidentally added on the QT, it should be backed out.

:+1:

TerryE:
Philip, you've added another core configuration file.  This is a fundamental configuration change.  This should have it's own issue raised and properly discussed by the contributors, before we decide on this.  "It appears that it got committed/merged as part of another PR that I submitted some time ago" isn't a valid reason for bypassing this debate.  

I prefer the unified config approach.  I have scripted this and proposed that we adopt this. I did this by raising the issue.   I offered to port this to any scripting language that the other contributors felt necessary.  However, none of the other contributors supported this, so I accepted this view and withdrew the script.

Raise the issue and lets discuss this properly.

Issue #1220, opened 2016-04-03
Change enduser_setup connected callback to fire when enduser_setup is done


pjsg:
I want to be able to write code like:

init.lua:

```
enduser_setup.start(function() dofile("init2.lua") end)
```

init2.lua:

```
-- start the http server
-- start the timers
-- etc, etc
```

However, the connected callback currently fires while the enduser_setup still has the http server active so I can't start another one. Adding a timer for 10 seconds seems the wrong approach as well.

Having to wrap all my initialization code in a pcall and then retrying it every second also seems the wrong approach, -- and it would require very careful coding to make sure that everything only gets initialized once.

Please have a callback that indicates when the wifi is working and the enduser_setup module has cleaned up after itself.

jfollas:
As a workaround, you should be able to call `enduser_setup.stop()` within the callback to tear down EUS before the 10-seconds elapse.

jmattsson:
And if that's causing you crashes, use manual mode (you'll need to get the ESP into STATIONAP mode before starting EUS in that case).

It does sound like the callback is firing too early as is though.

Issue #1180, opened 2016-03-21
EUS - HTTP stops responding while station is connecting


jfollas:
While @robertfoss is working on #1138, I thought I'd mention this issue that has been plaguing us for the past few weeks in case it leads to something that can be corrected in that PR:

I have a private fork of NodeMCU that also has a fork of the End User Setup module (I kept the original in the source so that it would keep in sync with the upstream repo, and just copied the code into a new module). We use the modified version of EUS on our device because there's a few additional endpoints that are implemented, and a couple of other changes unique to our needs. 

One of the endpoints is similar to the /status endpoint in that a client sends a request on a regular basis (2 second interval) and gets a JSON response back with information. With debugging turned on, we see this 2-second request working regularly right until the time that credentials are provided.

The code executes `wifi_station_set_config()` just fine (inside `do_station_cfg()`), and the config gets saved to the flash. But, as soon as the subsequent `wifi_station_connect()` executes, no more HTTP requests are accepted by the ESP until at least the wifi status != 1 (connecting). It doesn't matter what endpoint the browser tries, the ESP just quits answering TCP requests while the station is connecting and the web client (XHR) times out.

This wouldn't be a big deal to just wait out the Connecting status until there's either a success or failure, since it's only 5-10 seconds. But more often than not, it doesn't resume processing HTTP requests even after the station connects (even using manual mode to control the AP and stop()). So the web page never gets the final status that was requested.

This symptom of requests not being served sounds familiar: does anyone remember similar behavior, and what may have led up to it?

jfollas:
I took Robert's latest enduser_setup.c and reapplied all of my additions to it. The scenario is still there where requests are not serviced while the station is connecting (so that's probably normal), and so far, I'm getting requests served after the connection is established. 

I'll close this issue after testing is complete using all of the phones and tablets on my desk. :-/

jmattsson:
I can explain this behaviour. It's not that the http server stops responding as such, it's that the entire AP gets channel-shifted to whichever network's channel the station side is trying to join. That's largely a hardware limitation of trying to run both a station and an AP on the one radio. Things get even wackier if you try to join a non-existent network, whereby the ESP keeps switching channel searching for the network... Normally your client (phone, laptop etc) catches on to the change of channel after ~10sec or so (in my experience), but if the ESP is channel-hopping it has a *much* harder time of keeping up and staying in touch with the http server.

[...]

PR #1221, opened 2016-04-03
EUS: Wi-Fi Channel issue workaround, bug fixes, improvements

jfollas:
Squashed commits included:

Bug fixes and final implementation
- Added Content-Length: 0 to all headers
- Endpoint name checks not using trailing space so cache-busting techniques can be used (i.e., append a nonce to the URL)
- Track when connecting so APList scan doesn't take place during (which changes the channel)
- More debugging output added to assist in tracking down some issues

Added /status.json endpoint for phone apps/XHR to get JSON response

Station Status caching for wifi channel workaround + AJAX/CORS
- During checkstation poll, cache the last station status
- Shut down the station if status = 2,3,4 and channel is different than SoftAP
- Add Access-Control-Allow-Origin: \* to endpoint responses used by a service
- Add a /setwifi GET endpoint for phone apps/XHR to use (same parameters as /update endpoint). Returns a JSON response containing chip id and status code.
- Add handler for OPTIONS verb (needed for CORS support)

Wi-Fi Channel Issue Workaround
- Do a site survey upon startup, set SoftAP channel to the strongest rssi's channel
- Compare successful station connect channel to SoftAP's. If different, then defer the Lua success callback to the end. Shut down Station and start the SoftAP back up with original channel.
- After the 10 second shutdown timer fires, check to see if success callback was already called. If not, then call it while starting the Station back up.

HTTP Response and DNS enhancements
- If DNS's UDP buffer fills up, keep going as non-fatal. It's UDP and not guaranteed anyways. I've seen this occur when connecting a PC to the SoftAP and every open program tries to phone home at the same time, overwhelming the EUS DNS server.
- Support for detecting/handling pre-gzipped `enduser_setup.html` (and `http_html_backup`) payload. Nice for keeping the size of the `state->http_payload_data` as small as possible (also makes minimization not as critical)
- Corrected misuse of HTTP 401 response status (changed one occurrence to 400/Bad Request, and changed another to 405/Method Not Allowed)


jfollas:
This is the PR to address #1180. 

While testing with latest commits, discovered situations where the HTTP responses were malformed (content-length was off-by-one, content length missing from 404, etc) and caused issues with Fiddler while I was verifying correct operation (Fiddler was waiting for the remaining bytes that were promised to arrive, but then the connection was force closed... though I noticed that actual web browsers were more forgiving). This has been resolved in this PR and verified to work.

Also brought over some enhancements from our private fork implementation, like not treating a full DNS UDP buffer as a fatal error. When I connect my laptop to the EUS AP, every open browser tab and every open app tries to call home once the new network is up. This overwhelms the UDP receive buffers and used to lead to a Panic Reset. My thinking here is that UDP is not guaranteed, so it's safe to discard the overflows (anticipating that the request will be made again by the client).

And lastly, because configuring EUS via a mobile app is important to a lot of folks, I brought over our "web service" endpoints:

|Endpoint|Description|
|----|----|
|/setwifi|Identical to /update endpoint, but returns JSON instead of redirect to HTML
|/status.json|Endpoint to poll for status changes as JSON. Includes chipid and status code.

A pretty generic CORS implementation is also included for the benefit of things like Cordova that sends the `OPTIONS` verb and respects `Access-Control-Allow-Origin` response headers.

(This reminds me that I need to make documentation changes and add into this PR)

app/modules/enduser_setup.c
```diff
@@ -490,14 +555,26 @@ static int enduser_setup_http_load_payload(void)
...
 
   if (f == 0 || err == -1 || err2 == -1)
   {
     ENDUSER_SETUP_DEBUG("enduser_setup_http_load_payload unable to load file enduser_setup.html, loading backup HTML.");
 
-    int payload_len = LITLEN(http_header_200) + cl_len + LITLEN(http_html_backup);
+    c_sprintf(cl_hdr, http_header_content_len_fmt, c_strlen(http_html_backup));
+    cl_len = c_strlen(cl_hdr);
+    
+    if (http_html_backup[0] == 0x1f && http_html_backup[1] == 0x8b)
```
pjsg: If the content is gzipped, then it is likely that the strlen call a couple of lines up will give the wrong answer as there 
will be an enbedded null in the data.

jfollas: You're absolutely right. This was hand-merged, and I see in my fork that I used `sizeof` instead of `strlen`. Thanks for catching that!

robertfoss:
What is the benefit of a CORS implementation? Isn't it only useful for HTTPS scenarios?

Are these new json endpoints used by the html, or just designed for apps?
If at all possible, I would like the same API be used by the html and apps.

jfollas:
CORS isn't just HTTPS. It's pseudo-security built into browsers (i.e., XHR) that will prevent JavaScript from fetching resources from a domain that is different than the current web page's unless the `access-control-allow-origin` header lists the current page's domain (or wildcard). This is why interacting with web services on another domain often fails if they haven't properly set up CORS (or if they don't want their services to be accessible to the public). Note that `content-type` plays a role in CORS, so sometimes a request for a `text/plaintext` resource might work just fine, but a request for `application/json` data will trigger the CORS check.

In addition, a client will test CORS for each resource before making a regular request by sending a pre-flight OPTIONS request. The server responds with things like the list of verbs that can be used (GET, POST, etc) and the `access-control-allow-origin`, etc. This way, the client is not even supposed to attempt making a real GET request if CORS is just going to get in the way.  

In summary, CORS is not going to be needed in the typical EUS scenario, but without it being implemented, a whole class of apps may be locked out of being able to automate the wi-fi configuration through the EUS capabilities.

In our implementation, I was troubleshooting why a Cordova app had so much trouble fetching status data from the EUS "web server", so I printed out each request's contents to the DEBUG. That's when I saw OPTIONS requests coming through (which were being discarded before), and figured out that it was CORS that was interfering with that Cordova app.

We also originally tried to just use the `/update` and `/status` endpoints with apps (i.e., XHR), but had trouble because it was a 302 and HTML returned for `/update` and we needed more information from `/status` than just the plain-text status description (even more data than what I added to this PR). That's the reason for the two new endpoints intended for programmatic access by apps or anything else that isn't meant to consume HTML. 

Incidentally, the `enduser_setup.html` that we use has been converted to use these endpoints for status polling and form submission - I changed everything over to XHR in the JavaScript. If this PR is accepted, I'll submit that page (probably as an `/example`) but it doesn't make sense to submit it now without the API in place. The backup HTML stored in my firmware itself is a very lightweight pure HTML page that does a simple form submission using `/update` - just enough to allow the user to submit the wi-fi credentials should the regular `enduser_setup.html` not exist on the filesystem, but doesn't bloat the firmware size as much as the pretty page does. So, because of that, I don't think it makes sense to remove any of the existing endpoints in favor of using the two new endpoints.

robertfoss:
The CORS support seems a little bit over the top to me. But I'm certainly not a web-guy, so maybe implementing it actually is fine. I'm ok with it being included if it is necessary.

About supporting multiple types of endpoints: Let's just move the html in the firmware to use the JSON endpoints. Having multiple implementations of the same functionality is no good. And JSON is the correct way of doing this.

marcelstoer:
Looks pretty complete to me, well done.

Jason, you're using this from Cordova? Interesting... Why would Cordova itself send the pre-flight requests if it's just a native wrapper around a web application?

I was impressed you answered the CORS question in your own words 😉I always send folks to https://en.wikipedia.org/wiki/Cross-origin_resource_sharing when asked about it.

> configuring EUS via a mobile app is important to a lot of folks

Do you have any numbers to back up this claim?

> Let's just move the html in the firmware to use the JSON endpoints. Having multiple implementations of the same functionality is no good.

I agree.

jfollas:
> Jason, you're using this from Cordova?

Cordova and Xamarin are a couple of the mobile app technologies used to call our `/setwifi` endpoint once a mobile device has connected to the EUS SoftAP. That is, we don't just show the EUS web page in a webview, but rather our app shows its own form to the user, and then just does an XHR request to submit data to EUS (and to poll `status` and `aplist` messages, too). And, if a customer doesn't happen to have our app, then they can just open a web browser to do EUS configuration as it exists today.

At this point, I don't recall specifics, and the CORS discovery was made at another location (so I don't have the apps to try right now). But I think Cordova sends an `Origin: null` header with its HTTP requests for external resources. Since `Origin` is included, a CORS-configured website is supposed to include `Access-Control-Allow-Origin` in the response. I also specifically recall seeing pre-flight requests using `OPTIONS` when I had debugging enabled.

> Do you have any numbers to back up this claim?

Me. Plus all of my customers [who will be] using our firmware and mobile app. So that's thousands. ;-)  Likely others, too (i.e., @skycoders mentioned the same in #1138), but it's kind of the norm to configure IoT devices using a phone, and as nice as the EUS captive portal is, it's still slower than having an app just do it for you (especially after the second or third device that you are configuring).

This is the approach that we're using, and would like to contribute the code back to the project in the event that others find it useful, too.

> Let's just move the html in the firmware to use the JSON endpoints. Having multiple implementations of the same functionality is no good.

I'm not sure that it's a zero-sum game. I see utility in having both the existing endpoints, and the JSON API endpoints.

robertfoss:
What are the advantages of having duplicated json/non-json enpoints?
The disadvantages are maintenance and size of code as far as I see it.

Parsing json instead of just inputting raw data is pretty much equally hard in html.

jfollas:
Here's a repository that I put together to showcase the effects of gzipping the content: https://github.com/jfollas/nodemcu-eus-html

It includes the modified `enduser_setup.html` that I talked about earlier (under "squares" directory). Some user experience changes, and it uses the API endpoints.

It also includes a very "minimal" example that, in my mind, should be what's included in the source as `http_html_backup` because after being gzipped, only occupies 361 bytes. Being a backup, it provides the bare minimal functionality needed (a way to set the wifi credentials from a web page) for about 10% of the memory requirements needed for the nicer version. The user would then be in control over which experience they need, and how much memory they can afford to give up in order to run EUS on their device (i.e., if they have the memory to spare and want the nicer experience, then it requires a file to be added to SPIFFS).

robertfoss:
Gzipping content makes sense to me, especially since the way we store HTML in enduser_setup.c already is minified and pretty much unreadable. And also since HTTP supports gzipped content.

Would it be possible for EUS to serve some content gzipped (like the HTML) and some content unzipped (like the JSON (since we don't really want to do live gzipping on the ESP))?
If serving mixed content isn't possible, I don't think there is much of a point in keeping the static HTML gzipped, especially since we would have to unzip it on the ESP before sending it out.

@jfollas are you suggesting that we replace the current HTML shipped EUS with something more minimal? I don't particularly agree with that as it would increase the burden for the average developer using NodeMCU. Requiring a developer to write a HTML module to have a semi-reasonable experience with EUS does not sound terribly appealing for the average developer. 
However if you really care about those bytes taking up space, building a version without the backup HTML is probably the way to go.  Maybe adding a compile-time switch for disabling the backup HTML is a reasonable compromise.

marcelstoer:
> Would it be possible for EUS to serve some content gzipped (like the HTML) and some content unzipped (like the JSON)

Yeah, definitely.

> are you suggesting that we replace the current HTML shipped EUS with something more minimal?

I wouldn't do that. Compressing the current HTML is good enough IMO (i.e. most suitable for the majority of users).

> Maybe adding a compile-time switch for disabling the backup HTML is a reasonable compromise.

👍 but only if it's off by default.

jfollas:
> Would it be possible for EUS to serve some content gzipped (like the HTML) and some content unzipped (like the JSON)

This PR already does that. It determines whether the HTML is gzipped or not, and emits the header. Doesn't try to gzip any of the JSON.

robertfoss:
But if mixed gzipped and unzipped content isn't a problem, then that is
fine.

jfollas:
@robertfoss It's not a problem because compression is for the transport layer only, and each response is independent from other responses.

So, gzipped content (determined by looking at the first two bytes of the data to see if it's the magic bytes for gzip) will get a `Content-Encoding: gzip` header added, while other non-compressed responses won't. Modern browsers automatically accept both encodings, and if it's gzipped during transport, then it will be unzipped before passing it on to the next stage of processing.

robertfoss:
@jfollas The gzip stuff sounds good to me!

Changes I'd like to see:
- (optional) #define for disabling backup HTML.
- Remove duplicate HTTP endpoints.
- Switch HTML over to JSON.

jfollas:
@robertfoss To be clear, which endpoint(s) do you see as duplicate? Status is the only one that comes to mind. 

`/update` and `/setwifi` may have overlap in functionality, but serve two distinct purposes (`/update` enables the minimal web form that doesn't need javascript, and `/setwifi` enables API access to the same functionality).

robertfoss:
I don't see building a JS-less HTML page being particularly useful. Not useful enough to warrant keeping the non-JSON endpoints around.

So to be clear, all non-JSON endpoints that now have JSON equivalents should be removed.
(I've updated the above post to clarify this)

gsker:
If you don't provide a JS-less HTML page, then how do you program the device 
from a non-js browser?  Is there a GET that will do it or does it require a 
POST?
The EUS needs to have the lowest bar possible.

Gerry


--
gsker@skerbitz.org

robertfoss:
@gsker: Which browser does not support JS? Lynx? Curl? If you're worried about people who purposefully block JS in their browser, I'm sure other sites have broken mysteriously for them before. 

You could always just send a GET containing some JSON without ever using a browser.

gsker:
Yes;  wget, lynx, links, curl, telnet, perl, python etc.

I just think when working with simple devices that we should make sure
the standard (an HTML GET with a FORM) is supported for the lowest possible 
client.  And that it's clear what is expected when the 
default page is connected to.

I think javascript only is a bad idea.

--
gsker@skerbitz.org

marcelstoer:
> If you don't provide a JS-less HTML page, then how do you program the device
from a non-js browser?

By sending JSON to the defined endpoints. All of the listed _user agents_, curl et.al. are not browsers, are capable of doing that. I can't image a _realistic_ use case where EUS with JSON-only endpoints would indeed pose a real problem.

robertfoss:
I would agree with marcelstoer. If you can send a GET request, then you can send a GET request containing JSON.

jmattsson:
I haven't had a chance to test out these changes yet, but it sounds good! I see you too have been hit by the fatal DNS issue - it's been on my todo-list to submit that upstream for some now, but it looks like you've done it for me :)

And to hop on the "me too" train - sending a GET with JSON is typically simpler than GET with urlencoded params.

Do you expect these changes to fix the issue seen in #1220?

jmattsson:
Alright, test results (unfortunately not so good):

-  Something here is breaking the `manual()` mode. Regardless of whether EUS is configured for manual, the EUS here now changes the AP config including SSID.
- On loading the page (WP8.1) I get the EUS page show up, then shortly (~3s) thereafter it gets replaced by an error screen saying "Error: Cannot find server or DNS error", giving me no chance to actually interact with EUS. Attempting to reload the page yields the same error, sometimes after very briefly flashing up the actual page.

jfollas:
@jmattsson for the second point, was there already wi-fi config saved in the ESP's flash? Wondering if this is the same thing that we saw where the ESP is configured to connect to an AP that is not accessible, but that doesn't stop the ESP from channel hopping to try to find that AP.  We had to take the approach to execute `wifi.sta.config("", "")` before starting EUS in order to keep the SoftAP up on one channel.

jmattsson:
Yes, I can confirm that during that test the ESP was already *successfully* joined to a network, so the AP was already started up on the same channel as the STA.

robertfoss:
@jfollas Are you still up for rebasing these patches on dev?

I'd rather see this PR merged (and imperfect) than not merged at all.

marcelstoer:
> I'd rather see this PR merged (and imperfect) than not merged at all.

Same here. I merged the changes into my own fork's `dev` (with the intention to create a new PR).  However, either the changes don't play nice with the current base or I broke something while merging. EUS doesn't work reliably after the merge.

https://github.com/marcelstoer/nodemcu-firmware/commit/9da2fd15e6c1ce0a11e38cc2ff2aabbdd5e105e7

jfollas:
Just saw these replies in my inbox. Apologies for the inaction.

For a while there, I wasn't able to create good builds for some reason
(after the SDK version bump), but that seems to be resolved now.  Let me
see what I can do to get this PR working again.

jfollas:
This PR is working again with latest code changes. 

I replaced the previous embedded HTML/Javascript with a revised version that uses the JSON endpoints and does not constantly invoke the AP survey (which was causing the ESP's radio to skip all over the place). Because I grabbed this from my private source, the styling is a bit different than Robert's original page.

One other thing that I did was externalize http_html_backup into a *.def file that is included into enduser_setup.c. I did this for a few reasons:
- Kept the HTML out of the source code - smaller file to maintain/scroll through
- Allows the full HTML to be maintained separately under source control
- Using the included prepare.sh script, will gzip the HTML and produce the array declaration c source code (which is what's actually included as the .def file). 

One thing that we discovered in our project was that simply gzipping the HTML yielded a much smaller output than minifying the HTML/JavaScript (especially when using Zopfli). There was only a marginal size improvement by minifying the HTML before gzipping, so we don't even do that step (i.e., we maintain "pretty" HTML/JavaScript and then gzip that directly for embedding into the firmware). That's what I replicated here.

Hoping this is good enough now to finally be merged in, and then it can be further revised under separate PRs.

marcelstoer:
Thanks Jason, much appreciated. However, this doesn't look right. GitHub shows 249 modified files with +48,526 −2,927  line changes. What happened?

jfollas:
Ugh. Not sure. Let me see if I can figure out the proper git magic to
correct it.

marcelstoer:
With Git nothing is ever "beyond repair" but I often the lack the experience to repair it 😦 So, starting over in a new branch is always an option.

jfollas:
When I compare my branch to the dev branch, I see far fewer changes... so will try to just do a new PR as an easy way to get this working.

jfollas:
Resubmitted as #1583 which shows the correct number of commits/files

Issue #1164, opened 2016-03-17
Upgrade to current version of SPIFFS


pjsg:
We have various issues with spiffs at the moment. Large files, crashes etc. Our version has diverged from the original github repo. I propose to bring our version up to date *and* make it easy to pull new versions in the future.

Also, after discussions with the SPIFFS author, I'm going to change the parameters for the default spiffs configuration. This will be somewhat more efficient than the current settings. 

In order to make spiffsimg work nicely, I want to bring that into the tree and build it from the same source that we use -- in this way, we are guaranteed to build compatible file systems. I also want to move to using the magic header on the file system so that there is less danger of trashing a file system if it doesn't match the size of the remaining flash. 

This will certainly take some time to settle down, but it shouldn't be *too* painful.

Thoughts?

TerryE:
In principle this sounds like a good idea, but we should do a careful review of our historic changes just in case any are valid for regressing into the new SPIFFS version.

Also my personal preference would be to move the SPIFFS test suite out of the core app tree into say `tests/spiffs/` and exclude it from the standard make altogether as this suite won't even run on the ESP hardware without some further work.

pjsg:
I was also proposing to get rid of the test suite as it didn't work anyway (and it you could make it build, it would not run on the ESP and would have different parameters).

I agree on looking at the historic changes. The good news is that the SPIFFS maintainer uses spiffs on an ESP!

pjsg:
From looking at the historic changes, the only one that still seems relevant is the one which fixes a bunch of issues with long filenames. I filed a ticket pellepi/spiffs#74 to get appropriate changes made. 

All the other changes were changes to the data structures to do with alignment. I want to leave all of those out and just use the base code (and make sure that any fixes we need are done through the base project).

pjsg:
This will resolve #896 #1147 #1149 #1150

jmattsson:
I believe the only change we'd want to be keeping is the tuning of the working buffers in order to keep RAM usage in control. The only alignment changes we're currently running with (to my knowledge) are in relation to willy-nilly casting into the internal working buffer, which hopefully would be fixed properly upstream. Most of our issues with alignment-related crashes were tracked down to be an inconsistent behaviour of the platform flash read/write routines which were fixed, and removed the need for all(?) the previous alignment hacks applied to SPIFFS.

If there are layout changes to the actual file system, please ping me when this goes in so I can update `spiffsimg` as well.

[...]

Issue #1150, opened 2016-03-11
SPIFFS bugfix for duplicate filenames


eku:
I suggest to take over the bug fix for https://github.com/pellepl/spiffs/issues/74

pjsg:
This will be included in an upcoming PR.

marcelstoer:
Fixed by #1226.

Issue #1149, opened 2016-03-11
Spiffs cannot hold large file (larger than around 26857B)


MarsTechHAN:
Happened on both master branch and dev branch.
MASTER Commit: c8037568571edb5c568c2f8231e4f8ce0683b883

 If I use old firmware(such as 0.9.6 release) it can write all but return wrong length.

I want to write the GBK.bin(A dot matrix font lib for u8g to display Chinese word) into the Flash through SPIFFS. 

I try the ESPlorer's upload, but the nodemcu will stop response after around 26857B ( at a static point).
`Uploading to ESP file gbk.bin...Waiting answer from ESP - Timeout reached. Command aborted.`

Then I used mkspiffs, but when I use esptool to write the output bin inside, the nodemcu cannot detect the file, the filename is not ascii charater and size is -678239427394(a large minus number) 
but I can open it and read/write. （Have not check weather the data of the file is correct or not yet)

`Uploading to ESP file gbk.bin.001...Success
Uploading to ESP file gbk.bin.002...Success
Uploading to ESP file gbk.bin.003...Success
Uploading to ESP file gbk.bin.004...Success
Uploading to ESP file gbk.bin.005...Success
Uploading to ESP file gbk.bin.006...Success
Uploading to ESP file gbk.bin.007...Success
Uploading to ESP file gbk.bin.008...Success
Uploading to ESP file gbk.bin.009...Success
Uploading to ESP file gbk.bin.010...Success
Uploading to ESP file gbk.bin.011...Success
Uploading to ESP file gbk.bin.012...Success
Uploading to ESP file gbk.bin.013...Success
`
Now I split it into 42 files, each 24KB, and looks good :X but it is really inconvience, wish to solve the problem~ :D

devsaurus:
This seems to be related to #898. Though there's no fix at the moment.

MarsTechHAN:
Thanks for the reminding,  In addition, with the circumstance 2, I can open the file, read it but it does not show correct on the file.list() with looks like overflood number, which is really weird?

pjsg:
I am looking to replace the current version of spiffs with the current master version from the source repo. This can handle large files....

marcelstoer:
Fixed by #1226.

Issue #1147, opened 2016-03-10
ESP8266 flash write cycle limit


GelirDurum:
I have a problem with ESP8266 12E and 12F series about write commands. When I need to write data to the flash in ESP8266 it only allows to around 6800 times with (w+) command. I am using both integer and float and lastest and previous firmwares but nothing changes. An example simple code is below. With ESPlorer I run the codes and when both test1 lua and test2 lua reaches to 3400 file system starts to reset the module. I only activate module by formatting it. If I use just one file it reaches around 6800 cycle of writing. With (a+) command and 2 files I only managed to write up to total 13600 number somehow. Then again a format is needed. I do not how to erase and where to erase flash form lua file. Any help would be appreciated.

```
test=0
tmr.alarm(6, 100, 1, function() 
             file.open("test1.lua","w+")       
              file.writeline(test)
            file.close()
             file.open("test2.lua","w+")       
              file.writeline(test)      
            file.close()

            test=test+1
end)
```

Ondreyk:
I have this problem too. And in native C (not only in lua module).

pjsg:
On lua, does the module reset? If so, then can you provide the output of node.bootreason() and the contents of the mapfile that was output by the build.....

Ondreyk:
I have other observation(suspicion). With SPIFFS_DGB. I saw next picture.

```
page index byte len:         256
object lookup pages:         1
page pages per block:        16
page header length:          5
object header index entries: 107
object index entries:        125
available file descriptors:  4
free blocks:                 135
```


All static but except "free blocks". 
```

free blocks:                 133
free blocks:                 131
free blocks:                 129
...

free blocks:                 5
free blocks:                 3
Reset reason: REASON_EXCEPTION_RST
free blocks:                 3
Reset reason: REASON_EXCEPTION_RST
...
```

"free blocks" have leaks. It is happened if we run timer once and reset device. Sorry for my bad grammar.

pjsg:
I'm trying to debug this issue -- I suspect that it is another alignment error. It turns out that the current version of spiffs from the master repo has a (difference) alignment error in it. Unfortunately, I'm finding it difficult to build the nodemcu flavor of spiffs with the testcases. I'm trying to build it on my linux system. The other option would be to move to the newer version of spiffs and move it over, being careful not to edit most of the files (to make it easy to bring new patches over in the future).

Thoughts?

devyte:
[...]

PR #1226, opened 2016-04-09
Upgrade to latest spiffs and add building of file systems

pjsg:
This PR resolves a bunch of open tickets:

#1164 -- General upgrade to latest version
#1150 -- Fixed in the latest version of spiffs
#1149 -- Fixed in the latest version of spiffs
#1147 -- Fixed in the latest version of spiffs
#898 -- Fixed in the latest version of spiffs
#1064 -- Code merged in.

The configuration of spiffs was changed slightly after consultation with @pellepl . 

The filesystem is now formatted with the magic_number option. This means that, when the system boots, it can find the filesystem. This also means it formats the file system at boot if no valid filesystem is found. This should fix a bunch of problems with corrupt file systems causing mysterious problems. On the other hand, it also means that flashing images tends not to corrupt the file system (a bigger gap is left).

This also adds a mechanism to build a filesystem image which can be flashed. 

There are some integration kinks to work out with regard to the file system image and how to flash it. It turns out that it takes quite a while to flash a 4MB image!

The spiffs code is now dropped into the lua tree essentially unmodified -- this makes it easier to upgrade in the future. I did remove the test cases though.

We don't want to merge this until after the dev->master merge takes place....


devsaurus:
Very much appreciate the positive effects of this PR, thanks a lot Philip :smiley: Eg the big file limitation pops up here once in a while.

Though I see a couple of potential gotchas here:
> This also means it formats the file system at boot if no valid filesystem is found.

- As an impatient user, I might constantly reset / replug the board when it takes ages to format a big flash. It should be ensured that the user gets an indication of what's going on. Don't know if that's already the case with the `NODE_ERR()` messages in `nodemcu_init()`.
- As a developer, I would dislike being forced to idle through the formatting phases when swapping images for debugging etc. Can we have a switch for devs to suppress this? Maybe in `user_config.h`?

pjsg:
Yes, there is a message that says that the filesystem is being formatted. 

If you are swapping images a lot, then it doesn't reformat every time. It leaves a gap between the end of the image and the start of the filesystem when it formats one automatically. When the next image starts, it searches for a valid image, and then uses it *even if* it would have put it somewhere else. 

If you don't want to wait, then I think that the best thing to do would be to format a really small filesystem. This way you will end up with a file system that works (rather than having to figure out how to fake out things like the open of "init.lua"). I'll add an option to be able to specify the max size of the automatically formatted file system. I think that the smallest size is probably 32k bytes. This should format in a fraction of a second.

jmattsson:
Hang on, we already have the `SPIFFS_FIXED_SIZE` option, do we really need yet another option in this area?

pjsg:
You are absolutely right -- let me rework the size limit.

pjsg:
I took a look at this today -- and there is no SPIFFS_FIXED_SIZE option. There is a SPIFFS_FIXED_LOCATION which sets the starting position of the filesystem -- however it is somewhat cumbersome to use that to control the size of the filesystem. Both size and location parameters may be useful once we support OTA upgrades where each running image is not allowed to use the full flash memory....

jmattsson:
Wait? Haven't I submitted that back upstream yet? We introduced that a quite some time ago at DiUS. Have a look at the spiffs.c in [our dius branch[(https://github.com/DiUS/nodemcu-firmware/blob/dius/app/spiffs/spiffs.c). We also have the SPIFFS_SIZE_1M_BOUNDARY define for automatic-but-within-an-OTA-slot sizing, for when we want a per-OTA-image filesystem rather than shared-between-OTA-images filesystem.

pjsg:
Yes. I'll add the 1M boundary option and I'll verify that the fixed size option does essentially the same thing.....

pjsg:
There are now lots of combinations of setting the file system size and boundaries. I've verified some of them.... I would expect the rest to work.

tools/spiffsimg/main.c
```diff
@@ -0,0 +1,433 @@
...
+  {
+    if (!sz) {
+      if (!flashsize || !used) {
+	die("Missing flashSize or Used");
+      }
+      sz = flashsize - used - 4 * 4096;   // This leaves space for the parameter area
+      if (sz < 0x4000) {
+	die("Not enough space");
+      }
+      if (sz > 512000) {
```
jmattsson: Do you mean 512*1024 here?

tools/spiffsimg/main.c
```diff
@@ -0,0 +1,433 @@
...
+    die ("mmap");
+
+  if (create)
+    memset (flash, 0xff, sz);
+
+  spiffs_config cfg;
+  cfg.phys_size = sz;
+  cfg.phys_addr = 0;
+  cfg.phys_erase_block = 0x1000;
+  cfg.log_block_size = 0x1000 * (sz > 512000 ? 2 : 1);
```
jmattsson: And here?

jmattsson:
I finally had a few minutes for a closer look. Some thoughts:

- These are substantial changes. I think it would be prudent to hold off on the merge until we've done the cut to master. Would you agree?

- Using the filesize of the 0x10000.bin to deduce the SPIFFS location makes the assumption that the irom0_seg is set to 0x40210000 in the linker file. A safer approach would be to look at the _flash_used_end symbol instead, and would also give coherency between the "outside" way and the NodeMCU way of working out the location. E.g.: ```echo $((0x`xtensa-lx106-elf-objdump -t app/.output/eagle/debug/image/eagle.app.v6.out |grep _flash_used_end |cut -f1 -d" "` - 0x40200000))```

- You added a few options to spiffsimg, could you document them in the corresponding README.md too please?

- The `tools/Makefile` does not look like it should use the "recursion magic" as Espressif calls it, and you explicitly filter it out in the top Makefile anyway. You can safely remove all the PDIR and related Espressif stuff (including copyright notice) I believe. It's a plain old Makefile :)

I don't have time just yet to test all this for compatibility with what we're using here at work, but I can always submit some tweaks later on after it's merged and we decide to pull in the latest into our dius branch. Thanks heaps for incorporating the fixed-size and 1M-boundary options!

pjsg:
These changes are definitely *NOT* to be merged until after the merge to master. They will need time to settle down. I've been running them for a while, but there are lots of options and people will try different things.

I will make the other changes as you recommend. I took some of the spiffsimg makefile wrapping from another PR to this repo without thinking about it too much. 

I will update the README.md and get rid of the recursion magic.... 

Probably will be this weekend when I get a chance to work on it....

There is an implicit tradeoff here between the amount of memory devoted to the file system, and the gap between the end of the code and the start of the file system. The bigger the gap, the more likely it is that the file system will survive a reflash of the code. This is a huge win during development. For small flash chip sizes (and this may include the 1MB flash in the 8265) you probably don't want to waste much space -- especially for the production builds. I expect that people will want more control over these parameters.

jmattsson:
> The bigger the gap, the more likely it is that the file system will survive a reflash of the code. This is a huge win during development.

Oh absolutely, I wasn't suggesting you had to pack it right after _flash_used_end, only that you can use it to find the minimum address (which still needs padding to end of block etc).

pjsg:
Any thoughts on making the standard filename longer?

jmattsson:
Is it really necessary? We're on a highly memory constrained device, it seems... frivolous to spend more memory on something cosmetic such as filenames :)

devyte:
Per #1328 the use case explained there is use of uuid for filenames, which doesn't fit in 31 chars. In such a case it's not really cosmetic.
Also, the image gets 80 bytes larger, and there's no noticeable difference in ram (which is the tightest constraint). True, holding longer filenames in lua scripts would take more ram due to the longer strings, but I'd think that that should be handled by whoever uses longer filenames.

My personal experience has been that a size of 32 is too short, which has made me struggle. I would love to go up to 128, but 64 would certainly be an improvement.

oyooyo:
> Is it really necessary? We're on a highly memory constrained device, it seems... frivolous to spend more memory on something cosmetic such as filenames :)

I did a test in #1328 to see how big the effect would be. It suggested that increasing the max. filename length from 31 to 63 characters alone actually had no effect at all on both RAM and available SPIFFS storage space. The only difference was that the firmware image was 80 bytes larger.

Since I still have little experience, there's a good chance I simply did something wrong, so I would welcome if someone more experienced tries to check this for himself to see if he comes to same conclusion.

pellepl:
Looks correct.

The file names are stored in first logical index page of a file. Increasing
the size of file names will not make the page larger, but it will leave
less space for the page lookup in the first  index page. Meaning a file
will need to have extra index pages sooner than before wrt file size.

All in all, as long as you have filenames not being bigger than ~half
logical page size, the impact will be small. Some extra index tables in
flash, some extra look ups performancewise, and some dir entry structs in
ram will be bigger. But these are almost always on stack.

pjsg:
Also note that I doubled the page size (to 256) that we use in Nodemcu on the advice of @pellepl so it should be less of a concern. 

As I see it, there are two options -- one is to just document how to change the filename length and put the #define into user_config.h with the default value of 32. The other is to change it to (say) 64 and not document how to change it.

I'm leaning to the latter as having a single filename length on the platform means that there is more compatibility from build to build.

marcoskirsch:
It's true that this is a memory constrained device. But this firmware is about running Lua on it, and. Is largely to make it easier to use. 64 character file names makes things easier since it's less likely one would run into limitations,

I would vote to do follow @pjsg's suggestion.

marcelstoer:
Philip, I don't know whether you omitted the third option on purpose: keep the current limit in place and don't offer a `#define` in `user_config.h`. I consider the need for such long files a very rare use case. However, it's still one you can fulfill by tweaking the SPIFFS config.

pjsg:
We could do that -- though I would rather that it be documented so that people did not have to open an issue to find out how to do it.

oyooyo:
While I would personally vote for option 4 (increase the default limit _and_ move the setting to user_config.h), I would still prefer option 2 (increase the default, don't document how to change it) over option 1 (keep the default, but move setting to user_config.h). So it's 4 > 2 > 1 > 3 for me.

I guess most users will be able to cope with a 31 characters limit just fine. But then again, a few people have indeed run into this limit already (the issue tracker lists three issues concerning the filename length limit in the last three months). So why not simply increase the default value? As I get it, nobody has yet found an actual, noteworthy disadvantage of doing so.

marcelstoer:
I'd rather not hold up this PR any longer just because of this filename length discussion. There's no necessity to include it in this PR. If you decide to change the limit it could just as well go into a separate PR.

As I see it Johny and I are both somewhat reluctant to increase it, I'm not sure about yourself, though. Your call Philip. Should you decide to leave it as it is you could of course still ad a note about how to raise the limit to the docs.

pjsg:
I think that I don't want to couple the two issues. I'd like to get this merged -- and there will probably be a period of shakeout. I think that opening up an issue to increase the length of the filenames is probably the right way to go here.

docs/en/spiffs.md
```diff
@@ -0,0 +1,118 @@
+# SPIFFS File System
+
+The NodeMCU project uses the SPIFFS filesystem to store files in the flash chip. The technical details about how this is configured can be found below, along with vairous build time options.
```
marcelstoer: vairous -> various

marcelstoer: Add a link to https://github.com/pellepl/spiffs behind 'SPIFFS'?

docs/en/spiffs.md
```diff
@@ -0,0 +1,118 @@
...
+thing onto your microprocessor's storage instead of painstakingly upload
+file-by-file through your app on the micro? With spiffsimg you can!
+
+NodeMCU uses a SPIFFS filesystem that knows how big it is -- i.e. when you build a file system
+image, it must fit into the flash chip, and it cannot be expanded once flashed.
+It is important to give the `spiffimg` tool the correct size. You can provide either the `-c` option or both the `-U` and `-S` options.
+
+### Syntax 
+
+`spiffsimg -f <filename>
```
marcelstoer: I think you want to use 3 back ticks instead of one - otherwise it's all be on a single line.

docs/en/spiffs.md
```diff
@@ -0,0 +1,118 @@
...
+	[-o <offsetfile>]
+	[-c <size>] 
+	[-S <flashsize>]
+	[-U <usedsize>]
+	[-d]
+	[-l | -i | -r <scriptname> ]`
+
+### Supported operations:
+
+  * -f specifies the filename for the disk image. '%x' will be replaced by the calculated offset of the file system.
```
marcelstoer: I'd rather have all of the switches formatted as code (i.e. wrapped by back ticks).

docs/en/spiffs.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Known limitations:
+
+  * The block & page sizes are hard-coded to be compatible with nodemcu.
+  * Error handling is not entirely consistent, some errors result in an
+    early exit, others just print an error (both cause a non-zero exit though).
+  * Only flat SPIFFS is supported.
+
+
+### See also:
```
marcelstoer: If you go for the link in the intro paragraph I'd skip this one.

tools/spiffsimg/README.md
```diff
@@ -0,0 +1,66 @@
+# spiffsimg - Manipulate SPI Flash File System disk images
+
+Ever wished you could prepare a SPIFFS image offline and flash the whole
```
marcelstoer: I'd much rather link to nodemcu.readthedocs.io/en/dev/en/spiffs/ (or ../../docs/en/spiffs.md) than repeating that content here.

marcelstoer:
There's one already: #1328.

tools/spiffsimg/README.md
pjsg: I did that as well.

pjsg:
I think I'll just add a comment to the docs about changing the limit.... and then close this issue out.

PR #1233, opened 2016-04-10
Add net.server:getpeer() for UDP servers

TerryE:
See issue #223.  This fixes the bug in the Espressif LiWP UDP wrapper and adds getpeer processing for UDP.

### DIFF
--- a/app/lwip/app/espconn_udp.c
+++ b/app/lwip/app/espconn_udp.c
@@ -308,6 +308,11 @@ espconn_udp_recv(void *arg, struct udp_pcb *upcb, struct pbuf *p,
 		wifi_get_ip_info(0, &ipconfig);
 	}
 
+  precv->pespconn->proto.udp->remote_ip[0] = ip4_addr1_16(addr);
+  precv->pespconn->proto.udp->remote_ip[1] = ip4_addr2_16(addr);
+  precv->pespconn->proto.udp->remote_ip[2] = ip4_addr3_16(addr);
+  precv->pespconn->proto.udp->remote_ip[3] = ip4_addr4_16(addr);
+  precv->pespconn->proto.udp->remote_port = port;
 	precv->pespconn->proto.udp->local_ip[0] = ip4_addr1_16(&ipconfig.ip);
 	precv->pespconn->proto.udp->local_ip[1] = ip4_addr2_16(&ipconfig.ip);
 	precv->pespconn->proto.udp->local_ip[2] = ip4_addr3_16(&ipconfig.ip);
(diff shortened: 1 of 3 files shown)


app/lwip/app/espconn_udp.c
```diff
@@ -308,6 +308,11 @@ espconn_udp_recv(void *arg, struct udp_pcb *upcb, struct pbuf *p,
 		wifi_get_ip_info(0, &ipconfig);
 	}
 
+  precv->pespconn->proto.udp->remote_ip[0] = ip4_addr1_16(addr);
+  precv->pespconn->proto.udp->remote_ip[1] = ip4_addr2_16(addr);
+  precv->pespconn->proto.udp->remote_ip[2] = ip4_addr3_16(addr);
+  precv->pespconn->proto.udp->remote_ip[3] = ip4_addr4_16(addr);
+  precv->pespconn->proto.udp->remote_port = port;
```
robertfoss: Is the indentation correct here?

TerryE: The Lua / nodeMCU standard is a two space tabstop. The LwIP code uses hard tabs, and I missed this. I will update, but have you tried the function?

TerryE:
Don't merge this just yet. Even though this works, it looks like Espressif have implemented on alternative

TerryE:
This all a bit bizarre. The inline documentation for [`espconn_get_connection_info()`](/nodemcu/nodemcu-firmware/blob/dev/app/lwip/app/espconn.c#L825) is a cut and paste clone of that  for `espconn_regist_disconcb()`.  The description in the SDK states "Get the information about a TCP connection or UDP transmission" where the returned argument is "the connect to client info".  

The code doesn't return this info; it returns the address of an `remot_info` array. rather than the entry relating to this connection which (using the SDK example) would be `pcon_info[espconn->link_cnt]`, so ***the current uses of `espconn_get_connection_info()` will only return the correct IP if the link happens to be the first link***.

 Internally the espconn layer seems to use a set of `espconn_msg` linked lists to maintain its state strucutres and primarily used the address of the `struct_espconn` as a moniker search into this. 

So this patch works but seems to be adopting a different approach to espconn implementation, but the espconn implementation itself seems to be flawed -- at  least in this area.  

What to do?  I will need to think about this.

jmattsson:
The entire espconn layer has issues. Serious issues. I've griped about this in the past. The more I'm exposed to that code, the more I want to rip it out and replace it.

TerryE:
Seems a bold step that might have quite a few unintended consequences,, but I can see where you are coming from.

jmattsson:
Given the net module cleanup you're doing Terry, is this PR still relevant?

TerryE:
No its not.  Thanks for the reminder.

PR #1235, opened 2016-04-11
changed documentation for ws2812 module, …

cogitatio:
… which only works for pin GPIO2 at the moment


Alkorin:
Thanks, I still have some pending work on documentation... should try to find some time to finish it ... :(

Issue #1236, opened 2016-04-11
mDNS: response sent to source instead of multicast address


FrankX0:
Thanks to @pjsg the mDNS functionality works better than before. But I found there is still room to improve.

When a request is sent to the multicast address 224.0.0.251, the node should send the response to the multicast address. Currently the response is sent to the source address.
The issue is in the nodemcu_mdns.c file, function mdns_recv.
When changing all the destination addresses to &multicast_addr, the conversion from a .local name to an IP address now works correctly for most browsers (Chrome, Safari, Firefox), ping and dns-sd.
In my opinion the regularly sending (currently every 120 seconds) of ms_info is no longer required (tested by disabling this functionality).
Please note that MS Edge is still not satisfied with the response. I'll try to find out why.

I would have liked to implement this change myself in the repository, but do not yet feel comfortably enough to do so.

TerryE:
>.I would have liked to implement this change myself in the repository, but do not yet feel comfortably enough to do so.

There is always the lesser option of proposing a patch if you've worked out a solution.

FrankX0:
I implemented following patch in _nodemcu_mdns.c_, in function _mdns_recv_, changing the source address of the mdns send calls from _addr_ to _&multicast_addr_
So, from:
`mdns_send_service_type(i, addr, port);`
to
`mdns_send_service_type(i, &multicast_addr, port);`
for all occurrences of _mdns_send_service_type()_ and _mdns_send_service()_.

pjsg:
You should only send the response to the multicast address if it is a QM query. 

I filed a PR to resolve this....

FrankX0:
Resolved by #1239 
Thanks @pjsg!

PR #1239, opened 2016-04-13
Send a multicast respone if it is a QM query (and other MDNS fixes)

pjsg:
This fixes the bug reported in #1236 -- now mdns sends the response to the multicast address if it is a QM query and to the unicast address if it is a QU query.

Tested, and appears to work.

### DIFF
--- a/app/net/nodemcu_mdns.c
+++ b/app/net/nodemcu_mdns.c
@@ -59,6 +59,8 @@
 
 #define MDNS_NAME_LENGTH     68 //68
 
+#define DNS_RRTYPE_NSEC    47
+
 static const char* service_name_with_suffix    = NULL;
 #define DNS_SD_SERVICE       "_services._dns-sd._udp.local"
 #define PUCK_SERVICE_LENGTH  30
@@ -176,12 +178,12 @@ PACK_STRUCT_END
 
 PACK_STRUCT_BEGIN
 /** MDNS answer message structure */
-struct mdns_auth {
+struct mdns_a_rr {
 	PACK_STRUCT_FIELD(u32_t src);
 }PACK_STRUCT_STRUCT;
 PACK_STRUCT_END
 
-#define SIZEOF_MDNS_AUTH 4
+#define SIZEOF_MDNS_A_RR 4
 PACK_STRUCT_BEGIN
 /** MDNS service registration message structure */
 struct mdns_service {
@@ -234,7 +236,15 @@ mdns_compare_name(unsigned char *query, unsigned char *response) {
 		} else {
 			/* Not compressed name */
 			while (n > 0) {
-				if ((*query) != (*response)) {
+				char q = *query;
+				if (q >= 'A' && q <= 'Z') {
+				  q = q + 'a' - 'A';
+				}
+				char r = *response;
+				if (r >= 'A' && r <= 'Z') {
+				  r = r + 'a' - 'A';
+				}
+				if (q != r) {
 					return 1;
 				}
 				++response;
@@ -325,7 +335,7 @@ mdns_send_service_type(u16_t id, struct ip_addr *dst_addr, u16_t dst_port) {
 	err_t err;
 	struct mdns_hdr *hdr;
 	struct mdns_answer ans;
-	struct mdns_auth auth;
+	struct mdns_a_rr a_rr;
 	struct mdns_service serv;
 	struct pbuf *p ,*p_sta;
 	char *query, *nptr;
@@ -418,7 +428,7 @@ mdns_send_service(struct nodemcu_mdns_info *info, u16_t id, struct ip_addr *dst_
 	struct mdns_hdr *hdr;
 	struct mdns_answer ans;
 	struct mdns_service serv;
-	struct mdns_auth auth;
+	struct mdns_a_rr a_rr;
 	struct pbuf *p ,*p_sta;
 	char *query, *nptr;
 	char *query_end;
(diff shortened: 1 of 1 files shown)


FrankX0:
Thanks for fixing #1236 in this way!
Like reported earlier, most browsers work correctly now with the current (improved) mDNS implementation, except for MS Edge.
Looking at the Ethernet traffic, this browser first makes an AAAA request. The current code does not respond to that request. But in response, MS Edge will not do an A request.
According [Common Misbehavior Against DNS Queries for IPv6 Addresses](https://www.ietf.org/rfc/rfc4074.txt):

> Suppose that an authoritative server has an A RR but has no AAAA RR
   for a host name.  Then, the server should return a response to a
   query for an AAAA RR of the name with the response code (RCODE) being
   0 (indicating no error) and with an empty answer section (see
   Sections 4.3.2 and 6.2.4 of [1]).  Such a response indicates that
   there is at least one RR of a different type than AAAA for the
   queried name, and the stub resolver can then look for A RRs.

Would it be possible to implement this?

pjsg:
I'll update the PR later on this evening when I have tested this...

pjsg:
I added handling of missing RRs. It turns out that mdns uses NSEC records to signal this. THis now sends NSEC records as appropriate -- it sends the marker for the A record on the big packet and dynamically generates the NSEC if it gets a query.

FrankX0:
Still no luck with MS Edge.
When it requests a type A record, the current firmware responds with PTR, TXT, SRV, A and SREC records. It seems MS Edge does not seem how to deal with this response.
Would it be possible to create a function which only responds with an A record?

pjsg:
I'll get to this when I get back home....
---
Philip Gladstone. On the road, please excuse typos.
On Apr 17, 2016 08:11, FrankX <notifications@github.com> wrote:Still no luck with MS Edge.
When it requests a type A record, the current firmware responds with PTR, TXT, SRV, A and SREC records. It seems MS Edge does not seem how to deal with this response.
Would it be possible to create a function which only responds with an A record?

pjsg:
THis now responds directly to a A query (and also to AAAA). Lets hope that this works with Edge!!

pjsg:
I did download the Windows 10 EDGE test VM and it was able to resolve my test name correctly. Having said that, I don't know if all the paths are the same when Edge is running inside a VM. @FrankX0 Would appreciate a real test.

FrankX0:
I did some testing, but unfortunately less successful as yours.
Without the mDNS firmware running, the MS Edge browser will perform a type A and AAAA request:
```
Queries
    nodemcu3.local: type A, class IN, "QM" question
    nodemcu3.local: type AAAA, class IN, "QM" question
```
With the current firmware implementation running however, I do not see any mDNS request from the MS Edge browser.
When I disable the automatic transmission of the supported services, the browser will perform the type A and AAAA requests again:
```
Queries
    nodemcu3.local: type A, class IN, "QM" question
    nodemcu3.local: type AAAA, class IN, "QM" question
```
The firmware will then respond with a type A and an additional NSEC record:
```
Answers
        nodemcu3.local: type A, class IN, addr 192.168.1.106
Additional records
        nodemcu3.local: type NSEC, class IN, next domain name nodemcu3.local
```
Then the browser sends again a type A query with the previous answer (the retrieved IP address).
I think this is where things start to go wrong (maybe it is expecting 2 answers instead of 1 answer and 1 additional record):
```
Queries
        nodemcu3.local: type A, class IN, "QM" question
Answers
        nodemcu3.local: type A, class IN, addr 192.168.1.106
```
The firmware responds in the same way (type A and NSEC record). The browser responds with a type AAAA query and the NSEC record:
```
Queries
        nodemcu3.local: type AAAA, class IN, "QM" question
Answers
        nodemcu3.local: type NSEC, class IN, next domain name nodemcu3.local
```
And the following NodeMCU response:
```
Answers
    nodemcu3.local: type NSEC, class IN, next domain name nodemcu3.local
```
Browser request:
```
Queries
        nodemcu3.local: type A, class IN, "QM" question
        nodemcu3.local: type AAAA, class IN, "QM" question
Answers
        nodemcu3.local: type A, class IN, addr 192.168.1.106
        nodemcu3.local: type NSEC, class IN, next domain name nodemcu3.local
```
NodeMCU response:
```
Answers
        nodemcu3.local: type A, class IN, addr 192.168.1.106
Additional records
        nodemcu3.local: type NSEC, class IN, next domain name nodemcu3.local
```
After that the browser requests a NetBIOS name query.
Apparently MS Edge did not understand the mDNS responses to resolve the name.

pjsg:
I read the RFC a little more closely and it turns out that NSEC record is supposed to be an additional record. This wasn't consistent, so I fixed that. Also, unicast responses are supposed to have no more a TTL of 10 seconds, so I fixed that as well.

Just a question -- are you running these tests in AP mode, STA more or AP+STA?

As far as I can tell, in windows 10 there is native mdns support (at some level). Thus once the firmware starts sending out the announce message, it should be cached locally by the OS, and so when Edge tries to get to that name, the data should already be available.

The firmware does not handle multiple questions, but I'm not sure that that is the problem....

jmattsson:
Does this PR need any further testing before merge?

pjsg:
I was hoping that @FrankX0 would give it a whirl. In any case, the code is better now than it was and could be merged. The Edge problem is not a regression, but various other things do now work.

FrankX0:
Sorry, could not respond earlier. Although Edge can still not connect with the current implementation, I agree with merging. It is much better now as most browser do work witness the current implementation. We can always work out later why Edge is the odd one out.

pjsg:
Is there anything else that needs to be addressed?

I've tried running Edge in one of the VMs that microsoft provides, but I can't get the VM to do anything with mDns.....

marcelstoer:
Even if it should turn out to be true that not 100% of a all possible use cases are fully covered this PR is certainly a step forward -> merged.

PR #1240, opened 2016-04-13
Added ECCDSA elliptic curve hashing to the crypto library

kieransimkin:
.. then realised I could program in C and promptly left lua. But crypto should have some form of signing and ECDSA is the obvious choice for space constraints

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "app/crypto/micro-ecc"]
+	path = app/crypto/micro-ecc
+	url = https://github.com/kmackay/micro-ecc.git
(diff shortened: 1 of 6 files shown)


pjsg:
There are a bunch of issues with this:

* You should target the 'dev' branch
* You shouldn't disable compiler warnings -- fix the code
* You should have written documentation
* You are using rand() as a secure random number generator. It isn't.
* This should have been a new source file
* I don't see any way to select the curve.

kieransimkin:
last one is answered by a #define in uECC.h, for minimum size you only get one at a time. the rest I don't have time to do now

PR #1243, opened 2016-04-15
SNTP module enhancements

jmattsson:
Two small but important enhancements included here:

- Improved diagnostics by way of error code now passed to the error handler function
- Switch to use an ephemeral source port for NTP requests.

The necessity of the latter we discovered through the use of the former. It appears that some ISPs, in their finite wisdom, block all NTP-port traffic to their customers presumably as a misguided attempt at limiting DDoS attempts. So even though NTP is meant to run 53<->53, and it's only a subtype 7(?) control packet which has any chance of causing amplification, these ISPs block it all if it's destined to port 53 on the customer's end. Switching to an ephemeral source port resolved the problem in both our encountered cases.

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -58,6 +58,14 @@
 # define sntp_dbg(...)
 #endif
 
+typedef enum {
+  NTP_NO_ERR = 0,
+  NTP_DNS_ERR,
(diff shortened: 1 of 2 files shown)


marcelstoer:
Very useful, thanks DiUS.

PR #1252, opened 2016-04-18
Added support for Analog Devices ADXL345 accelerometer

jschmidlapp:


marcelstoer:
Can't comment on the C code but on a number of formalities. See https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md#writing-documentation:

- we appreciate descriptions for PRs
- modules in `user_modules.h` should be alphabetically sorted a-z
- modules in `mkdocs.yml` should be alphabetically sorted a-z
- functions in the module documentation should be alphabetically sorted a-z

app/modules/adxl345.c
```diff
@@ -0,0 +1,101 @@
...
+    return ret;
+}
+
+static int ICACHE_FLASH_ATTR adxl345_init(lua_State* L) {
+
+    uint32_t sda;
+    uint32_t scl;
+    uint8_t  devid;
+
+    if (!lua_isnumber(L, 1) || !lua_isnumber(L, 2)) {
```
TerryE: This check is redundant so drop it; the check integer calls do this anyway. Also use the arg check macro instead of an if predicate and and error: less source line and a better diagnostic for the app programmer.

jmattsson:
:+1: in general. I might have preferred `.read()` rather than `.data()` as the former suggests obtaining fresh data whereas the latter could conceptually refer to cached data. I don't know if we have any actual guidelines for naming conventions in this area though - anyone?

marcelstoer:
Good point Johny, I agree. When I first looked at the API `data()` felt somehow odd but I couldn't quite picture why. I personally always prefer semantically explicit vs. implicit function names. Hence, I'd much rather have `read()` and an *additional* function should there ever arise the need for a function that caches internally (not now).
To my knowledge we don't have any conventions. That'd be something for the "how to write a NodeMCU module" documentation (we used to have an issue/PR but I couldn't find it anymore).

jschmidlapp:
Sorry, been out of town for a few weeks.  Changing the API from data() to
read() makes sense to me.  I'll make the change to all of my new sensor
modules and push the changes.

-Jason

PR #1253, opened 2016-04-18
Added support for Honeywell HMC5883L digital compass

jschmidlapp:


marcelstoer:
Can't comment on the C code but on a number of formalities. See https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md#writing-documentation:

- we appreciate descriptions for PRs
- modules in `user_modules.h` should be alphabetically sorted a-z
- module documentation should be added to `mkdocs.yml`
- modules in `mkdocs.yml` should be alphabetically sorted a-z
- functions in the module documentation should be alphabetically sorted a-z

app/modules/hmc5883l.c
```diff
@@ -0,0 +1,114 @@
...
+    platform_i2c_send_stop(hmc5883_i2c_id);
+}
+
+static int ICACHE_FLASH_ATTR hmc5883_init(lua_State* L) {
+
+    uint32_t sda;
+    uint32_t scl;
+    uint8_t  devid_a, devid_b, devid_c;
+
+    if (!lua_isnumber(L, 1) || !lua_isnumber(L, 2)) {
```
TerryE: This check is redundant so drop it; the check integer calls do this anyway. Also use the arg check macro instead of an if predicate and and error: less source line and a better diagnostic for the app programmer.

jmattsson:
:+1: with similar comments as on #1252.

Issue #1254, opened 2016-04-18
Security issue: oversized CoAP hostnames can trigger a stack overflow in coap_request() and lead to remote code execution


jsandin:
### Impact:

Lua code that allows untrusted clients to specify the uri for a coap request is susceptible to remote code execution.  A malicious client can send a specially crafted coap uri containing an oversized host that overwrites the saved a0 register on the stack and causes execution to redirect to an address of the attacker's choosing when coap_request() returns.

This does not affect nodeMCU by default - only programs that allow clients to control the coap uri are affected.  The following (toy) example program is an example of a small Lua program that exposes this vulnerability to remote attackers:

`cs = coap.Client(); s=net.createServer(net.UDP); s:on("receive",function(s,c) cs:get("coap://"..c) end); s:listen(8889)`

If you paste this into nodeMCU and send an oversized string of As using netcat to port 8889 on the device, you can trigger a crash.

### Description:

The coap_request() function allocates a 32-byte 'host' buffer on the stack, and copies the host portion of a coap uri (constructed using coap_new_uri()) into this buffer, even if the size of the host portion of the uri exceeds 32 bytes.  An attacker that sends a uri containing an oversized hostname will cause a stack overflow that overwrites other variables on the stack as well as saved registers a12, a13, a14, a15, and a0.  When the function returns, it will used this modified saved a0 value as the target address to execute, allowing the attacker to redirect execution to an address of their choosing.

### Recommendation:

Either reject uri strings that contain hostnames exceeding 32 bytes, or allocate sufficient memory to store the hostname before performing the memcpy on line 357 of coap.c.

Enhance coap_new_uri() to perform further validation of CoAP uris and reject malformed uris.

### Technical details:

from nodemcu-firmware/app/modules/coap.c - this vulnerability is present in both the current development branch and 0.9.6 release:

```C
313 static int coap_request( lua_State* L, coap_method_t m )
314 {
315   struct espconn *pesp_conn = NULL;
316   lcoap_userdata *cud;
317   int stack = 1;
...
326   stack++;
327   pesp_conn = cud->pesp_conn;
328   ip_addr_t ipaddr;
329   uint8_t host[32];
...
348   coap_uri_t *uri = coap_new_uri(url, l);   // should call free(uri) somewhere
349   if (uri == NULL)
350     return luaL_error( L, "uri wrong format." );
351
352   pesp_conn->proto.udp->remote_port = uri->port;
[...]

PR #1260, opened 2016-04-23
Added length checking to prevent stack overflow

pjsg:
As recommended in the bug report #1254 , added length checking. Also I increased the size of the buffer to make it less likely to error out under normal circumstances.

Verified that the limit works.

However, something else terrible happens (sometimes) and I get a crash after a few seconds. I doubt that this is related..... I just opened #1261 which is a separate problem and much more difficult to exploit (if it is exploitable).

### DIFF
--- a/app/modules/coap.c
+++ b/app/modules/coap.c
@@ -326,7 +326,7 @@ static int coap_request( lua_State* L, coap_method_t m )
   stack++;
   pesp_conn = cud->pesp_conn;
   ip_addr_t ipaddr;
-  uint8_t host[32];
+  uint8_t host[64];
 
   unsigned t;
   if ( lua_isnumber(L, stack) )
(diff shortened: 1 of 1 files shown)


app/modules/coap.c
```diff
@@ -348,6 +348,9 @@ static int coap_request( lua_State* L, coap_method_t m )
   coap_uri_t *uri = coap_new_uri(url, l);   // should call free(uri) somewhere
   if (uri == NULL)
     return luaL_error( L, "uri wrong format." );
+  if (uri->host.length >= sizeof(host) - 1) {
```
marcelstoer: Where does the `- 1` come from?

pjsg:
There is an extra null that is added to the end of the hostname. In fact, this is the whole reason that the copy takes place -- the original buffer is not null terminated -- it is just a pointer and length into the URL. Also, I think that the only allowable value is actually an ip address -- it doesn't appear to support doing the lookup.

I think that the code really needs a lot of work, but it is worth fixing this specific issue anyway.

marcelstoer:
Thanks for the clarification and the extra comment in the source code.

PR #1263, opened 2016-04-24
Ws2812

Alkorin:
As promised but quite late sorry, the updated code with documentation.

Some improvements here: 
- Get rid of `ets_intr_lock`
- Add `ws2812.init()` to limit `ICACHE_RAM_ATTR` code
- Drop `ws2812.writergb`, it could be handled on LUA side (and then removed the naked malloc)
- Add support of RGBW strips

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -51,7 +51,9 @@ static void ICACHE_RAM_ATTR ws2812_write(uint8_t *pixels, uint32_t length) {
   // 0b00000111 => 000111 => [0]111000[1] => 10001110 => 01
   // 0b00110100 => 110100 => [0]001011[1] => 11101000 => 10
   // 0b00000100 => 000100 => [0]001000[1] => 11101110 => 11
-  uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
+  // Array declared as static const to avoid runtime generation
+  // But declared in ".data" section to avoid read penalty from FLASH
+  static const __attribute__((section(".data._uartData"))) uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
 
   uint8_t *end = pixels + length;
 
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -1,5 +1,6 @@
 #include "module.h"
 #include "lauxlib.h"
+#include "lmem.h"
 #include "platform.h"
 #include "c_stdlib.h"
 #include "c_string.h"
@@ -14,9 +15,35 @@ typedef struct {
   uint8_t values[0];
 } ws2812_buffer;
 
+// Init UART1 to be able to stream WS2812 data
+// We use GPIO2 as output pin
+static void ICACHE_FLASH_ATTR ws2812_init() {
+  // Configure UART1
+  // Set baudrate of UART1 to 3200000
+  WRITE_PERI_REG(UART_CLKDIV(1), UART_CLK_FREQ / 3200000);
+  // Set UART Configuration No parity / 6 DataBits / 1 StopBits / Invert TX
+  WRITE_PERI_REG(UART_CONF0(1), UART_TXD_INV | (1 << UART_STOP_BIT_NUM_S) | (1 << UART_BIT_NUM_S));
+
+  // Pull GPIO2 down
+  platform_gpio_mode(4, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
+  platform_gpio_write(4, 0);
+
+  // Waits 10us to simulate a reset
+  os_delay_us(10);
+
+  // Redirect UART1 to GPIO2
+  // Disable GPIO2
+  GPIO_REG_WRITE(GPIO_ENABLE_W1TC_ADDRESS, BIT2);
+  // Enable Function 2 for GPIO2 (U1TXD)
+  PIN_FUNC_SELECT(PERIPHS_IO_MUX_GPIO2_U, FUNC_U1TXD_BK);
+}
+
+
 // Stream data using UART1 routed to GPIO2
+// ws2812.init() should be called first
+//
 // NODE_DEBUG should not be activated because it also uses UART1
-static void ICACHE_RAM_ATTR ws2812_write(uint8_t pin, uint8_t *pixels, uint32_t length) {
+static void ICACHE_RAM_ATTR ws2812_write(uint8_t *pixels, uint32_t length) {
 
   // Data are sent LSB first, with a start bit at 0, an end bit at 1 and all inverted
   // 0b00110111 => 110111 => [0]111011[1] => 10001000 => 00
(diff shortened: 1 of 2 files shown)


marcelstoer:
Terry, Johny or Arnim could one of you please have a look so we can merge this before the dev-to-master snap?

app/modules/ws2812.c
```diff
@@ -14,30 +15,45 @@ typedef struct {
...
-  // Data are sent LSB first, with a start bit at 0, an end bit at 1 and all inverted
-  // 0b00110111 => 110111 => [0]111011[1] => 10001000 => 00
-  // 0b00000111 => 000111 => [0]111000[1] => 10001110 => 01
-  // 0b00110100 => 110100 => [0]001011[1] => 11101000 => 10
-  // 0b00000100 => 000100 => [0]001000[1] => 11101110 => 11
-  uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
-
+// Init UART1 to be able to stream WS2812 data
+// We use GPIO2 as output pin
+static void ICACHE_FLASH_ATTR ws2812_init() {
```
devsaurus: Please remove the `ICACHE_FLASH_ATTR` instances - they're redundant.

app/modules/ws2812.c
```diff
@@ -14,30 +15,45 @@ typedef struct {
...
+//
+// NODE_DEBUG should not be activated because it also uses UART1
+static void ICACHE_RAM_ATTR ws2812_write(uint8_t *pixels, uint32_t length) {
+
+  // Data are sent LSB first, with a start bit at 0, an end bit at 1 and all inverted
+  // 0b00110111 => 110111 => [0]111011[1] => 10001000 => 00
+  // 0b00000111 => 000111 => [0]111000[1] => 10001110 => 01
+  // 0b00110100 => 110100 => [0]001011[1] => 11101000 => 10
+  // 0b00000100 => 000100 => [0]001000[1] => 11101110 => 11
+  uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
```
devsaurus: `_uartData` is a local variable - is there a runtime penalty in case it's being initialized on the stack each time `ws2812_write()` is called?

app/modules/ws2812.c
```diff
@@ -91,11 +93,13 @@ static int ICACHE_FLASH_ATTR ws2812_writegrb(lua_State* L) {
 // Handle a buffer where we can store led values
 static int ICACHE_FLASH_ATTR ws2812_new_buffer(lua_State *L) {
   const int leds = luaL_checkint(L, 1);
+  const int colorsPerLed = luaL_checkint(L, 2);
 
   luaL_argcheck(L, leds > 0, 1, "should be a positive integer");
+  luaL_argcheck(L, colorsPerLed > 0, 2, "should be a positive integer");
```
devsaurus: Why is `colorsPerLed` not constrained to 3 & 4 as stated in the doc?

docs/en/modules/ws2812.md
```diff
@@ -1,57 +1,163 @@
 # WS2812 Module
 | Since  | Origin / Contributor  | Maintainer  | Source  |
 | :----- | :-------------------- | :---------- | :------ |
-| 2015-02-05 | [Till Klocke](https://github.com/dereulenspiegel) | [Till Klocke](https://github.com/dereulenspiegel) | [ws2812.c](../../../app/modules/ws2812.c)|
+| 2015-02-05 | [Till Klocke](https://github.com/dereulenspiegel), [Thomas Soëte](https://github.com/Alkorin) | [Till Klocke](https://github.com/dereulenspiegel) | [ws2812.c](../../../app/modules/ws2812.c)|
 
+ws2812 is a library to handle ws2812-like led strips.
+It works at least on WS2812, WS2812b, APA104, SK6812 (RGB or RGBW).
+
+The library uses UART1 routed on GPIO2 to generate the bitstream.
```
devsaurus: Please also mention pin index 4 for GPIO2 on NodeMCU boards.

app/modules/ws2812.c
Alkorin: Done

app/modules/ws2812.c
Alkorin: I could, but if somesay some WS2812-like leds have 5 colors... ?

docs/en/modules/ws2812.md
```diff
@@ -1,57 +1,163 @@
...
+
+#### Returns
+`(color)`
+
+#### Example
+```lua
+buffer:get(2) -- return the color of the third led
+```
+## ws2812.buffer:set()
+Setthe value at the given position
```
devsaurus: Nospace left :wink:

docs/en/modules/ws2812.md
```diff
@@ -1,57 +1,163 @@
...
 #### Parameters
-- `pin` is ignored. Only works with GPIO2.
-- `string` payload to be sent to one or more WS2812 LEDs.
-  It should be composed from an RGB triplet per element.
-    - `R1` the first pixel's Red channel (0-255)
-    - `G1` the first pixel's Green channel (0-255)
-    - `B1` the first pixel's Blue channel (0-255)<br />
-    ... You can connect a lot of WS2812 ...
-    - `R2`, `G2`, `B2` are the next WS2812's Red, Green, and Blue channel parameters
+ - `index` position in the buffer
```
devsaurus: The range for `index` is 0 to n, but the [Lua convention for arrays](https://www.lua.org/pil/11.1.html) is to start at index 1. Should this be applied here as well (and for `ws2812.buffer:get()`)?

app/modules/ws2812.c
Alkorin: Yep, fixed

docs/en/modules/ws2812.md
Alkorin: Done

app/modules/ws2812.c
devsaurus: I see, thanks.

devsaurus:
Apart from the questions above, I had a functional test of the driver. The overall improvements wrt SDK compliance are definitely paying off - I was able to mash up a VU meter with 8 WS2812 for the PCM playback and observed no drop-outs, no flickering, nothing. 
The new design of this driver is a big improvement IMO. Thanks a lot Thomas :smiley:

Alkorin:
I'll try to change the "index is 0 to n" this evening and all remarks will be fixed

docs/en/modules/ws2812.md
Alkorin: Changed in code and documentation

app/modules/ws2812.c
```diff
@@ -51,7 +51,7 @@ static void ICACHE_RAM_ATTR ws2812_write(uint8_t *pixels, uint32_t length) {
   // 0b00000111 => 000111 => [0]111000[1] => 10001110 => 01
   // 0b00110100 => 110100 => [0]001011[1] => 11101000 => 10
   // 0b00000100 => 000100 => [0]001000[1] => 11101110 => 11
-  uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
+  static const uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
```
devsaurus: I took the time to dig into this a bit further and found that the array contents will be placed into .rodata section in both cases. Thus no repeated initialization of the local variable on the stack - the complier is smart enough to determine that `_uartData` is never written to. So much for the good news.

The bad news is summarized in #504:
> The tradeoff is that 8bit and 16bit const data access gets a fair bit slower.

At the moment I see no other way to avoid this by mapping the array explicitly to the .data section like this:
```c
static const __attribute__((section(".data._uartData"))) uint8_t _uartData[4] = { 0b00110111, 0b00000111, 0b00110100, 0b00000100 };
```
Maybe there's a smarter approach.

Alkorin: In fact, without "static const", the array is initialized on the stack from .rodata. With "static const" the code access directly to the .rodata section. The function is then 30 bytes smaller with "static const": good point.

The question is: how slower it is ? It is hard tu measure because everything is interleaved with fifo cache. Which tradeoff is in use in NodeMCU code ? Speed or Free Memory ?

Alkorin: Ok, no comparison. It's 14x faster with _uartData in `.data` section...
Fixed

Alkorin:
Should I squash all commits ?

marcelstoer:
You could but it's actually no longer necessary. GitHub added a great new feature a few weeks back. Committers can now squash while merging if they want.

Alkorin:
Oh, nice feature :)

devsaurus:
:+1 for merge from my side.

marcelstoer:
Thank you Thomas and Arnim!

(Thomas see how GitHub allowed me to tune the final commit message in 0577c8a, really nice)

Issue #1264, opened 2016-04-25
Rotary Module documentation error


NicolSpies:
@pjsg 

- 1. Documentation for **rotary.on()** callback applies.
- 2. The following is stated:

> The first argument is the current position of the rotary switch, the second is the eventtype,..

-   I have tested the **first argument** returned with the callback to be **"eventtype"** rather than "pos" as stated and,
- the **second argument** returned with the callback to be **"pos"** rather than "eventtype" as stated

jmattsson:
Fix merged, thanks for reporting!

PR #1265, opened 2016-04-26
Fixed the documentation to match the implementation

pjsg:
As reported in #1264 the documentation had the arguments to the callback the wrong way round.

This PR fixes it!


NicolSpies:
Apologies, I missed that

PR #1273, opened 2016-04-29
wifi example in README needs a connect() call

TomK32:


marcelstoer:
See https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md, we don't accept PRs for `master`. Next dev-to-master snap is around the corner #1146.

dnc40085:
It would redundant to call `wifi.sta.connect()` after `wifi.sta.config()` since the function `wifi.sta.config()` calls [`wifi_station_connect()`](https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/wifi.c#L685) after WiFi credentials are set.

The only case it would be necessary to call `wifi.sta.connect()` is if the `auto` parameter was set to `0` (`wifi.sta.config("SSID", "password", 0)`).

marcelstoer:
@dnc40085 thanks for comment. If your docs were just a little more detailed we wouldn't see such "PRs".

I'll fix http://nodemcu.readthedocs.io/en/dev/en/modules/wifi/#wifistaconnect.

PR #1282, opened 2016-05-07
Update timer interval limits

dnc40085:
This PR fixes typos in the documentation for `tmr.alarm()`, `tmr.interval()` and `tmr.register()`.

This was brought to my attention by [a thread on esp8266.com](http://www.esp8266.com/viewtopic.php?f=23&t=9803).

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -60,6 +60,15 @@ tmr.softwd(int)
 #define TIMER_MODE_AUTO 1
 #define TIMER_IDLE_FLAG (1<<7) 
 
+#define STRINGIFY_VAL(x) #x
+#define STRINGIFY(x) STRINGIFY_VAL(x)
+
+// assuming system_timer_reinit() has *not* been called
+#define MAX_TIMEOUT_DEF 6870947  //SDK 1.5.3 limit (0x68D7A3)
+
+static const uint32 MAX_TIMEOUT=MAX_TIMEOUT_DEF;
+static const char* MAX_TIMEOUT_ERR_STR = "Range: 1-"STRINGIFY(MAX_TIMEOUT_DEF);
+
 typedef struct{
 	os_timer_t os;
 	sint32_t lua_ref;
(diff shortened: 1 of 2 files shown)


marcelstoer:
Actually, since the docs pages are tied to a specific branch those SDK version references should be removed IMO.

dnc40085:
That makes sense, I'll change it and push the new version later today

marcelstoer:
Also, our users would probably appreciate if they didn't have to figure out how many minutes 12884901ms and 6871948ms represent 😉 
And as you're at it, for future reference I think it'd be helpful to include something along the lines of

> Internally the mS value is multiplied by 1000 to be uS. This means a 32-bit signed integer has a maximum value of 0X7FFFFFF / 1000 or 2147483

(from the forum) on that page.

jmattsson:
This change looks wrong to me. The <= is the intent here. It was only in the 1.5.1 timer patch that the larger values stopped causing Strange Behaviour(tm). If someone is still seeing incorrect timer behaviour then it suggests they're not having the patch applied correctly.

dnc40085:
@jmattsson in the SDK programming guides, all releases SDK 1.5 and earlier (<=) had an upper limit of 0xFFFFFFF, starting with SDK 1.5.1 (>=) the upper limit started decreasing, SDK 1.5.1(0xC49BA5), SDK 1.5.2 (0x68DB8B) and SDK 1.5.3 (0x68D7A3). Unless the problem is in hardware, I have no idea why Espressif keeps decreasing the upper limit rather than just fixing the problem.

jmattsson:
Interesting. When Espressif provided the initial fix (binary patch for 1.5.0), the stated new limit was 0xC49BA5 which seems to be what then carried through into 1.5.1 then. At the very least the documented maximum value has been wrong in all releases <= 1.5.0+patch, and I'd argue the implementation has also been faulty. Sounds like maybe what we should have in the docs then is the 1.5.3 value (6870947) listed as the max value? Especially if values above that still yield bad behaviour on 1.5.1.

dnc40085:
Should I also update the upper limit in source to the SDK 1.5.3 value?

jmattsson:
Might as well imo - I don't trust any of those inbetween values now considering how the SDK docs have changed. Besides, if we do it now we won't forget during the upgrade to 1.5.3!

dnc40085:
Very true, I'll be sure to add that in to this PR.

dnc40085:
I've updated the timer module with the suggested changes and I also Improved error reporting to be more descriptive

app/modules/tmr.c
```diff
@@ -60,6 +60,10 @@ tmr.softwd(int)
 #define TIMER_MODE_AUTO 1
 #define TIMER_IDLE_FLAG (1<<7) 
 
+// assuming system_timer_reinit() has *not* been called
+static const uint32_t MAX_TIMEOUT = 0x68D7A3; //SDK 1.5.3 limit
+static const char* MAX_TIMEOUT_STR = "Range: 1-6870947";
```
marcelstoer: How about deducing 6870947 in `MAX_TIMEOUT_STR` from `MAX_TIMEOUT` just above? You'd eliminate a potential pitfall down the road when the max timeout value changes (should you forget to update the error message at the same time).

dnc40085: I was gonna do that but I couldn't figure out how without adding a bunch of unnecessary code (c_sprintf), any pointers?

marcelstoer: Sorry, I am not (no longer) a C programmer but isn't `c_sprintf` all you need?

devsaurus: You could get away with [Stringification](https://gcc.gnu.org/onlinedocs/cpp/Stringification.html), eg:
```c6870947
#define MAX_TIMEOUT_DEF 6870947  //SDK 1.5.3 limit
static const uint32_t MAX_TIMEOUT = MAX_TIMEOUT_DEF;
static const char* MAX_TIMEOUT_STR = "Range: 1-" #MAX_TIMEOUT_DEF;
```

dnc40085: Thank you very much @devsaurus, I'm still trying to figure out all the preprocessor tricks.

devsaurus:
Hm, I rechecked with u8g and ucg modules where I used this feature. Seems that stringification only works for macro parameters, not for macros themselves.
Sorry for my misleading sketch. You'd need to set up a macro and parametrize it with the numerical constant, I assume. Adds some overhead to the source, though. But I don't know any other technique to solve this challenge statically at compile time.

dnc40085:
@devsaurus I figured it out, thank you for pointing me in the right direction.

Here is what worked for me:
```C
#define STRINGIFY_VAL(x) #x
#define STRINGIFY(x) STRINGIFY_VAL(x)
#define MAX_TIMEOUT_DEF 6870947  //SDK 1.5.3 limit (0x68D7A3)
static const uint32 MAX_TIMEOUT=MAX_TIMEOUT_DEF;
static const char* MAX_TIMEOUT_STR = "Range: 1-"STRINGIFY(MAX_TIMEOUT_DEF);
```

marcelstoer:
> Is there anything else that needs to be changed?

No, not from my side, there's nothing that _needs_ to be changed.

However, I find the variable name `MAX_TIMEOUT_STR` a bit misleading. The time out _string_ would be `"6870947"`. In reality "Range: 1-6870947" is something like an error message. But as I said it's tiny detail.

dnc40085:
I changed `MAX_TIMEOUT_STR` to `MAX_TIMEOUT_ERR_STR`

Issue #1071, opened 2016-02-21
Potential PR: Add option to change esp init data byte 107 in lua


dnc40085:
Would it be useful to have the ability to change byte 107 of esp_init_data after firmware is burnt?
If so, where should it go, in `adc` module or `node` module?

It doesn't have to be limited to just byte 107, other bytes could be changed as well.

ABonner:
I'm assuming this is to assist with the ADC setting per the discussion in #911?

dnc40085:
Yes, that and other issues involving byte 107.

dnc40085:
@jmattsson I was looking around for issues having to do with the `adc` module/byte 107 and I ran across [your post](https://github.com/nodemcu/nodemcu-firmware/issues/1007#issuecomment-182664796) in issue #1007 suggesting the same PR. 
I was thinking that since there are a few more parameters in the init data [here is a spreadsheet](https://gist.github.com/dnc40085/a758f28f5d1bda1606ee/raw/a2d82a1fae73fed134036b6ec7592c76883e0d60/ESP8266_RF_init.xls) including an option to adjust the max tx power for the different MCS data rates and a low power mode which allows further attenuation of tx power.

What are your thoughts on which direction this should go?

jmattsson:
As I mentioned in that post, I'm not overly keen on having to expose the init data bytes into Lua in this manner in the first place. They're (very) advanced settings which shouldn't be twiddled under normal circumstances, and for the advanced use cases there is already the option of flashing a custom init data block. Documentation for the various bytes being in short supply doesn't further endear me to opening up this can of worms.

I think the ADC control byte is probably warranted to expose, and I'd say it should be a function on the `adc` module. Internally, the (re)writing of the init block should of course use a common API (extend the flash_api as/if needed, I think).

If, and it's a substantial if, we want to also expose WiFi tx power modes, I think they could sit well in an optional `wifi.powerctl` sub-module. It would need to be *really well documented* though for it to actually be useful. That would be the hardest part I'd say - radio performance is a tricky topic at the best of days, and looking at the spreadsheet does not make things very clear to me.

dnc40085:
I agree, this is not something to be taken lightly as there is very little documentation concerning the init data.

If it is decided that this should be implemented, I was planning to use Espressif's [Flash download tool](http://bbs.espressif.com/viewtopic.php?f=5&t=433) to figure out how the byes affected one another so it could be correctly implemented.

[...]

Issue #1009, opened 2016-02-03
Thoroughly document how NodeMCU handles esp_init_data_defaults


jmattsson:
This topic comes up with surprising frequency, so it could do with better explanations in the docs.

marcelstoer:
Do you plan to expand http://nodemcu.readthedocs.org/en/dev/en/flash/#upgrading-from-sdk-09x-firmware?

jfollas:
As a reader, part of my confusion comes from the fact that init_data exists in flash_api.c, so I'm not sure when I need to flash a .bin vs. allowing the ESP to write the init data itself.

marcelstoer:
Superseded by #1285.

PR #1285, opened 2016-05-10
Reworked ADC mode setting approach.

jmattsson:
First cut of moving the ADC mode control into Lua land, as discussed in #1071.

In theory this would also render #1009 moot.


pjsg:
I am not competent to evaluate this.... Sorry

marcelstoer:
Speaking with the nodemcu-build.com-maintainer hat on I'm looking forward to this. ADC config support has been the most requested feature the past few months (2/month maybe).

app/modules/adc.c
```diff
@@ -24,10 +24,49 @@ static int adc_readvdd33( lua_State* L )
...
+  lua_pushboolean (L, true);
+  return 1;
+}
+
 // Module function map
 static const LUA_REG_TYPE adc_map[] = {
   { LSTRKEY( "read" ),      LFUNCVAL( adc_sample ) },
-  { LSTRKEY( "readvdd33" ), LFUNCVAL( adc_readvdd33) },
+  { LSTRKEY( "readvdd33" ), LFUNCVAL( adc_readvdd33 ) },
+  { LSTRKEY( "force_init_byte107" ), LFUNCVAL( adc_init107 ) },
```
marcelstoer: Shouldn't the fact that byte 107 is affected be an implementation detail? How about something like `set_mode` or `init_mode`?

jmattsson: I umm'ed and ahh'ed about that a lot, but in the end I felt that trying to hide the fact that it's an init_data setting was counter productive. If anyone is using custom init_data blocks for RF tuning, I'd prefer them to be easily aware that they're mucking with it when/if they're using this ADC function. I could be convinced otherwise, but that was my reasoning for going with that name.

marcelstoer: I see your point, haven't thought of that, but I still think that detail shouldn't leak into Lua land. It's IMHO something for the docs.

> If anyone is using custom init_data blocks for RF tuning, I'd prefer them to be easily aware that they're mucking with it when/if they're using this ADC function.

I have no idea what "RF tuning" is or what RF stands for (radio frequency?) but how many use cases or users are we talking about here? One in a thousand?

As I know your reasoning now I'm undecided as well. Is it better to leak an implementation detail into the API or into the docs? Dunno. How likely is it that it's not going to be byte 107 anymore in the future?

jmattsson: Yep, RF = Radio Freq. And I honestly have no idea about number of users. Could be many, could be zero. We know Espressif keeps tweaking these numbers, so I consider it highly possible some users (or companies) have their own recommended settings.

I can't see it ever changing from byte 107 on the ESP8266 given how they've structured the init data. On the ESP32 the ADC setup is completely different as far as I can tell, so once we get that far I expect this function will be ifdef'd out (or turned into a no-op, maybe).

If you reckon the function should be a bit more agnostically named, what's your preferred option?

NicolSpies:
May I add that that the feature to make it possible to configure the ADC in a structured way may create a way to gain access to one or two other features of the system that can be very handy e.g. it is already possible to set the RF power output of the chip. If this is exposed in Lua and used with the RSSI it will be possible to reduce the output power to the minimum workable level and so reduce battery or power consumption in addition to the sleep modes. There may be other similar parameters that will make sense to expose.

app/modules/adc.c
marcelstoer: > If you reckon the function should be a bit more agnostically named, what's your preferred option?

I revisited my initial comment and `set_mode` or `init_mode` still sound fine to me.

jmattsson: In that case, can I counter-propose `force_init_mode`? I really want the `force` in there to make clear that this is not a simple change (and in fact needs a reboot afterwards) and also to back up the logic of the return value - `true` if the value had to be forcibly change, `false` if not. Without the `force` the sensible way would be to have those return values swapped, but that then would force the user to always add a `not` for the do-I-need-to-reboot test, which is just unfriendly.

marcelstoer: I actually did think about that for a few seconds but I dismissed it because (in contrast to you) I didn't find a sensible and logical explanation for it 😉 Your reasoning makes a lot of sense, go for it.

jmattsson: Okay, the name has been changed. The changes were dry-coded remotely though, so please double-check I didn't miss something.

marcelstoer:
I'm testing your branch on a [WeMos D1 mini](http://www.wemos.cc/Products/d1_mini.html).

```lua
adc.force_init_mode(adc.INIT_ADC)
-- restart
print(adc.readvdd33()) --> 65535 as expected
print(adc.read(0)) --> either 1 or 2, no sensor connected to A0
adc.force_init_mode(adc.INIT_VDD33)
-- restart
print(adc.readvdd33()) --> 2991 as expected
print(adc.read(0)) --> 65535, ok I guess even though not documented explicitly
```

docs/en/modules/adc.md
```diff
@@ -5,9 +5,37 @@
...
+####Example
+```lua
+-- in you init.lua:
+if adc.force_init_mode(adc.INIT_VDD33)
+then
+  node.restart()
+  return -- don't bother continuing, the restart is scheduled
+end
+
+print("System voltage (mV):", adc.read(0))
```
marcelstoer: Shouldn't this be `adc.readvdd33()` here?

docs/en/modules/adc.md
```diff
@@ -5,9 +5,37 @@
 
 The ADC module provides access to the in-built ADC.
 
-On the ESP8266 there is only a single-channel, which is multiplexed with the battery voltage. Depending on the setting in the "esp init data" (byte 107) one can either use the ADC to read an external voltage, or to read the system voltage, but not both.
+On the ESP8266 there is only a single-channel, which is multiplexed with the battery voltage. Depending on the setting in the "esp init data" (byte 107) one can either use the ADC to read an external voltage, or to read the system voltage (vdd33), but not both.
 
-The default setting in the NodeMCU firmware can be controlled via user_config.h at compile time, by defining one of ESP_INIT_DATA_ENABLE_READVDD33, ESP_INIT_DATA_ENABLE_READADC or ESP_INIT_DATA_FIXED_VDD33_VALUE. To change the setting at a later date, use Espressif's flash download tool to create a new init data block.
+Which mode to use the ADC in can be configured via the `adc.force_init_mode()` function. Note that after switching from one to the other a system restart is required before the change takes effect.
```
marcelstoer: Should we mention here that pushing the 'Reset' button in ESPlorer (soft restart) is not enough?

docs/en/modules/adc.md
```diff
@@ -5,9 +5,37 @@
```
marcelstoer: How about you as the maintainer of this module?

docs/en/modules/adc.md
NicolSpies: Note that after switching from one to the other a system restart by power cycling is required before the change takes effect.

docs/en/modules/adc.md
```diff
@@ -5,9 +5,37 @@
...
+Checks and if necessary reconfigures the ADC mode setting in the ESP init data block.
+
+####Syntax
+`adc.force_init_mode(mode_value)`
+
+####Parameters
+`mode_value` One of `adc.INIT_ADC` or `adc.INIT_VDD33`.
+
+####Returns
+True if the function had to change the mode, false if the mode was already configured. On a true return the ESP needs to be restarted for the change to take effect.
```
NicolSpies: True if the function had to change the mode, false if the mode was already configured. On a true return the ESP needs to be restarted by power cycling for the change to take effect.

docs/en/modules/adc.md
marcelstoer: To me (non-native) the term "power cycling" can be a bit misleading because you don't necessarily need to physically power-down / power-up the device. Calling `node.restart()` is enough.

docs/en/modules/adc.md
NicolSpies: +
 +####Example
 +```lua
 +-- in you init.lua:
 +if adc.force_init_mode(adc.INIT_VDD33)
 +then
 +  node.restart()
 +  return -- don't bother continuing, the restart is scheduled
 +end
 +
 +-- after restart usage:
 +print("System voltage (mV):", adc.readvdd33())

NicolSpies: --after restart usage:
print("System voltage (mV):", adc.readvdd33())

docs/en/modules/adc.md
jmattsson: Using the reset button in ESPlorer is fine, that just issues a `node.restart()` command (followed by some uart reconfig after the reboot). For ADC config purposes, any type of reboot is fine.

docs/en/modules/adc.md
jmattsson: Yep, that should be `adc.readvdd33()`, well spotted.

docs/en/modules/adc.md
jmattsson: Yeah I suppose I've been the one working on this module the most lately...

jmattsson:
Documentation has now been amended.

docs/en/modules/adc.md
marcelstoer: > Using the reset button in ESPlorer is fine

No, didn't work for me using the latest 0.2.0-rc3. Looks like I need to investigate what _exactly_ it does. Calling `node.restart()` manually is fine.

jmattsson: Oh-ho! There's finally been an ESPlorer update then! I'm on 0.2.0-rc2 here, and that version is just issuing a `node.res[tart()]` (the `tart()` bit gets drowned out by the wrong-baud reboot chatter).

marcelstoer: Actually it's darn obvious - but with a slight twist.
In 0.2.0-rc2 that button with the red icon was labeled 'Reset' but issued `node.restart()`. Note the slight mismatch between "reset" and "restart". In 0.2.0-rc3 the button issues `node.**reset()**` which is now in line with the button label but still a surprise for those used to the old behavior.

PR #1286, opened 2016-05-10
mention wifi config persistence in documentation

brantsch:
I consider it appropriate to mention that the wifi configuration is persistent, because it might catch users by surprise (if they have not read the module source code and Espressifs SDK documentation).

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -580,7 +580,8 @@ static int wifi_station_getconfig( lua_State* L )
 /**
   * wifi.sta.config()
   * Description:
-  * 	Set current Station configuration.
+  * 	Persistently set current station configuration.
+  * 	Configuration information will be retained until changed even if device is turned off.
   * 	Note: If there are multiple APs with the same ssid, you can connect to a specific one by entering it's MAC address into the "bssid" field.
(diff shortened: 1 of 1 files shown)


dnc40085:
Sounds like a good idea. However, typically it is suggested that PRs be submitted to the dev branch as opposed to the master branch.

On a related note, I've been considering writing a pull request to make saving of the wifi configuration to flash optional, instead of just saving to flash by default.

brantsch:
Sorry for picking the wrong branch. Your idea seems better, and very easy to implement, because the corresponding functions exist in the SDK.

marcelstoer:
See https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md

dnc40085:
I think it would still be a good idea to add this detail to the documentation for `wifi.setmode`, `wifi.sta.config` and `wifi.ap.config` in the dev branch before the merge to master, since it will be the default behavior for the master branch.

marcelstoer:
> I think it would still be a good idea to add this detail to the documentation

I definitely agree. We'd be happy to see a complete PR for the `dev` branch as you suggested.

PR #1291, opened 2016-05-11
Modifications to fix open drain level issues in OW.

karrots:
This code fixes issues with using OW in non-parasitic mode. Currently, this PR does not incorporate documentation updates from #1143. Fixes issue #1130.

Port of @pstolarz work on PaulStoffregen/OneWire#8

### DIFF
--- a/app/driver/onewire.c
+++ b/app/driver/onewire.c
@@ -107,7 +107,6 @@ uint8_t onewire_reset(uint8_t pin)
 
 	noInterrupts();
 	DIRECT_WRITE_LOW(pin);
-	DIRECT_MODE_OUTPUT(pin);	// drive output low
 	interrupts();
 	delayMicroseconds(480);
 	noInterrupts();
@@ -126,19 +125,12 @@ uint8_t onewire_reset(uint8_t pin)
 static void onewire_write_bit(uint8_t pin, uint8_t v)
 {
 	if (v & 1) {
-		noInterrupts();
-		DIRECT_WRITE_LOW(pin);
-		DIRECT_MODE_OUTPUT(pin);	// drive output low
-		delayMicroseconds(10);
-		DIRECT_WRITE_HIGH(pin);	// drive output high
-		interrupts();
-		delayMicroseconds(55);
+        onewire_read_bit(pin);
 	} else {
 		noInterrupts();
 		DIRECT_WRITE_LOW(pin);
-		DIRECT_MODE_OUTPUT(pin);	// drive output low
 		delayMicroseconds(65);
-		DIRECT_WRITE_HIGH(pin);	// drive output high
+		DIRECT_MODE_INPUT(pin);	// drive output high by the pull-up
 		interrupts();
 		delayMicroseconds(5);
 	}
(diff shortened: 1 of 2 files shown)


marcelstoer:
I can't test this. @pjsg, @nickandrew?

nickandrew:
I'll be able to test it 2-3 days hence.

marcelstoer:
Can we assume this to be clean because it's a port of an "upstream" fix?

marcelstoer:
@nickandrew what's the outlook / status here? It's the last issue blocking the master drop.

karrots:
@nickandrew any progress on testing this?

marcelstoer:
Merged it even w/o the blessing of a committer with OW know-how. Should it, against all odds, turn out to be buggy we have several weeks to fix it until the next master drop.

karrots:
Thanks Marcel.

cranphin:
Looking at this after spending a day figuring out why power = 0 broke temperature reading, glad to see a fix is already in the makes!
I'm a bit confused by why onewire_read_bit is called by onewire_write_bit though?
I'm guessing it tidies up the touch support in the source this fix is copied from (touch support isn't copied btw.), by making touch_bit and write_bit use the same code, but it's likely to confuse anyone in the future, especially without any comments to explain it :)

karrots:
To indicate a 1 on the bus the uC or device on the bus lets the bus float which allows the pull-up resistor pull the bus high.

A read operation lets the bus float. I only ported what @pstolarz proposed to change in the base library so I can't guess about why. I assume the reason is code reuse.

I didn't port touch support over because I only wanted to fix an issue not add new functionality.

pstolarz:
@cranphin - thats a good question. And the answer comes from the timings on the w1 bus during the reading/writing phases. I don't want to go into the electrical details but in short...

If you look at the w1 timings at:
https://datasheets.maximintegrated.com/en/ds/DS18B20.pdf, Fig. 16
you will see that write "1" and read "1" are identical in terms of voltage levels observed on the bus. The only difference is, the read "1" phase samples the bus at 12-14 us starting from the phase beginning, while the write phase doesn't.

Additionally (in our case) - the w1 master's GPIO is switched to the INPUT state (hight impedance) to drive the bus high by the pull-up. This really means that the write "1" is really read "1" with the bus sampling discarded.
This basically answers the question why, while writing "1" on the bus we really perform the reading phase. During the w1 writing phase none of the slaves connected to the bus interfere on it (short the bus to the ground), therefore the observed voltage level timing is as for the read "1" phase.

Going further into the timing details you may find that all the read/write "0"/"1" phases may be replaced (from the w1 master perspective) by only 2 of them, that is: write "1" with sampling and write "0". These operations are defined as the bus touching.
The w1 touch operation greatly simplifies all the w1 activities. At my original patch you can see a bus touching example replacing the original read/write approach. More (Linux kernel oriented) examples in my RPi dedicated *librasp* library.

cranphin:
Nice info :) I did figure how that would work, wasn't sure about why :)
Some inline. and method comments to explain details would help a lot against at people fix-breaking it at a later time ;)
On a side note, current dev build works well for me in non parasitic mode now, with power set to '0', yay! :)

cranphin:
Now I just need to figure out why I see spurrous temporary jumps/peaks up of 0.625 Celc., that's both with and without these changes tho (and not with my own C code, so it's specific to NodeMCU), but I'll put up a issue if I can't figure it out myself :)

nickandrew:
@cranphin Why is it confusingly using the same code? ... because that's what happens in embedded programming; and particularly when one operation can be defined in terms of another :smile:

PR #1297, opened 2016-05-16
Added support for tm1829 based led strips

sebi2k1:
- Squashed into single commit
- Added module documentation
- Rebased on dev


devsaurus:
Thanks for the contribution @sebi2k1.
I don't have any TM1829 hardware available but since rework of the WS2812 driver finished recently with #1263, I'd like to review this PR along the learnings we gathered there.

#### Locking interrupts vs stability
The driver locks interrupts - will it lock them for more than 50&nbsp;µs? Doing so can impact the overall system stability as discussed in https://github.com/nodemcu/nodemcu-firmware/pull/902#issuecomment-168850770. Also see [How does the non-OS SDK structure execution](https://nodemcu.readthedocs.io/en/dev/en/extn-developer-faq/#how-does-the-non-os-sdk-structure-execution) for details regarding the timing requirements.

In case it compromises stability then please investigate for other techniques than bit-banging to implement the low-level protocol. Eg WS2812 uses UART1 on GPIO2 to decouple the CPU & system latencies from bit stream generation (introduced with #951). It's now considered compliant with the SDK guidelines.

#### Lua API
As far as I can see, this driver adopts a similar API than WS2812's previous variant. WS2812 was enhanced by a buffer sub-module (see #902 and #1263) that could also be applicable for TM1829. Do you see a change to adapt this as well (maybe in a second iteration)?

app/include/user_modules.h
```diff
@@ -58,6 +58,7 @@
 #define LUA_USE_MODULES_WIFI
 //#define LUA_USE_MODULES_WS2801
 //#define LUA_USE_MODULES_WS2812
+#define LUA_USE_MODULES_TM1829
```
marcelstoer: New modules need to be disabled by default and ordered alphabetically.

marcelstoer:
As per https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md#writing-documentation the documentation page needs to be registered in `mkdocs.yml`.

marcelstoer:
Maybe worth noting that we had a (discarded) TM1829 contribution before: #984.

sebi2k1:
@marcelstoer My initial approach was similar to #984 (adapt timings, invert levels), but it failed due to some timing issue. I've optimised the algorithm to reduce memory access during timing critical section in favor of increased register use. I could verify the timing for 80/160MHz using an DSO and on my led strip.

Regarding bit bang vs UART. One bitstream takes roughly (2.4us * 3 + 6us). So depending how much leds are in the daisy chain it may take longer than 50us. It might work to disable/enable ISR only when writing a single byte, longer gaps between each byte are fine if < 500us.

@devsaurus I already noticed the buffer implementation and I'm interested in adding support. For my current project no sophisticated animation are required, so adding buffer support is not a priority ATM. But I'll definitely look into this. Same is true for UART support.

sebi2k1:
I integrated the requested changes and optimised the timing (one byte takes now 36us). To avoid long periods of disabled IRQs, IRQs are no longer locked for the complete frame but only for a single byte. Tested and verified here.

I cannot proceed with the UART support, since on my platform none of the TXDs is usable (GPIO2/GPIO15).

mkdocs.yml
```diff
@@ -74,5 +74,6 @@ pages:
         - 'wifi': 'en/modules/wifi.md'
         - 'ws2801': 'en/modules/ws2801.md'
         - 'ws2812': 'en/modules/ws2812.md'
+        - 'tm1829': 'en/modules/tm1829.md'
```
marcelstoer: Sorry for nagging but you should really study the contributing guidelines I pointed you to. The module list here is alphabetically sorted as well.

sebi2k1: Got ya. Was reading CONTRIBUTING.md from master branch initially and missed the whole docs/module stuff in dev branch.

app/include/user_modules.h
```diff
@@ -58,6 +58,7 @@
 #define LUA_USE_MODULES_WIFI
 //#define LUA_USE_MODULES_WS2801
 //#define LUA_USE_MODULES_WS2812
+//#define LUA_USE_MODULES_TM1829
```
marcelstoer: Should be on line 53...

sebi2k1: Fixed as well.

sebi2k1:
Are there any additional changes I have to made?

Alkorin:
I'm not a big fan to loose memory and time to invert RGB <=> BRG values. It could be handled in LUA's side. Or directly handled by the developper as I did in the WS2812 driver. Anyway the memcpy is useless you could read in the input buffer and write in your new buffer.

Or as you expose only the "RGB" call, you could invert the values during streaming.

sebi2k1:
Changing the order in C is likely more efficient than doing it in an interpreted language like LUA, but you are right the memcpy is useless it is just a leftover of the ws2801 driver I used to base my tm1829 support on. I will change that.

Are there any other changes you want me to do in order merge the pull request or do you not consider merging this pull request at all (e.g. cause of bitbanging)?

marcelstoer:
> Are there any other changes you want me to do in order merge the pull request or do you not consider merging this pull request at all

I can't test this and I don't understand enough of the technical foundation (C, hardware). Once @Alkorin and @devsaurus are happy with the implementation (as they mainly drove #1263) and see merit in having this module I'd take that as a 👍 for merging. However, that should be done only after the upcoming dev-to-master snap.

devsaurus:
PR looks ok and the input from @Alkorin should be considered. Although I can't check on real hardware, it's also a :+1: from me for a merge after updating master.

Alkorin:
I've checked the datasheets, it seems good.
From my point of vue, RGB is just a convention. I would prefer that the developper writes
`tm1829.write(5, string.char(B,R,G,B,R,G))` and avoid malloc / copy in C code

Just said in your API that write() send bytes as-is and it's the developper responsability to put bytes in the correct order. Your repsonsability is just to avoid 0xFF on the first byte to use PWM mode.

docs/en/modules/tm1829.md
```diff
@@ -0,0 +1,27 @@
...
+
+#### Parameters
+- `string` payload to be sent to one or more TM1829 leds.
+
+#### Returns
+`nil`
+
+#### Example
+```lua
+tm1829.write(5, string.char(255,0,0,255,0,0) -- turn the two first RGB leds to red using GPIO 5
```
Alkorin: Missint `)` to close the write()

sebi2k1:
Ok, will do the requested changes.

sebi2k1:
Let me reopen a clean PR.

PR #1300, opened 2016-05-18
Remove deploy section from Travis config

marcelstoer:
This is a follow-up to Johny's recent retroactive 1.4 tag/release at https://github.com/nodemcu/nodemcu-firmware/releases.

The deploy post-build section in `.travis.yml` has been there pretty much from day 1. It automatically pushed the built binaries to a GitHub release with the same name as the tag name.
As we don't want to host binaries on GitHub any longer but might still want to continue tagging master drops it's necessary to stop Travis from doing any post build actions for tags.


jmattsson:
:+1:

I was quite baffled when you pointed out there were in fact binaries attached to that release - I'd spent a few minutes editing it and *knew* there were no binaries, but Travis must have snuck 'em in right after I finished up and closed the tab. Sneaky guy!

PR #1307, opened 2016-05-21
Ensure that load_store exceptions trigger the exception handler

pjsg:
This fixes a problem where store instructions that reference the flash currently cause watchdog timeouts. This change makes them crash the system immediately and provides the correct data for node.bootreason().

The root cause was that nodemcu has its own load_store exception handler where certain exceptions can be fixed. The other exceptions ended up in a busy loop.

### DIFF
--- a/app/user/user_exceptions.c
+++ b/app/user/user_exceptions.c
@@ -72,7 +72,8 @@ void load_non_32_wide_handler (struct exception_frame *ef, uint32_t cause)
   else
   {
 die:
-    /* Turns out we couldn't fix this, trigger a system break instead
+    /* Turns out we couldn't fix this, so try and chain to the handler
+     * that was set by the SDK. If none then trigger a system break instead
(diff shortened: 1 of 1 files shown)


app/user/user_exceptions.c
```diff
@@ -73,6 +75,9 @@ void load_non_32_wide_handler (struct exception_frame *ef, uint32_t cause)
     /* Turns out we couldn't fix this, trigger a system break instead
      * and hang if the break doesn't get handled. This is effectively
      * what would happen if the default handler was installed. */
+    if (load_store_handler) {
```
jmattsson: Before I merge this, could I suggest you either update the comment to mention the chaining, or move the chaining above the comment?

pjsg: yes, I'll do that. -- good point

pjsg:
I fixed the comment......

PR #1313, opened 2016-05-25
EUS: add assigned IP address

FrankX0:
This change shows the assigned IP address when connected to the AP using EUS.

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -819,7 +819,7 @@ static void serve_status(struct tcp_pcb *conn)
 
 
         int state_len = c_strlen(s);
-	int ip_len = c_strlen(ip_addr);
+        int ip_len = c_strlen(ip_addr);
         int ssid_len = c_strlen(config.ssid);
         int status_len = state_len + ssid_len + ip_len + 1;
         char status_buf[status_len];
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -810,7 +810,7 @@ static void serve_status(struct tcp_pcb *conn)
 
 	wifi_get_ip_info(STATION_IF , &ip_info);
 
-	char ip_addr[15];
+	char ip_addr[16];
 	ip_addr[0] = '\0';
 	if (curr_state == STATION_GOT_IP)
 	{
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -787,7 +787,7 @@ static void serve_status(struct tcp_pcb *conn)
     "Failed to connect to \"%s\" - Wrong password.",
     "Failed to connect to \"%s\" - Network not found.",
     "Failed to connect.",
-    "Connected to \"%s\"."
+    "Connected to \"%s\" (%s)."
   };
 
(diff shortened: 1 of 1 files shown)


marcelstoer:
I also had a look at your feature request because I thought it might be an easy entry into NodeMCU firmware coding for me 😉
I'm not really qualified to judge because C is only a very vague memory for me. However, I believe you need a separate `case STATION_GOT_IP:` as it's the only case in which an IP address is available. If you try to smuggle it into the existing structure it'll become messy (as your code IMO shows). My approach would be this:

```c
case STATION_CONNECTING:
case STATION_WRONG_PASSWORD:
case STATION_NO_AP_FOUND:
{
  // call shared function
  enduser_setup_http_serve_header(conn, buf, output_len);
}
break;

case STATION_GOT_IP:
{
  // some specific IP handling
  // call shared function
  enduser_setup_http_serve_header(conn, buf, output_len);
}
break;
```

app/modules/enduser_setup.c
```diff
@@ -1525,4 +1538,4 @@ static const LUA_REG_TYPE enduser_setup_map[] = {
   { LNILKEY, LNILVAL}
 };
 
-NODEMCU_MODULE(ENDUSER_SETUP, "enduser_setup", enduser_setup_map, NULL);
+NODEMCU_MODULE(ENDUSER_SETUP, "enduser_setup", enduser_setup_map, NULL);
```
robertfoss: I'm not sure whats happening on this line, but I doubt that the change is needed or intended.

app/modules/enduser_setup.c
```diff
@@ -806,12 +806,25 @@ static void serve_status(struct tcp_pcb *conn)
...
+	char ip_addr[15];
+	ip_addr[0] = '\0';
+	if (curr_state == STATION_GOT_IP)
+	{
+		c_sprintf (ip_addr, "%d.%d.%d.%d", IP2STR(&ip_info.ip.addr));
+	}
+
+
         int state_len = c_strlen(s);
+	int ip_len = c_strlen(ip_addr);
```
robertfoss: Is the indentation right on this line?

app/modules/enduser_setup.c
marcelstoer: GitHub indicates that @FrankX0 replaced a line ending with newline by one without newline - and it doesn't like that (neither do I).

robertfoss:
Thanks for the PR!

I left some nitpicky feedback for you, but overall the patch looks good to me.

robertfoss:
Looks go to me.

app/modules/enduser_setup.c
```diff
@@ -806,12 +806,25 @@ static void serve_status(struct tcp_pcb *conn)
         wifi_station_get_config(&config);
         config.ssid[31] = '\0';
 
+	struct ip_info ip_info;
+
+	wifi_get_ip_info(STATION_IF , &ip_info);
+
+	char ip_addr[15];
```
jmattsson: This needs to be \[16\] to safely fit the \\0.

FrankX0:
Thanks for all your feedback so far!

Issue #1259, opened 2016-04-20
ESP_ and NODE- hostnames (if no hostname is set)


joysfera:
Currently the NodeMCU client in STA mode sends two different hostnames to DHCP server - on the very first connection to an AP it sends "ESP_XXXXXX" while on second and subsequent connections it sends "NODE-XXXXX" (where the XXXXX are last five/six digits of its MAC address). This is weird and confusing for the users. It would be better if it always sent the latter - "NODE-XXXXX" - especially when the underscore is not allowed character in a hostname for a DHCP, as I've heard.

The "NODE-XXXXX" comes from #878 and @dnc40085 writes there that he sets it in luaopen_wifi but for some reason the "ESP_XXXXXX" gets in for the first connection.

marcelstoer:
Thanks for the detailed analysis.

dnc40085:
Maybe the code needs to be moved to `user_main.c` so the hostname can be configured earlier

TerryE:
You can always set `WIFI_STA_HOSTNAME` in your build, but IIRC the underlying issue is that the Espressif code sets the DHCP hostname during its startup to the format with the _ in it and this isn't valid under the RFC and so isn't C accepted by some router firmware versions. Moving the code won't really help, IMO

marcelstoer:
In #875 Arnim suggest to file a bug with Espressif due to the use of the _ but I didn't find one. So, it likely wasn't reported yet.

devsaurus:
Last time when I visited this topic I found that `wifi_station_set_hostname` is of volatile nature. Meaning the configured hostname is lost during reset or power-on. If wifi is set to autoconnect then the default name was applied before `wifi_station_set_hostname` became effective. That would explain your observation @joysfera.

As a short term workaround you could try to _disable_ autoconnect once and configure & connect wifi from within your script each time the firmware starts up.

On the long term, Espressif should be notified that the DHCP hostname feature has room for improvement with respect to 2 related issues:
- default hostname is not compliant as discussed in #875 
- it would be beneficial if the hostname provided with `wifi_station_set_hostname` was stored in the non-volatile config area together with SSID, WPA keys, etc.

Any volunteers?

joysfera:
[...]

PR #1317, opened 2016-05-26
Move call to set the default wifi hostname to user_main.c (fix for #1259)

dnc40085:
This PR moves the call to set the default WiFi hostname from `luaopen_wifi (wifi.c)` to `user_rf_pre_init
(user_main.c)` as discussed in issue #1259.

This change will add ~440 us to boot time.


jmattsson:
I guess this is the right thing to do, even if it's not as clean structurally.

PR #1322, opened 2016-05-29
With this small patch, the DNS transaction ID space increases from 250 to 65536 states

ahupowerdns:
The ID space was artificially limited to only 250 values, making spoofing trivial. I tested this change and it works for me, and I checked the DNS random and id recovery logic and that works too.


ahupowerdns:
oops against wrong branch, apologies.

PR #1323, opened 2016-05-29
With this small patch, the DNS transaction ID space increases from 250 to 65536 states

ahupowerdns:
The ID space was artificially limited to only 250 values, making spoofing trivial. I tested this change and it works for me, and I checked the DNS random and id recovery logic and that works too.

### DIFF
--- a/app/lwip/core/dns.c
+++ b/app/lwip/core/dns.c
@@ -570,7 +570,7 @@ dns_send(u8_t numdns, const char* name, u8_t id)
   char *query, *nptr;
   const char *pHostname;
   u8_t n;
-  dns_random = os_random();
+  dns_random = (u16_t)os_random();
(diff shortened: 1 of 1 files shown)


app/lwip/core/dns.c
```diff
@@ -570,7 +570,7 @@ dns_send(u8_t numdns, const char* name, u8_t id)
   char *query, *nptr;
   const char *pHostname;
   u8_t n;
-  dns_random = os_random()%250;
+  dns_random = os_random();
```
dnc40085: since `os_random()` returns an unsigned long, shouldn't `os_random()` be cast to `u16_t`?

ahupowerdns: C does that for us automatically, but I added a cast which basically shows we know what we are doing.

jmattsson:
Sounds good. Have you tried getting this into the upstream SDK/LwIP as well? I fear we might lose this patch next time a new SDK+LwIP gets pulled in.

PR #1324, opened 2016-05-29
with this small fix, the DNS engine is RFC 5452 compliant because it …

ahupowerdns:
…now picks a random source port. This patch also moves the local UDP port range definition to udp.h from udp.c, which is probably a better place for it.

The original code did not deal with udp binding failures, this new code does one random port attempt and then sets port to 0 to get 'OS default' behaviour, and ASSERTs that that has to work.

### DIFF
--- a/app/include/lwip/udp.h
+++ b/app/include/lwip/udp.h
@@ -156,6 +156,11 @@ void             udp_input      (struct pbuf *p, struct netif *inp)ICACHE_FLASH_
 
 #define udp_init() /* Compatibility define, not init needed. */
 
+#ifndef UDP_LOCAL_PORT_RANGE_START
+#define UDP_LOCAL_PORT_RANGE_START 4096
+#define UDP_LOCAL_PORT_RANGE_END   0x7fff
+#endif
+
 #if UDP_DEBUG
 void udp_debug_print(struct udp_hdr *udphdr)ICACHE_FLASH_ATTR;
 #else
--- a/app/lwip/core/dns.c
+++ b/app/lwip/core/dns.c
@@ -254,8 +254,14 @@ dns_init()
       LWIP_ASSERT("For implicit initialization to work, DNS_STATE_UNUSED needs to be 0",
         DNS_STATE_UNUSED == 0);
 
-      /* initialize DNS client */
-      udp_bind(dns_pcb, IP_ADDR_ANY, 0);
+      /* initialize DNS client, try to get RFC 5452 random source port */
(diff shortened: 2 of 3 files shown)


jmattsson:
My concern here is that we're likely to lose this patch next time the SDK/lwIP gets upgraded. Anyone have any good ideas on how to manage that? Back when I had to patch something in lwIP I managed to get that to cause a linker error on upgrade if the patch wasn't carried across, but that's not exactly a sustainable approach.

ahupowerdns:
Hi @jmattsson - in general in open source projects, there are two solutions for this. One is trying to upstream the fix, which in this case means trying to get the SDK/lwIP people to merge it. The other, and you may have no choice, is to maintain an explicit patch-set against the upstream. This is how Linux distributions ship the Linux kernel. If you re-merge from SDK/lwIP, you apply the patch-set again and verify that it all works correctly. I can help with both approaches, please let me know!

karrots:
Is this an issue in 1.5.4 SDK? It mentions DNS optimization in the release notes.

djphoenix:
As I see in example of open LWIP (see #1415), this was already fixed by Espressif.

marcelstoer:
@djphoenix I don't think so. If you compare `dns.c` 

- in your 2.0 branch https://github.com/djphoenix/nodemcu-firmware/blob/b53608689df311a10238cb4e66264f7408c1c3e8/app/lwip/core/dns.c#L258
- with this PR branch https://github.com/ahupowerdns/nodemcu-firmware/blob/350741ffce8d256a1aa3c7b2e899002883660c6c/app/lwip/core/dns.c#L258

then you see that the UDP bind port remained untouched in 2.0.

@ahupowerdns I see @jmattsson's concerns but I'm not sure how to proceed. Being RFC 5452 compliant would certainly be nice.

djphoenix:
@marcelstoer my miss, sorry. So I vote up for this PR.

marcelstoer:
>  I managed to get that to cause a linker error on upgrade if the patch wasn't carried across, but that's not exactly a sustainable approach.

Johny, true but better than nothing at all. I was thinking that we should at least fence the code block in `dns.c` with something like 

```
/* PR #1324 */
new code
/* END PR #1324 */
```

once the PR has landed. Again, not ideal better than nothing...

marcelstoer:
Sigh...again no one noticed that this PR was for `master` rather than `dev`. I reverted my merge.

ahupowerdns:
apologies! Should I redo it for dev? Dev is where the new things happen I understand?

marcelstoer:
I just did that while you wrote the comment, no worries. In hind sight though it was the wrong decision. I usually let contributors redo their PRs.

karrots:
Slightly OT but it may be worth changing the default branch to dev so PRs automatically go to the correct location.

PR #1327, opened 2016-05-31
Fix gpio.serout (issue #428 and other minor bugs)

oyooyo:
gpio.serout does not work right now (issue #428). This pull request should provide a bugfixed and slightly improved version of gpio.serout and the gpio.serout documentation.


oyooyo:
Obsolete since PR #1359

PR #1332, opened 2016-06-01
Add releases section

marcelstoer:
I don't want to wait for #1291 any longer and proceed with the master drop. However, I wanted to document a summary of our release and merge policy in README before that.


devsaurus:
:+1:

PR #1333, opened 2016-06-01
Add SSL/TLS remarks to docs

### LINKED ISSUE #1325: add TLS cipherlist to documentation
please document the supported ciphers for TLS as many of them (especially RC4) are disabled in current distributions.

http.get "https://www.howsmyssl.com/a/check" shows:
TLS 1.1

TLS_RSA_WITH_AES_128_CBC_SHA
TLS_RSA_WITH_AES_256_CBC_SHA
TLS_RSA_WITH_RC4_128_SHA
TLS_RSA_WITH_RC4_128_MD5

marcelstoer:
fixes #1325

### DIFF
--- a/docs/en/modules/http.md
+++ b/docs/en/modules/http.md
@@ -19,6 +19,10 @@ For each operation it is also possible to include custom headers. Note that foll
 
 The `Host` header is taken from the URL itself, the `Connection` is always set to `close`, and the `User-Agent` is `ESP8266`.
 
+**SSL/TLS support**
+
+Take note of constraints documented in the [net module](net.md). 
+
 ## http.delete()
 
 Executes a HTTP DELETE request. Note that concurrent requests are not supported.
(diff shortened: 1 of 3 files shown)


devsaurus:
Why maintain 3 copies of the same information?

marcelstoer:
We currently don't have a good place to document such cross-cutting concerns. Where would you place that?

hex2a:
It would be better to document and maintain the cipher list in one place. (maybe in the net module documentation, as you should set the certificate with net.cert.verify)

devsaurus:
Like @hex2a proposed, net would be most intuitive also in my opinion. The other modules could briefly mention the list in net.

> if more modules are "affected".

I'm not aware of any - and a quick grep for `espconn_secure` didn't bring up further instances.

marcelstoer:
Fine for me. I updated `http.md` and `mqtt.md` to point to `net.md` instead and force-pushed the changes.

PR #1337, opened 2016-06-04
Fixed a typo in the wifi documentation

kbeckmann:
I found a typo in the docs and fixed it.

Changed wifi.getpymode() into the correct wifi.getphymode().


marcelstoer:
Thanks.

PR #1343, opened 2016-06-06
added fadeIn, faceOut and shift to ws2812 module

### LINKED ISSUE #1342: Extend ws2812 module with shift feature
### Missing feature
The ws2812 module was recently updated and the buffer concept was introduced. It also added a fade() method to fade out the buffer. As another extension I have implemented a shift method, which allows you to circulate the buffer forward (positive parameter) and backwards (negative parameter). This way you can quickly achieve animations, or move the buffer and overwrite the new first/last pixel afterwards.   
I have further extended the module to support fadeIn and fadeOut (fade method kept for backwards compatibility). I can provide a pull request for this, if this is seen as useful.

### Justification
fadeIn completes the already provided fadeOut. It will of course only work if the initial set of bytes allows to scale up, but nice animations will still be possible. 
The shift method is very useful to perform this task on e.g. LED circular modules or to do any other LED chasing features.

### Workarounds
This can be done in LUA, but it is runtime critical and not as memory efficient. For fast animations, it is better to have this in native code.

skycoders:
Fixes #1342.

This PR adds a shift (forward and backward) to the ws2812 buffer. It simply circulates the data like in a ring. It also adds a fadeIn (and for consistency fadeOut) method. The fade method has been kept for backward compatibility.

Committers supporting this PR: @marcelstoer

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -12,6 +12,8 @@
 
 #define FADE_IN  1
 #define FADE_OUT 0
+#define SHIFT_LOGICAL  0
+#define SHIFT_CIRCULAR 1
 
 
 typedef struct {
@@ -183,9 +185,10 @@ static int ws2812_buffer_fade(lua_State* L) {
 }
 
 
-static int ws2812_buffer_rotate(lua_State* L) {
+static int ws2812_buffer_shift(lua_State* L) {
   ws2812_buffer * buffer = (ws2812_buffer*)lua_touserdata(L, 1);
   const int shiftValue = luaL_checkinteger(L, 2);
+  const unsigned shift_type = luaL_optinteger( L, 3, SHIFT_LOGICAL );
 
   luaL_argcheck(L, buffer && buffer->canary == CANARY_VALUE, 1, "ws2812.buffer expected");
   luaL_argcheck(L, shiftValue > 0-buffer->size && shiftValue < buffer->size, 2, "shifting more elements than buffer size");
@@ -212,7 +215,14 @@ static int ws2812_buffer_rotate(lua_State* L) {
     // Move pixels to end
     os_memmove(&buffer->values[shift*buffer->colorsPerLed], &buffer->values[0], remaining_len);
     // Fill beginning with temp data
-    c_memcpy(&buffer->values[0], tmp_pixels, shift_len);
+    if (shift_type == SHIFT_LOGICAL)
+    {
+      c_memset(&buffer->values[0], 0, shift_len);
+    }
+    else
+    {
+      c_memcpy(&buffer->values[0], tmp_pixels, shift_len);
+    }
   }
   else
   {
@@ -221,7 +231,14 @@ static int ws2812_buffer_rotate(lua_State* L) {
     // Move pixels to end
     os_memmove(&buffer->values[0], &buffer->values[shift*buffer->colorsPerLed], remaining_len);
     // Fill beginning with temp data
-    c_memcpy(&buffer->values[(buffer->size-shift)*buffer->colorsPerLed], tmp_pixels, shift_len);
+    if (shift_type == SHIFT_LOGICAL)
+    {
+      c_memset(&buffer->values[(buffer->size-shift)*buffer->colorsPerLed], 0, shift_len);
+    }
+    else
+    {
+      c_memcpy(&buffer->values[(buffer->size-shift)*buffer->colorsPerLed], tmp_pixels, shift_len);
+    }
   }
   // Free memory
   luaM_free(L, tmp_pixels);
@@ -319,7 +336,7 @@ static int ws2812_buffer_write(lua_State* L) {
 static const LUA_REG_TYPE ws2812_buffer_map[] =
 {
   { LSTRKEY( "fade" ),    LFUNCVAL( ws2812_buffer_fade )},
-  { LSTRKEY( "rotate" ),  LFUNCVAL( ws2812_buffer_rotate )},
+  { LSTRKEY( "shift" ),  LFUNCVAL( ws2812_buffer_shift )},
   { LSTRKEY( "fill" ),    LFUNCVAL( ws2812_buffer_fill )},
   { LSTRKEY( "get" ),     LFUNCVAL( ws2812_buffer_get )},
   { LSTRKEY( "set" ),     LFUNCVAL( ws2812_buffer_set )},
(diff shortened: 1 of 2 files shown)


marcelstoer:
Well done. Someone familiar with the rest of the code base and with hardware to test this needs to verify and merge it, though. @devsaurus, @Alkorin one for you?

skycoders:
One question to the experienced ones here: I have thought some more about what I have implemented, and gave it another try with c_memcpy and os_memmove after including osapi.h to get it compiling, as c_memmove does not seem to exist. 
This does clean up the code, but after all the research I did, I could not find a clear hint about what memmove does with overlapping arrays. IF it copies everything into a buffer, this would be inefficient compared to my current code which one-by-one moves the pixels. But IF it treats the overlap similarly efficient, it would be better to use memmove, as all the loops are gone. 
What would be the preferred one from your view? I am happy to update the pull request.

app/modules/ws2812.c
```diff
@@ -165,6 +165,110 @@ static int ws2812_buffer_fade(lua_State* L) {
...
+    
+    
+  // check if we want to shift at all
+  if (shift == 0)
+  {
+    return 0;
+  }
+  
+
+  int * tmp_pixels = luaM_malloc(L, buffer->colorsPerLed * sizeof(int) * shift);
```
Alkorin: Why int* or sizeof(int), colors are uint8_t

app/modules/ws2812.c
```diff
@@ -165,6 +165,110 @@ static int ws2812_buffer_fade(lua_State* L) {
...
+    {
+      tmp_pixels[i] = *p++;
+    }
+    // Move pixels to end
+    // start with last pixel byte and move backwards in loop
+    p = &buffer->values[buffer->size*buffer->colorsPerLed-1];
+    uint8_t * q = &buffer->values[(buffer->size-shift)*buffer->colorsPerLed-1];
+    for (i = 0; i < buffer->size-shift; i++)
+    {
+      for (j = 0; j < buffer->colorsPerLed; j++)
```
Alkorin: You could merge these two loops for (...; buffer->colorsPerLed*(buffer->size-shift); ...)

app/modules/ws2812.c
```diff
@@ -165,6 +165,110 @@ static int ws2812_buffer_fade(lua_State* L) {
...
+    // Store the values which are moved out of the array (first n pixels)
+    uint8_t * p = &buffer->values[0];
+    for(i = 0; i < shift*buffer->colorsPerLed; i++)
+    {
+      tmp_pixels[i] = *p++;
+    }
+    // Move pixels to front
+    p = &buffer->values[0];
+    uint8_t * q = &buffer->values[shift*buffer->colorsPerLed];
+    for (i = 0; i < buffer->size-shift; i++)
```
Alkorin: Same as line 222

app/modules/ws2812.c
```diff
@@ -165,6 +165,110 @@ static int ws2812_buffer_fade(lua_State* L) {
...
+      {
+        *p++ = *q;
+        q++;
+      }
+    }
+    // Fill end with temp data
+    p = &buffer->values[(buffer->size-shift)*buffer->colorsPerLed];
+    for (i = 0; i < shift; i++)
+    {
+      for (j = 0; j < buffer->colorsPerLed; j++)
```
Alkorin: Same as line 222

app/modules/ws2812.c
skycoders: I copied this out of line 129 (colors array for fill) as this pattern was used there. Happy to change this.

Alkorin:
If you want to be faster, you could try to move 4 bytes at a time as ESP8266 is a 32bits CPU, but ... you should be carefull with alignment! And code will be less clean :/

skycoders:
@Alkorin I have just tested an implementation which needs only three lines to copy memory:

    // Store the values which are moved out of the array (last n pixels)
    c_memcpy(tmp_pixels, &buffer->values[(buffer->size-shift)*buffer->colorsPerLed], shift_len);
    // Move pixels to end
    os_memmove(&buffer->values[shift*buffer->colorsPerLed], &buffer->values[0], remaining_len);
    // Fill beginning with temp data
    c_memcpy(&buffer->values[0], tmp_pixels, shift_len);

This works great, I am only uncertain regarding memory usage and have no idea how to measure this. BUT the code is really clean. And I expect the OS to be as efficient as possible when doing this. So if you think this makes sense, I will update the pull request, eliminating all loops as well :)

app/modules/ws2812.c
Alkorin: Yes, I did this because luaL_checkinteger returns int. I could have cast luaL_checkinteger to uint8_t it there and win some cycles and few memory bytes, should check the assembly :)

But here, no need of (int), uint8_t is enough as it's your input and output type

Alkorin:
If it works, good for me ! Don't need to reinvent the wheel.
If you check de bootrom dump of memmove (http://cholla.mmto.org/esp8266/bootrom/boot.txt 4000e04c), the code load multiple 32 bits at a time and write them. More code but faster code. Anyway, as usual, bench the two solutions and keep the faster one ;) But as memmove is already there... it's free !

docs/en/modules/ws2812.md
```diff
@@ -150,6 +150,55 @@ Divide each byte of each led by the given value. Useful for a fading effect
...
+ - `value` value by which to divide each byte
+
+#### Returns
+`nil`
+
+#### Example
+```lua
+buffer:fadeOut(2)
+```
+## ws2812.buffer:shift()
```
devsaurus: Technically it's more a rotate than a shift.

skycoders: you are right. Shall I rename it?

docs/en/modules/ws2812.md
```diff
@@ -150,6 +150,55 @@ Divide each byte of each led by the given value. Useful for a fading effect
 ```lua
 buffer:fade(2)
 ```
+## ws2812.buffer:fadeIn()
```
devsaurus: How about one single method `ws2812.buffer:fade(value[, direction])`? This would keep the API lean.

app/modules/ws2812.c
```diff
@@ -165,6 +165,110 @@ static int ws2812_buffer_fade(lua_State* L) {
   return 0;
 }
 
+static int ws2812_buffer_fade_in(lua_State* L) {
```
devsaurus: Duplicated code with `ws2812_buffer_fade_out()`. If you unify the API as mentioned below, then there's no need for two separate functions anymore.

skycoders:
I have just tested the loop-single-move variant currently in this PR against the memcpy/memmove variant with my 300 LEDs strip and a delay of 1ms through a timer before executing the next shift call. The new memcpy/memmove variant seems to be faster - 3.1s compared to 3.4s round-trip. As there are other things to improve (fade method), I will update this and also rename the method to rotate.

docs/en/modules/ws2812.md
devsaurus: Yes, I'd prefer [rotate](https://en.wikipedia.org/wiki/Bitwise_operation#Rotate_no_carry) to distinguish it from [logical shift](https://en.wikipedia.org/wiki/Bitwise_operation#Logical_shift).

devsaurus:
I'd like to revise my request from above
- either rename to `rotate` if you intend to stick to the current functionality
- or keep `shift` and support both circular and logical shifts selected by a parameter

skycoders:
Done. I decided for shift with parameter. Module and docs updated.

devsaurus:
Code looks good, thanks @skycoders!
I will give it a try on hardware before merging it.

PR #1353, opened 2016-06-12
Drop examples folder

### LINKED ISSUE #1351: Drop the examples folder?
I only recently looked at the content we have in the `examples` folder. Is it (still) worth keeping?

I find [`fragment.lua`](https://github.com/nodemcu/nodemcu-firmware/blob/master/examples/fragment.lua) in particular a bit odd. It appears to be a collection of unrelated, undocumented and unmaintained pieces of code for the initial set of NodeMCU modules. Some may not even work with the current firmware anymore. In its current form that file probably is of very little value to anyone.

marcelstoer:
Fixes #1351.


devsaurus:
A lot of whitespace diffs, but :+1: overall as far as I can see.

marcelstoer:
> A lot of whitespace diffs

Yep, and I'm sorry about that. It was too late when I noticed and I didn't want to reverse that.

Issue #1339, opened 2016-06-04
Hardware timer (platform_hw_timer_*) causing freeze


vsky279:
I am building a module to radio control Somfy blinds via 433Mhz RF transmitter. A sequence of 0 and 1 with different delays needs to be transmitted as illustrated on the chart:
![image](https://cloud.githubusercontent.com/assets/12664840/15799330/53e6ecfe-2a59-11e6-8871-2693314cc4c0.png)

There are implementations for Arduino using delay functions but on ESP8266 I need to go asynchronous. That is why I have decided to use the hw_timer to achieve precise timing in the order of microseconds. 

### Actual behavior
I am indeed able to control blinds but after sending few commands the ESP8266 freezes and then goes restart triggered by hw watchdog (`rst cause: 4, boot mode:(3,0)`).

### Test code
The minimum code to reproduce the issue is

`modules\somfy.c`
```C
#include "os_type.h"
#include "osapi.h"

#include "module.h"
#include "lauxlib.h"
#include "platform.h"
#include "hw_timer.h"
#include "user_interface.h"

#define SYMBOL 640 // symbol width in microseconds

#define TIMER_OWNER ((os_param_t) 's')

static uint32_t   delay[10] = {9415, 89565, 4*SYMBOL, 4*SYMBOL, 4*SYMBOL, 4550, SYMBOL, SYMBOL, SYMBOL, 30415}; // 143 ms total
static uint8_t    signalindex;

static void ICACHE_RAM_ATTR sendCommand(os_param_t p) {
	(void) p;
	//NODE_DBG("%d\n", signalindex);
	signalindex++;
	if (signalindex<10) {
		platform_hw_timer_arm_us(TIMER_OWNER, delay[signalindex-1]);
	} else {
		platform_hw_timer_close(TIMER_OWNER);
	}
}

static int somfy_lua_sendcommand(lua_State* L) {
	if (!platform_hw_timer_init(TIMER_OWNER, NMI_SOURCE, TRUE)) {
		// Failed to init the timer
		luaL_error(L, "Unable to initialize timer");
	}
	NODE_DBG("Triggering hw timer (%d)\n",  system_get_time());
	platform_hw_timer_set_func(TIMER_OWNER, sendCommand, 0);
	signalindex=0;
	sendCommand(0);
	return 0;
}

static const LUA_REG_TYPE somfy_map[] = {
[...]

PR #1359, opened 2016-06-13
gpio.serout going asynchronous

vsky279:
Fixes issues of `gpio.serout` discussed in #1339.

This PR is adding an asynchronous mode to `gpio.serout` function. The existing implementation is synchronous and therefore blocks the stack. Transmission of a signal lasting more than 15 ms was obviously causing reboots. Now the transmission can be asynchronous with delays up to the limit of hw timer and virtually unrestricted duration.
The implementation is backwards compatible. The synchronous version is to be used for signal with sub-50 µs resolution (but restricted to max. overall duration).
Whether the function goes async is determined by `callback` parameter. If present the function goes async. Please see the updated documentation for details.

Hardware timer is used so this function cannot be used simultaneously with other modules using hw timer (e.g. `pcm`).

This PR also replaces PR #1327 which should be dropped (proposed changes are included in this PR too).

### DIFF
--- a/app/modules/gpio.c
+++ b/app/modules/gpio.c
@@ -180,7 +180,7 @@ typedef struct
   uint32 repeats;
   uint32 *delay_table;
   uint32 tablelen;
-  lua_State* L;
+  task_handle_t done_taskid;
   int lua_done_ref; // callback when transmission is done
 } serout_t;
 static serout_t serout;
@@ -188,9 +188,16 @@ static const os_param_t TIMER_OWNER = 0x6770696f; // "gpio"
 
 static void seroutasync_done (task_param_t arg)
 {
-  lua_rawgeti (serout.L, LUA_REGISTRYINDEX, serout.lua_done_ref);
-  lua_call (serout.L, 0, 0);
-  luaL_unref (serout.L, LUA_REGISTRYINDEX, serout.lua_done_ref);
+  lua_State *L = lua_getstate();
+  luaM_freearray(L, serout.delay_table, serout.tablelen, uint32);
+  if (serout.lua_done_ref != LUA_REFNIL) { // we're here so serout.lua_done_ref != LUA_NOREF
+    lua_rawgeti (L, LUA_REGISTRYINDEX, serout.lua_done_ref);
+    if (lua_pcall(L, 0, 0, 0)) {
+      // Uncaught Error. Print instead of sudden reset
+      luaL_error(L, "error: %s", lua_tostring(L, -1));
+    }
+    luaL_unref (L, LUA_REGISTRYINDEX, serout.lua_done_ref);
+  }
 }
 
 static void ICACHE_RAM_ATTR seroutasync_cb(os_param_t p) {
@@ -204,16 +211,12 @@ static void ICACHE_RAM_ATTR seroutasync_cb(os_param_t p) {
     if (serout.repeats && serout.index>=serout.tablelen) {serout.index=0; serout.repeats--;}
   } else {
     platform_hw_timer_close(TIMER_OWNER);
-    luaM_freearray(serout.L, serout.delay_table, serout.tablelen, uint32);
-    if (serout.lua_done_ref != LUA_NOREF && serout.lua_done_ref != LUA_REFNIL) {
-      task_post_low (task_get_id((task_callback_t) seroutasync_done), (task_param_t)0);
-    }
+    task_post_low (serout.done_taskid, (task_param_t)0);
   }
 }
 
 static int lgpio_serout( lua_State* L )
 {
-  serout.L = L;
   serout.pin = luaL_checkinteger( L, 1 );
   serout.level = luaL_checkinteger( L, 2 );
   serout.repeats = luaL_optint( L, 4, 1 )-1;
@@ -292,6 +295,7 @@ int luaopen_gpio( lua_State *L ) {
   }
   platform_gpio_init(task_get_id(gpio_intr_callback_task));
 #endif
+  serout.done_taskid = task_get_id((task_callback_t) seroutasync_done);
   return 0;
 }
 
--- a/app/modules/gpio.c
+++ b/app/modules/gpio.c
@@ -1,4 +1,5 @@
 // Module for interfacing with GPIO
+//#define NODE_DEBUG
 
 #include "module.h"
 #include "lauxlib.h"
@@ -8,6 +9,7 @@
 #include "c_types.h"
 #include "c_string.h"
 #include "gpio.h"
+#include "hw_timer.h"
 
 #define PULLUP PLATFORM_GPIO_PULLUP
 #define FLOAT PLATFORM_GPIO_FLOAT
@@ -160,8 +162,7 @@ static int lgpio_write( lua_State* L )
 
 #define DELAY_TABLE_MAX_LEN 256
 #define delayMicroseconds os_delay_us
-// Lua: serout( pin, firstLevel, delay_table, [repeatNum] )
-// -- serout( pin, firstLevel, delay_table, [repeatNum] )
+// Lua: serout( pin, firstLevel, delay_table[, repeat_num[, callback]])
 // gpio.mode(1,gpio.OUTPUT,gpio.PULLUP)
 // gpio.serout(1,1,{30,30,60,60,30,30})  -- serial one byte, b10110010
 // gpio.serout(1,1,{30,70},8)  -- serial 30% pwm 10k, lasts 8 cycles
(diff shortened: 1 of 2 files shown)


app/modules/gpio.c
```diff
@@ -170,42 +171,95 @@ static int lgpio_write( lua_State* L )
...
+  lua_State* L;
+  int lua_done_ref; // callback when transmission is done
+} serout_t;
+static serout_t serout;
+static const os_param_t TIMER_OWNER = 0x6770696f; // "gpio"
+
+static void seroutasync_done (task_param_t arg)
+{
+  lua_rawgeti (serout.L, LUA_REGISTRYINDEX, serout.lua_done_ref);
+  lua_call (serout.L, 0, 0);
```
dvv:
```lua
if (lua_pcall(serout.L, 0, 0, 0)) {
  // Uncaught Error. Print instead of sudden reset
  error(serout.L, "error: %s", lua_tostring(serout.L, -1));
}
```

app/modules/gpio.c
```diff
@@ -170,42 +171,95 @@ static int lgpio_write( lua_State* L )
...
+  NODE_DBG("%d\t%d\t%d\t%d\t%d\t%d\t%d\n", serout.repeats, serout.index, serout.level, serout.pin, serout.tablelen, serout.delay_table[serout.index], system_get_time()); // timing is delayed for short timings when debug output is enabled
+  if (serout.index < serout.tablelen) {
+    GPIO_OUTPUT_SET(GPIO_ID_PIN(pin_num[serout.pin]), serout.level);
+    serout.level = serout.level==LOW ? HIGH : LOW;
+    platform_hw_timer_arm_us(TIMER_OWNER, serout.delay_table[serout.index]);
+    serout.index++;
+    if (serout.repeats && serout.index>=serout.tablelen) {serout.index=0; serout.repeats--;}
+  } else {
+    platform_hw_timer_close(TIMER_OWNER);
+    luaM_freearray(serout.L, serout.delay_table, serout.tablelen, uint32);
```
devsaurus: I don't know how safe (or unsafe) it is to run memory allocation routines from within an ISR.
You could free the array in `seroutasync_done()` and decide afterwards whether to call out to Lua land or not.

app/modules/gpio.c
```diff
@@ -170,42 +171,95 @@ static int lgpio_write( lua_State* L )
...
+    if (serout.lua_done_ref != LUA_NOREF && serout.lua_done_ref != LUA_REFNIL) {
+      task_post_low (task_get_id((task_callback_t) seroutasync_done), (task_param_t)0);
```
devsaurus: Same here: `task_get_id()` calls realloc. Furthermore, each invocation consumes a small chunk of heap that can't be free'd at all.
Retrieve and store the task id *once* (eg. in `luaopen_gpio()`) and just do something like
```lua
task_post_low( serout.serout_taskid, ... )
```

app/modules/gpio.c
dvv: I believe housekeeping if for `__gc`.
May be worth to make serout an object for that. Plus we could add cancellability.

app/modules/gpio.c
```diff
@@ -170,42 +171,95 @@ static int lgpio_write( lua_State* L )
...
+}
+
 static int lgpio_serout( lua_State* L )
 {
-  unsigned clocks_per_us = system_get_cpu_freq();
-  unsigned pin = luaL_checkinteger( L, 1 );
-  unsigned level = luaL_checkinteger( L, 2 );
-  unsigned repeats = luaL_optint( L, 4, 1 );
-  unsigned table_len, i, j;
+  serout.L = L;
```
devsaurus: There was [a statement](https://github.com/nodemcu/nodemcu-firmware/issues/846#issuecomment-164651561) that tasks/callbacks should not re-use local states. Instead they should get it from
```lua
```

vsky279:
Thanks for all your comments! I hope I fixed them all.
Except the "to make serout an object" - I would need more hints if there's an agreement to proceed this way.

devsaurus:
Excellent, green :traffic_light: from my side.
Anything missing @dvv?

oyooyo:
Recently created issue #1364 isn't very helpful, but apparently, it's asking for a way to send infrared signals. I'm generally interested in this too, and I originally thought this would be possible with gpio.serout() as well, but meanwhile I realized it isn't.

The problem is that unlike when sending 433MHz signals, when sending IR signals using an IR LED it's not enough to simply turn some gpio pin on for a few hundred microseconds. Instead, the signal is usually modulated onto a 36-40kHz signal. For example, in case of the popular RC5 protocol, a single "1" bit is represented by 889µs with the IR LED completely turned off, followed by 32 impulses of turning the IR LED on for about 7µs, with spaces of about 21µs in between.
See [this image](http://www.sprut.de/electronic/ir/rc5t.gif) for a simplified illustration.

I'm wondering if it makes sense to futher extend gpio.serout() for such use cases, adding another function or how else to allow for sending IR signals as well. BTW, this is also a good example of why something like the synchronous version of gpio.serout is indeed still necessary, since those tight timings of 7 or 21µs shouldn't be possible with the asynchronous version.

Since some people are also interested in the opposite direction (something like gpio.serin() or so for receiving data from 433MHz, infrared etc.), I'm also beginning to think it might really make sense to move the code to its own module (named "gpioser" or so) , as already partly discussed in #1017 .

What's your opinion on this?

dvv:
1. IR send is here: https://github.com/dvv/nodemcu-thingies/blob/master/irsend.lua
2. I tend to separate IR sender into i) real flashing with IR led (38kHz PWM with circa 1/3 duty) and ii) mimicking TSOP (that is as though you've desoldered a real TSOP from a device and control its input leg directly with a ESP pin). The former requires pure sync serout (https://github.com/dvv/nodemcu-thingies/blob/master/irsend.lua#L27), the latter operates with pulses circa 0.5ms (https://github.com/dvv/nodemcu-thingies/blob/master/irsend.lua#L43) and is a candidate for async serout.
3. I'm very pro `.serin(pin, n_edges)` provided it can report durations at least with 10us resolution.

oyooyo:
@dvv:

- thanks for the hint towards your irsend.lua . I'm a bit confused though: Was that really what #1364 was referring to? Because since your irsend.lua is apparently LUA code that probably already works perfectly fine with existing nodemcu-firmware, I don't quite understand why the issuer wants to have it added into nodemcu-firmware? Can't he simply add your irsend.lua to his project and use it?

- I'm not 100% sure if I mean exactly the same as you do, but I think I'm in favor of something at least very similar: The actual timing values passed to a gpio.serout()-like function capable of sending arbitrary infrared signals should be agnostic of any eventual underlying modulation, so they should describe the waveform that for example a TSOP1736 outputs. How the modulated signal is being generated should be specified by a different argument.
For example, I envision a gpio.serout-like function capable of sending arbitrary modulated infrared signals to have a signature somewhat similar to this:
`sendir(pin, active_level, delay_times, duty_cycle_active_time, duty_cycle_inactive_time)`
So for example, sending a single "0" bit using the RC5 protocol would translate to:
`sendir(3, 0, {889, 889}, 7, 21)`
which would semantically mean: 
_"Send an infrared signal using an infrared LED connected to gpio pin 3. The IR LED is connected as "active low", so it is turned on if the state of the GPIO pin is 0/low. Send / turn on the (modulated) IR signal for 889µs, followed by 889µs with the IR LED turned off. During the phases when the IR LED is sending, generate the modulated IR signal by repeatedly turning the IR LED on for 7µs, then turning it off for 21µs (which results in a modulated signal with a period of 28µs = approx. 35700Hz, and 25% duty cycle)."_

- What particular use case(s) do have in mind for .serin? And in particular, what use case do you see that requires 10µs resolution?

At first I thought something like .serin() was necessary for receiving and decoding infrared and 433mhz protocols, but meanwhile I wonder if this isn't already possible with the existing gpio module.

dvv:
On `.serin()` -- It's a call to leave firmware provide Lua harness and only export very necessary and well-thought set of low-level time-tight routines. All high-level stuff is better written in Lua. Examples are IR send/recv, DHT*, 1-wire, all kinds of particular I2C "drivers" etc etc. Pure IMHO.

devsaurus:
Guys, I understand your intention to draft a serial-in concept, yet I propose you take this topic to a new, dedicated issue. Otherwise it might get lost here once this PR is merged & closed (what I'm going to do in a couple of hours).

devsaurus:
Thanks a lot to all who were involved! @vsky279 for your contribution and dedication to this topic, @oyooyo for the spark to evolve the somfy use case into a general purpose feature, and @dvv for sketching out the next `gpio` extensions :smiley:

marcelstoer:
For serial-in keep #1017 in mind.

PR #1362, opened 2016-06-15
Skip passing port in Host header if standard port.

### LINKED ISSUE #1361: HTTP Host header standard ports
### Expected behavior
HTTP module sends port as portion of `Host` header  value only if using non-standard ports 
### Actual behavior
HTTP module sends host:port even when using standard ports. While technically not against [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23) it does break some virtual host environments.

Link to another project that details this issue.
- https://github.com/request/request/issues/515
- https://www.joedog.org/2014/09/05/is-a-port-number-required-in-the-http-host-header/

### Test code
The code below when captured with Wireshark or similar packet capture will show `Host: httpbin.org:80`. Nginx or httpbin's code seems to strip the port 80 off the request before displaying it back to the user.

```Lua
http.get("http://httpbin.org/headers", nil, function(code, data)
    if (code < 0) then
      print("HTTP request failed")
    else
      print(code, data)
    end
  end)
```
### NodeMCU version
dev 500bbceded18fcf0900da948e8e9642ba3dc230e

### Hardware
Bug hardware agnostic.

karrots:
Fixes #1361.

Rationale: Current HTTP module sends port in `Host` header even if using standard 80 and 443 ports. This fixes the behavior with the exception of HTTPS on port 80 and HTTP on port 443. A more thorough fix would take a bit of rework of the module.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -181,7 +181,7 @@ static void ICACHE_FLASH_ATTR http_connect_callback( void * arg )
 	HTTPCLIENT_DEBUG( "Connected\n" );
 	struct espconn	* conn	= (struct espconn *) arg;
 	request_args_t	* req	= (request_args_t *) conn->reverse;
-
+	int len;
 	espconn_regist_recvcb( conn, http_receive_callback );
 	espconn_regist_sentcb( conn, http_send_callback );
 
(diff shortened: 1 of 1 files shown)


app/http/httpclient.c
```diff
@@ -200,16 +200,29 @@ static void ICACHE_FLASH_ATTR http_connect_callback( void * arg )
...
-			      "%s %s HTTP/1.1\r\n"
-			      "Host: %s:%d\r\n"
-			      "Connection: close\r\n"
-			      "User-Agent: ESP8266\r\n"
-			      "%s"
-			      "%s"
-			      "\r\n",
-			      req->method, req->path, req->hostname, req->port, req->headers, post_headers );
 
+	if ((req->port == 80) || (req->port == 443))
```
djphoenix: I think there should be "secure" flag check, not just omit both 80s and 443s

djphoenix:
IMHO if you point to RFC, you should use all of them :)

karrots:
RFC only says leave it off if default. Never actually says the port should be there when it's not default, though. So it's a bit ambiguous.

app/http/httpclient.c
karrots: I did miss there was already a req->secure field in the structure. I'll update the code to check that.

marcelstoer:
Thanks.

PR #1379, opened 2016-07-03
Net functionality rewrite on top of LWIP

### LINKED ISSUE #1080: Net module model review and reimplementation.
The net module attempts to provide a unified model for TCP and UDP and really fails to paper over the cracks in how these are structurally different and in particular that UDP is not connection orientated.  In the current implementation:
-  You can have one TCP server and one UDP server
-  You can create multiple listeners per TCP server (number TBD) to listen on more than one port, but you will need to use bound upvals to know which socket you are receiving on since this information isn't passed as a parameter.
-  The TCP server will receive up to 5 connections and create a net.socket for each.
-  You can reply on connected sockets
-  You can (also) create multiple outbound TCP connections and send and receive on these.
-   You can create multiple listeners per UDP server (number TBD) to listen on more than one socket
-   The UDP model is different in that the listen() method ignores any connection function and the on and send methods a bound to the server and not the socket.
-   If you are receiving from multiple hosts or on multiple ports, your send will go the the last port/IP address and not necessarily the one which initiated the send.
-   If you do try any if these "permitted" multiple combinations then your ESP will regularly PANIC or die with an exception.  

IMO, this is a mess.  
1.  The model should be clearly documented and the ESP should not crash.  We should either return a success fail status or through a Lua error, but the application should remain in control.  
2.  We should not try to make UDP look as if it is connection oriented.  Better that the on receive gets an IP and port and can whether to process it or not.  we shouldn't have a createConnection for UDP but the send should just accept an IP and a port.

Thoughts, comments?

djphoenix:
Fixes #1080


djphoenix:
Anyone prefer to keep old UDP code (on top of espconn), rewrite it on LWIP, or strip out when we have new-style API?

/cc @marcelstoer @pjsg @TerryE @jmattsson

djphoenix:
Has rewrote net.udp.resolve on LWIP because old version uses UDP socket creation (WHY???)... This chaining was absolutely unneeded for UDP resolution.
Breaking change on API: callback function now takes only IP (or nil if not resolved).

djphoenix:
TCP/SSL connections proposal:
1. Move "secure" connections to `crypto` (or another, maybe new) module.
2. Strip out secure connections code from `net` module.
3. Rewrite TCP connections on LWIP (and so remove limitation to 5 connections).
4. Freeze `net` module and then begin thinking about mbedTLS (or another TLS backend).

jmattsson:
Lots of stuff here! I don't have time to code-review at the moment, but here are my views on the rest:

- Really really avoid breaking the API if at all possible. For the dns.resolve(), just pass a `nil` as the first parameter to the callback. If someone had gone against what the docs say and used the `sk` anyway, that's too bad (for them).

- I think the reason the DNS resolution was on top of espconn was that it was in the SDK docs, whereas the LWIP version is not. I agree the espconn DNS resolution is a bad idea.

- I would like to keep the existing `net` API with its `secure` flag to get a TLS connection, both for backward compatibility as well as fitting conceptually. I could (easily?) be convinced that the actual implementation might be better off in its own module (e.g. `tls`) and the `net.xyz(..., secure)` is just a forwarder. That way for basic TLS functionality one can use the `net` module, but once one wants to start doing certificate verification etc it's time to head to the `tls` module. Besides, last I checked the cloud builder stats, the SSL support wasn't all that frequently used.

- Indeed, TCP connections should be native LWIP. We need to get all the current `net` support off espconn both for our own sanity and for ESP32 support.

- I did a bit of research into TLS stacks suitable for embedding, and mbedTLS seemed a prime candidate as long as we use the Apache licensed version. We're staying clear of GPL code for this project to keep things a bit simpler. Getting the non-secure portions off espconn is the higher priority though.

jmattsson:
Oh, and :+1: for doing all this! I believe Terry has been busy with other real-world stuff lately.

djphoenix:
@jmattsson thanks for support.
1. I'll revert back first parameter for dns.resolve. But what to do with createServer/Connection? It now covers just TCP. Create a alias to createUDPSocket if first is net.UDP, or finally strip it out?
2. OK, I will move secure versions to tls module with aliases in net. Also will refactor it at all, so removed 5 socks limit and use LWIP when possible (e.g. DNS resolution in connect).
3. So, when work with net module will seems done, this PR will be landed and I will begin research about mbedTLS integration.

TerryE:
>  I believe Terry has been busy with other real-world stuff lately

And in spades. I am doing 10hr days on my self-build doing the stuff that I haven't subcontracted out but having to keep aligned to their timescales.  On top of that the self-build forum that I use all the time as a reference source cratered, because the owner / SysAdmin decided to unilaterally shut it down.  I am the only one in this community with the Linux / PHP / MySQL/ SysAdmin skills to set up a replacement site and configure everything, so doing this has been my evening work.  I've told my wife that I am really missing my girlfriend Lua! :disappointed_relieved:  

I'll rebel tonight and review Yuri's work.

marcelstoer:
I agree with Johny on most things

- definitely don't break backwards compatibility
- keep the secure flag to get TLS (RTOS provides TLS 1.2 IIRC)
- I would hate to add both a firmware-wide "SSL-support flag" _plus_ an extra module if I need encrypted connections in my firmware
- however, whatever you do with SSL/TLS keep in mind that both the HTTP and MQTT modules also (can) support encrypted connections and they shouldn't break either

Side note, we usually appreciate if we can have such discussions _before_ you start working on a PR so that you can land the PR once all issues have been resolved: https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md

djphoenix:
@marcelstoer 
- About backward compatibility - it definitely **should be** broken for UDP part, because old concept (separation to server/client) is absolutely wrong. At another side, for secure connections, compatibility was preserved. Also kept "createServer/Connection" compatibility with UDP (by chaining into createUDPSocket).
- Extra module is needed when we integrate mbedTLS, because it (definitely) adds a some functional, and I propose move it outside net module, else it will grow a lot. So, for compatibility, we will keep "chains" to new module from original code.
- About another modules - their will be rewrote anyway, but not in this PR.

jmattsson:
Automatically enabling a `tls` module when `ENABLE_SSL` (or whatever the precise macro is) is defined is trivial, I'm happy to look at that if needed.

I'm not 100% sure on the UDP side of things, other than that the espconn model is conceptually wrong. If we can both clean it up *and* keep backwards compatibility through forwarders, that would be grand imo.

djphoenix:
@jmattsson We can't keep all of backward compatibility for UDP module. Espconn layer have not a lot of issues, but was misused in old-style UDP sockets. It was described on #1080.
> We should not try to make UDP look as if it is connection oriented.

djphoenix:
// work progress report
Not have enough time right now... Stayed on rewrite of TCP connections, this is huge part which need to be entire reviewed and reimplemented. I'll commit it nearer to weekend.
Also took a look aside mbedtls (on RTOS SDK without nodemcu) just for understand how it works. So it will be interesting...

djphoenix:
OK, now all functions of old net module was rewrote on top of LWIP with full backward compatibility... with some exceptions.
- UDP sockets does not have `connection` callback on `listen` method.
- UDP sockets does not have `connect` method and needs to pass `port`/`ip` to `send` method. **TODO: backward compatibility**
- TCP server's `timeout` does not work. **Searching for solution.**
- TCP socket's `hold`/`unhold` does not work. **Searching for solution.**

jmattsson:
Great progress! I'm hoping to spend some quality time with NodeMCU next week and looking forward to taking this for a spin over on the RTOS branch too!

For UDP `connect`, I don't remember what the current behaviour is, but if we could get the normal-sockets approach of effectively setting the "connected" ip/port as the default values to use with subsequent `send` calls that would probably be neat.

Some ideas, but since I haven't looked through the code recently they may be silly:
  - From memory timeouts in espconn are driven internally by regular os_timers. If we need to use another timer to handle timeouts, that shouldn't be a problem. I *thought* LwIP had its own internal "tick" based timeouts though. Maybe search for "tick" and backtrack from there?

- Hold/unhold would be receive window related. Try hunting for "receive", "recv" etc in LWIP to see if anything comes up.

app/modules/net.c
```diff
@@ -10,1364 +10,757 @@
...
+    lua_pushvalue(L, stack++);
+    luaL_unref(L, LUA_REGISTRYINDEX, ud->client.cb_sent_ref);
+    ud->client.cb_sent_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+  }
+  if (ud->type == TYPE_UDP_SOCKET && !ud->pcb) {
+    ud->udp_pcb = udp_new();
+    if (!ud->udp_pcb)
+      return luaL_error(L, "cannot allocate PCB");
+    udp_recv(ud->udp_pcb, net_udp_recv_cb, ud);
+    addr.addr = 0;
```
jmattsson: This nukes the address passed in, making all UDP being sent to 0.0.0.0. Not quite useful ;)
Changing to this makes things work a lot better:
```
ip_addr_t laddr;
laddr.addr = 0;
err_t err = udp_bind(ud->udp_pcb, &addr, 0);
```

app/modules/net.c
```diff
@@ -10,1364 +10,757 @@
...
+    if (ud->self_ref == LUA_NOREF) {
+      lua_pushvalue(L, 1);
+      ud->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+    }
+  }
+  if (!ud->pcb || ud->self_ref == LUA_NOREF)
+    return luaL_error(L, "not connected");
+  err_t err;
+  if (ud->type == TYPE_UDP_SOCKET) {
+    struct pbuf *pb = pbuf_alloc(PBUF_TRANSPORT, datalen, PBUF_RAM);
```
jmattsson: Does `pb` need to be tested for NULL to catch a failed alloc?

app/modules/net.c
```diff
@@ -10,1364 +10,757 @@
...
+    err = udp_sendto(ud->udp_pcb, pb, &addr, port);
+    pbuf_free(pb);
+    if (ud->client.cb_sent_ref != LUA_NOREF) {
+      lua_rawgeti(L, LUA_REGISTRYINDEX, ud->client.cb_sent_ref);
+      lua_rawgeti(L, LUA_REGISTRYINDEX, ud->self_ref);
+      lua_call(L, 1, 0);
+    }
+  } else if (ud->type == TYPE_TCP_CLIENT) {
+    err = tcp_write(ud->tcp_pcb, data, datalen, TCP_WRITE_FLAG_COPY);
   }
```
jmattsson: `err` seems to only be assigned, never checked. Is there a call missing to `luaL_lwip_checkerr()` before the return, or is it intentional that we're quiet about errors here?

jmattsson:
Current merge conflict is due to the removal of the set_size SSL stuff; trivial to fix.

app/modules/net.c
```diff
@@ -10,1364 +10,757 @@
...
+      int cb_connect_ref;
+      int cb_disconnect_ref;
+      int cb_reconnect_ref;
+    } client;
+  };
+} lnet_userdata;
+
+#pragma mark - LWIP errors
+
+int luaL_lwip_checkerr (err_t err, lua_State *L) {
```
jmattsson: Can I recommend not using the `luaL_` prefix since that's used by the official Lua aux lib? I see what you're going for though, so it's a soft recommendation only :)

Also, customarily the lua_State * is the first arg.

docs/en/modules/tls.md
```diff
@@ -0,0 +1,291 @@
...
+If event is "reconnection", the second parameter is the reason of connection error (string).
+
+#### Returns
+`nil`
+
+#### Example
+```lua
+srv = tls.createConnection()
+srv:on("receive", function(sck, c) print(c) end)
+srv:connect(443,"google.com")
```
jmattsson: The `:connect` should be moved *after* the `:on` registration to avoid a potential race - when pasting in the example into the console, the connection may be established before the callback registration has happened. I discovered this the hard way (a whole bunch of `printf`s later).

Comment also applies to the `net.socket:on()` entry.

djphoenix:
@jmattsson thx for review. Comments looks rational, I very busy right now, so I will continue (and maybe finish) work on it nearer to weekend.

jmattsson:
>    UDP sockets does not have connection callback on listen method.

This is sane.

>    UDP sockets does not have connect method and needs to pass port/ip to send method. TODO: backward compatibility

It would be nice, but I would consider merging even without this.

>    UDP socket's receive callback receives port/ip arguments before data. Maybe move to end of args to keep backward compatibility.

Yes, please move those to the end.

>    TCP server's timeout does not work. Searching for solution.

There are two options here. One is to use the regular TCP keepalive support (see `tcp_pcb->keep_idle`). The other is to add a timer to the socket userdata, and restart it on each send/receive event. I believe the latter approach is what the current interface is geared towards (and thus what espconn does internally).

>    TCP socket's hold/unhold does not work. Searching for solution.

I *think* what we would like to do here is announce a zero-sized receive window when `hold()` is called, and restore it on `unhold()`. This would mean working directly with the `tcp_pcb->rcv_wnd` field, and rather than calling `tcp_recved()` directly use a wrapper which tests our `.hold` flag and if set instead adjusts the "shadow" `rcv_wnd` (probably kept in our socket userdata) and does a direct `tcp_ack_now()`. If not in hold, the wrapper can use `tcp_recved()` directly.

---

I'm not going to claim that I have comprehensively tested each and every corner of this PR but other than what I've commented on so far, this is looking really good! Would love to merge this Soon(tm).

docs/en/modules/net.md
```diff
@@ -364,6 +396,44 @@ none
...
+
+The syntax and functional similar to [`net.server:listen()`](#netserverlisten), but callback parameter is not provided.
+
+## net.udpsocket:on()
+
+Register callback functions for specific events.
+
+The syntax and functional similar to [`net.socket:on()`](#netsocketon), only "received", "sent" and "dns" events is valid.
+
+**`received` callback have `port` and `ip` before `data` argument.**
```
jmattsson: Don't forget to update this when the args are moved. :)

Overall we should probably expand on the `net` docs, they've always been a bit light... but that's for later.

luismfonseca:
I wrote some tests for the `net` module [here](https://github.com/luismfonseca/nodemcu-mispec/blob/master/nodemcu/mispec_netclient.lua).

They pass both on the master and on this branch (in the last test in there, which uses TLS, I had to change to something other than `https://example.com` because of the issue #1485):
```lua
The net module (client), it should:
> 
  * create connection
  * resolve dns
  * connect to remote host via TCP
  * communicate to remote host via TCP with TLS
  * communicate to remote host via TCP

Completed in 6.10 seconds.
Success rate is 100.0% (0 failed out of 5).

```

This looks promising!

jmattsson:
Oooh, nice stuff @luismfonseca! @marcelstoer wouldn't it be fantastic if we could run such tests in CI too? But how to allow remote running of random-internet-provided scripts without opening up massive security holes wherever the hardware is hosted? (Should be perhaps fire up a discussion issue for this?)

marcelstoer:
> random-internet-provided scripts

Why internet-provided? Something like [`mispec_netclient.lua`](https://github.com/luismfonseca/nodemcu-mispec/blob/master/nodemcu/mispec_netclient.lua) should IMO be in our repo if we want to adopt that approach. And we should run all those tests parameterized (loop over array of URLs/domains).

> fire up a discussion issue for this?

There is one already: #516.

jmattsson:
Because we'd want to run the (in-repo) specs/tests on PR builds, and some pesky critter could add a spec that does all sorts of shenanigans to our CI hardware and surrounds :)

Let's move further discussion over to the issue you referenced (and thanks for pointing out I have goldfish memory >.> )

app/modules/net.c
djphoenix: ACK

app/modules/net.c
djphoenix: ACK

app/modules/net.c
djphoenix: ACK, miss check

app/modules/net.c
djphoenix: ACK

docs/en/modules/tls.md
djphoenix: ACK

djphoenix:
Rebased on master and fixed most of commented places. The two things remaining is Server's `timeout` and Socket's `hold`/`unhold`.

marcelstoer:
> Rebased on master

I hope you mean "Rebased on _dev_"?

djphoenix:
@marcelstoer sh... My brain was very confused in my work repo's tree. Of course, it was dev.

jmattsson:
FYI, I've used this as the basis for the `net` module on the `dev-esp32` branch. A bunch of grunt-work was needed to make it RTOS aware/safe, but that was always going to happen. Looking good so far!

luismfonseca:
Does this get closer to being merged?

Now that the merge request #1435 was merged...

marcelstoer:
Definitely closer than it was a few hours ago 😉 I take it this still passes your test suite?

Terry, [you wanted to look into this](https://github.com/nodemcu/nodemcu-firmware/pull/1379#issuecomment-252490665) but we know how busy you've been with your real-life project. Did you ever find the time to catch up here?

djphoenix:
Yep, let the party go on!

Definitely I finished work on SDK2.0.0, and this PR is my next target. And I have just two points that I should pass before I can tell "this is ready".

I will play with this points as soon as possible. Any tests on this branch are welcome.

djphoenix:
Aaand progress here! Reached server "timeout" setting (it I decided tight how it should work).
```lua
function server()
  local srv = net.createServer(net.TCP, 10)
  srv:listen(80, function(sk)
    local lp, la = sk:getaddr()
    local rp, ra = sk:getpeer()
    local l = la .. ':' .. lp
    local r = ra .. ':' .. rp
    local slr = l .. '-' .. r
    print('CONN', slr)
    sk:on('receive', function(s, dt)
      print('RECV <<<', slr)
      print(dt)
      print('<<<RECV')
    end)
    sk:on('disconnection', function(s)
      print('DIS', slr)
    end)
    sk:on('reconnection', function(s, e)
      print('ERR', slr, e)
    end)
  end)
  return srv
end
srv = server()
```

```
CONN	192.168.97.42:80-192.168.97.33:54757
-- Switched off WiFi on notebook
-- After 10s
ERR	192.168.97.42:80-192.168.97.33:54757	-8
```

djphoenix:
TCP hold done too:
```lua
function test(hold)
  local conn = net.createConnection(net.TCP, 0)
  conn:connect(8000, '192.168.97.33')
  local total = 0
  conn:on('connection', function(s)
    local lp, la = s:getaddr()
    local rp, ra = s:getpeer()
    local slr = la .. ':' .. lp .. '-' .. ra .. ':' .. rp
    print('CON', slr)
    s:send(
      'GET / HTTP/1.0\r\n' ..
      'Host: 192.168.97.33\r\n' ..
      'Connection: close\r\n' ..
      '\r\n'
    )
  end)
  conn:on('disconnection', function(s)
    print('DIS')
  end)
  conn:on('reconnection', function(s, e)
    print('ERR', e)
  end)
  conn:on('sent', function(s)
    print('SNT')
  end)
  conn:on('receive', function(s, d)
    total = total + #d
    print('RECV', #d, total)
    if hold then
      print('HOLD 1s')
      s:hold()
      tmr.create():alarm(1000, tmr.ALARM_SINGLE, function()
        print('UNHOLD')
        s:unhold()
      end)
    end
  end)
end
```

```
> test(false)
CON	192.168.97.42:11062-192.168.97.33:8000
SNT
RECV	1460	1460
RECV	1460	2920
RECV	1460	4380
RECV	1460	5840
RECV	1460	7300
RECV	1460	8760
RECV	1460	10220
RECV	552	10772
DIS
```
```
> test(true)
CON	192.168.97.42:7724-192.168.97.33:8000
SNT
RECV	1460	1460
HOLD 1s
UNHOLD
RECV	1460	2920
HOLD 1s
UNHOLD
RECV	1460	4380
HOLD 1s
UNHOLD
RECV	1460	5840
HOLD 1s
UNHOLD
RECV	1460	7300
HOLD 1s
UNHOLD
RECV	1460	8760
HOLD 1s
UNHOLD
RECV	1460	10220
HOLD 1s
UNHOLD
RECV	552	10772
HOLD 1s
DIS
UNHOLD
```

djphoenix:
Work on PR is done. Please, review (and test as possible) all changes again, and we are ready.

jmattsson:
Note to self to add the hold & timeout support on the ESP32 branch too.

docs/en/modules/net.md
```diff
@@ -3,11 +3,7 @@
 | :----- | :-------------------- | :---------- | :------ |
 | 2014-12-22 | [Zeroday](https://github.com/funshine) | [Zeroday](https://github.com/funshine) | [net.c](../../../app/modules/net.c)|
 
-**SSL/TLS support**
-
-NodeMCU includes [mbedTLS library](https://tls.mbed.org/). With default config it supports **TLS** 1.0 / 1.1 / 1.2 and most common cipher suites, including DH/ECDH key exchange methods.
-
-Full list of features you may get on [mbedTLS page](https://tls.mbed.org/core-features).
+** TLS operations moved to [tls](tls) module **
```
marcelstoer: ..."was moved to the \[TLS\](tls.md) module"

docs/en/modules/net.md
```diff
@@ -20,20 +16,23 @@ Creates a client.
 `net.createConnection(type, secure)`
 
 #### Parameters
-- `type` `net.TCP` or `net.UDP`
-- `secure` 1 for encrypted, 0 for plain
+- `type` `net.TCP` or `net.UDP`. UDP connections chained to [net.createUDPSocket()](#netcreateudpsocket)
+- `secure` 1 for encrypted, 0 for plain. Secure connections chained to [tls.createConnection()](tls#tlscreateconnection)
```
marcelstoer: \[tls.createConnection()\](tls.md#tlscreateconnection)

docs/en/modules/tls.md
```diff
@@ -0,0 +1,285 @@
+# tls Module
```
marcelstoer: TLS

docs/en/modules/net.md
```diff
@@ -20,20 +16,23 @@ Creates a client.
...
 #### Example
 
 ```lua
-net.createConnection(net.UDP, 0)
+net.createConnection(net.TCP, 0)
 ```
 
 #### See also
-[`net.createServer()`](#netcreateserver)
+[`net.createServer()`](#netcreateserver), [`net.createUDPSocket()`](#netcreateudpsocket), [`tls.createConnection()`](tls#tlscreateconnection)
```
marcelstoer: \[`tls.createConnection()`\](tls.md#tlscreateconnection)

docs/en/modules/net.md
```diff
@@ -3,11 +3,7 @@
 | :----- | :-------------------- | :---------- | :------ |
 | 2014-12-22 | [Zeroday](https://github.com/funshine) | [Zeroday](https://github.com/funshine) | [net.c](../../../app/modules/net.c)|
```
marcelstoer: Maybe you should be listed as maintainer here?

docs/en/modules/tls.md
```diff
@@ -0,0 +1,285 @@
...
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2016-07-04 | [PhoeniX](https://github.com/djphoenix) | [PhoeniX](https://github.com/djphoenix) | [tls.c](../../../app/modules/tls.c)|
```
marcelstoer: Please adjust the date to some day in December. Ideally it'd be the day we eventually merge the PR but we can't predict the future 😉 "Since" should give developers an indication when the module was add to the firmware in _our_ repository.

jmattsson:
> Also, do we think it's ok to have two very similar APIs for plain and encrypted socket connections rather than a single API as we have it now?

Imho yes. The espconn API (which our `net` API is effectively based on) is an oversimplification. Fine for a range of use cases, but woefully insufficient for some, including some rather important ones (such as mutual cert auth). As pert he old adage of "as simple as possible, but no simpler", I'd say espconn fell in the trap of "simpler". I think the compromise of having the backwards compatibility for the "easy" cases and the fuller API for other cases is a good one. Then again, I might have been the one suggesting it in the first place? >.>

docs/en/modules/net.md
djphoenix: Maybe, if it applicable. I'll change it.

docs/en/modules/tls.md
djphoenix: ACK

djphoenix:
> Johny once proposed to make TLS a regular module but loaded it automatically when CLIENT_SSL_ENABLE is set.
OK, I will add dependency and corresponding defines.

> I don't know what the best approach is. Also, the TLS module requires the net module. I guess that's ok but we should tell the developers (in tls.md).
ACK, I'll add a note.

> Also, do we think it's ok to have two very similar APIs for plain and encrypted socket connections rather than a single API as we have it now?
Of course, right now they're very similar. But next point is migrate TLS module on mbedTLS, and make some extensions to API (e.g. custom cert verification, cert-based auth, renegotiation, connection info, and many more features - discussion needed, I'll create it).

> A related question...I see CLIENT_SSL_ENABLE is used in net.c and mqtt.c but not in the HTTP code. How does that module get its TLS support?
Not sure I very familiar with it... Probably, espconn layer does not completely disabled when CLIENT_SSL_ENABLE is off.

devsaurus:
This PR doesn't work for me.

Some basic example code

```lua
function cb_connected()
    print("connected")
end
  
_conn = net.createConnection(net.TCP, 0)
_conn:on("receive", function() print("data received") end)
_conn:on("disconnection", function() print("disconnected") end)
_conn:connect(80, server_ip, cb_connected)
```

results in "connected" on `dev` branch but stays mute when I compile from this branch merged into `dev`:

```
git checkout -b djphoenix-net-udp-lwip dev
git pull https://github.com/djphoenix/nodemcu-firmware.git net-udp-lwip
make EXTRA_CCFLAGS="-DLUA_NUMBER_INTEGRAL"
```

djphoenix:
Does this work with connection callback registered with "on" method? I'll check callback set from "connect"...

djphoenix:
...ah, I missed it completely. Sorry. I will fix it.

djphoenix:
Fixed all things, re-check now.

app/include/user_modules.h
```diff
@@ -59,6 +59,7 @@
 //#define LUA_USE_MODULES_STRUCT
 //#define LUA_USE_MODULES_SWITEC
 //#define LUA_USE_MODULES_TM1829
+#define LUA_USE_MODULES_TLS
```
marcelstoer: To me it would be more logical to have `#if defined(CLIENT_SSL_ENABLE) && defined(LUA_USE_MODULES_NET)` here rather than in `tls.c`. What do you think?

devsaurus:
I'm observing a very bad real time performance with `hold()` / `unhold()`.

Adapted your script from https://github.com/nodemcu/nodemcu-firmware/pull/1379#issuecomment-267043262 to emit timestamps (roughly rounded to 1s) and the delta from UNHOLD to RECV:

```lua
function test(hold)
  local last_unhold = 0
  local conn = net.createConnection(net.TCP, 0)
  conn:connect(8000, '192.168.97.33')
  local total = 0
  conn:on('connection', function(s)
    local lp, la-- = s:getaddr()
    local rp, ra-- = s:getpeer()
    local slr-- = la .. ':' .. lp .. '-' .. ra .. ':' .. rp
    print('CON', slr)
    s:send(
      'GET /pulser.u8 HTTP/1.0\r\n' ..
      'Connection: close\r\n' ..
      '\r\n'
    )
  end)
  conn:on('disconnection', function(s)
    print('DIS')
  end)
  conn:on('reconnection', function(s, e)
    print('ERR', e)
  end)
  conn:on('sent', function(s)
    print('SNT')
  end)
  conn:on('receive', function(s, d)
    total = total + #d
    local now = tmr.now()/1000000
    print(now, 'RECV', #d, total, now-last_unhold)
    if hold then
      print('','HOLD 1s')
      s:hold()
      tmr.create():alarm(1000, tmr.ALARM_SINGLE, function()
        last_unhold = tmr.now()/1000000
        print(last_unhold, 'UNHOLD')
        s:unhold()
      end)
    end
  end)
end
```

Looks like an exponential growth in latency:

```
test(true)
CON	nil
SNT
22	RECV	1460	1460
	HOLD 1s
23	UNHOLD
24	RECV	1460	2920	1
	HOLD 1s
25	UNHOLD
26	RECV	1460	4380	1
	HOLD 1s
27	UNHOLD
29	RECV	1460	5840	2
	HOLD 1s
30	UNHOLD
36	RECV	1460	7300	6
	HOLD 1s
37	UNHOLD
50	RECV	1460	8760	13
	HOLD 1s
51	UNHOLD
77	RECV	1460	10220	26
	HOLD 1s
78	UNHOLD
131	RECV	1460	11680	53
	HOLD 1s
132	UNHOLD
240	RECV	1460	13140	108
	HOLD 1s
241	UNHOLD
```

Whereas on current `dev` there's no significant latency at all:

```
test(true)
CON	nil
SNT
13	RECV	1460	1460	13
	HOLD 1s
13	RECV	1460	2920	13
	HOLD 1s
13	RECV	1460	4380	13
	HOLD 1s
13	RECV	1460	5840	13
	HOLD 1s
13	RECV	1460	7300	13
	HOLD 1s
14	UNHOLD
14	RECV	1460	8760	0
	HOLD 1s
14	UNHOLD
14	UNHOLD
14	RECV	1460	10220	0
	HOLD 1s
14	UNHOLD
14	RECV	1460	11680	0
	HOLD 1s
14	UNHOLD
14	RECV	1460	13140	0
	HOLD 1s
14	RECV	1460	14600	0
	HOLD 1s
14	RECV	1460	16060	0
	HOLD 1s
14	RECV	1460	17520	0
	HOLD 1s
14	RECV	1460	18980	0
	HOLD 1s
15	UNHOLD
15	RECV	1460	20440	0
	HOLD 1s
15	UNHOLD
15	RECV	1460	21900	0
	HOLD 1s
15	UNHOLD
15	RECV	1460	23360	0
	HOLD 1s
15	UNHOLD
15	RECV	1460	24820	0
	HOLD 1s
15	UNHOLD
15	UNHOLD
15	RECV	1460	26280	0
	HOLD 1s
15	UNHOLD
15	UNHOLD
15	RECV	1460	27740	0
```

It's also remarkable that issuing `hold()` on `dev` doesn't throttle the stream immediately, but typically up to 5 callback invocations happen before that's achieved. `hold()` for LWIP is effective with the first instance.

TerryE:
> It's also remarkable that issuing hold() on dev doesn't throttle the stream immediately, but typically up to 5 callback invocations happen before that's achieved. hold() for LWIP is effective with the first instance.

IIRC, either the SDK documentation or the Espressif thread where this new function was introduced warned that hold was "hold eventually" and that up to 5 packets could be received before the hold took effect -- a bit like trying to close a line at the supermarket: they let the people already standing in the queue go through before closing it.

djphoenix:
Interesting point that this case affects only internet connections - local sockets works without timing issues.

Patched `net.c` to get behavior that more similar to current `dev`. So modified test code a bit:
```lua
function test(hold)
  local last_unhold = 0
  local conn = net.createConnection(net.TCP, 0)
  conn:connect(80, 'phoenix.dj')
  local total = 0
  conn:on('connection', function(s)
    print('CON')
    s:send(
      'GET /Gourmia_design_test2.pdf HTTP/1.0\r\n' ..
      'Host: phoenix.dj\r\n' ..
      'Connection: close\r\n' ..
      '\r\n'
    )
  end)
  conn:on('disconnection', function(s)
    print('DIS')
  end)
  conn:on('reconnection', function(s, e)
    print('ERR', e)
  end)
  conn:on('sent', function(s)
    print('SNT')
  end)
  conn:on('receive', function(s, d)
    total = total + #d
    local now = tmr.now()/1000000
    print(now, 'RECV', #d, total, now-last_unhold)
    if hold then
      print('','HOLD 1s')
      hold = false
      s:hold()
      tmr.create():alarm(1000, tmr.ALARM_SINGLE, function()
        last_unhold = tmr.now()/1000000
        print(last_unhold, 'UNHOLD')
        hold = true
        s:unhold()
      end)
    end
  end)
end
```

Now works as expected.

app/modules/net.c
```diff
@@ -10,1363 +10,780 @@
...
-  // "enhanced" end
-
-  lua_call(L, 2, 0);
-
-end:
-  if((pesp_conn->type == ESPCONN_TCP && pesp_conn->proto.tcp->remote_port == 0)
-    || (pesp_conn->type == ESPCONN_UDP && pesp_conn->proto.udp->remote_port == 0) ){
+  ud->client.wait_dns --;
+  if (ud->pcb && ud->type == TYPE_TCP_CLIENT && ud->tcp_pcb->state == CLOSED) {
+    tcp_connect(ud->tcp_pcb, &addr, ud->tcp_pcb->remote_port, net_connected_cb);
```
luismfonseca: > The tcp_connect() function can return ERR_MEM if no memory is available

From: http://lwip.wikia.com/wiki/Raw/TCP

luismfonseca:
First of, this was quite an effort Yuri! 👍 

I've spent the weekend refactoring my websocket client implementation while reading your code and the lwIP documentation.

I think that the memory usage is not as efficient... I'm running the following code on the current `dev`:
```
tmr.alarm(1, 2000, 1, function()
  n=net.createConnection(net.TCP)
  n:on("connection", function(x, _)
    print(node.heap())
    x:send("GET / HTTP1.1")
  end)
  n:on("sent", function(x)
    x:close()
    n = nil
  end)
  n:connect(80, "wwww.websocket.org")
end)
```
Gives: 
> 39536
> 39720
> 39536
> 39720
> 39536
> 39720
> ....

So it stabilizes at `39536`... In this branch, it goes:
> 39928
> 39744
> 39560
> 39376
> 39192
> 39008
> 38824
> (...)
> 29616
> 29616
> 29616
> 29616
> 29616

When I stop the alarm, it will (very) slowly recover the heap available. This could pose a problem for applications that keep creating several net connections (and are on the edge of going out of heap).

djphoenix:
So if it stopped, thi is not a memory leak but some deferred frees of a buffers. This may be an LWIP sockets in TIME_WAIT state. Interesting that espconn have not this issue... or have another.

app/modules/net.c
```diff
@@ -233,7 +233,7 @@ static err_t net_tcp_recv_cb(void *arg, struct tcp_pcb *tpcb, struct pbuf *p, er
     return tcp_close(tpcb);
   }
   net_recv_cb(ud, p, 0, 0);
-  tcp_recved(tpcb, p->len);
+  tcp_recved(tpcb, ud->client.hold ? 0 : TCP_WND);
```
jmattsson: is it really safe to use TCP_WND here rather than p->len (or possibly an accumulation of p->len from all packets received since the hold started)

jmattsson:
I should perhaps elaborate a bit more on that comment - I have two questions/possible concerns:
  1. Whether it impacts any buffer management when lying about the receive amount, and
  1. whether it runs the risk of prematurely ACKing in-flight data before having received it, and thus risk losing data from what's supposed to be a reliable TCP stream.

djphoenix:
@jmattsson 
1. Not seen any issues for me. Both with and without hold/unhold, after received all of data from another side, I got constant memory consumption and none failures. Wireshark reports all data frames are OK too. So anyway receive window are limited to TCP_WND in LWIP code.
2. Not seen any issues too - but maybe this is just a luck.

luismfonseca:
Following up on my previous comment, I've added `printf`s to basically every call that `espconn_tcp` layer does to `lwip` [[raw output]](http://pastebin.com/raw/wBSDQ64z).

It seems that it first will do a simple `tcp_close` call. But it keeps track of the connections in its `tcp_pcb_lists[]`.

After a few connections it will call `espconn_kill_oldest_pcb` and it will do:

```
tcp_abort(3fff0e68):espconn_kill_oldest
tcp_pcb_remove(3ffef73c, 3fff2928):espconn_kill_oldest
```

So it seems that it will wait (per lwip standards/timeouts for the connection to be closed) up until the point where it has multiple existing connections (?) or connections with an inactive state for a long period of time (judging by the `espconn_kill_oldest` code). Then, it will trigger the `tcp_abort` and `tcp_pcb_remove` - which oddly enough, is for a different `pcb`.

The question is whether we want this optimization in our `net` module. And if we do, we would benefit from something like the `espconn_tcp` - an intermediate layer that interacts with lwip and does its optimizations. The benefit being that other modules like `http` or `websocket` wouldn't need to reimplement them. As is, this merge request mixes `lua` directly with `lwip`, so it's not possible to reuse any code in `http` or `websocket`.

I've tried changing `TCP_FIN_WAIT_TIMEOUT` from `20000` to `2000` but it didn't make a difference in this case.

I've also tried changing `TCP_MSL` from `60000UL` to `30000UL` and that did seem to make an impact. So `lwip` is internally now freeing the connections earlier (this is still a bit different than the `espconn` behaviour). So with the same test, the heap go to to a lowest `35816` (which is much more acceptable than the previous `29616`).

Looking at LWIP code, it seems that if it is failing to allocate a new `pcb`, one thing it does is to kill off old connections that are in the `TIME_WAIT` state. It would be interesting to add a hook to lua GC somehow so that when memory was really running low, kill off all `TIME_WAIT` connections, regardless of their age.

jmattsson:
It's been a long road, but I think we're here. Does anyone have any objections to merging this now?

djphoenix:
There is at least a month to next master drop, so this is enough to test this huge PR roughly.
About `TIME_WAIT` I agree @luismfonseca: the way is add TIME_WAIT cleanup somewhere in lua GC, and also (maybe) tune LWIP config to clean TIME_WAIT sooner than now.

app/include/user_modules.h
marcelstoer: Bump - what's with this?

djphoenix: I think we should keep modules-list file as clean as possible. Maybe (as an another way) remove `#if defined` line, it may get code cleaner, but adds some points in dependency matrix.

devsaurus:
Did some more stress testing with my vncserver framework - no regressions observable.
IMO we have enough confirmation from several parties, so let's get the 🎱 rolling and start 2017 with LWIP. Happy new year 😄

PR #1413, opened 2016-07-22
Change 'formatting file system' message

marcelstoer:
The message currently printed to console when the fs is formatted is not ideal IMO. If the system just prints "Formatting file system." and then acts like it's actually stuck with no indication whatsoever (console progress ticker, device LED blinking, etc.) some users may cancel and reboot.

A quick-fix is either to print "Formatting file system..." to indicate that it's a currently running process or to say something like "Formatting file system. Don't abort, may take a while.". An alternative would be to also print a success message explicitly like so:

```
print "Formatting file system..."
success = fs_format()
if (success) {
  print "done"
} else {
  // print error msg
}
```

### DIFF
--- a/app/user/user_main.c
+++ b/app/user/user_main.c
@@ -101,7 +101,7 @@ void nodemcu_init(void)
 #if defined ( BUILD_SPIFFS )
     if (!fs_mount()) {
         // Failed to mount -- try reformat
-	c_printf("Formatting file system.\n");
+	c_printf("Formatting file system. Please wait...\n");
         if (!fs_format()) {
             NODE_ERR( "\n*** ERROR ***: unable to format. FS might be compromised.\n" );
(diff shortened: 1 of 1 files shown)


djphoenix:
Is there a chance to make it async? Maybe best solution is allow regular operations, but reject file-related?

pjsg:
The formatting doesn't take a long time, but it isn't quick either..... It isn't worth making it async (and it wouldn't be easy either) as it doesn't happen often (in particular, it doesn't happen on every firmware flash).

I'd make the message a bit more explicit (I couldn't see any reasonable way to have a progress meter)

marcelstoer:
As for async I agree.
> I'd make the message a bit more explicit

Any particular suggestion? How about what I suggested above in the PR description?

pjsg:
"Formatting file system. Please wait..."

marcelstoer:
Make sense, I pushed an update.

PR #1414, opened 2016-07-22
Re-write the 'upgrading firmware' section in the flash docs

marcelstoer:
As discussed in https://github.com/nodemcu/nodemcu-firmware/pull/1390#discussion_r71603589.

### DIFF
--- a/docs/en/flash.md
+++ b/docs/en/flash.md
@@ -4,7 +4,7 @@ Adafruit provides a really nice [firmware flashing tutorial](https://learn.adafr
 
     Keep in mind that the ESP8266 needs to be [put into flash mode](#putting-device-into-flash-mode) before you can flash a new firmware!
 
-## esptool
+## esptool.py
 > A cute Python utility to communicate with the ROM bootloader in Espressif ESP8266. It is intended to be a simple, platform independent, open source replacement for XTCOM.
 
 Source: [https://github.com/themadinventor/esptool](https://github.com/themadinventor/esptool)
@@ -15,7 +15,7 @@ Supported platforms: OS X, Linux, Windows, anything that runs Python
 
 Run the following command to flash an *aggregated* binary as is produced for example by the [cloud build service](build.md#cloud-build-service) or the [Docker image](build.md#docker-image).
 
-`esptool.py --port <USB-port-with-ESP8266> write_flash -fm <mode> -fs <size> 0x00000 <nodemcu-firmware>.bin`
+`esptool.py --port <serial-port-of-ESP8266> write_flash -fm <mode> -fs <size> 0x00000 <nodemcu-firmware>.bin`
 
 - `mode` is `qio` for 512&nbsp;kByte modules and `dio` for 4&nbsp;MByte modules (`qio` might work as well, YMMV).
 - `size` is given in bits. Specify `4m` for 512&nbsp;kByte and `32m` for 4&nbsp;MByte.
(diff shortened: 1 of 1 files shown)


docs/en/flash.md
```diff
@@ -50,22 +50,30 @@ Also, in some special circumstances, you may need to flash `blank.bin` or `esp_i
 
 If upgrading from [SPIFFS](https://github.com/pellepl/spiffs) version 0.3.2 to 0.3.3 or later, or after flashing any new firmware (particularly one with a much different size), you may need to run [`file.format()`](modules/file.md#fileformat) to re-format your flash filesystem. You will know if you need to do this if your flash files disappeared, or if they exist but seem empty, or if data cannot be written to new files.
```
pjsg: Can you change this para to say 

When flashing a new firmware (particularly with a much different size), the flash filesystem may be reformatted as the firmware starts. If it is not automatically reformatted, then it should be valid and have the same contents as before the flash operation. You can still run [`file.format()`](modules/file.md#fileformat) to re-format your flash filesystem. You will know if you need to do this if your flash files exist but seem empty, or if data cannot be written to new files. However, this should be an exceptional case.

docs/en/flash.md
```diff
@@ -50,22 +50,30 @@ Also, in some special circumstances, you may need to flash `blank.bin` or `esp_i
...
+Espressif changes the init data block (`esp_init_data_default.bin`) for their devices along the way with the SDK. So things break when a NodeMCU firmware with a certain SDK is flashed to a module which contains init data from a different SDK. Hence, this section applies to upgrading NodeMCU firmware just as well as *downgrading* firmware.
 
-Download SDK patch 1.5.4.1 (http://bbs.espressif.com/download/file.php?id=1572) and extract `esp_init_data_default.bin` from there. *Use this file together with the new firmware during flashing*. This SDK release introduced an additional sector for RF calibration data. Espressif recommends to clear this one as well in the [release notes](http://bbs.espressif.com/viewtopic.php?f=46&t=2376). The simplest way to set up the flash from scratch is to erase the complete flash before upgrading (or downgrading) SDK versions.
+A typical case that often fails is when a module is upgraded from a 0.9.x firmware to a recent version. It might look like the new firmware is broken, but the reason for the missing Lua prompt is related to the big jump in SDK versions.
```
pjsg: Add a comment that says that the default baud rate also changed between 0.9.6 and 1.5

docs/en/flash.md
marcelstoer: Done. I also moved it to the new "Upgrading" section. Makes more sense there for me.

docs/en/flash.md
devsaurus: Marcel, I can't find changes related to Philip's remark in your last commit. Do you intend to have further updates to the PR before merging?

marcelstoer: Oh, I unintentionally missed that, sorry. However, I don't think it really belongs here. I would not want to turn that note into a highly condensed change log. The thing about `socket:send` is just a prominent example of APIs that change between releases. The default baud rate is documented in the README and twice in the docs (uart module and "Uploading code") - I believe that should be enough? I also added it to the release tag description at https://github.com/nodemcu/nodemcu-firmware/releases/tag/1.5.1-master_20160603.

> Do you intend to have further updates to the PR before merging?

If you have more feedback I will 😃  I don't intend to merge it myself if not absolutely necessary anyway.

devsaurus: Thanks, wanted to check this aspect before merging.

docs/en/flash.md
```diff
@@ -48,24 +48,32 @@ Otherwise, if you built your own firmware from source code:
...
+Espressif changes the init data block (`esp_init_data_default.bin`) for their devices along the way with the SDK. So things break when a NodeMCU firmware with a certain SDK is flashed to a module which contains init data from a different SDK. Hence, this section applies to upgrading NodeMCU firmware just as well as *downgrading* firmware.
+
+A typical case that often fails is when a module is upgraded from a 0.9.x firmware to a recent version. It might look like the new firmware is broken, but the reason for the missing Lua prompt is related to the big jump in SDK versions.
+
+If there is no init data block found during SDK startup, the SDK will install one itself. If there is a previous (potentially too old init block), the SDK *probably* doesn't do anything with it but there is no documentation from Espressif on this topic.
+
+Hence, there are two strategies to update the SDK init data:
+
+- Erase flash completely. This will also erase the (Lua) files you uploaded to the device! The SDK will install the init data block during startup.
```
dnc40085: This isn't the case for either my esp-07(512k) or my esp-12(4M), after I erase the flash with `esptool.py ... erase_flash` then upload `0x00000.bin` and `0x10000.bin` with `esptool.py ... write_flash`, I get the error `rf_cal[0] !=0x05,is 0xFF` then it just spits out garbage after I pull the reset pin low.

In order to get the esp-07 to boot up I had to flash `esp_init_data_default.bin` to `0x7c000` and it booted up, simple enough.
The esp-12 was not so simple, I flashed `esp_init_data_default.bin` to `0x3fc000` and still got garbage, so I erased again then flashed it another time and got the same result. I spent the whole day messing with it and got nowhere.
Then I got the bright idea to try flashing `esp_init_data_default.bin` to `0x7c000` and `0x3fc000` then it finally booted up.

I'm not sure if this is a problem with the SDK or NodeMCU.

djphoenix: @dnc40085 dumb question, did you set up `write_flash`'s `flash_size` in `esptool.py`? SDK gets flash map based on several bits at begin of flash. These bits set by esptool, and default size is 4m (mega**bits**, for 4 megabytes you need 32m).

dnc40085: @djphoenix I guess I need to RTFM a little more, I assumed esptool.py automatically configured flash size on upload.

PR #1423, opened 2016-07-28
Update modules to be Lua coroutine compatible

### LINKED ISSUE #846: Updating modules that are incompatable with Lua coroutining

djphoenix:
Fixes #846.
Recreation of #1167 with resolved conflicts.

### DIFF
--- a/app/cjson/cjson_mem.c
+++ b/app/cjson/cjson_mem.c
@@ -2,19 +2,14 @@
 #include "../lua/lauxlib.h"
 #include <c_stdlib.h>
 
-static lua_State *gL;
 static const char errfmt[] = "cjson %salloc: out of mem (%d bytes)";
-
-void cjson_mem_setlua (lua_State *L)
-{
-  gL = L;
-}
 
 void *cjson_mem_malloc (uint32_t sz)
 {
   void *p = (void*)c_malloc (sz);
-  if (!p && gL)
-    luaL_error (gL, errfmt, "m", sz);
+  lua_State *L = lua_getstate();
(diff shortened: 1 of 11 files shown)


TerryE:
@djphoenix Yuri, thanks for doing this and finishing a job that I've left undone because of other commitments.  Much appreciated.  One thing to consider is the relevance of the `if(!gL) ...` logic which typically executes an an aborting return.  `lua_getstate()` use the static `lua_crtstate` in `lstate.c` which is initialised in `lua_open()` which is called immediately in `lua_main()` and is not changed (except in `LUA_CROSS_COMPILER` shutdown which is not relevant to any of these changes).

With one exception, all modules are initialised after this point so the now `if(!L) ...` branches are redundant and these code paths can never be tested.

The exception is that `platform_init()` is executed before posting the `start_lua` task, and this establishes the UART ISR leading to a potential race condition in `app/modules/uart.c` so we do need to keep this 
`if(!L) return false;` here.

djphoenix:
OK, fixed some entries that was found.

TerryE:
:smile: Thanks Yuri

PR #1424, opened 2016-07-28
Dynamic timers support

### LINKED ISSUE #1416: Feature request: dynamic software timers
### Missing feature
Now we have 7 (0-6) timer control functions. When we need more, we stuck in refactoring.

### API proposal
```lua
local timer = tmr.create( function(t) -- self in callback
  print("TICK")
  -- after return, deallocated if not repeat nor rearmed
end )
timer:arm(1000, true) -- delay, [repeat=false]
timer:arm_us(152342) -- delay_us, [repeat=false]
timer:disarm()
timer = nil -- deallocated if not armed, kept otherwise
```

### Implementation proposal
I have finished it in my local repo (on top of `ETSTimer` of course), if there is no rejections of idea, I'll cleanup code, write docs and make PR.

djphoenix:
Fixes #1416.

I'll write docs soon, now will be great if anyone test this stuff.

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -146,6 +146,7 @@ static int tmr_now(lua_State* L){
 }
 
 static timer_t tmr_get( lua_State *L, int stack ) {
+	// Deprecated: static 0-6 timers control by index.
 	luaL_argcheck(L, (lua_isuserdata(L, stack) || lua_isnumber(L, stack)), 1, "timer object or numerical ID expected");
 	if (lua_isuserdata(L, stack)) {
 		return (timer_t)luaL_checkudata(L, stack, "tmr.timer");
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -71,7 +71,7 @@ static const char* MAX_TIMEOUT_ERR_STR = "Range: 1-"STRINGIFY(MAX_TIMEOUT_DEF);
 
 typedef struct{
 	os_timer_t os;
-	sint32_t lua_ref;
+	sint32_t lua_ref, self_ref;
 	uint32_t interval;
 	uint8_t mode;
 }timer_struct_t;
@@ -95,11 +95,17 @@ static timer_struct_t alarm_timers[NUM_TMR];
 static os_timer_t rtc_timer;
 
 static void alarm_timer_common(void* arg){
-	timer_t tmr = &alarm_timers[(uint32_t)arg];
+	timer_t tmr = (timer_t)arg;
 	lua_State* L = lua_getstate();
 	if(tmr->lua_ref == LUA_NOREF)
 		return;
 	lua_rawgeti(L, LUA_REGISTRYINDEX, tmr->lua_ref);
+	if (tmr->self_ref == LUA_REFNIL) {
+		uint32_t id = tmr - alarm_timers;
+		lua_pushinteger(L, id);
+	} else {
+		lua_rawgeti(L, LUA_REGISTRYINDEX, tmr->self_ref);
+	}
 	//if the timer was set to single run we clean up after it
 	if(tmr->mode == TIMER_MODE_SINGLE){
 		luaL_unref(L, LUA_REGISTRYINDEX, tmr->lua_ref);
@@ -108,7 +114,11 @@ static void alarm_timer_common(void* arg){
 	}else if(tmr->mode == TIMER_MODE_SEMI){
 		tmr->mode |= TIMER_IDLE_FLAG;
 	}
-	lua_call(L, 0, 0);
+	if (tmr->mode != TIMER_MODE_AUTO && tmr->self_ref != LUA_REFNIL) {
+		luaL_unref(L, LUA_REGISTRYINDEX, tmr->self_ref);
+		tmr->self_ref = LUA_NOREF;
+	}
+	lua_call(L, 1, 0);
 }
 
 // Lua: tmr.delay( us )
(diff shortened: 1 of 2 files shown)


devsaurus:
Great, I'll have a look.

app/modules/tmr.c
```diff
@@ -135,20 +141,32 @@ static int tmr_now(lua_State* L){
 	return 1; 
 }
 
-// Lua: tmr.register( id, interval, mode, function )
+static int tmr_get( lua_State *L, int stack, timer_t *tmr ) {
+  if (lua_isuserdata(L, stack)) {
```
devsaurus: The code can emit a clearer error message if you query the element type first and then have `luaL_argcheck()` report a wrong type with something like "timer object or numerical ID expected".

app/modules/tmr.c
```diff
@@ -135,20 +141,32 @@ static int tmr_now(lua_State* L){
...
+    *tmr = (timer_t)luaL_checkudata(L, stack, "tmr.timer");
+    if (!*tmr) return luaL_error(L, "invalid userdata");
```
devsaurus: No need to check `tmr` for != NULL. `luaL_checkudata()` will error out if the metatable doesn't match AFAIK.

app/modules/tmr.c
```diff
@@ -307,8 +340,44 @@ static int tmr_softwd( lua_State* L ){
...
+  ud->self_ref = LUA_NOREF;
+  ud->mode = TIMER_MODE_OFF;
+  ets_timer_disarm(&ud->os);
+  return 1;
+}
+
+// Lua: tmr.delete()
+static int tmr_delete( lua_State *L ) {
+  timer_t ud = (timer_t)luaL_checkudata(L, 1, "tmr.timer");
+  if (!ud) return luaL_error(L, "wrong userdata");
```
devsaurus: See above.

devsaurus:
The following code PANICs when the timer expires:

```lua
tim = tmr.create()
tim:register(5000, tmr.ALARM_SINGLE, function(t)
    print("timer expired")
    print(t:state())
    t:unregister()
end)

tim:start()
print(tim:state())

-- PANIC when object is destroyed before timer expired
tim = nil
```

app/modules/tmr.c
```diff
@@ -108,7 +108,13 @@ static void alarm_timer_common(void* arg){
...
 		tmr->mode |= TIMER_IDLE_FLAG;
 	}
-	lua_call(L, 0, 0);
+  if (tmr->self_ref == LUA_REFNIL) {
+    uint32_t id = tmr - alarm_timers;
+    lua_pushinteger(L, id);
+  } else {
+    lua_rawgeti(L, LUA_REGISTRYINDEX, tmr->self_ref);
+  }
+	lua_call(L, 1, 0);
```
devsaurus: Please stick to a consistent indentation style.
If unsure, follow the Lua style: 2 space indent + 1TBS + no trailing spaces.

app/modules/tmr.c
```diff
@@ -307,8 +340,44 @@ static int tmr_softwd( lua_State* L ){
...
 
+static const LUA_REG_TYPE tmr_dyn_map[] = {
+	{ LSTRKEY( "register" ),     LFUNCVAL( tmr_register ) },
+	{ LSTRKEY( "alarm" ),        LFUNCVAL( tmr_alarm ) },
+	{ LSTRKEY( "start" ),        LFUNCVAL( tmr_start ) },
+	{ LSTRKEY( "stop" ),         LFUNCVAL( tmr_stop ) },
+	{ LSTRKEY( "unregister" ),   LFUNCVAL( tmr_unregister ) },
+	{ LSTRKEY( "state" ),        LFUNCVAL( tmr_state ) },
+	{ LSTRKEY( "interval" ),     LFUNCVAL( tmr_interval) }, 
+  { LSTRKEY( "__gc" ),         LFUNCVAL( tmr_delete ) },
```
devsaurus: Why not just call `tmr_unregister()` upon garbage collection? IMO the only difference is the condition for `ets_timer_disarm()`.

djphoenix:
Thanks @devsaurus for review. I will fix codestyle & logics and push it back today.

app/modules/tmr.c
djphoenix: ACK

app/modules/tmr.c
djphoenix: Is luaL_error (and other checks) breaks execution of C-level? Or we need to return anyway? Sorry, lua<->C integrations is blackbox for me.

app/modules/tmr.c
djphoenix: ACK

app/modules/tmr.c
devsaurus: The docs mention a long jump to terminate the C function: http://pgl.yoyo.org/luai/i/lua_error. These functions simply don't return to the calling context in case of an error.

devsaurus:
Very much appreciated @djphoenix. It's been just some minor aesthetic things while functionality is great.

djphoenix:
Fixed all of your comments and began to write docs.

djphoenix:
All done. Also fixed self-unref if mode was SINGLE or SEMI and timer was not rearmed during callbeck.

Now this works OK without leaks:
```lua
function setTimeout(t, f)
    return tmr.create():alarm(t, tmr.ALARM_SINGLE, f)
end

setTimeout(1000, function()
    print("TICK!")
end)
```

marcelstoer:
I'm missing the deprecation notes for the old-style functions in code and docs.

djphoenix:
My propose is now only add new functions. We may think about deprecation process in another thread.

marcelstoer:
You need to think about deprecating the old/existing stuff whenever you add something new that serves as a replacement. Deprecating doesn't mean removing but adding a note to others indicating that existing functionality will or might be removed in the future. That way we give developers some slack to transition to using the new API. Also, we remove reluctance as an obstacle to actually removing old API.

djphoenix:
Ohm. OK. Can you point me to example/instruction of deprecation notes? So I'll do it some later cause AFK.

marcelstoer:
We have no guidelines for that, sorry. In essence the word "deprecated" in code and docs is all it takes e.g. http://nodemcu.readthedocs.io/en/latest/en/modules/node/#nodekey-deprecated.

djphoenix:
@marcelstoer check it... And we *really* need guidelines to deprecation process, either for arguments or functions or whole modules.

devsaurus:
:+1: from my side.

eku:
@djphoenix  how long does it take until http://nodemcu.readthedocs.io/en/master/en/modules/tmr/ is updated?

djphoenix:
@eku you watching to master-branch docs, they're updating parallel with code at "master drops". Now this feature is only in dev branch, so look for dev docs ;)

marcelstoer:
@eku There's a fly-out menu in the lower left corner with which you, amongst others, can select the branch.

Issue #384, opened 2015-04-25
Feature request: Websockets


urmilparikh:
Any plan to add support for Websocket client / server?

marcelstoer:
IMO this shouldn't be built right into the firmware.

marcoskirsch:
Wrong layer of the st wrong layer of the stack.

marcelstoer:
Btw, this was brought up before in #30 in denied.

urmilparikh:
Thanks @marcelstoer, I got some clues from that thread. Hopefully will have something working soon!

urmilparikh:
Tried https://github.com/openresty/lua-resty-websocket, not enough memory :(

f4t:
what's the solution for this as of today ?

TerryE:
The previous poster approach might work with the dev140 when we've sorted out the issues.

Issue #30, opened 2014-12-13
Implementing Web Sockets


samehhady:
Hi Zeroday,

You think its possible to implement web sockets on ESP?

Think of it as we will be able to control all ESPs using a cloud server & there will be 2 way communication between all ESPs and the server!

Also this can provide a cloud based flashing tool for ESP, we will be able to flash the ESPs from anywhere!

kireevco:
I used rabbitmq with web-stomp and mqtt to connect nodemcu.
I would rather have mqtt working more stable on the ESP8266.

dvv:
There's also a good and lightweight https://developer.mozilla.org/en-US/docs/Web/API/EventSource

kireevco:
Thanks. mosquitto / rabbitmq has websocket support. I use rabbitmq on a raspberry pi as a central hub, works great.

urmilparikh:
@samehhady, it will be nice if you can share your lua or js code to implement Websocket over TCP connection.

I'm planning to do so, can leverage on your work.

Thanks!

creationix:
Websocket is a little heavy, but I think it can be done.  I tried porting my luvit websocket library but it used too much ram.  The protocol itself is pretty simple.  When I get a chance I'll write a pure lua websocket server so that an html5 based app can talk directly to the nodemcu.

pastukhov:
The main problem is md5 and sha1 algorithms, as i see.

creationix:
Those can also be implemented in lua, I've done it a few times before in JS.  Though I do wonder how much memory that would use.

creationix:
@pastukhov There is no MD5 is websocket, but you do need SHA1 and Base64 encoding.  Here is a sha1 function I just wrote using the `bit.*` library that should work.
https://gist.github.com/creationix/7ce3796e65b66549c4b0

pastukhov:
I'm getting 
```lua
> dofile('testsha1.lua')
string length overflow
```

creationix:
@pastukhov You probably don't want to run all the unit tests on the nodemcu.  One is calculating the sha1 of a string containing 1,000,000 "a"s long.  The values that will be used in websocket handshake are under 100 characters long.

ikaras:
Hi, all!
I tried to resolve issue with websocket client for nodemcu: https://github.com/ikaras/nodemcu-websocket-client
Take a look and be free to use it.

creationix:
[...]

PR #1433, opened 2016-08-02
Feature Websocket module

### LINKED ISSUE #1425: Websocket module
### Missing feature
A native [Websocket ](https://tools.ietf.org/html/rfc6455) module.

### Justification
The websockets allow a two-way communication between the client and server, so it's more well suited for IoT applications where one would like to stream data.
Being an upgradable protocol from HTTP this means the existing infrastructure can still handle websockets connections.

The nodemcu offers an `net` module for more raw control of the connection - this, however, is an higher-level protocol that sits on the same level of abstraction of the `http`.

The main advantage for a native module is heap usage.

### Workarounds
There are 2 projects that implement this in Lua [[1]](https://github.com/creationix/nodemcu-webide) [[2]](https://github.com/ikaras/nodemcu-websocket-client). Both use more than 10KB of heap.

### Implementation
I'm aware this is a non-trivial effort, therefore I'm also volunteering to do this myself. Although I feel I should reach out for possible implementation discussion & comments.
I'm also more interested in a client version, although after a client is done, a server should not be too hard.

luismfonseca:
Fixes #1425.

Websockets allow a two-way communication with servers which is better suited for realtime applications. Existing Lua impementations are not as fast and take over 10KB of heap for a single connection.

Since it has been a several years without programming in C or C++, I'm sure my code could be improved. Please let me know where. Also, I'd like some additional testing to other websockets servers (I've mainly tested against Pushers and websocket.org).

Committers supporting this PR: .


luismfonseca:
Hi, I know you guys are busy with the new releases and whatnot but I'd like to know what I can do to progress this further. (Or is this just rejected?)

Btw, for reference sake, this solves issues #30 and #384.

devsaurus:
I will review the code for formal aspects, but websockets is not a use case for me. We'd need someone else who can assess the overall design and fitness for typical use.

luismfonseca:
Ok, thank you.
Perhaps @creationix and @ikaras can provide input as well? (Seeing that both actually implemented a websocket interface).

app/modules/websocket.c
```diff
@@ -0,0 +1,332 @@
...
+  luaL_getmetatable(L, METATABLE_WSCLIENT);
+  lua_setmetatable(L, -2);
+
+  return 1;
+}
+
+static int websocketclient_onConnection(lua_State *L) {
+  NODE_DBG("websocketclient_onConnection\n");
+  ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);  
+  luaL_argcheck(L, ws, 1, "Client websocket expected");
```
devsaurus: All checks for `ws` after `luaL_checkudata()` are redundant since this function will effectively not return to your context in case of an error. It'll long jump to the calling Lua context instead. Ref. https://www.lua.org/manual/5.1/manual.html#lua_error.
Also applies to similar instances below.

app/modules/websocket.c
```diff
@@ -0,0 +1,332 @@
...
+
+  if (ws == NULL || ws->reservedData == NULL) {
+    return luaL_error(L, "Client websocket is nil.\n");
+  }
+  ws_data *data = (ws_data *) ws->reservedData;
+
+  if (lua_type(L, 2) != LUA_TNIL && lua_type(L, 2) != LUA_TFUNCTION && lua_type(L, 2) != LUA_TLIGHTFUNCTION) {
+    return luaL_error(L, "bad argument #2 to 'onConnection' (function or nil expected, got something else)");
```
devsaurus: Cosmetic: Use [`luaL_typerror()`](http://pgl.yoyo.org/luai/i/luaL_typerror) instead `luaL_error()`.

app/modules/websocket.c
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  if (data->onConnection != LUA_NOREF) {
+    luaL_unref(L, LUA_REGISTRYINDEX, data->onConnection);
+  }
```
devsaurus: Your intention is more obvious if you set `data-onConnection = LUA_NOREF;` here and below...

app/modules/websocket.c
```diff
@@ -0,0 +1,332 @@
...
+
+  if (lua_type(L, 2) == LUA_TNIL) {
+    data->onConnection = LUA_NOREF;
+  } else {
```
devsaurus: ...only test for case `lua_type(L, 2) != LUA_TNIL`.

app/modules/websocket.c
```diff
@@ -0,0 +1,332 @@
...
+
+  ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);  
+  luaL_argcheck(L, ws, 1, "Client websocket expected");
+
+  if (ws == NULL || ws->reservedData == NULL) {
+    return luaL_error(L, "Client websocket is nil.\n");
+  }
+  ws_data *data = (ws_data *) ws->reservedData;
+
+  if (data->onConnection != LUA_NOREF) {
```
devsaurus: Check for `LUA_NOREF` is redundant since [`luaL_unref()`](http://pgl.yoyo.org/luai/i/luaL_unref) handles this case.

app/modules/websocket.c
```diff
@@ -0,0 +1,332 @@
...
+
+static const LUA_REG_TYPE websocket_map[] =
+{
+  { LSTRKEY("createClient"), LFUNCVAL(websocket_createClient) },
+  { LNILKEY, LNILVAL }
+};
+
+static const LUA_REG_TYPE websocketclient_map[] =
+{
+  { LSTRKEY("onConnection"), LFUNCVAL(websocketclient_onConnection) },
```
devsaurus: Several modules like `net` and `mqtt` only expose one single `module.on()` method where the event is identified by a string. I suggest to use [`luaL_checkoption()`](http://pgl.yoyo.org/luai/i/luaL_checkoption) for decoding instead of the if/else logic found over there.

app/websocket/websocketclient.c
```diff
@@ -0,0 +1,816 @@
...
+#define WS_OPCODE_CLOSE 0x8
+#define WS_OPCODE_PING 0x9
+#define WS_OPCODE_PONG 0xA
+
+static char *cryptoSha1(char *data, unsigned int len) {
+  SHA1_CTX ctx;
+  SHA1Init(&ctx);
+  SHA1Update(&ctx, data, len);
+  
+  uint8_t *digest = (uint8_t *) os_zalloc(20);
```
devsaurus: Since you already included c_stdlib.h: `os_zalloc()` -> `c_zalloc()`.
Same below.

app/websocket/websocketclient.c
```diff
@@ -0,0 +1,816 @@
...
+  SHA1Final(digest, &ctx);
+  return (char *) digest; // Requires free
+}
+
+static const char *bytes64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
```
devsaurus: Does it make sense to use the `encoder` module here in the same way as it's done in [`crypto`](https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/crypto.c#L40)? This would be another instance of the #925 interdependencies.

app/websocket/websocketclient.c
```diff
@@ -0,0 +1,816 @@
...
+
+  ws->unhealthyPoints = 0; // received data, connection is healthy
+  os_timer_disarm(&ws->timeoutTimer); // reset ping check
+  os_timer_arm(&ws->timeoutTimer, WS_PING_INTERVAL_MS, true);
+
+
+  char *b = buf;
+  if (ws->frameBuffer != NULL) { // Append previous frameBuffer with new content
+    NODE_DBG("Appending new frameBuffer to old one \n");
+    char *newFrameBuffer = os_zalloc(ws->frameBufferLen + len);
```
devsaurus: Can the following code be substituted by `c_realloc()`?
Same below.

devsaurus:
As a general remark: your code bases on espconn which will vanish sooner or later. There are already efforts ongoing to re-base networking to LWIP in #1379. Don't know if it's worth considering this here already from the beginning.

djphoenix:
I really propose to reject new modules that use espconn layer. It's very bad and bugous.
If it uses TLS client mode, wait for #1435 and rebase it on mbedtls.

creationix:
My use case is websocket server which doesn't appear to be in here.

luismfonseca:
Thanks for reviewing @devsaurus.

I also feel that espconn layer is bad, so I understand the need to remove it. It would be nice to continue to work on this so it uses LWIP, but on my next MR. That wouldn't make me feel that I wasted so much time battling the espconn!

@creationix: Yes, you're right. But it would be mostly the same except from having a new method: `createServer(function(ws))` which would receive a function to handle new connections, the interface would be the same. And maybe another function to close the server.
I only added you to this discussion as you use websockets and nodemcu, and thus could provide an insight about the interfaces (even if it is only client side for now).

app/websocket/websocketclient.c
luismfonseca: Interesting. I supposed it could make some sense...

But at this moment this `websocketclient` isn't coupled to Lua, and I think it's better that way.

I thought about exposing some interface (`ifdefs` maybe, or a static global variable) that allowed the `malloc` to be overridden. That way, I could change it to use the luaM_malloc (which can potentially enforce a gc in desperate situations). Maybe this could also be fine tuned as well...

app/modules/websocket.c
luismfonseca: Yes, I noticed. Although I think it's too dynamic \ generic. In this case, every `on` receives a function, but their callbacks can have 1, 2 or 3 arguments depending on the type of callback.

It seems odd for me to couple a simple string instance to the number of arguments required... But maybe that's more lua-ish?

devsaurus: > It seems odd for me to couple a simple string instance to the number of arguments required... But maybe that's more lua-ish?

According to the [Lua wiki](http://lua-users.org/wiki/BindingEnumsToLua) strings are preferred.
For a more relevant example: [`mqtt.client:on()`](http://nodemcu.readthedocs.io/en/dev/en/modules/mqtt/#mqttclienton) has callbacks with different number of parameters. That's the beauty of Lua, IMO :smile:

app/websocket/websocketclient.c
devsaurus: > But at this moment this `websocketclient` isn't coupled to Lua, and I think it's better that way.

Good point, agreed. Maybe we would need a C interface for the encoder at some point.

app/modules/websocket.c
luismfonseca: Right, I also prefer module consistency.

I've updated the code to implement the callbacks this way.

app/modules/websocket.c
```diff
@@ -0,0 +1,288 @@
...
+  int handle = luaL_checkoption(L, 2, NULL, (const char * const[]){ "connection", "receive", "close", NULL });
+  if (lua_type(L, 3) != LUA_TNIL && lua_type(L, 3) != LUA_TFUNCTION && lua_type(L, 3) != LUA_TLIGHTFUNCTION) {
+    return luaL_typerror(L, 3, "function or nil");
+  }
+
+  switch (handle) {
+    case 0:
+      NODE_DBG("connection\n");
+
+      if (data->onConnection != LUA_NOREF) {
```
devsaurus: As mentioned earlier, [`luaL_unref()`](http://pgl.yoyo.org/luai/i/luaL_unref) handles `LUA_NOREF`. You can skip the "if" here and below.

docs/en/modules/websocket.md
```diff
@@ -0,0 +1,172 @@
...
+#### Example
+```lua
+ws = websocket.createClient()
+ws:on("connection", function()
+  ws:send('hello!')
+end)
+ws:connect('ws://echo.websocket.org')
+```
+
+## websocket:connect()
```
devsaurus: Please sort methods alphabetically.

devsaurus:
Apart from cosmetic topics, the final gating question IMO is: shall we accept a new module that's based on espconn?

luismfonseca:
How far away do you reckon both #1379 and #1435 are?

Like I said, I don't mind doing another MR to implement this using lwIP. I also think it will be better.
It would be nice having this version merged though.

marcelstoer:
> shall we accept a new module that's based on espconn?

👍 from my side if we have the commitment from the author / maintainer to port it to lwIP once that's ready - and we've got that here.

marcelstoer:
Btw Luís, since you just created your first NodeMCU module (I believe) can I invite you to edit / amend ["How to write a C module"](https://github.com/nodemcu/nodemcu-firmware/wiki/%5BDRAFT%5D-How-to-write-a-C-module) if you have any feedback.

luismfonseca:
Thanks for merging this!

Ok, I'll work on that. But I'd say that as it stands now, it is already helpful.

luismfonseca:
@marcelstoer I've added a small section for debugging and change the start a tiny bit, feel free to adjust as you see fit.

Issue #1406, opened 2016-07-18
MQTT client stays connected but becomes stale or unresponsive


NicolSpies:
### Expected behavior

If the MQTT connection becomes stale or unresponsive the MQTT Client should trigger the offline event and automatically reconnect if the client:connect method has been configured to auto reconnect.

### Actual behavior

1. On ESP startup MQTT client connects to the broker and subscribes to topics. Topics wit payloads can be published and received.
2. The connection is periodically confirmed by publishing a dummy message and verifying a successful return.
3. When the MQTT connection becomes stale or unresponsive the MQTT broker sends the LWT confirming the ungraceful client disconnection.
4. When this happens the client is still connected and the offline event does not fire. This is confirmed by the "already connected" message when trying to reconnect in this state.
5. If the stale connection is closed using the client:close() method, the offline event fires.
6. The auto reconnect option set for the client does not fire when the offline event has fired.
7. The client object has not been destroyed and when the client:connect method is executed again no events are triggered and the re-connection does not take place.
8. The only way to re-establish the MQTT connection is to restart the ESP and then everything happens normally until the connection becomes stale again. The stale state occurs about once every few hours.

### Test code

```Lua
-- The standard example code in the NodeMCU documentation is used
```
### NodeMCU version

Lua 5.1.4 on SDK 1.5.1(e67da894)
branch: dev
commit: adf717353a49362ef07b779b08ae333caa59b8aa
SSL: false
modules: adc,bit,cjson,enduser_setup,file,gpio,i2c,mdns,mqtt,net,node,rotary,rtctime,sntp,tmr,uart,wifi

### Hardware
ESP12-E is used in a standard configuration.

marcelstoer:
Isn't that a dup of #1394 and #1395 (which themselves seem dups)?

NicolSpies:
Yes, I have been re-testing and believe we can make this the root issue describing the sequence of events and link the other two issues that are inter related to this one

NicolSpies:
In point 6 of the actual behavior above, the auto reconnect does not fire when the MQTT connection has become stale or unresponsive, 

**HOWEVER,**

the **auto reconnect does fire and executes successfully** if the mqtt client goes offline **immediately after** a new mqtt connection has been established.

djphoenix:
@NicolSpies can I ask you to try patch from #1349? Maybe it will establish something...

NicolSpies:
[...]

PR #1445, opened 2016-08-08
MQTT memory / connection issues fix

### LINKED ISSUE #1349: MQTT Client never freed (memory leak)
### Expected behavior
After `failure` callback or `:close()` method and nullified pointer, all memory that used by client is freed (heap before creation == heap after destroy)

### Actual behavior
Memory leak

### Test code
mqtt.lua
```lua
local client = mqtt.Client( 'Test client', 120, nil, nil, 1 )
client:on('connect', function(client)
    print('CONN')
    client:subscribe("esp/mqtt/test", 0)
    client:publish("esp/mqtt/test", "test", 0, 0)
end)
client:on('message', function(client, topic, message)
    print("RECV", topic, message)
    client:close()
end)
return client
```
mqtt-unsecure.lua
```lua
dofile('mqtt.lua'):connect( "85.119.83.194", 1883,
    0, 0, nil,
    function(client, err)
        print("CONN FAIL", err)
    end
)
```
mqtt-secure.lua
```lua
dofile('mqtt.lua'):connect( "85.119.83.194", 8883,
    1, 0, nil,
    function(client, err)
        print("CONN FAIL", err)
    end
)
```

example output (disconnected):
```
> =node.heap()
43640
> dofile("mqtt-unsecure.lua")
CONN FAIL	-5
> = collectgarbage(), node.heap()
0	42696
> dofile("mqtt-secure.lua")
CONN FAIL	-5
> = collectgarbage(), node.heap()
0	41504
... after some time ...
> = collectgarbage(), node.heap()
0	41704
```
example output (wifi connection):
```
= collectgarbage(), node.heap()
0	43520
> dofile("mqtt-secure.lua")
CONN
RECV	esp/mqtt/test	test
> = collectgarbage(), node.heap()
0	43480
> dofile("mqtt-unsecure.lua")
CONN
RECV	esp/mqtt/test	test
> = collectgarbage(), node.heap()
0	43264
... after some time ...
> = collectgarbage(), node.heap()
0	43312
```
### NodeMCU version
Current master (1.5.1)

### Hardware
NodeMCU DevKit 1.0

djphoenix:
Fixes #1349, #1406.

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -125,6 +125,7 @@ static void mqtt_socket_disconnected(void *arg)    // tcp only
       mud->pesp_conn = NULL;
     }
 
+    mud->connected = false;
     luaL_unref(L, LUA_REGISTRYINDEX, mud->self_ref);
     mud->self_ref = LUA_NOREF; // unref this, and the mqtt.socket userdata will delete it self
   }
@@ -156,6 +157,14 @@ static void mqtt_socket_reconnected(void *arg, sint8_t err)
     pesp_conn->proto.tcp->local_port = espconn_port();
     socket_connect(pesp_conn);
   } else {
+#ifdef CLIENT_SSL_ENABLE
+    if (mud->secure) {
+      espconn_secure_disconnect(pesp_conn);
+    } else
+#endif
+    {
+      espconn_disconnect(pesp_conn);
+    }
     mqtt_socket_disconnected(arg);
   }
   NODE_DBG("leave mqtt_socket_reconnected.\n");
@@ -804,6 +813,9 @@ static int mqtt_delete( lua_State* L )
     c_free(mud->pesp_conn);
     mud->pesp_conn = NULL;    // for socket, it will free this when disconnected
   }
+  while(mud->mqtt_state.pending_msg_q) {
+    msg_destroy(msg_dequeue(&(mud->mqtt_state.pending_msg_q)));
+  }
 
   // ---- alloc-ed in mqtt_socket_lwt()
   if(mud->connect_info.will_topic){
@@ -873,8 +885,7 @@ static sint8 socket_connect(struct espconn *pesp_conn)
 #ifdef CLIENT_SSL_ENABLE
   if(mud->secure)
   {
-      espconn_secure_set_size(ESPCONN_CLIENT, 5120); /* set SSL buffer size */
-      espconn_status = espconn_secure_connect(pesp_conn);
+    espconn_status = espconn_secure_connect(pesp_conn);
   }
   else
 #endif
@@ -884,7 +895,7 @@ static sint8 socket_connect(struct espconn *pesp_conn)
 
   os_timer_arm(&mud->mqttTimer, 1000, 1);
 
-  NODE_DBG("leave socket_connect.\n");
+  NODE_DBG("leave socket_connect, heap = %u.\n", system_get_free_heap_size());
 
   return espconn_status;
 }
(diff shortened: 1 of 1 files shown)


marcelstoer:
We already established that it does fix the bugs indeed. So, all we need is someone to give this a formal code review.

app/modules/mqtt.c
```diff
@@ -873,8 +885,7 @@ static sint8 socket_connect(struct espconn *pesp_conn)
 #ifdef CLIENT_SSL_ENABLE
   if(mud->secure)
   {
-      espconn_secure_set_size(ESPCONN_CLIENT, 5120); /* set SSL buffer size */
```
devsaurus: You remove `espconn_secure_set_size()` here and perform it once during module initialization. I suspect that this  creates interdependencies to other modules that change the setting in the meantime. You can't be sure about the current value when execution reaches this point unless there's a global convention.

app/modules/mqtt.c
```diff
@@ -971,22 +982,21 @@ static int mqtt_socket_connect( lua_State* L )
...
-    if(mud->pesp_conn->proto.tcp)
-      c_free(mud->pesp_conn->proto.tcp);
-    mud->pesp_conn->proto.tcp = NULL;
-    c_free(mud->pesp_conn);
-    mud->pesp_conn = NULL;
+  struct espconn *pesp_conn = mud->pesp_conn;
+  if(!pesp_conn) {
+    pesp_conn = mud->pesp_conn = (struct espconn *)c_zalloc(sizeof(struct espconn));
+  } else {
+    espconn_delete(pesp_conn);
```
devsaurus: Is the state of `pesp_conn` equivalent after the two `if` branches? Can it be re-used below if it's deleted here? The API guide is not clear in this respect IMO.

app/modules/mqtt.c
```diff
@@ -971,22 +982,21 @@ static int mqtt_socket_connect( lua_State* L )
...
-  struct espconn *pesp_conn = NULL;
-  pesp_conn = mud->pesp_conn = (struct espconn *)c_zalloc(sizeof(struct espconn));
   if(!pesp_conn)
     return luaL_error(L, "not enough memory");
-
-  pesp_conn->proto.udp = NULL;
-  pesp_conn->proto.tcp = (esp_tcp *)c_zalloc(sizeof(esp_tcp));
+  if(pesp_conn->proto.udp) {
+    pesp_conn->proto.udp = NULL;
+    c_free(pesp_conn->proto.udp);
```
devsaurus: `c_free(NULL)` is a no-op. Should lines 995 and 996 be swapped?

app/modules/mqtt.c
```diff
@@ -1609,6 +1622,7 @@ static const LUA_REG_TYPE mqtt_map[] = {
 int luaopen_mqtt( lua_State *L )
 {
   luaL_rometatable(L, "mqtt.socket", (void *)mqtt_socket_map);  // create metatable for mqtt.socket
+  espconn_secure_set_size(ESPCONN_CLIENT, 4096);
```
devsaurus: Needs `#ifdef CLIENT_SSL_ENABLE`?

app/modules/mqtt.c
djphoenix: Don't see any issues here. Now it works for me, both for secure and unsecure variants.

app/modules/mqtt.c
djphoenix: Espconn have not depends to our configuration. However, it's a good point. ACK

app/modules/mqtt.c
djphoenix: Oh... Will research. It works, without leaks, but...

app/modules/mqtt.c
djphoenix: In case of set size for each connection there was some memory leaks... Maybe it fixed, will try to move back and test.

devsaurus: > In case of set size for each connection there was some memory leaks

Interesting point, would be important to know for the other modules as well.

app/modules/mqtt.c
devsaurus: Ok.

NicolSpies:
@djphoenix Hi Jury, I have been monitoring a unit the last 2 weeks **running on the existing dev without your fixes** and observe the following that might be of interest to the MQTT connection state.

1. Over a period of time the MQTT connection appears to be fine as I publish once a minute.
2. As soon as a message is received the connection drops.
3. I restart the ESP and a fresh MQTT connection is established.
4. Now messages can be received without any problems

If this phenomena corresponds to what you have already fixed, please ignore. As soon as the fix has been merged with dev I will load on the "field" ESP unit and continue the testing.

Regards

djphoenix:
Oh... Don't seen smth like this anyway. But situation should be more stable after merge.

djphoenix:
Fixed all commented places.

app/modules/mqtt.c
djphoenix: So yes, `proto` field is `union`. In case of MQTT, we will never touch UDP part. Removed entire block.

devsaurus:
Good to go, thanks Yury!

NicolSpies:
Thanks Yury, much much appreciated

PR #1450, opened 2016-08-11
Feature: HTTP following redirects

### LINKED ISSUE #1366: Feature Request: http functions could handle HTTP REDIRECT
The http.get() and other useful functions from http module could handle HTTP 30x return code (redirect) automatically (= repeat the request to the new URL provided by server in the Location header). 

Alternatively the http functions should return also the received http headers (apart from status and body) so that users could issue another request to redirected URL by themselves (when they find out that status code was 30x).

Reason for this is that we can upload IoT data directly to Google Sheets (for example) but there's a http redirect that needs to be processed.

Frankly, I'd love to get the http response headers in the http.get() callback for now. Parsing out the redirected URL from Location header would be piece of cake.

luismfonseca:
Fixes #1366 .

Maybe the HTTP module should be configurable globally. As in `http.config(config_data)`, where you could set the user agent and wether or not to follow redirections.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -20,6 +20,8 @@
 #include "httpclient.h"
 #include "stdlib.h"
 
+#define REDIRECTION_FOLLOW_MAX 20
+
 /* Internal state. */
 typedef struct request_args_t {
 	char		* hostname;
@@ -31,6 +33,7 @@ typedef struct request_args_t {
 	char		* post_data;
 	char		* buffer;
 	int		buffer_size;
+	int		redirect_follow_count;
 	int		timeout;
 	os_timer_t	timeout_timer;
 	http_callback_t callback_handle;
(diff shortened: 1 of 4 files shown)


jmattsson:
Looks pretty good, but it would be nice to have some safeguard against infinite redirects. Do you have a strategy in mind?

luismfonseca:
Well, since the module doesn't handle the state per-connection, the only option really is to add a global static variable to count the number of redirects and give up after some reasonable value (i.e. Chrome does 21).

This module should be using a state per request so it could support concurrent requests. Then it would only be another field counting the redirects followed thus far.

I can do the first option if you want.

app/http/httpclient.c
```diff
@@ -307,26 +307,56 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
...
-				if ( os_strstr( req->buffer, "Transfer-Encoding: chunked" ) )
-				{
-					int	body_size = req->buffer_size - (body - req->buffer);
-					char	chunked_decode_buffer[body_size];
-					os_memset( chunked_decode_buffer, 0, body_size );
-					/* Chuncked data */
-					http_chunked_decode( body, chunked_decode_buffer );
-					os_memcpy( body, chunked_decode_buffer, body_size );
+
+				if ( locationOffset != NULL && http_status >= 300 && http_status <= 307 ) {
```
marcelstoer: What about 308 i.e. RFC 7538?

marcelstoer:
> Maybe the HTTP module should be configurable globally. As in `http.config(config_data)`...

Yes, that'd be useful indeed.

> I can do the first option if you want.

Go for it.

app/http/httpclient.c
luismfonseca: You're right, I was only looking at the RFC 2616. I need to get with the times.

luismfonseca:
What I said before was true, the module now indeed has a session state per connection, so I just added a new variable there to count the number of redirects being followed.

I was also assuming the redirect path was a full url but it can be a relative path, so I implemented that as well.

Here are some sample requests to test all redirect types:
```
http.get("http://luisfonseca.xyz/redirect/infinite.php", nil, function(d,c) print(d,c) end, function(d,c) print(d,c) end) 
http.get("http://luisfonseca.xyz/redirect/absolute_path.php", nil, function(d,c) print(d,c) end, function(d,c) print(d,c) end) 
http.get("http://luisfonseca.xyz/redirect/relative_full_path.php", nil, function(d,c) print(d,c) end, function(d,c) print(d,c) end) 
http.get("http://luisfonseca.xyz/redirect/relative_same_folder.php", nil, function(d,c) print(d,c) end, function(d,c) print(d,c) end) 
```

docs/en/modules/http.md
```diff
@@ -13,6 +13,8 @@ Each request method takes a callback which is invoked when the response has been
 
 For each operation it is possible to provide custom HTTP headers or override standard headers. By default the `Host` header is deduced from the URL and `User-Agent` is `ESP8266`. Note, however, that the `Connection` header *can not* be overridden! It is always set to `close`.
 
+Redirects are followed automatically (note: the redirects will be followed up to a limit of 20 requests).
```
marcelstoer: HTTP redirects (HTTP status 300-308) are followed automatically up to a limit of 20 to avoid the dreaded redirect loops.

marcelstoer:
Really nice, almost good to go.

PR #1464, opened 2016-08-18
Write function in ws2812.buffer module

bogvak:

This new function allows to write string to ws2812.buffer or copy data from another ws2812.buffer,
String writing could be very useful while creating animation from saved data or from data coming from network.

Copying data from another ws2812.buffer is very useful while creating advanced animated effects, transition from one frame to other etc,

#### Format
`ws2812.buffer:write(data)`
where `data'` is `string` or another `ws2812.buffer`

#### Returns
`nil`

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -199,6 +199,49 @@ static int ws2812_new_buffer(lua_State *L) {
   return 1;
 }
 
+static int ws2812_buffer_write(lua_State* L) {
+  size_t length;
+  const char *buffer;
+
+  ws2812_buffer * destBuffer = (ws2812_buffer*)lua_touserdata(L, 1);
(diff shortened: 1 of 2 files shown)


marcelstoer:
@Alkorin what do you think of this?

Alkorin:
Hum, isn't it the same as buf:set(0, data) ? Where data is a string ?

The new thing is to copy from an existant buffer. But we could implement :set(N, buffer) instead of adding a new function no ?

bogvak:
Really sorry for long reply, sirs. You're definitely right I haven't paid enough attention that it's possible to pass full string to _buf:set_ function and it's better just to extend it to support _buffer_ passing as argument.

Will you commit me to do this?
Thanks!

Alkorin:
Do it and I will review it ;-)

marcelstoer:
> Do it and I will review it ;-)

This would go into a new PR then -> closing this one.

PR #1466, opened 2016-08-20
Set espconn's secure buffer size once at initialization

### LINKED ISSUE #1457: espconn_secure_set_size leaks discussion
As I mentioned at [MQTT PR](https://github.com/nodemcu/nodemcu-firmware/pull/1445#discussion_r74107955), sequential calls of `espconn_secure_set_size` **may** cause some leaks.

I tested this behavior on SDKs 1.5.4.1 and 2.0.0, and for both of them situation is **very** unstable (from normal pass to fatal error).

### Fix proposal

1. Choose default "secure size" like 4096 (maybe more or less) and put it in `user_config.h` for keep it user-configurable.
2. Grep all invocations of `espconn_secure_set_size` in modules and remove them.
3. Add **one** invocation right near usercode initialization.
4. ...
5. Remove entire espconn layer and move to mbedtls/lwip (WIP)

djphoenix:
Fixes #1457

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -477,7 +477,6 @@ static void ICACHE_FLASH_ATTR http_dns_callback( const char * hostname, ip_addr_
 
 		if ( req->secure )
 		{
-			espconn_secure_set_size( ESPCONN_CLIENT, 5120 ); /* set SSL buffer size */
 			espconn_secure_connect( conn );
 		} 
 		else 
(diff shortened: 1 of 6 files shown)


jmattsson:
:+1: looks good but haven't tested it

devsaurus:
:+1: testing not possible atm.

marcelstoer:
I tested HTTP and MQTT, all good (as expected).

PR #1470, opened 2016-08-24
Fix apa102 docs

### LINKED ISSUE #1469: apa102 documentation mention AGRB order but module expect ABGR
This is the example from current documentation for the apa102 module:
```lua
a = 31
g = 0
r = 255
b = 0
led_abgr = string.char(a, g, r, b, a, g, r, b) 
apa102.write(2, 3, leds_abgr)
```
Documentation says this should turn two APA102 to red, but it does turn two APA102 to green (in current dev & master).
If we use an ABGR order (as variable name suggests), it then turn two APA102 to red.

avernois:
Fixes #1469 

Documentation for APA102 says module expect channel information in AGRB order but module expects information in ABGR.
This PR change the documentation to reflect actual behaviour of the module.

- [X] This PR is compliant with the [contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md)

### DIFF
--- a/docs/en/modules/apa102.md
+++ b/docs/en/modules/apa102.md
@@ -19,13 +19,13 @@ Send ABGR data in 8 bits to a APA102 chain.
 - `data_pin` any GPIO pin 0, 1, 2, ...
 - `clock_pin` any GPIO pin 0, 1, 2, ...
 - `string` payload to be sent to one or more APA102 LEDs.
-  It should be composed from a AGRB quadruplet per element.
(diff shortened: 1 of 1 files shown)


devsaurus:
Sounds good to me.

@robertfoss as the originator of this module, do you think this is the right approach?

robertfoss:
Looks good to me!

Thanks @avernois !

PR #1492, opened 2016-09-06
Fix float math.pow when number was negative and power was odd

### LINKED ISSUE #1491: float exponentiation of negative values gives wrong result
### Expected behavior
(-0.5)^3 should result in -0.125

### Actual behavior
(-0.5)^3 results in +0.125

### Test code
```Lua
=(-0.5)^3
=(-0.5)*(-0.5)*(-0.5)
```
### NodeMCU version
branch: master
commit: 95e85c74e7310a595aa6bd57dbbc69ec3516e9c6

### Hardware
NodeMCU v2 - Lua based ESP8266

luismfonseca:
Fixes #1491.

- [X] This PR is compliant with the [contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) (if not, please describe why).

### DIFF
--- a/app/libc/c_math.c
+++ b/app/libc/c_math.c
@@ -59,6 +59,7 @@ double pow(double x, double y)
 {
     double frexp(), g, ldexp(), r, u1, u2, v, w, w1, w2, y1, y2, z;
     int iw1, m, p;
+    bool flipsignal = false;
 
     if (y == 0.0)
         return (1.0);
@@ -67,7 +68,7 @@ double pow(double x, double y)
         if (x == 0.0)
         {
             if (y > 0.0)
-                return (x);
+                return 0.0;
             //cmemsg(FP_POWO, &y);
             //return(HUGE);
         }
@@ -75,6 +76,11 @@ double pow(double x, double y)
         {
             //cmemsg(FP_POWN, &x);
             x = -x;
+
+            if (y != (int) y) { // if y is fractional, then this woud result in a complex number
+                return NAN;
+            }
+            flipsignal = ((int) y) & 1;
         }
     }
     g = frexp(x, &m);
@@ -125,7 +131,8 @@ double pow(double x, double y)
     p = 16 * m - iw1;
     z = ((((((q7 * w2 + q6) * w2 + q5) * w2 + q4) * w2 + q3) * w2 + q2) * w2 + q1) * w2;
     z = a1[p] + a1[p] * z;
-    return (ldexp(z, m));
+    double res = ldexp(z, m);
+    return flipsignal ? -res : res;
 }
 
 #if 0


luismfonseca:
I run some tests and this still doesn't look 100% ok.

Here are mi specs:
```lua
require 'mispec'

describe('math.pow(a, b)', function(it)
    it:should('return 1 if power to 0', function()
        ok(eq(math.pow(0, 0), 1))
        ok(eq(math.pow(1, 0), 1))
        ok(eq(math.pow(123, 0), 1))
    end)
    
    it:should('power to even numbers, always resulting in a positive', function()
        ok(eq(math.pow(2, 2), 4))
        ok(eq(math.pow(-2, 2), 4))
        ok(eq(math.pow(-2, 4), 16))
    end)

    it:should('power to odd numbers', function()
        --ok(eq(math.pow(2, 3), 8)) -- fails
        --ok(eq(math.pow(-2, 3), -8)) -- also fails
        ok(eq(math.pow(-0.5, 3), -0.125))
    end)
end)

mispec.run()
```

The result were:
```lua
math.pow(a, b), it should:
> 
  * return 1 if power to 0
  * power to even numbers, always resulting in a positive
  ! it failed:	mispec.lua:31: 16 is not equal to 16
stack traceback:
	mispec.lua:31: in function 'eq'
	mispec_mathpow.lua:13: in function <mispec_mathpow.lua:10>
	[C]: in function 'pcall'
	mispec.lua:112: in function <mispec.lua:109>

  * power to odd numbers
  ! it failed:	mispec.lua:31: -0.125 is not equal to -0.125
stack traceback:
	mispec.lua:31: in function 'eq'
	mispec_mathpow.lua:19: in function <mispec_mathpow.lua:16>
	[C]: in function 'pcall'
	mispec.lua:112: in function <mispec.lua:109>


Completed in 0.21 seconds.
Success rate is 33.3% (2 failed out of 3).
```

So it seems that `math.pow(2, 4) == 16` returns `false` in the `float` build. It returns `true` on the `integer` build though, so I think this function still needs some work.

devsaurus:
This change fixes the wrong calculation of `x^y` for `x < 0` and integer y. Though it still neglects fractional values of y.
The original code works around this by forcing x to positive range (which introduced the current error). Can this condition be reported to the upper caller as NaN or something else instead?

luismfonseca:
If the `x` is negative `y` is fractional, then the result is a complex number.

So I agree that it would make sense to return sort of error...

luismfonseca:
I've updated the PR, now returns `NaN` in that scenario.

devsaurus:
Works out of the box on Lua layer - thanks Luís!

PR #1499, opened 2016-09-14
Added crypto.new_hmac() feature.

jmattsson:

This is the matching HMAC feature to the existing `crypto.new_hash()` feature introduced by @ABonner some time ago. Not much new code, mostly just refactoring to make it easy to fit in.

Switched `crypto.new_hash()` (and by extension, also `crypto.new_hmac()`) to use `luaM_malloc()` instead of `os_malloc()` as per earlier discussions on memory allocation by modules.

### DIFF
--- a/app/crypto/digests.c
+++ b/app/crypto/digests.c
@@ -156,31 +156,22 @@ int ICACHE_FLASH_ATTR crypto_fhash (const digest_mech_info_t *mi,
 }
 
 
-int ICACHE_FLASH_ATTR crypto_hmac (const digest_mech_info_t *mi,
-   const char *data, size_t data_len,
-   const char *key, size_t key_len,
-   uint8_t *digest)
+void crypto_hmac_begin (void *ctx, const digest_mech_info_t *mi,
+  const char *key, size_t key_len, uint8_t *k_opad)
 {
-  if (!mi)
-    return EINVAL;
-
-  void *ctx = (void *)os_malloc (mi->ctx_size);
-  if (!ctx)
-    return ENOMEM;
-
   // If key too long, it needs to be hashed before use
+  char tmp[mi->digest_size];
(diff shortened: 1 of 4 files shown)


marcelstoer:
Would be nice if we could ship this with the next master drop but I don't have time for testing atm. @devsaurus @djphoenix @pjsg could you help out?

djphoenix:
@marcelstoer sorry, not enough time to test... I prefer to concentrate on mbedTLS which also covers all of libcrypto/libssl functions, so they'll be deprecated completely.

app/crypto/digests.c
```diff
@@ -191,16 +182,42 @@ int ICACHE_FLASH_ATTR crypto_hmac (const digest_mech_info_t *mi,
...
+   const char *key, size_t key_len,
+   uint8_t *digest)
+{
+  if (!mi)
+    return EINVAL;
+
+  struct {
+    uint8_t ctx[mi->ctx_size];
+    uint8_t k_opad[mi->block_size];
+  } *tmp = os_malloc (sizeof (*tmp));
```
devsaurus: You're not using `luaM_malloc()` here. Should this be changed as well or is handing `L` down to this layer considered bad practice?

jmattsson: Nothing in the `app/crypto` dir is Lua aware, and it didn't feel right to push Lua into that layer. At least not at this point, and definitely not piecemeal, imo.

devsaurus: I see, thanks.

marcelstoer:
Thanks Arnim, did you save the joy of pushing that "Squash and merge" button for me on purpose 😉 I have to get used to the new GitHub review model...

devsaurus:
Me too, was wondering what an approval actually triggers. Apparently nothing else beyond the :white_check_mark:

PR #1503, opened 2016-09-18
Fixes the spiffsimg issues (#1502)

### LINKED ISSUE #1502: Several issues with spiffsimg
Triggered by a question from one of our users on Stack Overflow I set out to use `spiffsimg` for the first time. I ran into 3 issues which may all be rooted in a lack of documentation - even though https://github.com/nodemcu/nodemcu-firmware/blob/master/docs/en/spiffs.md is quite extensive.

**Which `spiffsimg`**
It all started with the question from where to get `spiffsimg`. The docs don't say. I knew about [`/tools/spiffsimg`](https://github.com/nodemcu/nodemcu-firmware/tree/master/tools/spiffsimg) but I also knew about https://github.com/DiUS/spiffsimg.
I went with ours because it's newer.

**How to build**
So, I wanted to build it from `/tools/spiffsimg`.

```
$ make
cc -g -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -I. -I../../app/spiffs -DNODEMCU_SPIFFS_NO_INCLUDE --include spiffs_typedefs.h main.c ../../app/spiffs/spiffs_cache.c ../../app/spiffs/spiffs_check.c ../../app/spiffs/spiffs_gc.c ../../app/spiffs/spiffs_hydrogen.c ../../app/spiffs/spiffs_nucleus.c  -o spiffsimg
In file included from main.c:44:
In file included from ../../app/spiffs/spiffs.h:14:
../../app/spiffs/spiffs_config.h:17:10: fatal error: 'user_config.h' file not found
#include "user_config.h"
```

I fixed the include path and finally got a `spiffsimg` (plus a `spiffsimg.dSYM` folder).

**Invalid image file name**
The docs say

> `-f` specifies the filename for the disk image. '%x' will be replaced by the calculated offset of the file system.

and

> The disk image file is placed into the `bin` directory and it is named `0x<offset>-<size>.bin` where the offset is the location where it should be flashed, and the size is the size of the flash part.

Yet, the following command produced a `0x-msr.img` in the current folder (`/tools/spiffsimg`):

`$ ./spiffsimg -f 0x-msr.img -c 262144`

_-> @jmattsson @pjsg Which one is a user error, which one a bug in the code and which one a bug in the docs?_

pjsg:
Fixes #1502 

This fixes the bugs reported in #1502. In particular, now that the stadard docker container includes the
local `gcc`, the `spiffsimg` tool is always built. This will prevent any future breakage of the tool (as it will fail the build). `spiffsimg` now fails if you try and specify %x in the filename and it cannot calculate the offset.

The documentation was also improved. 

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -28,8 +28,10 @@ FLASH_USED_END = $$((0x`$(OBJDUMP) -t ../app/.output/eagle/debug/image/eagle.app
 
 all:	spiffsscript
 
-spiffsscript: remove-image
+spiffsimg/spiffsimg:
 	$(MAKE) -C spiffsimg CC=gcc
+
+spiffsscript: remove-image spiffsimg/spiffsimg
 	rm -f ./spiffsimg/spiffs.lst
 	echo "" >> ./spiffsimg/spiffs.lst
 	@$(foreach f, $(SPIFFSFILES), echo "import $(FSSOURCE)$(f) $(f)" >> ./spiffsimg/spiffs.lst ;)
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -5,6 +5,7 @@
 FSSOURCE ?= ../local/fs/
 FLASHSIZE ?= 4mb 32mb 8mb
 SUBDIRS = 
+HOSTCC ?= gcc
 
 OBJDUMP = $(or $(shell which objdump),xtensa-lx106-elf-objdump)
 
@@ -36,7 +37,7 @@ spiffsimg: spiffsimg/spiffsimg
 	@echo Built spiffsimg in spiffsimg/spiffsimg
 
 spiffsimg/spiffsimg: 
-	@$(MAKE) -C spiffsimg CC=gcc
+	@$(MAKE) -C spiffsimg CC=$(HOSTCC)
 
 spiffsscript: remove-image spiffsimg/spiffsimg
 	rm -f ./spiffsimg/spiffs.lst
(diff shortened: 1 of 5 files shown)


tools/Makefile
```diff
@@ -3,7 +3,7 @@
 #
 
 FSSOURCE ?= ../local/fs/
-FLASHSIZE ?= 4mb 32mb 8mb
+FLASHSIZE ?= 4mb 32mb
```
pjsg: Note that the 8mb flash size does not exist in common use....

marcelstoer: What is "common"? ESP8285 (e.g. PSF-A85) has 1MB flash.

pjsg: Ok -- I'll put it back and make the documentation align

tools/Makefile
```diff
@@ -28,8 +28,17 @@ FLASH_USED_END = $$((0x`$(OBJDUMP) -t ../app/.output/eagle/debug/image/eagle.app
...
-	$(MAKE) -C spiffsimg CC=gcc
+.PHONY: spiffsimg
+
+.PHONY: spiffsimg/spiffsimg
+
+spiffsimg: spiffsimg/spiffsimg
+	@echo Built spiffsimg in spiffsimg/spiffsimg
+
+spiffsimg/spiffsimg: 
+	@$(MAKE) -C spiffsimg CC=gcc
```
jmattsson: Do we have a $(HOSTCC) defined? Should we? CC=$(HOSTCC) would be clearer, and also allow for building with host clang if so specified on the command line.

pjsg: Fixed

tools/Makefile
pjsg: Weirdly, this doesn't show that the issue is now resolved.......

PR #1506, opened 2016-09-20
Add the esp-gdbstub to enable debugging

### LINKED ISSUE #1505: Add esp-gdbstub to allow for C code debugging
### Justification
Debugging the firmware is tricky without being able to use a source code debugger. Espressif 
does provide a stub that provides basic GDB support. We should support it.

I have the code in a branch and am working on it.

pjsg:
Fixes #1505.

This is mostly the esp-gdbstub with a small stub to enable it to be controlled from lua. If you don't enable the GDBSTUB module, then none of this is included. 

I don't know how robust the Espressif code is, I just copied it in (and then tweaked it slightly). It worked
fine for my use case.

### DIFF
--- a/app/esp-gdbstub/gdbstub.c
+++ b/app/esp-gdbstub/gdbstub.c
@@ -640,6 +640,7 @@ static void ATTR_GDBFN gdb_exception_handler(struct XTensa_exception_frame_s *fr
 
 static void ATTR_GDBFN gdb_flush_output_buffer() {
     if (obufpos > 0) {
+	int i;
 	gdbPacketStart();
 	gdbPacketChar('O');
 	for (i=0; i<obufpos; i++) gdbPacketHex(obuf[i], 8);
@@ -651,7 +652,6 @@ static void ATTR_GDBFN gdb_flush_output_buffer() {
 //Replacement putchar1 routine. Instead of spitting out the character directly, it will buffer up to
 //OBUFLEN characters (or up to a \n, whichever comes earlier) and send it out as a gdb stdout packet.
 static void ATTR_GDBFN gdb_semihost_putchar1(char c) {
-	int i;
 	obuf[obufpos++]=c;
 	if (c=='\n' || obufpos==OBUFLEN) {
   	    gdb_flush_output_buffer();
@@ -780,6 +780,7 @@ static void ATTR_GDBINIT install_uart_hdlr() {
 void ATTR_GDBINIT gdbstub_redirect_output(int enable) {
   if (enable) {
     os_install_putc1(gdb_semihost_putchar1);
+    uart_set_alt_output_uart0(gdb_semihost_putchar1);
   } else {
     gdb_flush_output_buffer();
     os_install_putc1(uart0_putc);
--- a/app/esp-gdbstub/gdbstub.c
+++ b/app/esp-gdbstub/gdbstub.c
@@ -13,6 +13,7 @@
 #include "c_types.h"
 #include "gpio.h"
 #include "xtensa/corebits.h"
+#include "driver/uart.h"
 
 #include "gdbstub.h"
 #include "gdbstub-entry.h"
@@ -89,8 +90,8 @@ extern void xthal_set_intenable(int);
 #define UART_TXFIFO_CNT 0x000000FF
 #define UART_TXFIFO_CNT_S                   16
 #define UART_FIFO( i )                          (REG_UART_BASE( i ) + 0x0)
-#define UART_INT_ENA(i)                     (REG_UART_BASE(i) + 0xC)
-#define UART_INT_CLR(i)                 (REG_UART_BASE(i) + 0x10)
+//#define UART_INT_ENA(i)                     (REG_UART_BASE(i) + 0xC)
+//#define UART_INT_CLR(i)                 (REG_UART_BASE(i) + 0x10)
 #define UART_RXFIFO_TOUT_INT_ENA            (BIT(8))
 #define UART_RXFIFO_FULL_INT_ENA            (BIT(0))
 #define UART_RXFIFO_TOUT_INT_CLR            (BIT(8))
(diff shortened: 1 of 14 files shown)


docs/en/modules/gdbstub.md
```diff
@@ -0,0 +1,59 @@
+# gdbstub Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2016-09-18 | [Philip Gladstone](https://github.com/pjsg) | [Philip Gladstone](https://github.com/pjsg) | [gdbstub.c](../../../app/modules/gdbstub.c)|
+
+This module provides basic source code debugging of the firmware when used in conjunction with a version of gdb built for the lx108. If you
```
jmattsson: lx108? You mean the lx106, right? 108 is ESP32 territory.

docs/en/modules/gdbstub.md
```diff
@@ -0,0 +1,59 @@
...
+```
+gdb bin/firmwarefile.bin
+target remote /dev/ttyUSB0
+```
+
+At this point, you can just poke around and see what happened, but you cannot continue execution.
+
+In order to do interactive debugging, add a call to `gdbstub.brk()` in your lua code. This will trigger a break instruction and
+will trap into gdb as above. However, continuation is supported from a break instruction and so you can single step, set breakpoints, etc.
+Note that the lx108 processor as configured by Espressif only supports a single hardware breakpoint -- so this means that you
```
jmattsson: I'm sure the ESP8266 is an LX106...

docs/en/modules/gdbstub.md
```diff
@@ -0,0 +1,59 @@
...
+
+## gdbstub.brk()
+Enters gdb by executing a `break 0,0` instruction.
+
+#### Syntax
+`gdbstub.brk()`
+
+## gdbstub.gdboutput()
+Controls whether system output is encapsulated in gdb remote debugging protocol. This turns out not to be as useful as you
+would hopei -- mostly because you can't send input to the nodemcu board. Also becuase you really only should make this call
```
jmattsson: Typos `hopei` and `becuase`.

docs/en/modules/gdbstub.md
```diff
@@ -0,0 +1,59 @@
...
+*after* you get gdb running and connected to the nodemcu. The example below first does the break and then switches to 
+redirect the output. This works (but you are unable to send any more console input).
```
jmattsson: Would it be possible to call the corresponding C functions through gdb to in effect send Lua commands? Or is it not possible to execute random functions via gdb? If it is possible, an example of how to do that would be great.

app/esp-gdbstub/gdbstub.c
```diff
@@ -0,0 +1,805 @@
...
+  	    gdb_flush_output_buffer();
+	}
+}
+
+#if !GDBSTUB_FREERTOS
+//The OS-less SDK uses the Xtensa HAL to handle exceptions. We can use those functions to catch any 
+//fatal exceptions and invoke the debugger when this happens.
+static void ATTR_GDBINIT install_exceptions() {
+	int i;
+	int exno[]={EXCCAUSE_ILLEGAL, EXCCAUSE_SYSCALL, EXCCAUSE_INSTR_ERROR, EXCCAUSE_LOAD_STORE_ERROR,
```
jmattsson: Note that the EXCCAUSE_LOAD_STORE_ERROR will never be caught here, because we have wrapped _xtos_set_exception_handler to explicitly intercept that code, since we're using that for our non-32bit load handler (see app/user/user_exceptions.c). It might be worth adding a note to such effect here.

app/esp-gdbstub/gdbstub.c
```diff
@@ -0,0 +1,805 @@
...
+//We use a small hack to replace that with a jump to our own handler, which then has the task of
+//decyphering and re-instating the registers the FreeRTOS code left.
+extern void user_fatal_exception_handler();
+extern void gdbstub_user_exception_entry();
+
+static void ATTR_GDBINIT install_exceptions() {
+	//Replace the user_fatal_exception_handler by a jump to our own code
+	int *ufe=(int*)user_fatal_exception_handler;
+	//This mess encodes as a relative jump instruction to user_fatal_exception_handler
+	*ufe=((((int)gdbstub_user_exception_entry-(int)user_fatal_exception_handler)-4)<<6)|6;
```
jmattsson: Ermagerrrd! Teh horrorz!

Cool hack though.

app/modules/gdbstub.c
```diff
@@ -0,0 +1,39 @@
+/*
+ * Module for doing various horrible things
```
jmattsson: Technically, I can't argue with that statement, but something a teensy bit more descriptive might be nice :)

app/esp-gdbstub/gdbstub.c
pjsg: Actually, it does work -- when this code runs, it ends up calling our exceptions code which remembers the custom handler for LOAD_STORE_ERROR. If our handler decides not to handle the LOAD_STORE exception, then it chains onto the one that was set.

app/esp-gdbstub/gdbstub.c
pjsg: This file isn't my code -- it came from espressif!

app/modules/gdbstub.c
pjsg: My bad -- I started off with a small private module that I wrote and I forgot to change the comment!

app/esp-gdbstub/gdbstub.c
jmattsson: Yeah yeah, I know :) It's very special code nevertheless!

app/esp-gdbstub/gdbstub.c
jmattsson: Ah, I see now. I didn't have chaining support in there originally and had forgotten that got added. I sit corrected!

PR #1526, opened 2016-10-04
Reimplemented esp_init_data_default.

jmattsson:
Fixes #1500.

Reimplemented esp_init_data_default to work around the pesky "rf_cal[0] !=0x05" hang when booting on a chip which doesn't have esp_init_data written to it.

It is no longer possible to do the writing of the esp_init_data_default
from within nodemcu_init(), as the SDK now hangs long before it gets
there.  As such, I've had to reimplement this in our user_start_trampoline
and get it all done before the SDK has a chance to look for the init data.
It's unfortunate that we have to spend IRAM on this, but I see no better
alternative at this point.

With this patch, NodeMCU should finally be back to "it just works" when flashed onto a fresh(ly erased) ESP, and the previously documented approach should be correct once more.

I'd really appreciate it if a few more people could try this PR before considering merging - while it's working fine for me, I'm always a bit hesitant messing with the early boot stuff.

### DIFF
--- a/app/user/user_main.c
+++ b/app/user/user_main.c
@@ -47,6 +47,49 @@ void TEXT_SECTION_ATTR user_start_trampoline (void)
   rtctime_early_startup ();
 #endif
 
+  /* Re-implementation of default init data deployment. The SDK does not
+   * appear to be laying down its own version of init data anymore, so
+   * we have to do it again. To see whether we need to, we read out
+   * the flash size and do a test for esp_init_data based on that size.
+   * If it's missing, we need to initialize it *right now* before the SDK
+   * starts up and gets stuck at "rf_cal[0] !=0x05,is 0xFF".
+   * If the size byte is wrong, then we'll end up fixing up the init data
+   * again on the next boot, after we've corrected the size byte.
+   * Only remaining issue is lack of spare code bytes in iram, so this
+   * is deliberately quite terse and not as readable as one might like.
+   */
+  SPIFlashInfo sfi;
+  SPIRead (0, (uint32_t *)(&sfi), sizeof (sfi)); // Cache read not enabled yet, safe to use
+  if (sfi.size < 2) // Compensate for out-of-order 4mbit vs 2mbit values
+    sfi.size ^= 1;
+  uint32_t flash_end_addr = (256 * 1024) << sfi.size;
+  uint32_t init_data_hdr = 0xffffffff;
+  uint32_t init_data_addr = flash_end_addr - 4 * SPI_FLASH_SEC_SIZE;
+  SPIRead (init_data_addr, &init_data_hdr, sizeof (init_data_hdr));
+  if (init_data_hdr == 0xffffffff)
+  {
+    // Note: ideally we should generate the actual init data from the
+    // SDK's bin/esp_init_data_default.bin during the build, and #include it
+    // straight into this array. E.g.:
+    //   xxd -i esp_init_data_default | tail -n13 | head -n11 > eidd.h
+    static uint8_t flash_init_data[128] __attribute__((aligned(4))) =
+    {
+      0x05, 0x00, 0x04, 0x02, 0x05, 0x05, 0x05, 0x02, 0x05, 0x00, 0x04, 0x05,
+      0x05, 0x04, 0x05, 0x05, 0x04, 0xfe, 0xfd, 0xff, 0xf0, 0xf0, 0xf0, 0xe0,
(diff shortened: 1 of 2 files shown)


jimparis:
Cool, I didn't realize we were able to run code before the SDK error.  Definitely a good solution!

I tested 5 cases:

1. Erase, flash as 2MB chip and flash init data to 0x1fc000 -- works
2. Erase, flash as 4MB chip and flash init data to 0x3fc000 -- works
3. Erase, flash as 2MB chip, don't flash init data -- works
4. Erase, flash as 4MB chip, don't flash init data -- works
5. No erase after test 4, Flash as 2MB chip, don't flash init data -- works

:+1:

app/user/user_main.c
```diff
@@ -47,6 +47,49 @@ void TEXT_SECTION_ATTR user_start_trampoline (void)
...
+  uint32_t flash_end_addr = (256 * 1024) << sfi.size;
+  uint32_t init_data_hdr = 0xffffffff;
+  uint32_t init_data_addr = flash_end_addr - 4 * SPI_FLASH_SEC_SIZE;
+  SPIRead (init_data_addr, &init_data_hdr, sizeof (init_data_hdr));
+  if (init_data_hdr == 0xffffffff)
+  {
+    // Note: ideally we should generate the actual init data from the
+    // SDK's bin/esp_init_data_default.bin during the build, and #include it
+    // straight into this array. E.g.:
+    //   xxd -i esp_init_data_default | tail -n13 | head -n11 > eidd.h
```
pjsg: Is there a reason that you didn't do this?

pjsg:
I tried this, and it worked in my case.....

app/user/user_main.c
jmattsson: @pjsg Time.

app/user/user_main.c
```diff
@@ -47,6 +47,49 @@ void TEXT_SECTION_ATTR user_start_trampoline (void)
...
+    static uint8_t flash_init_data[128] __attribute__((aligned(4))) =
```
marcelstoer: If this is indeed what I think it is then it's the same I complained about in #1500. The init data bytes shouldn't be hardcoded (will soon be outdated) but read from the SDK which is downloaded during the build. @vowstar fixed this in his PR with https://github.com/nodemcu/nodemcu-firmware/pull/1500/commits/7f3e4b8257875b3786b2793cff59bd0ecd44aa93.

marcelstoer:
I'm currently on the road and can't test but did one of you verify if the "old firmware" (0.9.x, 1.4, etc.) -> 1.5.4.1 upgrade runs smoothly? IIRC there were some fundamental init data changes between 1.5.1 and 1.5.4.1.

jimparis:
Upgrading from old firmware would still require an erase first.  This code only covers the case where the beginning of the init data is blank.  I don't think you can do anything more generic without running the risk of overwriting a user's custom init data.

app/user/user_main.c
jimparis: See  #1527

app/user/user_main.c
jimparis: See #1527

jmattsson:
This PR is superseded by @jimparis' #1527, closing.

jmattsson:
@marcelstoer Upgrade from 1.5.1 to 1.5.4.1 worked fine here.

Tested:
  1. Full chip erase.
  2. Flashed NodeMCU 1.5.1
  3. Let it boot.
  4. Flashed this branch.
  5. Booted fine, it just had to reformat the filesystem.

marcelstoer:
> running the risk of overwriting a user's custom init data

Can we reliably determine the size (i.e. length) of the init data block on the device? If so then we can compare it against the SDKs init block size and still overwrite if they don't match.

jmattsson:
No, the init data block is always 128 bytes, but there's no (sane) way to differentiate between a custom init block and an old SDK one.

If we had heaps of IRAM we could keep copies of all previous SDK defaults and compare against each, but we don't, so i think #1527 is as good as we're going to get at this point.

marcelstoer:
That would indeed be a waste of space if all we care about is whether the defaults were changed or not. So, if we could guard all functions that alter init data we'd only need a single bit of IRAM.

jmattsson:
To my knowledge there are no functions which alter the init data (other than our initial installation thereof). Custom init data typically gets flashed at the same time (or soon after) the NodeMCU image is flashed to the chip, and there is no way for us to detect that short of a full comparison. And even then, if the user explicitly wants/needs the values from e.g. SDK 1.4.0, we wouldn't be able to tell that intent apart from someone who just upgraded from a 1.4.0 build and would like to have the init data updated.

PR #1525, opened 2016-10-03
Rework flashing instructions to clarify need for SDK init data

jimparis:
Addresses stuff raised in #1500.

- [x] This PR is for the `dev` branch rather than for `master`.
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [x] I have thoroughly tested my contribution.
- [x] The code changes are reflected in the documentation at `docs/en/*`.

Since 1.5.4.1, flashing `esp_init_data_default.bin` is no longer optional, even on a fully erased chip.  Update documentation to clarify that it must be flashed anytime it is missing, or whenever the SDK version changes.  This restructures the document a bit in the process.

```bash
# Testing with and without flashing esp_init_data_default.bin on an ESP-WROOM-02:

git checkout 1.5.1-master_20160603
make clean
make
esptool.py erase_flash
esptool.py write_flash -m qio -fs 32m 0x00000 bin/0x00000.bin 0x10000 bin/0x10000.bin
# Boots NodeMCU 1.5.1 successfully

git checkout 1.5.4.1-master_20161001
make clean
make
esptool.py erase_flash
esptool.py write_flash -m qio -fs 32m 0x00000 bin/0x00000.bin 0x10000 bin/0x10000.bin
# Prints error "rf_cal[0] !=0x05,is 0xFF", fails to boot to NodeMCU

git checkout 1.5.4.1-master_20161001
make clean
make
esptool.py erase_flash
esptool.py write_flash -m qio -fs 32m 0x00000 bin/0x00000.bin 0x10000 bin/0x10000.bin 0x3fc000 ~/sdk-1.5.4.1/bin/esp_init_data_default.bin
# Boots NodeMCU 1.5.4.1 successfully
```

PR #1527, opened 2016-10-04
Reimplement esp_init_data_default without hardcoded data

jimparis:
Fixes #1500.

This builds upon #1526 from @jmattsson and replaces the hardcoded init data with contents from the SDK.


dnc40085:
Tested on an esp-12(32m), works as advertised.

jmattsson:
Hah, I wasn't aware of the `.incbin` directive. Learned something new, thanks :)

:+1: from me.

marcelstoer:
I suggest that the `flash.md` changes from #1525 be merged into this PR for consistency reasons. They need to be merged at the same time anyway.

jimparis:
Merged flash.md changes from #1525 and updated it to fit the current situation (which means that a lot of the detail was taken out, since NodeMCU users shouldn't have to care about it).

marcelstoer:
Well done, thanks a lot @jimparis! However, I see no value in removing the init data addresses (https://github.com/nodemcu/nodemcu-firmware/pull/1527/files#diff-48f98dba177e62bebd22ee90be1ee586L79).

vowstar:
Thanks, very cool solution!

PR #1530, opened 2016-10-06
Smaller lua buffer -- reducing stack requirements

### LINKED ISSUE #1523: Reduce the size of the luaL_Buffer object
### Missing feature
The current luaL_Buffer consumes 1k bytes of stack space. This is a lot on the ESP8266.

### Justification
Not much stack available!

Nearly all the code deals with the buffer size not being large enough. This issue is to fix any errors and then reduce the buffersize to (say) 128. Once the issues are fixed, then it *should* be possible to run with much smaller buffer sizes (e.g. 4) and the system should still work -- albeit a bit more slowly.

pjsg:
Fixes #1523 

Essentially this just changes the #define for the size of the internal LUA buffer used in luaL_Buffer objects. I only found one uses where the input size was uncontrolled. The onewire case was fixed by limiting the number of bytes to read to be the same as LUAL_BUFFERSIZE. 

The other case interesting case is in read_line in liolib.c It did limit the read to the buffer size (which is now smaller). However, the callers would not be happy with a shorter line if it wasn't properly terminated. I fixed this function to read to the end of the line and return the resulting string. It isn't clear that that is the right solution. 

I tried running the code and it *appeared* to work the same way as before. In particular I could reverse long strings (much bigger than 1k).

### DIFF
--- a/app/lua/liolib.c
+++ b/app/lua/liolib.c
@@ -343,28 +343,19 @@ static int read_line (lua_State *L, int f) {
 static int read_line (lua_State *L, int f) {
   luaL_Buffer b;
   luaL_buffinit(L, &b);
-  char *p = luaL_prepbuffer(&b);
-  signed char c = EOF;
-  int i = 0;
-  do{
+  signed char c;
+  do {
     c = (signed char)vfs_getc(f);
-    if(c==EOF){
+    if (c==EOF) {
       break;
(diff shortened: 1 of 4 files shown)


TerryE:
This is a simpler solution than my suggestion and reduces the problem by 4x.

PR #1535, opened 2016-10-10
Fixes the gpio.serout problem from #1534

### LINKED ISSUE #1534: Clean up async support in gpio.serout
### Expected behavior

You should be able to do another gpio.serout inside a prior serout callback.

### Actual behavior

It doesn't work

pjsg:
Fixes #1534 

This fixes the order of freeing of callbacks when calling them. It also cleans up the documentation a bit. It also removes a NODE_DBG call from the interrupt handler. It cleans up some potential memory leaks, but I suspect that there may be more if the user tries to have multiple serout calls running at once.

### DIFF
--- a/docs/en/modules/gpio.md
+++ b/docs/en/modules/gpio.md
@@ -75,12 +75,13 @@ The function works in two modes:
 * synchronous - for sub-50 µs resolution, restricted to max. overall duration,
 * asynchrounous - synchronous operation with less granularity but virtually unrestricted duration.
 
-Whether the asynchronous mode is chosen is defined by presence of the `callback` parameter. If present and is of function type the function goes asynchronous the callback function is invoked when sequence finishes. If the parameter is numeric the function still goes asynchronous but no callback is invoked when done.
+Whether the asynchronous mode is chosen is defined by presence of the `callback` parameter. If present and is of function type the function goes asynchronous and the callback function is invoked when sequence finishes. If the parameter is numeric the function still goes asynchronous but no callback is invoked when done.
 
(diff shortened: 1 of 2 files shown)


docs/en/modules/gpio.md
```diff
@@ -80,17 +80,17 @@ Whether the asynchronous mode is chosen is defined by presence of the `callback`
 For asynchronous version minimum delay time should not be shorter than 50 μs and maximum delay time is 0x7fffff μs (~8.3 seconds).
 In this mode the function does not block the stack and returns immediately before the output sequence is finalized. HW timer inf `FRC1_SOURCE` mode is used to change the states.
 
-Note that the synchronous variant (no or nil `callback` parameter) function blocks the stach and as such any use of it must adhere to the SDK guidelines (also explained [here](https://nodemcu.readthedocs.io/en/dev/en/extn-developer-faq/#extension-developer-faq)). Failure to do so may lead to WiFi issues or outright to crashes/reboots. Shortly it means that sum of all delay times multiplied by the number of repeats should not exceed 15 ms.
+Note that the synchronous variant (no or nil `callback` parameter) function blocks the stach and as such any use of it must adhere to the SDK guidelines (also explained [here](https://nodemcu.readthedocs.io/en/dev/en/extn-developer-faq/#extension-developer-faq)). Failure to do so may lead to WiFi issues or outright to crashes/reboots. Shortly it means that sum of all delay times multiplied by the number of cycles should not exceed 15 ms.
```
devsaurus: stack not 'stach'

pjsg: Fixed. I also cleaned up some other english style issues and made the extn-developers-faq link point to the right place (relative rather than absolute)

PR #1538, opened 2016-10-12
Add generic function to print deprecation notes

marcelstoer:
Follow-up to #1533.

This is my first cut. Happy to get your feedback.

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -151,6 +151,7 @@ static timer_t tmr_get( lua_State *L, int stack ) {
 	if (lua_isuserdata(L, stack)) {
 		return (timer_t)luaL_checkudata(L, stack, "tmr.timer");
 	} else {
+		platform_print_deprecation_note("Use the dynamic timer API rather than the static one.", "end of 2016");
 		uint32_t id = luaL_checkinteger(L, 1);
 		luaL_argcheck(L, platform_tmr_exists(id), 1, "invalid timer index");
 		return &alarm_timers[id];
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -151,7 +151,6 @@ static timer_t tmr_get( lua_State *L, int stack ) {
 	if (lua_isuserdata(L, stack)) {
 		return (timer_t)luaL_checkudata(L, stack, "tmr.timer");
 	} else {
-		platform_print_deprecation_note("Use the dynamic timer API rather than the static one.", "end of 2016");
 		uint32_t id = luaL_checkinteger(L, 1);
 		luaL_argcheck(L, platform_tmr_exists(id), 1, "invalid timer index");
 		return &alarm_timers[id];
--- a/app/platform/platform.c
+++ b/app/platform/platform.c
@@ -859,3 +859,8 @@ uint32_t platform_flash_mapped2phys (uint32_t mapped_addr)
   uint32_t meg = (b1 << 1) | b0;
   return mapped_addr - INTERNAL_FLASH_MAPPED_ADDRESS + meg * 0x100000;
 }
+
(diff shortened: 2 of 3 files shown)


app/modules/tmr.c
```diff
@@ -151,6 +151,7 @@ static timer_t tmr_get( lua_State *L, int stack ) {
 	if (lua_isuserdata(L, stack)) {
 		return (timer_t)luaL_checkudata(L, stack, "tmr.timer");
 	} else {
+		platform_print_deprecation_note("Use the dynamic timer API rather than the static one.", "end of 2016");
```
djphoenix: Maybe it will be helpful to print exact method (or at least module) that generates warning, for users who migrates large existing project with lots of code.

app/platform/platform.c
```diff
@@ -859,3 +859,8 @@ uint32_t platform_flash_mapped2phys (uint32_t mapped_addr)
   uint32_t meg = (b1 << 1) | b0;
   return mapped_addr - INTERNAL_FLASH_MAPPED_ADDRESS + meg * 0x100000;
 }
+
+void* platform_print_deprecation_note( const char *msg, const char *time_frame)
```
djphoenix: Nice feature: print warning only once for each deprecated invocation per boot. There may be tons of useless logs instead.

marcelstoer: I wouldn't want to keep track of whether a particular `msg` has already been printed or not. Besides, printing once only isn't enough IMO.

pjsg:
I think that printing this on every invocation is overkill. If we are to do this, then we must provide migration instructions from using the static timers to using dynamic ones. 

For example, what is the replacement for:

tmr.alarm(0, 1000, 1, function() do_something() end)

Note that I do not have to keep track of the timer object -- I an always get it back by using id 0. In particular, running the same call again will end up with ONE timer running rather than two. This is incredibly useful when developing code. 

It also isn't clear (to me) when timers get garbage collected. Does this happen when the application loses track of the timer object, or does it have to wait until the timer has stopped firing and the application has lost track of it?

marcelstoer:
The main problem I myself have with this PR is the mix of introducing a generic function to print deprecation notes _plus_ the use of that function for the timer module. Also, as Philip noted it isn't entirely clear yet how `tmr` _without_ that static functions is going to look like for all cases.

My proposal:
- cut the scope of this PR by removing the invocation of the new function
- create a new issue (for @djphoenix?) to discuss the removal of static timers in detail 

Agree?

app/modules/tmr.c
marcelstoer: Or did you mean `platform_print_deprecation_note` should generically detect who the caller is (module and function) and print that? How could that be done? Please clarify.

djphoenix: @marcelstoer if there can be a macro, maybe good resolution is __FUNC__ / __FILE__ pseudo-variable can be used. So don't remember how them exactly named, but in macro-way there are options.

marcelstoer:
> remove `time_frame`.

I absolutely see your point. However, I added it exactly to give us "positive motivation" 😉 to remove deprecated stuff rather that leaving deprecated for ages. We'd be free to update the time frame any time but a message including a time component would still convey a certain commitment from us.

app/platform/platform.c
jmattsson: It depends on much of a nuisance we want to make of our deprecation notes. I could even see an argument made for a two-stage approach, where initially we only whinge once, but closer to the date we keep whinging. Quite possibly over-kill though...

If you want to keep track of messages already printed, the easy option would be a singly-linked list which stores the msg pointer as the value. If you're wanting to be really tight you could probably use less than the standard 32bits, but that would not be warranted unless we have lots of deprecations in flight, which I hope we won't.

PR #1557, opened 2016-10-23
Make the MQTT PING functionality work better.

### LINKED ISSUE #1553: mqtt keep alive timer is not correct
### Expected behavior

When the esp8266 loses wifi communication it should activate the m:on("offline" ... after 120 Seconds (according to the following example)
### Actual behavior

The offline event activates randomly and after 10-15 minutes
### Test code

``` Lua
-- init mqtt client with keepalive timer 120sec
m = mqtt.Client("clientid", 120, "user", "password")

-- setup Last Will and Testament (optional)
-- Broker will publish a message with qos = 0, retain = 0, data = "offline" 
-- to topic "/lwt" if client don't send keepalive packet
m:lwt("/lwt", "offline", 0, 0)

m:on("connect", function(client) print ("connected") end)
m:on("offline", function(client) print ("offline") end)
```
### NodeMCU version

master
### Hardware

esp8266-12e

pjsg:
Fixes #1553.

When the keepalive PING is sent, the client is supposed to wait for up to the keepalive timer for a PINGRESP. If it doesn't get it, then the connection should be aborted. This now happens.

It should be noted that the timer passed in to the Client constructor is one keepalive time. It takes two expirations of this for the offline event to be called (one to send the PINGREQ, and one to not receive the response).

I also cleaned up the spaces in this file -- which were rather a mess. To see the diff without the whitespace changes, look at https://github.com/nodemcu/nodemcu-firmware/pull/1557/files?w=1

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -72,6 +72,7 @@ typedef struct lmqtt_userdata
   uint8_t secure;
 #endif
   bool connected;     // indicate socket connected, not mqtt prot connected.
+  bool keepalive_sent;
   ETSTimer mqttTimer;
   tConnState connState;
 }lmqtt_userdata;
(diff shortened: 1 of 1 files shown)


NicolSpies:
@pjsg, I have 3 units in the field connected via 3G that experience MQTT disconnects due to the connection. I will test on the units. Unfortunately can only get to them in a few days to upload new build.

PR #1435, opened 2016-08-03
Upgrade to SDK 2.0.0

djphoenix:
Fixes #1409, #1415, #1624, #1604

PR #1558, opened 2016-10-24
Fix for httpclient resetting the system (maybe)

### LINKED ISSUE #1546: HTTP request with "large" response body fails or resets device
This is a follow-up to #1485 and a more-to-the-point description of #1545. Seems like a serious regression bug in the HTTP module or some SDK characteristic the HTTP module relies upon.

### Expected behavior
HTTP GET to an HTTP resource should return the HTTP resource.

### Actual behavior
Depending on the response size (I believe) this either works, fails or resets the device.

### Test code
```Lua
http.get("http://nodemcu-build.com", nil, function(code, data)
  if (code < 0) then
    print("HTTP request failed")
  else
    print(code, data)
  end
end)
```
- http://frightanic.com/jquery-highlight.js ✅ , 2.7 KB
- http://frightanic.com 🔴 , 17.7 KB
- http://nodemcu-build.com 🔴 ♻️ , 6.6 KB
- http://nodemcu-build.com/about.php 🔴 ♻️ , 2.8 KB
- http://nodemcu-build.com/jumbotron-narrow.css ✅ , 433 B
- http://www.google.com, 🔴 , 66.9 KB
- http://clients5.google.com/pagead/drt/dn/, ✅ , 270 B
- http://httpbin.org 🔴 , 12.1 KB
- http://httpbin.org/get ✅ , 798 B

🔴 : fails
♻️ : reset

### NodeMCU version
Current `master` and `dev` with SDK 1.5.4.1. According to #1545 this worked fine with 1.5.1. Did the SDK used to buffer responses in earlier versions?

### Hardware
NodeMCU devkit 1.0.

pjsg:
Partially Fixes #1546.

The issue was that the reconcb called the disconnect cb. This was bad as it freed up all the control blocks and didn't disconnect the connection. I switched it over to called the timeout callback which does try and disconnect the connection and which then does the cleanup. 

I'm not entirely convinced that this is the right fix -- but I'm putting it out there in case anybody wants to build on it.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -411,13 +411,6 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
 }
 
 
-static void ICACHE_FLASH_ATTR http_error_callback( void *arg, sint8 errType )
-{
-	HTTPCLIENT_DEBUG( "Disconnected with error\n" );
-	http_disconnect_callback( arg );
-}
-
-
 static void ICACHE_FLASH_ATTR http_timeout_callback( void *arg )
 {
 	HTTPCLIENT_DEBUG( "Connection timeout\n" );
@@ -436,6 +429,13 @@ static void ICACHE_FLASH_ATTR http_timeout_callback( void *arg )
 		espconn_secure_disconnect( conn );
 	else
 		espconn_disconnect( conn );
+}
+
+
+static void ICACHE_FLASH_ATTR http_error_callback( void *arg, sint8 errType )
+{
+	HTTPCLIENT_DEBUG( "Disconnected with error: %d\n", errType );
+	http_timeout_callback( arg );
 }
 
 


pjsg:
I'll split it out into a seperate PR and try and move most of the various invocations of debug printfs over to it. The big win is that it uses hardly any stack.....

pjsg:
I cut this back to just the http client change. It appears to not crash and not leak memory. However, there still are problems in getting https urls from certain places.

marcelstoer:
All good from my side (I tested the URLs listed in the issue).

How about https://github.com/nodemcu/nodemcu-firmware/issues/1546#issuecomment-255258990?
> I realized that there are a number of `HTTPCLIENT_DEBUG` messages (not just "response too long") which shouldn't be _debug_ messages but regular error messages i.e. printed unconditionally.

Should that be a separate PR?

pjsg:
I suspect that the whole module needs a cleanup -- I'd rather leave that for later. Maybe it isn't worth doing until the mbed tls stuff is in....

marcelstoer:
> I suspect that the whole module needs a cleanup

I can't really judge, a number of people improved it since its inception. Yuri already [switched to mbedTLS](https://github.com/nodemcu/nodemcu-firmware/pull/1435#issuecomment-255750472) over in his branch but `http.c` is not affected. So, I don't think we have to wait.

Changing a debug message into a regular output to console is something I feel is in my reach 😉 I'll be submitting a PR then.

PR #1564, opened 2016-10-26
Version of printf that doesn't suffer from buffer overflows

pjsg:

I found this issue while trying to debug the httpclient issues. It turns out that if you try and log a long string with the existing printf implementation, then you get a stack overflow and everything dies. This is exactly the behavior that you don't want when you enable debugging. Also, from earlier experiments, it turns out that the printf implementation uses a fair amount of stack and this causes more issues when running with DEBUG enabled. This implementation solves both problems by outputting characters directly as they are generated. The downside is that it isn't a very complete version of printf -- but that is fine for debugging.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -13,6 +13,7 @@
  */
 
 #include "osapi.h"
+#include "../libc/c_stdio.h"
 #include "user_interface.h"
 #include "espconn.h"
 #include "mem.h"
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -645,10 +645,10 @@ void ICACHE_FLASH_ATTR http_put( const char * url, const char * headers, const c
 
 void ICACHE_FLASH_ATTR http_callback_example( char * response, int http_status, char * full_response )
 {
-	os_printf( "http_status=%d\n", http_status );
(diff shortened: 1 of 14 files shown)


marcelstoer:
Thanks Philip, can you please fix the build: https://travis-ci.org/nodemcu/nodemcu-firmware/builds/170759429#L5309-L5316

pjsg:
Rats!

Will do

TerryE:
@pjsg Philip, I've just been going through this and what we are doing here is modifying the lua core in  a very tactical way to solve a point issue.  This as an approach increases our code complexity and future maintenance load.  I think that this as a strategy merits further discussion.

The root issue is that the code lua code uses `printf()` calls in a limited number of places.  ( Lua itself does its own print format conversion to implement it own format specifiers and otherwise `puts()` to print this content).  The libraries also use `printf()` calls for debug.  

The current approach to printing is stack heavy, so enabling debug printing and cause a step change in stack usage compounding complexity of the debug task.

If we think that this stack issue is severe enough to merit reimplementing printf, then we should do that, and properly.  What we have here is a tactical fix to a point problem being cascaded into the Lua core.  I feel that in the long term, this is going in the wrong direction.

PR #1572, opened 2016-10-29
Improve SNTP module: Add list of servers and auto-sync

### LINKED ISSUE #1543: Improve sntp.sync to support a list of servers
### Missing feature

The `sntp.sync` should be able to take a list of servers to try, and then pick the best one.
### Justification

It is what other systems do.
### Workarounds

Yeah -- I've built the code.

pjsg:
Fixes #1543.

This is a large PR. It messes with the SNTP module and the RTCTIME modules. The RTCTIME module is used to keep track of the wall clock time. It (essentially) counts the clock ticks of the CPU, but these don't actually happen at the right frequency. Hence this PR allows fractional adjustment of the rate. In fact, on the systems that I have tested, the clock is pretty close to being right (with 5ppm). 

The SNTP module has two big changes. The first is to support passing in a list of servers (rather than a single server) and then it finds the best server and it uses that time. It also does multiple NTP packet exchanges so that the occasional delay (maybe an ARP round trip) doesn't cause problems.

The second change is to have an auto-sync mode where the sync operation is performed every 20 minutes or so. In this mode, because the time updates are regular, the clock can be conditioned and the frequency adjusted. 

More information is returned to the success callback so that interested people (aka time-nuts) can get more information about what is going on.

The end result is that the clock stays within a millisecond (or two) of the nominal time. Actually working out what the absolute offset is is quite difficult. NTP synchronized time has an offset when the clocks are down a cable modem (due to the difference in the upstream and downstream latency). 

This probably shouldn't be merged before the next drop to master.


marcelstoer:
It'd of course be ideal if the guys from DiUS (@jmattson et.al.) could review this as they contributed the affected modules..

jmattsson:
Whoa, lots of changes! I'll see if I can get @umisef to have a peek at the rtctime stuff.

From memory we have other modules which also use rtcmem slots, and I think those modules may need to be updated to not conflict with the additional slots used here now.

pjsg:
You are right -- this is a bit of a monster. I'm going to split it into a number of smaller PRs. The first one will be to just handle the list of names and other associated changes to sntp. The improved timekeeping can wait till later.

PR #1573, opened 2016-10-30
support for custom websocket headers

mrk-its:
This PR adds support for optional parameter 'headers' in websocket.client:connect. It is implemented similarly to http module, for example headers string must include '\r\n' line terminators.

Also:
- fix to allow for '\0's in received messages
- support lowercase headers in server response (currently connection fails if server responds with perfectly valid lowercase 'sec-websocket-accept' header)

Above fixes together are necessary to connect to Amazon's MQTT via websocket gateway.

### DIFF
--- a/app/modules/websocket.c
+++ b/app/modules/websocket.c
@@ -120,7 +120,6 @@ static int websocketclient_on(lua_State *L) {
   NODE_DBG("websocketclient_on\n");
 
   ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);  
-  luaL_argcheck(L, ws, 1, "Client websocket expected");
 
   ws_data *data = (ws_data *) ws->reservedData;
 
@@ -172,7 +171,6 @@ static int websocketclient_connect(lua_State *L) {
   NODE_DBG("websocketclient_connect is called.\n");
 
   ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);  
-  luaL_argcheck(L, ws, 1, "Client websocket expected");
 
   ws_data *data = (ws_data *) ws->reservedData;
 
@@ -207,7 +205,6 @@ static int websocketclient_config(lua_State *L) {
   NODE_DBG("websocketclient_config is called.\n");
 
   ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);
-  luaL_argcheck(L, ws, 1, "Client websocket expected");
 
   ws_data *data = (ws_data *) ws->reservedData;
 
@@ -246,7 +243,6 @@ static int websocketclient_send(lua_State *L) {
   NODE_DBG("websocketclient_send is called.\n");
 
   ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);  
-  luaL_argcheck(L, ws, 1, "Client websocket expected");
 
   ws_data *data = (ws_data *) ws->reservedData;
 
@@ -279,7 +275,6 @@ static int websocketclient_gc(lua_State *L) {
   NODE_DBG("websocketclient_gc\n");
 
   ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);  
-  luaL_argcheck(L, ws, 1, "Client websocket expected");
 
   ws->extraHeaders = realloc_headers(ws->extraHeaders, 0);
 
--- a/app/modules/websocket.c
+++ b/app/modules/websocket.c
@@ -45,7 +45,7 @@ static void websocketclient_onConnectionCallback(ws_info *ws) {
   }
 }
 
-static void websocketclient_onReceiveCallback(ws_info *ws, char *message, int opCode) {
+static void websocketclient_onReceiveCallback(ws_info *ws, int len, char *message, int opCode) {
   NODE_DBG("websocketclient_onReceiveCallback\n");
 
   lua_State *L = lua_getstate();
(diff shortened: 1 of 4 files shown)


marcelstoer:
@luismfonseca can we ask to review this?

djphoenix:
@mrk-its I think there is great improvements, but there is also some notes:
- Why not to use Lua tables to pass headers?
- It seems that you expect to get headers with proper case (e.g. `User-Agent`), but for webservers header names are case-insensitive. Maybe better to parse them properly? (With Lua tables it well be simple)

mrk-its:
@djphoenix
- I even planned to implement this with Lua tables, but then I saw in http module headers passed as single string, so I decided to be consistent with that. Second thing - underlying websocketclient library is Lua-independent, so passing headers as single string simplified implementation a bit.
- You are right - It would be better to parse them. Lets agree on API (headers passed via Lua tables vs single string) and I'll fix that.

luismfonseca:
Hi,

It does look good, I'll review it in depth and test this later on today :).

As for the Lua tables, I think you can use that on the Lua interface, and then send to the C side of things an array of array (a create new struct for this?). That way you still keep Lua out of the native implementation part.

Also, I was thinking that this could have its own function, i.e.: `websocket.config(params)`, where you could pass a table, like you do in `wifi.ap.config()` for example. My reasoning being that you'd usually want to use the same `User Agent` for all your websocket connections, and it could also be later expanded to support more configuration (i.e. timeouts).

But it might be an overkill, what do you think?

marcelstoer:
`websocket.config(params)` good idea, the user agent case would be reason enough for that.

mrk-its:
@luismfonseca @marcelstoer Ok, I'll add websocket.config(params), params.headers - table of headers. Thanks!

docs/en/modules/websocket.md
```diff
@@ -70,6 +66,27 @@ ws = nil -- fully dispose the client as lua will now gc it
...
+## websocket.client:config(params)
+
+Configures websocket client instance.
+
+#### Syntax
+`websocket:config(params)`
+
+#### Parameters
+- `params` table with configuration parameters. Following keys are recognized:
+  - `headers` table of extra request headers
```
marcelstoer: For absolute clarity please add here that these headers affect every request.

app/websocket/websocketclient.c
```diff
@@ -550,12 +580,31 @@ static void connect_callback(void *arg) {
   char *key;
   generateSecKeys(&key, &ws->expectedSecKey);
 
-  char buf[WS_INIT_HEADERS_LENGTH + strlen(ws->path) + strlen(ws->hostname) + strlen(key)];
-  int len = os_sprintf(buf, WS_INIT_HEADERS, ws->path, ws->hostname, ws->port, key);
+  header_t headers[] = {
+	  {"Upgrade", "websocket"},
+	  {"Connection", "Upgrade"},
+	  {"Sec-Websocket-Key", key},
```
luismfonseca: Nitpick: `Sec-WebSocket-Key`

luismfonseca:
It looks great, 👍.

app/websocket/websocketclient.c
```diff
@@ -509,7 +539,7 @@ static void ws_initReceiveCallback(void *arg, char *buf, unsigned short len) {
   }
 
   // Check server has valid sec key
-  if (strstr(buf, WS_HTTP_SEC_WEBSOCKET_ACCEPT) == NULL || strstr(buf, ws->expectedSecKey) == NULL) {
```
mrk-its: Hmm, I've removed this because checking presence of WS_HTTP_SEC_WEBSOCKET_ACCEPT was case-sensitive (for example it not worked for Amazon's MQTT websocket gateway which sends this header key lowercase). We still check if this header value (SHA-1 hash) is valid, so it is not big deal I think.
To do this properly we probably should parse all response headers and store them in header_t structs, instead of using this strstr search.

app/websocket/websocketclient.c
```diff
@@ -133,6 +127,37 @@ static void generateSecKeys(char **key, char **expectedKey) {
...
+      va_start(args2, buf);
+      for(header_t *header_set2 = va_arg(args2, header_t *); header_set2; header_set2 = va_arg(args2, header_t *))
+        for(header_t *header2 = header_set2; header2->key; header2++) {
+          if(header == header2)
+            goto ok;
+          if(!strcasecmp(header->key, header2->key))
+            goto skip;
+        }
+ok:
+      len += os_sprintf(buf + len, "%s: %s\r\n", header->key, header->value);
```
mrk-its: Are you sure about high stack usage in above os_sprintf call? os_sprintf writes to provided buffer, formatting string is very simple, I cannot imagine how it can use a lot of stack memory. Can you explain / point some docs?

app/modules/websocket.c
```diff
@@ -183,8 +185,59 @@ static int websocketclient_connect(lua_State *L) {
...
+  if(!new_size)
+    return NULL;
+  return (header_t *)c_malloc(sizeof(header_t) * (new_size + 1));
+}
+
+static int websocketclient_config(lua_State *L) {
+  NODE_DBG("websocketclient_config is called.\n");
+
+  ws_info *ws = (ws_info *) luaL_checkudata(L, 1, METATABLE_WSCLIENT);
+  luaL_argcheck(L, ws, 1, "Client websocket expected");
```
mrk-its: You are right - luaL_checkudata throws an error, I've removed following luaL_argcheck call (for this and 4 other similar occurrences)

app/websocket/websocketclient.c
mrk-its: OK, I've just replaced this os_sprintf call with simple string copy - @pjsg can you take a look now?

marcelstoer:
We're already past the 2-weeks-before-the-next-master-drop milestone but I feel this is pretty low risk. So, no problem. I added the proper milestone tag.

PR #1575, opened 2016-10-31
Add support to mix ws2812.buffer objects. 

### LINKED ISSUE #1574: Add support to ws2812.buffer to allow mixing...
### Missing feature

Easier operations on ws2812 buffers -- such as mixing
### Justification

In lighting, you often want to crossfade from one lighting setup to another. Having a builtin function can make this very efficient.
### Workarounds

No.

pjsg:
Fixes #1574 

This adds a number of new methods on the ws2812.buffer object. 

mix -- mix together a number of buffers in adjustable rations
power -- get the total power consumption of a buffer (useful if your power supply is limited)
load/dump -- convert a buffer to and from a string (in the raw string format)
__tostring -- format the buffer in a way that is useful for debug printing

Updated the documentation to match. Also improved the verification of the user data object.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -269,12 +269,21 @@ static int ws2812_buffer_shift(lua_State* L) {
   const int shiftValue = luaL_checkinteger(L, 2);
   const unsigned shift_type = luaL_optinteger( L, 3, SHIFT_LOGICAL );
 
-  luaL_argcheck(L, shiftValue > 0-buffer->size && shiftValue < buffer->size, 2, "shifting more elements than buffer size");
+  ptrdiff_t start = posrelat(luaL_optinteger(L, 4, 1), buffer->size);
+  ptrdiff_t end = posrelat(luaL_optinteger(L, 5, -1), buffer->size);
+  if (start < 1) start = 1;
+  if (end > (ptrdiff_t)buffer->size) end = (ptrdiff_t)buffer->size;
+
+  start--;
+  int size = end - start;
+  size_t offset = start * buffer->colorsPerLed;
+
+  luaL_argcheck(L, shiftValue > 0-size && shiftValue < size, 2, "shifting more elements than buffer size");
 
   int shift = shiftValue >= 0 ? shiftValue : -shiftValue;
 
   // check if we want to shift at all
-  if (shift == 0)
+  if (shift == 0 || size <= 0)
   {
     return 0;
   }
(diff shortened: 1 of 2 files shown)


app/modules/ws2812.c
```diff
@@ -132,8 +131,7 @@ static int ws2812_write(lua_State* L) {
   }
   else if (type == LUA_TUSERDATA)
   {
-    ws2812_buffer * buffer = (ws2812_buffer*)lua_touserdata(L, 1);
-    luaL_argcheck(L, buffer && buffer->canary == CANARY_VALUE, 1, "ws2812.buffer expected");
```
devsaurus: What was the purpose of this `CANARY_VALUE` anyway?

app/modules/ws2812.c
```diff
@@ -230,11 +224,10 @@ static int ws2812_buffer_fill(lua_State* L) {
 }
 
 static int ws2812_buffer_fade(lua_State* L) {
-  ws2812_buffer * buffer = (ws2812_buffer*)lua_touserdata(L, 1);
+  ws2812_buffer * buffer = (ws2812_buffer*)luaL_checkudata(L, 1, "ws2812.udata");
```
devsaurus: What is `ws2812.udata` - did you mean `ws2812.buffer`?

app/modules/ws2812.c
```diff
@@ -260,11 +253,10 @@ static int ws2812_buffer_fade(lua_State* L) {
 
 
 static int ws2812_buffer_shift(lua_State* L) {
-  ws2812_buffer * buffer = (ws2812_buffer*)lua_touserdata(L, 1);
+  ws2812_buffer * buffer = (ws2812_buffer*)luaL_checkudata(L, 1, "ws2812.udata");
```
devsaurus: `ws2812.buffer`?

docs/en/modules/ws2812.md
```diff
@@ -181,6 +181,87 @@ The number of given bytes must match the number of bytesPerLed of the buffer
...
+#### Returns
+A string containing the pixel values. 
+
+#### Example
+```lua
+local s = buffer:dump() 
+```
+
+## ws2812.buffer:load()
+Converts a string into a buffer (see also the `dump` method). The buffer must have the correct shape or an error will be thrown.
```
devsaurus: "correct shape" means that length has to match exactly? Or could the loaded string be smaller, filling only a portion of the buffer (with an optional index as offset)?

docs/en/modules/ws2812.md
```diff
@@ -181,6 +181,87 @@ The number of given bytes must match the number of bytesPerLed of the buffer
...
+buffer:mix(256 - crossmix, buffer1, crossmix, buffer2)
+
+-- multiplies all values in buffer by 0.75
+-- This can be used in place of buffer:fade
+buffer:mix(192, buffer)
+```
+
+## ws2812.buffer:power()
+Computes the total energy requirement for the buffer. This is merely the total sum of all the pixel values (which assumes that each color in each
+pixel consumes the same amount of power).
```
devsaurus: The datasheets don't provide any clue on this aspect. Do you have further information/links that back this? Same for the assumption that pixel values translate linearly to current consumption.

app/modules/ws2812.c
```diff
@@ -387,24 +462,59 @@ static int ws2812_buffer_set(lua_State* L) {
...
 
   lua_pushnumber(L, buffer->size);
 
   return 1;
 }
 
+static int ws2812_buffer_tostring(lua_State* L) {
+  ws2812_buffer * buffer = (ws2812_buffer*)luaL_checkudata(L, 1, "ws2812.buffer");
+
+  luaL_Buffer result;
```
devsaurus: Does this implementation work for RGB chains longer than 85 (aka `LUAL_BUFFERSIZE / 3`)?

app/modules/ws2812.c
pjsg: It was an attempt (I think) to verify that the right type of userdata was provided. I switched over to using the standard method.

docs/en/modules/ws2812.md
pjsg: The WS2812 is just three leds with a WS2811 chip. The datasheet for this chip (e.g. https://cdn-shop.adafruit.com/datasheets/WS2811.pdf ) indicates that each channel (R, G, B) is a 20ma constant current drive. The brightness is controlled by a variable duty cycle (per channel). I *assumed* that the duty cycle scaled linearly with the pixel value (it would be more difficult to to otherwise). This means that the *average* current is colorvalue / 255 * 20mA.

docs/en/modules/ws2812.md
pjsg: I've changed this to be more general purpose and allow an insert operation into the middle of another buffer. Also, the source can either be a buffer or a string.

app/modules/ws2812.c
pjsg: Yes. It works on my 144 led strip that is sitting on my kitchen table. Even my daughter said "Ooo!" (I have a couple of led chasers running)

devsaurus:
Cool stuff that you're adding, Philip :heart_eyes:
Let us know when you're done...

pjsg:
I should be done by friday -- I'm building an animated display as part of a hackathon project at work. It'll either work or not by Friday morning. Then I'l be done with ws2812!

pjsg:
I'm good with this. I beat on it pretty hard yesterday and didn't find any issues.

marcelstoer:
Then let's roll it in 😃

PR #1585, opened 2016-11-08
Update flash.md

robertio:
Fixes #\<GitHub-issue-number\>.

Make sure all boxes are checked (add x inside the brackets) when you submit your contribution, remove this sentence before doing so.

- [ ] This PR is for the `dev` branch rather than for `master`.
- [ ] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [ ] I have thoroughly tested my contribution.
- [ ] The code changes are reflected in the documentation at `docs/en/*`.

\<Description of and rational behind this PR\>
Hi, 
please update Flashing section with my two comments:

Flashing tricks for ESP-01 and ESP-12 from different forum posts like http://www.esp8266.com/viewtopic.php?f=23&t=11233&start=4 and http://www.esp8266.com/viewtopic.php?p=52965#p52965

PR #1587, opened 2016-11-09
Overhaul flashing docs once again

marcelstoer:
Addresses https://github.com/nodemcu/nodemcu-firmware/pull/1585#issuecomment-259273835.

### DIFF
--- a/docs/en/flash.md
+++ b/docs/en/flash.md
@@ -1,4 +1,4 @@
-Adafruit provides a really nice [firmware flashing tutorial](https://learn.adafruit.com/building-and-running-micropython-on-the-esp8266/flash-firmware).  Below you'll find just the basics for the two popular tools esptool and NodeMCU Flasher.
+Below you'll find all necessary information to flash a NodeMCU firmware binary to ESP8266 or ESP8285. Note that this is a reference documentation and not a tutorial with fancy screen shots. Turn to your favorite search engine for those. Make sure you follow a recent tutorial rather than one that is several months old!
 
 !!! attention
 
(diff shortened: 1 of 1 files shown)


docs/en/flash.md
```diff
@@ -28,7 +28,9 @@ Run the following command to flash an *aggregated* binary as is produced for exa
 - `size` is given in bits. Specify `4m` for 512&nbsp;kByte and `32m` for 4&nbsp;MByte.  If unsure, try a smaller size;
```
pjsg: Modern versions of the esptool.py actually auto-detect the flash size, so it is probably best to leave out the flash size. This functionality was added about a month ago. We may want to note the flash size issues in some notes somewhere.

docs/en/flash.md
```diff
@@ -41,6 +43,8 @@ Source: [https://github.com/nodemcu/nodemcu-flasher](https://github.com/nodemcu/
 
 Supported platforms: Windows
 
+Note that this tool was created by the initial developers of the NodeMCU firmware. It hasn't seen updates since September 2015 and is not maintained by the current NodeMCU *firmware* team. Be careful not to accidentally flash the very old default firmware the tool is shipped with.
```
pjsg: This tool is actively maintained by espressif (I followed the link to it higher up in the document): https://github.com/espressif/esptool

docs/en/flash.md
```diff
@@ -58,33 +62,57 @@ Otherwise, if you built your own firmware from source code:
...
 
-* The NodeMCU flash filesystem may need to be reformatted, particularly if its address has changed because the new firmware has a much different size than the old firmware.  If it is not automatically formatted, then it should be valid and have the same contents as before the flash operation. You can still run [`file.format()`](modules/file.md#fileformat) to re-format your flash filesystem. You will know if you need to do this if your flash files exist but seem empty, or if data cannot be written to new files. However, this should be an exceptional case.
+* The NodeMCU flash file system may need to be reformatted, particularly if its address has changed because the new firmware is different in size from the old firmware.  If it is not automatically formatted then it should be valid and have the same contents as before the flash operation. You can still run [`file.format()`](modules/file.md#fileformat) manually to re-format your flash file system. You will know if you need to do this if your flash files exist but seem empty, or if data cannot be written to new files. However, this should be an exceptional case.
```
marcelstoer: Ok, I can certainly do that.

docs/en/flash.md
marcelstoer: Oh nice, just found the [PR](https://github.com/espressif/esptool/pull/131), thanks. Don't you think it'd make sense to mention that only for [esptool.py >= 1.2](https://github.com/espressif/esptool/releases) you can skip the flash size parameter?

pjsg: I think that I would recommend leaving the parameter out. However, you should probably note that the esptool is under active development and that if it doesn't work, the first step should be to upgrade to the current version.

marcelstoer:
Philip, I processed your feedback and pushed a new version.

marcelstoer:
I take your thumbs-up as a go for merging.

pjsg:
Yes!

PR #1591, opened 2016-11-14
Add note on dev board usage of SPI bus 0

mdpye:
- [x] (NA) I have thoroughly tested my contribution.
- [x] (NA) The code changes are reflected in the documentation at `docs/en/*`.

Documentation only.

I spent time and effort researching the mapping from bus number to the SPI/HSPI pin groups and also ran in to confusion with bus 0 being already in use by my dev board. This note is what I wish I'd found when first opening this module.

### DIFF
--- a/docs/en/modules/spi.md
+++ b/docs/en/modules/spi.md
@@ -6,6 +6,16 @@
 All transactions for sending and receiving are most-significant-bit first and least-significant last.
 For technical details of the underlying hardware refer to [metalphreak's ESP8266 HSPI articles](http://d.av.id.au/blog/tag/hspi/).
 
+## A note on the SPI busses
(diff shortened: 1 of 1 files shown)


jmattsson:
:+1:

marcelstoer:
Should we also add a back-reference to this new chapter from each `id` parameter description ("SPI ID number: 0 for SPI, 1 for HSPI")? Something like "see note on bus IDs".

luismfonseca:
Back-reference too much noise, given how short the document is.

Maybe highlight this using the `note` syntax:
```
!!! note
    some test

```

marcelstoer:
Ok, I'll add that after the merge.

PR #1596, opened 2016-11-17
Improve SNTP module: Add list of servers and auto-sync [SNTP module only]

### LINKED ISSUE #1543: Improve sntp.sync to support a list of servers

pjsg:
Fixes #1543 (partial).

This is a large PR. It messes with the SNTP module.

The SNTP module has two big changes. The first is to support passing in a list of servers (rather than a single server) and then it finds the best server and it uses that time. It also does multiple NTP packet exchanges so that the occasional delay (maybe an ARP round trip) doesn't cause problems.

The second change is to have an auto-sync mode where the sync operation is performed every 20 minutes or so. In this mode, because the time updates are regular, the clock can be conditioned and the frequency adjusted (this feature will be enabled in a later PR) 

More information is returned to the success callback so that interested people (aka time-nuts) can get more information about what is going on.

This probably shouldn't be merged before the next drop to master.

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -47,6 +47,8 @@
 #include "rtc/rtctime.h"
 #endif
 
+#define max(a,b) ((a < b) ? b : a)
+
 #define NTP_PORT 123
 #define NTP_ANYCAST_ADDR(dst)  IP4_ADDR(dst, 224, 0, 1, 1)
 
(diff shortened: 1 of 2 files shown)


vsky279:
Fixes also issue (https://github.com/nodemcu/nodemcu-firmware/issues/1635).

PR #1601, opened 2016-11-18
Improve ROTable lookups -- DO NOT MERGE -- WIP

### LINKED ISSUE #1590: Some simple operations are mind-bogglingly slow
### Expected behavior

Invoking `tmr.now()` should be quick.

### Actual behavior

It actually takes hundreds of microseconds!

### Test code

```Lua
a = tmr.now()
b = tmr.now()
print ('delay', b - a)
```

It turns out that finding `tmr` is very slow. The code

```Lua
now = tmr.now
a = now()
b = now()
print ('Delay', b - a)
```

works much better. However, it doesn't seem reasonable for everybody to do this for all module accesses. 

It seems that improving the lookups should be doable. (Spoiler alert: Yes, it is possible. The trick is to do it without burning RAM)

pjsg:
Fixes #1590 

This is the current state of the branch that speeds up rotable lookups. For example the line

a = tmr.now

gets about a factor 10 faster. 

There are a lot of things going into this, see the discussion in #1590 for more information.

Yes -- this PR isn't done -- there are probably debug printfs in there still enabled etc.

This now builds (and runs) in both integer and float environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ ifdef DEBUG
   CCFLAGS += -ggdb -O0
   LDFLAGS += -ggdb
 else
-  CCFLAGS += -Os
+  CCFLAGS += -O2 
 endif
 
 #############################################################
@@ -185,7 +185,10 @@ DEP_LIBS_$(1) = $$(foreach lib,$$(filter %.a,$$(COMPONENTS_$(1))),$$(dir $$(lib)
 DEP_OBJS_$(1) = $$(foreach obj,$$(filter %.o,$$(COMPONENTS_$(1))),$$(dir $$(obj))$$(OBJODIR)/$$(notdir $$(obj)))
 $$(IMAGEODIR)/$(1).out: $$(OBJS) $$(DEP_OBJS_$(1)) $$(DEP_LIBS_$(1)) $$(DEPENDS_$(1))
 	@mkdir -p $$(IMAGEODIR)
-	$$(CC) $$(LDFLAGS) $$(if $$(LINKFLAGS_$(1)),$$(LINKFLAGS_$(1)),$$(LINKFLAGS_DEFAULT) $$(OBJS) $$(DEP_OBJS_$(1)) $$(DEP_LIBS_$(1))) -o $$@ 
+	$$(CC) $$(LDFLAGS) $$(if $$(LINKFLAGS_$(1)),$$(LINKFLAGS_$(1)),$$(LINKFLAGS_DEFAULT) $$(OBJS) $$(DEP_OBJS_$(1)) $$(DEP_LIBS_$(1))) -o $$@.tmp 
+	python $(TOP_DIR)/tools/addhashtables.py $$@.tmp > /tmp/hashtables.c
+	cd /tmp; $$(CC) -c hashtables.c
+	$$(CC) $$(LDFLAGS) $$(if $$(LINKFLAGS_$(1)),$$(LINKFLAGS_$(1)),$$(LINKFLAGS_DEFAULT) $$(OBJS) $$(DEP_OBJS_$(1)) $$(DEP_LIBS_$(1))) /tmp/hashtables.o -o $$@
 endef
 
 $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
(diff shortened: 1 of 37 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ ifdef DEBUG
   CCFLAGS += -ggdb -O0
   LDFLAGS += -ggdb
 else
-  CCFLAGS += -Os
+  CCFLAGS += -O2
```
pjsg: This change is huge -- it causes the constant strings to be 32-bit aligned.

TerryE: > This change is huge 

It is a slight increase in generated flash size, but I think that you mean that the number of non-aligned fetch exceptions is dramatically less.  Wouldn't it be better to disable this one specific size optimatisation since there are other differences between these two options.

pjsg: It turns out that I can't find the option that does the alignment of strings. There are only three reported differences (controlled by -f flags) between -Os and -O2. One of these is -finline-functions that we already override to off. The other two looked really obscure.

marcelstoer:
Looks like the CI build failed for the same reasons @djphoenix started #1566.

ld/nodemcu.ld
```diff
@@ -5,7 +5,7 @@ MEMORY
   dport0_0_seg :                        org = 0x3FF00000, len = 0x10
   dram0_0_seg :                         org = 0x3FFE8000, len = 0x14000
   iram1_0_seg :                         org = 0x40100000, len = 0x8000
-  irom0_0_seg :                         org = 0x40210000, len = 0x80000
+  irom0_0_seg :                         org = 0x40210000, len = 0x90000
```
marcelstoer: Nope, it fails because i**ram**1_0_seg is too small, not irom1_0_seg. Hence the reference to #1566 which originates from https://github.com/nodemcu/nodemcu-firmware/pull/1435#issuecomment-255995454.

djphoenix:
Yep. I will play with this PR and my LD patch tomorrow to find the way to fix all of this... anyway now we have two different reasons to review iram contents.

app/modules/wifi_eventmon.c
```diff
@@ -334,6 +335,7 @@ static const LUA_REG_TYPE wifi_event_monitor_reason_map[] =
   { LSTRKEY( "HANDSHAKE_TIMEOUT" ),        LNUMVAL( REASON_HANDSHAKE_TIMEOUT ) },
   { LNILKEY, LNILVAL }
 };
+LUA_TABLE_REG_2(wifi_event_monitor_reason_map);
 #endif
 
 const LUA_REG_TYPE wifi_event_monitor_map[] =
```
dnc40085: Shouldn't this have the `LUA_TABLE_REG_x()` around it as well?

pjsg: Good catch -- but actually no. It turns out that the table actually has to be defined in wifi.c as that is where the reference is. I've added a comment to the code to note this.

marcelstoer:
Philip, how much of a WIP is this? Would it be good to merge once you fix the conflicts?

pjsg:
Let me take a look at this. I'm back from a summer of vacations (and will soon have got over my jetlag). Is the ROTable implementation the same in 5.3 or is this mostly only useful in 5.1?

TerryE:
>  Is the ROTable implementation the same in 5.3 or is this mostly only useful in 5.1?

No, it's different. As I discuss in my paper, what I've done is to move towards a more unified approach for handling ROtables and RWtables at the higher Lua API levels.  However, I also ended up keeping the eLua linear vector implementation for luaR_entry vectors and a backwards compatibility mode, simply because too many of the add-on module would need major rework if we didn't and IMO, this would be a real barrier to the adoption of Lua 5.3.  

I adopted a different approach to ROtable acceleration: since Lua 5.3 already implements a small look-aside cache for accelerating repeated Cstring to TString lookup which works well, I decided to overload this with an equivalent look-aside functionality for ROtable entry resolution.  Testing in medium sized  Lua apps gives a rough 95% hit rate in my Lua test harness scenarios,and this pretty much eliminates the additional overhead of ROM lookup from doing ROM table linear scans and lots of strcmps.

pjsg:
Given the fact that this won't be used in the 5.3 code, I'm not sure that it makes sense to clean and polish it.

PR #1627, opened 2016-11-30
Next 1.5.4.1 master drop

marcelstoer:
Please approve.


pjsg:
Do we need to fix the iram/rom section contents? I was playing with a custom build last night with a lot of modules and the system didn't boot. It turned out that I could fix this by making one of the data structures in the websocketclient.c code into a static. I didn't try a plain dev build with all modules, but it is definitely worth a try.....

marcelstoer:
This was for #1566, right? Not sure...I'd vote against it. It's a rather far-reaching change which fixes something that is not critical (yet) for the large majority of our developers. 
I planned to merge it into `dev` (together with the EUS fix) right after this master drop.

pjsg:
I certainly want to try a clen dev build with lots of modules and see if it boots on a real nodemcu before we merge this one.

djphoenix:
I definitely vote for merge #1566 **after** master drop (so with sdk2.0.0), because this is huge change that should be tested in Dev as long as it can. For master drop - I vote up for this, it seems this is stable right now.

pjsg:
I now can't reproduce my problems from last night.....

devsaurus:
Shouldn't this have been a merge commit instead of a squash?

marcelstoer:
Oh bloody h... of course it should have been! I'm really sorry. Is this bad enough to warrant reverting?

devsaurus:
Github's compare for a PR from dev to master looks bad at the moment. The common baseline of the two branches isn't automatically detected and all commits since the last drop are scheduled again.
If feasible, this should be reverted IMO.

marcelstoer:
Hhmm, wouldn't know how to fix this properly. The tag on `master` could be dropped and this PR merge could be reverted, no problem. However, since we had merges on `dev` in the meantime we couldn't simply replay this PR, could we? We'd snap the current state of `dev` - unless all commits are cherry-picked.

marcelstoer:
Maybe this would be the way to go: http://stackoverflow.com/a/28497372/131929

pjsg:
You could revert the commit to master. Then do a PR from the correct commit on dev to master and merge that. It should work as those commits will not be on master (and never were on master).

devsaurus:
The revert branch will contain the original commits or just the squashed commit?

Like Philip proposed, we could create a new branch eg. `dev_for_drop` based on 85c3a24927b439a48a609ab8d3a4750bf8ca768a and start a PR for this.

pjsg:
The revert will just be a single commit that undoes the squashed commit.

marcelstoer:
Ok, hold on tight. Let's see if I manage to fix my mess...

marcelstoer:
Ok, it's all over and it didn't even hurt that much. Sorry again for all the trouble. I'll add a fat reminder to https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#handling-releases for next time.

Issue #1258, opened 2016-04-19
Executing HTTP request followed by another HTTP request


aadeon:
I wanted to execute a HTTP request (using http module) and when a response comes, immediately execute another request. The code is more or less is like this:

```
url = "http://google.com"

function firstRequest(callback)
  print("First request, connecting to " .. url)
  http.get(url, nil,
    function(code, data)
        print("First response: " .. code .. ", data: " .. data)
        callback()
    end)
end

function secondRequest()
  print("Second request, connecting to " .. url)
  http.get(url, nil,
    function(code, data)
        print("Second response: " .. code .. ", data: " .. data)
    end)
end


firstRequest(secondRequest)
```

Unfortunately the second request is never executed, in the output I always get:

```
First request, connecting to http://google.com
> First response: 302, data: ....
Second request, connecting to http://google.com
```

The response from the second request never comes back. What am I doing wrong?

I suspect that when the callback function of the first request is being executed it is not possible to "initiate" another HTTP request. If it is the cause, what is the proper way of achieveing such simple sequence of requests?

marcelstoer:
> I suspect that when the callback function of the first request is being executed it is not possible to "initiate" another HTTP request.

I never tried myself but why not, have you tried? This all being asynchronous and event-driven I'd expect that to work just fine.

aadeon:
There is output of the script I attached, so don't you think it's very likely that I tried? ;)

marcelstoer:
Grrr... mea culpa, sorry about that. I didn't study your example carefully enough and particularly had missed the last line (at that time I didn't think it was a complete example). Pity this seems to fall under #719 and should be discussed elsewhere.

aadeon:
Hi Marcel,

Thanks for your hint, it works!
I'm just wondering what is the explanation of such behaviour? I would like to avoid such pitfalls in the future...

aadeon:
Another question: after flashing custom firmware build with http://nodemcu-build.com/ (*including* `node` module), `node.task` seems to be nil. I used `dev` branch. Has it been removed or any special module has to be selected to make it work?

PR #1629, opened 2016-12-01
HTTP module can now chain requests

### LINKED ISSUE #1628: http module has warning in docs about callback
### Expected behavior

Making another http request from an http callback ought to work. Yes, the documentation says it won't, but that is crazytalk.

### Actual behavior

The second callback doesn't fir.

### NodeMCU version
dev

The root cause is that the http module clears the callback reference *after* the callback rather than before.

While this code is open, please add the http response headers as an argument to the callback. Also, try and reduce the memory consumption so that larger responses can be handled.

pjsg:
Fixes #1628 

This frees up the buffer and request control blocks before the callback is invoked. This ensures that another http request can be made from the callback. It also has the desirable side effect of eliminating one RAM copy of the response once the LUA code is entered. You still need two copies in memory for a short period of time.

Additionally, this adds a third argument to the callback which is a table containing the response headers with the keys being the header name converted to lower case.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -402,9 +402,20 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
 		}
 		if ( req->callback_handle != NULL ) /* Callback is optional. */
 		{
-			req->callback_handle( body, http_status, req->buffer );
-		}
-		http_free_req( req );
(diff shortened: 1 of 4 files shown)


pjsg:
This also fixes #1258 (which is the same issue).

jmattsson:
:+1:

Issue #1633, opened 2016-12-02
crypto.fhash() causes reboot when hash an empty file


kicer:
According to the documentation crypto.fhash() should:

> Compute a cryptographic hash of a a file.

Returns

> A binary string containing the message digest. To obtain the textual version (ASCII hex characters), please use crypto.toHex()

But when I try hash an empty file, will causes reboot:

```
> file.open("empty.lua", "w+")   --create empty file
> file.close()
> print(crypto.toHex(crypto.fhash("MD5", 'empty.lua'))) --SHA1 also causes reboot

 ets Jan  8 2013,rst cause:2, boot mode:(3,7)

load 0x40100000, len 24684, room 16 
tail 12
chksum 0x46
ho 0 tail 12 room 4
load 0x3ffe8000, len 2204, room 12 
tail 0
chksum 0xe6
load 0x3ffe889c, len 8, room 8 
tail 0
chksum 0x52
csum 0x52

NodeMCU 1.5.4.1 build unspecified powered by Lua 5.1.4 on SDK 1.5.4.1(39cb9a32)
lua: cannot open init.lua
```

md5sum on PC's:
```
$ md5sum /tmp/empty.file
d41d8cd98f00b204e9800998ecf8427e  /tmp/empty.file
```

Version used:
branch: dev
commit: 11ded3f, **the #1592 has Fixed**
modules: _default_+**crypto**

kicer:
@pjsg, After #1647 patched, fhash() empty file works as expected. Thanks~

```
> print(crypto.toHex(crypto.fhash("MD5", 'empty.lua')))
d41d8cd98f00b204e9800998ecf8427e
> print(crypto.toHex(crypto.fhash("SHA1", 'empty.lua')))
da39a3ee5e6b4b0d3255bfef95601890afd80709
> print(crypto.toHex(crypto.fhash("SHA256", 'empty.lua')))
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
> print(crypto.toHex(crypto.fhash("SHA384", 'empty.lua')))
38b060a751ac96384cd9327eb1b1e36a21fdb71114be07434c0cc7bf63f6e1da274edebfe76f65fbd51ad2f14898b95b
> print(crypto.toHex(crypto.fhash("SHA512", 'empty.lua')))
[...]

PR #1647, opened 2016-12-06
Various minor bits of cleanup

### LINKED ISSUE #1644: Various minor issues with gpio/perf/websocket
### Missing feature

Fix the following minor issues:

* websocket client keeps some data structures in RAM which should be in flash
* perf doesn't actually use the right PC when making its histogram
* gpio interrupt callback should indicate when the interrupt happened.

### Justification

This is cleanup.

### Workarounds

Yes, I've already fixes these in my branch.

pjsg:
Fixes #1644 and #1633 

The fixes are:

* Check the return code of the read function when doing crypto.fhash so that we don't pass negative lengths to the hashing functions
* Fix various assert failures in the LVM arising from rotables. No functional change
* Add the gpio interrupt time to the callback (and pass it from the interrupt handler)
* Get the PC right in the perf module
* Make the headers static in the websocket module
* Fix the documentation

### DIFF
--- a/app/modules/perf.c
+++ b/app/modules/perf.c
@@ -24,7 +24,6 @@ typedef struct {
   uint32_t bucket_count;
   uint32_t total_samples;
   uint32_t outside_samples;
-  uint32_t pc_offset;
   uint32_t bucket[1];
 } DATA;
 
@@ -39,7 +38,11 @@ static void ICACHE_RAM_ATTR hw_timer_cb(os_param_t p)
   uint32_t stackaddr;
 
   if (data) {
-    uint32_t pc = *(&stackaddr + data->pc_offset);
+    uint32_t pc;
+    asm (
+      "rsr   %0, EPC1;"    /* read out the EPC */
+      :"=r"(pc)
+    );
 
     uint32_t bucket_number = (pc - data->start) >> data->bucket_shift;
     if (bucket_number < data->bucket_count) {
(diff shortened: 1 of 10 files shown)


marcelstoer:
It was confirmed that this fixes #1633. So I guess all it'd need was for someone knowledgeable to give the code a quicke glance.

pjsg:
Bump -- can someone take a look please?

app/platform/platform.c
```diff
@@ -184,6 +184,7 @@ int platform_gpio_read( unsigned pin )
 static void ICACHE_RAM_ATTR platform_gpio_intr_dispatcher (void *dummy){
   uint32 j=0;
   uint32 gpio_status = GPIO_REG_READ(GPIO_STATUS_ADDRESS);
+  uint32 now = system_get_time();
```
jmattsson: Does `system_get_time()` live in IRAM?

app/modules/perf.c
```diff
@@ -39,7 +38,11 @@ static void ICACHE_RAM_ATTR hw_timer_cb(os_param_t p)
   uint32_t stackaddr;
 
   if (data) {
-    uint32_t pc = *(&stackaddr + data->pc_offset);
+    uint32_t pc;
+    asm (
+      "rsr   %0, EPC1;"    /* read out the EPC */
```
jmattsson: Are you sure you want the EPC1 here? I thought that register was only used during hardware exception handling.

app/platform/platform.c
pjsg: Yes. In particular, in the last build that I did, I got the following addresses:

```
                0x0000000040101a10                system_deep_sleep_local_2
                0x0000000040101ac4                system_get_time
                0x0000000040101adc                system_os_post
                0x0000000040101b10                system_rtc_mem_write
```

app/modules/perf.c
pjsg: The pragmatic answer is that it works. The fuller answer is that the timer interrupt option has three compare registers which can each generate interrupts (each of a particular level). It appears that the SDK initializes the timer that the hw_timer uses to use interrupt 1 and hence EPC1 is the right register to use.

jmattsson: Ah yes, of course, interrupts are delivered via hw exceptions too. Brain is in holiday mode, clearly >.>

Issue #1636, opened 2016-12-03
bug in bme280.c


eyaleb:
When converting the module to use i2c bust reads I noticed a mismatch from the doco in `bme280_lua_init()`.
I wonder if the doco is known to be wrong? After all a bad extraction of the parameters should produce wrong result from `bme280_compensate_H()` and I did not see a complaint.
### Expected behavior
The data-sheet says (**6.2.2 Trimming parameter readout**)
```
0xE4 / 0xE5[3:0]  -> dig_H4 [11:4] / [3:0]  signed short
0xE5[7:4] / 0xE6  -> dig_H5 [3:0] / [11:4]  signed short
0xE7              -> dig_H6                 signed char
```
### Actual behavior
The code for extracting the humidity parameters  does
```
bme280_data.dig_H4 = ((int16_t)r8u(reg) << 4 | (r8u(reg+1) & 0xF)); reg+=2;
bme280_data.dig_H5 = ((int16_t)r8u(reg+1) << 4 | (r8u(reg) >> 4)); reg+=2; 
bme280_data.dig_H6 = (int8_t)r8u(reg);
```
Note the `reg+=2` after extracting `H4`. `0xE5` should be split between `H4` and `H5`. The current code seems to extract `H5` from `0xE7/0xE6` then `H6` from `0xE8` (out of range). A `++reg` should be done instead.
Do I read this correctly? I am looking at the doco from Bosch dated
```
BME280: Final data sheet
Document revision            1.2
Document release date        October 26th, 2015
Document number              BST-BME280-DS001-11
Technical reference code(s)  0 273 141 185
```
### Test code
None, this is what I find looking at the code.
### NodeMCU version
Latest dev branch
### Hardware
None, this is what I find looking at the code.
### Notes
[Here](http://members.iinet.net.au/~eyaleb/bme280.c) is a version of `bme280` where I changed:
1 - Fixed the above bug
2 - improved registers read by using burst more
3 - Added `bme280_lua_read()` to efficiently read `T, P & H` (and optionally return QNH) in one call.

vsky279:
Thanks for spotting this bug. I have posted a PR https://github.com/nodemcu/nodemcu-firmware/pull/1652 to fix it based on your suggestion.

Please give me more time to study your other suggestions.

Issue #1632, opened 2016-12-02
bme280 module is inefficient


eyaleb:
### Expected behaviour
An efficient chip access using I2C burst mode.
One call to get all the readings.
### Actual behaviour
The many registers are read a byte at a time, some more that once.
Fetching Temp, Baro and Humi requires a number of API calls.
### Test code
See the actual code in `app/modules/bme280.c` and the example in the corresponding doco.
### NodeMCU version
latest dev branch.
### Hardware
Any ESP8266 device.
### Comment
I did a quick test creating a `T, P, H = bme280.read()` function. I also fetch the registers in a burst like this
```c
static int r8u_n(uint8_t reg, int n, uint8_t *buf) {
        int i;
        platform_i2c_send_start(bme280_i2c_id);
        platform_i2c_send_address(bme280_i2c_id, bme280_i2c_addr, PLATFORM_I2C_DIRECTION_TRANSMITTER);
        platform_i2c_send_byte(bme280_i2c_id, reg);
        platform_i2c_send_start(bme280_i2c_id);
        platform_i2c_send_address(bme280_i2c_id, bme280_i2c_addr, PLATFORM_I2C_DIRECTION_RECEIVER);
        while (n-- > 0)
                *buf++ = platform_i2c_recv_byte(bme280_i2c_id, n > 0);
        platform_i2c_send_stop(bme280_i2c_id);
}
// Always return T, P, H or none
static int bme280_lua_read(lua_State* L) {
        uint8_t buf[8];

        r8u_n(BME280_REGISTER_PRESS, 8, buf);   // registers are P[3], T[3], H[2]

// Must do Temp first since bme280_t_fine is used by the other compensation functions
        uint32_t adc_T = (uint32_t)(((buf[3] << 16) | (buf[4] << 8) | buf[5]) >> 4);
        if (adc_T == 0x80000 || adc_T == 0xfffff)
                return 0;
        lua_pushinteger(L, bme280_compensate_T(adc_T));

        uint32_t adc_P = (uint32_t)(((buf[0] << 16) | (buf[1] << 8) | buf[2]) >> 4);
        if (adc_P ==0x80000 || adc_P == 0xfffff)
                lua_pushinteger(L, 0);
        else
                lua_pushinteger(L, bme280_compensate_P(adc_P));

        uint32_t adc_H = (uint32_t)((buf[6] << 8) | buf[7]);
[...]

PR #1652, opened 2016-12-06
BME280: fixing humidity trimming parameter readout bug

vsky279:
Fixes https://github.com/nodemcu/nodemcu-firmware/issues/1636.

Fixes a bug in reading of trimming parameter `dig_H5` and `dig_H6` (Datasheet: 6.2.2 Trimming parameter readout). These parameters are used in compensation formulas for calculating humidity - i.e. humidity readout before this bug fix was not correct.

### DIFF
--- a/app/modules/bme280.c
+++ b/app/modules/bme280.c
@@ -286,7 +286,7 @@ static int bme280_lua_init(lua_State* L) {
 		reg = BME280_REGISTER_DIG_H2;
 		bme280_data.dig_H2 = r16sLE(reg); reg+=2;
 		bme280_data.dig_H3 = r8u(reg); reg++;
-		bme280_data.dig_H4 = ((int16_t)r8u(reg) << 4 | (r8u(reg+1) & 0xF)); reg+=2;
+		bme280_data.dig_H4 = ((int16_t)r8u(reg) << 4 | (r8u(reg+1) & 0xF)); reg++;
 		bme280_data.dig_H5 = ((int16_t)r8u(reg+1) << 4 | (r8u(reg) >> 4)); reg+=2;
 		bme280_data.dig_H6 = (int8_t)r8u(reg);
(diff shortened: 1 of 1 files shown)


marcelstoer:
@eyaleb can we ask you to verify this? [Your fix](http://members.iinet.net.au/~eyaleb/bme280.c) for the particular line is slightly different.

eyaleb:
The fix does
```c
bme280_data.dig_H4 = ((int16_t)r8u(reg) << 4 | (r8u(reg+1) & 0xF)); reg++;
```
My version does
```c
bme280_data.dig_H4 = (int16_t)reg[0] << 4 | (reg[1] & 0x0F); reg+=1;
```

The two are identical.

My version reads the data ahead (in a burst), so `r8u(reg)` is my `reg[0]` and `r8u(reg+1)` is my `reg[1]`.

The original code uses `reg` as the i2c register to read and my code uses `reg` as a pointer into the (already read) data in a buffer.

pjsg:
Ugh. This approach to reading the registers in individual i2c requests has problems. In particular, the [datasheet](https://cdn-shop.adafruit.com/datasheets/BST-BME280_DS001-10.pdf) says (at the start of section 4)

> To read out data after a conversion, it is strongly recommended to use a burst read and not
> address every register individually. This will prevent a possible mix-up of bytes belonging to
> different measurements and reduce interface traffic. Note that in I²C mode, even when pressure
> was not measured, reading the unused registers is faster than reading temperature and
> humidity data separately. 

Presumably the current code will sometimes give the wrong answers....

eyaleb:
@pjsg This is correct when reading the P/H/T. This fix is for the reading of the trimming parameters which I understand are static.

pjsg:
True -- but reading out the pressure, temp etc all does single byte reads. I.e. `r24u` does three `r8u` calls, each of which does a single byte transaction.

eyaleb:
Yes, reading the pressure, temp etc needs fixing, but this specific fix is for the incorrect extraction of a parameter. What @pjsg refers to is addressed in #1632 (node the introduction of `r8u_n()`) which should probably be marked as a **bug** .

PR #1658, opened 2016-12-10
Adds uart.check(0) to get the current uart parameters

### LINKED ISSUE #1655: Get auto-detected baud rate from uart module
The uart module auto-detection of baud rate works great, but how you change the parity, stopbits and databits while still using the auto-detection of baud rate feature?

Baud rate auto-detection only works if you don't call uart.setup(), but you have to call uart.setup() to change parity, stopbits or databits.

If you call uart.setup() you have to supply a baud rate, but you don't know what baud rate to supply!  If you supply a baud rate and make the call to uart.setup() it overwrites the auto-detected baud rate with the baud rate supplied.

My ideal solution would be to call uart.setup with only the id parameter, and it would then return the auto-detected baud rate, or return nil if baud rate not yet detected  (baud rate could then be used in subsequent calls to uart.setup in which parity, stopbits and databits could be set).

Baud rate auto-detection is an excellent feature, but its value is greatly diminished if we can't find out what baud rate is detected.

pjsg:
Fixes #1655.

This adds the `uart.check(0)` to return the current settings of the UART0. It actually reads from the registers (in the driver) to do this, so it will always return the right data. The only caveat is that the setting interface only allows setting one of the specified baud rates. Any other value is set to 9600. In order to make the developer experience better, the `uart.check` code adjusts the actual baud rate to the closest permitted value. This is needed as the autobaud code can pick some pretty funny rates that are really close to the nominal rate.

### DIFF
--- a/app/driver/uart.c
+++ b/app/driver/uart.c
@@ -356,3 +356,7 @@ uart_setup(uint8 uart_no)
 void ICACHE_FLASH_ATTR uart_set_alt_output_uart0(void (*fn)(char)) {
   alt_uart0_tx = fn;
 }
+
+int ICACHE_FLASH_ATTR uart_get_baudrate(uint8 uart_no) {
+  return UART_CLK_FREQ / (*(uint32_t *) (0x60000014 + 0xf00 * uart_no));
+}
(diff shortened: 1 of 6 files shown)


jmattsson:
Why `check()` and not simply `get()` or `getconfig()`? Check to me implies testing for some sort of error.

pjsg:
You are right -- I renamed to `getconfig` -- it makes more sense. I didn't want to use `get` as it sounded like a method to get a character from the uart.

Issue #1625, opened 2016-11-27
High power consumption during rtctime.dsleep


SergSlipushenko:
Hi, folks! I noticed that esp8266 consumes a lot (approx. 2mA) after rtctime.dsleep call. If I call node.dsleep than everything is ok. Is it known issue? I didn't manage to find answer in already existed issues here.

### Expected behavior
Power consumption < 0.1mA during deep sleep mode

### Actual behavior
Power consumption approximately  2.2mA during deep sleep mode.  It looks strange but some times it consumes 6.6mA or even works as expected. It looks like a bug and it is the main reason of reporting this issue :)

### Test code
```Lua
rtctime.dsleep(10000000,0)
```
and measure power consumption of esp8266 chip

### NodeMCU version
Current master (7b83bbb2ea134cd85ac9d63108603cc02c4e20f7)

### Hardware
ESP12-E chip. Please, note that it is not NodeMCU dev kit, so this issue can not be related with consumption of other components

jmattsson:
How reliably can you reproduce this? And are you able to probe the flash chip to see what state it gets left in?

SergSlipushenko:
To reproduce it just run rtctime.dsleep(10000000,0) and measure power consumption. In 9 of 10 cases it will be ~2mA. Rear it can be less then 0.1mA or surprisingly 6.6mA. But, again, usually it is 2mA... Also sometimes "something" happens and it is possible to get strike of 3-5 normal sleeps but then bug come back. Unfortunately I can't measure flash consumption because it is esp12-e solid module.

Basically, I already have found workaround for this bug, but with loosing rtc functionality :( . So if it is, kind of hardware limitation, it should displayed in docs, I guess

jmattsson:
I don't currently have the gear available to measure it myself at the moment. I'll see if I can get some spare lab time at $work.

However, this *used* to work. Once upon a time the `system_deep_sleep()` function also exhibited a similar bad behaviour, but Espressif fixed that after we reported it. It's peculiar that the rtctime.dsleep() would have developed such an issue since we explicitly guarded against this failure mode, but I suppose something must have changed in the SDK to do, well, *something*.

On an ESP12-E module you'll find the flash chip lines are also exposed as the pins on the "bottom" row of the module. I'd be interested in seeing whether any of the lines are left in a different state when this bug occurs vs when it doesn't. There might be a clue there. Then again, there might not.

[...]

PR #1659, opened 2016-12-10
rtctime.dsleep() current draw improvements

jmattsson:
Fixes some aspects of #1625.

- [ ] I have thoroughly tested my contribution. **I haven't, but others have; see referenced issue for discussion**
- [ ] The code changes are reflected in the documentation at `docs/en/*`. **n/a**

### DIFF
--- a/app/include/rtc/rtcaccess.h
+++ b/app/include/rtc/rtcaccess.h
@@ -54,4 +54,17 @@ static inline uint32_t rtc_reg_read(uint32_t addr)
   rtc_memw();
   return *((volatile uint32_t*)addr);
 }
+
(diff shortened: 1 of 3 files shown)


eyaleb:
Applied the three changed files to latest `dev` and tested. `node.dsleep()` is still low (23uA) and `rtctime.dsleep()` is a bit higher (39uA) but consistent. BTW Vdd is a bit higher (3.4v) in this test so the current is correspondingly higher than earlier tests in #1625 (vdd=3.3v and current=19uA/34uA).

jmattsson:
Great feedback, thanks!

Are there any objections to merging this in? While it's not a complete fix, it improves on the status quo a lot.

Issue #1631, opened 2016-12-02
bme280.init() is ambiguous


eyaleb:
### Missing feature
`bme280.init()` initialises the driver and the chip. We need to separate the two actions.

### Justification
In a battery operated setup the esp deep sleeps and on wakeup needs to initialise the driver (the `bme280.c` module). The chip was kept powered (sleeping too) and is holding the latest reading that should be fetched quickly before another reading starts in the background (`bme280.startreadout()`).

A chip init is not required and actually is detrimental in this context (which I consider common use).

### Workarounds
As always I can fiddle with the module and add an option to the `init()` routine to indicate if the chip needs to be initialised too.

vsky279:
When coding the module I have not thought about optimizing it for battery usage. But I agree it makes sense to do it.
I would suggest to add another optional parameter `cold_start` to `bme280.init()` which would be `true` by default. When `false` it would skip the unnecessary parts of the code.

PR #1662, opened 2016-12-10
BME280: improved reliability and efficiency

vsky279:
Fixes https://github.com/nodemcu/nodemcu-firmware/issues/1631 and https://github.com/nodemcu/nodemcu-firmware/issues/1632.

This PR improves reliability of the measurement by implementing the I2C burst mode reading as suggested by the [BME280 datasheet](https://cdn-shop.adafruit.com/datasheets/BST-BME280_DS001-10.pdf) discussed here https://github.com/nodemcu/nodemcu-firmware/pull/1652 and here https://github.com/nodemcu/nodemcu-firmware/issues/1632.
>To read out data after a conversion, it is strongly recommended to use a burst read and not
address every register individually. This will prevent a possible mix-up of bytes belonging to
different measurements and reduce interface traffic. Note that in I²C mode, even when pressure
was not measured, reading the unused registers is faster than reading temperature and
humidity data separately.

It speeds up initialization and readouts which should improve the endurance when run on batteries.

It also implements the `cold_start` parameter to `init()` method as suggested in https://github.com/nodemcu/nodemcu-firmware/issues/1631.

New method `read()` is implemented (and documented) which allows to read all measured parameters at once.

It is based on the modifications proposed of @eyaleb.

### DIFF
--- a/app/modules/bme280.c
+++ b/app/modules/bme280.c
@@ -23,14 +23,14 @@
 #define BME280_REGISTER_VERSION               (0xD1)
 #define BME280_REGISTER_SOFTRESET             (0xE0)
 #define BME280_REGISTER_CAL26                 (0xE1)
-#define BME280_REGISTER_TEMP                  (0xFA)
-#define BME280_REGISTER_PRESS                 (0xF7)
-#define BME280_REGISTER_HUM                   (0xFD)
+#define BME280_REGISTER_PRESS                 (0xF7)	// 0xF7-0xF9
+#define BME280_REGISTER_TEMP                  (0xFA)	// 0xFA-0xFC
+#define BME280_REGISTER_HUM                   (0xFD)	// 0xFD-0xFE
 
-#define BME280_REGISTER_DIG_T                 (0x88)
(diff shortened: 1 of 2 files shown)


vsky279:
@eyaleb could you please test the implementation?

eyaleb:
I did a quick check and it looks OK. I can see that you made more restructuring and left some clean-up of dead code to be done.

app/modules/bme280.c
```diff
@@ -144,24 +148,29 @@ static uint8_t w8u(uint8_t reg, uint8_t val) {
...
+	r8u_n(reg, 1, ret);
+	return ret[0];
 }
 
-static uint16_t r16uLE(uint8_t reg) {
-	uint8_t low = r8u(reg);
-	uint8_t high  = r8u(++reg);
-	return (high << 8) | low;
-}
+// static uint16_t r16u(uint8_t reg) {
```
devsaurus: Please either put these functions into a meaningful `#ifdef XYZ` / `#endif` or remove them completely if not used anymore.

vsky279: done

PR #1665, opened 2016-12-11
Added node.random implementation

### LINKED ISSUE #1374: Expose random number at 0x3FF20E44
### Missing feature

Random number generator. According to http://esp8266-re.foogod.com/wiki/Random_Number_Generator the easiest seems to be to offer a read method for 0x3FF20E44. Don't the crypto functions have something like that internally?
### Justification

Some guy over at esp8266.com built his own firmware with the disabled Lua math module enabled and then complained about bad results. Ok, not a good enough justitication but stil...
### Workarounds

Hand-woven randomness or use https://www.random.org/clients/http/.

pjsg:
Fixes #1374.

This adds node.random which behaves the same way that math.random does *except* that it uses the hardware random number generator for (more) unpredictably random numbers. It also gets large ranges right in the integer build.

### DIFF
--- a/docs/en/modules/node.md
+++ b/docs/en/modules/node.md
@@ -364,6 +364,33 @@ Nothing
 node.osprint(true)
 ```
 
+## node.random()
+
+This behaves like math.random except that it uses true random numbers derived from the ESP8266 hardware. It returns uniformly distributed
+numbers in the required range. It also takes care to get large ranges correct. 
+
+It can be called in three ways. Without arguments in the floating point build of NodeMCU, it returns a random real number with uniform distribution in the interval [0,1). 
+When called with only one argument, an integer n, it returns an integer random number x such that 1 <= x <= n. For instance, you can simulate the result of a die with random(6). 
+Finally, random can be called with two integer arguments, l and u, to get a pseudo-random integer x such that l <= x <= u.
+
+#### Syntax
+`node.random()`
+`node.random(n)`
+`node.random(l, u)`
+
+#### Parameters
+- `n` the number of distinct integer values that can be returned -- in the (inclusive) range 1 .. `n`
+- `l` the lower bound of the range
+- `u` the upper bound of the range
+
+#### Returns
+The random number in the appropriate range. Note that the zero argument form will always return 0 in the integer build.
+
+#### Example
+```lua
+print ("I rolled a", node.random(6))
+```
+
 # node.egc module
 
 ## node.egc.setmode()
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -467,6 +467,79 @@ static int node_osprint( lua_State* L )
   return 0;  
 }
 
+int node_random_range(int l, int u) {
+  // The range is the number of different values to return
+  unsigned int range = u + 1 - l;
+
+  // If this is very large then use simpler code
+  if (range >= 0x7fffffff) {
+    unsigned int v;
+
+    // This cannot loop more than half the time
+    while ((v = os_random()) >= range) {
+    }
+
+    // Now v is in the range [0, range)
+    return v + l;
+  }
+
+  // Easy case, with only one value, we know the result
+  if (range == 1) {
+    return l;
+  }
+
+  // Another easy case -- uniform 32-bit
+  if (range == 0) {
+    return os_random();
+  }
+
+  // Now we have to figure out what a large multiple of range is
+  // that just fits into 32 bits.
+  // The limit will be less than 1 << 32 by some amount (not much)
+  uint32_t limit = ((0x80000000 / ((range + 1) >> 1)) - 1) * range;
+
+  uint32_t v;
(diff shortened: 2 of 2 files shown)


docs/en/modules/node.md
```diff
@@ -364,6 +364,33 @@ Nothing
 node.osprint(true)
 ```
 
+## node.random()
+
+This behaves like math.random except that it uses true random numbers derived from the ESP8266 hardware. It returns uniformly distributed
+numbers in the required range. It also takes care to get large ranges correct. 
+
+It can be called in three ways. Without arguments in the floating point build of NodeMCU, it returns a random real number with uniform distribution in the interval [0,1). 
+When called with only one argument, an integer n, it returns an integer random number x such that 1 <= x <= n. For instance, you can simulate the result of a die with random(6).
```
devsaurus: dice?

luismfonseca:
I may be missing something obvious here, but what's wrong with:

```
int node_random_range(int l, int u) {
  unsigned int range = u - l + 1;

  switch(range) {
    case 0: return os_random();
    case 1: return l;
    default: return l + os_random() % range;
  }
}
```

Assuming that `os_random` will generate a uniform distribution, doesn't that mean that slicing the result also results in a uniform distribution?

docs/en/modules/node.md
pjsg: die is (in British English) the singular of dice (at least that is what I was always taught).

pjsg:
No. Consider the simpler case of a random generator returning 0,1,2 or 3 uniformly.

Now you want to have a random generator that returns 0,1 or 2 uniformly. If you just take the original generator and %3 it, then 0 ends up happening twice as often as 1 or 2. This may (or may not) be bad, but it isn't much more complicated to get it right. 

I'm a crypto person, so I like my random numbers unbiased!

docs/en/modules/node.md
jmattsson: Uttering "die in a fire" in a D&D game is not a hostile action, it's a call to action - get that d6 out of the camp fire! :)

Thanks for making the float range exclusive on the end. Yay sanity!

luismfonseca:
Ah I see, you're right that makes sense.

Although, I am still thinking about this... If the bits in the random number are also uniformly distributed, doesn't that mean that in the best case scenario we could run `os_random` once?

I.e. apply a mask for select the bits that are within the range... I suppose that's tricky because let's suppose we have a range of `5`, that's `101b`, so we would need a mask of `111`, but that can yield results up to `7`, and we can't really do the `%` because what you just explained, so... Hmm... Interesting.

pjsg:
It might make an interesting coding interview question. I would ask it incrementally. E.g. "Given a source of 32 bit random numbers, write a function to return a random number in the range l <= r <= u". Then go on to asking about uniformity etc.....

docs/en/modules/node.md
devsaurus: Cool, no further questions 😊

jmattsson:
Question: rather than re-rolling the random numbers, would it not be an option to right-shift the first one rolled until it fits below the limit? It would put a definite upper limit on the number of times through the loop. In theory the current loops could trigger the watchdog in the unlikely event of a long string of high-bit set numbers.

pjsg:
If you want unbiased numbers, you have to be very careful. The worst case is when you want the range of 0x80000001 numbers, and this means that half the time, the value returned is unacceptable. The cost of os_random is small (1us), and there is a 50% chance that you have to go around the loop again. So the probability that you will loop for a millisecond is 2^-1000. This is vanishingly small. If the random number generator stops working (and always returns the same number), then it could easily timeout. However, I suspect that you have got bigger problems at that stage.

jmattsson:
Ah yes, a range in the upper half there's not much to be done for; I was mostly thinking of the lower range loop, which I'd consider the more frequent path.

pjsg:
That is not so much a problem as the disallowed range is smaller so the probability of failure is correspondingly less.

computerlyrik:
@pjsg seems to be a great work towards a solution.

I am looking forward to get this into the dev branch for testing. I assume this is much faster than `math.random()`?
I do have some issues in a simulation using multiple `math.random()` in a `tmr` object. It kills itself due to long runtime.

pjsg:
How fast do you need it to be? It turns out that there are lots of ways to improve performance. In particular avoiding lookups in the global tables. I.e. add a Local r = node.randomOutside the loop and then use r() inside the loop. Also running the integer build is faster that the float build. Doubling the clock speed doesn't hurt either!On Dec 13, 2016 15:23, computerlyrik <notifications@github.com> wrote:@pjsg seems to be a great work towards a solution.
I am looking forward to get this into the dev branch for testing. I assume this is much faster than math.random()?
I do have some issues in a simulation using multiple math.random() in a tmr object. It kills itself due to long runtime.

PR #1674, opened 2016-12-18
Cron module

### LINKED ISSUE #1671: FR: cron module
As proposal to resolve #1670 

### Missing feature
Cron-like module (with possibility to run repeatable tasks).

### Justification
This is very important feature for IoT devices that runs some scheduled tasks (e.g. light controllers with daytime-related switching or dimming).

### Workarounds
Manual handle (e.g. minutely) wallclock-time and compare parts with template. Very complex at all.

My propose is to include parser (based on `ets_timer` and `gmtime`) in module and parse schedule list in C-level.

### API proposal
```lua
-- Every day at 09:00
local task_on = cron.schedule('0 9 * * *', function() light.switch(true) end)
-- Every day at 22:00
local task_off = cron.schedule('0 22 * * *', function() light.switch(false) end)
-- Remove task
task_on:unschedule()
-- Re-schedules task to 21:00
task_off:schedule('0 21 * * *')
-- Replaces task handler
task_off:handler(function() light.dim(20) end)
```

Discussion is welcome.

djphoenix:
Fixes #1671.


app/modules/cron.c
```diff
@@ -0,0 +1,223 @@
...
+  ud->cb_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+  return 0;
+}
+
+static int lcron_unschedule(lua_State *L) {
+  cronent_ud_t *ud = luaL_checkudata(L, 1, "cron.entry");
+  size_t i = lcron_findindex(L, ud);
+  if (i == -1) return 0;
+  luaL_unref(L, LUA_REGISTRYINDEX, cronent_list[i]);
+  memmove(cronent_list + i, cronent_list + i - 1, sizeof(int) * cronent_count - i - 1);
```
pjsg: Shouldn't the second arg be cronent_list + i + 1?

app/modules/cron.c
```diff
@@ -0,0 +1,223 @@
...
+  uint32_t diff = 60000 - tm.tm_sec * 1000 - tv.tv_usec / 1000;
+  if (tm.tm_sec == 59) {
+    t++;
+    diff += 60000;
+    gmtime_r(&t, &tm);
+  }
+  if (tm.tm_sec <= 1) {
+    cron_handle_time(tm.tm_mon + 1, tm.tm_mday, tm.tm_wday, tm.tm_hour, tm.tm_min);
+  }
+  ets_timer_arm_new(&cron_timer, diff, 0, 1);
```
pjsg: I think that you want this *before* you trigger the callbacks (which might take some time to execute)

pjsg:
Can you add a module level method that returns all of the cron.entry objects? I might want to clear all outstanding cron entries....

app/modules/cron.c
```diff
@@ -0,0 +1,223 @@
...
+  time_t t = tv.tv_sec;
+  struct tm tm;
+  gmtime_r(&t, &tm);
+  uint32_t diff = 60000 - tm.tm_sec * 1000 - tv.tv_usec / 1000;
+  if (tm.tm_sec == 59) {
+    t++;
+    diff += 60000;
+    gmtime_r(&t, &tm);
+  }
+  if (tm.tm_sec <= 1) {
```
pjsg: I'm not convinced that you want this check here. If the timer is delayed by 10 seconds (due to other activity), you still want to run the scheduled tasks.

app/modules/cron.c
djphoenix: Yep, misprint here.

app/modules/cron.c
djphoenix: Right point, thanks.

djphoenix:
@pjsg thanks for review, I fixed all of your points. About list of entries - don't think this may be usable for any reason. Maybe something like `cron.reset()` or `cron.clear()`?

pjsg:
Yes -- I think that having a `reset` method would solve the problem -- thanks.

djphoenix:
@pjsg Right, I added `reset` method and pushed changes again.

docs/en/modules/cron.md
```diff
@@ -110,9 +104,9 @@ ent:schedule("*/5 * * * *")
 
 ## cron.entry:unschedule()
 
-Disable schedule.
+Disables schedule.
 
-Disabled schedule may be enabled again by calling [`:schedule(mask)`](#cronentryschedule).
+Disabled schedules may be enabled again by calling [`:schedule(mask)`](cron.md#cronentryschedule).
```
djphoenix: @marcelstoer does local links requires filename specification?

marcelstoer: Not for viewing files on GitHub but for MkDocs (which converts to HTML on readthedocs.io) if I'm not mistaken.

djphoenix: For `mkdocs serve` it works. For readthedocs - not sure.

marcelstoer:
Maybe we should add a reference to this module to the [timer docs](http://nodemcu.readthedocs.io/en/latest/en/modules/tmr/)? As a way to circumvent the timer interval limit of 1:54:30.947?

djphoenix:
Yep, this is a good point that we should have IMO.

danjohn:
Please document, how many entries the crontab can carry (stored in RAM?), and if there is some kind of GC for stale (non-repeated) entrys.

For one-time events, it would also be beneficial for schedule() to accept a timestap instead of the description string (like cron.schedule(rtctime.get()+3*3600, doIn3hours()), and maybe GC that afterwards).

What do you think about this (non-unix-like-cron but simple&helpful) syntax extension?

marcelstoer:
For one-time events there's `tmr`, `cron` shouldn't be (ab)used for that.

danjohn:
I partially agree, but then tmr should be able to handle longer intervals (1 month or the like).

My application has to deal with sun rise/dip times, which are hopefully recalculated/rescheduled every midnight. A max. tmr interval of 1:54:31 doesn't help here, and I'd like to have a cron fallback (if reschedule fails on an unresponsive time server).

You're right in not mixing up cron with tmr, but it is a hassle to syncronize that time bases.
Suggestions:
*) Extend possible tmr intervals to 256*(1:54:30.947) OR (if that blows firmware possibilities)
*) new function tmr.alarmSec(seconds, callback()) OR
*) provide a tmr/cron "at" service (with cron-like syntax or timestamps) based on rtctime

I think, that problem is not very exotic and nettles many users.

pjsg:
With long time delays, you need to be careful about what you are waiting for. For example, if I set a timeout for 24 hours from now (as measured by the nodemcu), is that really what I want, or do I want to get a callback when the wall clock is 24 hours greater than now?

marcelstoer:
The `tmr` limit is imposed by the Espressif SDK IIRC, nothing we can do about.

danjohn:
@Phil: I don't see your precautions... is it about the unaccuracy of the nodemcu time base, or dst issues, or I missed the point (sorry)?

@Marcel:
I knew that this is an Essif limitation. But there is well something we can do about it, and render that strange limit obsolete:
6870947 is (oddly) 0x68d7a3; so I suppose, this is kept in a 32bit var (uInt32 or the like). If the full range could be used, it would give an interval of ~50days, more than enough!
Something like
 
>local delta=0
>function tmr.alarm(id, time, mode, f) {
>    if time<6870947 then existent:tmr.alarm(id, time, mode, f)
>    else {
>        delta=time
>        tmr.alarm(id, time, mode, function(di) {
>            delta=delta-6870947
>            tmr.alarm(x, delta, mode, self)
>        }
>}

translated to the C layer would do the trick, not wasting ressources in the lua level (extra timer, var, func) and *not changing the API* (except modifying the docs ;-)).

Is it worth filing a feature request on tmr for that?

PR #1682, opened 2016-12-24
Fix crash in sntp and add more reliable defaults servers

### LINKED ISSUE #1679: sntp sometimes crashes when there is lots of activity
### Expected behavior
The firmware doesn't crash

### Actual behavior
It sometimes does.

The problem is that the lua callback is called from the net callback which is not at lua base level. The solution is to use the task handling stuff to transfer control to the base level where the callback can be performed.

pjsg:
Fixes #1679.

The sntp code juts made the assumption that errors could be reported at any time. This was a bad assumption. The changes now make use of the task interface to schedule the error callback.

Other changes:

* Now defaults to using the nodemcu specific servers in the ntp pool (0 - 3 .nodemcu.pool.ntp.org). This makes the simplest configuration work in all (internet connected) cases.
* Fixes a problem in leap second calculation (it got the day offset from 1-mar-0000 wrong)
* Adds support for the KoD packet. This is required to use ntp pool servers.
* Speeds up the uint64 division by 1,000,000 dramatically.
* Gives priority to the previous ntp server chosen - this reduces server switching.

On the subject of division -- this is *really* slow -- especially the 64 bit division (this is what I discovered from the fixed version of the perf code -- this is in the random cleanup PR). Unfortunately there isn't a general mechanism for the compiler to use to convert uint64 division into something fast. (On many platforms, 32 bit division is converted into 64 bit multiplication with a shift afterwards). These constants were found by searching (~12 hours of x86 CPU time).

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -41,6 +41,7 @@
 #include "c_stdlib.h"
 #include "user_modules.h"
 #include "lwip/dns.h"
+#include "task/task.h"
 #include "user_interface.h"
 
 #ifdef LUA_USE_MODULES_RTCTIME
@@ -60,9 +61,10 @@
 # define sntp_dbg(...)
 #endif
 
-#define US_TO_FRAC(us)          ((((uint64_t) (us)) << 32) / 1000000)
+//#define US_TO_FRAC(us)          ((((uint64_t) (us)) << 32) / 1000000)
+#define US_TO_FRAC(us)          (div1m(((uint64_t) (us)) << 32))
 #define SUS_TO_FRAC(us)         ((((int64_t) (us)) << 32) / 1000000)
-#define US_TO_FRAC16(us)        ((((uint64_t) (us)) << 16) / 1000000)
+//#define US_TO_FRAC16(us)        ((((uint64_t) (us)) << 16) / 1000000)
 #define FRAC16_TO_US(frac)      ((((uint64_t) (frac)) * 1000000) >> 16)
 
 typedef enum {
@@ -70,7 +72,8 @@ typedef enum {
   NTP_DNS_ERR,
   NTP_MEM_ERR,
   NTP_SEND_ERR,
-  NTP_TIMEOUT_ERR
+  NTP_TIMEOUT_ERR,
+  NTP_MAX_ERR_ID     // must be last
 } ntp_err_t;
 
 typedef struct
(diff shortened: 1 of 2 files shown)


jmattsson:
Awww, you've got me all nostalgic with the div1m approach :)

:+1: from me, with the caveat I haven't actually tested it.

marcelstoer:
Johny, care to elaborate for the uninitiated? What _is_ "div1m"? I have the feeling I could learn something interesting here 😄

jmattsson:
Let me find a good text on the topic... alright, [this one](http://embeddedgurus.com/stack-overflow/2009/06/division-of-integers-by-constants/) looks pretty good. It doesn't cover the full story of what @pjsg has done in the [`div1m`](https://github.com/nodemcu/nodemcu-firmware/blob/057ab92d78956b181c4c80f4c487f95b60ab233b/app/modules/sntp.c#L162) function here, as the article is largely reciprocal multiplication, whereas @pjsg has done away with the mul/div altogether. The paper linked in the article does seem to cover it, but is a bit denser to read.

Basically, since shift operators are much much faster than division instructions, when you're dividing by a known constant it's typically possible to speed up the operation by implementing the division as a sequence of shift operations. You'd see this a fair bit back in the days when (game) graphics processing was all done on the main CPU, and people were squeezing every cycle in their engines. The most frequently seen case is of course when using powers of two, as that directly translates into a single shift.

For digging up the factors/steps for any arbitrary divisor X, I'm not actually aware of any quick or even optimal approaches - perhaps @pjsg could chip in with how he crunched them? I'd be curious too :)

pjsg:
The long and short is that you can evaluate   `a / b`   by doing   `a * (1/b)`  which is good if b is a constant. Except of course, you want to do this without floating point. So you do

```
(a * (2^n/b))/2^n
```

The divide by `2^n` is a shift, and the `2^n/b` is a constant.

Then observe that multiplication of 64-bit values on esp is expensive. But multiplication by a constant is equivalent to a sequence of adds and shifts. E.g.

```
a * 5 ==  (a <<2) + a

a * 7 == (a << 3) - a
```
So, in order to divide by an arbitrary constant, you want to construct a sequence of adds and subtracts that has the appropriate bit pattern, but is always *less* than the number that you want. 

If you are trying to compute `a / b`, then compute `c < 2^n / b, c as large as possible`.

```
q = (a * c) >> n;    // But q is a little small as c is a little small
error = a - q * b;    // error is small and positive and bounded by the max value of `a * (2^n/b - c)`

// this may be small enough that the following works. If not, then another approach is called for.
if (error >= b) {
   q++;
   error -= b;
}

// Now q is the result, and error is the remainder
```

This only works for unsigned values, but that is easy to handle in a wrapper function.

In this case, 1000000 is 2^6 * 5^6. We can cancel out the 2^6 factor, so we need to find

```
2^63 / (5^6)
```

I then wrote a program that tried all the combinations of shifting and adding (and subtracting) to find a small number of operations that would give sufficient accuracy.

The (paper that inspired me)[http://www.hackersdelight.org/divcMore.pdf] is this one.

PR #1683, opened 2016-12-26
Fix the error callback from not being called sometimes

### LINKED ISSUE #1680: MQTT: Connection failure callback not work
### Expected behavior
Connection failure callback reason should appear

### Actual behavior
Mqtt stuck.

### Test code
```Lua
    m = mqtt.Client("node-dev", 1)
    m:on("offline", function(con)
        print("E:Mqtt down!")
        m:close()
    end)
    m:connect("wf-dongle2.lan", 8883, 0,
      function(client) print("connected") end,
      function(client, reason) print("failed reason: "..reason)
    end)
```
### NodeMCU version
[http://nodemcu-build.com/builds/nodemcu-dev-11-modules-2016-12-23-17-08-53-integer.bin](url)
Dev branch, i guess it was commit 1452fbfd1394e4fe9073a2eb22c1a476d258a716
Code with  file, gpio, i2c, mqtt, net, node, pwm, rfswitch, tmr, uart, wifi.

### Hardware
Nodemcu-dev board

### Some debug : 
I found, that mqtt stuck only if try connect to  any *open* port of mqtt-server  - but w\o any listening service on it.
tcpdump result(on server side) proves it :
1) Port open, service down
Btw,  **only one try to connect!**

```
20:58:59.339071 IP (tos 0x0, ttl 255, id 8, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.36005 > wf-dongle2.lan.8883: Flags [S], cksum 0x3449 (correct), seq 6511, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 0008 0000 ff06 377e c0a8 017b c0a8  .,......7~...{..
        0x0020:  017a 8ca5 22b3 0000 196f 0000 0000 6002  .z.."....o....`.
        0x0030:  16d0 3449 0000 0204 05b4                 ..4I......
20:58:59.339395 IP (tos 0x0, ttl 64, id 56923, offset 0, flags [DF], proto TCP (6), length 40)
    wf-dongle2.lan.8883 > nodemcu-dev.lan.36005: Flags [R.], cksum 0x62c2 (correct), seq 0, ack 6512, win 0, length 0
        0x0000:  18fe 34d6 2f47 c83a 35d0 7690 0800 4500  ..4./G.:5.v...E.
        0x0010:  0028 de5b 4000 4006 d82e c0a8 017a c0a8  .(.[@.@......z..
        0x0020:  017b 22b3 8ca5 0000 0000 0000 1970 5014  .{"..........pP.
        0x0030:  0000 62c2 0000                           ..b...
```
Result - mqtt stuck

2) If i DROP packages from node, with simple rule:
```bash
iptables -A INPUT -i wlan0 -p tcp --destination-port 8883  -j DROP
```
Tcpdump result:
```
21:08:41.600489 IP (tos 0x0, ttl 255, id 23, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 0017 0000 ff06 376f c0a8 017b c0a8  .,......7o...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
21:08:42.595240 IP (tos 0x0, ttl 255, id 24, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 0018 0000 ff06 376e c0a8 017b c0a8  .,......7n...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
21:08:43.594872 IP (tos 0x0, ttl 255, id 25, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 0019 0000 ff06 376d c0a8 017b c0a8  .,......7m...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
21:08:44.596151 IP (tos 0x0, ttl 255, id 26, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 001a 0000 ff06 376c c0a8 017b c0a8  .,......7l...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
21:08:45.617655 IP (tos 0x0, ttl 255, id 27, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 001b 0000 ff06 376b c0a8 017b c0a8  .,......7k...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
21:08:46.626752 IP (tos 0x0, ttl 255, id 28, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 001c 0000 ff06 376a c0a8 017b c0a8  .,......7j...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
21:08:47.619849 IP (tos 0x0, ttl 255, id 29, offset 0, flags [none], proto TCP (6), length 44)
    nodemcu-dev.lan.25797 > wf-dongle2.lan.8883: Flags [S], cksum 0x5b71 (correct), seq 6695, win 5840, options [mss 1460], length 0
        0x0000:  c83a 35d0 7690 18fe 34d6 2f47 0800 4500  .:5.v...4./G..E.
        0x0010:  002c 001d 0000 ff06 3769 c0a8 017b c0a8  .,......7i...{..
        0x0020:  017a 64c5 22b3 0000 1a27 0000 0000 6002  .zd."....'....`.
        0x0030:  16d0 5b71 0000 0204 05b4                 ..[q......
```
Result: mqtt.client response correctly with, **after 5 connection attempts**:
```
> failed reason: -5
```

pjsg:
Fixes #1680 

There were a number of issues (now fixed);

* If the socket failed to connect before the timer went off, then it wasn't reported. It now is
* If the timer went off before the socket connected, then the connection attempt was not aborted.
* If auto-reconnect was set, then errors were never reported. Now they are reported on the first connection attempt. 

The downside is that there is a behavior change -- if the first connection fails, then reconnect is not performed.


devsaurus:
> The downside is that there is a behavior change -- if the first connection fails, then reconnect is not performed.

IMO this isn't really a downside but a clearer behavior. I like auto reconnect, but only in case a connection was established at all in the beginning. Observing that connection can be achieved at least once rules out a bunch of potential networking problems. Automatic algos can then keep on massaging the network layer as required.

devyte:
I mentioned this elsewhere, but I thought to bring it up here as well. Feel free to yell at me if I'm talking nonesense :)
I think it may make sense to keep trying to reconnect even if the first attempt fails. Consider the case where the ESP and the broker are on the same power rail, and power fails. When power comes back, both ESP and broker will start up at the same time, but the ESP will likely come online first (boot time is short). It will then try to connect to the broker, and fail, because the broker isn't online yet. If the auto-reconnect stops after the very first fail, it will never connect. The user would then need to add a whole additional layer on top to detect if the broker is online before trying to connect.
Wouldn't a better api be to specify a max number of auto-reconnect attempts, then give up and trigger a callback if all attempts failed?

devsaurus:
@devyte I noticed your post elsewhere after posting here and think that the case you describe should be considered.
My statement above originated from the understanding that the `autoreconnect` option means to automatically reconnect when the connection dropped - with the focus on the **re**. My view was probably a bit too narrow when assuming that a connection has been established initially.

Picking up your API proposal (not fully backwards compatible though):
`autoreconnect` n connection retries: 0 for `false`, > 0 for n retries, < 0 to retry infinitely, default 0

Does this sound meaningful?

devyte:
That looks about right! Thank you for taking my feedback into account :)

pjsg:
I'm think slightly differently -- how about this?

* Autoreconnect = 0    -- no auto reconnect behavior.  Error signalled via error callback
* Autoreconnect = 1    -- auto reconnects after first success. Error signalled via error callback.
* Autoreconnect = 2    -- auto connect/reconnect. *ALL* errors signalled via error callback.

The error callback gets an additional argument that says whether this is a 'final' error or not. A final error means that no further activity is expected.

Also, make sure that retries are done with an exponential backoff (with a max of 1000 seconds). 

I really don't want to have a mode where, if the server ip is wrong (or password is wrong), there is no error indication given. Of course, the programmer can ignore the error, or pass a nil as the error callback.

marcelstoer:
Thanks for picking this up Arnim.

> `autoreconnect` n connection retries: 0 for `false`, > 0 for n retries, < 0 to retry infinitely, default 0

It may be a perfectly valid approach in the C/embedded world, I have little to compare it with, but I don't like that. I prefer explicit APIs.
I'd also argue that the "n retries" feature won't be used much. Isn't it so that you either want retry or not and if so you want it to be infinite? Furthermore, which ever way the MQTT behaviour will eventually be  modified it should be consistent with the WiFi auto-(re)connect feature.

devyte:
@marcelstoer an explicit api is ok too. I don't really have a preference for how this is offered.

The name "autoreconnect" does not necessarily imply true|false. It can be the name of a feature, which takes many parameters. But like I said, I don't really have a preference for the api, so I'd rather leave that up to you :)

The n-retries approach is widely used in pretty much everything, including the Linux OS itself. It pretty much says: "retry n times, after that give up". Therefore, I would argue that it would be used quite often, especially in view of what @pjsg said.
Some users, such as me, would prefer to keep on trying for a very long time, or maybe even forever. This makes for a "soft connection" behavior, where connection errors can be recovered without restarting every single ESP in the network.
To explain this a bit: my use case is a wide wifi with many wifi routers interconnected via WDS. The thing is, the communication between the routers is not always stable, and sometimes I even have to reboot them. If I reboot the routers, all ESPs lose communication to the broker. If the autoreconnect is not handled properly, the ESPs would just give up and sit there staring at the wall. You could argue that specifying a timeout would be enough, but to be honest I've seen the wifi network take from a few seconds up to 15mins to recover. I prefer to have the ESPs just keep trying, eventually I get the wifi back up, and then the ESPs recover by themselves.

Also, like I explained in the previous post, all ESPs, the broker, and all wifi routers, among other devices, are on the same power circuit. If the power circuit goes down, when it comes back up all devices will power on at the same time, but will come online at very different times. I suspect that this is rather common among users, as is the case where the ESP is faster to come online that the broker. Therefore, if the first connection attempt by the ESP MQTT to the broker fails, it makes sense to keep retrying, at least for a while. This is also covered by the same n-retries approach, so it makes for simple usage for the programmer covering both cases in the same way.

marcelstoer:
> make sure that retries are done with an exponential backoff (with a max of 1000 seconds).

Yes, I believe that's important, too. I can't judge though whether 1k seconds is long enough.

devyte:
@pjsg 
you said:

> make sure that retries are done with an exponential backoff

Exponential backoffs are usually done to cover two cases at rather opposite ends of the spectrum: 
-quick and fast, where minimal overhead is desired
-long and slow, in which case you don't want to keep trying quickly, e.g.: overhead of waiting an extra few mins after an hour-long outage is ok

In this MQTT reconnect case, I don't think it's necessary to cover both ends of the spectrum. Personally, I feel that adding exponential backoff here would overcomplicate things, but that's just my own thought.

you said:

> Autoreconnect = 1 -- auto reconnects after first success

I assume that you want to force an error if the first connection attempt fails, and only try to auto-reconnect if the first try succeeds. This is a Bad Idea, like I explained previously.

you said:

> error callback gets an additional argument that says whether this is a 'final' error or not

This is a very good idea, it makes sense to know inside the callback whether more attempts will be tried or not.

you said:

> I really don't want to have a mode where, if the server ip is wrong (or password is wrong), there is no error indication given. Of course, the programmer can ignore the error, or pass a nil as the error callback.

The callback serves the purpose of error handling. If the callback receives the "final" argument, or "i-th of n" attempts argument, then an error indication can be implemented, either for each auto-reconnect attempt, or only for the final one (i.e.: I tried n times and I'm now giving up), or both.
If the programmer is too lazy to implement proper error handling, or ignores the error, or passes nil for the callback, then I'd say he deserves his fate. Error handling is fundamental in any half-way decent code.

marcelstoer:
I removed the 2.0.0 milestone as we're past the two-weeks-before-master-drop milestone and there are some unresolved issues (as per this discussion).

karrots:
It seems like auto-reconnection after the error without handling the error doesn't do anything towards fixing the error. Aside from connection time-out in @devyte's power example, everything else needs the Lua programmer or user to take action. If the call back worked correctly then the Lua programmer can decide to reconnect or stop and notify the user.

As it stands now the error handler doesn't always get called and if you try to stop the auto-reconnect by calling `close` on the client object the connections keep happening until the error handler happens again and gives a nil reference error.

It seems like first steps should be to fix the bugs in the system; then improve the system with preference changes (if still necessary).

1. Fix the error call back so it works consistently.
2. Fix the ability to stop auto-reconnect.

After those two items are fixed the Lua programmer can choose how to correctly deal with each error. Currently, they feel tied down because the system isn't working consistently.

pjsg:
The current approach is that the erro callback is *supposed* to be called at most once after a connect call is made. This is a final callback, no further action will be taken.

If autoreconnect means "always retry errors" then the programmer will get no error callbacks at all (there is no final error). 

I'm thinking that autoreconnect true means that errors are retried (no matter what the error is). I'm going to stick a reasonable timeout there -- say 10 seconds, However, *all* errors will be signalled to the error callback. An argument will be passed to say if there is a retry pending. Calling `close` would stop the retry.

Does this meet people's needs?

devyte:
Again, consider the case where the ESP and the broker power on at the same
time. The ESP will take a few seconds to come online and try to connect to
the broker. However, the broker could be a large system like a pc, or at
least a mini-version of one, generally linux-based. How long would such a
system take, from power on to the broker coming online? If you really must
put a timeout on retries, this must be taken into account in a robust
manner. Otherwise, after every power failure, all ESPs will need to be
restarted manually, because they will have timed out.

pjsg:
The programmer has two choices:

* They can manage the retries themselves. Don't set the auto-reconnect flag

* The system will continually retry (for ever). Each time the connection fails, the error callback will be called. If the programmer wants to stop after 3 hours of errors, then they can do that and call :close()

I suspect that most people will choose the second option (infinite retries).

pjsg:
Actually, now I stare at the mqtt code some more, I'm not sure that I can make it do anything consistent around reconnects.

I'm now feeling that the auto-reconnect option is a very bad idea. The programmer should deal with the reconnects by doing these from the error callback. The error callback should be guaranteed to be called at most once per connection. It should be called whenever the system gives up trying to get a good connection, or an existing connection breaks.

Also, given that the default for `clean_session` is true, it doesn't make sense to transparently reconnect anyway (as the broker throws away the state when the connection dies).

I'm now in favor of the following:

* Mark the auto-reconnect as deprecated with large warnings.
* Work on the reliability of the error callback
* Add an example showing how to do reconnect on error.

marcelstoer:
Philip, your proposal sounds reasonable, I'm ok with that.

> Mark the auto-reconnect as deprecated with large warnings.

Had we somehow resolved #1538 we could make sure programers don't miss that.

pjsg:
I've updated the documentation to make it clear that auto-reconnect should not be used. It particular, it interacts badly with `cleansession`.

I think that the underlying problem (of the missing callback) is also resolved here.

NicolSpies:
@pjsg Way back when I started reporting about the mqtt becoming "stale" or unresponsive in #1406 the reconnect functionality was part of the situation.  The following would have been nice:

If auto-reconnect is set, when the mqtt connection becomes "stale" or unresponsive, auto-reconnect is tried as many times as I have told it to try to reconnect. If it is does not succeed after the tries the "offline" event in triggered with a auto-reconnect "true" flag allowing me to "know" reconnect failed and will allow for specific code in the callback.
If auto-reconnect is not set, the "offline" event is triggered immediately with a auto-reconnect "false" flag.
This way the developer can let mqtt try to reconnect first before passing the "offline" callback.

Now, different scenarios I experienced complicates the suggestion above. The scenarios consists of a combination of one or more of the following:

ESP connection to AP failed.
ESP connected to AP but AP connection to the internet failed.
ESP connected to AP and AP connected to internet but connection to the mqtt broker failed.
Dependent on the existence and combination of the above, auto-reconnection is futile and hence my original response regarding the "ping" functionality to establish what has failed before attempting an reconnect.

Thus as a final thought regarding the auto-reconnect: It should only be attempted if none of the 3 conditions above are true or present. If this can be incorporated in the back-end functionality and returned as flags in the callback it will be great.

jmattsson:
Is this waiting for a particular milestone?

marcelstoer:
> Is this waiting for a particular milestone?

No, not really. But maybe for another approval (besides mine) from someone more familiar with the C-side of things 😉
Furthermore, the deprecation warning could now be embedded in the code (see https://github.com/nodemcu/nodemcu-firmware/pull/1683#issuecomment-274420203).

jmattsson:
👍 from me then, and +1 on adding deprecation warning in code

marcelstoer:
I just tried to add the deprecation warning to the code...and I failed 😞 

I added this

```
if ( auto_reconnect == RECONNECT_POSSIBLE ) {
    platform_print_deprecation_note("autoreconnect == 1 is deprecated", "in the next version");
}
```

to https://github.com/nodemcu/nodemcu-firmware/pull/1683/files#diff-904c0a57714312fb890c87ccf9beb2a4R1076 but the compiler complained about

> implicit declaration of function 'platform_print_deprecation_note'

Why is this? After all, `platform.h` _is_ included.

devsaurus:
> Why is this? After all, `platform.h` is included.

Yes, but the feature branch in this PR diverged before the deprecation function in https://github.com/nodemcu/nodemcu-firmware/pull/1538/commits/ba9d3af68f1e76a1e71fa9d87ae73b2258763851 was merged. `platform_print_deprecation_note()` is not included of this code base.

marcelstoer:
Oh sorry, silly me...why didn't I check. I didn't realize this PR is _that_ old. I'll merge it then and add the snippet to `dev` right afterwards.

marcelstoer:
There you go: b645100.

PR #1690, opened 2016-12-30
added 31250 baud rate for MIDI support

verneroberts:

MIDI works on 31250 baud rate so it makes sense to have this as an option

### DIFF
--- a/app/platform/platform.c
+++ b/app/platform/platform.c
@@ -361,6 +361,7 @@ uint32_t platform_uart_setup( unsigned id, uint32_t baud, int databits, int pari
     case BIT_RATE_4800:
     case BIT_RATE_9600:
     case BIT_RATE_19200:
+    case BIT_RATE_31250:
     case BIT_RATE_38400:
     case BIT_RATE_57600:
     case BIT_RATE_74880:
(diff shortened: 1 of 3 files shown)


app/platform/platform.c
```diff
@@ -361,6 +361,7 @@ uint32_t platform_uart_setup( unsigned id, uint32_t baud, int databits, int pari
     case BIT_RATE_4800:
     case BIT_RATE_9600:
     case BIT_RATE_19200:
+    case BIT_RATE_31250:
```
pjsg: You probably also want to add this bitrate to the uart_bitrates array. This will allow the get operation to return this bitrate.

verneroberts: cool, added.

djphoenix:
If it really works - it's something that really nice to have.

marcelstoer:
> MIDI works on 31250 baud rate

"MIDI & ESP" sounds interesting. What are you working on?

verneroberts:
@marcelstoer I've got a few ideas, I've initially created a record/playback loop device that plugs into my piano so I can play along with myself. It may not actually be feasible to reproduce the timing accurately enough from lua but that's some of the fun. Ultimately I'm keen to create midi files which can upload/download to a PC. I'm also playing with tunelling midi RX/TX over wifi so I can do lots of crazy things from a host PC.

Issue #778, opened 2015-11-19
Query re wifi.sta.config() 


alexgian:
This may not be an issue but just ignorance on my part - if so, my apologies in advance!
I am using NodeMCU 0.9.6 with Lua 5.1.4, build 20150704.

I can connect fine to networks that use normal text string passwords, but with networks where the password is a five-byte string I cannot.

I have tried a few different formats with no success.
Say the password is A1B2C3D4E5.  I have all tried the following, including their lowercase versions:

"A1B2C3D4E5"
"A1 B2 C3 D4 E5"
"A1-B2-C3-D4-E5"
"A1:B2:C3:D4:E5"
and even
"\161\178\195\212\229" (in which case it printed "stdin:1: pwd:0,8~64" but still didn't work)
Is there something I am missing here, or there some problem?

Thanks

karrots:
You'll want to try a newer build from the dev branch. See #719.

http://frightanic.com/nodemcu-custom-build/

dnc40085:
I believe the minimum password length for WPA/WPA2 is 8 characters. 
Is it possible to change the 5 byte password to an 8-63 byte password?

alexgian:
OK, thanks, dnc40085.
The WLANs I was trying were protected with WEP-64, which I will now assume is not compatible with NodeMCU.
Fortunately, I have the capability of changing them over to WPA2 (I'm not really sure what the difference is, anyway). I will do that and not waste more time.  

Does this mean, however, that someone stuck with a WEP-64 system would not be able to use NodeMCU?

karrots:
WPA2 should be supported by any 802.11g or better device.

alexgian:
OK, thanks, all done, closing this now.

dnc40085:
NodeMCU should be capable of connecting to a WEP encrypted network provided that the password is 8-63 bytes long. However...

WEP encryption should not be used by anyone since it is [no longer secure](http://www.isaac.cs.berkeley.edu/isaac/wep-faq.html). WiFi networks secured with WEP are [ridiculously easy to crack](http://www.freehowtohackwifi.com/wep-attacks/fern-wifi-cracker/), all you need is a backtrack or Kali Linux live cd, a wifi adapter that supports packet injection and about 30 minutes and it will spit out the password needed to connect to the network.

[...]

PR #1694, opened 2017-01-01
WPS functionality added

### LINKED ISSUE #1556: WPS support
Is there any plan to add support for WPS (auth via button)? This is crucial for applications where you have no other method of getting an SSID/password onto a device, so I am a little bit surprised that there is no support for it. The ESP8266 Arduino core (https://github.com/esp8266/Arduino) has support for it.

If nobody else is working on this I will make a start, but I'm interested to know if there is some reason that this hasn't been added yet.

FrankX0:
Fixes #1556 .

Although there seems to be some opposition against using WPS, I still think it is worthwhile making it available to whomever wants to use it.
Please bare in mind that my software/Github skills are somewhat limited.
Looking forward to your comments/acceptance.


app/modules/wps.c
```diff
@@ -0,0 +1,74 @@
...
+    lua_pushinteger(L, status);
+    lua_call(L, 1, 0);
+  }
+}
+
+// Lua: wps.start( function())
+static int ICACHE_FLASH_ATTR wps_start(lua_State* L)
+{
+  // retrieve callback arg (optional)
+  if (wps_callback_ref != LUA_NOREF)
```
pjsg: You don't need to do the test -- you can unref LUA_NOREF

app/include/user_modules.h
```diff
@@ -68,6 +68,7 @@
 //#define LUA_USE_MODULES_UCG
 //#define LUA_USE_MODULES_WEBSOCKET
 #define LUA_USE_MODULES_WIFI
+#define LUA_USE_MODULES_WPS
```
pjsg: This should default to OFF

docs/en/modules/wps.md
```diff
@@ -0,0 +1,63 @@
+# WPS Module
```
marcelstoer: A reference to this file needs to be added to /mkdocs.yaml.

marcelstoer:
The CI build fails with 

> section `.text` will not fit in region `iram1_0_seg`

Looks like the fix that @djphoenix provided in #1566 didn't last long.

marcelstoer:
I don't mind having this available as an off-by-default module. However, I think we should prominently mention the [security concerns](https://github.com/nodemcu/nodemcu-firmware/issues/778#issuecomment-158203376) our WiFi module maintainer @dnc40085 raised a long time ago in the docs.

FrankX0:
I guess we can add a statement to the WPS module documentation:

> Use with caution: there are (serious) security concerns about using WPS.

FrankX0:
@marcelstoer, @djphoenix: any ideas how to fix the CI build error?

marcelstoer:
I don't, not my area of expertise 😞

djphoenix:
Will take a look some later.

djphoenix:
Rebased this PR on current `dev` and built `all-modules + ssl` configuration. Build succeeds for `integer` version but fails for `float`.

After temporary increase `iram0` segment size I tried to link binary again and it succeeds, and so size of `iram` is `0x8045` (69 bytes overflow).

For `dev` branch `iram` size is `0x7d45` (difference with `wps` is 768 bytes).

*And the winner is...*
`libwps.a` is placed in `irom` at most. But size of all symbols in `text` section is...
```
$ gobjdump -h sdk/esp_iot_sdk_v2.0.0/lib/libwps.a | grep ' .text' | awk '{print $3}' | grep -v 0000000b
000000aa
00000094
00000082
00000068
000000aa
$ node
> 0xb + 0xaa + 0x94 + 0x82 + 0x68 + 0xaa
733
```
Very close to what we see...

The solution is patch LD script as [I was originally propose](https://github.com/nodemcu/nodemcu-firmware/pull/1566#issuecomment-260100969) - keep only `main`, `pp`, `phy`, `net80211` and `gcc` libs of SDK in `iram`. But I cannot guarantee that wps will keep work in `irom`. Cannot test this solution myself because in my router WPS completely disabled [for security reasons](http://www.howtogeek.com/176124/wi-fi-protected-setup-wps-is-insecure-heres-why-you-should-disable-it/).

FrankX0:
@djphoenix: I am willing to test, but it's not clear how to modify the current dev LD script.

djphoenix:
In `ld/nodemcu.ld` file, near line 105:
```diff
+    *libmain.a:*(.literal .text)
+    *libnet80211.a:*(.literal .text)
+    *libphy.a:*(.literal .text)
+    *libpp.a:*(.literal .text)
+    *libgcc.a:*(.literal .text)
-    *sdk/esp_iot_sdk_*lib/lib*.a:*(.literal .text)
```

FrankX0:
Thanks @djphoenix: no problems compiling.
WPS functionality still works OK.
I tested using the  physical button implementation, not through PIN.
This I have disabled (for what it is worth).

app/modules/wps.c
```diff
@@ -0,0 +1,73 @@
...
+#include "module.h"
+#include "lauxlib.h"
+#include "platform.h"
+
+static int wps_callback_ref;
+
+// Lua: wps.disable()
+static int ICACHE_FLASH_ATTR wps_disable(lua_State* L)
+{
+  wifi_wps_disable();
```
djphoenix: I think there should be error handler, as far as wps-related functions (in includes) returns bool as success/failure status.

app/modules/wps.c
```diff
@@ -0,0 +1,73 @@
...
+{
+  // retrieve callback arg (optional)
+  luaL_unref(L, LUA_REGISTRYINDEX, wps_callback_ref);
+
+  wps_callback_ref = LUA_NOREF;
+
+  if (lua_type(L, 1) == LUA_TFUNCTION || lua_type(L, 1) == LUA_TLIGHTFUNCTION)
+    wps_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+  else
+    return luaL_error (L, "Argument not a function");
```
djphoenix: With `luaL_checkanyfunction` no needed to manually raise an error.
And so safer way is check arguments *before* any internal state change.

app/modules/wps.c
```diff
@@ -0,0 +1,73 @@
...
+
+// Lua: wps.start( function())
+static int ICACHE_FLASH_ATTR wps_start(lua_State* L)
+{
+  // retrieve callback arg (optional)
+  luaL_unref(L, LUA_REGISTRYINDEX, wps_callback_ref);
+
+  wps_callback_ref = LUA_NOREF;
+
+  if (lua_type(L, 1) == LUA_TFUNCTION || lua_type(L, 1) == LUA_TLIGHTFUNCTION)
```
djphoenix: `luaL_checkanyfunction` should be there

app/modules/wps.c
```diff
@@ -0,0 +1,73 @@
...
+    wps_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```
djphoenix: There will be issue when more than one argument is passed, so call `lua_pushvalue(L, 1)` before `luaL_ref`

ld/nodemcu.ld
```diff
@@ -103,7 +103,11 @@ SECTIONS
     *(.init.literal)
     *(.init)
     
-    *sdk/esp_iot_sdk_*lib/lib*.a:*(.literal .text)
+    *libmain.a:*(.literal .text)
+    *libnet80211.a:*(.literal .text)
+    *libphy.a:*(.literal .text)
+    *libpp.a:*(.literal .text)
+    *libgcc.a:*(.literal .text)
```
djphoenix: I think this change is outside this PR's scope.
@marcelstoer @jmattsson should we move it in another issue/PR or keep there?

marcelstoer: I prefer a dedicated PR.

jmattsson: I'm *really* wary of this change altogether. As you already pointed out Yury, Espressif has put almost everything into `irom0.text`, except for what appears to be a few internal pieces. Presumably they had a really good reason for keeping those in IRAM. Maybe it would be worthwhile raising the large iram1_0_seg size with Espressif as a potential bug? If they respond, we'd know for certain whether it has to be kept in IRAM or not.

djphoenix: @jmattsson Johny, as I reviewed, a lot of Espressif code is a mess. I think there was some internal Espressif's issues e.g. "radius of the developers hands curvature does not comply with the platform requirements" :)
Anyway for WPS we decide that it works when fully putted in irom. And for most places it right (with exceptions for interrupts-related stuff). I haven't see any of interrupt-related code here.

jmattsson: Hah! Well I can't argue too strongly here seeing as I've used the "I've reviewed and tested it, and it works" argument myself in regard to flash vs ram.

Could we perhaps leave a comment in the `nodemcu.ld` file indicating that we're specifically excluding `ilbwps.a` due to size constraints?

djphoenix: @jmattsson I think we should move this discussion to #1710.
For @FrankX0 - just revert this part, we will fix this issue in other way.

marcelstoer: > For @FrankX0 - just revert this part, we will fix this issue in other way.

The cool thing is that GitHub allows us to do that yourselves right in the browser. You can't revert but you can resolve conflicts by editing code - which is just did.

marcelstoer:
I'll add the doc amendment we talked about at https://github.com/nodemcu/nodemcu-firmware/pull/1694#issuecomment-269920013 after merging.

Thanks Frank, Yury and Johny.

PR #1695, opened 2017-01-01
Replace old tmr API in docs with OO API

### LINKED ISSUE #1685: Upgrade tmr code in the docs
The non-OO tmr API is deprecated. Yet, all code snippets in our docs that rely on timer still use the old API.

-> `grep -r --include "*.md" --exclude "tmr.md" "tmr." .`

Same goes for https://en.wikipedia.org/wiki/NodeMCU#Connect_to_an_AP

marcelstoer:
Fixes #1685.

This is just one way of fixing the examples in the documentation. I'm open for alternatives.

### DIFF
--- a/docs/en/lua-developer-faq.md
+++ b/docs/en/lua-developer-faq.md
@@ -70,7 +70,7 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
 * Whilst the SDK provides a number of interrupt driven device drivers, the hardware architecture severely limits the memory available for these drivers, so writing new device drivers is not a viable options for most developers
 * The SDK interfaces internally with hardware and device drivers to queue pending events.
 * The registered callback routines are invoked sequentially with the associated C task running to completion uninterrupted.
-* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua [`tmr:alarm(interval, repeat, callback)`](modules/tmr.md#tmralarm) function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.  
+* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua [`mytimer:alarm(interval, repeat, callback)`](modules/tmr.md#tmralarm) function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.  
 
 The NodeMCU firmware simply mirrors this structure at a Lua scripting level:
 
--- a/docs/en/lua-developer-faq.md
+++ b/docs/en/lua-developer-faq.md
@@ -70,7 +70,7 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
 * Whilst the SDK provides a number of interrupt driven device drivers, the hardware architecture severely limits the memory available for these drivers, so writing new device drivers is not a viable options for most developers
 * The SDK interfaces internally with hardware and device drivers to queue pending events.
 * The registered callback routines are invoked sequentially with the associated C task running to completion uninterrupted.
-* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua `tmr.alarm(id, interval, repeat, callback)` function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.  
+* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua [`tmr:alarm(interval, repeat, callback)`](modules/tmr.md#tmralarm) function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.  
 
 The NodeMCU firmware simply mirrors this structure at a Lua scripting level:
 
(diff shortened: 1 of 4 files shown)


docs/en/lua-developer-faq.md
```diff
@@ -70,7 +70,7 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
 * The SDK interfaces internally with hardware and device drivers to queue pending events.
 * The registered callback routines are invoked sequentially with the associated C task running to completion uninterrupted.
-* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua `tmr.alarm(id, interval, repeat, callback)` function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.  
+* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua [`tmr:alarm(interval, repeat, callback)`](modules/tmr.md#tmralarm) function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.
```
devsaurus: `tmr:alarm()` implies that the module got overloaded by the variable's name. Maybe `mytimer:alarm()`?

docs/en/modules/tmr.md
```diff
@@ -34,7 +34,12 @@ To free up the resources with this timer when done using it, call [`tmr.unregist
 
 #### Example
 ```lua
-if not tmr.alarm(0, 5000, tmr.ALARM_SINGLE, function() print("hey there") end) then print("whoopsie") end
+if not tmr.create():alarm(5000, tmr.ALARM_SINGLE, function()
```
devsaurus: I like the `tmr.create():alarm()` thing - looks like a nice single-command substitution. But haven't actually tried it myself. Does it survive the next garbage collection before the timer expires?

docs/en/modules/tmr.md
```diff
@@ -235,25 +244,29 @@ If the specified timer is registered, returns whether it is currently started an
...
-`id`/`ref` timer id (0-6) or object
+`id`/`ref` timer id (0-6) or object, obsolete for OO API (→ [`tmr.create()`](#tmrcreate))
 
 #### Returns
 `true` if the timer was stopped, `false` on error
 
 #### Example
 ```lua
-if not tmr.stop(2) then print("timer 2 not stopped, not registered?") end
+if not tmr.create():stop() then print("timer not stopped, not registered?") end
```
devsaurus: Is this meaningful to create and immediately stop a timer?

marcelstoer: Not really, you're right. But the existing example (by itself) was just as meaningless. It's an example to demonstrate "`false` on error"...How would you like to see this improved?

docs/en/lua-developer-faq.md
marcelstoer: True, will fix.

docs/en/modules/tmr.md
marcelstoer: Yes, it does:

```lua
print("before: " .. collectgarbage("count")*1024)
if not tmr.create():alarm(5000, tmr.ALARM_SINGLE, function()
  print("hey there")
end)
then
  print("whoopsie")
end
print("while running: " .. collectgarbage("count")*1024)
```

Prints 
> before: 3072
> while running: 4096
> hey there

If you run `print("after: " .. collectgarbage("count")*1024)` after the timer finished it yields

> after: 3072

docs/en/upload.md
```diff
@@ -62,11 +62,12 @@ print("Connecting to WiFi access point...")
...
 wifi.sta.config(SSID, PASSWORD)
 -- wifi.sta.connect() not necessary because config() uses auto-connect=true by default
-tmr.alarm(1, 1000, 1, function()
+wifitimer = tmr.create()
+wifitimer:alarm(1000, tmr.ALARM_AUTO, function()
     if wifi.sta.getip() == nil then
         print("Waiting for IP address...")
     else
-        tmr.stop(1)
+        wifitimer:stop()
```
pjsg: This is an interesting example. Does the wifitimer object get gc'ed? I suspect that the callback holds a reference to the timer and hence it will never be GC'ed. It is that same (anti-)pattern that causes leakages in sockets in the net module.

I wondered whether the callback was invoked with the timer object as an argument -- but the documentation for `register` doesn't really describe the `func` argument.... I suspect that (like with the network callbacks), the func argument should be called with the timer object as the first parameter.... This makes it easy to write code that doesn't retain the extra reference.

docs/en/modules/tmr.md
devsaurus: Just use the `mytimer`? It has been introduced as a generic name for a random timer object above.

marcelstoer: I like them to be as self-contained as possible. Pushed a small improvement.

docs/en/upload.md
marcelstoer: Great input Philip! The callback _does_ get timer object back, I just tested:

```lua
mytimer = tmr.create()
mytimer:alarm(1000, tmr.ALARM_AUTO, function(cb_tmr) 
    cb_tmr:stop()
    print("hey there")
    print(cb_tmr:state())
end)
```
Prints
> hey there
false	1

and stops the timer. I'll document the missing `func` parameter and update this example.

docs/en/upload.md
```diff
@@ -62,11 +62,12 @@ print("Connecting to WiFi access point...")
...
 wifi.sta.config(SSID, PASSWORD)
 -- wifi.sta.connect() not necessary because config() uses auto-connect=true by default
-tmr.alarm(1, 1000, 1, function()
+wifitimer = tmr.create()
+wifitimer:alarm(1000, tmr.ALARM_AUTO, function(cb_timer)
     if wifi.sta.getip() == nil then
         print("Waiting for IP address...")
     else
-        tmr.stop(1)
+        cb_timer:unregister()
```
luismfonseca: I didn't test this, but it looks like a reference to `wifitimer` will linger on...

Perhaps the example should have a `wifitimer = nil` after this line to fully dispose of all the references so that it can be GC'ed. 

But this code would be better by avoiding the assignment to the `wifitimer` variable in the first place.

marcelstoer: True, thanks for the reminder, will fix. The reference to `wifitimer` is no longer necessary.

PR #1698, opened 2017-01-02
Dev

thomo:
Improvement

The UCG GraphicTest example first executes a full screen test and then a clip range test - and repeat this. But the clip range test takes 252 iterations (while the full screen only takes 4). The request change it so that both take 4 - so both takes the same time.


devsaurus:
Good find @thomo, thanks a lot!

PR #1702, opened 2017-01-04
Depend http client secure part on CLIENT_SSL_ENABLE

### LINKED ISSUE #1673: Problem: HTTP module is now too large for 512k Flash devices
### Expected behavior
Building a firmware with HTTP and only the bare minimum modules (file gpio net node tmr uart wifi) will run on a 512k flash ESP.  

### Actual behavior
The firmware clocks in at 582,544 bytes, and unsurprisingly doesn't fit/run.  It also won't run on 16MB modules (Wemos D1 Mini Pro).  The same commit built without the http module (7 modules) runs on both.

**512k Modules Errors**
```
Fatal exception (28): 
epc1=0x4000228b, epc2=0x00000000, epc3=0x00000000, excvaddr=0x000000b5, depc=0x00000000
```

**WeMos D1 Mini Pro Errors**
```

load 0x40100000, len 27148, room 16 
tail 12
chksum 0xe6
ho 0 tail 12 room 4
load 0x3ffe8000, len 2424, room 12 
tail 12
chksum 0x06
ho 0 tail 12 room 4
load 0x3ffe8978, len 136, room 12 
tail 12
chksum 0x4c
csum 0x4c
system param error, use last saved param!
rf_cal[0] !=0x05,is 0xCE
```

### NodeMCU version
Dev branch: c5c0143b2fa6402915beca2a50da7241247e1409

### Hardware
ESP-03, ESP-12 (512k), WeMos D1 Mini Pro

### Notes:
- The addition of the http module adds a whopping 171176 bytes or nearly 30% of the entire firmware!  That is kind of crazy! 
- Presumably the changes from sdk 2.0.0 caused the size of the http module to balloon.  A 10-module build (including http and a few others) from the last master was only 457kb.  
- Besides the extreme case of including ONLY the http module, it's kind of important that at least 10-12 modules (eg I2C and one or two sensor modules) fit on 512k.  Otherwise, these devices are more or less useless with the current firmware.  I realize they're a bit outdated and could have their chips upgraded, but I still have quite a few and if they're shielded, the upgrade is a fair bit of work.  For simple tasks, they've previously been fine.  
- Perhaps more importantly, this also effects the Wemos D1 Mini Pro (and probably other 16MB modules).  These appear to have the same limitations on firmware size that a 512k module would have, at least for now.  I haven't gotten a firmware build larger than 512k to work on them (but see [#1634] & [#1646]).  So the end result is that http will not work on these at all.

djphoenix:
Fixes #1673

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -27,7 +27,9 @@
 typedef struct request_args_t {
 	char		* hostname;
 	int		port;
+#ifdef CLIENT_SSL_ENABLE
 	bool		secure;
+#endif
 	char		* method;
 	char		* path;
 	char		* headers;
@@ -140,9 +142,11 @@ static void ICACHE_FLASH_ATTR http_receive_callback( void * arg, char * buf, uns
 	{
 		HTTPCLIENT_ERR( "Response too long (%d)", new_size );
 		req->buffer[0] = '\0';                                                                  /* Discard the buffer to avoid using an incomplete response. */
+#ifdef CLIENT_SSL_ENABLE
 		if ( req->secure )
 			espconn_secure_disconnect( conn );
 		else
+#endif
 			espconn_disconnect( conn );
 		return;                                                                                 /* The disconnect callback will be called. */
 	}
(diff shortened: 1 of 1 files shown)


mickley:
I can also confirm that this allows the http module to be used on 512 kByte modules.  The newer SSL does not fit though as @devsaurus says.  So there's no support now for SSL on 512 kByte modules.  Not the end of the world, but it might be worth documenting that.

PR #1710, opened 2017-01-09
Leave only some parts of SDK in iram

djphoenix:
Amendment for #1562/#1566, follow-up #1694 [comment](https://github.com/nodemcu/nodemcu-firmware/pull/1694#issuecomment-270260399).

As we decided, for example, `libwps` is safe to put it into `irom` (although Espressif marked some parts as "iram-forced"). And we cannot fit it into `iram` anyway (because `iram` segment size overflow).

So I bring back my original proposal to keep only some parts of SDK in iram. For now this is `main`, `net80211`, `phy`, `pp` and `gcc` libs (they're used in bootup process, in interrupts, and in other ways where iROM is unavailable). For another parts it is safe to put them in flash and free some instruction RAM.

### DIFF
--- a/ld/nodemcu.ld
+++ b/ld/nodemcu.ld
@@ -103,7 +103,11 @@ SECTIONS
     *(.init.literal)
     *(.init)
     
-    *sdk/esp_iot_sdk_*lib/lib*.a:*(.literal .text)
+    *libmain.a:*(.literal .text)
+    *libnet80211.a:*(.literal .text)
+    *libphy.a:*(.literal .text)
+    *libpp.a:*(.literal .text)
+    *libgcc.a:*(.literal .text)
 
     *(.iram.text .iram0.text)
 
--- a/ld/nodemcu.ld
+++ b/ld/nodemcu.ld
@@ -103,11 +103,24 @@ SECTIONS
     *(.init.literal)
     *(.init)
     
+    /* SDK libraries that used in bootup process, interruption handling
+     * and other ways where flash cache (iROM) is unavailable: */
     *libmain.a:*(.literal .text)
     *libnet80211.a:*(.literal .text)
     *libphy.a:*(.literal .text)
     *libpp.a:*(.literal .text)
     *libgcc.a:*(.literal .text)
+    
+    /* Following SDK libraries have .text sections, but not included in iRAM: */
+    /* *libat.a:*(.literal .text) - not used anywhere in NodeMCU */
+    /* *libcrypto.a:*(.literal .text) - tested that safe to keep in iROM */
+    /* *libdriver.a:*(.literal .text) - not used anywhere in NodeMCU */
+    /* *libespnow.a:*(.literal .text) - not used anywhere in NodeMCU */
+    /* *libmesh.a:*(.literal .text) - not used anywhere in NodeMCU */
+    /* *liblwip_536.a:*(.literal .text) - source-based library used instead */
+    /* *libpwm.a:*(.literal .text) - our own implementation used instead */
+    /* *libwpa.a:*(.literal .text) - tested that safe to keep in iROM */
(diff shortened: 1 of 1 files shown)


ld/nodemcu.ld
```diff
@@ -103,7 +103,11 @@ SECTIONS
     *(.init.literal)
     *(.init)
     
-    *sdk/esp_iot_sdk_*lib/lib*.a:*(.literal .text)
```
jmattsson: Could I request a comment here stating that we're intentionally excluding `libwps.a` even though it claims to have iram'd functions?

ld/nodemcu.ld
```diff
@@ -103,7 +103,11 @@ SECTIONS
...
+    *libmain.a:*(.literal .text)
+    *libnet80211.a:*(.literal .text)
+    *libphy.a:*(.literal .text)
+    *libpp.a:*(.literal .text)
+    *libgcc.a:*(.literal .text)
```
jmattsson: I don't think this list is complete. Have a look at the output of this:
```
nodemcu-firmware/sdk/esp_iot_sdk_v2.0.0/lib$ xtensa-lx106-elf-objdump -h *.a|grep -v '2\*\*0' | grep -v irom|grep -E 'format|.text|archive'
```
It seems the following libs also have some things in IRAM:
- ~~libat.a~~ (don't care, we don't use this)
- libcrypto.a
- libdriver.a (not sure if we're using any of this by now?)
- libespnow.a (not relevant yet, but might be worth listing here anyway)
- libmesh.a (as above)
- ~~liblwip_536.a~~ (crikey, espconn_buf.o has nearly 1.3k in IRAM! Lots of other bits too...)
- ~~libpwm.a~~ (we have our own pwm driver last I checked)
- libwpa.a
- ~~libwps.a~~ (intentionally excluded)

djphoenix: This list is incomplete for "all libs that have something in iRAM" but complete for "all libs that **should** have something in iRAM".
All libs that you listed above is called in flash-cache-safe ways. Originally in #1566 I was tested firmware carefully for hashing, WPA/WEP connections, and more ways, and doesn't found any issues with it. But later we choose the way to keep all SDK libs in iRAM.

Rationale of my way is to keep iRAM as clean as possible. Sure, originally Espressif wasn't just include "safe unaligned access" patch in core library, so I think there was something similar in their internal testing process (if it was generally), so after tons of changes, them kept listed libs in iRAM just for historical reasons.

jmattsson: Ah, sorry I forgot the testing you did back on that PR. And yeah, I'm not against kicking things out of IRAM if they don't *have* to be there, I'd just like it to be a very conscious decision, with a note about each lib we've done that to left in there.

Regarding unaligned accesses though, the constraints are the same for mapped flash and IRAM, but it's all a bit tangential to the main discussion :)

djphoenix: So okay, I was added a note about it. I think this is enough to go.
Also I propose to include SDK upgrade guide (to check iRAM/iROM contents) in `contributing.md`

jmattsson: Sounds like a good idea!

jmattsson:
Did you want Marcel or Arnim to have a look over this first, or are you happy for me to merge? And did you want to get the SDK upgrade not in here too?

djphoenix:
I think one review is enough, we have roughly tested this before.
About SDK upgrade flow - I think there will be a lots of discussions, and so I have not very good english skill to describe it right. So I propose to land this PR now and keep in mind (in Github issues) that we should work on contribution note.

marcelstoer:
> Did you want Marcel or Arnim to have a look over this first

Me? I have absolutely zero clue what you guys are doing here 😄 I'm just happy you fixed it - whatever that means.

PR #1727, opened 2017-01-15
make user/pass optional in MQTT client doc

Tropicao:
In the MQTT module documentation, about creating a MQTT client object, the "user" and "password" parameters are presented as mandatory. Having just started to use NodeMCU, I though I had to provide "fake" logins to be compatible with the API, hoping it would be ignored when connecting to a broker with no logins check (unfortunately, providing logins in this case prevent the module from connecting, tested with test.mosquitto.org).
It would be nice to aknowledge that those parameters are not necessary if the broker does not wait for logins.


marcelstoer:
Thanks

PR #1735, opened 2017-01-17
bme280: Correct code samples for negative values

### LINKED ISSUE #1734: bme280 reports wrong temperature around/below 0
### Expected behavior
Module returns actual temperature when reading it.

### Actual behavior
Around 0 degrees Celsius the temperature I read has 1-2° jumps in it and generally looks wrong:

I don't know the actual temperatures but I would guess something goes wrong when going below 0°. It already starts with the jump from 0 to 1 degree. I'm pretty sure this should not happen and the temperature was already below 0, before it jumped to -2°
I think the graph should have shown the temperature sink below 0 to roughly -2 and then rise again above 0.

I've never seen these jumps before when measuring higher temperatures! So I don't think it's my backend code.

### Test code
Nothing special here (I basically follow the example from the docs):

bmetype = bme280.init(PIN_BME280_SDA, PIN_BME280_SCL)
P = bme280.baro()
P = string.format("%d.%03d", P/10, P%10)
H, T = bme280.humi()
T = string.format("%d.%02d", T/100, T%100)

### NodeMCU version
	branch: master
	commit: 8e48483c825dea9c12b37a4db3d034fccbcba0bf
	SSL: true
	modules: bme280,cjson,dht,file,gpio,http,i2c,mqtt,net,node,spi,tmr,uart,wifi,ws2812,file,gpio,net,node,tmr,uart,wifi
 build 	built on: 2016-09-16 13:13

I know, rather old. But looking at bme280.c nothing changed since Aug 2, 2016

### Hardware
BME type: 2

dicer:
Fixes #1734.

- [x] This PR is for the dev branch rather than for master.
- [x] This PR is compliant with the other contributing guidelines as well (if not, please describe why).
- [x] The code changes are reflected in the documentation at docs/en/*.

The formatting/modulo will produce wrong values for negative numbers. Corrected the samples to make it work

### DIFF
--- a/docs/en/modules/bme280.md
+++ b/docs/en/modules/bme280.md
@@ -138,9 +138,9 @@ print(string.format("QNH=%d.%03d", QNH/1000, QNH%1000))
 
 H, T = bme280.humi()
 if T<0 then    
-  print(string.format("-%d.%02d", -T/100, -T%100))
+  print(string.format("T=-%d.%02d", -T/100, -T%100))
 else
-  print(string.format("%d.%02d", T/100, T%100))
+  print(string.format("T=%d.%02d", T/100, T%100))
 end
 print(string.format("humidity=%d.%03d%%", H/1000, H%1000))
 D = bme280.dewpoint(H, T)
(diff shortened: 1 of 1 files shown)


marcelstoer:
Looks good to me.

PR #1738, opened 2017-01-18
invert signal on rfswitch pin

ffedoroff:
since that code was migrated from arduino, I should to invert pin


marcelstoer:
Thanks for the fix and test.

PR #1743, opened 2017-01-19
Minify ESP8266 toolchain archive for CI

### LINKED ISSUE #884: Travis SDK contains a lot of bloatware
We copied the SDK used for Travis CI from George Hahn's repository.  It's 32Mb extra in ours.  Given that it's already a blob in our Repo, it's not worth changing at this time but if we do need to change it in the future 
-   Don't use GZ encoding; use XZ.  This is just _so_ much more efficient.  OK, compression takes ages, but we only uncompress during a Travis CI build and this is as fast as GZ uncompression.
-   Lose all of the unused stuff in the tarball -- that is only retain that which is needed to build nodemcu.  This includes c++ compiler, libraries and includes; gdb; all of the sources not used in our make.

This will more than half the size of the `.tar.xz` file compared to the `.tar.gz` version.

djphoenix:
Fixes #884

I used my script-set to build toolchain (on clean ubuntu 12.04 installation):
- `crosstool-NG` for `xtensa-lx106-elf` target, with `-mforce-l32` patch
- `libhal` and headers (integrated in toolchain sysroot)
- `lib[cm]irom` (integrated in toolchain sysroot)

So I removed `esptool.py` from archive and added it to `pip` in `.travis.yml`.

TAR archive was compressed with `xz -9e` to compress it as much as possible.

Also in this PR I removed `irom` size fix from `pr-build.sh` because it already applied in my previous LD fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,12 @@ cache:
   - directories:
     - cache
 before_install:
-- pip install --user hererocks
+- pip install --user hererocks esptool
 - hererocks env --lua 5.1 -rlatest
 - source env/bin/activate
 - luarocks install luafilesystem
 install:
-- tar -zxvf tools/esp-open-sdk.tar.gz
+- tar -Jxvf tools/esp-open-sdk.tar.xz
 - export PATH=$PATH:$PWD/esp-open-sdk/xtensa-lx106-elf/bin
 script:
 - lua tools/cross-lua.lua || exit 1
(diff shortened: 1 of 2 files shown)


TerryE:
Yup, Yuri's rebaselined the toolchain from V4.8.2 to 4.8.5 and dumped some of the never used elements.  The main saving though is moving from gzip to xz compression.  A good job. :smile: and :+1:

djphoenix:
Thanks guys. Let's merge.

PR #1756, opened 2017-01-23
TTL support in net module

### LINKED ISSUE #1455: Enable TimeToLive for net/UDP
8<------------------------ FEATURE REQUEST ------------------------------------
### Justification

TTL tells routers when to forward IP multicast packets.  A TTL of 0 or 1 restricts UDP to the same subnet; allowing a configurable TTL, or at least a value of 32 would support LUA apps running in on premise in more complicated networks.
Ref: [http://searchnetworking.techtarget.com/definition/time-to-live](url)
### Workarounds

Are there any workarounds you currently have in place because the feature is missing?
no

8<------------------------ END FEATURE REQUEST --------------------------------

djphoenix:
Fixes #1455.

Actually checked only that TTL values changes in Wireshark.
@mikep01 check that it works as you expect if you can.
Also tagging @luismfonseca as net internals expert :)

### DIFF
--- a/app/modules/net.c
+++ b/app/modules/net.c
@@ -638,6 +638,27 @@ int net_dns( lua_State *L ) {
   return 0;
 }
 
+// Lua: client/socket:ttl([ttl])
+int net_ttl( lua_State *L ) {
+  lnet_userdata *ud = net_get_udata(L);
+  if (!ud || ud->type == TYPE_TCP_SERVER)
+    return luaL_error(L, "invalid user data");
+  if (!ud->pcb)
+    return luaL_error(L, "socket is not open/bound yet");
+  int ttl = luaL_optinteger(L, 2, -1);
+  // Since `ttl` field is part of IP_PCB macro
+  // (which are at beginning of both udp_pcb/tcp_pcb)
+  // and PCBs declared as `union` there is safe to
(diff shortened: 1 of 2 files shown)


marcelstoer:
Ready for merge after the next master drop.

luismfonseca:
> Also tagging @luismfonseca as net internals expert :)

I'm not worthy of this title at all :o

It looks good, if anything I'd propose a different API. This `net` module has a total of `get*` functions, so I think that splitting the get to a separate function: `getttl` could be better. Although, now that I've just typed it, all those `t`s look a bit weird...

marcelstoer:
> It looks good, if anything I'd propose a different API. This `net` module has a total of `get*` functions

4 is not that many 😉 I'm fine with this function being both a setter and getter.

mikep01:
Just got 'round to this and I don't think I can test it.  The net UDP module i/f has changed with a relatively big impact to my lua code.  I'm going to have to stick with my old version rather than try to manage two separate code streams for my users (who already have the old firmware installed).  Sorry about that.

mikep01:
Updated my impl to use 2.1.0 (anticipating a fix for Krack), added and tested the TTL setting for my UDP connections.  It appears to be working well, the packets now flow across my network segments.

Issue #1752, opened 2017-01-21
Expose lwIP netif interface to Lua


wolfgangr:
8<------------------------ FEATURE REQUEST ------------------------------------
### Missing feature
Advanced/customized examination and modification of network configuration, using nodeMCU's lwIP network interface, from Lua land

### Justification
See the lwIP documentation about the functionalities of the netif module
http://lwip.wikia.com/wiki/Network_interfaces_management

I need it as a prerequisite for developping an ethernet driver ( see issue #1725).
I suppose it will also be required for configuring customized network configuration from lua land, once drivers for other PHY than the embeded WIFI are available.
It should also enhance the manipulation of routing and other advanced network setups with the onboard WIFI from Lua land, as far as provided form lwIP.

### Workarounds
Do it in C:
```
./app/include/lwip/netif.h
./app/include/lwip/netifapi.h
./app/include/netif/*.h
./app/lwip/api/netifapi.c
./app/lwip/core/netif.c
./app/lwip/netif/etharp.c
```

8<------------------------ END FEATURE REQUEST --------------------------------

Following the discussions of issue #1555 (exposing "ping"), I'd suggest to keep it in a file 
`app/modules/net_if.c` and call the lua functions `net_if.whatever()`.

Everybody happy with that proposal?

wolfgangr:
There are 2 layers of netif interfaces in lwip

* The direct interface, defined in `./app/include/lwip/netif.h`
* The API interface, definded in `./app/include/lwip/netifapi.h`

cmp [Adam Dunkel's Initial lwIP Paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.109.1795&rep=rep1&type=pdf) , Chapter 7 , Chapter 12.2

> The general design principle used is to let as much work as possible be done within the ap-
plication process rather than in the TCP/IP process. This is important since all processes use
the TCP/IP process for their TCP/IP communication. 

I think that we don't have "processes" or "threads" on the nodemcu, so we could use the "short way" and interfacing the TCP/IP direct, as available in `netif.h`.
However, given the above recommendation and the fact that the API interface is much simpler (it does not use callbacks), we start with the API interface in `netifapi.h`.

Accordingly, I'd suppose to call the function `net_if_api.whatever()` and the file `app/modules/net_if_api.c`
If we come across a functionality that does not work properly, we still can have the direct way in a seperate module `net_if.whatever()` in a file `app/modules/net_if.c`. 

[...]

PR #1757, opened 2017-01-23
Try merge ping to upstream

### LINKED ISSUE #1555: expose ping function to lua
### Missing feature

ping function
### Justification

ping function will help the developer to know if there is connectivity to web and search for other devices in local network
### Workarounds

I am currently using net module and doing post requests

wolfgangr:
Fixes #1555

- [x] I have thoroughly tested my contribution. (see below)

Exposing ping functionality for network diagnosis to lua land.

untested feature - I'v added a note inthe docs:
callback variant - allows you to register a lua function to be called upon each incoming ping, instead print on the serial console. 
See issue #1555 for reason

---

**API design**
- [x] Agreement on printing from C layer vs. callback -> no console output (except for debugging info)
- [ ] Update of `net_info.ping()` API accordingly?

**Documentation**
- [ ] Reflect updated API spec in doc file
- [ ] Align formatting of .md to existing files
- [ ] Other proposed corrections

**Implementation**
- [ ] Fix callbacks from lwip context. They shall not call out to Lua but need to post tasks.
- [ ] Remove caching of caller's lua state
- [ ] Fix memory leak of unbalanced zallocs
- [ ] Fix memory leak of Lua callback reference not unref'ed automatically
- [ ] Adapt parameter parsing to API doc
- [ ] Clarify uninitialized `ping_dns_lookup`


wolfgangr:
Hi Marcel,
Sorry for bothering you, many thanks for teaching me :-)))
Things getting better now?

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+    struct ping_resp *pingresp = (struct ping_resp*)data;
+
+    char ipaddrstr[16];
+    ip_addr_t source_ip;
+    
+    source_ip.addr = pingopt->ip;
+    ipaddr_ntoa_r(&source_ip, ipaddrstr, sizeof(ipaddrstr));
+
+    if (ping_callback_ref != LUA_NOREF) {
+      lua_rawgeti(gL, LUA_REGISTRYINDEX, ping_callback_ref);
```
devsaurus: Don't use the saved lua_state of `gL` (remove it completely) but obtain the main Lua state via
```c
```
Refer to this conclusion of a discussion we had in the past: https://github.com/nodemcu/nodemcu-firmware/issues/846#issuecomment-164651561

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+
+
+static void ping_by_hostname(const char *name, ip_addr_t *ipaddr, void *arg) {
+    struct ping_option *ping_opt = (struct ping_option *)c_zalloc(sizeof(struct ping_option));
+
+    if (ipaddr == NULL) {
+      c_printf("SEVERE problem resolving hostname - network and DNS accessible?\n");
```
devsaurus: This and below: `NODE_ERR()` instead of `c_printf()`?

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+  *                                           --     ping results
+  */
+
+static int net_info_ping(lua_State *L)
+{
+    const char *ping_target;
+    unsigned count = 4;
+
+    // retrieve address arg (mandatory)
+    if (lua_isstring(L, 1)) {
```
devsaurus: Since you're not checking for any other type of argument 1:
The whole if statement can be replaced by
```c
ping_target = luaL_checkstring(L, 1);
```

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+This module is a stub to collect common network diagnostic and analysis tools.
+
+##net_info.ping()
+ send ICMP ECHO_REQUEST to network hosts 
+ 
+**Synopsis:**:
+```
+net_info.ping(host [, count [, callback]])
```
devsaurus: Formally, your implementation is more like
```
net_info.ping(host [, [count] [, callback]])
```
I.e. `count` is optional even if a `callback` is given. Change the logic for processing the arguments if you want to conform to the Synopsis stated.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    // retrieve count arg (optional)
+    if (lua_isnumber(L, 2)) {
+	count = luaL_checkinteger(L, 2);
+    }
+   
+    // retrieve callback arg (optional)
+    if (ping_callback_ref != LUA_NOREF) 
+      luaL_unref(L, LUA_REGISTRYINDEX, ping_callback_ref);
```
devsaurus: Why unref a previous callback here? This means one has to call the ping function twice to free the callback of the first invocation.
I'd recommend to unref the callback directly after `lua_call()` in `ping_received()`.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+    ping_callback_ref = LUA_NOREF;
+       
+    if (lua_type(L, 3) == LUA_TFUNCTION || lua_type(L, 3) == LUA_TLIGHTFUNCTION)
+      ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```
devsaurus: `luaL_ref()` [pops the top element from the stack](http://pgl.yoyo.org/luai/i/luaL_ref) which could mess up the stack management at some point.
Recommended to generate a copy on the stack which is then popped by `luaL_ref()`:
```c
lua_pushvalue(L, 3);
ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+	       pingresp->seqno,
+	       pingresp->ttl,
+	       pingresp->resp_time);
+    }
+}
+
+
+
+static void ping_by_hostname(const char *name, ip_addr_t *ipaddr, void *arg) {
+    struct ping_option *ping_opt = (struct ping_option *)c_zalloc(sizeof(struct ping_option));
```
devsaurus: Here and for the `c_zalloc()` instance below:
Where is the corresponding `c_free()` for this memory chunk?

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+| :----- | :-------------------- | :---------- | :------ |
+| 2017-01-21 | [smcl](http://blog.mclemon.io/esp8266-contributing-to-the-nodemcu-ecosystem) | [wolfgangr](https://github.com/wolfgangr/nodemcu-firmware) | [net_info.c](../../../app/modules/net_info.c)
+
+
+This module is a stub to collect common network diagnostic and analysis tools.
+
+##net_info.ping()
+ send ICMP ECHO_REQUEST to network hosts 
+ 
+**Synopsis:**:
```
devsaurus: Please stick to the common layout for API docs: Use `#### Syntax` and also have a section for `#### Parameters` etc.

marcelstoer: You beat me to it 😉 
@wolfgangr An arbitrary example for what Arnim means can be seen here: http://nodemcu.readthedocs.io/en/latest/en/modules/net/#netserverlisten

app/modules/net_info.c
marcelstoer: ... and if you don't want to use `NODE_ERR` directly you can stick to  https://github.com/nodemcu/nodemcu-firmware/wiki/%5BDRAFT%5D-How-to-write-a-C-module#debug-and-error-messages

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
+/* 
+ * http://blog.mclemon.io/esp8266-contributing-to-the-nodemcu-ecosystem
+ * https://github.com/smcl/nodemcu-firmware/blob/cc04aaf92c1c076c30ef0b0eee43b3f924137440/app/modules/test.c
+ *
+ * extracted ping function
```
marcelstoer: How about "Source / inspiration for 'ping' function: <the-two-URLs>"? As this now a more general module we should expect more functions and thus more references in the future.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+ * adoptet jan 2017 for commit to nodeMCU by Wolfgang Rosner
```
marcelstoer: "2017-01 Adapted to NodeMCU by Wolfgang Rosner, https://github.com/wolfgangr/"

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
+# net_info Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2017-01-21 | [smcl](http://blog.mclemon.io/esp8266-contributing-to-the-nodemcu-ecosystem) | [wolfgangr](https://github.com/wolfgangr/nodemcu-firmware) | [net_info.c](../../../app/modules/net_info.c)
```
marcelstoer: Whether you want to put GitHub username "smcl" there or "Sean McLemon" is up to you. I personally see no reason to stick to the username - unless you want to link his GitHub account rather than his blog. Same applies to yourself of course.
We never formalized it anywhere but in this scenario Sean is the _origin_ and you are _contributor_ i.e. the one who eventually contributed code to NodeMCU.

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+```
+function ping_recv(bytes, ipaddr, seqno, ttl, ping)
+```
+* length of message
+* ip-address of pinged host
+* icmp_seq number
+* time-to-live-value
+* ping time in ms
+
+!!! Caution
```
marcelstoer: I wouldn't be happy to merge untested code/features, that just ain't us. Therefore, I'd like to see this whole disclaimer be removed - once @wolfgangr or someone else actually tested this feature.

docs/en/modules/net_info.md
wolfgangr: hmm....
as I statet, I have just portet code found somewhere else.
I get a rough grasp of your points, but can't follow them in detail.
Getting familiar with the details of C/LUA interfacing is still on my todo list.
But time is limited, and my priority is to have an ethernet driver running.
I'd suppose to postpone this issue until I have to learn the details of lua interfacing anyway.

Another option is to ask the original author of the code, @smcl, to take over.
As I see, he has just dropped into the discussin on topic #1758
Is there a formal way to get him on board at this stage of a pull request?

A third option would be to remove the callback functionality, since this requires most of the lua / stack interfacing. But this causes work, test and risk as well.
Personally, I'd prefer to fix things than to dispose them.

docs/en/modules/net_info.md
wolfgangr: I ust learned the full name of Sean McLemon when he dropped into issue #1758 .
I'd prefer to quote the full name.

>  Sean is the origin and you are contributor 

never mind.
Maybe even better than "Maintainer" - less obligations, more the one shot thing?

docs/en/modules/net_info.md
wolfgangr: Well, that comes back to the discussion above with @devsaurus .
YOU suggested skd of early PR, right? - So I tried to issue one.
How can anybody else test the code if it is not pbulic?
The ping is in a "works-for-me" state. My primary goal is the ethernet driver.
Yes, I've learned a lot on git and contributing the last days, many thanks for that.

But I don't like to run out of time without an ethernet driver working in the end, you understand?.

What about having a night sleep on this?
And maybe hoping for Sean McLemon jumping (back?) on the bandwagon?

app/modules/net_info.c
wolfgangr: Could you give me a hint on the reason, please?
Yes, I think generally it would be better to assemble a string an then only in the end decide what to do with it.
If only it were not that cumbersome in C .....

What about a mixed implementation?
C just giving the bare data, and lua doing the format?

But as I understood from your comment here
https://github.com/nodemcu/nodemcu-firmware/issues/1555#issuecomment-274259503
this is not so easily accomplished.

docs/en/modules/net_info.md
wolfgangr: so this boils down to use 
####Synopsis
instead of 
\*\*Synopsis:\*\*

OK, no problem.
I had tried with 3 hashes ###
but still got shown in the index and I didn't want to clobber that with details too deep.

@devsaurus 
> Please stick to the common layout for API docs

Is there a common layout for API docs defined somewhere (outside of your brain, of course)?
I really would have preferred to start from some accepted template, 
if I only could have found one...

app/modules/net_info.c
wolfgangr: Sorry, I still don't get the point.
Just searched the docu for NODE_ERR() and dbg_print(), but could not find anything.
Well, there is a dbg_print.c as a reverse engineering candidate?

Whats the difference, and also to c_printf?
Does NODE_ERR() cancel execution?
That might actually be a good idea in **some** situations.

I'm just crafting a netif interface full of c_printf .
As far as I understood, dbg_printf and the associated obscurity layers do not offer the formating power of printf, right?
Well, I could use sprintf instead, but does this make things better?

Or should we better clearly separate two situations:
- debugging / prototyping on the console, where occasional glitches do not hurt
- production code, which has to be perfectly stable

Both my ping and the netif part I'm working now is supposed to be a prototyping tool. 
I think in production, to write to the console does not make any sense at all, right?

I really would be glad to be able to print to some buffer using a printf / sprintf variant and decide at the end where to hand it to. But I'm afraid to do so I have to switch the language :-(

app/modules/net_info.c
pjsg: It isn't clear to me what context this is executing in. I suspect that it might as a direct callback from lwip. In that case, you mustn't reference any lua context. You have to use the task api to get the data queued down to the base level.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+    if (ping_callback_ref != LUA_NOREF) {
+      lua_rawgeti(gL, LUA_REGISTRYINDEX, ping_callback_ref);
+      lua_pushinteger(gL, pingresp->bytes);
+      lua_pushstring(gL, ipaddrstr);
+      lua_pushinteger(gL, pingresp->seqno);
+      lua_pushinteger(gL, pingresp->ttl);
+      lua_pushinteger(gL, pingresp->resp_time);
+      lua_call(gL, 5, 0);
+    } else {
+      c_printf("%d bytes from %s, icmp_seq=%d ttl=%d time=%dms\n",
```
pjsg: This feels like NODE_DBG -- it can happen normally.

app/modules/net_info.c
pjsg: Also, it would be better to throw a LUA error rather than just return silently.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+	return luaL_error(L, "no address specified");
+    }
+
+    // retrieve count arg (optional)
+    if (lua_isnumber(L, 2)) {
+	count = luaL_checkinteger(L, 2);
+    }
+   
+    // retrieve callback arg (optional)
+    if (ping_callback_ref != LUA_NOREF)
```
pjsg: Note that you can call unref on LUA_NOREF -- so just leave out the test.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+   
+    // retrieve callback arg (optional)
+    if (ping_callback_ref != LUA_NOREF) 
+      luaL_unref(L, LUA_REGISTRYINDEX, ping_callback_ref);
+    ping_callback_ref = LUA_NOREF;
+       
+    if (lua_type(L, 3) == LUA_TFUNCTION || lua_type(L, 3) == LUA_TLIGHTFUNCTION)
+      ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+    
+    gL = L;   // global L
```
pjsg: Never save L. You can use lua_getstate() to get an arbitrary lua_State blocks. Note that lua_States can go away at any time.

app/modules/net_info.c
```diff
@@ -0,0 +1,176 @@
...
+	ping_opt->coarse_time = 0;
+	ping_opt->recv_function = &ping_received;
+	
+	ping_start(ping_opt);
+	
+    } else {
+	ping_host_count = count;
+
+	struct espconn *ping_dns_lookup;
+	espconn_create(ping_dns_lookup);
```
pjsg: This doesn't make sense -- ping_dns_lookup is not initialized. Also the return code of espconn_create is not checked.

devyte:
Just some comments:
-I believe that use of espconn is discouraged. Yuri just did an awesome job
porting the net layer from the sdk and espconn to use lwip directly. It
would be great if that dnslookup code could be rewritten without it.
-Writing out to console does make sense, assuming that console means
stdout. There is a way to redirect stdout to a socket, which effectively
allows a remote Lua prompt over network.
What probably does not make sense is writing debug output in production.
-I'm probably too late to say this, but ping, traceroute, etc are network
utilities. In Linux (debian, ubuntu, et al) they are in the iputils
package. It would be more descriptive to call the module net_utils or
net_tools. Of course, like you said, the priority is to have an ethernet
interface, and compared to that, this is not really important.
-I find your work so far really great! Some time ago I was actually
thinking of having a network sentry, where it would periodically ping a
list of network devices, and send an email notification if any went down. I
was thinking of doing it in the main router (dd-wrt), but every time I
upgrade the firmware I'd have to set that up all over again. Your api is
perfect for this, it allows an ESP to be the sentry!

app/modules/net_info.c
wolfgangr: This the very reason-d'etre for this ping code: print received pings :-)

wolfgangr:
@devyte 
> espconn is discouraged.
> [...] use lwip directly

Glad to read that. So the question is when and who is rewriting the code.
Maybe, when I delve deeper into netif, the solution just falls onto my knees?

> Writing out to console does make sense, assuming that console means stdout. There is a way to redirect stdout to a socket, which effectively allows a remote Lua prompt over network.

Wasn't the concept of piping around with stdin / stdout / (and to a lesser extend stderr) the core of the success of 'nix & C ?  That's why I really hesitate to drop that in favour of a proprietary nodemcu supplement. 

So how do I do it? `c_printf` ? `dbg_printf` ? 
Is there a docu available how stdin/stdout/stderr map into the nodemcu environment and how to manipulate them? **How can I redirect stdout?**

> What probably does not make sense is writing debug output in production.

Do you want to say **stick to c_printf?** 

Maybe I should check for "normal" and "abnormal" conditions. From work in PERL (and others) I used to distinct
- normal operation -> stdout
- debug and mild errors -> stderr
- severe errors -> stderr and abandon execution (skd if "die") with cleanup and backtrace (usually provided by the environment, but redirectable)

Is there a clean mapping for this in the nodemcu environment?

I proved easy to redirect those e.g. to a web serving environment, and thus do interactive development over a browser by tow three-code-line wrappers. I did this with database presentation, as well as with a webmin extension.

>  It would be more descriptive to call the module **net_utils** or **net_tools.** 

I can **rename**, but if so, I'd like to do it now. But I think the flat namespace wihtout dependencies and call for high granularity makes it impossible to mimic the naming convention of linuxes. So rather than hunting an ever moving target, it may be preferrable to settle and then stick to it.

I think the flat independet module organisation may not live forever, so ther may be some larger reorganisation required in some distant future anyway. 

Do we have any **authority** to make a **binding decision**?

> -I find your work so far really great!

Thanks :-) You won't believe but this helps :-)))

marcelstoer:
> Do we have any authority to make a binding decision?

More less, yes...informally, though 😜 It's not like in a 1st class democracy but we're usually kind and forgiving with each other.
We're a handful of _collaborators_ (GitHub puts a "Collaborator" shield next to your name) who can approve and merge PRs. What I observed over the course of my participation is that if a collaborator says A and no other collaborator objects within a couple of days then it should be A.

docs/en/modules/net_info.md
marcelstoer: > The ping is in a "works-for-me" state

That's good and implies you _did_ test it, no?

docs/en/modules/net_info.md
marcelstoer: By contributing a new module you automatically become its 1st maintainer. With close to 50 modules we just don't have the capacity to maintain someone else's code w/o the commitment from the contributor.

wolfgangr:
@marcelstoer 
>  if a collaborator says A and no other collaborator objects within a couple of days then it should be A.

OK, so I just have to wait a few days, right :-)

wolfgangr:
regarding the "where to print" issue:

just scanning @marcelstoer pointer to the [How-to-write-a-C-module](https://github.com/nodemcu/nodemcu-firmware/wiki/%5BDRAFT%5D-How-to-write-a-C-module#c-module) from top to bottom and find this code snippet:

```
// debugging function (to return the current values of the module)
static int easygpio_info( lua_State* L )
{
    char buf[255];
    c_sprintf(buf, "pin clock:%d, pin output:%d, state:%d", PIN_CLK, PIN_DATA, STATE_INDICATOR);
    lua_pushfstring(L, buf);
    return 1;
}
```

This comes close to my goal of redirectable output.

The sad thing is that obviously I have to handle the tedious buffer allocation manually.
We all know that trick of just allocating buffers veeeerrrrryyyy ( [255] and evern more... ) large only provides false feeling of safetey and sacrifices valuable memory for nothing.

I was hoping to hand this off to the OS emulation. This is was OS are for the last 50 years or so.
Of course I know that nodemcu is not really an OS.

So does c_printf actually implement reliable buffer management, inclduding saving valuable memory (and most important) protection against overflow? If so, I'd really appreciate to continue with that and redirect output to some lua string at the end.

In my net_if #1752 I already encountered the need of hardly predictable buffer length.
I tried (and think I succeded) with malloc, but this  is dangerous to produce memory leaks.

app/modules/net_info.c
wolfgangr: @devsaurus , @marcelstoer 
> This and below: NODE_ERR() instead of c_printf()?

> . and if you don't want to use NODE_ERR directly you can stick to https://github.com/nodemcu/nodemcu-firmware/wiki/%5BDRAFT%5D-How-to-write-a-C-module#debug-and-error-messages

When I get this right, you want to have output dependent on some debug flag switched on.
However, I think that does not meet the expected typical use case of such network utility tools:

Somebody uses MASTER (presumably from build service) and has some network stuff enabled (both wireless and maybe once a day wired as well). He does **not switch any debug stuff on,** because he expect the **modules on nodemcu level working** without trouble. But he may still want to **debug his own local network** configuration - which is the standard use case of matured network utils like ping, traceroute or ifconfig.

Eg. having no hostname resolution because WIFI is down or DNS is not configured is not an issue for debugging nodemcu code, is it? Notifying the user that host resolution does not work is a perfect expectable behaviour of a ping tool - no need to debug the tool

Well, yes, the two tierd approach of "problem..." and "SEVERE problem..." is somewhat odd and presumably due to the fact that the template resides on the quirky deprecated `espconn` obscuration layer.
I just realized that there are cases (when wifi is down, e.g.) where we don't even have a ipaddr object, and there are cases where we have such, but the hostname could not be found.
Maybe, for the moment I rephrase it in "host name not found" and "name resolution not working"?

So we are talking about **different layers of debugging targets,** don't we?
So the **output of the net util tools has to go to stdout, not to stderr,** or it's conceptual counterparts in nodemcu and lua land.
Please correct me if I'm wrong.

pjsg:
I think that there are two different views of what this module is trying to acheive. I think that @wolfgangr wants something that can be run from the interactive serial prompt. I think that the other people (myself included) want something that can be integrated into an application. This latter case implies that the API is the most important aspect (because there is nobody around to look at printed output). 

I see the following use cases:

* An app that does regular connectivity tests and reports the results over mqtt to a monitoring station. The app wants to get individual round trip times and success / failure. It might use varying lengths of packets. It might do multiple pings to different machines at the same time. 

* An app that verifies that a system is reachable, and if it goes unreachable for a period of time, switches a relay off and then on to power-cycle the system. In this case, the app is interested in failure of individual messages (or maybe interested in failure of the `n` pings).

wolfgangr:
@pjsg 
I agree with your view of different use cases.
A full fledged net diag  tool should be able to be integrated in a lua land app.

However, as we see in the ping case at hand, this is a complex endeavour. 
Developing a command line tool in the fist instance is a first step to go, still a valuable ressource for SOME of the use cases, and a starting point for the integrateable tool.

From my current stage of work, I see four levels of integratability:
- printing to the command line (which may be redirected, as somebody mentioned above
- returning a string to lua on the stack, which may be printed or processed in any other way
- returning a data structure that resembles the underlying C struct for advanced postprocessing without the requirement to parse the string again
- registering a callback that can receive either the string or the data structure

Not any function may require all 4 levels, but some may. 
A callback is only (and always) required on blocking operations, as in the ping case.
But I think there should be some systematic approach to handle this quenstion in a coherent matter, so the fundamental discussion at hand is not to be repeated in every case.

And please allow stepwise implementation of the levels. 
Albeit I agree that the first two levels can possibly be handled in one turn.

docs/en/modules/net_info.md
wolfgangr: It means that for my needs I am happy with the console print, which is what I have tested.

I don't want to ping over web or mqtt or so, so there was no need for me to test the callback variant.
But I kept it there, because it was alredy in the example I took from Sean's tutorial. In case anybody had a need for - as it looks like.

I also see that  all the lua stack handling errors found by @devsaurus are not a big problem for one shot console calls, but of course they may be for permanent running apps.

app/modules/net_info.c
devsaurus: Hm, lwip callback context - I wasn't aware of such implications. Though I see this practice in `mqtt` as well. E.g. [`espconn_regist_recvcb(pesp_conn, mqtt_socket_received);`](https://github.com/nodemcu/nodemcu-firmware/blob/master/app/modules/mqtt.c#L546) while [`mqtt_socket_received()`](https://github.com/nodemcu/nodemcu-firmware/blob/master/app/modules/mqtt.c#L286) heavily interacts with Lua context.
Does it mean we have a general problem in the code base in case callbacks aren't fed through the task api?

app/modules/net_info.c
devsaurus: My point was to use NODE_ERR since I regard this as the default channel for emitting error messages. And yes, it can be masked by `user_config.h` if the user wants to skip hard errors. `c_printf()` prints unconditionally, that's all.

I'd even favor @pjsg's proposal to throw a Lua error instead, but this function is executed in some lwip-callback context. Guess the context issue has to be clarified first before.

docs/en/modules/net_info.md
devsaurus: > Is there a common layout for API docs

Guess we don't have a dedicated template, sorry. But you'll see a common pattern when looking at the other .md files.

docs/en/modules/net_info.md
devsaurus: I understand your headaches, really 😃 And don't take my review comments personally, your work is highly appreciated!
If ok from your side, I would resolve such cosmetic issues by adding commits to your branch myself.

devsaurus:
> Wasn't the concept of piping around with stdin / stdout / (and to a lesser extend stderr) the core of the success of 'nix & C ? That's why I really hesitate to drop that in favour of a proprietary nodemcu supplement.

I got the impression that the back and forth about printing info to stdout is just a symptom of an underlying mismatch. The concept of gluing function blocks together with pipes doesn't map very well to the firmware IMO. The firmware's counterpart is chaining functionality via callbacks. Terry's FAQ gives a good insight into the event-driven nature with [How is NodeMCU Lua different to standard Lua?](http://nodemcu.readthedocs.io/en/dev/en/lua-developer-faq/#how-is-nodemcu-lua-different-to-standard-lua).

With this in mind, my view on the four levels is:

> - printing to the command line (which may be redirected, as somebody mentioned above

This is purely for interactive use. No clear benefit for an application.

> - returning a string to lua on the stack, which may be printed or processed in any other way

Similar to level 1. An application would first need to extract the semantic information from the string.

> - returning a data structure that resembles the underlying C struct for advanced postprocessing without the requirement to parse the string again

This maps to returning a Lua table (applicable to complex / structured information which is returned by a single function call).

> - registering a callback that can receive either the string or the data structure

Equivalent to level 3. Just for asynchronous / event-driven cases as you mentioned.

> But I think there should be some systematic approach to handle this quenstion in a coherent matter

In the end I would clearly prefer levels 3 & 4 over 1 & 2. Pretty printing the information is just a special use case that can be built by encapsulating the function call (for level 3) or printing from the callback (level 4):

```lua
function list_addr()
  local ifcs = net_if.list_addr();
  for ifc, addr in pairs(ifcs) do
    print(string.format("%s IP: %08X netmask: %08X gw: %08X", ifc, addr.ip, addr.mask, addr.gw))
  end
end

function ping(ip)
  net_info.ping(ip, 4, function(bytes, ipaddr, seqno, ttl, t)
    print(string.format("%d bytes from %s, icmp_seq=%d, ttl=%d, time=%d", bytes, ipaddr, seqno, ttl, t))
  end)
end
```

Adding the wrappers to `init.lua` provides transparent support for interactive sessions 😉

Opinions?

marcelstoer:
Arnim, I think your analysis is spot on. I fully agree.

wolfgangr:
@devsaurus 
> printing info to stdout is just a symptom of an underlying mismatch. The concept of gluing function blocks together with pipes doesn't map very well to the firmware IMO. 

AAHHHHRGHHHH - and thousand thanks!
I think I got the point. 
May I rephrase it in my own words, so that we can check that I've got it?

Relying on **pipes** implies relying on the services of a real **preemptive multitasking OS**.
In **nodemcu land** we do not have such, and I think for time critical apps **we do not want such**, either. 
That's why I'm here at all. 
Raspi zero is announced at 5$ / piece and has a full fledged debian. 
But you can never **rely on stuff happening in the microsecond** domain.

So the cost of going away from a preemptive MT OS is to **sacrifice the pipe** concept in **favour** of **the event** driven execution model.

Having grasped that, I understand that keeping things **simple** and systematical means to implement **only** the **event driven / lua structure at the interface layer.**

So starting at print to "stdout" when developping code as I did it here
https://github.com/nodemcu/nodemcu-firmware/issues/1752#issuecomment-274885476
is as symptom that **my mind hasn't yet done the switch** required to the event driven concept.

Thanks for your patience in teaching me, lads :-)

wolfgangr:
sorry, **BUT** .... 
> Adding the wrappers to init.lua provides transparent support for interactive sessions 

OK. got hat. now let's consider **how to do that.**
Albeit referring to #1752 I think I continue here, since it's a conceptual question referring to ping as well and the logical continuation of the above.

Its not only the console session. People may forward pretty printed stuff on http or mqtt or whatever, and its not only pretty printing.

The netif structure is a little bit deep. Parsing stuff like MAC and IP from hex to human readable pretty print,  and parsing the 8 flag bits to their mnemnonic couterparts requires reading the manual or the .h file, or maybe call c functions already implemented.  When we go to manipulate things, we will want to implement advanced error checking.  

So if such interfacing tasks are implemented in lua land as @devsaurus proposes and I agree now, this is a little bit more than parsing a flat table out of the sleeves. So we will end up with quite **elaborated** (and properly **tested**!) code - now in **lua land** - that I think is worth to keep **in sync with** the (supposedly dynamic) development of the **C interface**. 

I just think of CPAN - a huge and ever growing PERL library pool for (among others) such tasks. I think in Python you can see the same. But they rely on a long matured standardized stable terface of e.g. Berkley / Unix network stack. So it is OK to have the higher level apps separated.

This is not what we have in the **nodemcu ecosystem**. Things are **still on the move**, as I see. Changing the bottom (eg. disposing espconnect in favour of lwip) will change the interface which will require changes in the accompanying lua code.

So I think the thin **flat layer in C shoult reside close the accompanied lua code on top of it.** to provide for **coherent maintenance**.
Do we want to have it in a **single module?** 
Remeber I had already raised that question but was bogged down somehow.
( please take this as an excuse for falling back to *nix thinking ;- ) )

Or is it ok or even desirable to have it **distinct modules?**
Hm... thinking in terms of managebility from the firmware point of view, and of **scarce memory** I 'd have it **out of the firmware,** actually.
And with this in mind, many of the above comments I did not understand yet make sense now.

On the other hand, **ROM is not that scarce** than RAM, so maybe we want it in the **firmware nonetheless?** Leave it to the **users choice?**

So this begs the question
Is there a place where such **lua code tailorized to meet an module interface** is supposed to be maintained?

Having it loosely strayed in tutorials, private git accounts, comments, wikis, blogs etc. is not the desireable situation of **coherent management.** I recall many situations where such code failed to work after os upgrade. This severly hurts the **usability of a platform** in anything **else than toy** applications.

A proposal for a workaround might be to **point in the doc file** to the **place where accompaning code lives.**
We even might simply put it as example **"code" into the docs?** 
Not nice, but workaround, and I think in many cases it is done this way already.
Or is there even a proper place I just haven not found yet?
`lua_examples` ? `lua_modules` ?
Do they go into the firmware? 
How is the proper way to point at them?

If we have code in `lua_modules` , the user will be able to include this in ROM in the **build service,** once this meets MASTER, right? But we **still can point** in the docs **to to the source** in the web, so **users may decide** for this in preference, eg. to save rom space, to avoid additional build/flash cycles, for test, for modification or whatever. 

So the question remains **how to keep this pointers** and the **web ressources** pointed to **in sync** with the **module version actually running**.

Is there a way already etablished?
Is there a simple way to do it, e.g. by versionized links and repositories?
If not, is there enough reqirement/pain to find a solution anyway?

app/modules/net_info.c
pjsg: We *may* have such a problem, or we may not. It is not always obvious when you need to do this. I tried to put together a debugging patch to detect this problem, but it got too complex. 

Timer callbacks leave you in the right context. I'm not sure about espconn callbacks -- though I *suspect* that they are ok. The lwip callbacks are *not* ok.

devsaurus:
You're bringing up an interesting topic which deserves a separate issue from my point of view. It might get lost here in this thread where focus should be on getting the PR flying.

devsaurus:
To quote myself

> focus should be on getting the PR flying

This thread has collected many loose ends so I tried to summarize the topics in the original post to have something like a list to check-off towards merge (for myself and maybe other reviewers). It's not necessarily in a specific order. Feel free to adapt or extend.

First step I see is revisiting the API definition.
My proposal as a consequence of the print/callback discussion: should the `net_info.pin()` API be updated since the callback seems to be mandatory?
```lua
net_info.ping(host, callback[, count])
```

Independent of that, the lwip context issue together with `gL` can be addressed in the code. @wolfgangr I offer to commit these changes to your PR branch if you agree.

wolfgangr:
@devsaurus 
> focus should be on getting the PR flying

As you probably know, this is not my own code, but a template I have imported from Sean McLemon's tutorial.
My first goal was to learn, which implies that for this night I get tired at chapter 23 of PIL - right before the relevant C interfacing.

> I offer to commit these changes to your PR branch

I understand and partially share your impatience. I think either I fix it completeley on my own, but then it will still take some days (with increasing risk that live threatens to come into way....)
Or you fix it completely, and I focus on the netif part, trying to apply what I've learned. 
You don't do me a big favour by offering me a moving target.
Since there seems to be some demand for the ping, I would not mind handing it off.

However, when I do it myself, I 'd consider to rewrite, flatten down and modularize it.
Some thoghts:

- replace the espconn host resolution by lwip DNS
- do host resolution as an independent step, separate from the ping itself, so we can get rid of the branching into two different ping-callbacks
- fix the stack handling. Maybe, the reason I did not understand it was not only due to my ignorance?
- take out all console printing and offer either a string representation as part of the final result or the __tostring mechanism as described in [pil](http://www.lua.org/pil/13.3.html) - which implies a redefinition of the API of course.
- move as much of the funcionality from C to lua, may be in form of a usage example in the doc, since I can't find a better place at the moment.
- fix the minor bugs/change requests one by one

I even consider to carry your thoght to the extreme and try to dispose of the C interface layer completely - see issue #1764 . However, this would really better be tried with a small start, as the current stage of my net_if resembles.

devsaurus:
> I think either I fix it completeley on my own, but then it will still take some days

That's fine for me.

wolfgangr:
OK, I've completed most of the reading assigned to me.
Most of your requests are now +- obvious to me, but one thing still puzzles me:

How the **get a Lua state  in the callback** functions?

I see that the existent static approach is not threadsave / not reentrant, so this is a no-go anyway.

I've learned that the stack might been reallocated during the time hostnames gets resolved and pings cycled around. So handing the lua state as an additional parameter to the callback is not option either, correct?
There is a `lua_getstate()` , intended to help out in some situations, but when I understand @pjsg correct
https://github.com/nodemcu/nodemcu-firmware/pull/1757#discussion_r97461970 I am **not** allowed to do so **in callbacks that return form lwip**. I suspect hat something like lua coexistence in nodemcu never was part of lwIP's desing specs, so we have to expect any lua stuff screwed up after executing lwip, OK?

To get out of this quagmire, we have to **place** a new **task** to the queue to get some kind of fresh lua context initialisation, based on which we may deliver our lwIP results to lua land.

So **any callback has to be split into two:**
**lwip** calls `task_post ...`  and this registers the **final callback** with the result to **lua**.
In our ping case at hand, we have to do this both for the hostname resolution `ping_by_hostname` (which I'd like to port to lwip) and for the `ping_received`.

So we end up in a **total of 4 callbacks, correct?**

Is there any tutorial or template around?

Is there a recommendation for the priority? I'd go for `TASK_PRIORITY_MEDIUM`. Users may wait for it, so its more important than some background housekeeping. But if there is more important (e.g. hardware related) stuff to do, pings can wait. Is this reasonable? Or should it even be user configurable?

wolfgangr:
RTFM: [Extension Developer FAQ](https://nodemcu.readthedocs.io/en/dev/en/extn-developer-faq/)

> Any **task handlers** are declared, typically in the **module_init** function by assigning task_get_id(some_task_**callback**) to a (typically globally) accessible **handle** variable, say XXX_callback_handle. This can then be used in an ISR or normal LLT to **execute a task_post**_YYY(XXX_callback_handle,param) where YYY is one of low, medium, high. The callback will then be executed when the SDK delivers the task.

So when I get this right, its a two step process:
First I have to get some Id from the handle, and when I finally want to trigger the callback, I have to use this ID, right? So I cannot directly register the task handler in the lwip call?

So when to do this with ping?
Do we really have to do some kind of **init** to be able just **to ping???**
And when we have done so, can we have **concurrent pings?** ist the code **reentrant** / **thread save** then?

Or can we initiate an Id **every single ping call**? But doesn't his add some extra layer of **ressources** which may **leak** memory and whatelse and are supposed to be **relased** again?

Grepping the build tree for an implementation example, I had a look at the rotary implementation and got really abhorred. Such ean effort for a simple ping? And this is **just the beginning of lwip** interfacing.

**Please tell me that I'm wrong.**
If not, I'll immediately trash all my nodemcu modules or sell them for a bargain :-(((

devyte:
The nodemcu code is not threaded, it's event driven. So thread safety is
not an issue.

wolfgangr:
... just finished gasping for breath ...

I see there is a NULL field as last argument of the `NODEMCU_MODULE` directive at the end of the module code. When I remeber right, this can hold not only test code, but also **module init** stuff, correct? 
So this might be the proper place where `task_get_id` might go, right?
Registering it only once means no leaks, no ressource management.

Are there any restrictions on the param field in the` task_post_YYYY` call?
Following the definition chain, I end up with 
`sdk/esp_iot_sdk_v2.0.0/include/ets_sys.h:32:typedef uint32_t ETSParam;`
so it is basically an arbitrary pointer.

So I could build a struct with two pointers in it:
-  to the the real callback
- to its arbitrary list of arguments
and hand this struct as param to ` task_post_YYYY`. 

This way, I can share one "generic task callback handler" for several callback functions in my module, and since several calls differ by their argument, I still have reentrant / thread save code.

Are you with me?

wolfgangr:
I just realize that there is no immediate reference to lua state in the `ping_by_hostname` callback.
So this one I still could use in its bare form, without envelopping it into a task?

But there are `c_printf` calls for the case that either the hostame resolution machine does not work or the hostname was not found. 
You asked me to replace these console prints by **NODE_ERR** macros - OK so far.

But are those **safe** for use in **lwIP callbacks**? 
Or do I have to wrap them into tasks as well?

wolfgangr:
@devyte 
> The nodemcu code is not threaded, it's event driven. So thread safety is not an issue.

well, but it still may be that concurrent pings may interleave.
Just think of somebody including ping in a web page, as available on many routers.
When two different clients call the ping page so that they temporarely overlap, ping functions may be called in a mixed manner by both.

Doesn't the event driven functionality implement some kind of threading, even if it maybe called slightly different by theoreticalists?

devyte:
No threading. When a request comes in, it is queued. This is the reason why
control must be returned asap, otherwise the accumulated requests in the
wifi stack could blow up the memory.

There are hardware interrupt routines which could interrupt normal code
execution, but they are very limited in what they do, and lots of care has
gone into making them as minimalistic as possible.

wolfgangr:
@devsaurus 
> My main concerns are the potential memory leaks (callback ref & unbalanced c_zalloc()s) 

**Yes, it does leak**, as node.heap() tells.
I think the callback were the easy thing, as you proposed the solution already.

The tricky thing is the `z_alloc` fo the `ping_option` structure.
I don't find a place in `net_info.c` where I could put the balancing free().
But when I put it **where logically it belongs**, it probably will **break backward compatibility**:

`net_info.c` relies on `app/lwip/app/ping.c` where we call into `ping_start`.
Look to some example output to explain what happens:
```
32 bytes from 192.168.1.12, icmp_seq=42 ttl=64 time=3ms
32 bytes from 192.168.1.12, icmp_seq=43 ttl=64 time=1ms
32 bytes from 192.168.1.12, icmp_seq=44 ttl=64 time=2ms
ping 4, timeout 0, total payload 128 bytes, 3965 ms
```
There is a timer routine `app/lwip/app/ping.c::ping_coarse_tmr` which maintains a counter. 
As long as this counter is positive, ping send/recv sequences are spawned off which end up in our callbacks, which print the repetitive lines (or call lua callback, depending on parameters).

But we must not free the `ping_option` struct there, because `ping_coarse_tmr` is still **running until** the last round where the counter is 0. There it prints the **last summary line.**
After that I think would be the right place to free `ping_option` struct.
But we don' get a callback after that any more.
So we had to put the free right into `app/lwip/app/ping.c` 

But wouldn't this risk to break backward compatibility?
I don't know who is relying on the `app/lwip/app/ping.c` code.
According to its comments, it is espressif code, not part of the lwIP core. 
And it lives in the same directory than espconn.c does.
So, ist it deprecated anyway?

Inserting a free into this code is a **policy decision** I do not dare to make as a newbie-maybe-contributor.
Rewriting the whole thing from scratch is beyond my time budget, I'm afraid.

My proposal were to put ping_opt as a static. I know this breaks reentrancy, but I think this is OK for most of the use cases.

devsaurus:
> My proposal were to put ping_opt as a static.

Yes, that would solve the memory issue for sure.

> I know this breaks reentrancy,

My feeling is that the whole ping stack is not reentrant even in the current state.
At the moment you can issue multiple pings, but they'll interfere with each other since they share global variable for function callback reference. A static ping_opt will "just" be another shared resource.

> After that I think would be the right place to free ping_option struct.
> But we don' get a callback after that any more.

What about the call to `ping_opt->sent_function()` at the end of `ping_coarse_tmr`? It looks like a "Last order, please" to me. Could net_if hook in there and free dynamic memory together with unref'ing Lua callbacks? lwip/ping is quite new to me, never looked into that. So this proposal might be nonsense.

wolfgangr:
@devsaurus 
> What about the call to `ping_opt->sent_function()` at the end of` ping_coarse_tmr`?

I think this is acutally a good idea. This line is not executed before any of the pings sent are either received or timed out, I suppose. In contrast to `ping_received`, which is too early when there are still further pings on the way.
I'll give it a try.

> My feeling is that the whole ping stack is not reentrant even in the current state.

I think it is reentrant for pings to numeric IP (i.e. without hostname resolution)  and without lua callback.

The **static** `ping_callback_ref` breaks reentrancy for the lua callback variant.
I could add another field to ping_option struct (maybe call  it "userdata") and thus keep the ` ping_callback_ref` seperate for each event sequence (I'm not supposed to call 'em threads, well). This should not break backward compatibility. So let's try it.

The **static** `ping_host_count` and the **static** `ping_host_ip` break reentrancy for pings by hostname.
Reason: The espconn host resolution does not provide a possibility to carry along userdata to its callback. 

I considered to drop the espconn in favour of the lwIP hostname lookup. This would allow for arbitrary arguments passed along.
However, there is this [warning](https://github.com/nodemcu/nodemcu-firmware/pull/1757#discussion_r97775498) raised by @pjsg :
>  I'm not sure about espconn callbacks -- though I suspect that they are ok. The lwip callbacks are not ok.

I could not track all the reason behind this, but I may assume that lwip callbacks can be generated out of network interrupts, so all lua context ist lost. Therefore, to keep the whole thing simple and to avoid the additional complexity of the task API, I'd stick with the espconn host lookup, since anyway the ping.c is obviously espconn, too.

This leaves us with code not reentrant only during the hostname lookup, which I think is typically much shorter than the following repetitive ping sequence. I'll place a note in the docs. In the rare case where this might cause a problem, the user can avoid it by resolving the hostname in a distinct step and call ping by IP afterwards.

devsaurus:
> I could add another field to ping_option struct (maybe call it "userdata")

What about `ping_option->reverse`?

> since anyway the ping.c is obviously espconn, too.

Do you mean `app/lwip/app/ping.c`? That would be lwip IMO.

BTW lwip also provides DNS lookup: `dns_gethostbyname()`. Don't know what else is required to get this working though.

wolfgangr:
@devsaurus
> What about ping_option->reverse?

As far as I can figure out "ping reverse" refers to something different.
I think misusing a field in other peoles code is not good programming style, is it?

>>    since anyway the ping.c is obviously espconn, too.

> Do you mean app/lwip/app/ping.c? That would be lwip IMO.

Sure?
from the [lwip-users](https://lists.nongnu.org/archive/html/lwip-users/2011-04/msg00089.html) mailing lsit:

> lwIP is the network protocol stack, ping is an application that uses the stack.  
lwIP supports ICMP (which the ping application uses) but it doesn't include the ping application internally, in the same way that it doesn't include a web server.

And the ping.c carries espcon copyright notice, in contrast to any other lwip files.

> BTW lwip also provides DNS lookup: dns_gethostbyname(). Don't know what else is required to get this working though.

Following @pjsg , we neet the task api. 'll give it a try. 

Next thing I need are some hours of leisure to do it...

devsaurus:
> As far as I can figure out "ping reverse" refers to something different

What exactly?
A `grep -r reverse` on `app/lwip` and `app/include/lwip` reveals no use of this field except the service to copy it when duplicating a struct espconn.

Refer to http://bbs.espressif.com/viewtopic.php?t=702 for intended use cases and to `app/modules/mqtt.c` for a real-life example.

In this repsect: `espconn.reverse` should be the place to store userdata for reentrant DNS lookup.

wolfgangr:
> >    As far as I can figure out "ping reverse" refers to something different

> What exactly?

Well, I remember that some RFC provides to call some remote host to ping to "me" (i.e the reverse way on the network), but I could not verify that by a quick google search. 
However, many people seem to refer to "reverse ping" as ping by IP + reverse host lookup.

> Refer to http://bbs.espressif.com/viewtopic.php?t=702 for intended use cases

Interesting :-). Thanks! This renders said question irrelevant ....
So I can rely on safely putting anything here? 
This begs the quest to leave `ping.c` completely untouched. That's waht interfacing to a library is supposed to be, isn't it?

wolfgangr:
@pjsg and all - can you help with the task api?
What's supposed to go into the `param` field in the `task_post` calls?
Where can I find docu / hints / pointers?

carved some hours to set up a task boilerplate following task.h and
 http://nodemcu.readthedocs.io/en/master/en/extn-developer-faq/

But instead of a safe **lua state** I get a **screwed** one resulting in lua **panic** and reboot

Current code state is here 
https://github.com/wolfgangr/nodemcu-firmware/blob/ping-debug-001/app/modules/net_info.c

Details and debug output I postet to the forum.
http://www.esp8266.com/viewtopic.php?f=6&t=13638

I think my usage of the param filed is not correct, but how do I have to do it correctly?

wolfgangr:
@pjsg 13 days ago Collaborator
> Also, it would be better to throw a LUA error rather than just return silently.

@devsaurus 12 days ago Collaborator

> I'd even favor @pjsg's proposal to throw a Lua error instead, but this function is executed in some lwip-callback context. Guess the context issue has to be clarified first before.

OK Gentlemen, I gave it serious try.
But a simple `nodemcu-firmware$ grep -nr "lua_error"` reveals that nobody tried this before (or did not succed it up to a successful PR). Sorry, but that's not a fair task for a beginner, is it?

wolfgangr:
Current interpretations of my test results for the protocol:

- `param` in` task_post `does behave as a transparent user pointer as expected
- allocation and scope of string literals is obviously implemented in a non standard way (details see [forum post](http://www.esp8266.com/viewtopic.php?f=6&t=13638) )
- having settled the string param issue, I see that the panic message is labeled with my own dummy error string:

```
 ### DEBUG: net_info ### :     ... pushed my msg to the stack...
PANIC: unprotected error in call to Lua API (I am the stupid hello task dummy)
```
Obviously, there is no real reason to panic, since then it would carry the real reasons error message.
Instead it looks like my **call to lua_error is mapped to panic**.

This begs the question whether the **lua state's environment** is to some extend **lost/modified** after cycling through the **task api**. 

Anybody experienced that already?
How to debug?
Where could I store / retrieve it?

wolfgangr:
Current interpretations of my test results for the protocol:

- `param` in` task_post `does behave as a transparent user pointer as expected
- allocation and scope of string literals is obviously implemented in a non standard way (details see [forum post](http://www.esp8266.com/viewtopic.php?f=6&t=13638) )
- having settled the string param issue, I see that the panic message is labeled with my own dummy error string:

```
 ### DEBUG: net_info ### :     ... pushed my msg to the stack...
PANIC: unprotected error in call to Lua API (I am the stupid hello task dummy)
```
Obviously, there is no real reason to panic, since then it would carry the real reasons error message.
Instead it looks like my **call to lua_error is mapped to panic**.

This begs the question whether the **lua state's environment** is to some extend **lost/modified** after cycling through the **task api**. 

Anybody experienced that already?
How to debug?
Where could I store / retrieve it?

wolfgangr:
I raised a separate issue #1789 for the panic problem, where I provide analysis and a proposal for a decision.

I'd go to split out the name resolution from the ping command, using the following api:
`net_info.host_to_ip("my.host.net", errorfunc)`

Ping then get's the API
`net_info.ping("W.X.Y.Z", count, trip-callback, summary-callback, error-callback)`
whith any malformed IP causing an error.

Following Terry's advice in [nodeMCU Unofficial FAQ](http://www.esp8266.com/wiki/doku.php?id=nodemcu-unofficial-faq), I'll drop my idea to produce a metatable with a __tostring mechanism to save spare memory.
Instead I'll do a simple print when the accompanying callback is missing. 

I'll configure it as #define directive, but I think it is not wise to suppress it by default in master, as it happens when I use NODE_ERR.

wolfgangr:
What about ICACHE_FLASH_ATTR ?
I think we should use it by default as long as there are no reason against it.
Even if it still does not restore the "normal" behaviour of literal string visibility.

wolfgangr:
I'll continue to drop my thoughts' eveolution here until I think the code reorganisation is settled.
Would it be better for the workflow and the readability to close the PR then and start a new one?
Or do you prefer to keep everything in one single albeit clobbered line?

marcelstoer:
> Would it be better for the workflow and the readability to close the PR then and start a new one?

I was just gonna propose that, glad you don't mind.

PR #1761, opened 2017-01-26
Feature/wifi clear config

### LINKED ISSUE #1706: Request: Clear Saved Wi-Fi config
### Justification
In an older version of NodeMCU, I would just do something like ```wifi.sta.config("","")``` and that would overwrite the saved config in flash. My understanding was that the SDK would not try to connect if the SSID was empty.

Changes to the wifi module now enforces a required SSID of at least one character:
https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/wifi.c#L676

### Workarounds
Set an invalid SSID, but SDK will try to connect to it: ```wifi.sta.config(" ", "")```. 

*Note: This also has a side effect of the SoftAP radio channel (since there's only one shared radio) changing constantly as the ESP tries to connect to a SSID that doesn't exist. When in STA+SoftAP mode, that is.*

jfollas:
Fixes #1706 .

Added `wifi.sta.clearconfig()` to remove the saved Wi-Fi configuration from flash.  Useful in a couple of scenarios:
- Factory Reset: To clear the saved configuration as a last step before giving the device to someone else
- End User Setup: If the currently configured network is accessible, then EUS will stop immediately after starting. If the currently configured network is not available, then the SoftAP will not lock onto a radio channel since the Station will be trying to find the previously configured AP (if Auto Connect is enabled, that is). A solution to both is to clear the config before starting EUS.

Also included: Eliminated the minimum password length requirement for station passwords. It was 8 characters, but this prevented WEP from being used with 5-character keys (ASCII representation of a 40-bit key). See the discussion in issue #1706.

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -110,16 +110,16 @@ static void wifi_scan_done(void *arg, STATUS status)
       }
       if(getap_output_format==1) //use new format(BSSID : SSID, RSSI, Authmode, Channel)
       {
-    	  c_sprintf(temp,MACSTR, MAC2STR(bss_link->bssid));
-    	  wifi_add_sprintf_field(L, temp, "%s,%d,%d,%d",
-    			ssid, bss_link->rssi, bss_link->authmode, bss_link->channel);
+        c_sprintf(temp,MACSTR, MAC2STR(bss_link->bssid));
+        wifi_add_sprintf_field(L, temp, "%s,%d,%d,%d",
+          ssid, bss_link->rssi, bss_link->authmode, bss_link->channel);
         NODE_DBG(MACSTR" : %s\n",MAC2STR(bss_link->bssid) , temp);//00 00 00 00 00 00
       }
       else //use old format(SSID : Authmode, RSSI, BSSID, Channel)
       {
-  	    wifi_add_sprintf_field(L, ssid, "%d,%d,"MACSTR",%d",
-  			  bss_link->authmode, bss_link->rssi, MAC2STR(bss_link->bssid),bss_link->channel);
-  	    NODE_DBG("%s : %s\n", ssid, temp);
+        wifi_add_sprintf_field(L, ssid, "%d,%d,"MACSTR",%d",
+          bss_link->authmode, bss_link->rssi, MAC2STR(bss_link->bssid),bss_link->channel);
+        NODE_DBG("%s : %s\n", ssid, temp);
       }
 
       bss_link = bss_link->next.stqe_next;
@@ -309,14 +309,14 @@ static int wifi_sleep(lua_State* L)
   sint8 wifi_fpm_do_sleep_return_value = 1;
   if(lua_isnumber(L, 1))
   {
-	  if(luaL_checknumber(L, 1) == 0)
-	  {
-		  desired_sleep_state = 0;
-	  }
-	  else if(luaL_checknumber(L, 1) == 1)
-	  {
-		  desired_sleep_state = 1;
-	  }
+    if(luaL_checknumber(L, 1) == 0)
+    {
+      desired_sleep_state = 0;
+    }
+    else if(luaL_checknumber(L, 1) == 1)
+    {
+      desired_sleep_state = 1;
+    }
   }
   if (!FLAG_wifi_force_sleep_enabled && desired_sleep_state == 1)
   {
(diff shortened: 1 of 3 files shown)


docs/en/modules/wifi.md
```diff
@@ -270,6 +270,24 @@ wifi.sta.changeap(4)
 - [`wifi.sta.getapinfo()`](#wifistagetapinfo)
 - [`wifi.sta.getapindex()`](#wifistagetapindex)
 
+## wifi.sta.clearconfig()
+
+Clears the currently saved WiFi station configuration. This will erase the saved configuration from the flash (useful for 
+factory-reset scenarios, or to prepare for [End-User Setup](enduser-setup) so that the SoftAP can lock onto a single channel)
```
dnc40085: >useful for factory-reset scenarios

The function `node.restore()` should be used to ensure complete erasure of WiFi configuration, since it completely overwrites the system parameter sectors that store WiFi configuration and other configuration information.

app/modules/wifi.c
```diff
@@ -506,8 +506,8 @@ static int wifi_station_get_ap_info4lua( lua_State* L )
     c_sprintf(debug_temp, " %-6d %-32s ", i, temp);
 #endif
 
-      memset(temp, 0, sizeof(temp));
-    if(strlen(config[i].password) >= 8)
+    memset(temp, 0, sizeof(temp));
+    if(strlen(config[i].password) >= 0) /* WPA = min 8, WEP = min 5 ASCII characters for a 40-bit key */
```
luismfonseca: If it's `5` for the 40-bit WEP, then shouldn't `5` be the new minimum?

jfollas: See #1706. tl;dr: 5 seems to be the smallest that would ever be used, but the SDK doesn't enforce a minimum.

docs/en/modules/wifi.md
jfollas: This whole time, I never knew about `node.restore()`.

> Reset default settings of following APIs: `wifi_station_set_auto_connect`, `wifi_set_phy_mode`, `wifi_softap_set_config` related, `wifi_station_set_config` related, `wifi_set_opmode`, and APs’ information recorded by `#define	AP_CACHE`.

app/modules/wifi.c
pjsg: I suspect that this ought to be >= 1.  I can't think of any way that strlen could return < 0.

app/modules/wifi.c
```diff
@@ -506,8 +506,8 @@ static int wifi_station_get_ap_info4lua( lua_State* L )
     c_sprintf(debug_temp, " %-6d %-32s ", i, temp);
 #endif
 
-      memset(temp, 0, sizeof(temp));
-    if(strlen(config[i].password) >= 8)
+    memset(temp, 0, sizeof(temp));
```
pjsg: I also wonder -- is the size of the ssid field the same as the password field? If not, then trouble is nearby. Would prefer to eliminate the temp variable and just do a pushlstring of the correct length.

app/modules/wifi.c
```diff
@@ -652,6 +652,30 @@ static int wifi_station_getconfig_default(lua_State *L)
...
+
+  memset(sta_conf.ssid, 0, sizeof(sta_conf.ssid));
+  memset(sta_conf.password, 0, sizeof(sta_conf.password));
+  memset(sta_conf.bssid, 0, sizeof(sta_conf.bssid));
+  sta_conf.bssid_set=0;
+
+  wifi_station_disconnect();
+
+  bool config_success;
+  if(save_to_flash) config_success = wifi_station_set_config(&sta_conf);
```
pjsg: please format so that else lines up with if. The clauses should be on lines by themselves. I would prefer to use { and } in each case.

jfollas: That was existing code - not changed for this PR. I'm all for applying the Boy Scout Rule, but think it makes it harder for reviewers to see what exactly is new.

app/modules/wifi.c
jfollas: Looking at the SDK docs, ssid=32, password=64. Looks like they're defining temp's length here as the max password length, and zeroing out before each write (so it is big enough to hold null-terminated strings for both password and ssid).

app/modules/wifi.c
jfollas: Yeah, strlen will never be <0. Looking at this now after having cooled down from when I made the change, I think >0 is what is appropriate for this code that is copying the passwords only if they exist. The question becomes, should "pwd" be included the Lua table even if there is no password for the entry?

app/modules/wifi.c
pjsg: Yeah -- it looks OK.

app/modules/wifi.c
pjsg: I'd leave out the pwd field if there is no password.

PR #1771, opened 2017-01-31
Add section about NodeMCU PyFlasher to flashing docs

marcelstoer:

I hesitate to add a section about one of _my_ tools. However, I do feel that it really adds value, I see that the new flasher is becoming popular.

The decision yours of course.

### DIFF
--- a/docs/en/flash.md
+++ b/docs/en/flash.md
@@ -42,7 +42,17 @@ Source: [https://github.com/nodemcu/nodemcu-flasher](https://github.com/nodemcu/
 
 Supported platforms: Windows
 
-Note that this tool was created by the initial developers of the NodeMCU firmware. It hasn't seen updates since September 2015 and is not maintained by the current NodeMCU *firmware* team. Be careful to not accidentally flash the very old default firmware the tool is shipped with.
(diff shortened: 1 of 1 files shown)


mickley:
@marcelstoer: does your tool detect the flash size for 16MB models and flash correctly as 1MB (e.g. #1646)?

marcelstoer:
That's a good question but you picked the wrong project 😉 It should be https://github.com/marcelstoer/nodemcu-pyflasher. BUT since NodeMCU PyFlasher is nothing but a GUI for esptool.py https://github.com/espressif/esptool would be even better. In fact there's issue https://github.com/espressif/esptool/issues/123 for that already.
So, my tool does neither set nor detect flash size while flashing.

mickley:
Thanks @marcelstoer.  That's what I gathered looking at your code, just wanted to make sure I was correct.  Yes, I know the esptool can't do this.  

In that case, it'd be very helpful to add a GUI component to your flasher to set flash size manually to something other than the auto detect you're using.  I can probably do it myself when I have time, but it should be pretty trivial to add a dropdown more radio buttons, keeping auto detect the default.

Issue #1739, opened 2017-01-18
Extend node.info([buildRef,[modules]])


spaccabit:
Hi at all,
I'm trying different builds with different settings to understand how nodeMcu.
The more difficulty that I met it is to identify the build of firmware and the selected modules.

I would suggest adding to node.info optional parameters:

buildRef - returns the short hash of the commit to which the build refers
modules - returns a table containing the list of modules selected on user_modules.h

djphoenix:
For cloud builds there are some information that printed on bootup.
Not sure how it works at all. @marcelstoer I think you may describe it?

So I agree for expose that data somewhere in Lua-level.

marcelstoer:
My cloud builder simply redefines `NODE_VERSION` and `BUILD_DATE` in `user_version.h`. The Git revision is extracted by running `git rev-parse HEAD`.

There are more details here https://github.com/nodemcu/nodemcu-firmware/issues/449#issuecomment-109283562.

HHHartmann:
I am working at #2799 and would like to add these informations too. 
The idea is to return all of them in node.info.

HHHartmann:
I am working at #2799 and would like to add these informations too. 
The idea is to return all of them in node.info.

HHHartmann:
fixed by #2830

Issue #1223, opened 2016-04-03
How best to generate configurations for the firmware build process?


pjsg:
I have a couple of NodeMCU projects under way. These have specific configurations of modules and fonts that are needed (and this is more than the default).

The documented approach for this is to edit `user_config,h` and `u8g_config.h`. The problem with this approach is that when these files are changed in `dev`. problems often arise. Further, changes to these files can get accidentally committed -- note that the `user_config.h` needs to be modified to test / workon most of the modules. 

Various people have their own solutions to this problem. I suggest that one of these solutions be made official and merged into codebase.

TerryE:
Philip, see #905.  I was wrong: it was a PR and not an issue, but I limped away licking my wounds. :laughing: 
At twenty 20,000ft, I think that whatever the mechanism we have to accept that the code uses the user_*.h files and we don't want to change that. 

Whatever the local configuration is it is by definition local so  IMO we should have a `local` or `site` subfolder with a `modules` and an `includes` subdirectory which are ignored by git.  The `local/includes`  folder should be ahead of the `app/includes` so the user always has the option of having a personal version of these standard includes files and of course there is nothing to stop the user nesting includes in this local version.

I will stick to my generate script, but I would prefer to output to a local directory.

pjsg:
I'm actually testing a different approach which only involves a small change to the Dockerfile for the build image that just allows passing through some make parameters. I already have a script which wraps the docker run command, and I can pass in the extra stuff that I want there.

TerryE:
Yup, but real men don't use Docker ;-?  Seriously, I just keep my entire dev environment for nodemcu on its own VM; nice and clean.  We can't have a solution that only works on Docker.

pjsg:
You should be able to go

make "INCLUDES=-include `pwd`/user_override.h"

and this will include user_override.h (from the root directory of the tree) into every file before compiling.

Real men may not use docker, but they do know how to write Fortran in any language!

TerryE:
I wrote my first Fortan program 48 years ago. Frightening!!

My concern about the optional user_override method is that we would have to restructure our whole user_\*.h approach.  As you can see in my utility in gist, it's a mess.  For example you include a module by defining a constant.  A compile line forced include is inserted as a prologue, so if the default user_modules.h includes module X then you still need to edit the user_modules.h.

[...]

PR #1787, opened 2017-02-09
Correcting BME280 IIR filter setting

### LINKED ISSUE #1786: Bug in IIR filter settings for BME280 - incorrect pressure
### Expected behavior
Changing the IIR filter coefficient in bme280.init() should not (greatly) affect the values returned

### Actual behavior

When the IIR filter coefficient uses setting 1 (filter coefficient of 2), the pressure readings are way off.  Instead of being in the 1000 hectopascal range (100 kPa), the returned value is around 660 hectopascals (66 kPa), far lower than any atmospheric pressure on earth.  

The filter coefficient performs averaging of samples, so with lower filter coefficients, even though more variability would be expected, nothing this extreme should happen.

Furthermore, appropriate pressure readings are returned for every other filter coefficient setting, including turning the IIR filter off completely.  

### Test code

```Lua

-- Initialize the bme280 module on pins 1 and 2, with an IIR coefficient of 2
bme280.init(1, 2, nil, nil, nil, nil, nil, 1)

-- Wait 500 ms before testing to allow the sensor to acquire multiple readings
tmr.create():alarm(500, tmr.ALARM_SINGLE, function()

    -- Get temperature, pressure, and humidity from the BME280
    temp, pressure, humid = bme280.read()

    -- Print out the pressure reading
    print(pressure)

end)

```
### NodeMCU version
dev branch, commit: a271ce4d34da4efb2a25114da7a5725d344c85ce

### Hardware
WeMos D1 Mini with 4MB flash

FrankX0:
Fixes #1786.

Parameters "inactive_duration" and "IIR_filter" were incorrectly written to the config register.
@mickley: can you test?
@vsky279: do you agree?

### DIFF
--- a/app/modules/bme280.c
+++ b/app/modules/bme280.c
@@ -249,8 +249,8 @@ static int bme280_lua_init(lua_State* L) {
 		
 	bme280_ossh = (!lua_isnumber(L, 5))?BME280_OVERSAMP_16X:(luaL_checkinteger(L, 5)&bit3); // 5-th parameter: humidity oversampling
 	
-	config = ((!lua_isnumber(L, 7)?BME280_STANDBY_TIME_20_MS:(luaL_checkinteger(L, 7)&bit3))<< 4) // 7-th parameter: inactive duration in normal mode
-		| ((!lua_isnumber(L, 8)?BME280_FILTER_COEFF_16:(luaL_checkinteger(L, 8)&bit3)) << 1); // 8-th parameter: IIR filter
+	config = ((!lua_isnumber(L, 7)?BME280_STANDBY_TIME_20_MS:(luaL_checkinteger(L, 7)&bit3))<< 5) // 7-th parameter: inactive duration in normal mode
+		| ((!lua_isnumber(L, 8)?BME280_FILTER_COEFF_16:(luaL_checkinteger(L, 8)&bit3)) << 2); // 8-th parameter: IIR filter
 	full_init = !lua_isnumber(L, 9)?1:lua_tointeger(L, 9); // 9-th parameter: init the chip too
 	NODE_DBG("mode: %x\nhumidity oss: %x\nconfig: %x\n", bme280_mode, bme280_ossh, config);
 	


mickley:
@FrankX0: 

I'm still working on getting a docker setup running smoothly (very limited time here), but I'll be happy to test when it hits the dev branch and I can use nodemcubuild.com

mickley:
I've gotten docker working and can confirm that this fixes the bug in the BME280 module.  Pressures are now correct when the IIR filter parameter is 1.  

It's difficult to accurately test the filter itself, but turning the filter off (0), vs using a filter coefficient of 16 (4) show clear differences in data variability so it appears the filter is working correctly.

marcelstoer:
> I'm still working on getting a docker setup running smoothly

Feel free to leave feedback with https://github.com/marcelstoer/docker-nodemcu-build (assuming you're using that image) if you think documentation should be improved.

mickley:
@marcelstoer the docker side of things worked pretty well, though its inability to run on Windows 7 is a real shame.  

What irked me more is that it's rather tedious to build, and difficult to figure out.  The section on building firmware in the Nodemcu docs is very brief.  The location for user_modules.h and user_config.h are not mentioned, and this isn't covered at all from your docker repository's Readme.  You might get more individuals interested in helping out if this was more accessible and well-documented.  

Editing user_modules.h, user_version.h, and user_config.h (and sometimes more) each time you want to build a new firmware is also rather tedious.  Why not have all of that in one file?  Better yet would be a way to preserve multiple build configurations (e.g. different sets of modules) without maintaining multiple sets of the user_*****.h files, especially if one could do that via scripting or something.  

I find it quite useful that your nodemcu-build.com adds the commit, current date, and included modules into the firmware's version info string.  This is really the only way to see which modules a firmware contains.  I'm surprised that there's no built-in support for this when building via docker.  I've already added the build date automatically, but again this just seems like something that should be built-in at least as an option.

vsky279:
@FrankX0: Yes, you are right. There was a bug in the code. I was misled by the fact that Bit 1 does not have any meaning. The documentation on page 28 says:

**Table 26: Register 0xF5 “config”**

| Register 0xF5 “config” | Name | Description |
| ------------------------ |:-------:| -------------:|
| Bit 7, 6, 5     | t_sb[2:0] | Controls inactive duration t_standby in normal mode. See Table 27 for settings and chapter 5.3.4 for details.|
| Bit 4, 3, 2     | filter[2:0] | Controls the time constant of the IIR filter. See Table 27 for settings and chapter 5.4.4 for details.|
| Bit 0            | spi3w_en[0] | Enables 3-wire SPI interface when set to ‘1’. See chapter 8.3 for details.|

I think we can merge.

marcelstoer:
> The section on building firmware in the Nodemcu docs is very brief.

True, could you maybe document what you were missing and prepare a PR to that effect?

> The location for user_modules.h and user_config.h are not mentioned

There's some info at https://github.com/nodemcu/nodemcu-firmware#build-options but I agree it doesn't really belong there (historical left-over).

> and this isn't covered at all from your docker repository's Readme. 

Indeed but I argue it doesn't belong _there_ but _here_.

> Editing user_modules.h, user_version.h, and user_config.h (and sometimes more) each time you want to build a new firmware is also rather tedious. Why not have all of that in one file?

Yep #1223.

>  I'm surprised that there's no built-in support for this when building via docker.

_I_ am surprised I had to build that myself for the cloud builder. Again, I argue it should be a standard firmware feature; #1739.

marcelstoer:
Thanks Frank.

PR #1820, opened 2017-02-25
DS18B20 module (asynchronous)

### LINKED ISSUE #1795: DS18B20 module does not approach correctly the delay needed for temperature conversion
### Expected behavior
Before reading DS18B20 sensor the reader needs to wait before the conversion is finished. This applies especially to parasit-powered sensor where I think the scratchpad will not be holding last reading (so no temperature is returned).

### Actual behavior
In the existing example this is "solved" by the ([commented](https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_modules/ds18b20/ds18b20.lua#L86)) `tmr.delay(1000000)`. This is not the right approach in the NodeMCU Lua. Waiting 750 ms, the minimum delay before conversion finishes, makes the chip unstable.
This becomes even more unbearable if more sensors are connected and more delays need to be called. Only one conversion can occur at time for a parasitic-powered sensors due to electric current restriction.

### NodeMCU version
Latest dev branch

### Hardware
Any

### Proposed solution
A callback approach should be implemented.
I propose the following implementation [ds18b20-async.lua](
https://github.com/vsky279/nodemcu-firmware/blob/DS18B20/lua_modules/ds18b20/ds18b20-async.lua).
I would even suggest that the previous example is removed as I think it is misleading.

*Issue:* my understanding was that I can start more conversions at time for powered sensors (non-parasitic) but this does not work correctly for me. I need to start conversions [one by one](https://github.com/vsky279/nodemcu-firmware/blob/DS18B20/lua_modules/ds18b20/ds18b20-async.lua#L17). In my hw setup several powered sensors are connected + 1 parasitic-powered.

### Example
```lua
t=require("ds18b20-async")

function readout(temp)
  local resp=""
  for addr, temp in pairs(temp) do
    -- print(addr, temp) -- better representation base64 is preferred (when encoder module is included)
    print(encoder.toBase64(addr), temp)
  end
end

t:readTemp(readout)
```

vsky279:
Fixes #1795.

Implements callback approach to DS18B20 sensors readouts. Eliminates the `tmr.delay()` approach in previous version of the module. See #1795 for details.


TerryE:
@vsky279, my apologies for not picking up this issue / commit and commenting earlier.  Also compare and contrast my rewrite version at [this gist](https://gist.github.com/TerryE/5ea4f51492468a46b095279b0fff69350).  My comments are broken down into functional deficits flaws, unnecessary interdependencies and good Lua style.

## Functional Issues
-   You don't support family search
-   You don't support `ow.skip()`
-   Every read generates a new setup of the OW and a full search enumeration.
-  The conversion function issues serial conversions to all temperature devices on the OW and does so serially.  Why?
-  The module sends diagnostic to the print device whether the end application wants this or not.  Diagnostics should only be issues if the application enables them.
-  Device discovery should be done at power up or first call and cached.  Without skip, you can do true concurrent readings of multiple sensors.  By caching these to an **lc** file, this is deep sleep friendly. 
-  The 750mSec conversion time is a spec target.  However there is no 100% guarantee that the conversion will complete in 750mSec and if it hasn't, then reading at 750mSec will incorrectly read the previous scratchpad value.   Either you should add an extra allowance, say extend this to 1 sec or in the case of powered devices probe the device for completion.  I decided on the former approach because it was simple, and I could still read 15 devices every 1sec if I wanted so.

## Unnecessary module interdependencies

-  You are introducing a dependency on `encoder`to convert the device address to ascii format.  Why do this?  The `('%02X:%02X:%02X:%02X:%02X:%02X:%02X:%02X '):format(addr:byte(1,8))` technique is just as fast and uses standard Lua.

## Style and performance improvements

-  The point about not doing work every cycle when it can be done once and cached as above.
-  You are using a lot of keyed table accesses in your source. Do a `luac -l ds18b20.lua |  grep UPVAL` to see what I mean.  Using such global / keyed access is a *lot* slower than local or upvalue access, *especially* for ROtable contents (try benchmarking it) and generate more code.  The standard way to make this code run faster is to copy all used modules into upvalues which are bound into the function closures once, for example:
```Lua
-- Used modules and functions
local table, string, ow, tmr, print, type, tostring = 
      table, string, ow, tmr, print, type, tostring
-- Local functions
local ow_setup, ow_search, ow_select, ow_skip, ow_read_bytes, ow_write, ow_crc8, ow_reset = 
      ow.setup, ow.search, ow.select, ow.skip, ow.read_bytes, ow.write, ow.crc8, ow.reset
local now = tmr.now
```

I realise that this level of optimization is not essential for this application, but these distro modules will be looked at by our end-user Lua application developers and should represent best programming practice. 

## Other comments

As to parasitic vs non-parasitic powering, I haven't really got my head around the benefits of any of the latter.  AFAIK, First you need special parts for the latter which are more expensive, and programming them is more cumbersome because there are constraints on parallel conversion.   I wire all my DS18b20s to a bus that is <2cm long which a bunch of 3 pin headers on it.  I am totally lost as to what losing one wire gains.  You need to drive the bus high during conversion with a MSFET or equiv since the ESP826 GPIO isn't really powerful enough to do this on its own.  Given that most applications will used powered chip versions, I really don't understand why we should constrain the implementation for parasitic compatibility.

devyte:
@TerryE said:

> I really don't understand why we should constrain the implementation for parasitic compatibility.

I have a use case for temperature sensors that **requires** 2-wire cable. A 3rd wire is not available because Reasons, so parasite power is a must, and like you said, I use a mosfet to drive the line.
I have a similar use case for a pressure sensor, and again only 2 wires are available. Cable length in this case is about 40m, though, so I don't know if it will work (I haven't received the parts yet).

In both cases it is the available wiring that mandates the implementation, not the difficulty of the programming or additional parts required like the mosfet.

TerryE:
@devyte, what you are emphasising is that there are valid needs for a two wire parasitic configuration, but my underlying point that the driver doesn't support standard three wire modes still applies.

vsky279:
It will be good to have a clean implementation of the module. I agree my implementation was not perfect but I needed to get rid off the `tmr.delay()` approach.

Few comments and reactions to your implementation:
- For me parasitic-power support is also a must.
- The module works only on float version. If integer support is not implemented it needs to be at least documented.

- > Every read generates a new setup of the OW and a full search enumeration.

   In my use case I sometimes connect and disconnect sensors without restarting so enumerating every time makes sense for me. For general module your approach is probably more suitable.

- > The conversion function issues serial conversions to all temperature devices on the OW and does so serially. Why?

   This is because I have combination of parasitic and non-parasitic devices on one bus. If there are more parasitic-powered devices I cannot start conversion on all of them. Maybe it's not the case when you power parasitic through MOSFET but for me ESP8266 GPIO is powerfull enough to power conversion on one parasitic device at time. So I start readouts on all non-parasitic and on the first parasitic device.

- Usage of `encoder` module is legacy from my implementation where I use Base64 format to store device address. Following commit changed it to hex conversion which is an overkill.

devsaurus:
> If there are more parasitic-powered devices I cannot start conversion on all of them. Maybe it's not the case when you power parasitic through MOSFET

IMO implementing the strong driver for parasite power mode with a MOSFET doesn't change anything regarding the overall constraint that the bus must remain on high level when a sensor in parasite mode is executing a "Convert T" command. This rules out any communication once the first parasite sensor started conversion.

@devyte:
> I use a mosfet to drive the line.

I'm interested to learn about the implications of this approach.
- How is the MOSFET connected to the ESP?
- Which features of the `ow` module for parasite mode do you actually use when you have a MOSFET in place? I'd guess that you don't use the `power` mode of the commands since your MOSFET does all the heavy lifting, right?
- Are there further features which we could add to the module to support this use case?

vsky279:
>overall constraint that the bus must remain on high level when a sensor in parasite mode is executing a "Convert T" command. This rules out any communication once the first parasite sensor started conversion.

I don't think there is any disagreement on this fact.

TerryE:
## Power Modes
 
@devsaurus Armin, the parasitic and non-parasitic templates have two different hardware design patterns.  The powered or non-parasitic is a lot simpler because you power the OW bus directly off the Vcc to the chip. My calibration rig will happily drive15 DS18B20s.  I calibrate them because once adjusted for calibration errors they are accurate to about 0.03 °C over the temperature range that I use and I prefer to work with a know accuracy.  In practice, I wire them as a radial configuration with the OW on a through-hole veroboard.  [This post](http://forum.buildhub.org.uk/ipb/topic/2261-logging-from-ds18b20s/#comment-34278) shows my small test rig.  The DQ line is driven directly from the GPIO but with a 10K pull-up to Vcc. 

I suspect that Lukáš configures his two-wire thermometers much more like fairy lights literally strung along a 2-wire cable.  The GPIO output of the ESP8266 is only rated at 12 mA which is nowhere near enough to drive a string of OW devices in parasitic mode, hence the need for a MOSFET to drive the line up and the inability to slipstream conversions across devices.

## Float vs Integer Versions

Hopefully this is going to go away soon as Lua 5.3 supports both number types natively, but in the meantime, I think that the Integer Version was largely there to solve a historic problem: the early NudeMCU 0.9x versions didn't have enough free memory to do anything usable with the float build.  Because of the  need for SDK2 support all current modules ship with at least 1Mb Flash, and Jonny's unaligned flash expcetion handling, plus my LCD patch now mean that we can trivially build configs with >300Kb SPIFFS and 40Kb+ free at boot.  

So today if you have an application that *needs* an interger beild then you are doing something wrong.

## Storing Keys in Hex

Why? Lua support binary strings.  The only possible reason to use hex is to  have them printable, and a `tostring` fuction is good enough for this.  

## Style and performance issues

Not using Lua best practice is just lazy, IMO.  I know that we probably don't need the margin, but we should start to set some minimum bar for our example Lua code because new programmers will use these files just as that: best practice examples.

vsky279:
## Float vs Integer Versions
Historically I'm using only the integer version. I am still using some older chips where I'm not sure that float would fit. I need to reconsider this.

## Storing Keys in Hex
I store a json file with mapping sensor address -> sensor name (`"{KBfJUwYAAJA=":"OUT.temp"}`) and I wanted something readable/editable. Of course hex would be fine but I choosed base64 encoding for no real reason.

## Style and performance issues
Definitely I agree. It's just that my programming skills do not include Lua best practices - maybe because of wrong Lua examples? :-)

TerryE:
@vsky279, the Wemos D1 modules cost about 3€ and the  ESP-01 1Mb part slightly less.  It's almost impossible to fit a current SDK 2.0 build on a ½Mb module anyway.    

Also have a look at how I use lc files much the same way as you use json.  

My surname is ellison and if you want ping me an email to my user ID at my .org.uk domain.  

PS. if you enable issues and review comments on your repo then we could ping=pong there.  

Also if you do a `luac -l ds18b20.lua | grep -cP 'GET(TABLE|GLOBAL)'` on your vs my modules then you will see that the counts are 108 vs 33 respectively and nearly all of mine are one-off at compile or closure., and I don't do algo stuff in the transaction loop that can be done as a one-off.

PR #1834, opened 2017-03-06
added noob 7bit address info

thirschbuechler:

to make the first i2c implementation for any noob (like me) more transparent, I propose to add the information that the address is 7bit only. Some datasheets don't include that, and it's just nice to have.
[Yes, one could just follow the link into `i2c.c`, but that's not very noob proof :) ]

### DIFF
--- a/docs/en/modules/i2c.md
+++ b/docs/en/modules/i2c.md
@@ -11,8 +11,10 @@ Setup I²C address and read/write mode for the next transfer.
 
 #### Parameters
 - `id` always 0
-- `device_addr` device address
+- `device_addr` device address (7-bit)<sup>[1](#7bit)</sup>
 - `direction` `i2c.TRANSMITTER` for writing mode , `i2c. RECEIVER` for reading mode
+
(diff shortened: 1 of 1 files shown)


docs/en/modules/i2c.md
```diff
@@ -11,9 +11,11 @@ Setup I²C address and read/write mode for the next transfer.
 
 #### Parameters
 - `id` always 0
-- `device_addr` device address
+- `device_addr` device address (7-bit)<sup>[1](#7bit)</sup>
```
marcelstoer: I'd put that differently, skipping the reference.

> `device_addr` 7-bit device address (8bit frame = 7bit address + R/W bit as LSB)

What do you think?

devsaurus: I also think that `device_addr` can go without a reference, but propose to explain the construction of the resulting "8bit frame" in the command description above. Something like "`device_addr` is used for the upper 7 bits followed by the `direction` bit". This relates to the wording in [chapter 3.1.10 of the I2C spec](http://www.nxp.com/documents/user_manual/UM10204.pdf) and hopefully rules out misunderstanding.

PR #1837, opened 2017-03-06
Improve wifi docs

### LINKED ISSUE #1744: inconsistency in module wifi
I mention this because it got me once and I will never get back those hours...
The [doco](http://nodemcu.readthedocs.io/en/dev/en/modules/wifi/#wifistastatus) for `wifi.sta.status()` does not say where these constants are:
```
Returns
number 0~5

- 0: STA_IDLE,
- 1: STA_CONNECTING,
- 2: STA_WRONGPWD,
- 3: STA_APNOTFOUND,
- 4: STA_FAIL,
- 5: STA_GOTIP.
```
`eventMonReg()` does say `wifi.STA_GOTIP` but it is deprecated and will probably be removed soon.

Even if we do not place the constants where they belong, the doco should clearly clarify, something like:
```
- 5: wifi.STA_GOTIP,
```
if we even want to mention the numeric value...
### Expected behavior
A function and its return values belong together, so one expects `wifi.sta.status()`, `wifi.sta.STA_GOTIP` etc.
### Actual behavior
We have `wifi.STA_GOTIP` instead.
### Test code
none
### NodeMCU version
dev
### Hardware
none

luismfonseca:
Fixes #1744.

I'm not sure about which part of the wifi is being deprecated as stated in the issue.

### DIFF
--- a/docs/en/modules/wifi.md
+++ b/docs/en/modules/wifi.md
@@ -993,14 +993,14 @@ Gets the current status in station mode.
 `nil`
 
 #### Returns
-number： 0~5
+The current state which can be one of the following:
 
-- 0: STA_IDLE,
-- 1: STA_CONNECTING,
-- 2: STA_WRONGPWD,
-- 3: STA_APNOTFOUND,
-- 4: STA_FAIL,
-- 5: STA_GOTIP.
+- `wifi.STA_IDLE`
+- `wifi.STA_CONNECTING`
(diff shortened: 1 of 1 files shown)


marcelstoer:
> I'm not sure about which part of the wifi is being deprecated as stated in the issue.

Me neither, the documentation doesn't say anything at all about that.

PR #1838, opened 2017-03-07
TIME_WAIT sockets fixes

### LINKED ISSUE #1836: TIME_WAIT sockets issues
Now we have two (*or more? Point me if I miss something*) critical issues related to sockets kept in TIME_WAIT state:
- Cannot open new `server` instance when any socket in `TIME_WAIT` state with matched target port is present
- "Memory leak" (actually high consumption that rolls back after time)

I think we have a several ways to go (or maybe mix them together):
- [x] Reduce `TIME_WAIT` state timeout for socket, so them will be closed earlier
- [x] ~Kill blocking `TIME_WAIT` sockets when calling `listen` method~. Enable `SO_REUSEADDR` flag.
- [ ] Kill `TIME_WAIT` sockets somewhere in Lua GC (I not so familiar with lua-level, anyone can explain this?)

Objections? I think we should resolve this ASAP (**before** upcoming master drop).
/cc @marcelstoer @devsaurus

djphoenix:
Fixes #1836 (at most)

### DIFF
--- a/app/include/lwip/tcp_impl.h
+++ b/app/include/lwip/tcp_impl.h
@@ -130,8 +130,7 @@ u32_t            tcp_update_rcv_ann_wnd(struct tcp_pcb *pcb)ICACHE_FLASH_ATTR;
 #define TCP_OOSEQ_TIMEOUT        6U /* x RTO */
 
 #ifndef TCP_MSL
-/* See https://github.com/nodemcu/nodemcu-firmware/issues/1836 for details */
-#define TCP_MSL 5000UL /* The maximum segment lifetime in milliseconds */
+#define TCP_MSL 60000UL /* The maximum segment lifetime in milliseconds */
 #endif
 
 /* Keepalive values, compliant with RFC 1122. Don't change this unless you know what you're doing */
--- a/app/include/lwip/tcp_impl.h
+++ b/app/include/lwip/tcp_impl.h
@@ -130,7 +130,7 @@ u32_t            tcp_update_rcv_ann_wnd(struct tcp_pcb *pcb)ICACHE_FLASH_ATTR;
 #define TCP_OOSEQ_TIMEOUT        6U /* x RTO */
 
 #ifndef TCP_MSL
-#define TCP_MSL 60000UL /* The maximum segment lifetime in milliseconds */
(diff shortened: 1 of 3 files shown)


app/include/lwip/tcp_impl.h
```diff
@@ -130,7 +130,7 @@ u32_t            tcp_update_rcv_ann_wnd(struct tcp_pcb *pcb)ICACHE_FLASH_ATTR;
 #define TCP_OOSEQ_TIMEOUT        6U /* x RTO */
 
 #ifndef TCP_MSL
-#define TCP_MSL 60000UL /* The maximum segment lifetime in milliseconds */
+#define TCP_MSL 5000UL /* The maximum segment lifetime in milliseconds */
```
marcelstoer: Maybe leave a comment there indicating _why_ this was changed so next time lwIP is updated it shows up in the diff. Can even include a reference to the GitHub issue number. If we don't document such changes there's a real risk that subsequent SDK updates will revert them w/o us noticing.

app/include/lwipopts.h
```diff
@@ -1455,7 +1455,7 @@
  * SO_REUSE==1: Enable SO_REUSEADDR option.
  */
 #ifndef SO_REUSE
-#define SO_REUSE                        0
+#define SO_REUSE                        1
```
marcelstoer: dito

djphoenix:
@marcelstoer updated

devyte:
Rather than make changes inside the lwip includes, which is is really a
deviation of the stock lwip code, why not add external defines for
configuration? The includes say #ifndef blah, #define blah.

djphoenix:
@devyte Agree for `tcp_impl.h` file change, moved it into `lwipopts.h`

djphoenix:
Objections here?

marcelstoer:
No, but since only 2/3 of #1836 are completed I thought you had planned to add more commits to _this_ PR. I understand you want to create a separate PR at a later point for the 3rd issue, right?

djphoenix:
Yep, I think this PR is huge step forward, and I haven't meet Lua GC anyway. So I will try to work on it some later (if nobody finished it earlier). But now these small changes gives big improvements, so I vote for land it as-is.

PR #1859, opened 2017-03-12
Extend node.dsleep() to support instant sleep.

### LINKED ISSUE #1776: There is a use case for a fast deep sleep feature
### Missing feature
Ability to enter dsleep quickly.
### Justification
I have an app that completes in under 500ms. `node.dsleep()` takes about 100ms extra which is a significant cost in battery power.

I now added a `node.dsleepInstant()` lua function with an implementation that is almost identical to `node.dsleep()`. I copied `node.c:node_deepsleep()` except for this small change:
```
      system_deep_sleep( us );
is replaced with
      {
        void system_deep_sleep_instant(uint32 time_in_us);
        system_deep_sleep_instant( us );
      }
```
Deep sleep is entered in 1-2ms and uses the same low power.

This feature is documented in section `4.5` of [`9b-esp8266_low_power_solutions_en.pdf`](https://espressif.com/sites/default/files/documentation/9b-esp8266_low_power_solutions_en.pdf).

BTW, I wonder if the `rtctime` module, which has a private fast dsleep routine, can use this feature too.

jmattsson:
Fixes #1776.

- [ ] I have thoroughly tested my contribution. **I have not, but @eyaleb has quite comprehensively done it in #1776**

This introduces the "instant" deep sleep feature which Espressif added but largely hid away. See #1776 for further details. Instant deep sleep effectively shaves off ~100ms on the going-to-sleep, which can be quite significant for battery-powered devices.

Due to the lack of documentation, I haven't made the instant deep sleep the default since I don't fully know what the difference between the two deep sleep functions are.

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -53,6 +53,9 @@ static int node_deepsleep( lua_State* L )
     else
       system_deep_sleep_set_option( option );
   }
+  bool instant = false;
+  if (lua_isnumber(L, 3))
+    instant = lua_tointeger(L, 3);
   // Set deleep time, skip if nil
   if ( lua_isnumber(L, 1) )
   {
@@ -61,7 +64,12 @@ static int node_deepsleep( lua_State* L )
     if ( us < 0 )
       return luaL_error( L, "wrong arg range" );
     else
-      system_deep_sleep( us );
+    {
+      if (instant)
+        system_deep_sleep_instant(us);
+      else
+        system_deep_sleep( us );
+    }
   }
   return 0;
 }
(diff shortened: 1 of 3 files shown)


jmattsson:
Huh! I never knew it was possible to link *into* a pdf!

And I have no idea how you managed to commit onto my branch? I assume it's some github custom deviltry? :)

Learned something new here - cheers!

marcelstoer:
> I never knew it was possible to link *into* a pdf!

That feature has been around for many years (10ish?) but I don't know how widely adopted it's become in browsers and PDF readers/plugins.

> I have no idea how you managed to commit onto my branch?

Cool, isn't it! When you create a PR GitHub offers this "Allow edits from maintainers" option (next to the green create-PR button), it's checked by default.

jmattsson:
Ah, *that's* what that box does. I figured it was referring to the PR meta-info. Live & learn, eh? :)

PR #1875, opened 2017-03-22
HDC1080 Module Add

saucompeng:
New Module


app/modules/hdc1080.c
```diff
@@ -0,0 +1,111 @@
...
+    platform_i2c_send_start(hdc1080_i2c_id);
+    platform_i2c_send_address(hdc1080_i2c_id, hdc1080_i2c_addr, PLATFORM_I2C_DIRECTION_RECEIVER);
+
+    for (i=0; i<2; i++) {
+	data[i] = platform_i2c_recv_byte(hdc1080_i2c_id, 1);
+    }
+
+    platform_i2c_send_stop(hdc1080_i2c_id);
+
+    temp = ((data[0]*256+data[1])/pow(2,16))*165-40;
```
pjsg: Does this work in the integer build? In the float build it seems a pity that you discard the fractional component of the temperature.

docs/en/modules/hdc1080.md
```diff
@@ -0,0 +1,56 @@
...
+This module provides access to the [HDC1080](http://www.ti.com/product/HDC1080) . Low Power, High Accuracy Digital Humidity Sensor with Temperature Sensor.
+
+## hdc1080.readTemperature()
+Samples the sensor and returns temperature value.
+
+#### Syntax
+`hdc1080.readTemperature()`
+
+#### Returns
+Temperature data (integers)
```
pjsg: In what units?

docs/en/modules/hdc1080.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## hdc1080.readHumidity()
+Samples the sensor and returns temperature value.
+
+#### Syntax
+`hdc1080.readHumidity()`
+
+#### Returns
+Humidity data (integers)
```
pjsg: I guess that this is a percentage from 0 - 100??

app/modules/hdc1080.c
```diff
@@ -0,0 +1,111 @@
...
+    uint8_t data[2];
+    int temp;
+    int i;
+
+    platform_i2c_send_start(hdc1080_i2c_id);
+    platform_i2c_send_address(hdc1080_i2c_id, hdc1080_i2c_addr, PLATFORM_I2C_DIRECTION_TRANSMITTER);
+    platform_i2c_send_byte(hdc1080_i2c_id, HDC1080_TEMPERATURE_REGISTER);
+    platform_i2c_send_stop(hdc1080_i2c_id);
+
+    os_delay_us(20000);
```
pjsg: This seems long. The max time taken for the conversion is < 7ms for 14 bit accuracy. I don't see why you are requesting 14 bit accuracy when you only return an integer temperature. If you requested 11, then the conversion time is under 4ms.....

Similarly for the humidity....

docs/en/modules/hdc1080.md
saucompeng: This is percentage, I will fix

app/modules/hdc1080.c
saucompeng: True, I will change it.

app/modules/hdc1080.c
saucompeng: True, I change it to float.

docs/en/modules/hdc1080.md
saucompeng: Sorry :) in Centigrade

app/modules/hdc1080.c
saucompeng: @pjsg  I'm new  at github, Which branch I should make pull request? Master or dev ?

pjsg: You are doing it right -- the PR should be to nodemcu:dev  -- we then merge this to master every 2-3 months when things look good.

app/modules/hdc1080.c
pjsg: You need to be careful -- there are two builds of the nodemcu firmware -- one is integer only and the other includes floats. This makes this sort of thing complicated!

saucompeng: https://github.com/nodemcu/nodemcu-firmware/blob/dev/docs/en/modules/dht.md

> If using float firmware then `temp` and `humi` are floating point numbers. On an integer firmware, the final values have to be concatenated from `temp` and `temp_dec` / `humi` and `hum_dec`.

Is it true method  ?

PR #1886, opened 2017-03-28
Support for extra collection of data in end user configuration (#1302)

### LINKED ISSUE #1302: Enduser setup (EUS) - Provide more hooks for code
So my use case is this: 

I want to use EUS for board set-up like it was meant  to be used, but I also want extra form fields (a human-readable identifier for the board and a secret code). I know I can edit the HTML, but I also want to get at all the values entered on the form on the lua side.

For more background: I am planning a little thermostat project where the board sends temperature and humidity readings identified by the human-readable id entered on the EUS form, and the data HMAC-SHA1 signed using the secret also from the EUS form). Trying to make it as easy to set up as possible.

rvowles:
This patch includes the ability to request extra fields during
end user configuration, have them validated by a LUA callback,
and have the results stored in a JSON file. The user will need
to use a json library (sjson) to read the data.

Fixes #1302 .

This addresses the issue where the user wishes to collect more data at start up and ensure that it is validated. Once the data is validated it is stored in a json file for the init.lua to load and process as it sees fit. 

I have included the .editorconfig for various editor environments that seems to match the project's settings, extra ignore for IDEA/CLion as well as the code for the functionality. My C is rusty so any feedback on how to do something more cleanly would be welcome. Please be nice. :-)


rvowles:
no problem :-)

app/modules/enduser_setup.c
```diff
@@ -625,6 +626,228 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+               int *wifi_ssid_loc, int *wifi_pwd_loc) {
+  int p_file = 0;
+  lua_State *L = NULL;
+
+  if (param_count > 2)
+  {
+    ENDUSER_SETUP_DEBUG("enduser: opening enduser.json for write");
+
+    // setup the file output
+    p_file = vfs_open("enduser.json", "w");
```
marcelstoer: As the supported custom HTML file is called `enduser_setup.html` I suggest to call this one `enduser_custom_parameters.json` or so.

rvowles: I'm not entirely sure I follow that logic? I mean, I don't mind, but I always thought shorter and meaningful is better when you have limited memory. But I will change it.

marcelstoer: You're right, it may not be 100% logical but "short" is sometimes the opposite of "meaningful" 😉 

My point was that `enduser.json` is IMHO too short or too generic. You can't deduce the content of the file from its name.

rvowles:
this is just a ping - do I need to do anything else here?

marcelstoer:
Nope, your part is done, thanks 😄 However, a maintainer or contributor familiar with the EUS code (@robertfoss, @jfollas,@jmattsson,@FrankX0 ) should approve and/or merge this.

Ognian:
I'm eagerly awaiting this to be merged, so that I can test it. Is there any chance to get this merged soon? Shouldn't the "changes requested" state be cleared too?
Sorry if disturbing.
Thanks
Ognian

rvowles:
It is now May 10, is there any chance this will be progressed?

app/modules/enduser_setup.c
```diff
@@ -625,6 +626,228 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+  for (pos = 0; pos < data_len; pos++) {
+    if (data[pos] == '?' && (pos + 1 < data_len)) {
+      fields[counter] = &data[pos + 1];
+      break;
+    }
+  }
+
+  data[data_len] = 0; // this must be true by definition
+
+  for (pos = 0; pos < data_len && counter < MAX_HTTP_FIELDS && state != STATE_FIN; pos++) {
```
pjsg: shouldn't this loop start at where the previous one left off?

app/modules/enduser_setup.c
```diff
@@ -625,6 +626,228 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+    //std::cout << "val" << pos << ":" << data[pos] << std::endl;
+    if (data[pos] == '&' || data[pos] == ';') { // w3c states & or ; separate params
+      if (lastPos != -1) {
+        data[pos] = 0; // terminate the string
+        counter++;
+
+        if (counter < MAX_HTTP_FIELDS) {
+          fields[counter] = NULL;
```
pjsg: There is a loop above that initializes all of these to NULL. Do you need to do it again?

app/modules/enduser_setup.c
```diff
@@ -625,6 +626,228 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+      }
+
+      file_field_counter++;
+      vfs_write(p_file, "\"", 1);
+
+      enduser_setup_http_urldecode(file_buf, fields_str[p_count], c_strlen(fields_str[p_count]), max_len);
+      if (L != NULL) {
+        lua_pushlstring(L, file_buf, c_strlen(file_buf));
+      }
+      vfs_write(p_file, file_buf, c_strlen(file_buf));
```
pjsg: I don't see any escaping of " or \ when writing out the JSON.....

app/modules/enduser_setup.c
```diff
@@ -1674,6 +1957,13 @@ static int enduser_setup_init(lua_State *L)
     ENDUSER_SETUP_DEBUG("enduser_setup_init: Debug callback has been set");    
   }
 
+  if (!lua_isnoneornil(L, 4))
+  {
+    lua_pushvalue (L, 4);
+    state->lua_validation_cb_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```
pjsg: I note that this (like the other assignments above) can leak memory if there was a previously set callback. This can happen if you call init twice... The second time it detects that it was already initialized, and skips the initializtion of the state block. Then it overwrites the pointers. Yes, it will return an error, but it is too late then.

marcelstoer:
> My concern with doing the querystring parsing in C is the increased possibility of user-provided data causing the firmware to panic.

Triggered by this comment I took a look at the code myself (not being a firmware coder mind you). I share Jason's concerns and found something else that bothers me: "hand-woven" JSON handling.

Why can't `/setwifi` and `/update` accept a single additional well defined parameter (e.g. `config`) whose value represents a URL-encoded JSON object? Furthermore, shouldn't the module have an implicit dependency to the SJSON module rather than deserializing the data into a file manually? Would address Philip's comment regarding escaped values as well. However, if the module accepted a serialized JSON object rather than individual querystring parameters JSON-handling wouldn't be required anymore at all as it would simply pipe the config string into a file, no?

pjsg:
The other option would be to construct a lua table out of the parameters and pass that into the callback The callback can write it to a file (using sjson) or do something else.

rvowles:
Ok, given these suggestions are the anti-thesis of this PR I have closed it. It suits my purpose so I will just maintain my fork.

marcelstoer:
Richard, could you maybe elaborate why you think the suggestions were the anti-thesis of your implementation? I see it more as an _elaboration_ of the foundation you laid.

rvowles:
HI Marcel - there were two types of suggestions that started to come out - code reviews (which didn't change the fundamental nature of the PR) and alternative ways this whole process should work suggestions.

The latter suggestions require a complete re-write of this PR. What is being suggested is quite different to what this PR seeks to achieve (simple key values, no dependence on sjson if you don't use it, etc). 

I understand there is a clash between what the project desires and what I need right now, thats always true in open source projects :-)

Jcrash29:
Is this still a PR that needs to get merged into master?

I'm currently exploring options to input two extra pieces of data from the endUser_Setup page. 

What methods would people recommend currently?

marcelstoer:
The OP closed this PR after we suggested some improvements\*. That's his choice and we have to accept that. Hence, unless someone else picks up where the OP left off and adds the missing pieces this is unlikely to get merged.

\* FWIW, I still don't see how `/setwifi?config=%7B%22foo%22%3A%20%22bar%22%2C%22fooz%22%3A%20%22bazz%22%7D` (URL encoded JSON object) would be so fundamentally different from `/setwifi?foo=bar&fooz=bazz` (apart from its advantages). Once `config` is URL _de_ coded it can be dumped into a file as-is. SJSON is not required for that but it's helpful to use it for reading the config again. On the client-side producing the `config` value is simple:
```
var json, config = {};
config.foo = "bar";
config.fooz = "bazz";
json = JSON.stringify(config);
encodeURIComponent(json);
```

Jcrash29:
@marcelstoer  I'm going to attempt to pick up where OP left off. For now I will highjack this thread with questions. 

I spent some time last night studying the code base can you correct me if I'm wrong:

The change you mentioned above about adding this to /update or /setwifi would go in the C code and get added to the SJSON file by the C code. Than the user in LUA would decode the message in a similar method to what you showed?

Is that accurate?

PR #1889, opened 2017-03-29
Update wifi event monitor documentation

### LINKED ISSUE #1737: Is there a race in eventmon?
This is not a simple missed trigger bug but a design (SDK or nodemcu?) issue.
### Expected behaviour
Registering an event should inform of the event happening.
### Actual behaviour
The event does not always trigger.
### Test code
If I do this early (to stop dhcp)
```lua
wifi.sta.setip({ip="...",netmask="...",gateway="..."})
```
then (at times much) later I attend to the wifi
```lua
local eventmon = wifi.eventmon
eventmon.register(eventmon.STA_CONNECTED, function(T)
    print("status="..wifi.sta.status()
end)
```
I usually see
```
status=5
```
However, if the registration is too late, after the connection is established, then no trigger is generated and a wdt reset takes place.

I do not see a simple way to ensure I **always** get notified. I now do this:
```lua
wifi.sta.setip({ip="...",netmask="...",gateway="..."})
...
local handled = false
eventmon.register(eventmon.STA_CONNECTED, function(T)
    if not handled then
        handled = true
        print("status="..wifi.sta.status())
        whatever...
    end
end)
if not handled and 5 == wifi.sta.status() then
    handled = true
    print("handling missed event")
    eventmon.unregister(eventmon.STA_CONNECTED)
    whatever...
end
```
but is setting `handled` properly serialised?
### NodeMCU version
Latest `dev`
### Hardware
Any

dnc40085:
Update wifi event monitor documentation.

Fixes #1737.

A lack of information about when the wifi event monitor callbacks should be registered is causing confusion, the new information provided should help to clarify the function of the event monitor .

### DIFF
--- a/docs/en/modules/wifi.md
+++ b/docs/en/modules/wifi.md
@@ -399,7 +399,7 @@ none
 Registers callbacks for WiFi station status events.
 
 !!! note
-    Please update your program to use the [`wifi.eventmon`](#wifieventmon-module) API, as the `wifi.sta.eventmon___()` API is to be depreciated. 
+    Please update your program to use the [`wifi.eventmon`](#wifieventmon-module) API, as the `wifi.sta.eventmon___()` API is deprecated. 
 
 ####  Syntax
 - `wifi.sta.eventMonReg(wifi_status[, function([previous_state])])`
--- a/docs/en/modules/wifi.md
+++ b/docs/en/modules/wifi.md
@@ -398,8 +398,8 @@ none
 
 Registers callbacks for WiFi station status events.
 
-  !!! Note
-   Please update your program to use the [`wifi.eventmon`](#wifieventmon-module) API, as the `wifi.sta.eventmon___()` API is to be depreciated. 
+!!! note
+    Please update your program to use the [`wifi.eventmon`](#wifieventmon-module) API, as the `wifi.sta.eventmon___()` API is to be depreciated. 
 
 ####  Syntax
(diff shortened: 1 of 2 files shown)


docs/en/modules/wifi.md
```diff
@@ -398,6 +398,9 @@ none
 
 Registers callbacks for WiFi station status events.
 
+!!! note
+    Please update your program to use the [`wifi.eventmon`](#wifieventmon-module) API, as the `wifi.sta.eventmon___()` API is to be depreciated.
```
marcelstoer: > is to be depreciated

1.  That should be "deprecated" I guess?
2.  "to be" confuses me a little here, is it deprecated or not?

dnc40085: >That should be "deprecated" I guess?
"to be" confuses me a little here, is it deprecated or not?

You're right, I found this [Wikipedia page on deprecation](https://en.wikipedia.org/wiki/Deprecation), and it is indeed "deprecated".

marcelstoer:
So, it seems those functions really are deprecated (sorry, didn't know). How about adding `platform_print_deprecation_note` to them in code?

Example usage: 
- https://github.com/nodemcu/nodemcu-firmware/pull/1887/files#diff-ae644cf8a7a5d6c9cff655fcf9e16aedR53
- https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/modules/net.c#L319

dnc40085:
>How about adding platform_print_deprecation_note to them in code?

Done

PR #1895, opened 2017-04-02
Add module for TCS34725 colour sensor.

### LINKED ISSUE #1894: No support for TCS34725 colour sensor
8<------------------------ FEATURE REQUEST ------------------------------------
### Justification
I want to detect colour with nodemcu.

### Workarounds
Write it. Creating issue prior to pull request.

8<------------------------ END FEATURE REQUEST --------------------------------

tjhowse:
Fixes #1894.


app/modules/tcs34725.c
```diff
@@ -0,0 +1,371 @@
...
+	uint8_t scl;
+	uint8_t config;
+	uint8_t ack;
+
+	if (!lua_isnumber(L, 1) || !lua_isnumber(L, 2)) {
+		return luaL_error(L, "wrong arg range");
+	}
+	sda = luaL_checkinteger(L, 1);
+	scl = luaL_checkinteger(L, 2);
+	platform_i2c_setup(TCS34725_BUS_ID, sda, scl, PLATFORM_I2C_SPEED_SLOW);
```
devsaurus: The I2C setup should not be part of a module's `init` function, see #1887 where the existing instances are being reworked. Please remove it here together with the `sda` and `scl` parameters.
NB there's no requirement to implement both variants with and without i2c setup. Just start this new module with a cleanly without.

app/modules/tcs34725.c
```diff
@@ -0,0 +1,371 @@
...
+	platform_i2c_send_stop(TCS34725_BUS_ID);
+	
+	return (high << 8) | low;;
+}
+/**************************************************************************/
+/*!
+		@brief	Finishes enabling the device
+*/
+/**************************************************************************/
+uint8_t tcs34725EnableDone(lua_State* L)
```
devsaurus: Is a `os_timer_disarm()` missing here?

docs/en/modules/tcs34725.md
```diff
@@ -0,0 +1,98 @@
...
+## tcs34725.init()
+
+Initializes module. Initialization is mandatory before values can be read.
+
+#### Syntax
+
+`tcs34725.init(sda, scl)`
+
+#### Parameters
+- `sda` - SDA pin
```
devsaurus: Don't include I2C setup, see above.

app/modules/tcs34725.c
```diff
@@ -0,0 +1,371 @@
...
+	uint16_t g;
+	uint16_t b;
+	uint16_t c;
+	
+	if (!_tcs34725Initialised)
+	{
+		tcs34725Init(L);
+	}
+
+	/* ToDo: Insert a blocky delay until the data is ready! */
```
devsaurus: How long would the delay be?
Can the delay be implemented on Lua level instead here?

app/modules/tcs34725.c
```diff
@@ -0,0 +1,371 @@
...
+	dbg_printf("id: %x\n",id);
+	if (id != 0x44) {
+		luaL_error(L, "No TCS34725 found.");
+		return 0;
+	}
+
+	/* Enable the device */
+	// TODO this doesn't quite work as it should. Enable is non-blocking.
+	if (tcs34725Enable(L) != 0) {
+		luaL_error(L, "TCS34725 Not initialising.");
```
devsaurus: `tcs34725Enable()` doesn't provide any post-init status as you already noted.

app/modules/tcs34725.c
```diff
@@ -0,0 +1,371 @@
...
+	sda = luaL_checkinteger(L, 1);
+	scl = luaL_checkinteger(L, 2);
+	platform_i2c_setup(TCS34725_BUS_ID, sda, scl, PLATFORM_I2C_SPEED_SLOW);
+	
+	/* Make sure we have the right IC (0x44 = TCS34725 and TCS34721) */
+	id = tcs34725Read8(TCS34725_ID);
+	dbg_printf("id: %x\n",id);
+	if (id != 0x44) {
+		luaL_error(L, "No TCS34725 found.");
+		return 0;
```
devsaurus: Here and below: `luaL_error()` implicitly leaves the function, see [`luaL_error`](http://pgl.yoyo.org/luai/i/luaL_error).
You could use the `return luaL_error()` pattern for clarity.

FrankX0:
Thanks for adding this sensor to the firmware!
A few remarks:

- Please add a `setup` function which does the same as `init` but without initialising the I2C bus (and remove the `init` function). This in accordance with #1887.
- Add the module to `user_modules.h` and make it non-default.
- Make an entry for `tcs34725.md` in `mkdocs.yml`.

devsaurus:
Regarding `tcs34725Enable()`:

My understanding is that `tcs34725.enable()` triggers a timed 2-stage process. This starts off asynchronously while the Lua script continues, but the script is supposed to leave the hardware alone until the process finished, right?
In this case `tcs34725.enable()` should have a callback function parameter which gets triggered once the enable process is done. Like in
```lua
tcs34725.enable(function()
  print("hardware enabled")
  clear,red,green,blue=tcs34725.raw()
end)
```

tjhowse:
Not quite ready yet. Still chasing a timing issue with very long integration times. I might have to deliver the result of raw() to a callback.

tjhowse:
Is there anything else I need to do to complete this PR? I think it should be ready to go.

app/modules/tcs34725.c
```diff
@@ -0,0 +1,359 @@
...
+{
+	dbg_printf("Enable begun\n");
+
+	if (lua_type(L, 1) == LUA_TFUNCTION || lua_type(L, 1) == LUA_TLIGHTFUNCTION) {
+		if (cb_tcs_en != LUA_NOREF) {
+			luaL_unref(L, LUA_REGISTRYINDEX, cb_tcs_en);
+		}
+		lua_pushvalue(L, 1);
+		cb_tcs_en = luaL_ref(L, LUA_REGISTRYINDEX);
+		self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```
devsaurus: If I understand the code correctly  then `self_ref` contains in the end also the reference to the callback function. This would be redundant. IMO the whole `self_ref` thingy can be removed.

app/modules/tcs34725.c
```diff
@@ -0,0 +1,359 @@
...
+	/* Ready to go ... set the initialised flag */
+	_tcs34725Initialised = true;
+	
+	/* This needs to take place after the initialisation flag! */
+	tcs34725SetIntegrationTime(TCS34725_INTEGRATIONTIME_2_4MS, L);
+	tcs34725SetGain(TCS34725_GAIN_60X, L);	
+	
+	lua_rawgeti(L, LUA_REGISTRYINDEX, cb_tcs_en); // Get the callback to call
+	luaL_unref(L, LUA_REGISTRYINDEX, cb_tcs_en); // Unregister the callback to avoid leak
+	lua_rawgeti(L, LUA_REGISTRYINDEX, self_ref);
```
devsaurus: Remove `self_ref` as proposed below.
If you decide to keep `self_ref`, then tell `lua_call()` that there are **2** items available on the stack.

app/modules/tcs34725.c
```diff
@@ -0,0 +1,359 @@
...
+	uint8_t high = tcs34725Read8(++reg);
+	
+	return (high << 8) | low;
+}
+/**************************************************************************/
+/*!
+		@brief	Finishes enabling the device
+*/
+/**************************************************************************/
+uint8_t tcs34725EnableDone(lua_State* L)
```
devsaurus: Don't use cached Lua state in a callback function.
Best practice is to get a Lua state with
```c
lua_State *L = lua_getstate()
```

Refer to https://github.com/nodemcu/nodemcu-firmware/issues/846#issuecomment-164651561 for details.

devsaurus:
Marcel, the merge-window before next master drop is still open, right?
Just double checking in case I missed something and mess up things when merging this asap.

Issue #1862, opened 2017-03-12
Add a "How To Debug" chapter to Extension Developer FAQ


marcelstoer:
We currently have [module documentation for `gdbstub`](http://nodemcu.readthedocs.io/en/latest/en/modules/gdbstub/). It's quite a bit more than a simple API documentation but not enough to qualify as a real "How to". IMO it would be helpful to have recipe-style instructions in the [extension developer FAQ](http://nodemcu.readthedocs.io/en/latest/en/extn-developer-faq/)s for people who have never worked with GDB. Such a chapter would also prove to be useful in bug processing. We could send people there with a clear conscience and ask them to do some of the debugging themselves.

TerryE:
Marcel, one thing that I keep asking myself is:  why we don't implement the standard Lua debug library?  It is really quit useful in bug hunting end-user Lua apps.

As to the **gdb** stub, I personally feel that we should assume that C developers will already be familiar with **gdb** (and if they aren't, then there are already a lot of useful tutorials out there).  What we could do with is a more focused explanation of how to do host / target debugging using the **gddbstub** and its strengths and limitations.

marcelstoer:
Nice feedback Terry, thanks.

> why we don't implement the standard Lua debug library? It is really quit useful in bug hunting end-user Lua apps.

I'm sure we wouldn't reject a PR if someone provided an implementation, would we?

> As to the gdb stub, I personally feel that we should assume that C developers will already be familiar with gdb

I couldn't tell. What I do know though is that in 17+ years of Java development I never interacted with the Java debugger directly - the complexity is all nicely tucked away in the IDEs.

> and if they aren't, then there are already a lot of useful tutorials out there

That must be true but nothing stops us from adding a reference to your favourite tutorial.

> What we could do with is a more focused explanation of how to do host / target debugging using the gddbstub and its strengths and limitations.

That's essentially what I had in mind. Philip's documentation contains aspects of that already.

I created this isssue here because I personally would profit from such an FAQ chapter. _Reading_ and _debugging_ someone else's code in a language that you are not/no longer/not yet familiar with is the first step into understanding. That's my status with this firmware code.

TerryE:
@marcelstoer @jmattsson @pjsg after my experiences on debugging this #2068, I have decided that we definitely need this.  We also need a more developer friendly way of debugging and testing C library code.

[...]

Issue #1861, opened 2017-03-12
Add a sleep/dsleep chapter to Lua Developer FAQ


marcelstoer:
@TerryE's FAQ has a nice chapter about [techniques for reducing RAM and SPIFFS footprint](http://nodemcu.readthedocs.io/en/latest/en/lua-developer-faq/#techniques-for-reducing-ram-and-spiffs-footprint). I believe it'd be very useful to add a related chapter about "How to reduce power consumption and how to sleep/dsleep".

We have documentation on [`node.sleep`](http://nodemcu.readthedocs.io/en/latest/en/modules/node/#nodedsleep), [`wifi.nullmodesleep`](http://nodemcu.readthedocs.io/en/latest/en/modules/wifi/#wifinullmodesleep), [`rtctime.dsleep`](http://nodemcu.readthedocs.io/en/latest/en/modules/rtctime/#rtctimedsleep), and [`rtcfio.sleep_until_sample`](http://nodemcu.readthedocs.io/en/latest/en/modules/rtcfifo/#rtcfifodsleep_until_sample). What we're missing is the bigger picture. When to use which? Which function is intended for which use case? What is the best strategy for (quasi-)autonomous, battery/solar-powered devices? What are the gotchas? It'd also be nice to document the required wiring graphically (linking to `node.sleep`).

@eyaleb or @jmattsson would you have time to put together a draft (assuming you agree with my request)?

jmattsson:
Good idea, but at the moment I'm still somewhat overloaded :/  While I've started to catch up on things a wee bit, I've still got something like 250 outstanding github notifications to make my way through.

TerryE:
I've just finished doing the plumbing and central heating system for my new house.  And then I want to crack on with the Lua 5.3 implementation which kinda got stalled with the house work.  Sorry.  In terms of the FAQ, there's bits that really need updating like once we've got the early memory leaks out of the way, using UpVals is solid, fast and efficient, but there are gotcha's which need explained for the beginners.

So what I am saying is that this is a great idea, but not one that I've got time for at the moment.

flip111:
@TerryE i was going to open a new issue, but i saw you mentioning lua 5.3. Is Lua 5.3 support in the works? That be pretty sweet, i just checked the lua homepage and 5.1.* seems pretty old.

TerryE:
@flip111, see #1661 for this discussion.

TerryE:
[...]

PR #1899, opened 2017-04-03
Version 2.0 of the Lua Develoer FAQ

### LINKED ISSUE #937: Rebaselining the nodeMCU Lua official FAQ
If we are going to move the FAQ into the distro, then we can hardly call it unofficial (see #449 for previous issue).  This is to discuss the necessary changes to prepare the FAQ for moving into the official document set.

TerryE:
Fixes #937.

This is a major rewrite of the Lua Developer FAQ to bring it up to date with the current state of development.

This type of FAQ is always going to tend to be the views of one author and in the case me.  The commnets I would be really interested in are:

-  Any typo or grammar fixes.
-  Points where you think that I am factually incorrect or have missed some emphasis.
-  Other Qs which NodeMCU application developers really need addressing / answering.

Note that the MD format is pretty diff hostile for this level of change, so you might just want to take a view on the [new document](/TerryE/nodemcu-firmware/blob/b094e382ee23e8fa74a6c8d4a129c1ef14905c8b/docs/en/lua-developer-faq.md) as a whole.


marcelstoer:
> Other Qs which NodeMCU application developers really need addressing / answering.

#1861 and #1862. I wasn't going to mention them but since you explicitly asked for such feedback I saw no point in withholding them.

docs/en/lua-developer-faq.md
```diff
@@ -1,12 +1,28 @@
...
- 
 ## What is this FAQ for?
 
-This FAQ does not aim to help you to learn to program or even how to program in Lua.  There are plenty of resources on the Internet for this, some of which are listed in [Where to start](#where-to-start).  What this FAQ does is to answer some of the common questions that a competent Lua developer would ask in learning how to develop Lua applications for the ESP8266 based boards running the [NodeMcu](http://NodeMCU.com/index_en.html) firmware.
+This FAQ does not aim to help you to learn to program or even how to program in Lua.  There are plenty of resources on the Internet for this, some of which are listed in [Where to start](#where-to-start).  What this FAQ does is to answer some of the common questions that a competent Lua developer would ask in learning how to develop Lua applications for the ESP8266 based boards running the [NodeMcu firmware](https://github.com/nodemcu/nodemcu-firmware).  This includes the NodeMCU Devkits.  However, the scope of the firmware is far wider than this as it can be used on any ESP8266 module.
+
+## What has changed since the first version?
```
marcelstoer: As I said in the earlier discussion I still think that this whole chapter doesn't belong here (I suggested README,
https://www.facebook.com/NodeMCU or even https://en.wikipedia.org/wiki/NodeMCU instead).

docs/en/lua-developer-faq.md
```diff
@@ -311,35 +378,16 @@ Note that if you use `require("XXX")` to load your code then this will automatic
...
-* The NodeMCU development team recommends that you consider using a tailored firmware build, which only includes the modules that you plan to use before developing any Lua application. Once you have the ability to make and flash custom builds, the you also have the option of moving time sensitive or logic intensive code into your own custom module. Doing this can save a large amount of RAM as C code can be run directly from Flash memory. If you want an easy-to-use intermediate option then why note try the [cloud based NodeMCU custom build service](https://nodemcu-build.com)?. 
+### How to reduce the size of the firmware?
+
+* We recommend that you consider using a tailored firmware build; one which only includes the modules that you plan to use in developing any Lua application. Once you have the ability to make and flash custom builds, the you also have the option of moving time sensitive or logic intensive code into your own custom module. Doing this can save a large amount of RAM as C code can be run directly from Flash memory. If you want an easy-to-use intermediate option then why note try the [cloud based NodeMCU custom build service](https://nodemcu-build.com).
```
marcelstoer: > We recommend that you consider...

Users don't really have a choice now 😉 On the first mention of the term `build` you could link to `./build.md`.

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
+*  The ESP8266 runs the SDK over the native hardware, so there is no underlying operating system to capture errors and to provide graceful failure modes.  Hence system or application errors can easily "PANIC" the system causing it to reboot. Error handling has been kept simple to save on the limited code space, and this exacerbates this tendency. Running out of a system resource such as RAM will invariably cause a messy failure and system reboot.
+*  Note that in the 3 years since the firmware was first developed, Espressif has developed and released a new RTOS alternative to the non-OS SDK, and and the latest version of the SDK API reference recommends using RTOS.  Unfortunately, the richer RTOS has a significantly larger RAM footprint.  Whilst our port to the ESP-32 (with its significantly larger RAM) uses the ESP-IDF which is based on RTOS, the ESP8266 RTOS versions don't have enough free RAM for a RTOS-based NodeMCU firmware build to have sufficient free RAM to write usable applications.
```
marcelstoer: You could link to https://espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf#page=16 here.

marcelstoer: ... and https://github.com/espressif/esp-idf

docs/en/lua-developer-faq.md
```diff
@@ -17,23 +33,23 @@ The NodeMCU firmware implements Lua 5.1 over the Espressif SDK for its ESP8266 S
...
+The NodeMCU Lua firmware is an ESP8266 application and is therefore be layered over the ESP8266 SDK. However, the hooks and features of Lua enable it to be seamlessly integrated without losing any of the standard Lua language features.  The firmware has replaced some standard Lua modules that don't align well with the SDK structure with ESP8266-specific versions.  For example, the standard  `io` and `os` libraries don't work, but have been largely replaced by the NodeMCU `node` and `file` libraries.  The `debug` and `math` libraries have also been omitted to reduce the runtime footprint (`modulo` can be done via `%`, `power` via `^`).
```
marcelstoer: >  is therefore be layered over the ESP8266 SDK

Did you mean to write " is therefore _to_ be layered "?

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
-* The LTR implementation means that you can't easily extend standard libraries as you can in normal Lua, so for example an attempt to define `function table.pack()` will cause a runtime error because you can't write to the global `table`. (Yes, there are standard sand-boxing techniques to achieve the same effect by using metatable based inheritance, but if you try to use this type of approach within a real application, then you will find that you run out of RAM before you implement anything useful.) 
+*  The ESP8266 uses a combination of on-chip RAM and off-chip Flash memory connected using a dedicated SPI interface.  Code can be executed directly from Flash (the ESP hardware first faults the Flash code into an L1 cache in RAM, but this largely transparent in terms of programming ESP8266 applications).  The Lua firmware largely runs out of Flash.  Both the RAM and the Flash memory are *very* limited when compared to systems than most application programmers use.
```
marcelstoer: > to systems than most application programmers use

"that" instead of "than"?

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
+*  Over the last two years, both the Espressif non-OS SDK developers and the NodeMCU team have made a range of improvements and optimisations to increase the amount of RAM available to to developers, from a typical 15Kb or so with Version 0.9 builds to some 45Kb with the current firmware Version 2.x builds.
+*  The early ESP8266 modules were typically configured with 512Kb Flash.  Fitting a fully featured Lua build with a number of optional libraries and still enough usable Flash to hold a Lua application was a struggle.  However the code-size of the SDK has grown significantly between the early versions and the current 2.0 version.  Applications based on the current SDK can no longer fit in 512Kb Flash memory, and so all currently produced ESP modules now contain a minimum of 1Mb with 4 and 16Mb becoming more common. The current NodeMCU firmware will fit comfortably in a 1Mb Flash and still have ample remaining Flash memory to support Lua IoT applications.
```
marcelstoer: You could mention here that we froze the [`1.5.4.1-final` branch](https://github.com/nodemcu/nodemcu-firmware/tree/1.5.4.1-final) to still support those 512Kb modules. Even though it was introduced less than a month ago the stats of my build service indicate that there actually _is_ demand for it.

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
-* Non-Lua processing (e.g. network functions) will usually only take place once the current Lua chunk has completed execution.  So any network calls should be viewed at an asynchronous request. A common coding mistake is to assume that they are synchronous, that is if two `socket:send()` are on consecutive lines in a Lua programme, then the first has completed by the time the second is executed.  This is wrong.  Each `socket:send()` request simply queues the send operation for dispatch.  Neither will start to process until the Lua code has return to is calling C function. Stacking up such requests in a single Lua task function burns scarce RAM and can trigger a PANIC.  This true for timer, network, and other callbacks.  It is even the case for actions such as requesting a system restart, as can be seen by the following example:
+* The runtime system runs in interactive-mode.  In this mode it first executes any `init.lua` script.  It then "listens" to the serial port for input Lua chunks, and executes them once syntactically complete.
```
marcelstoer: Hyperlink-maniac speaking again 😉 How about linking to `upload.md#initlua`?

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
+* Non-Lua processing (e.g. network functions) will usually only take place once the current Lua chunk has completed execution.  So any network calls should be viewed at an asynchronous request. A common coding mistake is to assume that they are synchronous, that is if two `socket:send()` are on consecutive lines in a Lua programme, then the first has completed by the time the second is executed.  This is wrong. A `socket:send()` request simply queues the send task for dispatch by the SDK.  This task can't start to process until the Lua code has returned to is calling C function to allow this running task to exit. Stacking up such requests in a single Lua task function burns scarce RAM and can trigger a PANIC.  This true for timer, network, and other callbacks.  It is even the case for actions such as requesting a system restart, as can be seen by the following example:
 
 ```lua
 node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
```
marcelstoer: To drive the point home it'd IMO be helpful to add a (stripped) sample output.

docs/en/lua-developer-faq.md
TerryE: Marcel, I agree that I could replace this with a reference, but IMO the source needs to be _wthin_ our documentation set.  Perhaps you could consider reviewing our history and adding this?

-   I don't know who owns the NodeMCU project Facebook page, but Facebook is closed resource and many of us may not even have a Facebook about (me for one) and really object to getting bonbarded to "sign up to Facebook messages".
-   I am a regular Wkipedia contributor, and this is not how WP works.  Whilst articles can include a history section, as a contributor you are still not allowed to add primary research to a WP page. So for example, if a history page is included on this site and the content is both non-falsifiable and therefore non-controversial,  then a WP contributor might be able to update the History by quoting this reference as a source.

So my proposal is that you draft a more general "changes in NodeMCU section for the documentation and I reference this here.  We've all done so much over the last two years, that I feel that we should be open and explicit about what we've achieve.

docs/en/lua-developer-faq.md
TerryE: Yup, since the FAQ is written in a generally personal voice, the shift to first person works; however,   the grammar still isn't quite right here, so I'll update accordingly and reference the build section.  

One of the thing that has changes since V1 is that it's primary host was esp8266.com and decent documentation didn't exist.  I have tried to integrate intra-documentation references and avoid too much duplication when practical.

docs/en/lua-developer-faq.md
TerryE: Thanks, I will add the links.

docs/en/lua-developer-faq.md
TerryE: Sorry, I am a sloppy writer (my fingers don't fully obey my thinking :smiley:\)and my stuff always needs proofing.  What I am trying to say is the that in some sense the NodeMCU firmware is just another NodeMCU application so the phrase should read "and must therefore be layered over the ESP8266 SDK."

docs/en/lua-developer-faq.md
TerryE: Thanks.

docs/en/lua-developer-faq.md
TerryE: Yes, but the target link needs improving IMO -- however, that is a separate issue.

TerryE:
Marcel, thanks for your feedback.

docs/en/lua-developer-faq.md
marcelstoer: You have a point here - more than one actually.

> I don't know who owns the NodeMCU project Facebook page

"We" do. I don't know who else has write access to it but at least the originators and I do.

> but Facebook is closed resource and many of us may not even have a Facebook account 

True, I keep forgetting that (what a shame it is closed btw). I don't have a personal account either and I personally couldn't care less about FB but I just accept that it's a useful medium for a project like ours.

You're right about WP as well.

I'll be happy to create a "Changes in NodeMCU" section and transfer your chapter to it. It'll be  on the [README](https://github.com/nodemcu/nodemcu-firmware/blob/dev/README.md), though. The space between the "Releases" and "Support" sections looks appropriate to me.

marcelstoer:
There's more to come, I'm only half-way through 😉

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
-* Non-Lua processing (e.g. network functions) will usually only take place once the current Lua chunk has completed execution.  So any network calls should be viewed at an asynchronous request. A common coding mistake is to assume that they are synchronous, that is if two `socket:send()` are on consecutive lines in a Lua programme, then the first has completed by the time the second is executed.  This is wrong.  Each `socket:send()` request simply queues the send operation for dispatch.  Neither will start to process until the Lua code has return to is calling C function. Stacking up such requests in a single Lua task function burns scarce RAM and can trigger a PANIC.  This true for timer, network, and other callbacks.  It is even the case for actions such as requesting a system restart, as can be seen by the following example:
```
vsky279: This true for timer -> This is true for timer

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
+
+*  If you are running out of memory, then you might not be correctly clearing down Registry entries.  One example is as above where you are setting up timers but not unregistering them.  Another occurs in the following code fragment. The `on()` function passes the socket to the connection callback as it's first argument `sck`.  This is local variable in the callback function, and it also references the same socket as the upvalue `srv`.  So functionally `srv` and `sck` are interchangeable.  So why pass it as an argument?  Normally garbage collecting a socked will automatically unregister any of its callbacks, but if you use a socket as an upvalue in the callback, the socket is now referenced through the Register, and now it won't be GCed because it is referenced.  Catch-22 and a programming error, not a bug.
```
vsky279: socked will -> socket will

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
+* It is important to understand that a single Lua function is associated with / bound to any event callback task. This function is executed from within the relevant NodeMCU library C code using a `lua_call()`. Even system initialisation which executes the `dofile("init.lua")` is really a special case of this.  Each function can invoke other functions and so on, but it must ultimately return control to the C library code which then returns control the SDK, terminating the task. 
+* By their very nature Lua `local` variables only exist within the context of an executing Lua function, and so locals are unreferenced on exist and any local data (unless also a reference type such as a function, table, or user data which is also referenced elsewhere) can therefore be garbage collected between these `lua_call()` actions.
```
vsky279: unreferenced on exist -> unreferenced on exit

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
 
 ### Why is it importance to understand how upvalues are implemented when programming for the ESP8266?
 
-Routines directly or indirectly referenced in the globals table, **_G**, or in the Lua Registry may use upvalues. The number of upvalues associated with a given routine is determined by the compiler and a vector is allocated when the closure is bound to hold these references.  Each upvalues is classed as open or closed. All upvalues are initially open which means that the upvalue references back to the outer functions's register set.  However, upvalues must be able to outlive the scope of the outer routine where they are declared as a local variable.  The runtime VM does this by adding extra checks when executing a function return to scan any defined closures within its scope for back references and allocate memory to hold the upvalue and points the upvalue's reference to this.  This is known as a closed upvalue. 
+The use of upvalues is a core Lua feature, that are explain in detail in PiL.  Any Lua routines defined within an outer scope my use them.  This can include routines directly or indirectly referenced in the globals table, **_G**, or in the Lua Registry.
```
vsky279: that are explain -> that are explained

docs/en/lua-developer-faq.md
devsaurus: > but this *is* largely transparent in terms of programming ESP8266 applications

docs/en/lua-developer-faq.md
```diff
@@ -46,15 +62,23 @@ The NodeMCU libraries act as C wrappers around registered Lua callback functions
...
+*  The ESP8266 uses a combination of on-chip RAM and off-chip Flash memory connected using a dedicated SPI interface.  Code can be executed directly from Flash (the ESP hardware first faults the Flash code into an L1 cache in RAM, but this largely transparent in terms of programming ESP8266 applications).  The Lua firmware largely runs out of Flash.  Both the RAM and the Flash memory are *very* limited when compared to systems than most application programmers use.
+*  Over the last two years, both the Espressif non-OS SDK developers and the NodeMCU team have made a range of improvements and optimisations to increase the amount of RAM available to to developers, from a typical 15Kb or so with Version 0.9 builds to some 45Kb with the current firmware Version 2.x builds.
```
devsaurus: "to to"

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
-* The registered callback routines are invoked sequentially with the associated C task running to completion uninterrupted.
-* In the case of Lua, these C tasks are typically functions within the Lua runtime library code and these typically act as C wrappers around the corresponding developer-provided Lua callback functions.  An example here is the Lua [`mytimer:alarm(interval, repeat, callback)`](modules/tmr.md#tmralarm) function.  The calls a function in the `tmr` library which registers a C function for this alarm using the SDK, and when this C function is called it then invokes the Lua callback.  
+*  The SDK uses a small number of Interrupt Service Routines (ISRs) to handle short time critical hardware interrupt related processing. These are very short duration and can interrupt a running task for up to 10µSec.  (Modifying these ISRs or adding new ones is not a viable options for most developers.)
+*  All other service and application processing is split into code execution blocks, known as **tasks**.  The individual are executed one at a time and run to completion. No task can never pre-empt another.
```
devsaurus: > The individual *tasks* are executed

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
-* Excessively long-running Lua functions can therefore cause other system functions and services to timeout, or allocate memory to buffer queued data, which can then trigger either the watchdog timer or memory exhaustion, both of which will ultimately cause the system to reboot.
-* By default, the Lua runtime also 'listens' to UART 0, the serial port, in interactive mode and will execute any Lua commands input through this serial port.
+* The SDK invokes a startup hook within the firmware on boot-up.  This firmware code initialises the Lua environment and then attempts to execute the Lua module `init.lua` from the SPIFFS file system.  This `init.lua`  module can then be used to do any application initialisation required and to call the necessary timer alarms or library calls to bind and callback routines to implement the tasks needed in response to any system events.
+* By default, the Lua runtime also 'listens' to UART 0, the serial port, in interactive mode and will execute any Lua commands input through this serial port.  Using the serial port in this way is the most common method of developing and debugging Lua applications on the ESP8266/
```
devsaurus: Interactive commands are also fed into the task scheme described above. I.e. long running code can delay or even block them and the firmware seems to be unresponsive.

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
+    } do  
+    dev['write'..reg] = function(o, dataByte)
+                          write(MCP23008addr, regAddr, dataByte)
+                        end
+    dev['read'..reg]  = function(o)
+                          return read(MCP23008addr, regAddr)
+                        end
+  end
+``` 
+*  This loop is compiled once when the module is required.   Th opcode vectors for the read and write functions are created during the compile, along with a header which defines how many upvalues and locals are used by each function. However, these two functions are then bound _four_ times as different functions (e.g. `mcp23008.writeIODOR()`) and each closure inherits it's own copies of the upvalues it uses so the `regAddr` for this function is `0x00`). The upvalue list is created when the closure is created and though some Lua magic, even if the outer routine which initially declared then is no longer in scope and is GCed (Garbage Collected), the Lua RTS ensures that upvalue still persist whilst the closure persists.
```
devsaurus: > *The* opcode vectors

> inherits *its* own copies

> and *through* some Lua magic

> initially declared *them* is no longer

> ensures that *upvalues* still persist

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
+end)
+```
+
+*  One way to check the registry is to use the construct `for k,v in pairs(debug.getregistry()) do print (k,v) end` to track the registry size.  If this is growing then you've got a leak.
+
+### How do I track globals
+
+* See the Unofficial LUA FAQ: [Detecting Undefined Variables](http://lua-users.org/wiki/DetectingUndefinedVariables).
+
+* My approach is avoid using then unless I have a *very* good reason to justify this. I track them statically by running a `luac -p -l XXX.lua | grep GLOBAL` filter on any new modules and replace any accidental globals by local or upvalued local declarations.
```
devsaurus: > avoid using *them*

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
+*  One way to check the registry is to use the construct `for k,v in pairs(debug.getregistry()) do print (k,v) end` to track the registry size.  If this is growing then you've got a leak.
+
+### How do I track globals
+
+* See the Unofficial LUA FAQ: [Detecting Undefined Variables](http://lua-users.org/wiki/DetectingUndefinedVariables).
+
+* My approach is avoid using then unless I have a *very* good reason to justify this. I track them statically by running a `luac -p -l XXX.lua | grep GLOBAL` filter on any new modules and replace any accidental globals by local or upvalued local declarations. 
+
+* On NodeMCU, _G's metatable is _G, so you can create any globals that you need and then 'close the barn door' by assigning
+`_G.__newindex=function(g,k,v) error ("attempting to set global "..k.." to "..v) end` and any attempt to create new globals with now throw an error and give you a traceback of where this has happened.
```
devsaurus: > new globals *will* now throw

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
 
-One further complication is that some library functions don't correctly dereference expired callback references and as a result their upvalues may not be correctly garbage collected (though we are tracking this down and hopefully removing this issue).  This will all be manifested as a memory leak. So using upvalues can cause more frequent and difficult to diagnose PANICs during testing. So my general recommendation is still to stick to globals for this specific usecase of passing context between event callbacks, and `nil` them when you have done with them.
+One further complication is that some library functions don't implicitly dereference expired callback references and as a result their upvalues may not be garbage collected and this application error can be be manifested as a memory leak. So using upvalues can cause more frequent and difficult to diagnose PANICs during testing. So my general recommendation is still to stick to globals during initial development, and explicitly dereference resources by setting them to `nil` them when you have done with them.
```
devsaurus: double *them*

docs/en/lua-developer-faq.md
```diff
@@ -179,31 +238,37 @@ status = mail.send(to, subject, body)
...
 
 ### How do I avoid a PANIC loop in init.lua?
 
 Most of us have fallen into the trap of creating an `init.lua` that has a bug in it, which then causes the system to reboot and hence gets stuck in a reboot loop.  If you haven't then you probably will do so at least once.
 
 - When this happens, the only robust solution is to reflash the firmware.
 - The simplest way to avoid having to do this is to keep the `init.lua` as simple as possible -- say configure the wifi and then start your app using a one-time `tmr.alarm()` after a 2-3 sec delay.  This delay is long enough to issue a `file.remove("init.lua")` through the serial port and recover control that way.
-- Also it is always best to test any new `init.lua` by creating it as `init_test.lua`, say, and manually issuing a `dofile("init_test.lua")` through the serial port, and then only rename it when you are certain it is working as you require.
+-  Another trick is to poll a spare GPIO input pin in your startup.  I do this omy boards by taking this GPIO plus Vcc to a jumper on the board, so that I can set the jumper to jump into debug mode or reprovision the software.
```
devsaurus: > I do this *on my* boards

docs/en/lua-developer-faq.md
```diff
@@ -290,8 +357,8 @@ Note that if you use `require("XXX")` to load your code then this will automatic
...
+   * **Upload your `.lc` files to the PC and disassemble then there**. There are a number of Lua code disassemblers which can list off the compiled code that you application modules will generate, `if` you have a script to upload files from your ESP8266 to your development PC.  I use [ChunkSpy](http://luaforge.net/projects/chunkspy/)  which can be downloaded [here](http://files.luaforge.net/releases/chunkspy/chunkspy/ChunkSpy-0.9.8/ChunkSpy-0.9.8.zip) , but you will need to apply the following patch so that ChunkSpy understands eLua data types:
```
devsaurus: > and disassemble *them* there

> the compiled code that *your* application modules will generate

docs/en/lua-developer-faq.md
```diff
@@ -311,35 +378,16 @@ Note that if you use `require("XXX")` to load your code then this will automatic
...
-
-Here bigger means less RAM used.  
-
-Of course you should still use functions to structure your code and encapsulate common repeated processing, but just bear in mind that each function definition has a relatively high overhead for its header record and stack frame (compared to the 20 odd KB RAM available).  *So try to avoid overusing functions. If there are less than a dozen or so lines in the function then you should consider putting this code inline if it makes sense to do so.*
+Functions have fixed overheads, so in general the more that your group your application code into larger function, then  less RAM used will be used overall.  The main caveat here is that if you are starting to do "copy and paste" coding across functions then you are wasting resources.  So of course you should still use functions to structure your code and encapsulate common repeated processing, but just bear in mind that each function definition has a relatively high overhead for its header record and stack frame.  *So try to avoid overusing functions. If there are less than a dozen or so lines in the function then you should consider putting this code inline if it makes sense to do so.*
```
devsaurus: > the more that *you* group your application

> then less RAM will be used overall

docs/en/lua-developer-faq.md
```diff
@@ -64,108 +88,143 @@ node.restart(); for i = 1, 20 do print("not quite yet -- ",i); end
...
+*  Because locals are lexically scoped you can also refer to local variables in an outer scope and these are still accessible within the inner scope. Such variables are know as **upvalues**..
+*  Lua variable can be assigned two broad types of data: **values** such as numbers, booleans, and strings and **references** such as functions, tables and userdata.  You can see the difference here  when you assign the contents of a variable `a` to `b`.  In the case of a value then it is simply copied into `b`.  In the case of a reference, both `a` and `b` now refer to the *same object*, and no copying of content takes place.  This process of referencing can have some counter-intuitive consequences. For example, in the following code by the time it exists, the variable `timer2func` is out of scope.  However a reference to the function has now been stored in the Lua registry by the alarm API call, so it and any upvalues that it uses will persist until it is eventually entirely dereferenced (e.g. by `tmr2:unregister()`.
+```Lua 
+  do
+    local tmr2unc = function() ds.convert_T(true) tmr1:start() end
```
vsky279: local tmr2unc = function() ds.convert_T(true) tmr1:start() end
->
local tmr2func = function() ds.convert_T(true); tmr1:start() end
+ correction of `tmr2unc` on the following line

docs/en/lua-developer-faq.md
vsky279: This example, although very elegant, is a bit complex for what is being explained here.

I would prefer to see something very elementary. The following trivial example helps me understand:

The following two lines of code have the same functionality but the second one consumes less memory (50 bytes less in this case) and is more efficient.
```lua
tmr.create():alarm(1000, tmr.ALARM_AUTO, function() print(node.heap()) end)
```
```lua
local node_heap=node.heap; tmr.create():alarm(1000, tmr.ALARM_AUTO, function() print(node_heap()) end)
```

The first callback function every time it is called needs to search for a member of the `node` table called `heap` (opcode `GETGLOBAL` + `GETTABLE`) but it also needs to store the string "heap" so it knows what to look for.

The second example uses just the upvalue `node_heap`. At the time the timer is printing the amount of free memore it no longer needs to know the string "heap" and does not have to look it up in the table node`. It uses the reference to the function `node.heap()` stored in the upvalue (opcode `GETUPVAL`). As noted it is more efficient and less memory demanding.

TerryE:
Thanks guys for all this feed back.  I will go through the doc in the next couple of days and incorporate this.  

Any Qs or topics that you feel should be in this, that I've missed?

jmattsson:
Poke?

TerryE:
On my TODO list. I'll push out an update this w/e.

joysfera:
Please cover also ```dofile()```. Most information in NodeMCU refer to ```require()``` only and is rather complicated. ```dofile()``` is very straigtforward to use, but it's not clear (to me, at least) how to use it efficiently. Simple example: I moved all DS18B20 handling to a separate file. If I call dofile() on it, will it stay in memory for all the 750+ milliseconds? How to set up the conversion timer to allow freeing memory in the meantime and also afterwards?

Another very typical example: a server. How to implement a memory efficient handling of receive events? Can we call dofile() from within the receive handlers? Is it smart? I guess it's a tradeoff between speed and memory but we generally miss memory, not speed. Perhaps the handlers can contain just dofile() and nothing else? Will it free the memory as soon as possible? What is the impact on speed, anyway?

Last but not least: is it possible to free the memory used by init.lua if one of the last commands is ```dofile('main.lc')``` or will the init.lua code stay in memory forever? If yes then we need to write very short init.lua to save memory.

marcelstoer:
I that `dofile()`/memory discussion you can also refer to/copy/integrate (or challenge 😜) NodeMCU Lua stuff by "the boss": http://nodemcu.a1w.ca/post/150675512189/coding-nodemcu-lua-like-a-boss

TerryE:
>  [NodeMCU Lua stuff by "the boss"] (http://nodemcu.a1w.ca/post/150675512189/coding-nodemcu-lua-like-a-boss)
Actually Marcel, whoever this guy or gal is, he / she is terribly ill informed and giving bad advice, so I don't recommend that we link to it.  The advice seems to be don't use locals or globals (and by implication _any_ variables) or tables or functions, which is all just plain silly.  We could with some targeted advice in the FAQ, but the main thing is that Lua coders should have and understanding of the rules for the life of any variable and how the GC works.  You can write some pretty serious embedded applications and still stay within the RAM constraints of the ESP8266 if you understand these.

marcelstoer:
> don't use locals or globals (and by implication any variables) or tables or functions, which is all just plain silly

I agree, you may reduce the memory footprint of your application but any such code will be terribly hard to read and thus hard to understand and maintain.

docs/en/lua-developer-faq.md
TerryE: Actually the `;` here is syntactic sugar and the code compiles and runs fine without it, but I agree that the code is clearer with the separator.

docs/en/lua-developer-faq.md
TerryE: I agree that using locals will in general generate less code and execute faster, but that is a different point than the issue about locals used as upvalues inside functions will still persist even though the binding function is not longer scoped. 

The second example is complicated but it does underline the basic point that a function is compiled once but can be bound many times at run time.

TerryE:
@joysfera Petr, the GC in Lua is really quite simple:  if a variable or resource is referenced directly or indirectly then it won't be GCed.  Conversely if it isn't then it can be.  Doing a `collectgarbage()` will do just that: collect any resources eligible for collection (because of the way that it works, this is a two-pass process so the sweep must take place twice to collect all eligible resources).

Tables, userdata, function closures, strings are all treated as collectable resources.  So if they are referenced then the won't be collected and if they aren't then they can be.  The only reason that a required module sticks around is that the `require()` function sticks a reference to the module in the table `package.module`, so if you set `package.module.mymodule = nil` **and** you haven't got any other references to it then it can be collected.  Conversely if you do something like `_G.myfunc = loadfile("myfunc.lua")` then it won't be.

TerryE:
I've just pushed a version with these changes in it.  Unless anyone can see  any show-stoppers, then I suggest that we merge this.  It's not perfect, but it never will be.  I would really prefer something more fluid that we could collaborate on, such as a wiki pageset and which we occasionally snap to the readthedocs reference, but I've already lost that one :smile:

joysfera:
I have just a few global variables and functions, all the rest is local, most of the stuff is loaded on demand only yet I have less than 23 kB free. This is not enough for TLS, unfortunately. That's why I am looking for any advice how to write code that keeps as much memory free as possible.
Maybe the TLS problem could be attacked from a different angle: the mongoose-os.com says that they somehow optimized the TLS footprint so that each connection takes just 2 kB of memory...

marcelstoer:
> Unless anyone can see any show-stoppers, then I suggest that we merge this.

Not that anything is wrong but [you promised](https://github.com/nodemcu/nodemcu-firmware/issues/1861#issuecomment-291894112) to include #1861 😜 Sorry for being such a pest.

TerryE:
Marcel, I want to print this off and consider again.  I asked you guys if you had any other FAQs, and I've considered a list. Some are very ESP-specific and some are really general Lua, but not really something that you worry too much about except when you are running in a tight RAM configuration as in the ESP8266.

-  Discussion of the sleep modes and how you use them within a Lua application.
-  How can you find out what (RAM) resources are tied up in your running application and what you can  do to limit these.
-  How does the GC work?
-  How do I override ROM-based resources, such as extending `string` or `table`? 

The problem is that there's a lot of ground to cover here and a lot is covered in other references that I call out, but people don't seem to read them.  maybe we just need to break this up more.  

Need to think, but let's get this out for new and  move this sort of stuff into a separate issue/PR.

Issue #1131, opened 2016-03-07
Add optional callbacks to wifi.sta.connect() and wifi.sta.disconnect()


alex3kov:
I'm aware of `wifi.sta.eventMonReg()`. I'm thinking about writing several different functions in one program. Each will connect wifi at the start and disconnect at the end. They will do different things (obviously) and `wifi.sta.eventMonReg()` can only hold one function. So I would have to register/unregister `eventMonReg` each time and it wouldn't be **quite** deterministic due to asynchronicity:

```
function test1()
  wifi.sta.eventMonReg(wifi.STA_GOTIP, function()
    --do things here
    wifi.sta.eventMonReg(wifi.STA_GOTIP, "unreg")
    wifi.disconnect()
  end)
  wifi.connect()
end
```

...what if wifi manages to connect before event monitor gets registered? I realize it's unlikely, but who knows - mega-fast WiFi router with DHCP already aware of ESP's MAC, some random delay in ESP to register event monitor and voila, `eventMonReg`'s callback won't fire. I think this would look better, be more EDD(Event-Driven Development)-like and deterministic:

```
function test2()
  wifi.connect(function()
    --do things here
    wifi.disconnect()
  end)
end
```

TerryE:
Alex, that's the second or 3rd time you've referred to EDD.  Can you please explain for poor old farts who have been in this business for 40 years but totally missed this acronym?  Good old google doesn't help.

The second thing is that this is mainly a wrapper around the SDK API call `wifi_station_connect()` which doesn't have any callback so you have to poll `wifi_station_get_connect_status()` either at a Lua level or in a C wrapper.  But "at 20,000ft" you are quite right programming connections is a mess from an applications programmer's PoV.

Not sure at this stage what we do about it.

dnc40085:
@TerryE I had to look it up myself and I'm pretty sure EDD means Event Driven Development.

TerryE:
OK figures. Alex is right. Developing a nodeMCU Lua app is more like developing a jQuery or DB triggered app. Giving a more rigorous temporal dependency model would be better.

dnc40085:
@TerryE I don't believe it would be that complicated to add the functionality described here after the event monitor in PR #1018 is merged, no polling of `wifi_station_get_connect_status()` would be necessary.

pjsg:
[...]

PR #1903, opened 2017-04-05
Add event callbacks to wifi.sta.config() and wifi.ap.config() and more

dnc40085:
Add feature requested in issue #1131.

This PR adds the functionality requested in issue #1131, add optional callbacks to `wifi.sta.connect()` and `wifi.sta.disconnect()`. The option to register event callbacks as part of the Station/SoftAP configuration process has also been implemented.

### DIFF
--- a/docs/en/modules/wifi.md
+++ b/docs/en/modules/wifi.md
@@ -401,7 +401,8 @@ Sets the WiFi station configuration.
 	- `save` Save station configuration to flash. 
 		- `true` configuration **will** be retained through power cycle. 
 		- `false` configuration **will not** be retained through power cycle. (Default).
-	- Event callbacks will only be available if `WIFI_SDK_EVENT_MONITOR_ENABLE` is enabled in `user_config.h`
+	- Event callbacks will only be available if `WIFI_SDK_EVENT_MONITOR_ENABLE` is uncommented in `user_config.h`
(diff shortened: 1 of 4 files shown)


docs/en/modules/wifi.md
```diff
@@ -1143,7 +1144,8 @@ Sets SSID and password in AP mode. Be sure to make the password at least 8 chara
 	- `save` save configuration to flash.
 		- `true` configuration **will** be retained through power cycle. (Default)
 		- `false` configuration **will not** be retained through power cycle.
-	- Event callbacks will only be available if `WIFI_SDK_EVENT_MONITOR_ENABLE` is enabled in `user_config.h`
+	- Event callbacks will only be available if `WIFI_SDK_EVENT_MONITOR_ENABLE` is uncommented in `user_config.h`
+		- Please note: To ensure all SoftAP events are handled at boot time, all relevant callbacks must be registered as early as possible in `init.lua` with either `wifi.ap.config()` or `wifi.eventmon.register()`.
```
TerryE: One comment about this. The way this reads, the cb's are optional, which they are technically, but I feel that we should add a caveat to the effect that:

Application developers​ should not assume that the WiFi is connected at any time during initialization start-up.  This should be validated either by using a wifi event callback or of polling the status on a timer.

Or words to that effect.

PR #1913, opened 2017-04-17
Move the documentation for wifi.nullmodesleep() to the correct spot

dnc40085:

As I was checking some recent changes in `docs/en/wifi.md`, I noticed that the documentation for `wifi.nullmodesleep()` was in the incorrect spot, so I moved it.


marcelstoer:
Thank you very much for such dedicated proof-reading.

PR #1916, opened 2017-04-18
Update ws2812.md

flip111:
update link to telnet example


marcelstoer:
Thanks for spotting but please note that PRs must be created against `dev` : https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#working-with-git-and-github

PR #1927, opened 2017-04-22
ADC extension for ESP32

zelll:

Supports ADC1 in ESP32.

### DIFF
--- a/components/modules/Kconfig
+++ b/components/modules/Kconfig
@@ -47,6 +47,13 @@ depends on LUA_BUILTIN_DEBUG
 endmenu
 
 
+config LUA_MODULE_ADC
+  bool "ADC module"
+  default "n"
+  help
+      Includes the adc module. This module provides access to the
+      adc1 hardware.
+
 config LUA_MODULE_BIT
   bool "Bit module"
   default "n"
--- a/components/modules/adc.c
+++ b/components/modules/adc.c
@@ -1,0 +1,69 @@
+// Module for interfacing with adc hardware
+
+#include "module.h"
+#include "lauxlib.h"
+#include "platform.h"
+
+
+// Lua: config( bits )
+static int adc_set_width( lua_State *L )
+{
+  int bits = luaL_checkinteger( L, 1 );
+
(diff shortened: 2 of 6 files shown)


devsaurus:
Looks good, thank you very much!

Once Espressif enables ADC2 support, how will it be integrated?
Eg by separate functions in `adc` module or an extra `id` parameter?

zelll:
ADC2 can not work with WiFi/BT, so personally I don't think it's important for most users. But it would be great if someone wants many ADC channels.

Maybe, like this?
```
adc.setwidth(bits, adc = 2)
adc.read(8) , adc.read(17)
```

devsaurus:
Unfortunately, the [Technical Reference Manual](http://espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf) doesn't provide any details on ADC1 and ADC2, so it's kind of guessing how the API for ADC2 will look like.
Since the current IDF has dedicated functions like `adc1_config_width()`, I'd expect that there will also be dedicated ones for ADC2 like `adc2_config_width()`. This could be handled with a parameter for each Lua function:
```lua
adc.setwidth(adc.ADC1, 8)
adc.setwidth(adc.ADC2, 12)

print(adc.read(adc.ADC1, 0)
print(adc.read(adc.ADC2, 1)
```

But that's guessing, what is your view?

zelll:
Yes, there should be dedicated functions for ADC2 in IDF. https://github.com/espressif/esp-idf/pull/466 does in this way.

I thought we could add a second parameter for ADC2 ` adc.setwidth(channel [, adc = adc.ADC1]) `, but it maybe better to add it as 1st parameter now ` adc.setwidth(adc.ADCn, channel) `...

devsaurus:
Thanks for the link, I wasn't aware of this PR.

Regarding the instance id as the first parameter, it's just by convention that many other existing modules already do it this way.

zelll:
Added adc.ADC1 as 1st parameter and updated docs.

devsaurus:
Thanks again for sharing!

PR #1953, opened 2017-05-04
Update init.lua example to use new wifi.sta.config() implementation

### LINKED ISSUE #1951: init.lua uses old WiFi code
When @dnc40085 updated the `wifi.sta.config()` implementation to use the table notation we forgot to update https://nodemcu.readthedocs.io/en/latest/en/upload/#initlua accordingly.

dnc40085:
Fixes #1951 .

When I updated the function `wifi.sta.config()` to accept a table for station configuration, I forgot to update the `init.lua` example in `docs/en/upload.md` accordingly.

The new version of the example also makes use of the event monitor rather than a timer to catch station events.

### DIFF
--- a/docs/en/upload.md
+++ b/docs/en/upload.md
@@ -59,7 +59,7 @@ end
 -- Define WiFi station event callbacks 
 wifi_connect_event = function(T) 
   print("Connection to AP("..T.SSID..") established!")
-  print("Wating for ip...") 
+  print("Waiting for ip...") 
 end
 
 wifi_got_ip_event = function(T) 
@@ -70,6 +70,10 @@ wifi_got_ip_event = function(T)
 end
 
 wifi_disconnect_event = function(T)
+  if T.reason == wifi.eventmon.reason.ASSOC_LEAVE then 
+    --the station has disassociated from a previously connected AP
+    return 
+  end
   local retry_ct = 3
   print("WiFi connection to AP("..T.SSID..") has failed!")
   print("Disconnect reason:"..T.reason)
--- a/docs/en/upload.md
+++ b/docs/en/upload.md
@@ -59,10 +59,13 @@ end
 -- Define WiFi station event callbacks 
 wifi_connect_event = function(T) 
   print("Connection to AP("..T.SSID..") established!")
-  print("Waiting for ip...") 
+  print("Waiting for IP address...")
+  if disconnect_ct ~= nil then disconnect_ct = nil end  
 end
 
 wifi_got_ip_event = function(T) 
+  -- Note: Having an ip address does not mean there is internet access!
+  -- Internet connectivity can be determined with net.dns.resolve().    
   print("Wifi connection is ready! IP address is: "..T.IP)
   print("Startup will resume momentarily, you have 3 seconds to abort.")
   print("Waiting...") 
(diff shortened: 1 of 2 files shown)


docs/en/upload.md
```diff
@@ -56,21 +56,46 @@ function startup()
     end
 end
 
+-- Define WiFi station event callbacks 
+wifi_connect_event = function(T) 
+  print("Connection to AP("..T.SSID..") established!")
+  print("Wating for ip...")
```
jmattsson: Typo. Should be `Waiting`.

jmattsson:
👍 in general

marcelstoer:
Woah, a really elaborate template, thanks. Did you consider to drop the "Inspired by " note at the end (it has deviated quite a bit)? I'll some notes/questions inline. 

@jmattsson what does your cautious "in general" mean? I interpret this as if you were not fully satisfied with the proposal?

docs/en/upload.md
```diff
@@ -56,21 +56,63 @@ function startup()
     end
 end
 
+-- Define WiFi station event callbacks 
+wifi_connect_event = function(T) 
+  print("Connection to AP("..T.SSID..") established!")
+  print("Waiting for ip...")
```
marcelstoer: How about "Waiting for IP address..."?

docs/en/upload.md
```diff
@@ -56,21 +56,63 @@ function startup()
...
+  local total_tries = 3
+  local reason_string=""
+  print("\nWiFi connection to AP("..T.SSID..") has failed!")
+
+  --There are many possible disconnect reasons, the following iterates through 
+  --the list and returns the string corresponding to the disconnect reason.
+  for key,val in pairs(wifi.eventmon.reason) do
+    if val == T.reason then
+      print("Disconnect reason:"..T.reason.."("..key..")")
+      reason_string=key
```
marcelstoer: Why do you keep `reason_string`? I don't see it being used anywhere...

docs/en/upload.md
```diff
@@ -56,21 +56,63 @@ function startup()
...
+  -- total_tries: how many times the station will attempt to connect to the AP.
+  local total_tries = 3
+  local reason_string=""
+  print("\nWiFi connection to AP("..T.SSID..") has failed!")
+
+  --There are many possible disconnect reasons, the following iterates through 
+  --the list and returns the string corresponding to the disconnect reason.
+  for key,val in pairs(wifi.eventmon.reason) do
+    if val == T.reason then
+      print("Disconnect reason:"..T.reason.."("..key..")")
```
marcelstoer: How about `"Disconnect reason: "..val.." ("..key..")"`?

docs/en/upload.md
```diff
@@ -56,21 +56,63 @@ function startup()
...
+  tmr.create():alarm(3000, tmr.ALARM_SINGLE, startup)
+end
+
+wifi_disconnect_event = function(T)
+  if T.reason == wifi.eventmon.reason.ASSOC_LEAVE then 
+    --the station has disassociated from a previously connected AP
+    return 
+  end
+  -- total_tries: how many times the station will attempt to connect to the AP.
+  local total_tries = 3
```
marcelstoer: As a default value I would increase this significantly (or drop it altogether) to accommodate for intermittent  AP reboots.

jmattsson:
Oh, just the typo at the time :)
No objections from me, merge whenever you're satisfied!

docs/en/upload.md
dnc40085: I was going to add the disconnect reason to the final abort message, but I decided not to and I forgot to remove this variable.

docs/en/upload.md
dnc40085: does 30 sound good?

marcelstoer: It does but I haven't tested how this translates to _time_. Would it be safe to assume that APs take anywhere between 30-60s to boot?

dnc40085: My Comcast xfinity gateway(Arris technicolor tc8305c) takes about 2.5 min to reboot and the interval between `NO_AP_FOUND` disconnect events is about 2.25 seconds, so I think a value of 75 should cover it.

PR #1961, opened 2017-05-07
Documenting creating integer build

### LINKED ISSUE #1957: Issue with integer management in runtime
### Expected behavior
If the 31st bit is set, it shouldn't cause the variable to become 0XFFFFFFFFF

### Actual behavior
See test code, just comment out either first or second line to feed value v differently. First line works fine without the 31st bit set, second line results in the value being 0xFFFFFFFF (incorrectly).

Interestingly the second invocation to bit.set(v,31) works fine indicating the problem is with the initial parsing of the value (still into the function) and not the internal checkInteger function itself (although I might be wrong). I have tested this with both bit and rfswitch (where I found it) modules, so it's not the module.

### Test code
```Lua
--v=0x7555AAAA  -- WORKS
v=0xF555AAAA  -- FAILS and trashes value with 0xFFFFFFFF

print(string.format("%X\n", v)) -- Works in both cases
v=bit.set(v,31)
print(string.format("%X\n", v)) -- Fails in v=F555AAAA case, ok in v=7555AAAA
v=bit.set(v,31) 
print(string.format("%X\n", v))  -- Second invocation in v=7555AAAA case is fine (it was set F555AAAA in prior bit.set() call)
```
### NodeMCU version
master, 22e1adc4b06c931797539b986c85e229e5942a5f

### Hardware
NodeMCU, ESP8266, 4MB flash

FrankX0:
Fixes #1957 .

Documenting how to create an integer build.

### DIFF
--- a/docs/en/build.md
+++ b/docs/en/build.md
@@ -57,11 +57,18 @@ the firmware to lock onto that baud rate (between 1200 and 230400).
 
 ### Integer build
 By default a build will be generated supporting floating-point variables.
-To reduce memory size an integer build can be created.  You can change this by
-uncommenting `LUA_NUMBER_INTEGRAL` in `app/include/user_config.h`:
+To reduce memory size an integer build can be created.  You can change this 
+either by uncommenting `LUA_NUMBER_INTEGRAL` in `app/include/user_config.h`:
 
 ```c
 #define LUA_NUMBER_INTEGRAL
+```
+
+OR by overriding this with the `make` command as it's [done during the CI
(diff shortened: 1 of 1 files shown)


devsaurus:
You can also get an integer build without modifying the include file. Taken from [`.travis.yml`](https://github.com/nodemcu/nodemcu-firmware/blob/master/.travis.yml):

> make EXTRA_CCFLAGS="-DLUA_NUMBER_INTEGRAL -DBUILD_DATE='\"'$BUILD_DATE'\"'"

PR #1999, opened 2017-06-06
Only handle errors if we have not parsed a complete object

### LINKED ISSUE #1989: sjson should trim trailing null bytes
### Expected behavior
sjson.decode() (and possibly encode()) should trim trailing null bytes. This is particularly important when the json string was encrypted and usually terminated by a null byte. (Null bytes in-between characters may be another issue since they are allowed in lua strings)
### Actual behavior
sjson.decode() throws a FOUND_NULL_BYTE exception when string has trailing null bytes. The following snippet shows this behavour and confirms that decoding goes through when trailing null bytes are removed.
### Test code

```Lua
=sjson.decode('{}')
table: 3fff0408
> cypher=crypto.encrypt("AES-ECB","1234567891234567", '{}')
> text=crypto.decrypt("AES-ECB","1234567891234567", cypher)
> =text
{}
> =sjson.decode(text)
stdin:1: JSON parse error: FOUND_NULL_BYTE
stack traceback:
	[C]: in function 'decode'
	stdin:1: in main chunk
> text=text:gsub('%z.*','')
> =sjson.decode(text)
table: 3fff0408
> 
```
### NodeMCU version
Nodemcu dev branch built from nodemcu-build.com with sjson and crypto libraries.

pjsg:
Fixes #1989 and #2005 .

This ignores errors that happen in the parser after the object is parsed. This make `sjson` behave like `cjson`.


pjsg:
I included the fix for #2005 as well....

Issue #1883, opened 2017-03-26
Rewrite of the Lua DS18B20 module


TerryE:
I am not sure at this stage whether this is a bugrep or a missing feature.  I am using DS18B20s in anger and decided that I'd rewrite the ds18b20 lua module.  My reasons were as follows:

-  I wanted to make the module tmr compatible, and to this end I've split the read function into a `convert_T()` call which issues the convert command and a separate `read()` which collects the response from a single thermometer which and be collected on a ALARM_SEMI callback.
-  The module now supports multiple DS18B20s correctly in that a `convert_(true)` does an `ow.skip()` instead of an `ow.select()` and starts all DS18B20s converting in parallel.
-  The module saves the search output to `ds18b20_save.lc` and reloads this subsequently on first use so that you don't need to do the search each waking after deep sleep.
-  I've also used Lua best practice for locals and upvalues to minimise size / runtime.

Here is my test harness which shows the difference at an application level:
```Lua
local ds = require("ds18b20")
local to_string, print, gpio2 = ds.to_string, print, 4
local t1, addrs

ds.init(gpio2)
addrs = ds.get_addrs()

if addrs and #addrs > 0 then
  print("Total DS18B20 sensors: ")
  for k,v in ipairs(addrs) do print (' ',to_string(v)) end
  local tmr1,tmr2 = tmr.create(), tmr.create()
  tmr1:register(1000, tmr.ALARM_SEMI, function()
    local t = {}
    for i = 1, #addrs do t[i] =  ("%8.3f"):format(ds.read(addrs[i])) end
    print(unpack(t))
  end)
  tmr2:alarm(30000, tmr.ALARM_AUTO, function()
    ds.convert_T(true)
    tmr1:start()
  end)
else
  print("No OW thermometers found")
end
```
You can happily crank this up to collect half a dozen probe temperatures each second.

TerryE:
@devsaurus @pjsg @jmattsson I've added a gist copy [here](/TerryE/5ea4f51492468a46b095279b0fff6935).  I'd be interested in your views.  I am rewriting a few of these for my own use anyway, and I'd like your views on whether it is worth replacing the existing libraries as we do so.

What I've done here is to make the code a lot more runtime efficient as well as address a few functional holes.

karrots:
Gist link needs to be fixed. Github is improperly prepending `github.com` instead of `gist.github.com`

TerryE:
[...]

PR #2003, opened 2017-06-11
add ds18b20 module

fetchbot:

DS18B20 module in C rather than Lua #1996, #1883, #1820.

### DIFF
--- a/app/modules/ds18b20.c
+++ b/app/modules/ds18b20.c
@@ -1,0 +1,207 @@
+//***************************************************************************
+// DS18B20 module for ESP8266 with nodeMCU
+// fetchbot @github
+// MIT license, http://opensource.org/licenses/MIT
+//***************************************************************************
+
+#include "module.h"
+#include "lauxlib.h"
+#include "platform.h"
+#include "osapi.h"
+#include "driver/onewire.h"
+
+//***************************************************************************
+// OW ROM COMMANDS
+//***************************************************************************
+
+#define DS18B20_ROM_SEARCH				(0xF0)
+#define DS18B20_ROM_READ				(0x33)
+#define DS18B20_ROM_MATCH				(0x55)
+#define DS18B20_ROM_SKIP				(0xCC)
+#define DS18B20_ROM_SEARCH_ALARM		(0xEC)
+
+//***************************************************************************
+// OW FUNCTION COMMANDS
+//***************************************************************************
+
+#define DS18B20_FUNC_CONVERT			(0x44)
+#define DS18B20_FUNC_SCRATCH_WRITE		(0x4E)
+#define DS18B20_FUNC_SCRATCH_READ		(0xBE)
+#define DS18B20_FUNC_SCRATCH_COPY		(0x48)
+#define DS18B20_FUNC_E2_RECALL			(0xB8)
+#define DS18B20_FUNC_POWER_READ			(0xB4)
+
+//***************************************************************************
+
+static uint8_t ds18b20_bus_pin = 3;
+static uint8_t ds18b20_device_index;
+static uint8_t ds18b20_device_par;
+static uint8_t ds18b20_device_conf[3];
+static uint8_t ds18b20_device_rom[8];
+static uint8_t ds18b20_device_scratchpad[9];
+static double ds18b20_device_scratchpad_temp;
+static int ds18b20_device_scratchpad_temp_dec;
+static uint8_t ds18b20_device_scratchpad_conf;
+static uint8_t ds18b20_device_res = 12;			// 12 bit resolution (750ms conversion time)
+
+os_timer_t ds18b20_timer;						// timer for conversion delay
+int ds18b20_timer_ref;							// callback when readout is ready
+
+static int ds18b20_lua_readoutdone(void);
+
+// Setup onewire bus for DS18B20 temperature sensors
+// Lua: ds18b20.setup(OW_BUS_PIN)
+static int ds18b20_lua_setup(lua_State *L) {
+	// check ow bus pin value
+	if (!lua_isnumber(L, 1) || lua_isnumber(L, 1) == 0) {
+		return luaL_error(L, "wrong 1-wire pin");
+	}
+	
+	ds18b20_bus_pin = luaL_checkinteger(L, 1);
+	MOD_CHECK_ID(ow, ds18b20_bus_pin);
+	onewire_init(ds18b20_bus_pin);
+}
+
+// Change sensor settings
+// Lua: ds18b20.setting(ROM, RES)
+static int ds18b20_lua_setting(lua_State *L) {
+	// check rom table and resolution setting
+	if (!lua_istable(L, 1) || !lua_isnumber(L, 2)) {
+		return luaL_error(L, "wrong arg range");
+	}
+	
+	ds18b20_device_res = luaL_checkinteger(L, 2);
+	
+	if (!((ds18b20_device_res == 9) || (ds18b20_device_res == 10) || (ds18b20_device_res == 11) || (ds18b20_device_res == 12))) {
+		return luaL_error(L, "Invalid argument: resolution");
+	}
+	
+	for (uint8_t i = 0; i < 8; i++) {
+		lua_pushinteger(L, i+1);
+		lua_gettable(L, -3);
+		if (lua_isnumber(L, -1)) {
+			ds18b20_device_rom[i] = lua_tonumber(L, -1);
+		}
+		lua_pop(L, 1);
+	}
+	
+	// no change to th and tl setting
+	ds18b20_device_conf[0] = 0x4B;
+	ds18b20_device_conf[1] = 0x46;
+	ds18b20_device_conf[2] = ((ds18b20_device_res - 9) << 5) + 0x1F;
+	
+	// set conversion delay once to max if sensors with higher resolution still on the bus
+	ds18b20_device_res = 12;
+	
+	onewire_reset(ds18b20_bus_pin);
+	onewire_select(ds18b20_bus_pin, ds18b20_device_rom);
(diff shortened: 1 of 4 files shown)


marcelstoer:
I guess you'll want to drop `/lua_modules/ds18b20` with this as well?

TerryE:
TBH, I don't understand the desire to use C and write C libraries where Lua ones work fine.  Lua developers can always tweak a Lua module if their needs aren't 100% aligned to the current module implementation -- and can do so without having to get into C or rebuild the firmware.  In this case the core One-wire C library encapsulates everything that needs to be done in C.  So why reimplement in C?

joysfera:
I've been told that almost nothing needs to be running in memory - most if not all code can be loaded on demand, like dofile("ds18b20.lc"). The available NodeMCU documentation covers 'require()' use only, though. No detailed word about 'dofile()', IIRC. I think it would be great if NodeMCU beginners could see some large-ish example programs that would show how to keep only the necessary minimum of Lua code in memory. Then the need for offloading Lua code to C modules would disappear, perhaps.

The DS18B20 is actually a good example of a code that might not be simple to keep out of memory: it needs one timer to wait those 750+ milliseconds between decoding and reading the temperature, and usually another timer to measure temperatures say every minute. How to write the dofile() or require() in a way that ensures the memory is indeed freed immediately and completely?

devyte:
If the code is linear, as in just a single pass through and no functions,
then that is true: nothing is kept in mem.

If there are functions declared, then each of those gets loaded into memory
first, then they get called whenever. I have no idea what happens to the
loaded functions after the dofile returns.

Also, is it possible to pass arguments to a dofile call? If not, then
parameter passing must be through some global variable scheme, which is
rather fragile.

Finally, if a callback needs to be setup for something, which is quite
common, then you'll probably need a closure of some kind. I found that in
most cases that is not viable to do, unless the code loaded with dofile is
itself an entire function that gets loaded.

Of course, I'm not a lua expert, so please take my comments with a grain of
salt, but that has been my experience.

marcelstoer:
@joysfera I think the `dofile()` vs. `require()` discussion could be fed into #1899 for @TerryE to be addressed in the FAQ update?

TerryE:
> I think the dofile() vs. require() discussion could be fed into #1899 for @TerryE to be addressed in the FAQ update?
@marcelstoer Marcel, thanks for the prompt.  I am coming to a checkpoint in my Lua 5.3 -- time to migrate into the Xtensa build environment -- so I will do this refresh first. :smile:

app/modules/ds18b20.c
```diff
@@ -0,0 +1,207 @@
...
+		lua_pushinteger(L, i+1);
+		lua_gettable(L, -3);
+		if (lua_isnumber(L, -1)) {
+			ds18b20_device_rom[i] = lua_tonumber(L, -1);
+		}
+		lua_pop(L, 1);
+	}
+	
+	// no change to th and tl setting
+	ds18b20_device_conf[0] = 0x4B;
```
nickandrew: Can you put the magic numbers into #defines?

docs/en/modules/ds18b20.md
```diff
@@ -0,0 +1,119 @@
...
+#### Returns
+- `INDEX` index of the sensor on the bus
+- `ROM` sensors 64-bit lasered rom code
+	* `ROM0` LSB, 8-bit family code
+	* `ROM1` - `ROM6` 48-bit serial number
+	* `ROM7` MSB 8-bit crc
+- `RES` temperature resolution
+- `TEMP` temperature
+- `TEMP_DEC` temperature decimals for integer firmware
+- `PAR` sensor parasitc flag
```
nickandrew: Typo

docs/en/modules/ds18b20.md
```diff
@@ -0,0 +1,119 @@
...
+
+-- print only parasitic sensors
+ds18b20.read(
+	function(ind,r0,r1,r2,r3,r4,r5,r6,r7,res,temp,tdec,par)
+		if (par == 1) then
+			print(ind,string.format("%02X%02X%02X%02X%02X%02X%02X%02X",r0,r1,r2,r3,r4,r5,r6,r7),res,temp,tdec,par)
+		end
+	end);
+
+-- print if temperature is greater or less then a defined value
```
nickandrew: Typo

marcelstoer:
Can we have a quick voting among the maintainers as to whether to include this module? Terry already voted against it I understand.

TerryE:
@marcelstoer, my vote is more of an abstain rather than a :-1:  It's just that I personally don't see the point of doing this one in C and wouldn't personally use it, but I wouldn't wish to block its inclusion if other committers think that there is value in having it.

devsaurus:
My experience with this sensor is that people use it in many different ways. I guess that neither a C nor a Lua module would cover all. I regarded the old Lua module more like a template to start from and adapt to one's needs. The C module has a more "official" character and need more support effort on the long run.
We have a couple of modules in C that would also work as plain Lua code, so there's no point in blocking this one IMO.

nickandrew:
The argument against a C module is that it needs to be compiled-in. Arguments for are that it may use less memory, and be faster - which is good to ensure that any timing constraints on the 1wire bus are met. I believe the timing in this module is not critical (as it uses the onewire lib). I'd probably use this module in my devices.

All up I give it a weak thumbs-up; merge it and if problems are reported in future then reconsider.

marcelstoer:
How about we merge both #1996 and this one but cross-reference them in the documentation (I could do that after the merge)?

nickandrew:
I just had a look at #1996 but I don't have an opinion on its value, yet. I don't think it should hold up #2003.

TerryE:
OK, I've had a look at this code.  

I accept that this might be a good approach for @fetchbot 's specific requirements, but it is very functionally limited and I really don't see the point of doing this in C as it will run a _lot_ slower than my or @vsky279 Lukas' Lua version.  Why? and other points:

-  The `read()` function does a broadcast convert request and then uses an OW search to enumerate the OW address space (not even limiting it to DS18B20 device types).  This is a slow binary-chop scan.  Surely this only needs to be once or at least optional so you can do fast reads of specified thermometers?

-  The documentation is misleading as the `read()` returns nothing.  Instead it executes the supplied cb _once per DS18B20 found_ passing the return data as input parameters.

-  It makes a lot more sense to return the ROM1..ROM8 bytes as a single 8-byte string.  Lua strings aren't the same as C strings and can be binary.  You can always use the `byte` and `substr` functions to pick them apart.  Ditto the input ROM array to `setting()`.

-  The callback is called using a `lua_call` rather than a `lua_pcall` which means that any errors in the calback will panic the system (though to be fair, this is a common failing of most modules).

I still will register an abstain rather than a :-1: --  so long as we get resolution / consensus on the first 3 points.

@nickandrew Nick what are your thoughts?  Am I being too pedantic?

nickandrew:
@TerryE I think those are fair comments. My temp code treats the device IDs as a string or array rather than individual bytes (I also munge the byte order in a way that suits me). How much slower is the discovery algorithm than reading only a specified sensor? My temp code always does a select-all; convert; then goes through the algorithm to process all responses.

I need to spend more time looking at this :grimacing:

TerryE:
BTW,

> I guess you'll want to drop `/lua_modules/ds18b20` with this as well?

I for one will stick to Lua version.  It's faster and easier to tailor.

marcelstoer:
I believe that making the wrong decision is sometimes much worse than not making a decision at all. I cross-linked the C/Lua module docs: http://nodemcu.readthedocs.io/en/latest/en/modules/ds18b20/

Fisherman48:
Missing leading zero of decimal part?
Using integer firmware and concatenate temp and tdec strings ex:
..romhex.."="..temp..","..tdec.." " I get 6,62 when I expect 6,0625
Other examples with 3 ds18b20 and decreasing temp:
18,250 18,250 18,437
 18,62 18,62 18,250
 18,62 18,62 18,250
 18,62 18,125 18,250
 18,0 18,62 18,250
Help

devsaurus:
> ..romhex.."="..temp..","..tdec.." " I get 6,62 when I expect 6,0625

Guess you need to format tdec with leading zeros:

```lua
print(string.format("%d,%04d", temp, tdec))
```

Fisherman48:
Strange behaviour at 0,0 C?
Integer master modules , powered by Lua 5.1.4 on SDK 2.1.0(116b762)
adding temp + tdec in excel

Never show zero C temp- temp+tdec show -1 C when I expect zero. 
Se discontinuity in graph [graph
image
](url)

`
local ow_pin = 3
ds18b20.setup(ow_pin)

dstimer = tmr.create()
dstimer:register(60*1000, tmr.ALARM_AUTO, function()
    
    -- read all sensors and print all measurement results
    ds18b20.read(
        function(ind,rom,res,temp,tdec,par)
            print(ind,string.format("%02X:%02X:%02X:%02X:%02X:%02X:%02X:%02X",string.match(rom,"(%d+):(%d+):(%d+):(%d+):(%d+):(%d+):(%d+):(%d+)")),res,temp,tdec,par)
        end,{});
    end)

dstimer:start()`

/Fisherman

Fisherman48:
After a cold night the same effect occurs at all negative x.0 temps from  -2.0, -3.0 to -9.0
Is it something due to scaling and 2'nd complement for neg temps for integer version of module?
/Fisherman

TerryE:
@Fisherman48, have tried extending the wait time [Line 188](../blob/5073c199c01d4d7bbbcd0ae1f761ecc4687f7217/app/modules/ds18b20.c#L188) out to 800 or 900 mSec. 760 is only 10mSec over the datasheet minimum, and the code might just be jumping in too early before the reading is complete in cold weather.  I use 1sec in my Lua version but then again I do all of my device converts in parallel then loop around the devices to do the read, so it still works out a lot faster.

PR #2008, opened 2017-06-21
Add sqlite3 module

### LINKED ISSUE #1863: sqlite3 support
I did a port of sqlite3 to nodemcu and would like to know if it would be interesting to the nodemcu community too.
The motivation was that I am about to build an access control system using RFIDs tags that would depend on a database with allowed tags and keys built on a pc or on a pi 3 or beagle bb. This database would be broadcasted to nodemcus via a secured connection over wifi using ssl or something like that.
The first thing I needed was to first enable nodemcu to work with sqlite3 databases and this is what I got.
This patch is based on sqlite3 (of course) and luasqlite3 (half way there) with a esp8266 sqlite3 vfs developed by me.

Would this patch be of interest of nodemcu community?

luizfeliperj:
Fixes #1863 .

This module adds the sqlite3 support to NomeMCU.


marcelstoer:
As for the failing build

> `.irom0.text` will not fit in region `irom0_0_seg`

Sigh, there we go again...

dnc40085:
@marcelstoer perhaps the previous fix that was in `tools/pr-build.sh` could be added back in.

Section removed  from `tools/pr-build.sh` in PR #1743:
``` bash
cd "$TRAVIS_BUILD_DIR"/ld || exit
 # increase irom0_0_seg size for all modules build
 sed -E -i.bak 's@(.*irom0_0_seg *:.*len *=) *[^,]*(.*)@\1 0xD0000\2@' nodemcu.ld
 cat nodemcu.ld
```

marcelstoer:
Hmm, not sure I understand...

With b09cac0 (came after PR #1743) @pjsg extended it from `0xD0000` to `0xE0000`. If we were to adjust `tools/pr-build.sh` again we'd have to enlarge the segment from `0xE0000` to `0xF0000` for it to be effective, no?

luizfeliperj:
The sqlite3 library size occupies exactly 230720 bytes, 0xF0000 could not be enough.

dnc40085:
> If we were to adjust tools/pr-build.sh again we'd have to enlarge the segment from 0xE0000 to 0xF0000 for it to be effective, no?

Going on the assumption that the purpose of the Travis CI build is to make sure that the new modules/additions play nice with the preexisting modules/code and not to make sure that it will work with the limitations of the ESP8266 (1024 kB limit on program size, RAM, etc.), I would say it's safe to set the `irom0_0_seg` and `iram1_0_seg` much higher values for PR builds.

Maybe 0x200000 for `irom0_0_seg` and 0x10000 for `iram1_0_seg` to cover future additions.

I tried this locally and it compiled fine, but spiffs caused the build to fail with `Not enough space: fatal error` which isn't surprising since it's probably going by the 1024kB limit.

app/include/user_modules.h
```diff
@@ -77,6 +77,7 @@
 //#define LUA_USE_MODULES_WS2801
 //#define LUA_USE_MODULES_WS2812
 //#define LUA_USE_MODULES_XPT2046
+//#define LUA_USE_MODULES_SQLITE3
```
dnc40085: user_modules.h line 19:
> // New modules should be disabled by default and added in alphabetical order.

nickandrew:
It's a bummer if this module can never pass CI due to CI needing to link *every* module.

Is it possible to split the CI into two passes, one with every module except sqlite3 enabled and the 2nd with only sqlite3 enabled?

marcelstoer:
I think it'd be easier to just implement @dnc40085 's proposal of increasing the memory params.

TerryE:
@marcelstoer, @dnc40085, @nickandrew you are the guys who have commented on this PR / merge.  Have any of you actually built and used this sqlite3 implementation?  I didn't review it in depth, but the way it's built seems rather odd.

I use a 384Mb VM for my dev sandpit and this has always worked brilliantly -- until I started trying to build with a recent `dev`:  I can't even compile the single sqllite3.c file as `ar` runs out of RAM+swap, so my workaround is to remove sqllite3 from the app `SUBDIRS` list.   I now notice that you've materially increased the `iram1_0_seg` to get this to link, But the `iram_0_seg` is dictated by the ESP8266 hardware, so again I've got to question: 

_has anyone actually got this running, as opposed to getting a clean build?_  

If not, then why have we merged it?

marcelstoer:
> has anyone actually got this running, as opposed to getting a clean build?

Yes but I can only speak for myself. I did run a few smoke tests. As you know I can't assess the code quality 😦. My test firmware had the default modules like the cloud builder plus HTTP, SQLite and SSL enabled.

As for the `iram1_0_seg` discussion, we've had this last summer in the comments above. The large increase is only effective for PR builds.

luizfeliperj:
Compiling sqlite3 module requires a lot of compiler effort so it can generate the smallest output possible, I think a 384Mb VM is just not enough to run the compiler to compile it.

TerryE:
> I think a 384Mb VM is just not enough to run the compiler to compile it.

This issue here is that whether or not a developer is using sqlite3, and very few will, this change means that _any_ developer now has to use a 512Mb VM to do a NodeMCU build. This isn't good, is it?

luizfeliperj:
I understand that this can be an issue, as I use docker to compile nodemcu this was never an issue for me. I can split the sqlite3 amalgamation file in multiple files to make it easier to the compiler, but that would be at cost of larger code output, would that be ok?

TerryE:
@luizfeliperj  Luiz, thanks for this positive suggestion.

Splitting the aggregated C file into a few files, say 2 or 3, will make a tiny difference to the generated code size but collapse the memory requirement for the compile.  Also you've got about 30 or so defines in your `Makefile`.  Why not stick these into a `sqlite3_config.h` file and put it in the directory since it is an internal header.  This will make the log easier to read.

luizfeliperj:
@TerryE I will do that and update when it is ready

TerryE:
@luizfeliperj, thanks for this. 😄

luizfeliperj:
@TerryE , I tried to break the sqlite3.c file in few evenly distributed files but I couldn't, during the amalgamation lots of tricks are used and lots of function calls are defined as static, it will be hard to figure out which I need to export, I can keep trying but it will take a reasonable amount of time

For now, the best I can do is use the .c files sqlite3 uses to produce the amalgamation, that will, for sure, ease the work for the compiler, but compromising the file size of the object.

Should I upload the sqlite3 module without amalgamation to dev?

TerryE:
@luizfeliperj Luiz,  I am about to go to bed, but I think that we could do a compromise with some Makefile magic, so that the sqlite3 subdir is only compiled if the module is enabled in `user_config.h`.  That way everyone wins without you having to do a lot of work. I'll look at this tomorrow, and suggest something for you to try.  

Regards Terry :smile:

TerryE:
The make magic is to add the following to the `ifndef PDIR` branch in the `app/Makefile`
```make
OPT_MKTARGETS      := coap crypto dht http mqtt pcm sjson sqlite3 tsl2561 u8glib ucglib websocket
SEL_MKTARGETS      := $(shell $(HOSTCC) -E -dM include/user_modules.h | sed -n '/^\#define LUA_USE_MODULES_/{s/.\{24\}\(.*\)/\L\1/; p}')
OPT_SEL_MKTARGETS  := $(foreach tgt,$(OPT_MKTARGETS),$(findstring $(tgt), $(SEL_MKTARGETS)))
OPT_SEL_COMPONENTS := $(foreach tgt,$(OPT_SEL_MKTARGETS),$(tgt)/lib$(tgt).a)
```
And use the `OPT_SEL_MKTARGETS` and `OPT_SEL_COMPONENTS` in the respective lists.  Note for these libraries, I've also standardised the naming convention.  Hence for the **OPT_MKTARGETS  **, the companion directory is only compiled if the corresponding module is selected.

luizfeliperj:
@TerryE , removing the makefile defines and pushing them to a file, my proposal to fix it is in PR #2297 .

PR #2017, opened 2017-06-28
Update documentation for wifi.sta.setaplimit()

### LINKED ISSUE #2010: wifi (doc) questions
Hi/,
please add to docs (+code, maybe):

* wifi.sta.clearconfig() immediately breaks an existing connection (odd, if you just want to write a new config to flash and keep online status)
* after that, wifi.sta.get[default]config() returns old values (should be nil)
* wifi.sta.getapindex() returns 1 (should be nil)

It's not clear from the docs, how to juggle with wifi configs: How to decide when to clear/recover to a default config (virgin/crashed device), in which slot does a config(save=true) land, how to rewrite a specific slot, does some algo automagically select the best config (signal strength), does it switch networks when the device is moving?

I suppose, the SDK imposes that shortcomings, sorry for couriosity.
TY

dnc40085:
Fixes #2010

This PR adds a note to the `wifi.sta.setaplimit()` documentation that lets the developer know that the new setting will not take effect until restart.

### DIFF
--- a/docs/en/modules/wifi.md
+++ b/docs/en/modules/wifi.md
@@ -899,7 +899,10 @@ Set Maximum number of Access Points to store in flash.
  - This value is written to flash
 
 !!! Attention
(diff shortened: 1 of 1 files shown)


marcelstoer:
Thanks

Issue #1721, opened 2017-01-12
http.get() Memory Leak


zhujieo:
8<------------------------ BUG REPORT -----------------------------------------
My Egnlish is bad,Sorry!

I have problem:
    when Nodemcu connected my web server the memory is not leak,
    but when I close my web server, the nodemcu is disconnect the web service,it's memory is leak.

### Test code
Provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) which will reproduce the problem.
```Lua
wifi.setmode(wifi.STATION)
wifi.sta.config('****','*****')
wifi.sta.connect()

tmr.alarm(0,2000,1,function ()
    print(node.heap())
    http.get("http://192.168.1.149:8000/devices/command", nil, function(code, json)
        if (code < 0) then
            print('error')
        else
            print('ok')
        end
    end)
end)
```

ESPlorer Report:

NodeMCU 1.5.4.1 build 2016/12/31 powered by Lua 5.1.4 on SDK 1.5.4.1(39cb9a32)
> 40520
42848
42320
ok
42216
ok
42232
ok
42232
HTTP client: Connection timeout
ok
41464
HTTP client: Connection timeout
ok
41464
ok
41464
ok
41464
ok
41464
ok
41464
ok
41464
ok
41464
ok
41464
ok
41448 . -----------(This time I closed my server)
HTTP client: Disconnected with error: -9
HTTP client: Connection timeout
41264
HTTP client: Disconnected with error: -9
HTTP client: Connection timeout
40976
HTTP client: Disconnected with error: -9
HTTP client: Connection timeout
40728
HTTP client: Disconnected with error: -9
HTTP client: Connection timeout
40464
HTTP client: Disconnected with error: -9
HTTP client: Connection timeout
40184
HTTP client: Connection timeout
HTTP client: Disconnected with error: -9
HTTP client: Connection timeout
......

### NodeMCU version
Master

NodeMCU 1.5.4.1 build 2016/12/31 powered by Lua 5.1.4 on SDK 1.5.4.1(39cb9a32)

### Hardware
Nodemcu with nothing attached

8<------------------------ END BUG REPORT -------------------------------------

djphoenix:
Can you check that problem still appears on latest `dev`?

zhujieo:
OK, Let me try it. Thanks.

zhujieo:
In the latest dev ,this problem is still have.

[...]

PR #2020, opened 2017-06-30
fix #2007 http: Call callback in all cases

### LINKED ISSUE #2007: HTTP.get no callback on error
http.get on a local IP that is not available (not assigned) fails but does not call callback

### Expected behavior
The callback should be called  **allways**, even on all failures

in the example it should print
callback

### Actual behavior
The callback is not called

-- > http.get("http://192.1.1.254/",nil,function() print "callback" end)
-- > HTTP client: Disconnected with error: -11
-- HTTP client: Connection timeout
-- HTTP client: Connection timeout

### Test code

```Lua
-- http.get("http://192.1.1.254/",nil,function() print "callback" end)
```
### NodeMCU version
	branch: master
	commit: c8ac5cfb912ff206b03dd7c60ffbb2dafb83fe5e
	SSL: true
	modules: bit,file,gpio,http,i2c,net,node,ow,pwm,rtctime,sjson,sntp,tmr,uart,wifi,tls
 build 	built on: 2017-06-17 17:53

### Hardware
Standard ESP6288 E-12 Dev Board

HHHartmann:
Call callback with errorcode -1 if no connection could be establioshed

Fixes #2007 and #1721.

the PR checks the returnvalue of espconn_(secure_)disconnect. If no connection was established it fails and no disconnected callback is to be expected.
In this case the callback is called directly

I allso added some more error logging which I can remove on request.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -456,7 +456,7 @@ static void ICACHE_FLASH_ATTR http_timeout_callback( void *arg )
 	}
 	if ( conn->reverse == NULL )
 	{
-		HTTPCLIENT_DEBUG( "reverse is NULL" );
+		HTTPCLIENT_DEBUG( "Connection request data (reverse) is NULL" );
 		return;
 	}
 	request_args_t * req = (request_args_t *) conn->reverse;
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -451,20 +451,32 @@ static void ICACHE_FLASH_ATTR http_timeout_callback( void *arg )
 	struct espconn * conn = (struct espconn *) arg;
 	if ( conn == NULL )
 	{
+		HTTPCLIENT_ERR( "Connection is NULL" );
 		return;
 	}
 	if ( conn->reverse == NULL )
 	{
+		HTTPCLIENT_ERR( "reverse is NULL" );
 		return;
 	}
 	request_args_t * req = (request_args_t *) conn->reverse;
+	HTTPCLIENT_ERR( "Calling disconnect" );
 	/* Call disconnect */
+	sint8 result;
 #ifdef CLIENT_SSL_ENABLE
 	if ( req->secure )
-		espconn_secure_disconnect( conn );
+		result = espconn_secure_disconnect( conn );
 	else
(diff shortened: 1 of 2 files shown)


app/http/httpclient.c
```diff
@@ -451,20 +451,32 @@ static void ICACHE_FLASH_ATTR http_timeout_callback( void *arg )
...
+		HTTPCLIENT_ERR( "Connection is NULL" );
 		return;
 	}
 	if ( conn->reverse == NULL )
 	{
+		HTTPCLIENT_ERR( "reverse is NULL" );
 		return;
 	}
 	request_args_t * req = (request_args_t *) conn->reverse;
+	HTTPCLIENT_ERR( "Calling disconnect" );
```
marcelstoer: Shouldn't this be `HTTPCLIENT_DEBUG`?

marcelstoer:
Thanks you. In #2007 we established that there's also an inconsistency in our documentation wrt the disconnection handling. Can I ask you to clean this up as well in this PR?

app/http/httpclient.c
```diff
@@ -451,20 +451,33 @@ static void ICACHE_FLASH_ATTR http_timeout_callback( void *arg )
 	struct espconn * conn = (struct espconn *) arg;
 	if ( conn == NULL )
 	{
+		HTTPCLIENT_DEBUG( "Connection is NULL" );
 		return;
 	}
 	if ( conn->reverse == NULL )
 	{
+		HTTPCLIENT_DEBUG( "reverse is NULL" );
```
marcelstoer: It may be helpful to print a more explanatory message. The meaning of "reverse" may not be totally obvious.

HHHartmann: @marcelstoer As to the documentation inconsistency: I am not aware of it, please enlighten me so I can add it.

The fix above establishes the behavior as in the docs AFAICS.

HHHartmann:
@marcelstoer Documentation is a good point, already forgot about that problem. But there you go ...

HHHartmann:
@marcelstoer well I intentionally left out the period as it had been omitted in the other paragraphs as well ;-)

PR #2027, opened 2017-07-04
ESP32: Ledc driver module

larsstenberg:

Started on LEDC module, this is work in progress started pull request to get early feedback since this is my first module.


jmattsson:
I suspect we might need some strategy around the timer usage, but this looks like a good start. @devsaurus you're way more across the ESP32 stuff than I am at the moment, would love your thoughts here :)

jmattsson:
@larsstenberg give us a shout when you want us to have a thorough review here :)

larsstenberg:
Well now would be ok :)

larsstenberg:
Makes sense, i will make a object oriented approach instead =)

larsstenberg:
ledc

Seems to be working great =)

PR #2028, opened 2017-07-05
Unbreak build when SPIFFS_CACHE==0.

jmattsson:

It seems at some point we lost the ability to build if `SPIFFS_CACHE` is defined to be `0` in `user_config.h`. The spiffs code bails out due to `SPIFFS_CACHE_WR` not being defined. This fixes that problem by providing a default definition for it.


jmattsson:
I acknowledge we (DiUS) seem to be the only ones using the filesystem as read-only, so it's probably not worth pushing this issue upstream.

PR #2032, opened 2017-07-08
ws2812 mix: add all, then divide 

nwf:

This flips the ws2812 framebuffer `mix` function to add all the contributions together before dividing; this allows one to, as I do, ensure that a channel stays on even in light of other operations: `basefb:fill(1,1,1) ; outfb:mix(255,basefb,256/dimfactor,infb)`, which results in dimming that doesn't zero out color channels.

No docs changes necessary, I think.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -399,17 +399,19 @@ static int ws2812_buffer_mix(lua_State* L) {
 
   size_t i;
   for (i = 0; i < cells; i++) {
-    int val = 0;
+    int32_t val = 0;
     for (src = 0; src < n_sources; src++) {
-      val += ((int)(source[src].values[i] * source[src].factor) >> 8);
+      val += (int32_t)(source[src].values[i] * source[src].factor);
     }
+
+    val >>= 8;
 
     if (val < 0) {
       val = 0;
     } else if (val > 255) {
       val = 255;
     }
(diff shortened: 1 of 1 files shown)


marcelstoer:
@Alkorin does this look reasonable to you?

Alkorin:
What's the size of an 'int' ?
Would be better to use an uint32_t to avoid any undefined behaviour

marcelstoer:
@nwf do you plan to address Thomas' feedback?

nwf:
Mm, sorry, I am busy writing my PhD thesis, so time is basically, uh, zero.  That said, I don't object, but I'm not sure a uint32_t is really necessary; a uint16_t would handle mixing hundreds of framebuffers without overflow, no?

Alkorin:
As `source[src].values` is an `uint8_t` and factor an `int`, you already have an overflow, as only a `int32_t` may store the result. If factor were only an `uint8_t`, an `uint16_t` can only store the result of 1 framebuffer (255 * 255).

That's why I think `int32_t` is a better solution.

nwf:
@Alkorin Erm, yes, of course.  OK.  I'll make the change when I get a minute.

nwf:
Let's that on for size (thesis document done, too!)

nwf:
@Alkorin @marcelstoer ping

PR #2035, opened 2017-07-10
Added mention of another Lua book that covers Lua 5.1 to the Lua developer FAQ.

luaqr:

This PR is a simple documentation update to the Lua developer FAQ that adds mention of another modern Lua book that covers Lua 5.1.


marcelstoer:
> This PR is compliant with the other contributing guidelines 

Nope, because it's for `master` rather than `dev` -> please change and re-submit

TerryE:
I would :-1: any attempt to add this link. My rationale is that is just a piece of self advertising by the author, and I've look for references and reviews of the book and I can't find any.  From the introduction and contents list, it seems to add nothing over PiL, and certainly nothing which extends Lua into its use in this sort of IoT environment.

My advice to all Lua programmers and library developers is that if you want to buy a book on Lua, then start with PiL, 3rd edition.

PR #2038, opened 2017-07-11
Fix the SPIFFS blocksize issues

### LINKED ISSUE #2030: spiffsimg creates incompatible filesystem for small fs sizes
### Expected behavior
The bundled `spiffsimg` to generate usable filesystems.

### Actual behavior
For small file systems (128k or less I believe), NodeMCU does not recognise the filesystem and promptly reformats it.

### Test code

In `user_config.h`, something like:
```
#define SPIFFS_FIXED_LOCATION 0xA0000
#define SPIFFS_FIXED_SIZE 0x10000
```
Prepare a filesystem of that size:
```
tools/spiffsimg/spiffsimg -f test.fs -c 0x10000 -i
> import Makefile Makefile
```
Build and flash firmware + fs at the appropriate spots. Boot and you'll get told that it's formatting the filesystem, and your Makefile is nowhere to be seen.

### NodeMCU version
`dev`

### Cause
In the olden days we determined the logical block size to use depending on the overall filesystem size. This logic remains in `spiffsimg/main.c`. In the NodeMCU `spiffs.c` however, the handling of this has changed to be incompatible with `spiffsimg`. I don't see an obvious way of resolving this conflict, as now the address depends on the block size, and the block size depends on the size, which in turn depends on the address. @pjsg I'm very tempted to lob this into your court to figure out :)

pjsg:
Fixes #2030.

- [x] @jmattsson has thoroughly tested my contribution.

This change makes the mounting code try and mount the filesystem with both the small and the large blocksize. Thus it shouldn't matter which blocksize was used by spiffsimg.

### DIFF
--- a/app/spiffs/spiffs.c
+++ b/app/spiffs/spiffs.c
@@ -73,7 +73,7 @@ static bool myspiffs_set_location(spiffs_config *cfg, int align, int offset, int
  * Returns  TRUE if FS was found
  * align must be a power of two
  */
-static bool myspiffs_set_cfg(spiffs_config *cfg, int align, int offset, bool force_create) {
+static bool myspiffs_set_cfg_block(spiffs_config *cfg, int align, int offset, int block_size, bool force_create) {
   cfg->phys_erase_block = INTERNAL_FLASH_SECTOR_SIZE; // according to datasheet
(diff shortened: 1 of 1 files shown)


jmattsson:
Thanks Philip!

PR #2054, opened 2017-07-28
dht_read11 DHT12 Fix

### LINKED ISSUE #2050: dht_read() reads data in a wrong way?
At line 122, 123 and 182, 183 in file [dht.c](https://github.com/nodemcu/nodemcu-firmware/blob/master/app/dhtlib/dht.c), a macro is used to combine lower and higher bits of sensor data bytes. The macro is defined as
```c
```
I think it is equivalent to byte_high * 256 + byte_low, while it should be byte_high * 10 + byte_low according to documents of dht sensor (say [this one](https://akizukidenshi.com/download/ds/aosong/DHT11.pdf)). 

The code works fine with dht11 sensor because all lower bits are 0 and the app won't come to this line. I'm afraid it will go wrong when reading other dht sensor (say dht12).

I might misunderstand the document or the code, but it confused me. Can any one tell me am I making mistakes? Is there something wrong?

AdrianDroid:
Fixes #2050.

Due to shorthand implementation of DHT11,  DHT12 will fail to DHTxx logic.
However, DHT12 use the same data representations, thru DHT12 will have the wrong reading when using dht_read().
and shorthand implementation of DHT11 of its checksum will make dht_read11() fail in DHT 12.
this fix allows DHT12 user to call dht_read11() to have the correct reading.

### DIFF
--- a/app/dhtlib/dht.c
+++ b/app/dhtlib/dht.c
@@ -41,6 +41,7 @@
 #endif /* ifndef HIGH */
 
 #define COMBINE_HIGH_AND_LOW_BYTE(byte_high, byte_low)  (((byte_high) << 8) | (byte_low))
+#define COMBINE_INT_AND_FRACTION(integer,  fraction)  ((integer * 10) + (fraction))
 
 static double dht_humidity;
 static double dht_temperature;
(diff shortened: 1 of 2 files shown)


marcelstoer:
I don't have hardware to verify and the DHT module has a really crappy API IMO (`dht.read(), dht.read11(), dht.readxx()`) but I think this fix should be handled differently. How about:
- rename `dht_read11` to `dht_read11_12`
- do the same in `app/modules/dht.c` for `dht_lapi_read11`
- register a `read12` Lua function that maps to `dht_lapi_read11_12`

On the other hand, one may ask why we should have a dedicated Lua function for 11 and 12 but not for the others....

nickandrew:
If dht11 and dht12 use the same read logic then the function should be called dht_read1x() and then map dht_read11 and dht_read12 to that function.

marcelstoer:
That'd be a good alternative Nick, thanks. The only downside I see is that `read1x` implies that it will work for _all_ DHT10-something sensors. That's a bit dangerous as we can't anticipate what the characteristic of a potential DHT15 sensor will be - should there ever be such a sensor.

Ultimately what I'd like to see is an expressive _Lua_ API with a dedicated `read` function for each DHT sensor. That would give us total freedom as to how those functions are mapped to C functions. And it would give our users an extremely stable interface to program against.

I know that such a refactoring is well beyond the scope of this PR but any changes to the DHT module should IMO be aligned to this idea.

nickandrew:
I'm assuming that dht_read1x will work for all 1x sensors until it's shown otherwise, and then we can cross that bridge. So in line with your ultimate desire, the end user should call a function named according to their device.

Come to think of it, dht_read11() and dht_read12() aren't a great naming standard, it should be dht11_read() & dht12_read() or read_dht11() & read_dht12(). The 11/12 should be next to the dht.

devsaurus:
> it should be dht11_read() & dht12_read()

Sounds like an object-oriented view, appealing idea. Though read() would be the only method, a real object feel like overkill though.

marcelstoer:
@AdrianDroid do you plan to incorporate the changes suggested here?

AdrianDroid:
updated to #2071

PR #2056, opened 2017-08-02
Refactor timer suspend portion of node.sleep (pmsleep)

### LINKED ISSUE #2048: pmSleep and Timer suspend enabled by defaul despite runtime impact on all code.
### Expected behaviour

PR  #1231.  as this case updated `sdk-overrides/include/osapi.h`
```C
#define os_timer_arm(timer_ptr, duration, mode) do{swtmr_register(timer_ptr); \
  ets_timer_arm_new(timer_ptr, duration, mode, 1);}while(0);
```
and similarly for disarm, and these `swtmr` both add a hidden runtime memory and call overhead overhead and the ability to fail silently due to out-of-memory conditions.

-  No change to the core RTS which changes the timing and runtime nature of another module should be introduced without explicitly highlighting this runtime impact as part of the change.  The documentation should also explicitly point this out.

-  In such cases, the  `user_config.h` should also disable the feature by default, with a referenced to the issue in the documentation.

### Actual behaviour

The two key parameters which confiigure this functionality [lines 118:119](nodemcu/nodemcu-firmware/blob/master/app/include/user_config.h#L118]) are enabled by default.  They shouldn't be.  And the additional explanation needs to be added to the documentation.

### NodeMCU version

Current dev and master

### Test code & Hardware

N/A

dnc40085:
Fixes #2048.

The previous version of the timer suspend code was large, difficult to read and had some problems.
This PR changes the way the timer suspend uses memory, rather than storing the necessary data in memory blocks allocated with `c_malloc()`, the new implementation stores its data in the Lua registry and in doing so a lot of extra code was able to be removed.

**!!!IMPORTANT!!!** Some functions were removed in this PR! 
~~No deprecation notices have been included for the following functions~~
A message has been left in the following functions notifying the developer that the feature has been removed.
* `tmr.suspend()` 
* `tmr.suspend_all()`
* `tmr.resume()`
* `tmr.resume_all()`


marcelstoer:
> Some functions were removed in this PR! No deprecation notices have been included...

Ouch! I can't gauge how big of an impact this has on our users. I guess those functions weren't the most popular ones, though.

Would it make much of a difference if you left the functions in place for now but replaced their body with a "Sorry, this was removed" message?

dnc40085:
@marcelstoer No problem, I'll submit a PR shortly.

marcelstoer:
You could just push to your existing branch and this PR will update.

dnc40085:
My mistake, I meant push a commit.

TerryE:
@dnc40085   Hi dnc, can we have a discussion about the requirement and implementation here?

Surely the underlying requirement is that we can add the `node.sleep()` and `wifi.suspend()` functionality for low-power usecases where the application will be putting the CPU into modem-sleep or light-sleep to save power(15mA and 0.4mA chip consumption respectively)? 

These functions are really only need if the ESP application is battery powered and needs to use the WiFi intermittently. IMO, other usecases form the bulk of ESP applications, so my recommendation this implementation _must not_ impact the runtime performance or RAM footprint of these majority usecases.   Therefore, such functionality should not be enabled by default.

It should also have as small an impact on the runtime performance or RAM footprint of applications which employ light-sleep.  

As I understand this, the SDK PWM and S/W timer functions use the FRC1 and FRC2 H/W timers for sequencing. These don't work with light-sleep as the CPU and its internal clock are halted and hence the H/W timer NMIs won't get delivered.  The 2.1 SDK API guide only mentions restrictions "timers" in  Sec 3.7 which could include FRC2 use.

The S/W timer code is currently pretty lightwieght. The caller is responsible for allocating timer structures and the API marshals these in its linked timer list. No malloc/free calls are made.

The S/W timer is used by both SDK internal and application functionality.  I assume that we only need to have a treatment for the application ones, and what we it seems that we need to do is to stop any application timers before reqesting a sleep and then resuming them with a correction offset (whatever that should be) on awaking.  This is a one per sleep / wake cycle overhead and not a once per every timer call one.

As I see it, the issue -- if we want to do this at once per sleep -- is that when we walk the timer linked list we need a method of descriminating which list entries are application ones.  We've got some 35 `xx_timer_setfn()` call in 22 C files; some 30 unique C cb addresses in our firmware, of which only a dozen might be used in a typical build.  Surely it is up the to the individual module to determine how to handle the impact of light sleep? For example:
-  let the time slip right by the sleep time
-  adjust time and drop any past timers
-  adjust time and deliver timer immediately for past timers.

Why not use a macro to allow modules to declare a policy for each cb address?  This could even use an inline static flag byte so this only calls the sleep hook once per boot, instead of once per register / unregister.  OK this isn't transparent at a coding level, but I personally would rather have 30 extra source lines to add and no other runtime RAM/CPU hit than a chunky piece of code with malloc / frees on every timer register / unregister.

marcelstoer:
> @dnc40085 Hi dnc, can we have a discussion about the requirement and implementation here?

Did this happen offline or has this stalled?

TerryE:
It hasn't happened. Sorry to be terse, but getting ready for my son's wedding tomorrow 😀

marcelstoer:
@dnc40085 can we still count on your support for this project? Your GitHub history went blank after #2079 last August...?

TerryE:
If @dnc40085 has gone onto greater things, then I guess that I'll have to add this to my long butt list because I don't think that we can leave it as it is.

dnc40085:
Hello, my apologies for being gone for so long, life got in the way.

I have a new commit that I'm putting some finishing touches on, I've implemented the changes suggested by @TerryE above.

I'll try to get it finished and push the new commits this weekend.

dnc40085:
I think I'm gonna need to create a new PR, I messed up somewhere

PR #2057, opened 2017-08-02
Add ws2812 open-drain mode

nwf:

WS2812-alikes generally expect 5V supply rails and so may not consider the 3.3V logic high of the ESP8266 to be logic high.  As such, one should consider using a level shifter.  However, the ESP8266 is 5V tolerant and supports open-drain GPIO, so a lower component-count option is to externally pull the WS2812 data line high with a resistor to 5V and have the ESP float, rather than drive, the line during logic high pulses.  In testing, this works great with the APA106 I have handy.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -10,7 +10,8 @@
 
 #define CANARY_VALUE 0x32383132
 #define MODE_SINGLE  0
-#define MODE_DUAL    1
+#define MODE_SINGLE_OPEN 1
+#define MODE_DUAL    2
 
 #define FADE_IN  1
 #define FADE_OUT 0
@@ -29,7 +30,7 @@ typedef struct {
 // You HAVE to redirect LUA's output somewhere else
 static int ws2812_init(lua_State* L) {
   const int mode = luaL_optinteger(L, 1, MODE_SINGLE);
-  luaL_argcheck(L, mode == MODE_SINGLE || mode == MODE_DUAL, 1, "ws2812.SINGLE or ws2812.DUAL expected");
+  luaL_argcheck(L, mode >= MODE_SINGLE && mode <= MODE_DUAL, 1, "ws2812 mode expected");
 
   // Configure UART1
   // Set baudrate of UART1 to 3200000
@@ -46,7 +47,11 @@ static int ws2812_init(lua_State* L) {
   }
 
   // Pull GPIO2 down
-  platform_gpio_mode(4, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
+  if (mode == MODE_SINGLE_OPEN) {
+    platform_gpio_mode(4, PLATFORM_GPIO_OPENDRAIN, PLATFORM_GPIO_FLOAT);
+  } else {
+    platform_gpio_mode(4, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
+  }
   platform_gpio_write(4, 0);
 
   // Waits 10us to simulate a reset
@@ -594,6 +599,7 @@ static const LUA_REG_TYPE ws2812_map[] =
   { LSTRKEY( "FADE_IN" ),        LNUMVAL( FADE_IN ) },
   { LSTRKEY( "FADE_OUT" ),       LNUMVAL( FADE_OUT ) },
   { LSTRKEY( "MODE_SINGLE" ),    LNUMVAL( MODE_SINGLE ) },
+  { LSTRKEY( "MODE_SINGLE_OPEN" ),    LNUMVAL( MODE_SINGLE_OPEN ) },
   { LSTRKEY( "MODE_DUAL" ),      LNUMVAL( MODE_DUAL ) },
   { LSTRKEY( "SHIFT_LOGICAL" ),  LNUMVAL( SHIFT_LOGICAL ) },
   { LSTRKEY( "SHIFT_CIRCULAR" ), LNUMVAL( SHIFT_CIRCULAR ) },
--- a/docs/en/modules/ws2812.md
+++ b/docs/en/modules/ws2812.md
@@ -23,7 +23,21 @@ Initialize UART0 (TXD0) too if `ws2812.MODE_DUAL` is set.
 `ws2812.init([mode])`
 
 #### Parameters
-- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted) or `ws2812.MODE_DUAL`
+- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted), `ws2812.MODE_SINGLE_OPEN`, or `ws2812.MODE_DUAL`
+
+WS2812-alikes generally expect 5V supply rails and when so powered may not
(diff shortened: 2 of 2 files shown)


app/modules/ws2812.c
```diff
@@ -594,6 +599,7 @@ static const LUA_REG_TYPE ws2812_map[] =
   { LSTRKEY( "FADE_IN" ),        LNUMVAL( FADE_IN ) },
   { LSTRKEY( "FADE_OUT" ),       LNUMVAL( FADE_OUT ) },
   { LSTRKEY( "MODE_SINGLE" ),    LNUMVAL( MODE_SINGLE ) },
+  { LSTRKEY( "MODE_SINGLE_OPEN" ),    LNUMVAL( MODE_SINGLE ) },
```
nickandrew: Shouldn't this be LNUMVAL(MODE_SINGLE_OPEN) ?

docs/en/modules/ws2812.md
```diff
@@ -23,7 +23,11 @@ Initialize UART0 (TXD0) too if `ws2812.MODE_DUAL` is set.
 `ws2812.init([mode])`
 
 #### Parameters
-- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted) or `ws2812.MODE_DUAL`
+- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted), `ws2812.MODE_SINGLE_OPEN`, or `ws2812.MODE_DUAL`
+
+In `ws2812.MODE_SINGLE_OPEN` you will have to pull the WS2812 data line
+high.  As the ESP8266 is 5V tolerant but 3.3V may not rise to the logic high
+level of a WS2812, this arrangement may increase reliability of the signal.
```
nickandrew: This paragraph could do with fleshing out to the high quality of your PR description.

app/modules/ws2812.c
nwf: Er, yes, of course.  That'll teach me to be quick about my edits. :(

docs/en/modules/ws2812.md
nwf: Done.  New patch forthcoming.

TerryE:
> However, the ESP8266 is 5V tolerant 

The datasheet lists Vih as 3.6V not 5V, so whilst this might work, you are running the chip out of manufacturer spec. Are you using a voltage device?

nwf:
I confess to having not consulted the datasheet, but apparently the word on the street is that the datasheet is wrong (or, at least, incomplete).

Apparently (god, I feel so dirty linking to a twitter-posted screenshot of a facebook group, but it's as close to the original source as I've found), the CEO of Espressif asserts that the GPIOs are 5V tolerant at 3.3V supply: https://twitter.com/ba0sh1/status/759239169071837184.

http://download.arduino.org/products/UNOWIFI/0A-ESP8266-Datasheet-EN-v4.3.pdf offers "All digital IO pins are protected from over-voltage with a snap-back circuit connected between the pad and ground. The snap back voltage is typically about 6V, and the holding voltage is 5.8V." but is otherwise mum on absolute maxima.
I do note that there is no maximum given for VOH, and given that open-drain is considered to be an output mode, perhaps this should be taken to mean that the chip can, in fact, tolerate up to snapback?  I don't know.

In any case, I can add a disclaimer to the docs, if you like.

docs/en/modules/ws2812.md
```diff
@@ -23,7 +23,15 @@ Initialize UART0 (TXD0) too if `ws2812.MODE_DUAL` is set.
 `ws2812.init([mode])`
 
 #### Parameters
-- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted) or `ws2812.MODE_DUAL`
+- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted), `ws2812.MODE_SINGLE_OPEN`, or `ws2812.MODE_DUAL`
+
+WS2812-alikes generally expect 5V supply rails and when so powered may not
+consider the 3.3V logic high of the ESP8266 to be logic high. As such, one
+should consider using a level shifter.  However, the ESP8266 is 5V tolerant
```
nickandrew: Perhaps "the ESP8266 I/Os are 5V tolerant"

nwf: Mm, fair.  Will revise.

docs/en/modules/ws2812.md
```diff
@@ -23,7 +23,15 @@ Initialize UART0 (TXD0) too if `ws2812.MODE_DUAL` is set.
 `ws2812.init([mode])`
 
 #### Parameters
-- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted) or `ws2812.MODE_DUAL`
+- `mode` (optional) either `ws2812.MODE_SINGLE` (default if omitted), `ws2812.MODE_SINGLE_OPEN`, or `ws2812.MODE_DUAL`
```
devsaurus: What about DUAL mode operation - can the other IO be configured to open-drain as well?

docs/en/modules/ws2812.md
```diff
@@ -23,7 +23,15 @@ Initialize UART0 (TXD0) too if `ws2812.MODE_DUAL` is set.
...
+
+WS2812-alikes generally expect 5V supply rails and when so powered may not
+consider the 3.3V logic high of the ESP8266 to be logic high. As such, one
+should consider using a level shifter.  However, the ESP8266 is 5V tolerant
+and supports open-drain GPIO, so a lower component-count option is to
+externally pull the WS2812 data line high with a resistor to 5V and have the
```
devsaurus: Please embed or link a reference schematic showing the exact configuration and resistor value.

docs/en/modules/ws2812.md
nwf: I was not sure where ws2812 currently configures the second pin; there isn't the code I expected in `ws2812_init`, so I left it be.  Yes, in principle, I think there could be a `MODE_DUAL_OPEN`.

docs/en/modules/ws2812.md
nwf: Is that really necessary?  None of the other documentation, AFAIK, contains schematics, and there are modules that talk to things much more complicated than a WS2812.

The exact value of the resistor does not matter terribly much (though I think something ~10K is typical), but it does, in principle, vary with the design of the WS2812-alike's input circuitry.

In any case, would the following ASCII-art suffice?
```
5V >----[10K]---+
                |
GPIO4 >---------+------> WS2812 Data In
```

nwf:
@devsaurus An excellent question.  I went and hooked this up to an oscilloscope and I am mystified now as to why it seemed to matter in my earlier testing.  (Truly, the ws2812 signal was unreliable and when I added this code it got better... of course, I also changed the hardware at the same time, so perhaps it was just a bad connection!)

In any case, observationally, the line rings up to about 4V or so, but no, it levels off around 3V.  I'm going to retract this PR for the moment, since it seems like `MODE_SINGLE_OPEN` is not actually doing what I anticipated.  Perhaps I misunderstand how to configure the GPIOs, or perhaps this is because the ESP8266 module I use has other circuitry (notably, an LED) connected to GPIO4.

Sorry for the noise, everyone.

PR #2060, opened 2017-08-04
LuaOTA Provisioning System

TerryE:
-  This PR is for the `dev` branch rather than for `master`.
- This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- I have thoroughly tested my contribution.

## Description of and rationale behind this PR

See README.md.  Currently the Lua examples are not documented in the documentation folder so I have included the documentation as an in-directory README.  The intro is

> Espressif use an optional update approach for their firmware know as OTA (over the air).   This module offers an equivalent facility for Lua applications developers, and enables  module development and production updates by carrying out automatic synchronisation  with a named provisioning service at reboot.
> 
> This `luaOTA` provisioning service uses a different approach to  [enduser setup](https://nodemcu.readthedocs.io/en/dev/en/modules/enduser-setup/). The basic concept here is that the ESP modules are configured with a pre-imaged file system that includes a number of files in the luaOTA namespace.  (SPIFFS doesn't implement a directory hierarchy as such, but instead simply treats the conventionaldirectory separator as a character in the filename.  Nonetheless, the "luaOTA/" prefix serves to separate the lc files in the luaOTA namespace.)

### DIFF
--- a/lua_examples/luaOTA/_doTick.lua
+++ b/lua_examples/luaOTA/_doTick.lua
@@ -1,0 +1,76 @@
+tmr.stop(0)--SAFETRIM 
+-- function _doTick(self)
+
+  -- Upvals
+  local self                = ...
+  local wifi,net            = wifi,net 
+  local sta                 = wifi.sta
+  local config,log,startApp = self.config,self.log,self.startApp
+  local tick_count          = 0
+  
+  local function socket_close(socket) --upval: self, startApp
+    if rawget(self,"socket") then
+      self.socket=nil                        -- remove circular reference in upval 
+      pcall(socket.close,socket)
+      return startApp("Unexpected socket close")
+    end
+  end
+
+  local function receiveFirstRec(socket, rec)  -- upval: self, crypto, startApp, tmr
+    local cmdlen   = (rec:find('\n',1, true) or 0) - 1
+    local cmd,hash = rec:sub(1,cmdlen-6), rec:sub(cmdlen-5,cmdlen)
+    if cmd:find('"r":"OK!"',1,true) or cmdlen < 16 or
+      hash ~= crypto.toHex(crypto.hash("MD5",self.secret .. cmd):sub(-3)) then
+      print "No provsioning changes required"
+      self.socket = nil
+      self.post(function()  --upval: socket
+                  if socket then pcall(socket.close, socket) end 
+                end)
+      return startApp("OK! No further updates needed")
+    end
+    -- Else a valid request has been received from the provision service free up
+    -- some resources that are no longer needed and set backstop timer for general 
+    -- timeout.  This also dereferences the previous doTick cb so it can now be GCed.
+    collectgarbage()
+    tmr.alarm(0, 30000, tmr.ALARM_SINGLE, self.startApp)
+    return self:_provision(socket,rec)
+  end
+
+  local function socket_connect(socket) --upval: self, socket_connect
+    print "Connected to provisioning service"
+    self.socket = socket
+    socket_connect = nil                  -- make this function available for GC
+    socket:on("receive", receiveFirstRec)
+    return self.socket_send(socket, self.config)
+  end
+
+  local conn 
+  return function() -- the proper doTick() timer callback
+    tick_count = tick_count + 1
+    log("entering tick", tick_count, sta.getconfig(false), sta.getip())
+    if (tick_count < 20) then             -- (wait up to 10 secs for Wifi connection)
+      local status, ip = sta.status(),{sta.getip()}
+      if (status == wifi.STA_GOTIP) then
+        log("Connected:", unpack(ip))
+        if (config.nsserver) then
+          net.dns.setdnsserver(config.nsserver, 0)
+        end
+        conn = net.createConnection(net.TCP, 0)
+        conn:on("connection",     socket_connect)
+        conn:on("disconnection",  socket_close)
+        conn:connect(config.port, config.server)
+
+        tick_count = 20
+      end
+
+    elseif (tick_count == 20) then        -- assume timeout and exec app CB
+      return self.startApp("OK: Timeout on waiting for wifi station setup")
+
+    elseif (tick_count == 26) then        -- wait up to 2.5 secs for TCP response
+      tmr.unregister(0)
+      pcall(conn.close, conn)
+      self.socket=nil
+      return startApp("OK: Timeout on waiting for provision service response")
+    end
+  end
+-- end
--- a/lua_examples/luaOTA/_doTick.lua
+++ b/lua_examples/luaOTA/_doTick.lua
@@ -21,7 +21,7 @@ tmr.stop(0)--SAFETRIM
     local cmd,hash = rec:sub(1,cmdlen-6), rec:sub(cmdlen-5,cmdlen)
     if cmd:find('"r":"OK!"',1,true) or cmdlen < 16 or
       hash ~= crypto.toHex(crypto.hash("MD5",self.secret .. cmd):sub(-3)) then
-      print "No provsioning changes required"
+      print "No provisioning changes required"
       self.socket = nil
       self.post(function()  --upval: socket
                   if socket then pcall(socket.close, socket) end 
--- a/lua_examples/luaOTA/_doTick.lua
+++ b/lua_examples/luaOTA/_doTick.lua
@@ -20,7 +20,7 @@ tmr.stop(0)--SAFETRIM
     local cmdlen   = (rec:find('\n',1, true) or 0) - 1
     local cmd,hash = rec:sub(1,cmdlen-6), rec:sub(cmdlen-5,cmdlen)
     if cmd:find('"r":"OK!"',1,true) or cmdlen < 16 or
-      hash ~= crypto.toHex(crypto.hash("MD5",self.secret .. cmd):sub(-3)) then
+      hash ~= crypto.toHex(crypto.hmac("MD5", cmd, self.secret):sub(-3)) then
       print "No provisioning changes required"
       self.socket = nil
       self.post(function()  --upval: socket
--- a/lua_examples/luaOTA/check.lua
+++ b/lua_examples/luaOTA/check.lua
@@ -1,0 +1,66 @@
+--SAFETRIM
+--------------------------------------------------------------------------------
+-- LuaOTA provisioning system for ESPs using NodeMCU Lua
+-- LICENCE: http://opensource.org/licenses/MIT
+-- TerryE  15 Jul 2017
+--
+-- See luaOTA.md for description and implementation notes
+
+--------------------------------------------------------------------------------
+
+-- upvals
+local crypto, file, json,  net, node, table, tmr, wifi =
+      crypto, file, sjson, net, node, table, tmr, wifi
+local error, pcall   = error, pcall 
+local loadfile, gc   = loadfile, collectgarbage
+local concat, unpack = table.concat, unpack or table.unpack 
+
+local self = {post = node.task.post, prefix = "luaOTA/", conf = {}}
+
+self.log     = (DEBUG == true) and print or function() end 
+self.modname = ...  
+
+--------------------------------------------------------------------------------------
+-- Utility Functions
+
+setmetatable( self, {__index=function(self, func) --upval: loadfile
+    -- The only __index calls in in LuaOTA are dynamically loaded functions. 
+    -- The convention is that functions starting with "_" are treated as 
+    -- call-once / ephemeral; the rest are registered in self 
+    func = self.prefix .. func
+    local f,msg = loadfile( func..".lc")
+    if msg then f, msg = loadfile(func..".lua") end
+    if msg then error (msg,2) end
+    if func:sub(8,8) ~= "_" then self[func] = f end
+    return f
+  end} )
+
+function self.sign(arg)  --upval: crypto, json, self
+  arg = json.encode(arg)
+  return arg .. crypto.toHex(crypto.hash("MD5", self.secret .. arg):sub(-3)) .. '\n'
+end
+
+function self.startApp(arg) --upval: gc, self, tmr, wifi
+  gc();gc()
+  tmr.unregister(0)
+  self.socket = nil
+  if not self.config.leave then wifi.setmode(wifi.NULLMODE,false) end
+  local appMod    = self.config.app   or "luaOTA.default"
+  local appMethod = self.config.entry or "entry"
+  if not arg then arg =  "General timeout on provisioning" end
+  self.post(function() --upval: appMod, appMethod, arg 
+              require(appMod)[appMethod](arg)
+            end)
+end 
+
+function self.socket_send(socket, rec, opt_buffer)
+  return socket:send(self.sign(rec) .. (opt_buffer or ''))
+end
+
+self.post(function() -- upval: self
+            -- This config check is to prevent a double execution if the 
+            -- user invokes with "require 'luaOTA/check':_init( etc>)" form  
+            if not rawget(self, "config") then self:_init() end 
+          end)
+
+return self
--- a/lua_examples/luaOTA/check.lua
+++ b/lua_examples/luaOTA/check.lua
@@ -37,7 +37,7 @@ setmetatable( self, {__index=function(self, func) --upval: loadfile
 
 function self.sign(arg)  --upval: crypto, json, self
   arg = json.encode(arg)
-  return arg .. crypto.toHex(crypto.hash("MD5", self.secret .. arg):sub(-3)) .. '\n'
+  return arg .. crypto.toHex(crypto.hmac("MD5", arg, self.secret):sub(-3)) .. '\n'
 end
 
 function self.startApp(arg) --upval: gc, self, tmr, wifi
--- a/lua_examples/luaOTA/luaOTAserver.lua
+++ b/lua_examples/luaOTA/luaOTAserver.lua
@@ -164,7 +164,7 @@ receive_and_parse = function(esp)
   local packed_cmd, sig = line:sub(1,#line-6),line:sub(-6)
 -- print("reply:", packed_cmd, sig)
   local status, cmd = pcall(json.decode, packed_cmd)
-  if hmac(packed_cmd):sub(-6) == sig then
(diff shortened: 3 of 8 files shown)


TerryE:
This is very much a work in progress, but given some of the recent developer feedback, then I am offering now because I think it quite timely, even it its current form.   The server is stripped out one from what I actually use for my projects, but my live version has a lot of extra project specific stuff that we just be confusing.  

As a Lua example, I think that is has two strong  advantages for its inclusion:

-  The auotmatic sync model of doing updates and configuring devices is just so much easier for most projects, IMO.  This pretty much removes the need to use esplorer and simplifies production development.  I use an PRi3 with a WD PiDrive for my always-on provisioning server, and I just edit the files on my laptop (through an SSH tunnels sftp share) and the changes get automatically cascaded onto the devices.

-  We don't seem to have many good examples of how to write scalable Lua apps.  The self autoloader is very scalable, and leave no aggregated RAM footprint.  I use the same technique for my actual apps.

At this stage I would welcome any feedback, questions, or comments.

NicolSpies:
Thanks for the example how to write scalable Lua apps, I am always keen to investigate and refactor what I have.

I devised an OTA system using bitbucket requiring SSL resulting in the requirement to use the TLS module. I ran into receive buffer size limitation with larger (few KB) lua files. My way around the limitation is to split the files to 1KB, send across, stitch together and "flash" to single function containing .lc files using flashmod.

Are there any limitations on the size of the .lc files your provisioning system can OTA ?

TerryE:
@NicolSpies Nicol,

> I devised an OTA system using bitbucket requiring SSL resulting in the requirement to use the TLS module. I ran into receive buffer size limitation with larger (few KB) lua files.

IMO, Whilst IoT devices run a bare-bones OS or SDK without proper protection and authentication services, they should stay on local Intranet and have no direct access to Internet services, push or pull. All of mine only "talk" to a management server which can do outward connections to the Internet.  Inbound requests come through a gateway server.  In this usecase, I am not really interested in command protocol level privacy,  though I am interest in trusted communication, which is why I plump for a signature approach based on a shared secret, because this avoids all of the TLS overhead.
  
> My way around the limitation is to split the files to 1KB, send across, stitch together and "flash" to single function containing .lc files using flashmod.

Using large memory resources is bad news in the ESP8266 and should really be avoided because this fragments the heap space badly, and large allocs can fail even when you've notionally got 15K or more of notional RAM  free.  If you look, you will see that I also use 1K chunks and use a reader in my load to aavoid needing to concatenate these.  The `lc` file is still a single chunk though.

If you note: once the ESP goes into re-provisioning mode, it always reboots on completion, so "production boot are always clean and not contaminated by heap fragmentation from compilation. 

> Are there any limitations on the size of the .lc files your provisioning system can OTA ?

I haven't stress tested this to see how large an `lc`file I can load,  as I tend to keep mine down to a max of 3-4K anyway and just dynamically load them as needed.  I can still use `luac.cross` to load larger `'lc` files with a chunk-write apptoach, the same way I do my uploads, but my attitude is "why bother?" which is why I hadn't coded this case.

NicolSpies:
@TerryE Terry,

I like the gateway server security paradigm when a site has a few endpoints, it however adds cost when only one endpoint per remote site exists.

You gave me a good idea, instead of sending lua files across and then using flasmod on the remote ESP to generate the individual lc files, I could do that upfront and OTA the myriad of small single function flashmod generated lc files.

All this will "go away" when using the ESP32 but extra RAM is no excuse for sloppy code and bad practices.

lua_examples/luaOTA/_doTick.lua
```diff
@@ -0,0 +1,76 @@
...
+      pcall(socket.close,socket)
+      return startApp("Unexpected socket close")
+    end
+  end
+
+  local function receiveFirstRec(socket, rec)  -- upval: self, crypto, startApp, tmr
+    local cmdlen   = (rec:find('\n',1, true) or 0) - 1
+    local cmd,hash = rec:sub(1,cmdlen-6), rec:sub(cmdlen-5,cmdlen)
+    if cmd:find('"r":"OK!"',1,true) or cmdlen < 16 or
+      hash ~= crypto.toHex(crypto.hash("MD5",self.secret .. cmd):sub(-3)) then
```
nwf: md5 is a weak hash and this use is open to extension attacks (that is, given a signed command "foo", I can generate a signed command "foobar" without knowing the secret.)  Please use something stronger, like crypto.hmac('sha1', cmd, self.secret).  (Comment applies in several places.)

Consider using base64, rather than hex, for a slightly shorter hash, too?

TerryE: @nwf Nathan, yes MD5 is weak hash and I would no longer use it for password mangling on an application such as mediaWiki or a forum package such as phpBB3 ot InvsionPower (and I administer all of these, BTW), but it is only weak in the sense that it can be brute-force attacked in multistream processors such as a CUDA architecture; hack in and get a password file of say 10K accounts then the attacker would probably crack perhaps 30-50% of typical user passwords in an 24hr compute.  If the IoT device and the server used say "wTRfe$523'dfaw;WEqwe2" then hell would freeze over before the attacker cracked it. Yuo don't have scale advantages of the "attack N thousand at once).  Incidentally  I originally used SHA1, but MD5 is in most Lua PC and \*nix repositories,  SHA1 isn't and I wanted to keep this simple.

nwf: Well, at least consider the use of an HMAC construction to prevent the extension attack.

TerryE: The change on the client side is quite straightforward, but not so on the server-side. We'd need the user to understand how to use luarocks and install a package such as **luaossl**  So why not suggest a patch?

lua_examples/luaOTA/luaOTAserver.lua
```diff
@@ -0,0 +1,240 @@
...
+
+    assert(fa, "File %s is required but not in sources directory" % name)
+    fp[#fp+1] = name .. ":" .. fa.modification
+    f[i] = {name = name, mtime = fa.modification, 
+            size = fa.size, content = read_file(fullname) }
+    assert (f[i].size == #(f[i].content or ''), "File %s unreadable" % name )
+  end
+
+  assert (#f == #fp, "Aborting provisioning die to missing fies",0)
+  return inventory, md5.sumhexa(concat(fp,":"))
```
nwf: To perform a HMAC-MD5 with no additional packages, it suffices to use
```
function hmd5(key, data)
  if key:len() > 64 then key = md5.sum(key) end
  key = key .. string.rep('\0',64-key:len())
  local ki = md5.exor(string.rep(string.char(0x36),64),key)
  local ko = md5.exor(string.rep(string.char(0x5c),64),key)
  return md5.sumhexa(ko..md5.sum(ki..data))
end
```

This function passes all the RFC2022 (https://tools.ietf.org/html/rfc2202) test vectors, as can be verified by running

```
print(hmd5(string.rep(string.char(0x0B),16), "Hi There") == "9294727a3638bb1c13f48ef8158bfc9d")
print(hmd5("Jefe", "what do ya want for nothing?") == "750c783e6ab0b503eaa86e310a5db738")
print(hmd5(string.rep(string.char(0xaa),16), string.rep(string.char(0xdd),50)) == "56be34521d144c88dbb8c733f0e8b3f6")
print(hmd5("\1\2\3\4\5\6\7\8\9\10\11\12\13\14\15\16\17\18\19\20\21\22\23\24\25",string.rep(string.char(0xcd),50)) == "697eaf0aca3a3aea3a75164746ffaa79")
print(hmd5(string.rep(string.char(0x0c),16),"Test With Truncation") == "56461ef2342edc00f9bab995690efd4c")
print(hmd5(string.rep(string.char(0xaa),80),"Test Using Larger Than Block-Size Key - Hash Key First") == "6b1ab7fe4bd7bf8f0b62e6ce61b9d0cd")
print(hmd5(string.rep(string.char(0xaa),80),"Test Using Larger Than Block-Size Key and Larger Than One Block-Size Data") == "6f630fad67cda0ee1fb1f562db3aa53e")
```

lua_examples/luaOTA/check.lua
```diff
@@ -0,0 +1,66 @@
...
+    local f,msg = loadfile( func..".lc")
+    if msg then f, msg = loadfile(func..".lua") end
+    if msg then error (msg,2) end
+    if func:sub(8,8) ~= "_" then self[func] = f end
+    return f
+  end} )
+
+function self.sign(arg)  --upval: crypto, json, self
+  arg = json.encode(arg)
+  return arg .. crypto.toHex(crypto.hash("MD5", self.secret .. arg):sub(-3)) .. '\n'
```
nwf: To perform a HMAC-MD5 with no additional packages, it suffices to use
```
function hmd5(key, data)
  if key:len() > 64 then key = md5.sum(key) end
  key = key .. string.rep('\0',64-key:len())
  local ki = md5.exor(string.rep(string.char(0x36),64),key)
  local ko = md5.exor(string.rep(string.char(0x5c),64),key)
  return md5.sumhexa(ko..md5.sum(ki..data))
end
```

This function passes all the RFC2022 (https://tools.ietf.org/html/rfc2202) test vectors, as can be verified by running

```
print(hmd5(string.rep(string.char(0x0B),16), "Hi There") == "9294727a3638bb1c13f48ef8158bfc9d")
print(hmd5("Jefe", "what do ya want for nothing?") == "750c783e6ab0b503eaa86e310a5db738")
print(hmd5(string.rep(string.char(0xaa),16), string.rep(string.char(0xdd),50)) == "56be34521d144c88dbb8c733f0e8b3f6")
print(hmd5("\1\2\3\4\5\6\7\8\9\10\11\12\13\14\15\16\17\18\19\20\21\22\23\24\25",string.rep(string.char(0xcd),50)) == "697eaf0aca3a3aea3a75164746ffaa79")
print(hmd5(string.rep(string.char(0x0c),16),"Test With Truncation") == "56461ef2342edc00f9bab995690efd4c")
print(hmd5(string.rep(string.char(0xaa),80),"Test Using Larger Than Block-Size Key - Hash Key First") == "6b1ab7fe4bd7bf8f0b62e6ce61b9d0cd")
print(hmd5(string.rep(string.char(0xaa),80),"Test Using Larger Than Block-Size Key and Larger Than One Block-Size Data") == "6f630fad67cda0ee1fb1f562db3aa53e")
```

lua_examples/luaOTA/luaOTAserver.lua
```diff
@@ -0,0 +1,240 @@
...
+-- Encode a response buff, add a signature and any optional buffer
+------------------------------------------------------------------
+send_command = function(esp, resp, buffer)
+  if type(buffer) == "string" then 
+    resp.data = #buffer 
+  else 
+    buffer = ''
+  end
+  local rec = json.encode(resp)
+  rec = rec .. md5.sumhexa(secret .. rec):sub(-6) .."\n"
```
nwf: To perform a HMAC-MD5 with no additional packages, it suffices to use
```
function hmd5(key, data)
  if key:len() > 64 then key = md5.sum(key) end
  key = key .. string.rep('\0',64-key:len())
  local ki = md5.exor(string.rep(string.char(0x36),64),key)
  local ko = md5.exor(string.rep(string.char(0x5c),64),key)
  return md5.sumhexa(ko..md5.sum(ki..data))
end
```

This function passes all the RFC2022 (https://tools.ietf.org/html/rfc2202) test vectors, as can be verified by running

```
print(hmd5(string.rep(string.char(0x0B),16), "Hi There") == "9294727a3638bb1c13f48ef8158bfc9d")
print(hmd5("Jefe", "what do ya want for nothing?") == "750c783e6ab0b503eaa86e310a5db738")
print(hmd5(string.rep(string.char(0xaa),16), string.rep(string.char(0xdd),50)) == "56be34521d144c88dbb8c733f0e8b3f6")
print(hmd5("\1\2\3\4\5\6\7\8\9\10\11\12\13\14\15\16\17\18\19\20\21\22\23\24\25",string.rep(string.char(0xcd),50)) == "697eaf0aca3a3aea3a75164746ffaa79")
print(hmd5(string.rep(string.char(0x0c),16),"Test With Truncation") == "56461ef2342edc00f9bab995690efd4c")
print(hmd5(string.rep(string.char(0xaa),80),"Test Using Larger Than Block-Size Key - Hash Key First") == "6b1ab7fe4bd7bf8f0b62e6ce61b9d0cd")
print(hmd5(string.rep(string.char(0xaa),80),"Test Using Larger Than Block-Size Key and Larger Than One Block-Size Data") == "6f630fad67cda0ee1fb1f562db3aa53e")
```

(Sorry, took a moment to land this at the relevant line in the changes!)

lua_examples/luaOTA/luaOTAserver.lua
```diff
@@ -0,0 +1,240 @@
...
+
+
+-- Decode a response buff, check the signature and any optional buffer
+----------------------------------------------------------------------
+receive_and_parse = function(esp)
+  local line = esp:receive("*l")
+  local packed_cmd, sig = line:sub(1,#line-6),line:sub(-6)
+-- print("reply:", packed_cmd, sig)
+  local status, cmd = pcall(json.decode, packed_cmd)
+  if md5.sumhexa(secret .. packed_cmd):sub(-6) == sig then
```
nwf: As above, re: HMAC.

lua_examples/luaOTA/luaOTAserver.lua
TerryE: '@nwf Nathaniel, thanks for this.  You are right, it is a more elegant approach. I will make these changes tomorrow and push them through once I am sure that I've got this implemented and tested correctly. :smile:

TerryE:
Poke for @marcelstoer, et al.  This is a Lua example.  IMO, useful and better than many of the existing examples that  we supply.  Any reason for not committing it?

marcelstoer:
> Any reason for not committing it?

I was able to set aside enough time to study this. If this is a good at it seems (I have no reason to doubt that) shouldn't we promote/advertise it prominently? How about linking to it from the documentation or even README?

TerryE:
In terms of the documentation, we don't have an a standard approach for the Lua examples. I included a README.md in this subdirectory. Should we do this or use docs? Re you happy to link the into the documentation YAML?  I defer to you views here.

marcelstoer:
>  I included a README.md in this subdirectory

I think this is great and appears to be a pattern others before you have been following. However, if we don't advertise the existence of Lua OTA it may be hard to find for people who don't actively look for it. If there was a mention somewhere in our documentation (FAQ maybe?) or even on https://github.com/nodemcu/nodemcu-firmware/blob/master/README.md we could help adoption.

Btw, https://github.com/TerryE/nodemcu-firmware/tree/dev-luaOTA/lua_examples/luaOTA#initlua shows some markdown syntax issue.

TerryE:
Since I've started hammering my working version of this code, I have made a couple of key changes:

1.  I have switched to server-side `luac.cross` compilation.  The reason for this is RAM constraints during provisioning.  The core provisioning code takes up about 15Kb when running, and on-~ESP8266 compilation sees to add at least a 2-3× overhead so the largest single module that you can compile under the provisioning system is around 4Kb, which is a PITA.  Moving all compilation to the host removes this overhead so you can use compilation units up to ~10Kb happily which makes code sizes so much more convenient.

2.  For some reason that I have yet to understand (and I would appreciate any suggestions), connecting to the WiFi takes about 4-5secs (my RPis take about a second).  It's the WiFi connection that takes the time, not the provisioning server dialogue.  So I've optimized the process to use a flag in RTCmem and setting this to non-zero triggers a provisioning check.

Even so, this works really well, in that you can pretty much transparently rsync your chips to a shadow PC directory and avoid all the Esplorer bollocks.

If there is any interest, then I can push these changes as well.

NicolSpies:
@TerryE , in my case, in order to also score on the provisioning overhead, I suppose I can try and reduce the resources used by flashmod or stop using it altogether and standardize on technique used in your "reference" example. I will have to investigate how your standard approach differs from the flashmod approach. I assume yours do not have the overheads.

Please push the changes, always interest from my side regarding refactoring and learning from others.

TerryE:
@NicolSpies Nicol, we are really doing a segue into another topic or issue, and that is the best way to code so that the minimum resources are tidied up in RAM but without making the application coding unwieldy.  IMO, Flashmod is too restrictive and just using Lua correctly gives as good if not better RAM utilisation.  So long as you correctly orphan collectable references, the GC is pretty brutal is cleaning out RAM.

I tend to use locals and upvalues a lot since these are _fast_ (since they remove rotable lookup access) and they also have a lean runtime footprint.  

I also keep a very close eye on global and registry references which can cause horrible marking cascades which prevent GCing.

I use `lc` files as these avoid the compile hit in terms of RAM (and luaOTA does this seemlessly) , but avoid those generated by `node.compile` since these discard the line number info which is a total pain for debugging, (since my LCD patch, keeping line numbers has a very low memory footprint anyway).   

I also tend to use modules for encapsulation even though this carries a table overhead, and use autoload techniques.  My shortest autoloader is:
```Lua
setmetatable(self, {__index=function(this,value) 
  local r = package.loaders[2](value)  -- this is the lua loader which resolves using package.path
  if type(r) == "function" then return r end 
end})  -- the set binding is 6 instructions and the function 7
```
This does the load on every resolution but if you are doing something like:
```
local phase2 = self.phase2
```
then this doesn't matter since the closure is cached in the local phase2 and it will be automatically referenced when the local goes out of scope.

nwf:
@TerryE I'd like to see your changes if for no other reason than as a worked example of server-side luac.cross use.  I slam into compilation E:M failures so often that I've taken to just pushing the lua files and rebooting the devices with an init.lua that looks for lua files to compile.

TerryE:
@nwf, I am on GMT so its beddy-byes now.  I'll do the push tomorrow.   Your doing essentially what I did in the current luaOTA, but don't use node.compile for code that isn't 100% stable as this strips out line number metadata.  My LCD patch dropped the over head from 4byte / Instruction to ~1 byte per non-blank source line which typically is a 15-30× saving.

TerryE:
@marcelstoer, I've hammering this and make a few changes to improve performance.  I'll issue a new PR to cover this work.

PR #2064, opened 2017-08-05
Comment out pmsleep and timer_suspend options in user_config.h

dnc40085:

This PR comments out `ENABLE_TIMER_SUSPEND` and `PMSLEEP_ENABLE` in `user_config.h`.
These features currently have some issues that will take some time to resolve and to prevent further delay of the master drop, the features have been disabled in `user_config.h`

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -114,8 +114,8 @@ extern void luaL_assertfail(const char *file, int line, const char *message);
 #define WIFI_SDK_EVENT_MONITOR_ENABLE
 #define WIFI_EVENT_MONITOR_DISCONNECT_REASON_LIST_ENABLE
 
-#define ENABLE_TIMER_SUSPEND
-#define PMSLEEP_ENABLE
+////#define ENABLE_TIMER_SUSPEND
+//#define PMSLEEP_ENABLE
 
 
 #define STRBUF_DEFAULT_INCREMENT 32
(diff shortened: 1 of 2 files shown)


TerryE:
Oops, it looks as is we've missed some conditionals around calling the `swTimer` routines in `tmr.c`

dnc40085:
@TerryE that's weird, it compiled fine locally for both float and integer versions

marcelstoer:
Doesn't this effectively render the improvements you made in 0844a5d moot and break backwards compatibility?

TerryE:
What this does is to make the pmSleep functionality disabled by default.  Any developer that wants it only needs to uncomment the two `#define` statements.  This a quick change that can be incorporated in the dev->master drop that you deferred.   

The last dev->master drop slipped in some fundamental changes to how our timer work.  This was a mistake, IMO; if making these changes configurable but at least not enabled by default to correct this mistake is a break in backwards compatibility,  then it's the sort of break that we should still be doing.

marcelstoer:
I wasn't trying to suggest this be changed somehow but trying to verify my assessment of the impact. To conclude:
- We're all good for the master drop.
- It really seems pointless for a function to print "this was removed" when the function isn't even compiled in. Not that this _couldn't_ be solved if desired...

marcelstoer:
In hind sight, why didn't it occur to us that this change should be reflected in the tmr documentation? *scratching my head*...

melstav:
As someone who recently wasted a not insignificant amount of time trying to figure out why calls to node.sleep kept erroring out, I have to say that I concur with the previous comment.

PR #2066, opened 2017-08-06
Makefile: optionally wrap the C compiler invocations

nwf:

If `ccache` is installed, use it when building the firmware.  This dramatically speeds up the build, at least on my system.  Using @marcelstoer's docker image modified to include `ccache` and an otherwise unmodified checkout, `time`-ing the first build took just over 90 seconds while the second, with a completely hot cache, took a just over 25.  In comparison, a completely stock build without `ccache` at all took 78 seconds.

I believe there are no documentation changes necessary since this is an opportunistic use of `ccache`.  To be sure, I built the docker builder without ccache installed and tested that things built even with this change in place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,9 +76,9 @@ else
 	endif
 	CCFLAGS += -ffunction-sections -fno-jump-tables -fdata-sections
 	AR = xtensa-lx106-elf-ar
-	CC = xtensa-lx106-elf-gcc
+	CC = $(WRAPCC) xtensa-lx106-elf-gcc
 	NM = xtensa-lx106-elf-nm
-	CPP = xtensa-lx106-elf-cpp
+	CPP = $(WRAPCC) xtensa-lx106-elf-gcc -E
 	OBJCOPY = xtensa-lx106-elf-objcopy
 	FIRMWAREDIR = ../bin/
     UNAME_S := $(shell uname -s)


nwf:
The travis build failures are mysterious to me, but I encountered them locally in testing, too.  I fixed them for me by "the other" commit in marcelstoer/docker-nodemcu-build#28, which, among other things, makes /root a symlink to /tmp so that one doesn't get EPERM when attempting to traverse these hard-coded? paths.

drawkula:
Using `ccache` does not need changes:
```
nodemcu-firmware$ alias quiet
alias quiet='>/dev/null 2>/dev/null '
nodemcu-firmware$ git branch 
* dev
  master
nodemcu-firmware$ quiet make clean
opt/nodemcu-firmware$ ccache -z
Statistics cleared
nodemcu-firmware$ quiet make CC='ccache xtensa-lx106-elf-gcc' CPP='ccache xtensa-lx106-elf-g++'
nodemcu-firmware$ ccache -s
cache directory                     /home/yeti/.ccache
cache hit (direct)                   406
cache hit (preprocessed)               0
cache miss                             5
called for link                        1
unsupported source language            1
unsupported compiler option          412
files in cache                     66249
cache size                           1.8 Gbytes
max cache size                       2.0 Gbytes
```

nwf:
I consider it slightly cleaner for the Makefile to probe for itself rather than to require someone to repeat "xtensa-lx106-elf-gcc" and "xtensa-lx106-elf-g++", strings the Makefile already knows, on the command line.  DRY and all that.

drawkula:
> I consider it slightly cleaner for the Makefile to probe for itself rather than to require someone to repeat "xtensa-lx106-elf-gcc" and "xtensa-lx106-elf-g++", strings the Makefile already knows, on the command line.

I consider it cleaner not to add features already existing to the code again and again, especially if it turns into an automatism taking control out of my hands. It would need an "ignore `ccache`" flag to be fully equivalent my way doing it.

Just add some few lines to the documentation about using `ccache` and `distcc` and this "problem" is solved.

nwf:
Done as suggested, though the tone of your comments is unnecessarily harsh, @drawkula.

I am still tempted to add `WRAPCC?=$(shell command -v ccache 2>/dev/null)` so that it defaults to using the cache (and could be overridden with `make WRAPCC=`), but I'll let people weigh in on that.  Meanwhile, I will go set WRAPCC in docker-nodemcu-build#28.

drawkula:
>  though the tone of your comments is unnecessarily harsh, @drawkula.

That may be a translation problem. I'm not born with an english tounge.

TerryE:
This might be a useful option, but I would not support it as a default feature. My reasons are:
-  The biggest pain / time delay in doing rebuilds of the firmware is flashing the bins, and this does nothing to help this.
-  Most builds in practice are incremental, so any marginal benefit collapses here.
-  I use a lean VBox VM running inside my Ubuntu Dev laptop.  I like to keep the VHD container small.  I don't really want the overhead of a 2Gb cache.
So for me: no material benefit.

nwf:
I use @marcelstoer's build system and every build ends up being a full build, so ccache is amazingly beneficial.  I suspect something about the mechanism by which the compilers are unpacked.  I believe that incremental builds are less impacted by ccache.

The most recent changes should be a no-op by default, and the default I proposed would only kick in if ccache was installed.  I am OK with the default being to leave WRAPCC unset.

marcelstoer:
This looks well thought out to me and I see how this will beneficial for all who use my Docker image.

So 👍 from my side.

TerryE:
Why not have a target or ENV variable enable it? That we the docker builds could use it by default, but people that didn't want to use it wouldn't have to start modifying makefiles.  Though to be honest, I always modify makefiles anyway, but for reasons outside this issue.

drawkula:
I'm perfectly happy with...

    make CC='ccache xtensa-lx106-elf-gcc' CPP='ccache xtensa-lx106-elf-g++'

...and do not fully understand why a change is needed at all.

nwf:
@drawkula That invocation continues to function correctly, so you are free to use it.  We've been over this.

TerryE:
IMO, @drawkula / Yeti's argument is the stronger.  Why change the Makefile if we don't need to to do a ccache based Docker build?

nwf:
Man, apparently Makefiles are touchy subjects.  This two-line, default no-op change really brings opinions out of the woodwork.  I'll state mine more fully and, if people are not convinced, people should feel free to close the PR.

We have Makefiles at all because they allow us not to repeat ourselves; they are a reasonably modular, declarative mechanism for managing the task of building an artifact out of smaller pieces.   We could eliminate the Makefile entirely for a shell script, or running the compiler invocations by hand, or, for that matter, executing the compiler by means of pencil and paper.  I expect that everyone would find these proposals varying degrees of wrong, and that the repetition required would be some part of the argument in all cases.

The nodemcu Makefile is "the" place that knows that the names of the compilers are "xtensa-lx106-elf-g{cc,++}" and that is, I think, the right place for such knowledge.  Adding a wrapper variable of this form allows other components to *modularly* consume that knowledge and transform it, while repeating the name of the compiler on the command line involves duplicating the compiler's name, either on the command line or in another script.  Thus, if we were to transition to, say, Clang or some other C compiler, multiple things would need to be updated.

However, in writing this response I have doubtless spent more time than the modularity would save me or anyone else, so, so much for that. :)

marcelstoer:
> This two-line, default no-op change really brings opinions out of the woodwork.

I know that you know that this is one of the "joys" of working on open-source projects. We're all entitled to our opinions and we're encouraged to make ourselves heard.

> I'll state mine more fully and, if people are not convinced, people should feel free to close the PR.

To a certain degree I share your frustration. Very often _I_ tend to be the one around here who eventually merges or closes a PR if there have been long discussions that seem to lead nowhere. That's a bit unfortunate for based on my (relevant) skills I'm the least likely of the maintainers to be in that position. However, I do understand that out of respect for the opinions of the rest of gang we/I have to be carefully not to rush ahead too early.

PR #2071, opened 2017-08-12
DHT12 Fix

### LINKED ISSUE #2050: dht_read() reads data in a wrong way?

AdrianDroid:
Fixes #2050.

- [] I have thoroughly tested my contribution.
- [] The code changes are reflected in the documentation at `docs/en/*`.

Due to shorthand implementation of DHT11,  DHT12 will fail to DHTxx logic.
However, DHT12 use the same data representations, thru DHT12 will have the wrong reading when using dht_read().
and shorthand implementation of DHT11 of its checksum will make dht_read11() fail in DHT 12.
this fix allows DHT12 user to call dht_read11() to have the correct reading.

### DIFF
--- a/app/dhtlib/dht.c
+++ b/app/dhtlib/dht.c
@@ -41,6 +41,7 @@
 #endif /* ifndef HIGH */
 
 #define COMBINE_HIGH_AND_LOW_BYTE(byte_high, byte_low)  (((byte_high) << 8) | (byte_low))
+#define COMBINE_INT_AND_FRACTION(integer,  fraction)  ((integer * 10) + (fraction))
 
 static double dht_humidity;
 static double dht_temperature;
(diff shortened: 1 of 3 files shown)


marcelstoer:
You introduced a new Lua function `dht.read12` but the documentation is missing. Why?
You didn't tick the "I have thoroughly tested my contribution" statement above. You sure about this i.e. you _haven't_ thoroughly tested?

marcelstoer:
No feedback from author in a month -> closing

Issue #806, opened 2015-12-01
Potential PR: over-the-air upgrade support


jmattsson:
So over at [DiUS](http://www.dius.com.au) we're currently working on getting some over-the-air upgrade support into our NodeMCU. It's something that could be contributed back upstream if there's a real interest in it (and thus get it tested even more), so I'm wanting to gauge said interest. Before you get all excited, let me tell you the trade-offs that come with it.

First of all, we're **not basing it around the official SDK's FOTA support**. After researching it, we decided against using it because:
- it is not well documented
- there is no evidence of any support for automatic rollback to the previous firmware in case of a bad upgrade
- it requires using different linker scripts
- you have\* to build two different firmwares, one for slot 0 and another for slot 1, then pick the correct firmware when you're upgrading
- there is no access to source code for the boot loader
- the FOTA boot loader appears to have a lot of legacy/backwards-compatibility support in it, which in our eyes increases the risk of bugs in it

Because of this, we have decided to implement something which:
- uses a minimal boot-loader
- has the boot-loader set up the hardware watchdog before booting NodeMCU, to catch really broken upgrades
- has a concept of "in-test" for a new firmware, and a maximum number of boots allowed before the system rolls back to the previous firmware version
- requires active acknowledgment after booting a newly upgraded firmware to have it become permanent (so loading a firmware which breaks WiFi won't necessarily brick the unit)
- uses only a single firmware image, built using the normal linker script
- can repack a NodeMCU firmware into an OTA image (though said firmware still needs to be built with the appropriate module enabled in order to boot properly)
- allows us to pack a NodeMCU firmware _and_ (read-only) filesystem into _each_ image, or have a shared filesystem, depending on the use-case

Specifically, our approach is not compatible/interchangeable with the SDK's FOTA, since:
- we have the concept of "in-test" firmwares
- we use a different header format (to prevent any accidental confusion)
- despite poking around in the disassembly of the official FOTA stuff, we could find no reliable way to inter-operate with its idea of which meg of flash should be mapped (we were hoping we could set things up so the SDK got the right idea about flash mapping, but no such luck)

However, the major drawbacks are:
- it will only ever support ESP modules with >= 2meg flash chip, since it relies on the ESP's hardware mapping of the flash to switch between firmwares, and that is only possible at megabyte boundaries
[...]

PR #2078, opened 2017-08-16
ESP32: OTA module

larsstenberg:

Early pull request to discuss OTA support.

_API_
*Start*
session = ota.start();

*Then depending if you want to write from LUA or let C code download the binary*
session:write(string.char(0x00, 0x0B, 0x0C, 0x0D));
session:download("www.cacharge.com", 80, "/kalle.png");

*Complete*
session:complete();

### DIFF
--- a/components/modules/otahttp.c
+++ b/components/modules/otahttp.c
@@ -1,0 +1,200 @@
+// Module for downloading over-the-air (OTA) firmware upgrades via http
+
+#include <string.h>
+#include <sys/socket.h>
+#include <netdb.h>
+#include <stdarg.h>
+
+#include "module.h"
+#include "lauxlib.h"
+#include "platform.h"
+
+#include "esp_system.h"
+#include "esp_log.h"
+#include "esp_ota_ops.h"
+#include "lwip/dns.h"
+
+#include "ota_common.h"
+
+#define BUFFSIZE 1024
+#define TEXT_BUFFSIZE 1024
+
+static const char *TAG = "otahttp";
+
+#define OTAHTTP_ERROR(L, format,...) \
+  ESP_LOGE(TAG, format, ##__VA_ARGS__); \
+  lua_pushboolean (L, false); \
+  return 1;
+
+typedef struct {
+  bool completed;
+  ip_addr_t ip_address;
+} otahttp_ip_lookup;
+
+static void dns_lookup_callback(const char *name, const ip_addr_t *ipaddr, void *callback_arg)
+{
+    otahttp_ip_lookup* lookup = callback_arg;
+    lookup->ip_address = *ipaddr;
+    lookup->completed = true;
+}
+
+static ip_addr_t dns_lookup(const char *domain) {
+  otahttp_ip_lookup lookup;
+
+  lookup.completed = false;
+  IP_ADDR4( &(lookup.ip_address), 0,0,0,0 );
+
+  err_t  err = dns_gethostbyname(domain, &(lookup.ip_address), dns_lookup_callback, &lookup);
+  if(err == ERR_OK)
+    dns_lookup_callback(domain, &(lookup.ip_address), &lookup);
+  else if (err == ERR_INPROGRESS)
+    while(!lookup.completed);
+
+  return lookup.ip_address;
+}
+
+static bool connect_to_http_server(const char* domain, uint16_t port)
+{
+    ESP_LOGI(TAG, "Server: %s Server Port:%d", domain, port);
+
+    ip_addr_t ip_address = dns_lookup(domain);
+
+    if(ip_address.u_addr.ip4.addr == 0) {
+      ESP_LOGE(TAG, "Failed to resolve IP address");
+      return -2;
+    } else {
+      ESP_LOGI(TAG, "DNS found: %i.%i.%i.%i\n", ip4_addr1(&ip_address.u_addr.ip4), ip4_addr2(&ip_address.u_addr.ip4), ip4_addr3(&ip_address.u_addr.ip4), ip4_addr4(&ip_address.u_addr.ip4) );
+    }
+
+    int http_connect_flag = -1;
+    int socket_id = socket(AF_INET, SOCK_STREAM, 0);
+    if (socket_id == -1) {
+        ESP_LOGE(TAG, "Create socket failed!");
+        return -1;
+    }
+
+    struct sockaddr_in sock_info;
+    memset(&sock_info, 0, sizeof(struct sockaddr_in));
+    sock_info.sin_family = AF_INET;
+    sock_info.sin_addr.s_addr = ip_address.u_addr.ip4.addr;
+    sock_info.sin_port = htons(port);
+
+    // connect to http server
+    http_connect_flag = connect(socket_id, (struct sockaddr *)&sock_info, sizeof(sock_info));
+    if (http_connect_flag == -1) {
+        ESP_LOGE(TAG, "Connect to server failed! errno=%d", errno);
+        close(socket_id);
+        return -3;
+    } else {
+        ESP_LOGI(TAG, "Connected to server");
+        return socket_id;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


larsstenberg:
@devsaurus or @marcelstoer, anyone have any input so i can go on =)

marcelstoer:
I'm not a firmware developer. Hence, all I could have told you is that as per [contribution guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md) we generally prefer to lead such discussions in an issue _before_ a PR is submitted. Since that wouldn't have been constructive I didn't 😉

devsaurus:
On a general level I'm pretty unbiased how OTA should be handled. We never had a canonical solution for the esp8266 and defaulting to the IDF's OTA way seems ok IMO. @jmattsson as an OTA stakeholder what do you think 😃?

Some thoughts regarding the design:

Why are sessions objects? Can there be more than one single ota session active at one time?

I understand that you embedded a http download method in the module in parallel to the raw binary write. That's pragmatic though not the cleanest approach. Http protocol handling *should* go into a dedicated module like on the esp8266 branch, and `ota` *should* consume payload on top of that. But we don't have `http` yet...
If others agree, I would be ok to have it this way until a dedicated `http` module materializes, and deprecate the embedded http code then.

`LUA_MODULE_OTA` seems to have a dependency to a specific partition table configuration. Can this be modeled with [kconfig syntax](https://github.com/espressif/esp-idf/blob/master/tools/kconfig/kconfig-language.txt)? E.g. enforce the required configuration in case ota is enabled.

In any case thank you very much for this contribution - OTA for nodemcu has a long history of pending user requests 😉

jmattsson:
Well I must admit I'm not up-to-date on the IDF's idea of OTA. Most likely it doesn't cater to the way we'd be using it at $work (see #806 if you're keen on grok'ing that), so I don't think I have more skin in the game than anyone else at this point. Besides, we're not using ESP32s at $work yet anyway.

I pretty much second what @devsaurus said. I think it would make more sense to have the OTA module be a singleton rather than return a session object - I can't see how there could be multiple sessions in progress. Having a way to manually cancel an in-progress upgrade would be nice, at least when using the `write()` approach rather than `download()`.

Breaking out the download support into either a full `http` module (asking a lot, I know), or just refactor the download function into an e.g. `otahttp` module would be nice.

Oh, and I see a potential buffer overflow in the use of `read_until()`, where `total_len - i` needs to be passed rather than `total_len`.

Overall a big thumbs up though!

larsstenberg:
New api:
ota.start();

ota.write(string.char(0xE9, 0x0B, 0x0C, 0x0D));
OR
otahttp.download("172.16.1.103", 9501, "NodeMCU.bin");

ota.complete();

larsstenberg:
I would like to have the possibility to set and retrieve partitions from LUA, so the user can decide boot partition.

What module would this be put in? And is it a good idea?

jmattsson:
I'd say that picking a boot partition is intrinsically linked to OTA upgrades, so keeping it in the `ota` module would make sense to me.

Last I checked the partition management exposed by the IDF was quite limited (and with encryption, somewhat tricky), so I don't think we'd want to try to go full `parted`/`fdisk` just yet.

larsstenberg:
Sure thing @jmattsson, i add it to the OTA module. Thanks for feedback

larsstenberg:
-- Get boot partition
for z, y in pairs(ota.partition.get_boot()) do
    print(z, y)
end

-- Get running partition
for z, y in pairs(ota.partition.get_running()) do
    print(z, y)
end

-- Set factory partition
factory_partition = ota.partition.find_first(ota.partition.TYPE_APP, ota.partition.SUBTYPE_ANY, "factory");
success = ota.partition.set_boot(factory_partition.handle);
print(success)

-- Find partitions by type, subtype and name
for k, v in pairs(ota.partition.find(ota.partition.TYPE_APP, ota.partition.SUBTYPE_ANY, "factory")) do
    print(v.label);
end

-- Find partitions by type and subtype
for k, v in pairs(ota.partition.find(ota.partition.TYPE_APP, ota.partition.SUBTYPE_ANY)) do
    print(v.label);
end

larsstenberg:
Regarding the dependency to have an OTA partition table, i think its kind of hard to configure with KConfig since a user can have a custom partition scheme.

larsstenberg:
If you guys thinks the API looks nice please notify so i can start write documentation.

jmattsson:
If we're limiting the partition stuff to OTA, is there real value in exposing the type/sub-type information? Or would it make more sense to hide that from the user and only use the partition labels?

larsstenberg:
-- Set factory partition
print(ota.partition.set_boot("factory"))

-- Get boot partition
print(ota.partition.get_boot().label)

-- Get running partition
print(ota.partition.get_running().label)

-- Find info for certain partition
for k, v in pairs(ota.partition.find("factory")) do
    print(k, v);
end

-- Find all app partitions
for k, v in pairs(ota.partition.find_all()) do
    print(v.label);
end

components/modules/otahttp.c
```diff
@@ -0,0 +1,205 @@
...
+  otahttp_ip_lookup lookup;
+
+  lookup.completed = false;
+  IP_ADDR4( &(lookup.ip_address), 0,0,0,0 );
+
+  err_t  err = dns_gethostbyname(domain, &(lookup.ip_address), dns_lookup_callback, &lookup);
+  if(err == ERR_OK)
+    dns_lookup_callback(domain, &(lookup.ip_address), &lookup);
+  else if (err == ERR_INPROGRESS)
+    while(!lookup.completed);
```
jmattsson: This hard loop here isn't good, nor the lack of synchronisation on the `completed` flag between threads. There are multiple constructs available to suspend a task which would be far better. The canonical way of doing it is properly async. Have a look at the `net` module for inspiration.

components/modules/otahttp.c
```diff
@@ -0,0 +1,205 @@
...
+    OTAHTTP_ERROR(L, "Send GET request to server failed");
+  } else {
+    ESP_LOGI(TAG, "Send GET request to server succeeded");
+  }
+
+  bool found_body = false, flag = true;
+  int packet_counter = 0;
+
+  /* Deal with all receive packets */
+  while (flag) {
```
jmattsson: Again, this would be better off as an asynchronous streaming of the HTTP data. The logic is sound, it just shouldn't block all other Lua activity while it is running.

components/modules/otahttp.c
```diff
@@ -0,0 +1,205 @@
...
+      ESP_LOGI(TAG, "Have written image length %d", image_length);
+    }
+  }
+
+  if(!found_body) {
+    OTAHTTP_ERROR(L, "Error: did not found body!");
+  } else {
+    ESP_LOGI(TAG, "Total Write binary data length : %d", image_length);
+
+    lua_pushboolean (L, true);
```
jmattsson: ...and rather than returning the success flag here, it'd be done by invoking a Lua callback. A callback of the canonical format `callback(err, other_data)` would be ideal, where `err` is null if there was no error.

components/modules/otahttp.c
larsstenberg: Hmm, not sure how this would be done. A FreeRTOS task and callback to LUA once a result is reached?

jmattsson: That would be one option (though you'd need to post an event to the Lua task so the actual Lua callback fires from that context - see `task/include/task/task.h`, it's easy). Another option would be to take a look at how the `net` module handles asynchronous network i/o and go down the same route. Both have merit, I'd be happy to approve either approach at this point.

abuisman:
Having OTA support would be great. What is needed to move this into a release?

marcelstoer:
Johny & Arnim, ok to close this as the original author didn't address the required changes?

jmattsson:
Sadly, I have to concur.

Issue #2080, opened 2017-08-17
Cron not working for latest release


abhignair:
**Please note that the cron module works using build from branch 2.1.0-master_20170521 but doesn't work using the latest release (master branch) from the cloud build service.**

8<------------------------ BUG REPORT -----------------------------------------
### Expected behavior
Output using nodemcu firmware from branch 2.1.0-master_20170521.

NodeMCU 2.1.0 build unspecified powered by Lua 5.1.4 on SDK 2.1.0(116b762)
> Time Sync  2017/08/17 08:32:21
Every Minute: 2017/08/17 08:32:21
Every Minute: 2017/08/17 08:33:00
Every Minute: 2017/08/17 08:34:00

...and so on with output of same format as above with different time every minute on new line.

### Actual behavior
Output using nodemcu firmware from the cloud build service.

NodeMCU custom build by frightanic.com
	branch: master
	commit: b62fae918b3627209fe434950cffb2506cee9a31
	SSL: false
	modules: cron,file,gpio,net,node,rtctime,sntp,tmr,uart,wifi
 build 	built on: 2017-08-17 06:35
 powered by Lua 5.1.4 on SDK 2.1.0(116b762)
> Time Sync  2017/08/17 07:16:31

...and no further output.
### Test code
Provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) which will reproduce the problem.
```Lua
-- add code here
function sync_success()
    tm = rtctime.epoch2cal(rtctime.get())
    print(string.format("Time Sync  %04d/%02d/%02d %02d:%02d:%02d", tm["year"], tm["mon"], tm["day"], tm["hour"], tm["min"], tm["sec"]))
    cron.schedule("* * * * *", function(e) 
		      tm = rtctime.epoch2cal(rtctime.get())
		      print(string.format("Every Minute: %04d/%02d/%02d %02d:%02d:%02d", tm["year"], tm["mon"], tm["day"], tm["hour"], tm["min"], tm["sec"]))
    end)
end

wifi.eventmon.register(wifi.eventmon.STA_GOT_IP, sntp.sync(nil, sync_success, nil, nil))
    
wifi.setmode(wifi.STATION)
    
station_cfg={}
station_cfg.ssid="xxxx"
station_cfg.pwd = "xxxx"
wifi.sta.config(station_cfg)
wifi.sta.connect()
```
### NodeMCU version
[...]

PR #1697, opened 2017-01-02
Add tracking and control of the rate error in the clock crystal.

pjsg:
Fixes #1696 

- [x] This PR is for the `dev` branch rather than for `master`.
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [x] I have thoroughly tested my contribution.
- [x] The code changes are reflected in the documentation at `docs/en/*`.

This includes control of the rtc clock rate (with resolution of 1 part in 4 billion). This automatically driven by the sntp module.

Also this PR only stores data in the battery backed ram when necessary. This speeds things up as accessing the bbram is slow. 

Yes, there are a lot of commits, but many of these have already been merged...

PR #2084, opened 2017-08-23
cron: don't include rtctime_internal.h

nwf:
Fixes nodemcu/nodemcu-firmware#2080

That file is supposed to only be included once because it does things
like declare static globals.  As it stands, cron doesn't believe time is
ticking.  With this patch applied, cron gets itself going again.

Requesting review from @pjsg and @djphoenix .

### DIFF
--- a/app/modules/cron.c
+++ b/app/modules/cron.c
@@ -8,7 +8,6 @@
 #include "c_string.h"
 #include "ets_sys.h"
 #include "time.h"
-#include "rtc/rtctime_internal.h"
 #include "rtc/rtctime.h"
 #include "stdlib.h"
 #include "mem.h"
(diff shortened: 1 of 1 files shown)


djphoenix:
@marcelstoer @devsaurus I think this is a moment when we *should* get aside from our release plan and cherry-pick this PR into `master`, because `cron` module doesn't work at all (if I understand situation right).

nwf:
@djphoenix Yeah, it looks like that file's inclusion policy changed with d93465cd and cron's #include was overlooked.

marcelstoer:
Maybe not really "overlooked"...the cron module was committed just a few days before #1697 was initially submitted. So, it could have been bad timing. That PR has than been sitting idle for half a year.

nwf:
@marcelstoer Ah, the joys of non-textual conflicts.  I apologize if my "overlooked" comment came across as offensive; it was not meant to besmirch any developer's reputation, merely as recognition that these things happen. :)

marcelstoer:
> I apologize if my "overlooked" comment came across as offensive

No worries, I didn't think it was. I think that I speak for all of us when I say "thank you for your contributions".

marcelstoer:
Let's get the 🎱 rolling. I'll first merge this will then create another full release.

PR #2086, opened 2017-08-24
Unbreak Travis-CI build for dev-esp32 branch

jmattsson:
After being handed a `trusty` container rather than the old `precise` container, things didn't work so well any more.

Changes:
- now forcing use of the OS python version, rather than new "magic" versions in /opt
- installing GNU make 4.1, since our build depends on that these days

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ install:
   - export PATH=$PWD/tools/toolchains/esp8266/bin:$PWD/tools/toolchains/esp32/bin:$PATH
 script:
 - dpkg -l | grep serial
+- which python python2.7 python3
+- python --version
+- dkpg -L python-serial
 - python -c "import serial; print(serial.__file__)"
 - python3 -c "import serial; print(serial.__file__)"
 - export BUILD_DATE=$(date +%Y%m%d)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 sudo: required
 dist: trusty
 language: cpp
+# we need at least GNU Make 4.0, which isn't available in trusty, so grab it
+# manually from xenial
 before_install:
 - wget http://mirrors.kernel.org/ubuntu/pool/main/m/make-dfsg/make_4.1-6_amd64.deb
 - sudo dpkg -i make_4.1-6_amd64.deb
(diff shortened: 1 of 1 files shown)


jmattsson:
Crumbs for @marcelstoer

.travis.yml
```diff
@@ -1,16 +1,21 @@
-sudo: false
+sudo: required # due to manual install of make 4.1
```
marcelstoer: I was gonna dismiss this but I now see that Travis finally supports sudo even for containerized environments. Previously requiring sudo would force your build to be run outside a container which routinely delayed it for up to 30min (until a slot became available). All good now.

marcelstoer:
Thanks Johny.

.travis.yml
jmattsson: You're not running in their docker container when you have sudo:required though. I couldn't find a way to get make 4.1 installed with sudo:false, otherwise I would've stayed with that. I'd hoped to sneak a modern make in via sources:debian-sid, but that *really* didn't work, and there is no xenial support anywhere to be found.

If you find a way, feel free to shift things around here. You can see the stuff I've already tried in the original branch.

marcelstoer: That's what I thought, yeah. Then I checked the log at https://travis-ci.org/nodemcu/nodemcu-firmware/builds/267828141 and expanded "Build system information". Since it's got Docker version  info there I was assuming that it'd be running on Docker now despite sudo.

jmattsson: Hmm. I don't know what to think anymore! :)

karrots: Looks like a full VM still if sudo is set to required.

https://docs.travis-ci.com/user/reference/overview/

They could still be abstracting things with Docker on the VM to make it easer to deploy.

marcelstoer: Thanks Jonathan, wanted to dig into the docs tonight but you beat me to it. They claim "20-50s boot time" for a sudo VM which I find acceptable.

jmattsson: Maybe at some point it would be worthwhile to suss out the performance of running our own docker instance there since they also support that. That way we wouldn't be as beholden to their rather old build environments. Not that I'm volunteering, I don't know much about docker, and no real bandwidth to change that... :)

PR #2090, opened 2017-08-27
CoAP module: Improve part of the CoAP client feature, and redesign the API

tiannian:
In the original CoAP implementation, Client features is not perfect. Server's Endpoint is restricted to a fixed path.

The redesigned API made the following changes:
- Combine Client and Server features on CoAP Peer objects
- Cancel the original Server Endpoint, you can directly register the lua function and handle the original data packets
- Cancel the direct dependency on ESP SDK's UDP Sockets

The CoAP protocol stack needs to implement the following features:
- [x] Basic Request / Response model support.
- [x] separate response (Support by lua code).
- [x] CoAP ping (Partly Support).
- [ ] Observe
- [ ] Resource discovery
- [ ] Cache
- [ ] Proxy
- [ ] Multicast CoAP


macduff:
Is this going to be integrated into the repo at some point?  I'm wondering, because right now I'm using the original coap implementation with some minor modifications for a project.  I'd like to know if that work will be deprecated in favor of this new API.

app/coap/coap.c
```diff
@@ -442,28 +442,29 @@ unsigned int coap_encode_var_bytes(unsigned char *buf, unsigned int val) {
...
 {
     int res;
     pkt->hdr.ver = 0x01;
     pkt->hdr.t = t;
     pkt->hdr.tkl = 0;
     pkt->hdr.code = m;
+    message_id ++;
     pkt->hdr.id[0] = (message_id >> 8) & 0xFF;  //msgid_hi;
     pkt->hdr.id[1] = message_id & 0xFF; //msgid_lo;
     message_id++;
```
pjsg: Is this second increment of message_id required?

app/coap/node.c
```diff
@@ -99,6 +99,23 @@ coap_queue_t * coap_pop_next( coap_queue_t **queue ) {		// this function is call
...
+coap_queue_t * coap_find_node(coap_queue_t *queue, const coap_tid_t id){
+  coap_queue_t *p, *q;
+  if ( !queue ) 
+    return NULL;
+  q = queue;
+  if (q->id == id) {
+    return q;  
+  }
+  while (q && q->id != id) {
+    p = q;
```
pjsg: Is this line still needed?

docs/en/modules/coap.md
```diff
@@ -6,225 +6,148 @@
...
-Issues a PUT request to the server.
-
-#### Syntax
-`coap.client:put(type, uri[, payload])`
-
-#### Parameters
-- `type` `coap.CON`, `coap.NON`, defaults to CON. If the type is CON and request fails, the library retries four more times before giving up.
-- `uri` the URI such as "coap://192.168.18.103:5683/v1/v/myvar", only IP addresses are supported i.e. no hostname resoltion.
-- `payload` optional, the payload will be put in the payload section of the request.
+You must return the first argument after processing the request.
```
pjsg: I don't understand this statement.... Are you saying that the callback function must return the first argument? Why?

app/modules/coap.c
```diff
@@ -18,586 +18,416 @@
...
+static int lua_coap_peer_recv(lua_State* L){
+    // 收到请求
+    coap_peer_t *peer = (coap_peer_t *) luaL_checkudata(L,1,"coap_peer");
+    coap_pdu_t *pdu = coap_new_pdu();
+    int len,rc;
+    const char *data = luaL_checklstring( L, 2 ,&len);
+    const char *ip = luaL_checkstring( L, 3 );
+    pdu->port = luaL_checkinteger( L, 4 );
+    
+    c_memcpy(pdu->ip,ip,16);
```
pjsg: No -- don't copy like this. Validate the string length and then use strcpy

pjsg:
This is (presumably) incompatible with the current coap implementation. Is there an easy way to migrate from the old version to the new version?

marcelstoer:
No feedback from @tiannian on Philip's review -> closing

PR #2097, opened 2017-09-08
Fix data loss in TCP streams.

jmattsson:
- [n/a] The code changes are reflected in the documentation at `docs/en/*`.

So it turns out that very occasionally^ the lwip stack hands over chained `pbuf`s to the `tcp_recv()` callback. As only the first `pbuf` in the chain was passed into Lua, this meant that data went missing from the middle of the TCP stream (which one would certainly *not* expect). The fix here is to simply do multiple callbacks into Lua until the end of the chain is reached (though I've never seen more than two `pbuf`s in the chain so far).

With this patch I'm now always successfully getting the full TCP stream, whereas before I never did.

^) During each 1meg transfer, I saw it happen 1-4 times.

### DIFF
--- a/app/modules/net.c
+++ b/app/modules/net.c
@@ -203,19 +203,22 @@ static void net_recv_cb(lnet_userdata *ud, struct pbuf *p, ip_addr_t *addr, u16_
     return;
   }
 
+  int num_args = 2;
+  char iptmp[16] = { 0, };
+  if (ud->type == TYPE_UDP_SOCKET)
+  {
+    num_args += 2;
+    ets_sprintf(iptmp, IPSTR, IP2STR(&addr->addr));
+  }
+
   lua_State *L = lua_getstate();
   struct pbuf *pp = p;
   while (pp)
   {
-    int num_args = 2;
     lua_rawgeti(L, LUA_REGISTRYINDEX, ud->client.cb_receive_ref);
     lua_rawgeti(L, LUA_REGISTRYINDEX, ud->self_ref);
     lua_pushlstring(L, pp->payload, pp->len);
     if (ud->type == TYPE_UDP_SOCKET) {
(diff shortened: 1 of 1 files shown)


app/modules/net.c
```diff
@@ -202,20 +202,26 @@ static void net_recv_cb(lnet_userdata *ud, struct pbuf *p, ip_addr_t *addr, u16_
...
+  {
+    int num_args = 2;
+    lua_rawgeti(L, LUA_REGISTRYINDEX, ud->client.cb_receive_ref);
+    lua_rawgeti(L, LUA_REGISTRYINDEX, ud->self_ref);
+    lua_pushlstring(L, pp->payload, pp->len);
+    if (ud->type == TYPE_UDP_SOCKET) {
+      num_args += 2;
+      char iptmp[16];
+      bzero(iptmp, 16);
+      ets_sprintf(iptmp, IPSTR, IP2STR(&addr->addr));
```
djphoenix: Can be moved outside a loop

jmattsson: Yes, though it becomes (slightly) harder to follow, and you double up on the `if` statement as you'll need to push the port/ip each time through the loop regardless. Given that we'd normally be doing the `ets_sprintf` for each pbuf anyway, I felt the tradeoff wasn't worth the readability cost. If you really reckon it is, I can change it.

djphoenix: Agree for "harder to follow"...
Will accept 'FIXME' comment there as a solution for now.

joysfera:
wondering if this patch could fix also the mysterious MQTT issues (where bits of TCP data were missing randomly)

jmattsson:
Afaiu mqtt uses espconn, so this shouldn't be applicable there. Someone is welcome to prove me wrong though :)

PR #2120, opened 2017-09-25
Added softuart support

crasu:
I hacked together very basic softuart support over the weekend by porting parts from the code in this repo https://github.com/plieningerweb/esp8266-software-uart. It is not yet in a mergable state. If you think it might be useful to others I can prepare a proper pull request.

Things that need to be addressed
* Use functions from platform.c
* Code formatting (is there a style guide for nodemcu somewhere)
* User documentation
* Clean up unused code parts


app/include/user_modules.h
```diff
@@ -78,6 +78,7 @@
 //#define LUA_USE_MODULES_WS2801
 //#define LUA_USE_MODULES_WS2812
 //#define LUA_USE_MODULES_XPT2046
+#define LUA_USE_MODULES_SOFTUART
```
eku: Why is this module active by default?

app/include/user_config.h
```diff
@@ -21,7 +21,7 @@ extern void luaL_assertfail(const char *file, int line, const char *message);
 
 // This enables lots of debug output and changes the serial bit rate. This
 // is normally only used by hardcore developers
-// #define DEVELOP_VERSION
+#define DEVELOP_VERSION
```
eku: ?!?

app/modules/softuart.c
```diff
@@ -0,0 +1,477 @@
...
+// Function for printing individual characters
+void softuart_putchar(char data)
+{
+    NODE_DBG("putchar called\r\n");
+    Softuart* s = &softuart;
+	unsigned i;
+	unsigned start_time = 0x7FFFFFFF & system_get_time();
+
+	//Start Bit
+	GPIO_OUTPUT_SET(GPIO_ID_PIN(s->pin_tx.gpio_id), 0);
```
djphoenix: Move `GPIO_ID_PIN(s->pin_tx.gpio_id)` to variable (used many times)

app/modules/softuart.c
```diff
@@ -0,0 +1,477 @@
...
+	softuart_pin_t pin_tx;
+	softuart_buffer_t buffer; //TODO volatile ??? not needed as no read anymore
+    uint32_t baudrate;
+	uint16_t bit_time;
+    uint16_t need_len;
+    uint16_t end_char;
+    int uart_receive_rf;
+} Softuart;
+
+Softuart softuart = {false};
```
djphoenix: Why using static variable instead of object-model?

app/modules/softuart.c
```diff
@@ -0,0 +1,477 @@
...
+	unsigned start_time = 0x7FFFFFFF & system_get_time();
+
+	//Start Bit
+	GPIO_OUTPUT_SET(GPIO_ID_PIN(s->pin_tx.gpio_id), 0);
+	for(i = 0; i < 8; i ++ )
+	{
+		while ((0x7FFFFFFF & system_get_time()) < (start_time + (s->bit_time*(i+1))))
+		{
+			//If system timer overflow, escape from while loop
+			if ((0x7FFFFFFF & system_get_time()) < start_time){break;}
```
djphoenix: Adjust remaining count instead of breaking?

app/modules/softuart.c
```diff
@@ -0,0 +1,477 @@
...
+		}
+		GPIO_OUTPUT_SET(GPIO_ID_PIN(s->pin_tx.gpio_id), chbit(data,1<<i));
+	}
+
+	// Stop bit
+	while ((0x7FFFFFFF & system_get_time()) < (start_time + (s->bit_time*9)))
+	{
+		//If system timer overflow, escape from while loop
+		if ((0x7FFFFFFF & system_get_time()) < start_time){break;}
```
djphoenix: Adjust? (again)

Also, maybe use static func for this delay? (used twice)

app/modules/softuart.c
```diff
@@ -0,0 +1,477 @@
...
+	}
+	GPIO_OUTPUT_SET(GPIO_ID_PIN(s->pin_tx.gpio_id), 1);
+
+	// Delay after byte, for new sync
+	os_delay_us(s->bit_time*6);
```
djphoenix: Why not to use previous loop-based delay here?

app/modules/softuart.c
crasu: Thank you for your response. I would like to be as close as possible to the hardware uart module. What do you think of using ids instead of objects?

djphoenix: I think IDs may be confusing here. Software UART may be used on any pair of pins, so...

pjsg:
Need to add softuart.md to the documentation tree....

app/modules/softuart.c
```diff
@@ -0,0 +1,398 @@
...
+    uint32_t result = 1;
+    uint32_t baudrate = (uint32_t) luaL_checkinteger(L, 1);
+    uint8_t tx_gpio_id = (uint8_t) luaL_checkinteger(L, 2);
+    uint8_t rx_gpio_id = (uint8_t) luaL_checkinteger(L, 3);
+
+    if(softuart.enabled)
+        return luaL_error( L, "SOFTUART already configured." );
+
+    if(!softuart_gpio_valid(tx_gpio_id)) {
+        return luaL_error(L, "SOFTUART tx GPIO not valid");
```
smil2k: Hi,

It would be beneficial to allow RX or TX only usage by accepting NIL as RX/TX pin. Of course in that case you cannot use read/write methods.

app/modules/softuart.c
```diff
@@ -0,0 +1,398 @@
...
+            os_delay_us(s->bit_time);
+
+            //done
+        }
+
+        //clear interrupt
+        GPIO_REG_WRITE(GPIO_STATUS_W1TC_ADDRESS, gpio_status);
+
+        // Reactivate interrupts for GPIO0
+        gpio_pin_intr_state_set(GPIO_ID_PIN(s->pin_rx), GPIO_PIN_INTR_ANYEDGE);
```
smil2k: Why triggering on ANY edge. If inverted mode not supported, then triggering is on FALLING edge only.

marcelstoer:
Are we going to see further development here (addressing questions and requested changes) or can we close this?

crasu:
Just close it. There does not seem to be much interested in it.

HHHartmann:
Please don't.
I was just thinking of using it to circumvent the limitation of node.output() not catching the error output.
I would just connect the main UART to this module.
This would help being able to use the ESP by network without having to connect the serial.
Or am I missing something?

marcelstoer:
> Please don't.

Well, for that to happen the author (or someone else) would have to be willing to address the review feedback and fix the merge conflicts.

PR #2129, opened 2017-10-05
Restore module MCP23017 from old-master

bigjohnson:

I re add the MCP23017 lua module from old-master that disappear in new, and addad two examples.

### DIFF
--- a/lua_examples/mcp23017/mcp23017_leds_A.lua
+++ b/lua_examples/mcp23017/mcp23017_leds_A.lua
@@ -1,0 +1,44 @@
+---
+-- @description Shows control of 8 GPIO pins/LEDs via I2C with the MCP23017 I/O expander.
+-- Tested on odeMCU 0.9.5 build 20150213. 
+-- @date March 02, 2015
+-- @circuit Connect 8 LEDs withs resistors to the GPIO pins of the MCP23017.
+--  Connect GPIO0 of the ESP8266-01 module to the SCL pin of the MCP23017. 
+--  Connect GPIO2 of the ESP8266-01 module to the SDA pin of the MCP23017.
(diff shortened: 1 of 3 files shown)


lua_examples/mcp23017/mcp23017_leds_A.lua
```diff
@@ -0,0 +1,44 @@
+---
+-- @description Shows control of 8 GPIO pins/LEDs via I2C with the MCP23017 I/O expander.
+-- Tested on odeMCU 0.9.5 build 20150213.
```
marcelstoer: Hhhmm, what about recent versions?

bigjohnson: In the master code there are
https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_examples/mcp23008/mcp23008_leds.lua
https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_examples/mcp23008/mcp23008_buttons.lua
https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_modules/mcp23008/mcp23008.lua
files with the same phrase and no one remove them.
The 23008 and 23017 are almost the same chip, the only difference is that 23017 has more io and their registers.
If none remove the 23008 module and examples also the 23017 module and examples should be added.
If you need to test the module I can test it with my board https://www.youtube.com/watch?v=uh4x8T8cQ-E

marcelstoer: > with the same phrase and no one remove them

It's never too late to improve. If we add new files they need to be accurate. You might even want to update the description and add your name as someone who tested and verified.

marcelstoer:
No more feedback from author in two months -> closing.

PR #2132, opened 2017-10-12
EndUser_setup will now be able to accept multiple other inputs via th…

Jcrash29:
Hey guys,

Feel free to tear the code apart, this is my first time making a PR to project with which I didn't know the dev team.

This code is working, but still has some debug code in there for items I was confused about.

What it does:

Allows for inputs to be added to the .html file for enduser_setup and those extra inputs can be added to a JSON file for later extraction and use.

links to similar PR's:

https://github.com/nodemcu/nodemcu-firmware/pull/1886


app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+ *
+ *
+ *
+ */
+static int enduser_setup_write_file_with_extra_configuration_data(
+char * configuration_string, int configuration_length
+)
+{
+  int p_file = 0;
+  lua_State *L = NULL;
```
jmattsson: Stylistically I'd prefer to see these declared right before use rather than old-C style top-of-function.

app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+	
+  ENDUSER_SETUP_DEBUG("enduser: opening enduser_custom_parameters.json for write");
+
+  // setup the file output
+  p_file = vfs_open("enduser_custom_parameters.json", "w");
```
jmattsson: May I suggest a shorted filename? Perhaps "eus_custom.json"? Feels like a lot of characters to type otherwise.
 
Actually, is it guaranteed to even be JSON? If not, maybe "eus_custom.txt"?

app/modules/enduser_setup.c
```diff
@@ -675,6 +721,27 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
...
+    int config_str_len = enduser_setup_srch_str(config_str_start, "& ");
+    /* Create a new variable for storing the decoded string.
+     * Add one more size(char) to allow for \0 */
+    uint8 decoded_config_str[config_str_len+1];
+    ENDUSER_SETUP_DEBUG("Extra Config Data found");
+	
+
+    err |= enduser_setup_http_urldecode(decoded_config_str, config_str_start, config_str_len, sizeof(decoded_config_str));
+	
+    enduser_setup_write_file_with_extra_configuration_data(decoded_config_str, config_str_len);
```
jmattsson: The return value of `enduser_setup_write_file_with_extra_configuration_data()` is being ignored here. It probably shouldn't be. Also, that's one long function name. Quite descriptive though, can't argue with that :)

app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+
+  // setup the file output
+  p_file = vfs_open("enduser_custom_parameters.json", "w");
+  if (p_file == 0)
+  {
+    ENDUSER_SETUP_DEBUG("Can't write to file!");
+    return 1;
+  }
+  
+  /* Not certain what this does */
```
jmattsson: This gets you the correct Lua context to work on. Note that this `lua_State` stack might already be in use, so care must be taken to ensure that the stack is left in the same condition as when the function was entered. Weird Stuff(tm) happens otherwise, of the variety that's hard to track down.

app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+  L = lua_getstate();
+  
+  if (L != NULL) 
+  {
+    lua_pushlstring(L, configuration_string, configuration_length);
```
jmattsson: Without the `lua_call()` below, this will end up leaving the stack in an inconsistent state. I'd suggest moving this down to the same block, there's no reason not to from what I can see.

app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+  }
+  
+  vfs_write(p_file, configuration_string, configuration_length);
```
jmattsson: The return code from the write should be paid attention to and propagated if error'ed.

app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
...
+  vfs_close(p_file);
+/*  
+  if (L != NULL) 
+  {
+    ENDUSER_SETUP_DEBUG("enduser: calling lua for validation");
+
+    lua_call(L, 1, 1);
```
jmattsson: The idea of a validation callback is good, but you'll need to pay attention to the return value in such case, and also the file shouldn't be written if the validation fails. If the callback is just for notification the current location is good. You'll also need to actually push the callback function onto the stack before the argument in order to be able to `lua_call()` it. There are plenty of examples hanging around in the modules, have a look for `lua_call()`, `luaL_ref()`, `luaL_unref()` and `lua_rawgeti()`.

app/modules/enduser_setup.c
```diff
@@ -675,6 +721,27 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
...
+
+  char *config_str = (char *) ((uint32_t)strstr(&(data[6]), "config="));
+  if(config_str != NULL)
+  {
+    int config_field_len = LITLEN("config=");
+    char *config_str_start = config_str + config_field_len;
+    int config_str_len = enduser_setup_srch_str(config_str_start, "& ");
+    /* Create a new variable for storing the decoded string.
+     * Add one more size(char) to allow for \0 */
+    uint8 decoded_config_str[config_str_len+1];
```
jmattsson: Just a note of caution that the stack space is rather limited. It would be safer to `os_malloc()` this buffer (and later `os_free()` it. As-is, someone could smash the stack by submitting a really long config parameter.

app/modules/eus/enduser_setup.html
```diff
@@ -160,6 +160,8 @@ <h3>Connect device to your Wi-Fi</h3>
 				</div>
 				<input id=ssid type=text autocorrect=off autocapitalize=none placeholder='Wi-Fi Name' />
 				<input id=wifi_password type=text autocorrect=off autocapitalize=none autocomplete=off placeholder=Password />
+				<input id=TwitterAPI type=text autocorrect=off autocapitalize=none autocomplete=off placeholder='Twitter API Handler' />
+				<input id=TwitterMessage type=text autocorrect=off autocapitalize=none autocomplete=off placeholder='Twitter Message' />
 				<button id=submit type=button>Save</button>
```
jmattsson: These additions shouldn't be part of the default page.

app/modules/eus/enduser_setup.html
```diff
@@ -242,7 +244,11 @@ <h4 id='st'>Updating Status...</h4>
 		}
 		function submit() {
 			submitted = true;
-			var url = '/setwifi?wifi_ssid=' + encodeURIComponent($('#ssid').value) + '&wifi_password=' + encodeURIComponent($('#wifi_password').value);
+			var json, config = {};
+			config.twitterAPI = $('#TwitterAPI').value;
+			config.twitterMessage = $('#TwitterMessage').value;
+			json = JSON.stringify(config);
+			var url = '/setwifi?wifi_ssid=' + encodeURIComponent($('#ssid').value) + '&wifi_password=' + encodeURIComponent($('#wifi_password').value) + '&config=' + encodeURIComponent(json);
 			clearTimeout(rs);
```
jmattsson: Likewise, these shouldn't be part of the default page. It'd be a good example in the docs though.

app/modules/eus/http_html_backup.def
```diff
@@ -1,205 +1,213 @@
 static const char http_html_backup[] = {
```
jmattsson: Again, changes should not be part of the default. Same goes for `http_html_backup.def`.

app/modules/enduser_setup.c
```diff
@@ -626,6 +626,52 @@ static void do_station_cfg (task_param_t param, uint8_t prio)
 }
 
 
+/**
+ *
+ *
+ *
+ *
```
jmattsson: SPAAAAAAAAAAAAAAAAAAAAACE! :)

Jcrash29:
@jmattsson  THANKS! I'll start making these changes. Really appreciate you taking the time to look at this!

marcelstoer:
Johny, I'm really happy you took the time to review this PR. I was afraid no one else from us maintainers besides me thought this be a very useful addition to the current module.

app/modules/eus/enduser_setup.html
marcelstoer: I agree with Johny here. I'm a bit surprised to see no changes to the documentation. I wouldn't want to change the default page but see documented how to change it in

- the module documentation
- or by  creating a nice example at https://github.com/nodemcu/nodemcu-firmware/tree/master/lua_examples and linking documentation to it

davidcbittner:
Hi, Has this been added? Is there any documentation available how to make use of this functionality? I had posted a request earlier about adding the host name as one of the parameters that should be set during the configuration. This is really a fundamental field which IMO should be added to the default HTML page but if it's not I'm certainly willing to save my own on there but not sure how to actually implement this.
Thanks
David

marcelstoer:
> Has this been added?

As you can see this PR is still open i.e. the code is not available in the NodeMCU repository yet. Hence, if you wanted to use it right now you'd have to manually build from the author's repository https://github.com/Jcrash29/nodemcu-firmware.

> Is there any documentation available how to make use of this functionality?

If you follow the conversation above you'll also see that we requested a couple of changes from the author before we'd merge the feature into our repo. One of the changes asks for (more/better) documentation.

marcelstoer:
@Jcrash29 I just realized you based your work on `master` and created the PR accordingly. This is not in line with [our contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#working-with-git-and-github). We can continue the conversation here until everything is ready but eventually this has to land on `dev`. I propose you keep committing changes to _your_ `master` (they'll show up here). When ready you can update your `dev`, carry over all changes and create a new PR.

davidcbittner:
Marcel, thank you for the reply. I am a bit new to github and learning the procedures, flags, etc…

David

and7ey:
Hi, is anyone working to get it implemented?

Jcrash29:
Sorry not currently.

I've got a couple other projects I am addressing.

I did implement several of the code changes recommended above, however have
to sufficiently tested them to create a commit.

and7ey:
I would suggest to modify the code in the following way:

1. change GET to POST when do `/setwifi` or `/update`. It will allow to pass more data.
2. modify javascript `submit()` function to send all text fields data instead of listing them manually in the code.

PR #2148, opened 2017-10-26
Fixes problems with time not preserved through dsleep

### LINKED ISSUE #2144: rtc time not kept through deep sleep
### Expected behavior

Setting the rtc time using `rtctime.set(0,0)` enables the realtime clock so that after a deep sleep, a call to `rtctime.get()` returns the curren (up-) time. This works on a version build from master on 2017-06-17. 
The example code outputs "phase 1", goes to deepsleep and then prints "phase 2".

### Actual behavior

Using a current master build (2017-10-12), reading the time using `rtctime.get()` always returns 0.
The example code outputs "phase 1" in an endless loop. I suppose that the changes of commit d93465c have something to do with the changed behaviour?

### Test code
```Lua
    local sec, usec = rtctime.get()
    if (sec == 0) then
        print("phase 1, initializing time + sleeping 2s");
        rtctime.set(0, 0)

        tmr.create():alarm(100, tmr.ALARM_SINGLE, function()
            rtctime.dsleep(2 * 1000 * 1000)
        end)
        
    else
        print(string.format("phase 2, uptime: %ds", sec));
    end
```
### NodeMCU version

Working version:
````
	branch: master
	commit: c8ac5cfb912ff206b03dd7c60ffbb2dafb83fe5e
	SSL: false
	modules: adc,bit,cron,file,gpio,mqtt,net,node,rtcmem,rtctime,sntp,tmr,uart,wifi,ws2812
 build 	built on: 2017-06-17 20:54
````

Broken version:
```
	branch: master
	commit: 2e67ff5a639a13260fd4fb3c3b627ccdc2845616
	SSL: false
	modules: adc,bit,cron,ds18b20,file,gpio,mqtt,net,node,rtcmem,rtctime,sntp,tmr,uart,wifi,ws2812
 build 	built on: 2017-10-12 19:32

```

### Hardware

ESP8266 ESP-12F on custom adapter board (http://www.mikrocontroller-elektronik.de/esp12e-tutorial-einstieg-mit-dem-esp8266-modul/)

### PR DESCRIPTION
**Pull request by pjsg**
Fixes #2144.

When I moved the rtc variables into memory (from the very slow rtc memory), the .bss initialization would clear them out after they were initialized. This caused no end of problems. Also, the sleep tracking values were updated *after* the last save to the rtc memory.

I'm currently running a long term test. The long term sleep (of 100 seconds) seems to cause a drift of under 1000ppm.

### DIFF
--- a/app/include/rtc/rtctime_internal.h
+++ b/app/include/rtc/rtctime_internal.h
@@ -506,11 +506,6 @@ extern void rtc_time_enter_deep_sleep_final(void);
 
 static void rtc_time_enter_deep_sleep_us(uint32_t us)
 {
-  if (rtc_time_check_wake_magic())
-    rtc_time_set_sleep_magic();
-  else 
-    bbram_save();
-
   rtc_reg_write(0,0);
   rtc_reg_write(0,rtc_reg_read(0)&0xffffbfff);
   rtc_reg_write(0,rtc_reg_read(0)|0x30);
@@ -535,6 +530,11 @@ static void rtc_time_enter_deep_sleep_us(uint32_t us)
 
   uint32_t cycles=rtc_time_us_to_ticks(us);
   rtc_time_add_sleep_tracking(us,cycles);
+
+  if (rtc_time_check_wake_magic())
+    rtc_time_set_sleep_magic();
+  else 
+    bbram_save();
 
   rtc_reg_write(RTC_TARGET_ADDR,rtc_time_read_raw()+cycles);
   rtc_reg_write(0x9c,17);
--- a/app/include/rtc/rtctime_internal.h
+++ b/app/include/rtc/rtctime_internal.h
@@ -187,8 +187,9 @@
 #define CPU_BOOTUP_MHZ    52
 
 #ifdef RTC_DEBUG_ENABLED
+#warn this does not really work....
 #define RTC_DBG(...)         do { if (rtc_dbg_enabled == 'R') { dbg_printf(__VA_ARGS__); } } while (0)
-static bool rtc_dbg_enabled;
+static char rtc_dbg_enabled;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/include/rtc/rtctime_internal.h [hunk 1]
```diff
@@ -187,8 +187,9 @@
 #define CPU_BOOTUP_MHZ    52
 
 #ifdef RTC_DEBUG_ENABLED
+#warn this does not really work....
```
pjsg: Debugging this code is a nightmare as some of it runs before the BSS segment is initialized and you can't print anything...

pjsg:
image

This shows around 24 hours of run with lots of 100 second deep sleeps. The blue line is the offset of the clock from actual after wakeup (left axis). The red line is calibration values (on the right axis). I think that the room was cooling down over the 24 hours (it has been getting cooler here). 

This looks good to me.

marcelstoer:
I don't understand the code but the graphs are convincing 😉

mkresse:
This PR seems to fix the issue and works for me.

marcelstoer:
Thanks for the confirmation.

mkresse:
Unfortunately I discovered only now an issue with the updated version. When powered up (after some seconds being turned off), `rtctime.get()` returns a random value instead of zero. This value remains constant even after a reset but changes to a different random value after a power down.

Only when rtctime is initialized via `rtctime.set(0)`, it behaves as the previous version: after resetting the ESP, `rtctime.get()` constantly returns 0 until `rtctime.set(0)` is called again.

Or to describe it another way: an init.lua with:

    print(rtctime.get())

now prints a random value after being powered up, while formerly it reliably printed 0.

marcelstoer:
@mkresse can you please verify #2154

PR #2151, opened 2017-10-28
BME680 module initial commit

### LINKED ISSUE #2114: Module for BME680
### Missing feature
Module for BME680 is missing

### Justification
The new [BME680 sensor ](https://www.bosch-sensortec.com/bst/products/all_products/bme680)from bosch is the follow-up chip of the already supportet BME280.
Beside temperature, humidity and air pressure, **AIR QUALITY** (VOCs) could be messured.
In my opinion the BME680 is a cheap solution to messure IAQ-Index (Indoor Air Quality) for IoT and home automatisation solutions like automatic aeration.

### Workarounds
The only workaround is to directly communicate via I2C to the Chip.

vsky279:
Fixes #2114.

This is a simple driver of the Bosch's BME680 sensor. In addition to the BME280 it is able to measure the indoor air quality. Unfortunately the IAQ (Indoor Air Quality) Index calculation algorithm is not publicly available. So we need to live with the gas resistance value which should still be some proxy to the IAQ.

Definitely it would be good if someone can further test the module before it is merged.


docs/en/modules/bme680.md
```diff
@@ -0,0 +1,159 @@
+# BME680 module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2017-10-28 | [vsky279](https://github.com/vsky279) | [vsky279](https://github.com/vsky279) | [bme680.c](../../../app/modules/bme680.c)|
+
+Copyright (C) 2017 - 2018 Bosch Sensortec GmbH
```
marcelstoer: What is this copyright for?

marcelstoer:
>  it would be good if someone can further test the module

I'm afraid that due to lack of hardware this will probably not happen. However, a code review can still be conducted.

docs/en/modules/bme680.md
vsky279: I'm using parts of the [bme680.c](https://github.com/BoschSensortec/BME680_driver/blob/2a51b9c0c1899f28e561e6701caa22cb23201cfc/bme680.c) driver from Bosch - it says we should retain the copyright. I think maybe just one line should be enough.

marcelstoer: I saw that, yes. However, I thought leaving the copyright in the source code would be enough. It feels somewhat out of place here and makes it look like Bosch contributed the NodeMCU module (and wrote this documentation).

vsky279: This seems reasonable. I will make a block in the code with these functions and will put the (c) note there.

pjsg:
I see that this uses doubles to perform calculations... Does this work in an integer build?

vsky279:
Yes, it does. This is even tested in the BME280 module.

KT819GM:
Temperature readings looks wrong (upper side by about 10C). Any advice how to add compensation for this? Humidity and resistance calculations seems OK. 
Before have tried this module with ESP32, comparing readings with BME280 and had less than degree difference. 
Using sample code from sample code with: Heater temp 360C, heater time 200ms, readout delay 250ms. Wrong temperature reading renders dew calculations wrong.

PR #2152, opened 2017-10-29
Make it build without INTERRUPT_ENABLE

### LINKED ISSUE #2092: Make fails if GPIO_INTERRUPT_ENABLE is not defined
### Expected behaviour
Node MCU should still compile

### Actual behaviour
`modules/gpio.c` fails to compile because of a debug addition by @pjsg at lines [134-135](https://github.com/nodemcu/nodemcu-firmware/blob/2e67ff5a639a13260fd4fb3c3b627ccdc2845616/app/modules/gpio.c#L135) as `pin_int_type` is only declared if `GPIO_INTERRUPT_ENABLE` is defined. 

### NodeMCU version
dev and master

I can't be bothered to raise a PR to fix this, since its been like this for two years and no one has noticed, but I'll incorporate the fix in my next dev PR.

pjsg:
Fixes #2092 .

- x ] I have thoroughly tested my contribution.

Makes it build without having to have GPIO_INTERRUPT_ENABLE defined (the default is to have it defined).

### DIFF
--- a/app/modules/gpio.c
+++ b/app/modules/gpio.c
@@ -132,7 +132,13 @@ static int lgpio_mode( lua_State* L )
 
 NODE_DBG("pin,mode,pullup= %d %d %d\n",pin,mode,pullup);
 NODE_DBG("Pin data at mode: %d %08x, %d %d %d, %08x\n",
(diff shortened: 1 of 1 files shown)


TerryE:
I picked this up as well and have fixed it in my current PRs, but since you've beat me to it .... :smile:

PR #2153, opened 2017-10-29
Make the CRON module work reliably

### LINKED ISSUE #1924: Cron Bug while Disabled schedules called again
Make sure you read and understand http://nodemcu.readthedocs.io/en/dev/en/support/.
Use one of the two templates below and delete the rest.

### Expected behavior
```Lua
first = cron.schedule("* * * * *", function(e)
  print("First Tick")
end)
second = cron.schedule("*/2 * * * *", function(e)
  print("second Tick")
end)
third = cron.schedule("*/3 * * * *", function(e)
  print("third Tick")
end)

first:unschedule()

first:schedule("* * * * *")
```
At each minute it should print First Tick
### Actual behavior
For Each Two and third Minutes I am getting second Tick and Third Tick but not the first tick
### Test code
```Lua
first = cron.schedule("* * * * *", function(e)
  print("First Tick")
end)
second = cron.schedule("*/2 * * * *", function(e)
  print("second Tick")
end)
third = cron.schedule("*/3 * * * *", function(e) 
  print("third Tick")
end)
first:unschedule()
first:schedule("* * * * *")
```
#
NodeMCU version
development branch -dev

pjsg:
Fixes #1924.

There were two bugs -- the memmove moved the wrong number of bytes (which could have led to a crash) and the lua_touserdata calls were trying to get the pointer from the wrong stack location. It is (somewhat) of a miracle that it worked at all!

### DIFF
--- a/app/modules/cron.c
+++ b/app/modules/cron.c
@@ -104,7 +104,7 @@ static size_t lcron_findindex(lua_State *L, cronent_ud_t *ud) {
   size_t i;
   for (i = 0; i < cronent_count; i++) {
     lua_rawgeti(L, LUA_REGISTRYINDEX, cronent_list[i]);
-    eud = lua_touserdata(L, 1);
+    eud = lua_touserdata(L, -1);
     lua_pop(L, 1);
     if (eud == ud) break;
   }
@@ -141,7 +141,7 @@ static int lcron_unschedule(lua_State *L) {
   size_t i = lcron_findindex(L, ud);
   if (i == -1) return 0;
   luaL_unref(L, LUA_REGISTRYINDEX, cronent_list[i]);
-  memmove(cronent_list + i, cronent_list + i + 1, sizeof(int) * cronent_count - i - 1);
+  memmove(cronent_list + i, cronent_list + i + 1, sizeof(int) * (cronent_count - i - 1));
   cronent_count--;
   return 0;
 }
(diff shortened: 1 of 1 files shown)


app/modules/cron.c
```diff
@@ -141,7 +141,7 @@ static int lcron_unschedule(lua_State *L) {
   size_t i = lcron_findindex(L, ud);
   if (i == -1) return 0;
   luaL_unref(L, LUA_REGISTRYINDEX, cronent_list[i]);
-  memmove(cronent_list + i, cronent_list + i + 1, sizeof(int) * cronent_count - i - 1);
+  memmove(cronent_list + i, cronent_list + i + 1, sizeof(int) * (cronent_count - i - 1));
```
djphoenix: Cheers, good catch! Dumb mistake from me...

app/modules/cron.c
```diff
@@ -104,7 +104,7 @@ static size_t lcron_findindex(lua_State *L, cronent_ud_t *ud) {
   size_t i;
   for (i = 0; i < cronent_count; i++) {
     lua_rawgeti(L, LUA_REGISTRYINDEX, cronent_list[i]);
-    eud = lua_touserdata(L, 1);
+    eud = lua_touserdata(L, -1);
```
djphoenix: Not sure what it is... But you definitely know inner mechanics better.

pjsg: The rawgeti *pushes* the integer onto the stack. This is element -1. Element 1 is the first argument in the current lua stackframe.

PR #2155, opened 2017-10-29
Another dsleep fix (or rather a not-deep-sleep case)

### LINKED ISSUE #2154: More problems with (lack of) deep sleep
Copied from https://github.com/nodemcu/nodemcu-firmware/pull/2148#issuecomment-340289308

> Unfortunately I discovered only now an issue with the updated version. When powered up (after some seconds being turned off), rtctime.get() returns a random value instead of zero. This value remains constant even after a reset but changes to a different random value after a power down.
>
> Only when rtctime is initialized via rtctime.set(0), it behaves as the previous version: after resetting the ESP, rtctime.get() constantly returns 0 until rtctime.set(0) is called again.
> 
> Or to describe it another way: an init.lua with:
> 
> print(rtctime.get())
> now prints a random value after being powered up, while formerly it reliably printed 0.

pjsg:
Fixes #2154.

We now always reset the configuration if the rtc contains data that is not a valid deep sleep setting. If it is valid, that seems to work as well.

### DIFF
--- a/app/include/rtc/rtctime_internal.h
+++ b/app/include/rtc/rtctime_internal.h
@@ -506,11 +506,6 @@ extern void rtc_time_enter_deep_sleep_final(void);
 
 static void rtc_time_enter_deep_sleep_us(uint32_t us)
 {
-  if (rtc_time_check_wake_magic())
-    rtc_time_set_sleep_magic();
-  else 
-    bbram_save();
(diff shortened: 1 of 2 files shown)


marcelstoer:
With @mkresse having tested this I think we're good to go.

PR #2171, opened 2017-11-16
esp8266: WiFi method to adjust maximum TX power

### LINKED ISSUE #2164: Add API to set WiFi RF power
### Missing feature
Add API to set WiFi RF power

### Justification
When power supply is not a problem you might want to crank up the volume on the WiFi to reach that AP. It's available in Arduino: https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ESP8266WiFiGeneric.cpp#L285
There's `	system_phy_set_max_tpw` in the SDK: http://espressif.com/sites/default/files/documentation/2c-esp8266_non_os_sdk_api_reference_en.pdf#page=31 -> chapter 3.1.13

@dnc40085 whadayathink?

### Workarounds
Move the device closer to the AP...

mdeneen:
Fixes #2164 

I added the requested method to set maximum TX power.  I wasn't sure if the API should be expressed in dBm or these quarter-dBm units that the espressif SDK uses, so I went with quarter-dBm.  There is no SDK method to obtain the current setting, so I did not add this to the Lua binding.

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -303,9 +303,7 @@ static int wifi_setmaxtxpower( lua_State* L )
 {
   unsigned power;
   power = luaL_checkinteger( L, 1 );
-
-  if ( power > 82 )
-    return luaL_error( L, "tx power out of range (0->82)");
+  luaL_argcheck(L, (power > 0 && power < 83), 1, "tx power out of range (0->82)");
 
   system_phy_set_max_tpw( (uint8_t) power);
   return 1;
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -298,6 +298,19 @@ static int wifi_getphymode( lua_State* L )
   return 1;
 }
 
+// Lua: wifi.setmaxtxpower()
+static int wifi_setmaxtxpower( lua_State* L )
+{
+  unsigned power;
+  power = luaL_checkinteger( L, 1 );
+
+  if ( power > 82 )
+    return luaL_error( L, "tx power out of range (0->82)");
+
+  system_phy_set_max_tpw( (uint8_t) power);
+  return 1;
+}
+
 #ifdef PMSLEEP_ENABLE
 /* Begin WiFi suspend functions*/
 #include "pmSleep.h"
@@ -1796,6 +1809,7 @@ static const LUA_REG_TYPE wifi_map[] =  {
   { LSTRKEY( "getchannel" ),     LFUNCVAL( wifi_getchannel ) },
   { LSTRKEY( "setphymode" ),     LFUNCVAL( wifi_setphymode ) },
   { LSTRKEY( "getphymode" ),     LFUNCVAL( wifi_getphymode ) },
+  { LSTRKEY( "setmaxtxpower" ),  LFUNCVAL( wifi_setmaxtxpower ) },
 #ifdef PMSLEEP_ENABLE
   { LSTRKEY( "suspend" ),        LFUNCVAL( wifi_suspend ) },
   { LSTRKEY( "resume" ),         LFUNCVAL( wifi_resume ) },
--- a/docs/en/modules/wifi.md
+++ b/docs/en/modules/wifi.md
@@ -213,7 +213,7 @@ physical mode after setup
 
 ## wifi.setmaxtxpower()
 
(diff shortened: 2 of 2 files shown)


pjsg:
Is this power saved or needs to be set on each boot? This has implications as to whether I should do a set power to max on every boot. Please update the documentation accordingly.

mdeneen:
Philip,

Thanks for the feedback. I have added additional documentation. 

If you're looking to permanently update the tx power, you can apparently modify esp_init_data_default.bin, byte 34.  Perhaps one could do this in a hex editor?

marcelstoer:
> There is no SDK method to obtain the current setting

Do we know what the _default_ value is? Would `wifi.setmaxtxpower(42)` increase or decrease the default TX power?

mdeneen:
I believe that it depends on the SDK, whatever they put in
esp_init_data_default.bin.  Byte 34 on mine is set to 82, which is the
maximum.

Is this file included in the final NodeMCU flash image?  If not, then the
default may depend on whatever was flashed on your esp8266 when it was
made.

On Nov 18, 2017 16:59, "Marcel Stör" <notifications@github.com> wrote:

There is no SDK method to obtain the current setting

Do we know what the *default* value is? Would wifi.setmaxtxpower(42)
increase or decrease the default TX power?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nodemcu/nodemcu-firmware/pull/2171#issuecomment-345474523>,
or mute the thread

marcelstoer:
> Is this file included in the final NodeMCU flash image?

Yes, as per our [flashing documentation](https://nodemcu.readthedocs.io/en/latest/en/flash/#sdk-init-data) NodeMCU writes this file on startup if necessary.

> ...depends on the SDK, whatever they put in esp_init_data_default.bin.  Byte 34 on mine is set to 82

Can I ask you to amend the description in the function documentation you wrote accordingly? Maybe add a (relative) link to the init-data chapter in the flashing documentation. So, if the default is 82 the documentation could hint that _lowering_ the value could help reduce power consumption for battery-driven devices.

app/modules/wifi.c
```diff
@@ -298,6 +298,19 @@ static int wifi_getphymode( lua_State* L )
   return 1;
 }
 
+// Lua: wifi.setmaxtxpower()
+static int wifi_setmaxtxpower( lua_State* L )
+{
+  unsigned power;
+  power = luaL_checkinteger( L, 1 );
+
+  if ( power > 82 )
```
devsaurus: Please use [luaL_argcheck()](http://pgl.yoyo.org/luai/i/luaL_argcheck) for checking function parameters.

docs/en/modules/wifi.md
```diff
@@ -211,6 +211,20 @@ physical mode after setup
 #### See also
 [`wifi.getphymode()`](#wifigetphymode)
 
+## wifi.setmaxtxpower()
+
+Sets WiFi maximum tx power.  This setting is not persisted across power cycles, and the documentation does not specify if the setting persists after deep sleep.
+
```
devsaurus: Maybe mention that the default setting is taken from a byte in the init data.

app/modules/wifi.c
mdeneen: I'll look at argcheck.  I don't see a reason to check for less than zero, though, because it's an unsigned number. 

Should I submit a pull request for other wifi functions which are missing argcheck?

mdeneen: Ah, I see that the function that didn't have arg checks was explicitly checking values against an enum.

mdeneen:
wow, you're faster than I thought.  I have one more doc commit if you want it.

marcelstoer:
Just commit it to your branch and I'll pick it up from there.

mdeneen:
https://github.com/mdeneen/nodemcu-firmware/commit/cc63243644b215104f6ee562b7fb0ae3fe7af8a9

mdeneen:
Thanks for merging this, Marcel.

dddsssasa:
I try method wifi.setmaxtxpower(int) and measurement current and this method doesnt affect input current. I also tested this features on arduino core firmware and method in this firmware affect input current.  (between 250uA - 350uA / 220V)

PR #2176, opened 2017-11-18
Addition of a Bloom Filter object

### LINKED ISSUE #2175: Feature request: bloom filter object
### Missing feature
Would like to have a bloom filter object.

### Justification
Given the small amount of memory on the esp8266 it is not easy to remember whether a particular object/event has been seen before. The standard way is to put the previous object into a table as a key and then you can check later to see if a new object was previously seen. This requires saving all of the objects and is heavy on the memory consumption.

A bloom filter solves this problem.

### Workarounds
You can use the table approach, but you tend to run out of memory,

pjsg:
Fixes #2175.

This is a fairly simple implementation of a bloom filter object. The limits on how many objects you can store vary based on how much memory is available. However, storing 10,000 objects with a 1 false positive error rate consumes under 8k bytes of memory.


app/modules/bloom.c
```diff
@@ -0,0 +1,192 @@
...
+ * Philip Gladstone, N1DQ
+ */
+
+#include "module.h"
+#include "lauxlib.h"
+#include "c_types.h"
+#include "../crypto/sha2.h"
+
+#if defined(LUA_USE_MODULES_BLOOM) && !defined(SHA2_ENABLE)
+#error Must have SHA2)ENABLE set for BLOOM module
```
devsaurus: ) -> _

docs/en/modules/bloom.md
```diff
@@ -0,0 +1,103 @@
+# bloom Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2017-11-13 | [Philip Gladstone](https://github.com/pjsg) | [Philip Gladstone](https://github.com/pjsg) | [bloom.c](../../../app/modules/bloom.c)|
+
+
+This module implements a bloom filter. This is a probabilistic data structure that is used to test for set membership. There are two operations -- `add` and `check` that allow 
+arbitrary strings to be added to the set or tested for set membership. Since this is a probabilistic dta structure, the answer returned can be incorrect. However,
```
devsaurus: dta -> data

docs/en/modules/bloom.md
```diff
@@ -0,0 +1,103 @@
...
+if the string *is* a member of the set, then the `check` operation will always return `true`.
```
devsaurus: Maybe add a link to https://en.wikipedia.org/wiki/Bloom_filter for further reading.

pjsg:
Thanks for fixing those issues 👍

PR #2188, opened 2017-11-29
QMC5883L support for hmc5883l.c module (#2187)

### LINKED ISSUE #2187: QMC5883L support next to HMC5883
### Missing feature
QMC5883L support next to HMC5883

### Justification
Because many cheap magnetometer modules on EBAY or ALIEXPRESS utilise QMC5883 instead of HMC5883 it would be nice to support both in nodemcu. Both modules are similar, but they have different registers structure and some additional adjustments must be done to have it working.

I've created my own version of /app/modules/hmc5883l.c working correctly with QMC sensors, but currently don't have an idea how to deal with supporting both modules in NodeMCU - create separate modules in lua for QMC and HMC, or make existing HMC library "smart" and recognize what type of chip to use. That's the open question.
I'm going to fork nodemcu-firmware, and share my changes. Also appreciate if you could help me with finding proper way of integrating it with current firmware.

### External resources
HMC5883L datasheet: https://cdn-shop.adafruit.com/datasheets/HMC5883L_3-Axis_Digital_Compass_IC.pdf
QMC5883L datasheet: http://wiki.epalsite.com/images/7/72/QMC5883L-Datasheet-1.0.pdf

Main differences between both chips:
- I2C address (QMC: 0x0D vs HMC: 0x1E)
- different ODRs (QMC: 10Hz, 50Hz, 100Hz, 200Hz vs HMC: 0.75Hz, 1.5Hz, 3Hz, 7.5Hz, 15Hz, 30Hz, 75Hz)
- data register address (QMC: 0x00:05 vs HMC: 0x03:08)
- data order (QMC: 0x00.XLSB, 0x01.XMSB, 0x02.YLSB, 0x03.YMSB, 0x04.ZLSB, 0x05.ZMSB vs HMC: 0x03.XMSB, 0x04.XLSB, 0x05.ZMSB, 0x06.ZLSB, 0x07.YMSB, 0x08.YLSB)
... and few more

infrat:
Fixes #2187 .

QMC5883L chip support introduced to hmc5883l.c library. Unfortunately, I don't have magnetometer based on HMC5883L chip to test it. My changes are tested ONLY with QMC5883L!

### DIFF
--- a/app/modules/hmc5883l.c
+++ b/app/modules/hmc5883l.c
@@ -9,56 +9,93 @@
 #include "c_stdlib.h"
 #include "c_string.h"
 
-static const uint32_t hmc5883_i2c_id = 0;
-static const uint8_t hmc5883_i2c_addr = 0x1E;
+#define QMC5883L_ADDR 0x0D
+#define HMC5883L_ADDR 0x1E
 
-static uint8_t r8u(uint32_t id, uint8_t reg) {
+static const uint32_t i2c_id = 0;
+static uint8_t i2c_addr;
+
+static uint8_t r8u(uint32_t id, uint8_t i2c_addr, uint8_t reg) {
     uint8_t ret;
 
     platform_i2c_send_start(id);
-    platform_i2c_send_address(id, hmc5883_i2c_addr, PLATFORM_I2C_DIRECTION_TRANSMITTER);
+    platform_i2c_send_address(id, i2c_addr, PLATFORM_I2C_DIRECTION_TRANSMITTER);
     platform_i2c_send_byte(id, reg);
     platform_i2c_send_stop(id);
     platform_i2c_send_start(id);
-    platform_i2c_send_address(id, hmc5883_i2c_addr, PLATFORM_I2C_DIRECTION_RECEIVER);
+    platform_i2c_send_address(id, i2c_addr, PLATFORM_I2C_DIRECTION_RECEIVER);
     ret = platform_i2c_recv_byte(id, 0);
     platform_i2c_send_stop(id);
     return ret;
 }
 
-static void w8u(uint32_t id, uint8_t reg, uint8_t val) {
-    platform_i2c_send_start(hmc5883_i2c_id);
-    platform_i2c_send_address(hmc5883_i2c_id, hmc5883_i2c_addr, PLATFORM_I2C_DIRECTION_TRANSMITTER);
-    platform_i2c_send_byte(hmc5883_i2c_id, reg);
-    platform_i2c_send_byte(hmc5883_i2c_id, val);
-    platform_i2c_send_stop(hmc5883_i2c_id);
+static void w8u(uint32_t id, uint8_t i2c_addr, uint8_t reg, uint8_t val) {
+    platform_i2c_send_start(id);
+    platform_i2c_send_address(id, i2c_addr, PLATFORM_I2C_DIRECTION_TRANSMITTER);
+    platform_i2c_send_byte(id, reg);
+    platform_i2c_send_byte(id, val);
+    platform_i2c_send_stop(id);
 }
 
-static int hmc5883_setup(lua_State* L) {
+static uint8_t xmc5883_identify() {
     uint8_t  devid_a, devid_b, devid_c;
 
-    devid_a = r8u(hmc5883_i2c_id, 10);
-    devid_b = r8u(hmc5883_i2c_id, 11);
-    devid_c = r8u(hmc5883_i2c_id, 12);
+    devid_a = r8u(i2c_id, HMC5883L_ADDR, 10);
+    devid_b = r8u(i2c_id, HMC5883L_ADDR, 11);
+    devid_c = r8u(i2c_id, HMC5883L_ADDR, 12);
 
-    if ((devid_a != 0x48) || (devid_b != 0x34) || (devid_c != 0x33)) {
-        return luaL_error(L, "device not found");
+    if ((devid_a == 0x48) && (devid_b == 0x34) && (devid_c == 0x33)) {
+        return HMC5883L_ADDR;
     }
 
-    // 8 sample average, 15 Hz update rate, normal measurement
-    w8u(hmc5883_i2c_id, 0x00, 0x70);
+    uint8_t  devid;
 
-    // Gain = 5
-    w8u(hmc5883_i2c_id, 0x01, 0xA0);
+    devid = r8u(i2c_id, QMC5883L_ADDR, 0x0D);
+    if (devid == 0xFF) {
+        return QMC5883L_ADDR;
+    }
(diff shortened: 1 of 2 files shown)


app/modules/hmc5883l.c
```diff
@@ -9,69 +9,139 @@
...
+    return luaL_error(L, "device not found");
+}
+
 
+static int xmc5883_init(lua_State* L) {
     uint32_t sda;
     uint32_t scl;
 
-    platform_print_deprecation_note("hmc5883l.init() is replaced by hmc5883l.setup()", "in the next version");
+    platform_print_deprecation_note("hmc5883l.init() is replacd by hmc5883l.setup()", "in the next version");
```
pjsg: This introduces a spelling mistake.

app/modules/hmc5883l.c
```diff
@@ -106,11 +176,22 @@ static int hmc5883_read(lua_State* L) {
     return 3;
 }
 
+static int xmc5883_read(lua_State* L) {
+    uint8_t i2c_addr;
+    i2c_addr = xmc5883_identify();
```
pjsg: This introduces a delay on each call of read(). Why not save the i2c_addr in xmc5883_i2c_addr (as part of setup/init) and then just access the right address immediately.

infrat: I was considering approach you suggested, but I wasn't really sure if nodemcu modules stores their states between each LUA command call - that's why I decided to determine chip type each time. But if it's possible to store such value, then I'll do it the way you suggest.

infrat:
Changes commited

devsaurus:
I'd merge this one if there's confirmation that HMC5883L isn't broken. @pjsg maybe?

infrat:
@devsaurus hold on for a moment, please. Few days ago I've found an minor issue with my change, when used with HMC/QMC disconnected from I2C bus. By default in such case it should throw error luaL_error(L, "device not found") but it doesn't happen (I suspect it's because module identification done here: https://github.com/infrat/nodemcu-firmware/blob/ff7bb2d525815ab65a56b03ec19776a3387f12d8/app/modules/hmc5883l.c#L55 )
After Christmas I should prepare fix for that.

It doesn't affect the library in way making it unusable, but it's a bug and I should fix it.

pjsg:
I don't have either device....

edtamarin:
I have used the HMC library from the latest master build successfully, and the "device not connected" error is thrown when the HMC is not on the bus. I have not tested this particular HMC build.

However, I had some additions to the module in mind, since, at least for the HMC chip, I think the module is too simple and can be made much better. I have opened #2206 for that and planned to make some changes over the Christmas holidays myself. Unfortunately, I only have the HMC chip so I can add features only for that at this time. I would like to hear your thoughts on this.

marcelstoer:
A fundamental problem is that the module name is equal to the chip name. Hence, someone looking for support for QMC5883L would be disappointed because he/she would only see HMC5883L in the modules list. The same applies for search engines of course.

We can change the title on line 1 in `hmc5883l.md` and we can [change the navigation item name](https://github.com/nodemcu/nodemcu-firmware/blob/master/mkdocs.yml#L58) to something like "hmc5883l / qmc5883l" (I suggest we do both) but you would still say `hmc5883l.read()` even if you run it against a QMC5883L. I'm not sure if we just have to live with that.

infrat:
Yeah, I agree with you but in fact it's quite hard to deal with that. For the moment, I see two ways: we can do it as is done in this PR, or create separate library for QMC. The question is which is better? Even I can't say which is better to me. Both solutions has pros and cons.

devsaurus:
I studied the datasheets of the two chips and understand the headaches here better now. Both only share the same model number 5883 and the fact that they're magnetometers. Apart from these, register interface and operation modes differ. Initially I though that QMC5883 would be the bigger/newer/extended version of HMC5883 but no - they're different architectures IMO.

> For the moment, I see two ways: we can do it as is done in this PR

I would avoid that for the reasons above.

> or create separate library for QMC

Not my favorite solution but provides a better separation.

Third alternative (same topic as https://github.com/nodemcu/nodemcu-firmware/pull/1242#issuecomment-212131673): Remove the chip-specific module and integrate HMC5883 and QMC5883 (and future ones) in a `magnetometer` module. Users derive objects from this class and run the `setup()` and `read()` methods on the object
```lua
hmc = magnetometer.hmc5883l(i2c_id)
newconfig = hmc:setup({MS=0, DO=4, MA=3, GN=1, MD=1}) -- normal measurement, 15Hz, 8 avg, single-shot mode
-- wait some time or wait for GPIO trigger by DRDY
x, y, z = hmc:read()

qmc = magnetometer.qmc5883l(i2c_id)
```

Implementations on C level would be quite specific to the chip type, same for the `setup()` arguments.

### Pros
- Single `magnetometer` module (or any other swift name)
- Clear separation, no need to support multiple architectures with one piece of code
- Extensible with further chip types
- Enables consistent module doc

### Cons
- Additional command to derive an object
- RAM usage for Lua object
- All-or-nothing option, qmc5883 routines use flash memory even if you only need hmc eg.
- Increased complexity on C level (slightly, IMO)

What do you think - would this third option be a feasible way out?

infrat:
@devsaurus if you compare block diagrams of both chips - they're very similar. From architectural point of view there are not too much differences between them. They have different registers addressing, and probably clock is little bit different too. Sad thing is that many manufacturers use QMC and HMC interchangeably what makes customers little bit confused (like in my case).

Regarding module code, third option you suggested I like the most. It seems to be clear enough and straightforward. Do we have any modules using such architectural approach?

devsaurus:
> Regarding module code, third option you suggested I like the most. It seems to be clear enough and straightforward. Do we have any modules using such architectural approach?

u8g and ucg are the only ones if I remember correctly, but they are quite complex to serve as a blueprint. Given there's consensus to head into this direction (@marcelstoer, @pjsg?) then I'd volunteer to set up such a module including current hmc5883l code. I don't intend to offload this maintenance effort to you 😃.

infrat:
> we're heading into a different (better) direction

OFC **direction**! because it's a magnetometer ;)

@devsaurus I think I can try to setup both modules in the way you suggested, so don't worry. I'd be really happy to have some influence on this project. At least, if I wouldn't manage I'll ask you for some help.
In this moment I'm focused on other ESP-related project, but hopefully I'll be able to find some time to do this soon. Just give me few weeks. I hope it's not really urgent?

marcelstoer:
Ok, then let's close this PR and either @infrat or @devsaurus will create a new one.

devsaurus:
> I think I can try to setup both modules in the way you suggested

Thanks you very for your support! It's highly appreciated.

It's not urgent at all from my point of view. Feel free to come up even with a prototype if you look for early feedback.

PR #2192, opened 2017-12-04
Add ChiliPeppr as option for IDE

chilipeppr:
ChiliPeppr's web IDE is ready for primetime. I'd love to get this into the docs.

### DIFF
--- a/docs/en/upload.md
+++ b/docs/en/upload.md
@@ -36,6 +36,18 @@ Source: [https://github.com/4refr0nt/luatool](https://github.com/4refr0nt/luatoo
 
 Supported platforms: OS X, Linux, Windows, anything that runs Python
 
+# ChiliPeppr ESP32 Web IDE
+
+> This is a new method via the browser of editing/uploading your Lua code to your ESP32 devices. The workspace has a Serial Port JSON Server that you run locally or remotely on your computer to let your browser talk direct to your serial ports. The Ace editor is embedded which does code highlighting and intellisense for Lua. There is a file browser for your ESP32, a serial port console, restart button, file saving, run, upload/run, and upload all. A pre-compiled binary is available with all modules compiled in so you don't have to compile the firmware on your own. There is also sample code including pure Lua libraries for websockets, joystick control, etc. There is also an XBM image creator for OLED displays that let you cut/paste PNG, JPEG, or GIF images or even drag/drop of animated GIFs to create Lua code.
+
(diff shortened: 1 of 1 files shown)


chilipeppr:
The rationale for this pull request is that ChiliPeppr's web IDE is a huge step forward for making the Nodemcu firmware for ESP32 relevant and easier to use for the developer community. ChiliPeppr's ability to let people fork the workspace and fork the widgets in it, creates an opportunity for Lua to become the dominant programming language for ESP32 with a bigger developer community than other competing platforms like Micropython, Javascript (Duktape), and the Arduino IDE.

marcelstoer:
@drawkula gave a 👎, care to explain why?

chilipeppr:
To add more color to the plans, I recently posted about a pure Lua
websockets client I wrote in the issues list. One of the next steps for
that is to let you program your esp via websockets instead of just via
serial port at 115kbaud. Chilipeppr would give you the Lua library to
install into init.lua so you can now program via TCP. Serial port Json
server is, at it's core, a websockets server. So this is a natural
progression.

If you watch my video on uploading animated gifs to esp32, it's slow, so
this would help.

It will also help pave the way to remote OTA updates for lua code.

chilipeppr:
Excellent!

marcelstoer:
The ESP32 branch version of this file is a bit behind the baseline. That should be fixed separately though.

- We never established criteria as for when to list a specific tools. It's difficult to be fair to all the tool developers. That's why we keep a disclaimer-like paragraph in the master version.
- If we allow ChiliPeppr to be listed in the ESP32 branch version why not on dev/master?

devsaurus:
> - If we allow ChiliPeppr to be listed in the ESP32 branch version why not on dev/master?

For the reasons you gave in your first point, I'd say: too many to list in the docs.

With ESP32 it's different in my opinion. We have just generic support as a by-product from esp8266-based tools like ESPlorer - fortunately - and I couldn't name any tool or IDE that will work without hiccups on the ESP32. Now ChiliPeppr's ESP32 workspace is the first true IDE targeting the ESP32 Lua up to my knowledge. This, and its good documentation for getting users started from scratch is worth mentioning IMO. Note that I'm not affiliated with ChiliPeppr 😄 I had a first look just recently.

The ESP32 situation might change in the future and I'd be gladly confronted with the same luxury issue like the esp8266 platform (although I doubt this will happen soon).

devsaurus:
I've added a note about the overall fitness of ESP8266 tools for the ESP32-based firmware (bded03540623248ec763440591c920b2072c3d7a).

PR #2198, opened 2017-12-18
Dev

bastos1505:

According with documentation the function file.readline should return up to 1024 bytes or up to find EOL. The behavior was return no more than 256 byte.
The problem was found in the line 476 of file: app/modules/file.c
static int file_readline( lua_State* L )
{
  GET_FILE_OBJ;

  return file_g_read(L, **FILE_READ_CHUNK**, '\n', fd);
}
In bold is the correction
[file.c.zip](https://github.com/nodemcu/nodemcu-firmware/files/1569782/file.c.zip)


pjsg:
I think that you created the pull request with the wrong branches. You want to make sure that there is only a single diff in your PR.

I'm going to decline this one as you are trying to merge 'dev' to 'master'.....

PR #2201, opened 2017-12-19
Update file.c

bastos1505:
Buffer size of funtion readline is wrong  (Line 480)
Should be FILE_READ_CHUNK = 1024 bytes instead 256 bytes

Buffer size wrong.

### DIFF
--- a/app/modules/file.c
+++ b/app/modules/file.c
@@ -477,7 +477,7 @@ static int file_readline( lua_State* L )
 {
   GET_FILE_OBJ;
 
-  return file_g_read(L, LUAL_BUFFERSIZE, '\n', fd);
+  return file_g_read(L, FILE_READ_CHUNK, '\n', fd);
 }
 
 // Lua: write("string")


TerryE:
@bastos1505 Carlos, this is the wrong approach for 99% of usecases, therefore it is the wrong implementation.  For the majority of users the line length will be less than `LUAL_BUFFERSIZE` (256) so the string will be allocated on the stack which is cheap and reasonably low memory footprint.  

We dropped the chunk size from 1024 to 256 because the ESP8266 only has ~48K RAM and doing this helped RAM usage a lot.

Using `FILE_READ_CHUNK` (1024) for line reads means that the chunksize is malloced and freed on every line read which is going to thrash the heap.  This is hammering almost all users to solve your point problem. So sorry, my vote is :-1: 

I feel that we should either change the documentation to reflect the current implementation and leave it to application programmers to handle this >256 byte line in **Lua**.

If the typical line length is around 40 chars, then its bad enough doing sliding `vfs_read()` calls of 256 bytes so  characters are "read" 8 times or so, but the current `file_g_read`  approach becomes silly with 1024 sliding frames.   

The alternative is to reimplement `file_g_read` to read a long line in chunks or enrich the vfs interface to allow proper record based reads.

pjsg:
Ah -- sorry that I merged it. I hadn't appreciated the other issues.

TerryE:
@pjsg, I really don't think that this is one that we should keep. Have a walk through the file.c code and see what I mean. This combined with the VFS abstraction layer is going to hit performance.
We shouldn't be doing this so solve on user's point issue that could be trivially coded around.

pjsg:
In a sense I feel guilty about the original change -- I probably made it when changing the buffer size from 1k to 256b.

bastos1505:
@TerryE, I understand your preocupation about memory heap. So, if change from 256b to 1K in the file.readline() function is bad, I believe is it still bad in file.read(). Maybe the documentation for both function should be changed to 256b so, don't?
If so, one more bug should be opened to accomplish the modification for file.read() to 256b.

TerryE:
The algorithm for reading in line by line is just _really_ dumb in terms of performance.  Just think in terms of a typical line length of 40 bytes, say:

1.  Issue seek to start of next line
2.  Allocate and read in 1K bytes from the file
3.  Scan serially though this 1K buffer for an EOL
4.  Allocate a Lua string for the (40 byte) line
5.  Dump the 1K buffer
6.  Return the Lua variable
7.  Go back to (1) for next line

No wonder people complain about reading from SPIFFS being really slow.  Given that readline is probably the most utilised file function, I just seem so sensible to me :stuck_out_tongue_winking_eye: to code this in such a terrible fashion and then slow it down more for good luck.

PR #2204, opened 2017-12-20
Add support for the wifi monitor mode in the SDK

### LINKED ISSUE #2203: Implement a way to get access to the raw management frames being received.
### Missing feature
it is not currently possible to get access to the management frames that can be received in monitor mode.

### Justification
I'm prepared to do the work!

### Workarounds
None. I want to survey the surrounding APs, and the existing scan api is not very useful.

pjsg:
Fixes #2203 

This is the addition of a new module to get access to (initially) the management frames that are being broadcast by APs. It look longer than I expected due to some undocumented subtleties in the SDK. This was finally clarified in https://github.com/espressif/ESP8266_NONOS_SDK/issues/72 as to what I was doing wrong (though I don't think that a null pointer exception is a very friendly way of signalling a problem!)

This works remarkably well on boards like the Wemos D1 Mini Pro which has an external antenna (but you do need to resolder a jumper to make the external antenna work).


marcelstoer:
I haven't tested but this looks really helpful. I would prefer having the documentation into its own file (rather than appending to WiFi). One file per module you configure in `user_modules.h` or so.

pjsg:
OK -- I wondered about that. The wifi.eventmon stuff is in the same file, but it is enabled somewhat differently. The wifi.md file is already pretty large. I'll refactor the docs.

pjsg:
Updates complete. Also improved docs a bit. I suspect that there are other use cases for this monitor mode that could be supported in the future.

PR #2214, opened 2017-12-29
Update mbedtls

nwf:

There are no documentation updates.  This is related to work on #1707 but is unlikely to fix matters by itself; the updates should be uncontroversial. :)


marcelstoer:
Thanks Nathaniel!

Btw, I asked the ARM people to fix their "bad message length" messages a month ago: https://github.com/ARMmbed/mbedtls/issues/1169. Looks like they accepted it.

FrankX0:
Thanks Nathaniel!
I did some initial testing to judge the current state.
These are my findings, requesting data trough both the http- and tsl-module.

https://testsafebrowsing.appspot.com
- **http**: ends with an error (HTTP client: Disconnected with error: 8) but the debug information shows that the page was actually retreived correctly.
- **tsl**: returns "HTTP/1.1 404 Not Found"

https://howsmyssl.com/a/check

- **http**: ends with an error (HTTP client: Disconnected with error: 8) but the debug information shows that the page was actually retreived correctly.
This time the function seems to be run twice, with the second run producing the expected result.
- **tsl**: returns: "HTTP/1.1 301 Moved Permanently"

https://www.httpbin.org/get
- **http**: client handshake failed (Reason:[-0x7780]), Disconnected with error: 9
- **tsl**: client handshake failed (Reason:[-0x7780])

So in the first two cases the http-module returns an error although the page was retrieved correctly.
In case of the tsl-module it seems in both cases an incorrect page was requested.

Hope this helps.

marcelstoer:
> https://howsmyssl.com/a/check
> - tsl: returns: "HTTP/1.1 301 Moved Permanently"

That's correct, they redirect you to https://www.howsmyssl.com/a/check

FrankX0:
Good point!
This explains the second call when using the http module: when using the full address, there is only one call, with the same result BTW.
When using the full address with the tls module, I however still get the 301 error.

marcelstoer:
> the updates should be uncontroversial

Exactly, we just need to make sure things are not getting _worse_ with this update 😉

marcelstoer:
It might maybe make sense to look at the mbedTLS changes Espressif will provide with the next SDK update (2.2): https://github.com/espressif/ESP8266_NONOS_SDK/compare/v2.1.0...master

FrankX0:
After some additional testing, things start to look better.
These are the results so far.

URL|net/TLS|
----|---|
www.howsmyssl.com|OK|
testsafebrowsing.appspot.com|OK|
requestb.in|fatal alert message (msg 40)|
www.httpbin.org|fatal alert message (msg 80)|
github.com|OK|

Issue #1352, opened 2016-06-12
add a general RGB LED library


skycoders:
### Missing feature

After working a lot with RGB LEDs together with NodeMCU, I was wondering if it makes sense and is of interest to create some native code LED library supporting cross-LED-type functionality. 
### Justification

There are several modules for digital RGB LEDs (tm1829,ws2801,ws2812) as well as the possibility to drive analog RGB LEDs with pwm. All of them need to do similiar stuff, especially color manipulation. For color fade or gradient effects, the HSV color model provides much easier ways than doing this in RGB space. Also keeping a color while completely dimming it down is not possible with RGB. Value conversion between HSV and RGB could be done in lua, but this is inherently inefficient. A LED library  module providing such things could help here. I have read that e.g. the [fastled.io](http://fastled.io/) library is successfully running on ESP8266 already, so maybe porting it or parts of it could do the job. I am only wondering how we can make it available natively within the single LED driver modules, because performing transformations native, returning all the values to lua and then sending them back to the LED module does not sound that great to me as well. 
One idea is to build a HSV-based buffer similar to the ws2812.buffer, but again, that would be limited to this one module.
### Workarounds

Do all the work in lua is possible, though probably inefficient and more memory consuming.

Any comments from the LED fans?

eku:
https://github.com/FastLED/FastLED
https://github.com/patrickjahns/RGBWWLed

PR #2215, opened 2018-01-02
ws2812 effects library

skycoders:
Contributes to idea in  #1352.

This PR adds two new modules: A color library, extracted to an own module because it is helpful for other RGB LEDs as well (e.g. PWM) and a ws2812 effects module. 

The color utils provide conversion between RGB and HSV color spaces, and a convenience function to get color wheel values based on an angle.

The ws2812 effects module provides some nice effects for ws2812 LED strips. This is an initial set, and the module itself is probably not perfect, but it does some nice stuff which might be useful for others as well. The effects have been natively implemented because LUA implementations are much too slow. In order to test the module, it must be enabled in the user_modules configuration before building the firmware.


app/include/user_modules.h
```diff
@@ -28,7 +28,8 @@
 //#define LUA_USE_MODULES_BME280
 //#define LUA_USE_MODULES_BME680
 //#define LUA_USE_MODULES_COAP
-//#define LUA_USE_MODULES_CRON
+//#define LUA_USE_MODULES_COLOR_UTILS
+#define LUA_USE_MODULES_CRON
```
devsaurus: Please disable module again.

app/include/user_modules.h
```diff
@@ -48,7 +49,7 @@
 #define LUA_USE_MODULES_MQTT
 #define LUA_USE_MODULES_NET
 #define LUA_USE_MODULES_NODE
-#define LUA_USE_MODULES_OW
+//#define LUA_USE_MODULES_OW
```
devsaurus: Please enable module again.

app/include/user_modules.h
```diff
@@ -68,7 +69,7 @@
 //#define LUA_USE_MODULES_SWITEC
 // #define LUA_USE_MODULES_TCS34725
 //#define LUA_USE_MODULES_TM1829
-#define LUA_USE_MODULES_TLS
+//#define LUA_USE_MODULES_TLS
```
devsaurus: Please enable module again.

docs/en/modules/color-utils.md
```diff
@@ -0,0 +1,62 @@
+# COLOR_UTILS Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2017-12-30 | [Konrad Huebner](https://github.com/skycoders) | [Konrad Huebner](https://github.com/skycoders) | [color_utils.c](../../../app/modules/color_utils.c)|
+
+The color_utils module provides some basic color transformations useful for color LEDs. It is used by the WS2812_EFFECTS module.
+
+
+## color\_utils.hsv2grb()
```
devsaurus: What's the purpose of the backslash here and below?

app/modules/ws2812_effects.c
```diff
@@ -0,0 +1,1074 @@
...
+    os_free((void *) state);
+  }
+  // Allocate memory and set all to zero
+  size_t size = sizeof(ws2812_effects) + buffer->colorsPerLed*sizeof(uint8_t);
+  state = (ws2812_effects *) os_zalloc(size);
+  // initialize
+  state->speed = SPEED_DEFAULT;
+  state->mode_delay = DELAY_DEFAULT;
+  state->brightness = BRIGHTNESS_DEFAULT;
+  state->buffer = buffer;
```
devsaurus: You'll need to [`luaL_ref()`](http://pgl.yoyo.org/luai/i/luaL_ref) the buffer object to prevent it from being garbage collected after this function exists.

app/modules/ws2812_effects.c
```diff
@@ -0,0 +1,1074 @@
...
+}
+
+
+/**
+* Set the active effect mode
+*/
+static int ws2812_effects_set_mode(lua_State* L) {
+
+  luaL_argcheck(L, state != NULL, 1, LIBRARY_NOT_INITIALIZED_ERROR_MSG);
+  const int mode = luaL_checkinteger(L, 1);
```
devsaurus: Quoting [`luaL_checkoption()`](http://pgl.yoyo.org/luai/i/luaL_checkoption):
> The usual convention in Lua libraries is to use strings instead of numbers to select options.

I know that our firmware is not 100% consistent here, but please think about turning the mode argument into string for the above reason. See `file`, `gpio` and other modules for how to use it.

docs/en/modules/ws2812-effects.md
```diff
@@ -0,0 +1,154 @@
...
+  - `HALLOWEEN` fills the strip with purple and orange pixels and circles them
+  - `CIRCUS_COMBUSTUS` fills the strip with red/white/black pixels and circles them
+  - `LARSON_SCANNER` is the K.I.T.T. scanner effect, based on the color set through `ws2812_effects.set_color()`
+  - `COLOR_WIPE` fills the strip pixel by pixel with the color set through `ws2812_effects.set_color()` and then starts turning pixels off again from beginning to end.
+  - `RANDOM_DOT` sets random dots to the color set through `ws2812_effects.set_color()` and fades them out again
+  - `CYCLE` takes the buffer as-is and cycles it. With the second parameter it can be defined how many pixels the shift will be. Negative values shift to opposite direction.
+- `effect_param` is an optional effect parameter. See the effect modes for further explanations. It can be an integer value or a string.
+
+#### Returns
+`nil`
```
devsaurus: Please add an example that shows how to use the functions.

devsaurus:
First of all, I like these additions to enhance the RGB capabilities of the firmware.

My main concern is that we're missing a more generic model. So far there's the buffer class being part of the ws2812 module with some basic functions and effects. Now there's another module adding effects to the buffer objects. If this is ever extended for other LED modules it probably will need refactoring on the API level.

On a more specific level: The effects are only applicable to a single stripe at the moment. Adding support for dual mode and up to 8 stripes for the ESP32 needs a different approach. Probably an own effects class where you derive objects for each buffer/stripe from. What are your thoughts here?

skycoders:
I share your thoughts, however I thought, if I don't try to bring this into the firmware somewhen, I will never do so - that is why I initiated the pull request. 

Right now, I don't think that the "old" PWM based strips will really need something like this, as they are just too limited. However, I also stumbled upon SK6812, which is similar to APA102, but not identical. Also, having the current buffer objects with RGB(W) values is suboptimal, as they do not allow lossless brightness adjustments. My final hope is to have a buffer object in the HSV value range and perform effects on this buffer, and then have different writers for each LED strip type, and individual strip connected to the chip. 

Unfortunately, I am no LUA module pro, and also miss the time to get all this done fast enough. Maybe there is someone who wants to join? I'd also love to look at the ESP32. I have one, but so far didn't manage to build the firmware.

So, before spending too much time in trying to fix the individual comments, it maybe makes sense to get a common understanding of how and where we should go from here. Thanks!

docs/en/modules/color-utils.md
skycoders: my one editor had an issue without the backslash, but Atom doesn't - I will remove them all.

app/modules/ws2812_effects.c
skycoders: I am fine with adjusting this, I just "cloned" the way it is done in the ws2812 module itself (for fading and single/dual mode). This was the closest thing to look at

app/modules/ws2812.c
```diff
@@ -399,19 +420,17 @@ static int ws2812_buffer_mix(lua_State* L) {
...
   size_t i;
   for (i = 0; i < cells; i++) {
-    int32_t val = 0;
+    int val = 0;
     for (src = 0; src < n_sources; src++) {
-      val += (int32_t)(source[src].values[i] * source[src].factor);
+      val += ((int)(source[src].values[i] * source[src].factor) >> 8);
     }
 
-    val >>= 8;
```
pjsg: I think that this was pulled out of the loop to enable smooth fading. Why move it back?

skycoders: Thanks. That was probably an accident, I did not intend to change this.

app/include/user_modules.h
```diff
@@ -29,7 +29,7 @@
 //#define LUA_USE_MODULES_BME680
 //#define LUA_USE_MODULES_COAP
 //#define LUA_USE_MODULES_COLOR_UTILS
-#define LUA_USE_MODULES_CRON
+//define LUA_USE_MODULES_CRON
```
marcelstoer: Don't fiddle with the default modules!

app/include/user_modules.h
```diff
@@ -49,7 +49,7 @@
 #define LUA_USE_MODULES_MQTT
 #define LUA_USE_MODULES_NET
 #define LUA_USE_MODULES_NODE
-//#define LUA_USE_MODULES_OW
+#define LUA_USE_MODULES_OW
```
marcelstoer: dito

app/include/user_modules.h
```diff
@@ -69,7 +69,7 @@
 //#define LUA_USE_MODULES_SWITEC
 // #define LUA_USE_MODULES_TCS34725
 //#define LUA_USE_MODULES_TM1829
-//#define LUA_USE_MODULES_TLS
+#define LUA_USE_MODULES_TLS
```
marcelstoer: dito

skycoders:
I copied the current DEV branch user_modules.h file and just added the two modules of this PR. This should be OK now hopefully. Never intended to change them.

PR #2217, opened 2018-01-03
Fix batch multi-line UART input discarding all but first line

jmattsson:
- [n/a] The code changes are reflected in the documentation at `docs/en/*`.

Currently, if sending Lua commands to the ESP over UART one needs to have a bit of a break between each command. This is most noticeable if you're feeding commands from another micro controller to the ESP.

I finally tracked down the cause of this. It turns out that we're at most ever handing one line over to the LVM for each input-has-arrived notification. If that input consisted of multiple lines, we end up silently gobbling them up. This PR now splits up such input and gets each line through to the LVM. You'll still want to avoid filling up the UART input buffer so the LVM gets a chance to process it, but sending a few short lines in a batch is no longer a problem.

### DIFF
--- a/app/lua/lua.c
+++ b/app/lua/lua.c
@@ -599,11 +599,12 @@ static bool readline(lua_Load *load){
         {
           /* Get a empty line, then go to get a new line */
           c_puts(load->prmt);
+          continue;
         } else {
           load->done = 1;
           need_dojob = true;
+          break;
         }
-        continue;
       }
 
       /* other control character or not an acsii character */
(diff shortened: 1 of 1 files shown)


pjsg:
I think this problem bit me this evening......

TerryE:
Johny, the standard Lua algo here is that that the interpreter tries to compile up the source up to the EOL, and if the parser give an incomplete statement error then it them waits until he next line is received and tacks this on ... until the string read in is syntactically complete and correct or until a compile error occurs.  The current NodeMCU 5.1 implementation is flaky and doesn't handle continuations robustly, and I ended up rewriting this for the 5.3 implementation to bring this closer to the standard Lua algo.  This also impact `node.input()`.  I'll compare this to my Lua 5.3 code.

Issue #2145, opened 2017-10-21
Create unit tests in Lua


marcelstoer:
It would be extremely helpful to have a single Lua script that anyone could run to test hardware-independent modules.

Whenever we cut a release we do so w/o having done any structured and reproducible testing. I understand and accept that it's difficult to do serious (automated) testing for sensor/actor modules because it requires the respective hardware to be installed. However, it'd be fairly straightforward to test all other modules at least partially e.g. WiFi (basic stuff), timer, net/tls, HTTP, sJSON, SNTP, perf, MQTT (against public broker), file, cron, bit etc.

This is related to #516 (i.e. a light version of it).

HHHartmann:
I strongly agree.

A short research brought up the following page: [lua-users.org/wiki/UnitTesting](http://lua-users.org/wiki/UnitTesting )
I have not tested any of them but some are marked as minimal or tiny which would propably be a good place to start at. Some might not be suitable out of the box though. [minctest-lua](https://github.com/codeplea/minctest-lua) for example uses os.clock().

I would also like to include a check for required module(s) in each test or testfile so they could be marked appropriately in the results if the prerequisites fail. (or not be executed at all).

To make sure that arbitrary sets of tests can be run by individuals I could imagine to have a naming schema (like "test-MyFavoriteTest") and a script that runs all matching files.
That way only needed tests can be downloaded to the esp and run without managing any lists.

just my 2c

marcelstoer:
>  minctest-lua for example uses os.clock()

And `io.*`, `math.*`, and `debug.*`... I don't expect we can use any of those other than for inspiration.

> include a check for required module(s) in each test or testfile so they could be marked appropriately in the results if the prerequisites fail.

👍

> To make sure that arbitrary sets of tests can be run by individuals 

I believe it's important to keep the execution & preparation effort minimal. Copy single file to device and run it. If each test checks whether the required module is available I don't see the need for arbitrary sets.

A have a few ideas how to improve this but I prefer to see working software rather than discussing the perfect solution for weeks and months 😉

TerryE:
Marcel,  there is already an established [test suite](https://www.lua.org/tests/) for standard Lua and I use it quite a lot for stress testing host variants of my build, e.g. the LCD and the current LFS patches as well as my 5.3 work.  The main issue is that it's got far too big a footprint to use on standard NodeMCU builds.  Having an LFS will help a lot here, plus a robust Lua provisioning system to download the test overlays.

[...]

PR #2231, opened 2018-01-17
added support for ads1015

### LINKED ISSUE #2223: support for ads1015
### Missing feature
When using ads1015 chip, the existing ads1115 module can be used. However, not everything maps 1:1.

- The single shot mode, module programs an esp timer with the values which are incorrect. The interval values are longer than necessary (e.g. 10ms is programmed  where 1ms would be sufficient)

- Interpretation of constant values used to program sampling frequency is different (e.g. ads1115.DR_8SPS is equivalent of ads1015.DR_128SPS)

### Justification
ads1015 breakout boards are is available from adafruiit and ali

### Workarounds

- ads1115module can be used in continuous mode

- use ads1115 module as-is with known limitations

paweljasinski:
Fixes #2223.

This PR introduces support for ADS1015 based on the code for ADS1115.
In addition this PR includes:
* support for up to 4 converters (hardware limit).
* ads1115.setup() does not perform i2c general reset anymore. Reset is available as a separate function.

Due to formatting changes (removal of the tabs), it is probably good idea to use `?w=1` when looking at the diff.

### DIFF
--- a/app/modules/ads1115.c
+++ b/app/modules/ads1115.c
@@ -111,6 +111,7 @@ static const uint8_t ads1115_i2c_id = 0;
 static const uint8_t general_i2c_addr = 0x00;
 static const uint8_t ads1115_i2c_reset = 0x06;
 static const char metatable_name[] = "ads1115.device";
+static const char unexpected_value[] = "unexpected value";
 
 typedef struct {
     uint8_t i2c_addr;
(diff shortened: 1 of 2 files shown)


devsaurus:
This appears to be a good extension of the current module, thanks a lot!

Apart from adding support for ADS1015 chips, it also enables having more than one converter attached to the ESP. I see that you arranged for a descriptor array which covers up to four of them. That caught my interest as your implementation is now very close to an object-oriented approach. Except for the API, which acquired complexity with the I2C address selection parameter for each command.

How about moving the API also towards objects, for example:
```lua
conv1 = ads1115.setup(<i2c_id>, ads1115.ADDR_GND, ads1115.ADS1115)
conv2 = ads1115.setup(<i2c_id>, ads1115.ADDR_VDD, ads1115.ADS1015)

conv1:setting(<gain>, <dr>, ...)
conv2:setting(<gain>, <dr>, ...)

conv1:startread()
...
conv1:read()
```

The `<i2c_id>` aims for future compatibility with the ESP32 which supports multiple I2C buses.

You could even differentiate between the sub-types with dedicated constructors instead of the generic `setup()`:
```lua
conv1 = ads1115.ads1115(<i2c_id>, ads1115.ADDR_GND)
conv2 = ads1115.ads1015(<i2c_id>, ads1115.ADDR_VDD)
```
This would be very close to the approach we discussed in #2188.

What do you think?

paweljasinski:
@devsaurus, yes, the oop API looks better, particularly when esp32 and multiple i2c buses are considered. Is there a mechanism to share module code between 8266 and 32 branches?

marcelstoer:
>  Is there a mechanism to share module code between 8266 and 32 branches?

There isn't, sorry.

paweljasinski:
I didn't squash to make move to oop API easier for review.

paweljasinski:
How about testing artifacts? I have a couple of scripts which together with a hardware schematics could be checked in.

marcelstoer:
> I didn't squash to make move to oop API easier for review.

Good decision, thanks.

> How about testing artifacts?

This is/was discussed in #2145 -> no established mechanism yet.

app/modules/ads1115.c
```diff
@@ -581,11 +557,23 @@ static const LUA_REG_TYPE ads1115_map[] = {
     {   LSTRKEY( "COMP_1CONV" ),    LNUMVAL(ADS1115_CQUE_1CONV)     },
     {   LSTRKEY( "COMP_2CONV" ),    LNUMVAL(ADS1115_CQUE_2CONV)     },
     {   LSTRKEY( "COMP_4CONV" ),    LNUMVAL(ADS1115_CQUE_4CONV)     },
-    {   LSTRKEY( "ADS1015"),        LNUMVAL(ADS1115_ADS1015)        },
-    {   LSTRKEY( "ADS1115"),        LNUMVAL(ADS1115_ADS1115)        },
     {   LSTRKEY( "CMODE_TRAD"),     LNUMVAL(ADS1115_CMODE_TRAD)     },
     {   LSTRKEY( "CMODE_WINDOW"),   LNUMVAL(ADS1115_CMODE_WINDOW)   },
     {   LNILKEY, LNILVAL                                            }
```
devsaurus: The `__gc` index is not defined - I suspect memory leaks when the device object is disposed by the garbage collector while `timer_ref` still holds a reference to the callback. Similarly, an armed os_timer would need to be shut down gracefully.

app/modules/ads1115.c
```diff
@@ -225,36 +226,39 @@ static int ads1115_lua_reset(lua_State *L) {
...
+}
+
+static int ads1115_lua_register(lua_State *L, uint8_t chip_id) {
     if (!lua_isnumber(L, 1) || !lua_isnumber(L, 2)) {
         return luaL_error(L, "wrong arg range");
     }
-    uint8_t i2c_addr = luaL_checkinteger(L, 1);
+    uint8_t i2c_id = luaL_checkinteger(L, 1);
+    if (i2c_id != 0) {
+        return luaL_error(L, "Invalid argument: i2c_id");
```
devsaurus: It's recommended to use [`luaL_argcheck()`](http://pgl.yoyo.org/luai/i/luaL_argcheck) when checking arguments.
I understand that you follow the pattern in the existing code, but unfortunately that was already inconsistent.

app/modules/ads1115.c
paweljasinski: Somehow I convinced myself that it will not happen, but on the second iteration I believe it will be worse than memory leak.
1. Timer is armed.
2. Device object is garbage collected.
3. Timer fires, with ads_ctrl referencing user data which can be anything at this moment.

app/modules/ads1115.c
```diff
@@ -513,6 +495,16 @@ static int ads1115_lua_read(lua_State *L) {
     return 3;
 }
 
+static int ads1115_lua_delete(lua_State *L) {
+    ads_ctrl_ud_t *ads_ctrl = luaL_checkudata(L, 1, metatable_name);
+    if (ads_ctrl->timer_ref != LUA_NOREF) {
+        os_timer_disarm(&ads_ctrl->timer);
+        lua_rawgeti(L, LUA_REGISTRYINDEX, ads_ctrl->timer_ref);
```
devsaurus: Why push timer_ref onto the stack? It's redundant at least and might leak memory since the return below says 0 results (not sure, can't run the code without a chip).

paweljasinski: A bit of cargo on my side - I took a snipped out of readoutdone without fully understanding what it does. Fixing it explained some other anomaly I have seen.

paweljasinski:
I gave a try to mispec mentioned in #2145. Here is the result: https://github.com/paweljasinski/nodemcu-mispec/blob/master/nodemcu/mispec_ads1115.lua

PR #2249, opened 2018-02-04
Add support for HTU21D/SHT21 temperature and humidity sensors

pawel-sw:

HTU21D is cheap digital temperature and humidity sensor.


app/modules/htu21.c
```diff
@@ -0,0 +1,101 @@
...
+    if (remainder & (uint32_t) 1 << (23 - i)) //Check if there is a one in the left position
+      remainder ^= divsor;
+
+    divsor >>= 1; //Rotate the divsor max 16 times so that we have 8 bits left of a remainder
+  }
+
+  return (uint8_t) remainder;
+}
+
+static uint16_t ICACHE_FLASH_ATTR read_with_crc_check(lua_State *L, uint8_t reg)
```
pjsg: You don;t need the ICACHE_FLASH_ATTR as all nodemcu functions default to flash unless otherwise specified.

s-pw: thank you, I've removed the `ICACHE_FLASH_ATTR`

devsaurus:
Please add I2C id as parameter to `htu21.read()` and check against `platform_i2c_exists()`. This seems redundant on ESP8266 platform, but helps future compatibility with ESP32.

HHHartmann:
Hmm, this PR unfortunately is so old that it does not contain several changes we did. Pity that it wasn't picked up earlier.
Given that  we now encourage pure Lua modules over C modules unless there is a reason for C (e.g. timing) this PR should be rewritten in Lua.
Same probably applies to #2261.

PR #2250, opened 2018-02-04
Add support for HTU21D/SHT21 temperature and humidity sensors (ESP32)

s-pw:

HTU21D is cheap digital temperature and humidity sensor.
Module for ESP32

### DIFF
--- a/components/modules/Kconfig
+++ b/components/modules/Kconfig
@@ -98,6 +98,12 @@ config LUA_MODULE_GPIO
   help
       Includes the GPIO module (recommended).
 
+config LUA_MODULE_HTU21
+  bool "HTU21D/STH21 module"
+  default "n"
+  help
+      Includes the htu21 module.
+
 config LUA_MODULE_I2C
   bool "I2C module"
   default "y"
--- a/components/modules/htu21.c
+++ b/components/modules/htu21.c
@@ -1,0 +1,29 @@
+#include "module.h"
+#include "lauxlib.h"
+#include <stdint.h>
+
+#include "platform.h"
+
+
+static int htu21_read(lua_State *L)
+{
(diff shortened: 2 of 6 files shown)


devsaurus:
Thanks for your contribution @s-pw!

Comparing this PR with #2249 I see that you introduced a `platform_htu21_*` layer for ESP32. The same pattern is visible in your PRs for BH1750. IMO this is a valid approach since I2C interfaces are not yet fully abstracted to platform layer on ESP32. Which means:
* `platform_htu21` (and the Lua) module is restricted to `i2c.SW` interface
* support for `i2c.HW0|1` can be added in the future to `components/platform/htu21.c`
* once `platform_i2c_*` also abstracts the HW interfaces, then `platform_htu21` can be simplified or even obsoleted

But there is currently no way to distinguish the I2C interface. Therefore I propose that you add the i2c id to the APIs on Lua and platform level:
* `htu21.read(i2c_id)`
  The Lua module passes `id` to platform_htu21 and checks for `PLATFORM_OK` result
* `platform_htu21_read(unsigned id, uint8_t reg)`
  Checks for support of requested `id` and returns PLATFORM_ERR if not supported.

s-pw:
Thanks, I like the idea. I'll update modules for both ESP32 and ESP8266. ESP8266 won't be just for compatibility as the sensor supports speed up to i2c.FAST (400kHz). There is software i2c interface that could support i2c.FAST and i2c.FASTPLUS : https://github.com/pasko-zh/brzo_i2c

marcelstoer:
This applies to  #2249,  #2250 (this one),  #2259, and  #2261

@devsaurus @pjsg Do you have any idea why these modules were never merged? Was it because of @TerryE's comment https://github.com/nodemcu/nodemcu-firmware/pull/2261#issuecomment-373763423?

PR #2255, opened 2018-02-06
Added support for custom rfswitch protocol definition (id = 0)

thornley-david:

Exposes the ability to configure a custom protocol definition
Add support for start/stop bit framing
Fixed issue with gpio line kept high (keying/flooding radio)

### DIFF
--- a/app/modules/rfswitch.c
+++ b/app/modules/rfswitch.c
@@ -28,6 +28,8 @@ typedef struct Protocol {
   HighLow one;
   /** @brief if true inverts the high and low logic levels in the HighLow structs */
   bool invertedSignal;
+  int frameStartStop;       // 0=none, 1=START, 2=STOP, 3=BOTH
+  bool invertStartStop;
 } Protocol;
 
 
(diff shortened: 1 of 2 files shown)


app/modules/rfswitch.c
```diff
@@ -76,19 +98,25 @@ void transmit(HighLow pulses, bool invertedSignal, int pulseLength, int pin) {
...
       if (value & (1L << i))
         transmit(p.one, p.invertedSignal, pulse_length, pin);
       else
         transmit(p.zero, p.invertedSignal, pulse_length, pin);
     }
+    if((p.frameStartStop & 2) > 0) transmit(p.invertStartStop ? p.one : p.zero, p.invertedSignal, pulse_length, pin);
+
     transmit(p.syncFactor, p.invertedSignal, pulse_length, pin);
   }
+  platform_gpio_write(pin, 0);      // If we leave gpio 0 high on last transmit, we flood the rf band (so we should always set low this after sending waveforms)
```
landswipe: Looking back at the previous fix #1738, the correct thing to do here might be to set the GPIO pin state to p.invertedSignal

`platform_gpio_write(pin, p.invertedSignal); `

app/modules/rfswitch.c
```diff
@@ -76,19 +98,25 @@ void transmit(HighLow pulses, bool invertedSignal, int pulseLength, int pin) {
  */
 void send(unsigned long protocol_id, unsigned long pulse_length, unsigned long repeat, unsigned long pin, unsigned long value, unsigned int length) {
   platform_gpio_mode(pin, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
-  Protocol p = proto[protocol_id-1];
+  Protocol p = custom;
+  if(protocol_id > 0) proto[protocol_id-1];
```
pjsg: I think that an assignement to p is missing here.

tongtong15:
mute the thread

app/modules/rfswitch.c
landswipe: yes this is an oversight, i'll fix and update. thanks!

app/modules/rfswitch.c
landswipe: I reviewed this again, and in one of my usage scenarios I actually set invertedSignal to true. Since my hardware's 433 transmitter is active high, it should still 0 as I have set it here. @ffedoroff who authored #1738 might be able to comment on whether or not this is still sufficient for his scenario given he originally moved away from this approach.

landswipe:
fixed that embarrassing assignment mishap ;)

app/modules/rfswitch.c
```diff
@@ -51,14 +53,34 @@ typedef struct Protocol {
...
-  { 450, { 23,  1 }, {  1,  2 }, {  2,  1 }, true }      // protocol 6 (HT6P20B)
+  { 350, {  1, 31 }, {  1,  3 }, {  3,  1 }, false, 0, false },    // protocol 1
+  { 650, {  1, 10 }, {  1,  2 }, {  2,  1 }, false, 0, false },    // protocol 2
+  { 100, { 30, 71 }, {  4, 11 }, {  9,  6 }, false, 0, false },    // protocol 3
+  { 380, {  1,  6 }, {  1,  3 }, {  3,  1 }, false, 0, false },    // protocol 4
+  { 500, {  6, 14 }, {  1,  2 }, {  2,  1 }, false, 0, false },    // protocol 5
+  { 450, { 23,  1 }, {  1,  2 }, {  2,  1 }, true,  0, false }     // protocol 6 (HT6P20B)
 };
 
+Protocol custom = { 350, {  1, 31 }, {  1,  3 }, {  3,  1 }, false, false };  // Defaults to same as protocol 1
```
pjsg: It would be nice if this was the same as protocol 1 -- there is a missing 0 near the end of the initialization. It doesn't really make a difference, but it ought to be fixed.

app/modules/rfswitch.c
```diff
@@ -76,19 +98,25 @@ void transmit(HighLow pulses, bool invertedSignal, int pulseLength, int pin) {
  */
 void send(unsigned long protocol_id, unsigned long pulse_length, unsigned long repeat, unsigned long pin, unsigned long value, unsigned int length) {
   platform_gpio_mode(pin, PLATFORM_GPIO_OUTPUT, PLATFORM_GPIO_FLOAT);
-  Protocol p = proto[protocol_id-1];
+  Protocol p = custom;
+  if(protocol_id > 0) p = proto[protocol_id-1];
```
pjsg: Can you add a check for having the protocol id in range somewhere? It should probably throw a lua error if it is out of range. Otherwise I suspect that rfswitch.send(9999999, ...) will cause the system to fault.

marcelstoer:
@landswipe do you plan to address the latest comments by @pjsg?

landswipe:
Yes I will... I had forgotten about this, thanks for the reminder :)

marcelstoer:
This is now stale and has merge conflicts. Shall we close it?

PR #2260, opened 2018-02-12
Fix a number of issues with the gpio.pulse family of functions

### LINKED ISSUE #2254: Issues with gpio.pulse
### Expected behavior
Handling of adjusts should work even when value is large
### Actual behavior
It gets confused if the new delay exceeds min or max

### NodeMCU version
dev

Also, increase the maximum value of delay above one second.

pjsg:
Fixes #2254.

I'm now trying to make progress on my clock project and discovering some bugs in the existing code, and missing functionality. This PR adds the missing features and resolves the bugs.

### DIFF
--- a/app/modules/gpio_pulse.c
+++ b/app/modules/gpio_pulse.c
@@ -142,8 +142,8 @@ static void fill_entry_from_table(lua_State *L, pulse_entry_t *entry) {
   while (lua_next(L, -2)) {
     // stack now contains: -1 => value; -2 => key; -3 => table
     if (lua_type(L, -2) == LUA_TNUMBER) {
-      int pin = lua_tonumber(L, -2);
-      int value = lua_tonumber(L, -1);
+      int pin = luaL_checkint(L, -2);
+      int value = luaL_checkint(L, -1);
 
       if (pin < 0 || pin >= GPIO_PIN_NUM) {
         luaL_error(L, "pin number %d must be in range 0 .. %d", pin, GPIO_PIN_NUM - 1);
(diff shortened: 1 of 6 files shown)


app/modules/gpio_pulse.c
```diff
@@ -137,6 +137,69 @@ static int gpio_pulse_delete(lua_State *L) {
...
+  if (lua_type(L, -1) != LUA_TTABLE) {
+    luaL_error(L, "All entries must be tables");
+  }
+
+  lua_pushnil(L);             // key position
+  while (lua_next(L, -2)) {
+    // stack now contains: -1 => value; -2 => key; -3 => table
+    if (lua_type(L, -2) == LUA_TNUMBER) {
+      int pin = lua_tonumber(L, -2);
+      int value = lua_tonumber(L, -1);
```
devsaurus: What if value at -1 isn't LUA_TNUMBER or a convertible string?

app/modules/gpio_pulse.c
```diff
@@ -137,6 +137,69 @@ static int gpio_pulse_delete(lua_State *L) {
...
+        entry->gpio_set |= BIT(pin_num[pin]);
+      } else {
+        entry->gpio_clr |= BIT(pin_num[pin]);
+      }
+    } else {
+      const char *str = lua_tostring(L, -2);
+
+      if (strcmp(str, "delay") == 0) {
+        entry->delay = lua_tonumber(L, -1);
+        if (entry->delay < 0 || entry->delay > DELAY_LIMIT) {
```
devsaurus: I expect that this won't throw an error if value at index `-1` is not convertible to a number. [`lua_tonumber()`](http://pgl.yoyo.org/luai/i/lua_tonumber) returns 0 in this case and the range check is happy.

`luaL_checknumber()` or an explicit test for `lua_isnumber()` if delay == 0 would detect nonsense.

Applies to further instances of `lua_tonumber()` below.

pjsg:
The changes aren't as big as they look as I moved the guts of one function into its own function so that I could use it elsewhere.

I did fix all the unchecked calls so that there is more error detection (and it actually found an error in my current application!)

pjsg:
This also includes a fix to the issue in https://stackoverflow.com/q/48790455/131929 -- I changed gpio.mode(, gpio.OUTPUT) to actually enable the driver.

pjsg:
I also addressed the comments in #2265

Issue #2241, opened 2018-01-28
BME280 module does not allow coexistence of two BME280s on the same I2C bus


mickley:
8<------------------------ BUG REPORT -----------------------------------------
### Expected behavior
The BME280 supports two different I2C addresses, so two BME280s can be used on the same bus.  This should be possible under nodemcu

### Actual behavior
The bme280 module doesn't allow the I2C address to be set and instead looks for devices on the first address, and failing to find one, looks on the second address.  Therefore, if 2 devices are present, only the first can be used.  

It would be better if BME280 functions had an optional address parameter to allow for multiple devices.  

### NodeMCU version
Dev: https://github.com/nodemcu/nodemcu-firmware/commit/ed56d949ee4d00ec10e29ec91d87bcc9b76b9d86

### Hardware
WeMos D1 Mini, with two BME280s connected to the same i2c bus.

8<------------------------ END BUG REPORT -------------------------------------

Fisherman48:
Just shift sda - scl wires to the second and call setup
i2c.setup(0, 3, 4, i2c.SLOW)
i2c.setup(0, 4, 3, i2c.SLOW)
I have done this with 6 BME280 using 3 pins (charlieplexing 3x2=6)

mickley:
*sigh*.  I'm shocked you're closing this with such a hack solution to a rather simple problem.  The whole point of the I2c bus is to NOT have to do this.  It's an acceptable short-term hack, but not really a fix.

For example, when one has hundreds of setups and only 4 rather long wires leading to the sensors, while it is technically possible to split sensors up and switch the wires, it's really not worth the time.

heikomat:
I too am facing the exact same issue, and using the same pin as clock for one and data for the other device seems really hacky to me aswell.

marcelstoer:
Yeah, closing it might actually have discouraged some folks from fixing it properly.

TerryE:
This isn't a bug per se.  The current behaviour is a functional limitation and removing this is really an enhancement.  If @heikomat or @mickley want to do the work on the module, test and submit as a PR then I will be happy to review and merge it.

heikomat:
@TerryE I'll see what i can do 👍

heikomat:
I've got it working. Will open a PR soon.

heikomat:
Done. @TerryE i'd appreciate a review :)
#3124

PR #2261, opened 2018-02-12
Add support for BH1750FVI ambient light sensors

pawel-sw:


devsaurus:
Please add I2C id as parameter to the Lua functions, also see https://github.com/nodemcu/nodemcu-firmware/pull/2249#issuecomment-370220435.

s-pw:
@TerryE it is a stateless component. If you store i2c_id during setup, you won't be able to use 2 sensors on 2 separate I2C buses.

TerryE:
> @TerryE it is a stateless component. If you store i2c_id during setup, you won't be able to use 2 sensors on 2 separate I2C buses.

So what happens when you setup bus 0 and attempt to read bus 15243523 or even bus 1?    This is an ambient light sensor.. If the manufacturers really thought that the same bus master was going to need to read two sensors then the'd have exposed a couple of address pins as pull up / down pins on the IC.

mickley:
I'm currently using a setup with two BH1750FVI sensors and my own lua module, and could forsee needing to use more than two.  The BH1750 actually does have an address pin too, so two are supported on the same bus, which is what I am doing.  This module doesn't have an option to use I2C address 0x5C (@s-pw).

Moreover, I think needlessly restricting useful features and forcing hack-style programming for stuff like multiple sensors is bad form.  I've had the same problem with BME280, which doesn't support multiple sensors on the same bus (#2241 ).  These are the sorts of basic things that should be supported for every sensor!

TerryE:
> I think needlessly restricting useful features and forcing hack-style programming for stuff like multiple sensors is bad form

In which case the setup should return a userdata object and all further references should use use Lua class features.  I didn't suggest that you should hack this.  Do single instance _or_  multi-instance properly.

PR #2267, opened 2018-02-20
Update mbedTLS to 2.7.0

nwf:
While here, export mbedTLS's debugging threshold knob to Lua code.  Tested through HTTP module.


marcelstoer:
I ran a few test cases from #1707 as well as non-TLS HTTP requests. Other than what we already know I didn't notice any problems.

PR #2283, opened 2018-03-01
Fix SNTP so that it retries DNS lookups if it needs to

### LINKED ISSUE #2277: sntp.sync fails even after Internet connection is established (in auto-repeat mode)
### Expected behavior
`sntp.sync` must perform sync successfully after Internet connection is established.

### Actual behavior
Here are the specific steps to reproduce:

1. Disconnect the Internet connection from your hotspot/router
2. Boot and connect ESP8266 to the router (This is important. ESP8266 must freshly boot and `sntp.sync` must fail on 1st attempt)
3. Start `sntp.sync` in auto-repeat mode
4. Reconnect the Internet after 2-3 failed attempts of `sntp.sync`
5. Actual Behavior: `sntp.sync` still keeps on failing with error code `4`

Note: the pattern of errors: error code `1` (DNS lookup failed) followed by many `4`(Timeout, no NTP response received) is necessary to reproduce the problem.

### Test code
Providing a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) which will reproduce the problem.
```Lua

	branch: master
	commit: 5073c199c01d4d7bbbcd0ae1f761ecc4687f7217
	SSL: false
	modules: file,gpio,net,node,rtctime,sntp,tmr,uart,wifi
 build 	built on: 2018-02-26 00:15
lua: cannot open init.lua
> =wifi.setmode(wifi.STATION)==wifi.STATION
true
> =wifi.sta.getip()
nil
> =wifi.sta.config({ssid="<ssid>", pwd="<password>"})
true
> =wifi.sta.connect()
> =wifi.sta.getip()
10.10.10.1	255.255.255.0	10.10.10.254
> sntp.sync(nil, function(sec) print(sec) end, function(err_code, err_desc) print(err_code, err_desc) end, 1)
> 1	0.nodemcu.pool.ntp.org
4	nil
4	nil
4	nil
4	nil
```

Note: after ESP printed `1`, `4`, `4`; the Internet connection was re-established and `sntp.sync` kept on failing with same error code.

### NodeMCU version
Branch: master
Revision: 5073c199c01d4d7bbbcd0ae1f761ecc4687f7217 (latest)

### Hardware
NodeMCU version 2 Amica

pjsg:
Fixes #2277

This retries the DNS lookups on each sync operation if the last sync operation ended in a timeout with no good responses, or if the dns lookups failed on the previous attempt. THis deals with the case when you start without a connection to the internet and the dns lookups fail. Then you get connected, and the next sync will do the lookups and then sync the time. Those IPs will continue to be used until they stop working (maybe because the connection goes away). Then the DNS lookups are retried until good answers are received.

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -225,6 +225,10 @@ static void sntp_handle_result(lua_State *L) {
   const uint32_t MICROSECONDS = 1000000;
 
   if (state->best.stratum == 0) {
+    // This could be because none of the servers are reachable, or maybe we haven't been able to look 
+    // them up.
+    state->lookup_pos = 0; // Reset for next time.
+    server_count = 0;      // Reset for next time.
     handle_error(L, NTP_TIMEOUT_ERR, NULL);
     return;
   }
@@ -614,10 +618,11 @@ static int sntp_getoffset(lua_State *L)
 
 static void sntp_dolookups (lua_State *L) {
   // Step through each element of the table, converting it to an address
-  // at the end, start the lookups
-  //
+  // at the end, start the lookups. If we have already looked everything up,
+  // then move straight to sending the packets.
   if (state->list_ref == LUA_NOREF) {
     sntp_dosend(L);
+    return;
   }
 
   lua_rawgeti(L, LUA_REGISTRYINDEX, state->list_ref);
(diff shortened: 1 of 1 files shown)


FrankX0:
Works great! Thanks.

PR #2297, opened 2018-03-12
Removing large number of #defines from Makefile and push them to a separate file

### LINKED ISSUE #1863: sqlite3 support

luizfeliperj:
Fixes #1863 ("Feature enhancement").

This pull request removes logs of defines from inside the makefile and pushes them into a separated file. This file is sources by the compiles by a flag that forces its inclusion on all source files within sqlite3 directory (-include config_ext.h).


TerryE:
Why is  SQLITE _DEFAULT_ LOOKASIDE 125×512? That's more RAM then we have on the ESP. Have you any non-trivial Lua application where SQLite is working?

luizfeliperj:
This is 125 x 512 bytes which is 64kb, doesn't the ESP specs RAM as 80Kib?
Anyway, it is really too big, what would be the best configuration for the lookaside?

TerryE:
> what would be the best configuration for the lookaside?

Luiz, if you do a `node.heap()`  on a virgin restarted ESP them you will typically see that you have 40-44Kb total RAM available.  Yes there is more RAM but some of this is used as a ROM bootstrap an ISRs, the SDK and a smaller amount for the core NodeMCU firmware.   You should really get SQLite3 working with a non-trivial application before you raise a PR.  Getting to a clean compile means nothing.

luizfeliperj:
@TerryE Never hit this limit, I did no tests using a ten of kilobytes large database. The library use should be kept trivial otherwise one would blow the time limit for the callback, because "The SDK states that if any tasks run for more than 15 mSec, then services such as WiFi can fail.", according to the NodeMCU docs.

I submitted the PR because others shown interest in it, having a broader audience helps fixing the problems we may face.

TerryE:
@luizfeliperj, I think that you miss my point.  The FS on a ESP8266 can be up to 4Mb, so their may be applications where SQLite3 is relevant. It's just that my initial reaction was that I couldn't see how anyone could get this working for a chip will so little memory. A full relational DB seemed unrealistic.  However, others think differently so I accept this as a consensus.  But this isn't an academic exercise.  If the orginal developer hasn't got it working an a real-life application which needs / uses it and reasonably stress tested it, then all that will happen is that when some other Lua (not C) developer tires to build her or his application using the library then we will get issues raised: "I can't get this to work".  Can you fix it?  And by then you may have lost interest in ESP8266s as a project and we have to respond.

So this has work robustly for a reasonable set of usecases that it might be used in.  Or we can't distribute it.

TerryE:
@luizfeliperj Luìz, I will merge this as it improved the make and does not impact the implementation otherwise. But see my new issue which I will just open

PR #2298, opened 2018-03-13
Ignore lookup errors on sync request unless all lookups fail

### LINKED ISSUE #2296: sntp dns lookups maybe incorrect behavior
### Expected behavior

I'm not sure -- I'm conflicted.

### Actual behavior

If you give a list of dns names to the sntp.sync method, then the call fails if *any* of them cannot be looked up. This seems bad. In my case, I had provided a couple of IP address and a couple of dns names, and the call failed because the internet was not reachable. [We got hit by the big storm in the north-eaast, and I was running on generator power for a couple of days and no internet.] My internal NTP servers continued to operate on generator power. 

### Solutions

1. The current behavior is correct

2. The sync operation should happen if *any* of the dns names or addresses can be converted into an ip address. There are then two sub cases: (a) at least one of the failed lookups should be reported on the error callback; (b) no error should be reported unless *all* of the lookups fails. 

I'm conflicted over which of the two subcases is better.

### NodeMCU version
dev

pjsg:
Fixes #2296.

This was largely driven by my experience of being disconnected from the internet as a result of a severe winter storm. I was unable to sync even though some of the targets were valid and would respond. This change also protects against someone prespecifying some named hosts and then not renewing the domain name. While there is at least one valid address, the sync will happen.

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -386,14 +386,13 @@ static void sntp_dns_found(const char *name, ip_addr_t *ipaddr, void *arg)
   if (ipaddr == NULL)
   {
     sntp_dbg("DNS Fail!\n");
-    task_post_low(tasknumber, (uint32_t) name);
   }
   else
   {
     serverp[server_count] = *ipaddr;
     server_count++;
-    task_post_low(tasknumber, SNTP_DOLOOKUPS_ID);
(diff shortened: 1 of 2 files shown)


TerryE:
Thanks guys.

Issue #2292, opened 2018-03-08
Execution of Lua out of flash memory


TerryE:
### New Feature

This issue supersedes #2068, which contains all of the historic discussion.  My [devLFS](https://github.com/TerryE/nodemcu-firmware/tree/dev-LFS) branch contains a working version of the functionality described in detail in this [LFS whitepaper](https://gist.github.com/TerryE/68da481ce231c35aecdf31a80be190a6).

This is simply an Alpha cut so I have committed an as-is snapshot of my system (that is with development options still enabled) and so it is not ready for incorporation into dev, hence this isn't a PR at this stage.  However, the top level highlights are as follows:

-  You can use `luac.cross` to compile your Lua project into an overlay image on your host machine. Once copied to the ESPs SPIFFS, you can then load this into Flash memory and execute Lua modules directly out of this Lua Flash Store (**LFS**) in just the same way as the Lua Firmware executes from flash memory.  I need to benchmark this a bit better but it looks like you can compile projects of up to _**roughly 5,000 lines of Lua into Flash**_. 

-  You will still be limited by the 48Kb RAM available for data and any (typically in development) Lua modules that you want to load directly from SPIFFS, but this in effect at least doubles the size of Lua application that you can run on the ESP.

-  This patch was a total bitch to debug, so in developing it I have also had to sort out a number of other issues.
   -  I have made some changes to the make system, so that `luac.cross` is now just built as part of the standard make, and it doesn't require Lua to be installed on the host.
   -  `luac.cross`has a couple of new options `-f` builds a LFS image and  `-e` executes a Lua file on the host using the NodeMCU firmware version. (Unlike #2278, this only supports the core, `io` and `os` libraries and but doesn't attempt to support any `app/modules` libraries.  It is really there to do first-pass debugging of the Lua core on the host with a full host debugger support.)
     -  The remote debugger now works properly and is _really_ useful for debugging C code on the ESP.  You are still constrained by ESP H/W support for a single hardware break/watch point being active at any time, but you can now successfully `step`, `next` and `continue` from breakpoints, as well as examining data in your application, as well enabling lua_asserts to throw you into the debugger, even if you haven't started in debug mode..  See #2146 for more details. 

I'll do some more tidy up and improve the whitepaper in the next few days, but the reason for this new issue
is to enable a dialogue with the other committers and developers based on the actual code rather than a concept, so that I can sort some basic feedback stuff before raising the PR.   Enjoy.

[...]

PR #2301, opened 2018-03-17
LFS evaluation version -- second release

TerryE:
See #2292.  

### This is a BIG patch

- [x] I have tested my contribution (see qualification below).

See the discussion in #2292.  This PR:
-  Implements LFS for both integer and float builds. (Though note that this is disabled with the default `user_config.h`).
-  Moves `luac.cross` into the main make and removes the need for a Lua install on the host.
-  Optimises the make so that options libraries such as sqlite3 are only compiled if the coreesponding module is enabled.
-  Fixed the remote GDB function and makes it usable.

I and @pjsg have build and tested this in light use for both float and integer builds.  It probably needs a lot more testing before we switch on LFS as default, and as of this request, it can be configured as a build option.  

I'll be moving my own home automation system over to this in the next few weeks.  The days of 1,000 line+ Lua applications on the ESP8266 have arrived.

I don't recommend merging until the reviews and other contributors have evaluated this patched functionality and performance.


devsaurus:
Linker warning for `app/lua/luac_cross`:
```
.output/host/release/obj/loslib.o: In function `os_tmpname':
loslib.c:(.text+0x28c): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
```

`tmpnam` originates from `luaconf.h` and it appears that `LUA_USE_POSIX` is not defined. The latter would be set by `LUA_USE_LINUX` which is also undef. Is there a host OS detection missing in the make flow?

app/lua/lua.c
```diff
@@ -602,12 +455,11 @@ static bool readline(lua_Load *load){
         {
           /* Get a empty line, then go to get a new line */
           c_puts(load->prmt);
-          continue;
         } else {
           load->done = 1;
           need_dojob = true;
-          break;
```
pjsg: I can't figure out what is going on here....

docs/en/modules/node.md
```diff
@@ -496,6 +496,37 @@ provides more detailed information on the EGC.
...
+Reload the LFS (Lua Flash Store) with the flash image provided. Flash images are generated on the host machine using the `luac.cross`commnad.
+
+#### Syntax
+`node.flash.rebuild(imageName)`
+
+#### Parameters
+`imageName` The of name of a image file in the filesystem to be loaded into the LFS.
+
+#### Returns
+_Not applicable_.  The ESP will load the LFS image and immediately reboot.  Control is not returned to the calling application.
```
pjsg: There are a bunch of error conditions that cause this method to return.

app/Makefile
```diff
@@ -20,93 +20,64 @@ FLAVOR = debug
 ifndef PDIR # {
 GEN_IMAGES= eagle.app.v6.out
 GEN_BINS= eagle.app.v6.bin
-SPECIAL_MKTARGETS=$(APP_MKTARGETS)
+OPT_MKTARGETS      := coap crypto dht http mqtt pcm sjson sqlite3 tsl2561 u8glib ucglib websocket
```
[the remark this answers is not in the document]
devsaurus: Can confirm, linker fails with lots of undefined references to u8g functions when `LUA_USE_MODULES_U8G` is defined.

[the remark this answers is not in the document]
TerryE: @devsaurus Arnim, re the warning: this has been there since eLua days, so this is a case of no change.  The NodeMCU make only works under the Xtensa toolchain on Linux, so this is a global issue with our build.  Maybe we should do an OS detect and abort on anything else, But this isn't an issue _introduced_ by this patch.

@pjsg, Agreed and this needs fixing, so I'll have a look at the Travis build and work out what is going wrong, but the assumption in the normal make is that you enable a given module by setting the corresponding define in `user_modulea.h`.  This Travis make seems to be overriding this by a backdoor and forcing all modules into a link, despite the settings of these defines.  This is a QA artiffact as such a biuld will never be capable of being downloaded onto a real ESP module.  It isn't a real-life issue but one of the way the Travis build is configured.  I just need to mirror this short circuit in the make and detect a Travis build and similarly force the build of all of the associated subdirs..

app/include/user_modules.h
```diff
@@ -65,13 +57,13 @@
 //#define LUA_USE_MODULES_SJSON
 //#define LUA_USE_MODULES_SNTP
 //#define LUA_USE_MODULES_SOMFY
-#define LUA_USE_MODULES_SPI
+//#define LUA_USE_MODULES_SPI
```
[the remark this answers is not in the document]
TerryE: Correct. I should back this out. but the set of settings in the default build needs to be revisited in a separate issue.  E.g. I can't hink of any of the devices that we currently support in our `modules` library that use SPI, so why do we include it by default?  We include MQTT but not json.  I'll make this change to minimize the out of scope side-effects, but this doesn't reduce the issue that that default set needs revisiting.

app/lua/Makefile
```diff
@@ -24,7 +25,8 @@ STD_CFLAGS=-std=gnu11 -Wimplicit
 #   makefile at its root level - these are then overridden
 #   for a subtree within the makefile rooted therein
 #
-#DEFINES += 
+#DEFINES += -DDEVELOPMENT_TOOLS -DDEVELOPMENT_USE_GDB -DNODE_DEBUG -DBREAK_ON_STARTUP_PIN=1
```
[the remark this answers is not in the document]
TerryE: Correct. These shouldn't be defaults, and that is why the char 1 on this line is a "**#**" char,, so the make ignores this line.  However, the most common alternative to the default is going to be uncommenting this line by removing the # char, which is why I have included in in this form.

app/lua/lflash.c
```diff
@@ -0,0 +1,224 @@
...
+static uint32_t curOffset;
+
+#define ALIGN(s)     (((s)+sizeof(size_t)-1) & ((size_t) (- (signed) sizeof(size_t))))
+#define ALIGN_BITS(s) (((uint32_t)s) & (sizeof(size_t)-1))
+#define ALL_SET      cast(uint32_t, -1)
+#define FLASH_SIZE   LUA_FLASH_STORE
+#define FLASH_PAGE_SIZE INTERNAL_FLASH_SECTOR_SIZE
+#define FLASH_PAGES  (FLASH_SIZE/FLASH_PAGE_SIZE)
+
+#define BREAK_ON_STARTUP_PIN  1  // GPIO 5 or setting to 0 will disable pin startup
```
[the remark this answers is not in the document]
TerryE: Yes, Good suggestion,  We need to add this to the sectino on debugging and leave it prefixed by a `//` to comment it out by default.

app/lua/lflash.c
```diff
@@ -0,0 +1,224 @@
...
+    return 0;
+   
+  if (vfs_read(fd, &fh, sizeof(fh)) != sizeof(fh) ||
+      fh.flash_sig != FLASH_SIG)
+    return 0;
+
+  if (vfs_lseek(fd, -1, VFS_SEEK_END) != fh.flash_size-1 ||
+      vfs_lseek(fd, 0, VFS_SEEK_SET) != 0)
+    return 0;
+
```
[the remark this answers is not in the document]
TerryE: Hummm. Excellent suggestion as (as we know :stuck_out_tongue_winking_eye:) the VM will barf if an Integer LFS is loaded into a Float RTL or v.v.  We could set one bit in the Flash ID in word 0 to to 0/1 depending on whether the underlying build is integer.  I'll do this in the tweaks commit.

app/lua/lgc.h
```diff
@@ -79,6 +79,7 @@
 #define VALUEWEAKBIT	4
 #define FIXEDBIT	5
 #define SFIXEDBIT	6
```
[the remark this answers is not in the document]
TerryE: If you look at the other bits, then you will see that bit 3 is also similarly denormalised.  We've only got eight bits in the marked byte, and they've already all been allocated.  Just as with `FIXEDSTACK` , `FINALIZED and `KEYWEAK` markings, this isn't an issue in practice  -- in that only `Proto` and `TString` types can be **LFS** GCobjects and neither are super fixed.

That being said, it took a lot of head scratching and extra `lua_assert`s to make sue that this is true. :smile:

app/lua/lua.c
TerryE: The original logic is -- how should i put it -- tangled.  To be honest I can't remember why I removed the break and instead fell through the the following continue.    

What the original code did was to break on receiving an EOL whereas this continue to receive another char.  I'll try both variants out and see which works the best.

Whatever, we've got whole blocks of code `#if 0`ed out since being hacked out for the first project commit.  IMO, if they don't do anything and will never be reintroduced then they should be removed so we can see the wood for the trees.

docs/en/modules/node.md
TerryE: Quite possibly.  The documentation all needs tightening up as well as adding `lua_examples` directory for sampe LFS init modules.  However, my suuggestion is that we should first reach a consensus on (a) whether this patch has enough merit to be include in dev, and (b) at what point do we swap to it LFS enabled being the default.

These documentation issues (whilst important) are still secondary to these decisions.

app/Makefile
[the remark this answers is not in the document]
TerryE: > Can confirm, linker fails with lots of undefined references to u8g functions when `LUA_USE_MODULES_U8G` is defined.
I'll take a look at this and fix. :)

app/lua/lflash.c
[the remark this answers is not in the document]
pjsg: I was thinking that you could have two different magic header numbers -- one for float and one for integer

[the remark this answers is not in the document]
TerryE: Having one bit flipped gives two different magic numbers, so I think we are basically saying the same thing.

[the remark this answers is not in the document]
TerryE: I also note that this reload function simply returns `nil` if there is an issue (as opposed to not returning if all is OK).  Would is be better to return some error code depending on the error?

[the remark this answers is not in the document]
pjsg: Probably would be good to return an error code....

app/Makefile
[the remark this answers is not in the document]
TerryE: My tweak to the magic works fine but
```
spiffsimg/spiffsimg -f ../bin/0x%x-4mb.img  -S 4mb -U 0x7c2d0 -r ./spiffsimg/spiffs.lst -d
```
is dying with "Not enough space: fatal error"  Why? swapping the **-S 4mb** for ** -c 0x8000** works fine.

app/lua/lflash.c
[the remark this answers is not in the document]
TerryE: The fact that the call returns at all indicates a fatal error, but maybe we should just return a error message if we do return

app/lua/lflash.c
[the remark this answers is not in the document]
TerryE: Whatever, this define doesn't belong in `lflash.c`.

app/Makefile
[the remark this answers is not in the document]
TerryE: OK, I've bottomed this.  The underlying issue is that most of these adjunct directories have the same root name as the module, e.g. `coap` uses `coap`, etc. The two u?g modules don't follow this pattern as `u8g` uses `u8glib` and ditto `ucg`.  I just had to add a couple of extra magic lines to handle this second case.

jmattsson:
`make clean` appears to do some dependency generation for luac_cross? I don't see any actual files, but a directory structure springs into existing under `app/lua/luac_cross/.output`

jmattsson:
```
make[2]: Entering directory '/home/johny/src/DiUS/nodemcu-firmware/tools/spiffsimg'
g -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -I. -I../../app/spiffs -I../../app/include -DNODEMCU_SPIFFS_NO_INCLUDE --include spiffs_typedefs.h -Ddbg_printf=printf main.c ../../app/spiffs/spiffs_cache.c ../../app/spiffs/spiffs_check.c ../../app/spiffs/spiffs_gc.c ../../app/spiffs/spiffs_hydrogen.c ../../app/spiffs/spiffs_nucleus.c  -o spiffsimg
make[2]: g: Command not found
Makefile:8: recipe for target 'spiffsimg' failed
make[2]: [spiffsimg] Error 127 (ignored)
```
Looks like `$(CC)` has been unset by the time spiffsimg is being built

jmattsson:
```
# echo 'print("Hello, Lua!")' > local/fs/init.lua
# make
....
make[1]: *** No rule to make target '../local/lua/*.lua', needed by 'LFSimage'.  Stop.
```

tools/Makefile
```diff
@@ -16,10 +17,22 @@ OBJDUMP = $(or $(shell which objdump),xtensa-lx106-elf-objdump)
 SPIFFSFILES ?= $(patsubst $(FSSOURCE)%,%,$(shell find $(FSSOURCE) -name '*' '!' -name .gitignore ))
 
 #################################################################
-# Get the filesize of /bin/0x10000.bin
+# Get the filesize of /bin/0x10000.bin and SPIFFS sizing
 #
```
jmattsson: On a default build with LFS enabled (set to 64k), the spiffs image gets built as `0x70000-*mb.img`. Yet, the actual file system location is at `0xb0000`.

I think this just reinforces that a partition table would be a very good thing...

app/platform/common.c
```diff
@@ -67,22 +67,54 @@ uint32_t platform_flash_get_num_sectors(void)
 #endif // #ifdef INTERNAL_FLASH_SECTOR_SIZE
 }
 
+static uint32_t allocated = 0;
+static uint32_t phys_flash_used_end = 0;  //Phyiscal address of last byte in last flash used sector 
+
+uint32_t platform_flash_reserve_section( uint32_t regsize, uint32_t *start )
```
jmattsson: Don't take this the wrong way Terry, but I really don't think this is the way to go. We've kept on bolting on various flash allocation bits over the years. I think it's time we bite the bullet and switch to a partition table like on the ESP32. That way we can easily tell where the spiffs image(s?) are, the LFS area, SDK reserved areas, etc.

With the approach here we're effectively introducing race/ordering problems where things may move around unexpectedly in flash. Having a partition table would keep things a lot cleaner and safer.

app/lua/lflash.c
```diff
@@ -0,0 +1,224 @@
...
+#define BYTE_OFFSET(t,f) cast(size_t, &(cast(t *, NULL)->f))
+/*
+ * Rehook address chain to correct Flash byte addressed within the mapped adress space 
+ * Note that on input each 32-bit address field is split into 2×16-bit subfields
+ *  -  the lu_int16 offset of the target address being referenced
+ *  -  the lu_int16 offset of the next address pointer. 
+ */
+
+static int rebuild_core (int fd, uint32_t size, lu_int32 *buf) {
```
jmattsson: I'd love for this to be more easily exposed (so not `static` for starters). The way we'd like to use the LFS feature at $work is to pre-build the entire image and flash (or OTA-upgrade) that, with the actual Lua code not even knowing how to `node.flash.reload()`.

I'd originally envisaged just adding a flag to `luac.cross` to spit out non-PIC addresses, but having seen the format now, I'm thinking it would be easier to write a separate utility that post-processes the `flash.img` from PIC to relocated-ready-to-be-dumped-into-flash format. That will need `rebuild_core` though (possibly with `flashBlock` passed in as a function pointer for flexibility).

app/lua/lflash.c
```diff
@@ -0,0 +1,224 @@
...
+  while (1) {}
+
+  return 0;
+}
+
+
+/*
+ * Return a C closure pointing to the Flash Index function
+ */
+LUAI_FUNC int luaN_index (lua_State *L) {
```
jmattsson: Currently one seems to need to do `node.flash.index()('somemodule')`. Is there any real reason for that much indirection? It would seem much more "normal" to do `node.flash.index('somemodule')`.

Also, I'd much prefer to have the index itself returned as a table (`local t = node.flash.index()`) so it can be read programatically, compared to just printed out (what `node.flash.index()()` currently does).

Am I missing something here that would prevent this simplification?

TerryE: The previous fact that Lua ran from RAM and C ran from flash lead to the practice of "when in doubt, code it in C", even when a direct Lua approach is simpler and more transparent therefore easier to modify for the typical Lua developer.  In this case it would be easier for the Lua application programmer to say `lfs.someroutine(args)` to invoke `someroutine` in the **LFS**. Yes, we could hard core this in C but this is only a couple of lines of Lua which would be in your unit routine in the LFS, so the _only_ cludgy reference would be the first `node.flash.index()'init'()` call and if we allowed the user to define `LUA_INIT_STRING` in the config then with
```Lua
#define LUA_INIT_STRING "pcall(function()node.flash.index()'init'()end)"
```
then they wouldn't even need to do that.

app/lua/lflash.c
TerryE: I think that I mentioned somewhere that `luac.cross` is so fast that it easier just to add a `-a` switch to fix the LFS at a fixed address, and run the `luac.cross` multiple times if necessary.

This still leaves the need to do address calculation in your make.  The alternative is to embed this functionality in a smarter esptool.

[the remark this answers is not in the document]
TerryE: See comment below.  simpler to add `-a` option to `luac.cross` to produce an absolute version.

app/lua/lflash.c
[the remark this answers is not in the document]
TerryE: See answer to repeat of this below: **tl;dr** version: that's what Lua is for :smile:

tools/Makefile
TerryE: My concern with adding a partition table is that on the ESP8266, at least,  it would be just that: a bolt-on addition, and not properly integrated into the flash loader.  So how do you allocate partitions on the ESP8266.  Do we try to adopt the same conventions as on the ESP32, and if so how do we integrate them?

IMO, it would be better have one of two apporaches:
1.  You can define the base and size of both SPIFFS and LFS as config defines.  If you do then
    -  You can fix the absolute address of the LFS using the `-a` option, and ditto with spiffsimg
    -  `user_main.c` should do some bounds checks to make sure that everything fits properly
2.  You don't in which case _both_ fall back to using the platform flash allocator, and in this case there is nothing to stop the make process generate an internal table using the defines at (1) and the allocator uses these if set.   

Incidentally you can set the LFS size to any multiple of 4Kb up to 256Kb.

The botch with SPIFFS is that the format is _not_ PI, (though it could trivially be so).  If it were then the SPIFFS images would also be PI.  This would make life _**so**_ much simpler.

app/platform/common.c
TerryE: See comment above.  This is really an issue it it's own right and maybe worth tracking as such.

app/lua/lflash.c
TerryE: As far as provisioning goes, I would class DiUS and me as advanced users, each with our own sweet spot.

In my case I have my own provisioning system.  I _very_ rarely want to change the base firmware (for example my home automation system is still running a build of dev from last Aug).  But I do use my own provisioning system a _lot_.  This probes a flag byte in RTCmem and if set then the ESP end does a reprovision on reboot.  I've integrated this into my ESP apps as a reprovision command, so each ESP config has its own master directory hierarchy on the host provisioning server.  If I want to make a change then I just change the master file on the server and issue a preprovision command to the ESP app, and bang -- magic happens -- and a few seconds later the ESP is rebooted with the new app running.  I am still testing the LFS changes but this is all pretty transparent to the app -- other than LFS aallows you to code your apps in a simpler cleaner way because you don't have to worry about the code footprint in RAM.

However, all this does rely on you having a network path from each ESP to some provisioning server.  So this isn't a good solution for a lot of embedded IoT uses.

The way that I currently do the bootstrap is to have a `init.lua` stub which either runs the `node.flash.reload()` if LFS isn't loaded or it chains into the init module in the LFS.  From there on in, all encapsulation is done by Lua features.

app/lua/lflash.c
TerryE: With the switch to LFS, I would also like to see a switch to the core dogma:

-   _Only use C if you can't do it in Lua or there are string performance reasons why it must be done in C._

Just think:  change a line of C and you have to reflash the entire firmware.  Change one line of Lua and you (or your provisioning subsystem) only needs to replace one SPIFFS file.  A mega hassle vs. a few seconds.

For example we've got a C ds18b20 module which is slow and far too limited functionality.  My Lua version is both richer and a lot faster.  So why use C?  The old answer was because C ran from flash so took up less RAM footprint.  This argument no longer applies.

As to this LFS init routine, if you do a `-p`  option on the `luac.cross` then you will see that this doesn't use a table which some C routine parses, the code is a directly emitted Lua function.

And as I have said elsewhere is really that hard to include something like the following in the LFS:init module?
```Lua
do
  local ndx = node.flash.index 
  local lfs_t = { __index = function(_, name)
      local fn = ndx(name)
      if type(fn) == 'function' then return fn end -- or return nil implied
    end}
  getfenv().lfs = setmetatable(lfs_t,lfs_t)
end
``` 
The RAM footprint is a table with a single entry in it and a LClosure with a single upval.  The LFS footprint is a code fragment of 16 instructions and a function of 10 instructions.

tools/Makefile
jmattsson: I'd be happy enough to go with your approach, especially in the interest of getting LFS going. As long as the size/location gets set at compile time so we don't have any allocation races/ordering issues at runtime I think we're good.

As for spiffs, last I checked it's position independent - the phys/logic address translation takes place in the read/write accessor routines passed in to spiffs on initialisation. Not sure what makes you think it isn't position independent?

app/lua/lflash.c
jmattsson: If you think a `-a` switch is the better approach, sure, I don't mind.

app/lua/lflash.c
jmattsson: You earlier argued that both mine and your uses were "advanced" (probably rightly so), so I think I'll level that same comment back at you here :)  Meta tables aren't exactly the most obvious things when one starts out with Lua.

I personally was looking at `node.flash.index` as a drop-in replacement for `dofile` (or even `require`). It's something existing developers are familiar with both concept wise and syntax wise, and it matches how the files/modules get interned into the LFS. Where before you would've `mymod = dofile('mymod.lc')` you could (assuming suggested change) `mymod = node.flash.index('mymod')`. To me that would be a nice path into the LFS.

To address the dogma shift, yes, I agree in principle, but we'd need to provide an easy way for users to build the `flash.img`. Right now Marcel's cloudbuilder is saving everyone from needing their own build environment, which is fabulous from a usability perspective. It does make it hard to start using LFS with this though. Say @marcelstoer, do you think it would be feasible to also have an LFS builder service? A user uploads their `.lua` files and we spit out a either a PIC `flash.img` to be `node.flash.reload()`ed or an `lfs.bin` to be `esptool`'d straight onto the device? If I had time and somewhere to host it I could probably put something together (insert bemoaning of lack of time here).

pjsg: This is my `init.lua`

```
if file.rename('flash.img', 'loaded.img') then
  node.flash.reload('loaded.img')
end
node.flash.index()('init')()
```

Then my `init.lua` in the ROM starts with

```
local index = node.flash.index and node.flash.index()
local function loader_flash(module)
  local r = index(module)
  return type(r) == 'function' and r -- or nil otherwise
end
if index then package.loaders[3] = loader_flash end
```

jmattsson: @pjsq Maybe you can explain what I'm missing here? Why do we need to do
```
local index = node.flash.index and node.flash.index()
```
rather than
```
local index = node.flash.index
```
What feature is gained by requiring that function call? Why can't that be hidden inside the `node.flash.index` function itself?

jmattsson:
Something's not right here. Default build of this PR, with LFS enabled (still 64k).

```
$ cat local/lua/wtf.lua 
function wtf(arg)
  print(arg)
  return 1
end
```
Build, flash, `node.flash.reload()` the `flash.img` file, reboot the esp.
```
NodeMCU 2.1.0 build unspecified powered by Lua 5.1.4 on SDK 2.1.0(116b762)
> wtf=node.flash.index()('wtf')
> =wtf('hi')
> =wtf('hi')
hi
1
>
```
Why isn't the first call to my `wtf` function giving me any output or returning a value?

app/lua/lgc.c
```diff
@@ -538,7 +555,7 @@ static void atomic (lua_State *L) {
   size_t udsize;  /* total size of userdata to be finalized */
   /* remark occasional upvalues of (maybe) dead threads */
   remarkupvals(g);
-  /* traverse objects cautch by write barrier and by 'remarkupvals' */
+  /* traverse objects caucht by write barrier and by 'remarkupvals' */
```
jmattsson: I think you mean "caught" :rofl:

app/lua/lobject.h
```diff
@@ -340,13 +203,7 @@ typedef TValuefields TValue;
...
 #define setsvalue2n	setsvalue
 
-#ifndef LUA_PACK_VALUE
-#define setttype(obj, tt) (ttype(obj) = (tt))
-#else // #ifndef LUA_PACK_VALUE
-/* considering it used only in lgc to set LUA_TDEADKEY */
-/* we could define it this way */
-#define setttype(obj, _tt) ( ttype_sig(obj) = add_sig(_tt) )
-#endif // #ifndef LUA_PACK_VALUE
+#define setttype(obj, stt) ((void) (obj)->value, (obj)->tt = (stt))
```
jmattsson: Hmm? Are you trying to silence some compiler warning here or something?

app/lua/lua.c
```diff
@@ -468,11 +324,8 @@ int lua_main (int argc, char **argv) {
 
 void lua_handle_input (bool force)
 {
-  while (gLoad.L && (force || readline (&gLoad)))
-  {
+  if (gLoad.L && (force || readline (&gLoad)))
```
jmattsson: Whoa, this change makes me nervous. I explicitly changed it to a `while` to fix the issue of lost commands on the serial. If more than one line buffers up the LVM would fall behind in its input processing.

app/platform/platform.c
```diff
@@ -879,7 +879,7 @@ uint32_t platform_s_flash_write( const void *from, uint32_t toaddr, uint32_t siz
   if(SPI_FLASH_RESULT_OK == r)
     return size;
   else{
-    NODE_ERR( "ERROR in flash_write: r=%d at %08X\n", ( int )r, ( unsigned )toaddr);
+    NODE_ERR( "ERROR in flash_write: r=%d at %p\n", ( int )r, ( unsigned )toaddr);
```
jmattsson: We actually have %p support in the printf? I thought we didn't (but I could be getting myself confused with one of the other platforms I work on).

app/platform/platform.c
```diff
@@ -928,15 +928,26 @@ int platform_flash_erase_sector( uint32_t sector_id )
...
   bool b1 = (cache_ctrl & CACHE_FLASH_MAPPED1) ? 1 : 0;
-  uint32_t meg = (b1 << 1) | b0;
-  return mapped_addr - INTERNAL_FLASH_MAPPED_ADDRESS + meg * 0x100000;
+  return ((b1 << 1) | b0) * 0x100000;
+}
+ 
+uint32_t platform_flash_mapped2phys (uint32_t mapped_addr)
+{
+ uint32_t meg = flash_map_meg_offset();
+ return (meg&1) ? -1 : mapped_addr - INTERNAL_FLASH_MAPPED_ADDRESS + meg ;
```
jmattsson: `(meg <= 0)` a bit clearer as to what the checks here (and a few lines further down) do?

app/user/user_exceptions.c
```diff
@@ -73,13 +72,15 @@ void load_non_32_wide_handler (struct exception_frame *ef, uint32_t cause)
...
+     * that was set. (This is typically a remote GDB break). If none 
+     * then trigger a system break instead and hang if the break doesn't 
+     * get handled. This is effectively what would happen if the default 
+     * handler was installed. */
     if (load_store_handler) {
       load_store_handler(ef, cause);
+      return;
     }
-    asm ("break 1, 1");
+     asm ("break 1, 1");
```
jmattsson: Surplus whitespace, or intentional?

tools/Makefile
TerryE: > Not sure what makes you think it isn't position independent?

`spiffimg` requires you to specify an absolute location for the image.  I'll meed to have a look at the source to see why it does this, but if SPIFFS is PI then `spiffimg` should be as well.  This would make life easier for everyone.

app/lua/lflash.c
TerryE: @jmattsson the and trick is simple because 
-  if the LFS isn't compiled in then `node.flash` returns `nil` and so evaluating `node.flash.index` will throw an exception, causing a panic.
-  if the LFS is compiled in, but the LFS isn't loaded, then `node.flash.index` returns `nil` and so evaluating `node.flash.index()` will throw an exception, causing a panic.  In the case of running _from LFS_, the `node.flash.index and ` is redundant because this code can assume that LFS is both compiled and loaded, and this makes life a lot simpler.  

My preferred `init.lua`approach is:
```Lua
pcall(function() local f=node.flash; return f.index and f.index()('init')() or f.reload('.img') end)
```
which is guaranteed not to panic. and I do the update reload using my previsioning system. 

I chose the Lua route because (a) it is simple; and (b) it leaves it to the individual developer / project to implement their own Lua encapsulation that's right for them.  This got the patch out there to evaluate.  I don't think that using meta tables is that much of an issue as we can include a default in Lua examples and refer to it in the API reference. Novice Lua programmers can use it as-is and more advanced ones can use it as a starting point to modify.

I will give an analogy: Enduser_setup is written in C, by I regularly see Lua developers post issues "I don't know how to write C modules, but how do I modify it to do XYZ?"  If it was in Lua, then they could just tweak the code.

----
This all being said, changing the API is pretty straight forward, so long as we first agree what we want to change it to.  Let's pick up this general point on #2292

app/lua/lgc.c
TerryE: A bit naff replacing one spelling error by another !!

app/lua/lobject.h
TerryE: There are two `tt` fields, one in the TValue and (in the case of GCObjects) one in the GCO header, and the code should confuse the two.  (Lua 5.3 does this by renaming the latter  to `tt_`.)  The `ttype(o)` and `setttype(o,t)` macros should only be used on the TValue type.  

The `(void) (obj)->value` guard is compiled but discarded during code optimisation, but it still throws a compile error is these are used on the GCOs, and this allowed me to fix a few cases that had been missed in the eLua changes.

The reason that this is important for LFS is that the TValue form is a 32 bit integer and so testing this doesn't throw a `l8ui` exception.  What I intend to do is to replace the relatively small number of GCO `tt` tests by a macro form which generates the extra `extui` instruction and avoids the unaligned exception.

tools/Makefile
jmattsson: As the guy who wrote the initial `spiffsimg` before @pjsg co-opted into NodeMCU: nah, that's not what that argument really says. That's just one way to do the max fs size calculation :)
I readily confess that it should've all been documented better though...

app/platform/platform.c
TerryE: It's fairly recent, but still already there:  #2062

app/lua/lobject.h
jmattsson: Aaaah, that makes sense. Cool hack. Maybe worth leaving a comment along those lines there for future reference? E.g. `// guard against confusion between GCO's and TValue's "tt" field by referencing the "value" member too`

app/lua/lua.c
TerryE: To be honest, I can't remember why I did this.  I will take a look at the history.  My first pass through merged in the changes from a vanilla Lua-5.1 variant and I may just have missed this in the merge.  

But the whole interactive interface is very flaky anyway.  It is really designed for a more POSIX-like environment.  The way it determines compilation units is to stack the lies and try:
-  line 1
- line 1 + line 2
- line 1 + line 2 + line 3
- ...

Until the sequence compiles.  It then executes the brick and treats the next line as line 1.  In the meantime there is no flow control (hardware or XON/XOFF) on the input uart so it is _very_ easy to overflow the 128 byte input FIFO on the UART, with or without this change.

Also  the whole of the interactive interface including the input / output redirector is extremely flaky.   

However all of this is separate to this change so I will back this one out here.

app/user/user_exceptions.c
TerryE: Surplus

tools/Makefile
TerryE: Ah, the penny drops re the difference between the `-c` and `-S`, so yes, it should have been better documented :grin:  In practice we have two use cases for a SPIFFS:
1.  The Lua app makes little use of SPIFFS, so it just needs to be "big enough"  (because of the flashing and startup overhead).  I typically use a 64Kb and work up from that is 64Kb multiples, simply because during development testing, it is often quicker to reflash the SPIFFS than use other ways to update it.
2/  The Lua developer isn't sure how much SPIFFS is going to used, so when in doubt then make it as big as possible.
I suspect that most developers start with (2) but if you reflash your firmware a lot then you end up doing (1).

app/lua/lobject.h
TerryE: In fact I just copied the same hack from somewhere else.  I should really back it out or make it a `//#define` variant because I really only need this when i need to do type checking in macros and I suspect that it does leave the code in in `-O0`.

app/lua/lgc.c
TerryE: @jmattsson, re your previous example, this is a mindfart on your part. :grinning: It does exactly what I would expect it to do, and would have the same result as `wtf=loadfile('wtf.lua')` is this was in the SPIFFS.  Do a `luac.cross -l -o /dev/null local/lus/wtf.lua` to see what this generate
-  The first execution executes the following with replaces the global `wtf` with the closure:
```
	1	[4]	CLOSURE  	0 0	; 0x258b2d0
	2	[1]	SETGLOBAL	0 -1	; wtf
	3	[4]	RETURN   	0 1
```
-  `wtf` now points to the closed function so the second call execute the following giving the response you expected.
```
	1	[2]	GETGLOBAL	1 -1	; print
	2	[2]	MOVE     	2 0
	3	[2]	CALL     	1 2 1
	4	[3]	LOADK    	1 -2	; 1
	5	[3]	RETURN   	1 2
```

jmattsson: Umm.. what? I'm not reassigning wtf between the invocations. I'm assigning the global `wtf` exactly once, then calling it twice. I sure as sundown don't expect my `wtf` variable/function to be reassigned just because I invoked it (given I don't reassign it in the function). Kindly explain my supposed mindfart further.

TerryE: No, you are getting caucht (joke) is the subtleties of Lua.  If you've got 15 mins, then just Skupe me or ping me an email and I'll Skype you.

jmattsson: As Terry promised me: ./facepalm
Okay, the penny dropped:`function wtf() end` is of course the same as `wtf = function() end`, and the initial `wtf` was the file/module closure itself.

app/lua/lflash.c
```diff
@@ -0,0 +1,224 @@
...
+
+#include "c_fcntl.h"
+#include "c_stdio.h"
+#include "c_stdlib.h"
+#include "c_string.h"
+
+/*
+ * Flash memory is a fixed memory addressable block that is serially allocated by the
+ * luac build process and the out image can be downloaded into SPIFSS and loaded into
+ * flash with a node.flash.load() command. See luac_cross/lflashimg.c for the build
```
jmattsson: Nitpick, the command is currently `node.flash.reload()`, and subject to iteration :)

jmattsson:
Here's one way of solving the build-fails-if-no-local/lua-files-exist issue:
```
diff --git i/tools/Makefile w/tools/Makefile
index 608e985..745bda5 100644
--- i/tools/Makefile
+++ w/tools/Makefile
@@ -69,8 +69,13 @@ spiffsscript: remove-image LFSimage spiffsimg/spiffsimg
        $(foreach sz, $(FLASHSIZE), spiffsimg/spiffsimg -f ../bin/0x%x-$(sz).img  $(FLASH_SW) $(sz) -U $(FLASH_FS_LOC) -r ./spiffsimg/spiffs.lst -d; )
        @$(foreach sz, $(FLASHSIZE), if [ -r ../bin/spiffs-$(sz).dat ]; then echo Built $$(cat ../bin/spiffs-$(sz).dat)-$(sz).bin; fi; )
        
-LFSimage: $(LUASOURCE)*.lua
-       ../luac.cross -f -o $(FSSOURCE)flash.img $(LUASOURCE)*.lua
+LFSSOURCES:=$(wildcard $(LUASOURCE)/*.lua)
+ifneq ($(LFSSOURCES),)
+LFSimage: $(LFSSOURCES)
+       ../luac.cross -f -o $(FSSOURCE)flash.img $(LFSSOURCES)
+else
+LFSimage: ;
+endif
 
 remove-image:
        $(foreach sz, $(FLASHSIZE), if [ -r ../bin/spiffs-$(sz).dat ]; then rm -f ../bin/$$(cat ../bin/spiffs-$(sz).dat)-$(sz).bin; fi; )
```

app/lua/lflash.c
TerryE: Yup, but as we discussed these two might be better stripped back.  Whatever ... will pick this up on the sweep.

marcelstoer:
Espressif claims they squeezed [17KB extra memory out of their RTOS SDK 2.0](https://twitter.com/EspressifSystem/status/978879387549528064)...niice. Regardless of what we do here it'd be nice if they gave the NON-OS SDK just as much love.

TerryE:
Yup, but when you go through the details in their documentation, then you see that the way that they do this is to drop the instruction cache from 32K to 16K and use the extra 16K freed by doing this as heap.
Given that most of our execution of from flash and this relies on the instruction cache to get reasonable runtime, _this isn't a patch that I'd want on my ESPs_

app/lua/lflash.c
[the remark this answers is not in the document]
TerryE: This entry needs removing.  

The macro also remained to `DEVELOPMENT_BREAK_ON_STARTUP_PIN` to have a unfied prefix with related macro constants.

app/platform/common.c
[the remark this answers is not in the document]
TerryE: Johny, One step too far on this patch.  Sorry.

app/platform/platform.c
TerryE: The platform API already had the mapping one way.  I just added the inverse map using the same checks.

app/user/user_exceptions.c
TerryE: I'd lose the while(1) {} and replace it by a while  around the break.  But a job for another day.  

The break will instantly reboot the ESP if the debugger exception handler is not installed If is then you will enter the debugger.

TerryE:
So I have been working through the correction list for updates to this patch.  The update list is far larger than the majority of other PRs !! So what I am proposing to do is to rebasline the patch so that incorporates two commits:
-  The first of these is a roll-up of the four initial commits on the Feb PR.
-  The second includes all of the review additions, so that reviewers can focus on this second commit.  

Note that this will also the eLua bugifix and #2333 so that as well as being able to move out all of the program and string data into LFS, the remaining RAM-base TValue-related resources will be 75% of the footprint (if you define ` LUA_PACK_TVALUES_ENABLE` in your build.

@jmattsson, I am surprised that you haven't picked up and commented on this little gem! :smile:

jmattsson:
@TerryE Why? I iz veri bizi... :(
Looks like an easy win (provided there are no gremlins in the corner cases), so go for it.

As for rebaselining the patch, I'd recommend not doing it. Github does a good job of remembering which version comments were applied to, so I think you might end up complicating things while trying to simplify them. Just push more commits here as-is; we can always tidy up if needed.

(And geez, stop taking all the fruit at once - surely we'll want to be able to find these juicy bits in the future too and feel all good about further freeing up memory! ;-P )

marcelstoer:
> The Travis CI build failed

Did you look into that already?

TerryE:
> Did you look into that already?
Is the Pope catholic?  :wink:

TerryE:
Sorry but there were a lot of review comments and so a lot of changes to a large patch, hence a large change.  Here is a summary of these updates.


## Configuration

### `.gitignore`

-  ignores the TLD `luac.cross` and the `local` hierarchy.

### `app/lua/linit.c`

-  This is not a user module and is a core lua component, so has been moved back into
`app/lua. The inclusion of the builtin components is now no longer predicated on the
`LUA_USE_BUILTIN_*` macros, since the the firmware has never been tested with any of
these components removed.  They are in practice mandatory.

-  This now also generate a stripped down set of libraries for cross compiler builds to
facilitate the `luac.cross -e` option.

### app/include/user_modules.h

-  The `LUA_USE_BUILTIN_*` macros have been removed since they are no longer used.

-  The patch no longer changes any of the `LUA_USE_MODULES_*` settings, and specfically
`LUA_USE_MODULES_SPI` is defined by default.

### app/include/user_config.h

-  The extra options for `LUA_FLASH_STORE` and `LUA_COMPACT_TVALUES` have been added.

-  The content has been reordered logically and the inline comments to help developers 
to configure the firmware have been much expanded.

-  An extra define `LUA_INIT_STRING` can be defined and the commented out example 
shows how to use this to load the LFS module `_init` instead of running the SPIFFS 
`init.lua` file.


## Functional

### `.gdbinit` and `.gdbinitlua`
-  The Lua macros need to be accessible for both host `luac.cross` and target `lua`
debugging, so I have split these off into a separate `.gdbinitlua` that can be sourced
from either debug enviroment.

### `app/Makefile`
-  Subdirectory skipping now handles `u8glib` and `ucglib` correctly.

### `app/lua/lflash.c`
-  Validates the LFS signature header to ensure that the image is build with the same
TValue size as the firmware.

-  Skip the position independent address relocation process in the case of absolute 
images.

-  Change to the processing of node.flash.index() routine, `luaN_index()`. This 
previously returned a Lua Closure of the index routine if present.  The routine now 
calls this closure directly and prefixes the return list with the base address and 
length of the LFS in the case of the index listing varains.  See the new Lua example
in `lua_examples/LFS/LFS_init.lua` for its use.

### `app/lua/lobject.h`
-  Add `LUA_PACK_TVALUES` TValue packing.

### `app/lua/lobject.c`
-  Fixed my bug in luaO_log2 (I forgot that this is wrapped in macro which offset by 
1, so my version was 1 out)

### `app/lua/lparser.`
-  Apply eLua correction: remove unneeded lua_malloc/free of small objects now 
allocated on stack.

### `app/lua/ltable.c`
-  Fix eLua bug: tables with `t->nzise == 0` and `t->node != dummynode` were being 
incorrectly resized.

### `app/lua/ltablib.c`
-  Apply eLua correction: misleading indentation in `addfield()`

### `app/lua/luaconf.h`
-  LUA_USE_UNIX is set if LUA_CROSS_COMPILER defined, thus removing the "tmpnam" 
warning in luac_cross make.

### `app/lua/lua.c`
-   Line 327 in lua.c Johny's fix: backout `while` replaced by an `if`.

### `app/lua/luac_cross/Makefile`
-  Added extra condtional to suppress the DEPENDS generation on a clean.  

### `app/lua/luac_cross/luac.c`

-  Add `-a` option to specify an absolute address to create LFS image.  Note that this
option also implies `-f`.

### `app/lua/luac_cross/lflashimg.c`

-  New `absolute` parameter to `dumpToFlashImage()` If this is non-zero, then an 
absolute rather than PIC image is generated.
-  The LFS signature header has six variants for the three sizes of TValue: 
`LUA_INTEGRAL` (8 bytes), `LUA_PACK_TVALUES_ENABLED` (12 bytes) and default float
build (16 bytes), and PI vs absolute formats.
-  Flash copy now handles all three TValue variants (8/12/16 bytes).

### `app/module/node.c`

-  remove extra table level `flash.`

## Documentation

### `docs/en/modules/node.md`

-  Updated documentation of  `node.flashindex()` and `node.flashreload()`

devsaurus:
Terry, I face errors during compilation:
```
make[2]: *** swTimer: No such file or directory.  Stop.
make[2]: Leaving directory 'nodemcu-firmware_devsaurus/app'
../Makefile:272: recipe for target '.subdirs' failed
make[1]: *** [.subdirs] Error 2
make[1]: Leaving directory nodemcu-firmware_devsaurus/app'
```

Somehow #2287 seems to be incompletely applied in your LFS branch.

TerryE:
> Somehow #2287 seems to be incompletely applied in your LFS branch.

Thanks, I'll push a fix in the next 12 hours.

jmattsson:
Ooooh! I'll try to find some time early next week to go through the updates!

tools/Makefile
```diff
@@ -39,13 +63,21 @@ spiffsimg: spiffsimg/spiffsimg
...
+	@echo "" >> ./spiffsimg/spiffs.lst
 	@$(foreach f, $(SPIFFSFILES), echo "import $(FSSOURCE)$(f) $(f)" >> ./spiffsimg/spiffs.lst ;)
-	@$(foreach sz, $(FLASHSIZE), spiffsimg/spiffsimg -U $(FLASH_USED_END) -o ../bin/spiffs-$(sz).dat -f ../bin/0x%x-$(sz).bin -S $(sz) -r ./spiffsimg/spiffs.lst -d; )
+	$(foreach sz, $(FLASHSIZE), spiffsimg/spiffsimg -f ../bin/0x%x-$(sz).img  $(FLASH_SW) $(sz) -U $(FLASH_FS_LOC) -r ./spiffsimg/spiffs.lst -d; )
 	@$(foreach sz, $(FLASHSIZE), if [ -r ../bin/spiffs-$(sz).dat ]; then echo Built $$(cat ../bin/spiffs-$(sz).dat)-$(sz).bin; fi; )
-	
+
+ifneq ($(LFSSOURCES),)
+LFSimage: $(LFSSOURCES)
+       ../luac.cross -f -o $(FSSOURCE)flash.img $(LFSSOURCES)
```
devsaurus: Spaces instead of a TAB. Causes `Makefile:75: *** missing separator.  Stop.`

TerryE: Bloody `gedit` preference.  Already picked this on up.

devsaurus:
I'm getting wdt resets with the latest changes:
```
=node.heap()
43864
> 
----------------------------
flash.img       : 9808 bytes
----------------------------
Total file(s)   : 1
Total size      : 9808 bytes

Total : 30371 bytes
Used  : 10291 bytes
Remain: 20080 bytes

> node.flashreload'flash.img'

 ets Jan  8 2013,rst cause:4, boot mode:(3,6)

wdt reset
load 0x40100000, len 26640, room 16 
tail 0
chksum 0x5f
load 0x3ffe8000, len 2184, room 8 
tail 0
chksum 0xe6
load 0x3ffe8888, len 136, room 8 
tail 0
chksum 0x37
csum 0x37
Formatting file system. Please wait...
```

I had to change from `node.flash.reload` to `node.flashreload` and thought the rest works as before. Am I missing something?

devsaurus:
Ok, I got it back to working condition. The only difference compared to yesterday was to uncomment `SPIFFS_MAX_FILESYSTEM_SIZE` in `user_config.h`. This generates a single bin file for the flash image instead of 3 for different sizes (I flashed the 4mb version back then).

I probably can't comment on the code, but thought to give LFS a try and understand the characteristics from a user's perspective.

### Preparing the build
0. Fixed [`tools/Makefile`](https://github.com/nodemcu/nodemcu-firmware/pull/2301#discussion_r183235349)
1. Copied `ftpserver.lua` from #2345 to `local/lua`
2. Uncommented `#define LUA_FLASH_STORE 0x10000` in `user_config.h`
3. Uncommented and selected SPIFFS location `#define SPIFFS_FIXED_LOCATION        0x80000` in `user_config.h`
4. Uncommented `#define SPIFFS_MAX_FILESYSTEM_SIZE    0x10000` in `user_config.h`
5. `cd tools; make` to generate `local/fs/flash.img` from `local/lua/ftpserver.lua`
6. Built with `make` in firmware root dir

### Flashing the firmware
Pretty straight forward:
```
esptool.py --baud 460800 write_flash -fm qio 0x00000 bin/0x00000.bin 0x10000 bin/0x10000.bin 0x80000 bin/0x80000-0x10000.img
```

### Using LFS
After firmware boot:
```lua
node.flashreload'flash.img'
```
FW restarts after some seconds (upon wdt, looks like intended boot reason).

Configured WIFI to station mode, got IP.

Started the FTP server:
```lua
node.flashindex'ftpserver'().createServer('test','12345')
```
and connected with ftp client. All works fine :+1:

### Some benchmarking
A quick comparison of LFS with the old-fashioned method in terms of available heap size:

| State | Heap LFS | Heap non-LFS |
| --- | --- | --- |
| Got IP | 42104 | 42104 |
| Module loaded | 40648 | 34952 |
| Client connected | 39856 | 34176 |
| `ls` executed | 39656 | 33952 |
| Client disconnected | 40184 | 34488 |
| Some seconds passed | 40408 | 34712 |
| Cycling last 4 steps | 40408 | 34712 |

Note: Module is loaded with `dofile("ftpserver.lua").createServer("test","12345")` for the non-LFS run.

So the LFS method constantly saves ~5.5&nbsp;kByte of heap for this example. Pretty impressive!

docs/en/modules/node.md
```diff
@@ -145,6 +145,36 @@ none
 #### Returns
 flash ID (number)
 
+## node.flashindex()
+
+Returns the function reference for a function in the LFS (Lua Flash Store).
+
+#### Syntax
+`node.flashindex()`
```
jmattsson: `node.flashindex(modulename)`

docs/en/modules/node.md
```diff
@@ -145,6 +145,36 @@ none
...
+
+#### Parameters
+`modulename`  The name of the module to be loaded.  If this is `nil` or invalid then an info list is returned
+
+#### Returns
+-  In the case where the LFS in not loaded, `node.flashindex` evaluates to `nil`, followed by the flash and mapped base addresss of the LFS  
+-  If the LFS is loaded and the function is called with the name of a valid module in the LFS, then the function is returned in the same way the `load()` and the other Lua load functions do.
+-  Otherwise an extended info list is returned: the Unix time of the LFS build, the flash and mapped base addresses of the LFS and its current length, and an array of the valid module names in the LFS. 
+
```
jmattsson: Would love a usage example of both `modulename` == nil and not nil :)

docs/en/modules/node.md
```diff
@@ -145,6 +145,36 @@ none
...
 
+## node.flashindex()
+
+Returns the function reference for a function in the LFS (Lua Flash Store).
+
+#### Syntax
+`node.flashindex()`
+
+#### Parameters
+`modulename`  The name of the module to be loaded.  If this is `nil` or invalid then an info list is returned
```
jmattsson: Would it not be more friendly to raise an error and/or return `nil` if you give it an invalid `modulename`? I don't see the info list as being what anyone would really want if they provide a typo'd module name.

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
+-- First time image boot to discover the confuration
```
jmattsson: configuration

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+NODEMCU=~/nodemcu        # The root of your NodeMCU file hierarchy
+SRC=$NODEMCU/local/lua   # your source directory for your LFS Lua files.
+BIN=$NODEMCU/bin
+ESPTOOL=$NODEMCU/tools/esptool.py
+
+$ESPTOOL --port $USB erase_flash   # Do this is you are having load funnies
+$ESPTOOL --port $USB --baud 460800  write_flash -fm dio 0x00000 \
+  $BIN/0x00000.bin 0x10000 $BIN/0x10000.bin
+#
+# Now restart your module and use whatever your intective tool is to do the above
```
jmattsson: interactive

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+-- 'luac.cross -a' options and the flash adresses of the LFS and SPIFFS.
+--
+--[[  So you would need these commands to image your ESP module:
+USB=/dev/ttyUSB0         # or whatever the device of your USB is
+NODEMCU=~/nodemcu        # The root of your NodeMCU file hierarchy
+SRC=$NODEMCU/local/lua   # your source directory for your LFS Lua files.
+BIN=$NODEMCU/bin
+ESPTOOL=$NODEMCU/tools/esptool.py
+
+$ESPTOOL --port $USB erase_flash   # Do this is you are having load funnies
```
jmattsson: if

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+-- It is a good idea to add an _init.lua module to your LFS and do most of the 
+-- LFS module related initialisaion in this.  This example uses standard Lua
+-- features to simplify the LFS API.  
+--
+-- The first adds a 'LFS' table to _G and uses the __index metamethod to resolve
+-- functions in the LFS, so you can execute the main function of module 'fred'
+-- by doing LFS.fred(params)
+--
+-- The second adds the LFS to the require searchlist so that you can require a
+-- Lua module 'jean' in the LFS by simply doing require "jean".  However not that
```
jmattsson: note

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+  end)
+
+-----------------------------------------------------------------------------------
+--
+-- The debug.getstrings function can be used to get a listing of the RAM (or ROM
+-- if LFS is loaded), as per the following example, so you can do this at the 
+-- interactive prompt or call it as a debug function during a running application. 
+--
+do
+  local a=debug.getstrings'RAM'
```
jmattsson: I think something ate the brackets here too?

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+--
+-- You also want to do autoload the LFS, for example by adding the following:
+--
+if node.flashindex() == nil then 
+  node.flashreload('flash.img') 
+end
+
+tmr.alarm(0, 1000, tmr.ALARM_SINGLE, 
+  function()
+    local fi=node.flashindex; return pcall(fi and fi'_init')
```
jmattsson: Something ate the brackets around the `fi` call?

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+  print ('local preload='..table.concat(a,','))
+end
+
+-----------------------------------------------------------------------------------
+--
+-- File: LFS_dummy_strings.lua
+--
+-- luac.cross -f will generate a ROM string table that includes all strings 
+-- referenced in the loaded modules.  If you want to preload other string constants
+-- hen the trick is to include a dummy module in the LFS.  You never need to call
```
jmattsson: then. less chickens that way.

lua_examples/lfs/lfs_fragments.lua
```diff
@@ -0,0 +1,124 @@
...
+-- running application and append these lines to this file.
+--
+local preload = "?.lc;?.lua", "@init.lua", "_G", "_LOADED", "_LOADLIB", "__add", 
+"__call", "__concat", "__div", "__eq", "__gc", "__index", "__le", "__len", "__lt", 
+"__mod", "__mode", "__mul", "__newindex", "__pow", "__sub", "__tostring", "__unm", 
+"collectgarbage", "cpath", "debug", "file", "file.obj", "file.vol", "flash", 
+"getstrings", "index", "ipairs", "list", "loaded", "loader", "loaders", "loadlib", 
+"module", "net.tcpserver", "net.tcpsocket", "net.udpsocket", "newproxy", "package", 
+"pairs", "path", "preload", "reload", "require", "seeall", "wdclr"
+
```
jmattsson: Okay @TerryE, this file is amazingly good and useful. Can I suggest you mention it in the API docs for `node.flashindex()`?

app/include/user_config.h
```diff
@@ -1,37 +1,206 @@
...
+
+//#define LUA_FLASH_STORE 0x10000
+
+
+// By default Lua executes the file init.lua at start up.  The following
+// define allows you to replace this with an alternative startup.  Warning:
+// you must protect this execution otherwise you will enter a panic loop.
+// The example provided executes the LFS module "_init" at startup or fails
+// through to the interactive prompt.
+// ********* WARNING THIS OPTION ISN'T CURRENTLY WORKING
```
jmattsson: :(

jmattsson:
I'm still getting this:
```
make[2]: Entering directory '/home/johny/src/DiUS/nodemcu-firmware/tools/spiffsimg'
g -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -I. -I../../app/spiffs -I../../app/include -DNODEMCU_SPIFFS_NO_INCLUDE --include spiffs_typedefs.h -Ddbg_printf=printf main.c ../../app/spiffs/spiffs_cache.c ../../app/spiffs/spiffs_check.c ../../app/spiffs/spiffs_gc.c ../../app/spiffs/spiffs_hydrogen.c ../../app/spiffs/spiffs_nucleus.c  -o spiffsimg
make[2]: g: Command not found
Makefile:8: recipe for target 'spiffsimg' failed
```
which is caused by `$(CC)` evaluating to empty when the spiffsimg build happens. This is on a plain `dev` branch with this PR applied, and `user_config.h` having the LFS option enabled.

jmattsson:
Something's not quite right with the finding of the spiffs base address when LFS is enabled:

```
> local _,mapa,fa=node.flashindex(); return ('0x%x, 0x%x, 0x%x'):format(mapa,fa,file.fscfg())
0x40270000, 0x70000, 0x90000
```
i.e. spiffs is at 0x90000 as far as the firmware is concerned. Yet:
```
$ make spiffs-image
make -C tools
make[1]: Entering directory '/home/johny/src/DiUS/nodemcu-firmware/tools'
...
spiffsimg/spiffsimg -f ../bin/0x%x-4mb.img  -S 4mb -U 0x6f8a0 -r ./spiffsimg/spiffs.lst -d;   spiffsimg/spiffsimg -f ../bin/0x%x-32mb.img  -S 32mb -U 0x6f8a0 -r ./spiffsimg/spiffs.lst -d;   spiffsimg/spiffsimg -f ../bin/0x%x-8mb.img  -S 8mb -U 0x6f8a0 -r ./spiffsimg/spiffs.lst -d; 
make[1]: Leaving directory '/home/johny/src/DiUS/nodemcu-firmware/tools'
$ ls bin
0x00000.bin  0x70000-32mb.img  0x70000-8mb.img
0x10000.bin  0x70000-4mb.img 
```
So here we have spiffs images marked to go into the LFS area?

Over in the `mapfile` we see that `_flash_used_end` sits *before* the LFS area, which might be the source of my woes. 
```
 0x000000004026f8a0                _flash_used_end = ABSOLUTE (.)
```
For the fs-location detection in `tools/Makefile` to work, that symbol would need to be shifted forward. Considering what LFS is, that would seem the correct thing?

jmattsson:
Other than the above stuff, this seems to be checking out. Some additional stuff might shake out over time, but I wouldn't let that hold back the merge. @TerryE consider this to have my blessing once the above is addressed. Let's make this happen, this stuff is far too good not to! :)

jmattsson:
@TerryE I've got a proposed fix for the `_flash_used_end` issue over at https://github.com/DiUS/nodemcu-firmware/tree/TerryE-dev-LFS%2Bstatic-reserved (see commit  7a115376eaa7d0c684ed0c727efa0b4431664e3b). It's switching the runtime reservations to compile-time, which allows the offline fs location detection to work as in the past.

TerryE:
I agree that having a mix of run-time and compile-time allocation is entirely unsatisfactory.  Happy to switch to either.  I'll go through your suggestion and incorporate in my next patch.

Some of the LFS changes still need wider testing coverage before we can enable this as a default, but given that we can disable LFS by default for now, I also see no reason not to pull it into dev once I fix the above minor issues.

app/include/user_config.h
TerryE: I do mean to get this working soon, but it was just a Q of priorities.  It's not absolutely essential for the first merge.

lua_examples/lfs/lfs_fragments.lua
TerryE: I am rewriting the LFS paper to make it more orientated to the Lua app developer.  I have covered this in there.  My intention is to hoist this paper or some derivative into the formal documentation set.

lua_examples/lfs/lfs_fragments.lua
TerryE: This is actually standard Lua "sugar" syntax: if a function has a single array or string parameter then the brackets are optional hence `fi'_init'`, `fi('_init')` and even `fi(('_init'))` all generate the same code.

docs/en/modules/node.md
TerryE: This follows the same convention as the Lua `pcall()` and in practice simplifies Lua coding.  I don't really see users calling this directly, but rather the `_init` stub in LFS will set up the environment, e.g.
```Lua
do
  local index = node.flashindex 
  local lfs_t = {
    __index = function(_, name)
        local fn_ut, ba, ma, size, modules = index(name)
        if not ba then 
          return fn_ut
        elseif name == '_time' then
          return fn_ut
        elseif name == '_config' then
          local fs_ma, fs_size = file.fscfg()
          return {ba, ma, fs_ma, size, fs_size}
        elseif name == '_list' then
          return modules
        else
          return nil
        end
            end,
    __newindex = function(_, name, value)
        error("LFS is readonly. Invalid write to LFS." .. name, 2)
      end
  }
  rawset(getfenv(),'LFS', setmetatable(lfs_t,lfs_t))

  local function loader_flash(module)
    local fn, ba = index(module)
    return ba and "Module not in LFS" or fn 
  end
  package.loaders[3] = loader_flash

end
```
now `LFS.somemodule(x,y,z)` just works and the LFS properties `_time`, `_config` and `_list` access the other bits that you need.

I'll explain this example in the LFS paper.

TerryE:
Have a read of my draft [LFS white paper](https://gist.github.com/TerryE/68da481ce231c35aecdf31a80be190a6) that I suggest we hoist into the documentation set.

app/include/user_config.h
```diff
@@ -1,37 +1,206 @@
 #ifndef __USER_CONFIG_H__
 #define __USER_CONFIG_H__
```
marcelstoer: With all the relevant documentation now inside this file could you maybe add a note with a link to it to the first paragraph at https://nodemcu.readthedocs.io/en/latest/en/build/#build-options?

TerryE: Will do :smile:

TerryE:
I've just found a bug when porting my latest telnet version to LFS. (It works fine on the latest master.) This LFS version fails on `node.input()` handling.  Any node.input() causes the readline processor to go into a loop; only terminating on WD timeout.  Entering this minimal command at the interactive prompt shows this:

```Lua
node.task.post(function()node.input "='hello'"; end)
```

Sorry guys.

TerryE:
I've just found a bug when porting my latest telnet version to LFS. (It works fine on the latest master.) This LFS version fails on `node.input()` handling.  Any node.input() causes the readline processor to go into a loop; only terminating on WD timeout.  Entering this minimal command at the interactive prompt shows this:

```Lua
node.task.post(function()node.input "='hello'"; end)
```

Sorry guys.

**PS**. Now fixed.

TerryE:
As some of you might have noticed, I may seem to have been side-tracked working on getting LFS-optimised versions of my provisioning system, together with the telnet and ftp servers.  But there was some rationale to this: I wanted to hammer LFS myself in a real development context before pushing the "final" PR to `dev`.  It has been 100% solid.  Ditto the use of 12 byte TValues, so I suggest we we also move to 12 byte TValues on this PR as default for floating point builds.

Another observation: @jmattsson was absolutely spot-on in requesting an "absolute" option for LFS images.  I use a script file to rebuild my LFS and reimage it directly to a test ESP module.  The whole cycle takes less than 10 seconds, so I find it easier just to develop directly in LFS and not even bother with moving in-test modules onto SPIFFS, because it is simpler and faster.  I do use the relocatable version but only for production luaOTA applications.

jmd13391:
Is there a rough ETA for LFS to land in DEV?

TerryE:
I've got about a day's work and it's next on my TODO list. 😚

TerryE:
Sorry for the delay guys. I am trying to track down a weird bug to do with the new method of SPIFFS and LFS inter-operating though linker-based static allocation. 
 -  If LFS is disabled, then everything is OK.  
 -  If LFS is enabled but with a zero sized LFS store then ditto.
 -  If LF has any store allocated then something else (?SPIFFS?) is stomping of the top end of flash (The LFS code is never called).

Lot's of head scratching.  Just waiting for that "oh shit, that's it" moment.

jmattsson:
@TerryE forget to move the `flash_used_end` symbol in the linker file perhaps?

Try running a delta against the branch I linked above some time ago. I had both LFS & SPIFFS going happily with static allocation.

TerryE:
I will have my own palm slap when I find it. Nothing so obvious. It's a sod. If I write then read verify the flash, then it's OK. But if I run the firmware and read it gets stomped. I've added a temp #ifdef code so that I can set the LFS to 0 len and that works but 64K and it runs the same exec path (basically a no-op if the flash signature isn't correct) and the stomp occurs. And very early. I'll find it today.

TerryE:
OK, definitely a head scratch here.  

-  I am using 32mbit (4Mb) Wemos parts for my test harness.  If I explicitly initialise the Flash header to 4Mb size using the `-fs 32m` option then everything seems to work fine.  

-  If I don't then a flash sector near the end of my image gets erased, stomping some code.  I've put in some diagnostics and this occurs before `nodemcu_init()` is posted.  More investigation needed.  Sorry guys.

jmattsson:
...if you don't, then the SDK uses whatever is in there to look for its config data, and if there's a bad checksum writes new config blocks there (last three sectors, as determined by the flash header). We used to catch an incorrect flash size byte before we got into the SDK to avoid just such a problem, so maybe that's gotten broken somehow? You haven't accidentally stopped the early init stuff from being called, have you? (e.g. by dropping the custom entry point setting)
This is the `user_start_trampoline()`, normally set as the entrypoint by `ld/nodemcu.ld`.

TerryE:
@jmattsson Johny, I've just noticed that we cross posted, having come to the same conclusion.  See my PS to my previous comment.  But thanks for the input :)

jmattsson:
Right, so we need to move the fixing-up of the size-byte into the trampoline too to ensure we have everything in order before the SDK can run? Not happy about the extra code in IRAM in that case, but maybe it's necessary.

TerryE:
OK, having thought about this one, this isn't an LFS issue, but more one that larger images such as LFS builds can excite.  Therefore I have raised #2407 to keep track of this but I am assuming that any LFS users will avoid this issue by the steps listed in #2407.

TerryE:
OK, take it apart guys.  Please note #2407.  The main changes are the incorporation on all of the review feedback, the merge with dev to resolve parallel changes to `node.c`, including:

-  Johny's static allocation of the LFS region.
-  `LUA_INIT_STRING` now works but be careful to avoid an infinite error loop, so start by tweaking the example supplied in `user_config.h`
-  `lua.c`exposed the internal of the `lua_Load` structure and this was messily manipulated in `node.c` and `coap.c`.  This is now hidden and an access routine `lua_put_line()` is provided to provide encapsulated access.

Also note that because the LFS region is inside the 0x10000.bin file, you can't flash the bin file in the same `esptool.py` command as this has region overlap detection.  The workaround is a variant of this:

```bash
# Pick the LFS region address from the ELF image
export LFS_BASE=$(xtensa-lx106-elf-objdump -t ../app/.output/eagle/debug/image/eagle.app.v6.out |grep flash_region_base| cut -b 4-8)
# Compile the LFS sources into the correct absolute  bin LFS image 
luac.cross -a 0x402$LFS_BASE -o $BIN/lfs-0x$LFS_BASE.img $LFS_SOURCE/*.lua
# Flash the firmware if needed.  Note the no_reset flag
esptool --port /dev/ttyUSB0 --baud 460800 --after no_reset write_flash -fm dio  0x00000 $BIN/0x00000.bin 0x10000 $BIN/0x10000.bin
# Now flash the LFS (and optionally a new SPIFFS)
esptool --port /dev/ttyUSB0 --baud 460800 write_flash -fm dio 0x$LFS_BASE  $BIN/lfs-0x$LFS_BASE.img
```
This takes about 7sec to reimage the ESP firmware and LFS with the new Espressif `esptool.py`

TerryE:
@cwrseck @devsaurus @djphoenix @dnc40085 @dtran123 @drawkula @georeb 
@HHHartmann @jmattsson @joysfera @marcelstoer @NicolSpies @nwf @pjsg 

You have all commented on / reviewed this LFS development in the past.  I would very much like to merge this PR into `dev`, but I need at least one independent reviewer to take it for a ride to make sure that I haven't made any serious cockups in this latest set of changes following the previous feedback.  So can I request that if you can find the time, then can you please try it out and post back here.  Thanks.

We still need to add cloud-builder support, but even so it is going to be a lot more useful having the base PR in `dev` rather than my fork.

dnc40085:
Aside from the error [`Makefile:75: *** missing separator.  Stop.`](https://github.com/nodemcu/nodemcu-firmware/pull/2301/files#r183235349) the current version builds and runs my application without issue. :thumbsup:

jmattsson:
I was hoping to have had a chance to test out the latest at work this week, but it's looking increasingly unlikely that I'll find time. I was pretty happy overall with it at the last pass, and I've had a quick look through the latest commits and don't see anything that makes be go "eep!" :)
Don't let my absence hold this back any longer, my feeling is that it's good enough to get into dev by now. We can always continue to fix & tweak as necessary.

devsaurus:
I'll try to test drive the latest commits tonight.

devsaurus:
Looks good :+1:

PR #2307, opened 2018-03-19
Add MQTT module for esp32

### LINKED ISSUE #1946: mqtt module for esp32?
Just wonder when can we have mqtt feature for esp32? very looking forwards to this.

goughy:
Fixes #1946.

This pull request adds support for the mqtt protocol and is based on https://github.com/tuanpmt/espmqtt as suggested in issue #1946.  This implementation simplifies mqtt support over the AWS IOT library implementation and has tried to maintain compatibility with the existing NodeMCU MQTT module.  Where the underlying library does not provide support (eg. table-based unsubscribe function), it has been noted in the doc.

Secure (TLS) MQTT support has not been tested (in progress), but there is no reason it should not work out of the box.

### DIFF
--- a/components/modules/mqtt.c
+++ b/components/modules/mqtt.c
@@ -1,0 +1,626 @@
+// Module for interfacing with an MQTT broker
+
+#include "module.h"
+#include "lauxlib.h"
+#include "platform.h"
+#include "task/task.h"
+
+#include "mqtt.h"
+#include <string.h>
+
+task_handle_t hConn;
+task_handle_t hOff;
+task_handle_t hPub;
+task_handle_t hSub;
+task_handle_t hData;
+
+//used as a holder to copy received data from the MQTT task
+//to the main Lua task.  Due to the async nature of the tasks
+//we must copy the data to deliver it intact to the Lua
+//callback
+typedef struct lmqtt_ctx
+{
+	mqtt_client * client;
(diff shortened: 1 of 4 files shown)


components/modules/mqtt.c
```diff
@@ -0,0 +1,717 @@
...
+	}
+
+	if( lua_isnumber(L, 5) )
+	{
+		mqtt_cfg->disable_clean_session = (luaL_checknumber( L, 5 ) == 0);
+		n++;
+	}
+	lua_pop( L, n ); //remove parameters
+
+	lua_newtable( L );
```
devsaurus: I suggest to build the client's metatable not programmatically but from a rotable, e.g. like in the [`ws2812 module`](https://github.com/nodemcu/nodemcu-firmware/blob/dev-esp32/components/modules/ws2812.c#L500).

docs/en/modules/mqtt.md
```diff
@@ -0,0 +1,266 @@
+# MQTT Module
```
devsaurus: `mkdocs.yml` is missing a reference to this file.

components/modules/mqtt.c
goughy: @devsaurus sure thing.  I will look into it.

goughy:
Thanks for the review.  I will try and find some time in the next few days to modify and test and update the PR.

Cheers,

PR #2316, opened 2018-03-25
DS18b20 fix decimals for negative temperatures

### LINKED ISSUE #2313: DS18b20 integer readout: decimals wrong for negative temperature
While calibrating DS18b20 using ice-water, I noticed readouts flipping from 0 to 0.9 degrees C. 

Upon checking /app/modules/ds18b20.c I see this calculation for the decimals:

ds18b20_device_scratchpad_temp_dec = ((double)(ds18b20_device_scratchpad[0] & 0x0F) / 16 * 1000);

Which does not take into account the sign bit(s) from scratchpad[1] so negative decimals are calculated as positive

ref datasheet:
https://datasheets.maximintegrated.com/en/ds/DS18B20.pdf page 6

I'll see if I find time to come up with a better function

peturdainn:
Fixes #2313 by changing the way temp and temp_dec are calculated, taking care not to lose the sign of the decimals part

Checked positive and negative temperatures with integer and floating point builds

### DIFF
--- a/app/modules/ds18b20.c
+++ b/app/modules/ds18b20.c
@@ -192,6 +192,7 @@ static int ds18b20_lua_read(lua_State *L) {
 
 static int ds18b20_read_device(uint8_t *ds18b20_device_rom) {
 	lua_State *L = lua_getstate();
+	int16_t ds18b20_raw_temp;
 
 	if (onewire_crc8(ds18b20_device_rom,7) == ds18b20_device_rom[7]) {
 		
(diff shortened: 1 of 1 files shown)


marcelstoer:
Sorry, as per [our guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#working-with-git-and-github) we only accept PRs against `dev` -> closing

PR #2318, opened 2018-03-25
Dev wifi/wifi_eventmon malloc/free update

### LINKED ISSUE #2303: Fixing modules wthich don't use the Lua memory allocator
The following modules include bare malloc / free API calls: `coap`, `crypto`, `enduser_setup`, `file`, `http`, `mqtt`, `net`, `pcm`, `rotary`, `sntp`, `sqlite3`, `tls`, `tm1829`, `u8g_glue`, `websocket`, `wifi`, `wifi_eventmon`, `wifi_monitor`, `ws2812`. 

The lua allocator should _always_ be used in Lua modules. The reason for this is that the Lua allocator is a thin wrapper on top of the standard allocator, but handles out of memory conditions differently.  If this occurs, the Lua system runs the GC and then retries the allocation. It only fails is there there is not enough memory at this point.  Also instead of returning a null pointer, a Lua error is thrown.
   
### Expected behavior

Libraries should not raise a memory allocation error if  the EGC mode `ON_ALLOC_FAILURE` is set and enough memory is available after GC to satisfy the allocation.

### Actual behavior

The application can crash with an out of memory error even though there is enough heap available to satisfay the request. 

### NodeMCU version
`master` and `dev`

dnc40085:
Partial-fix for #2303 .

This PR is intended to remedy the issue of the wifi module(wifi.c) and wifi_eventmon(wifi_eventmon.c) not using the Lua memory allocator.

### DIFF
--- a/app/modules/wifi_eventmon.c
+++ b/app/modules/wifi_eventmon.c
@@ -59,7 +59,7 @@ static sint32_t event_queue_ref = LUA_NOREF;
 
 static void wifi_event_monitor_handle_event_cb(System_Event_t *evt)
 {
-  EVENT_DBG("\n\twifi_event_monitor_handle_event_cb is called\n");
+  EVENT_DBG("was called (Event:%d)", evt->event);
 
 #ifdef LUA_USE_MODULES_WIFI_MONITOR
   if (hook_fn && hook_fn(evt)) {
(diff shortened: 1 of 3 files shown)


app/modules/wifi_eventmon.c
```diff
@@ -79,38 +74,67 @@ static void wifi_event_monitor_handle_event_cb(System_Event_t *evt)
...
     }
-    c_memcpy(temp->evt, evt, sizeof(System_Event_t)); //copy event data to new struct
+    lua_rawgeti(L, LUA_REGISTRYINDEX, event_queue_ref);
 
-    if(wifi_event_queue_head == NULL && wifi_event_queue_tail == NULL)// if queue is empty add item to queue
-    {
-      wifi_event_queue_head = wifi_event_queue_tail = temp;
-      EVENT_DBG("\n\tqueue empty, adding event and posting task\n");
+    System_Event_t* evt_tmp = lua_newuserdata(L, sizeof(System_Event_t));
+    c_memset(evt_tmp, 0, sizeof(System_Event_t));
```
pjsg: this line is really needed as the entire datastriucture is overwritten in the next line

dnc40085: >this line is really needed as the entire datastriucture is overwritten in the next line
I assume you meant that it really NOT needed, if so then I can remove it, I added it because I was having a problem due to unreferencing the userdata before processing the event

PR #2319, opened 2018-03-26
Tie in the EGC with the SDK's heap knowledge.

jmattsson:
Added `node.egc.meminfo()` to expose LVM usage (to make the regular
`node.egc.ON_MEM_LIMIT` option usable).

Extended the `node.egc.ON_MEM_LIMIT` option to also take negative limits,
in which case that's taken as a request to keep a certain amount of heap
available for non-Lua use.

This came up in the discussion with Terry last week as something we want in mainline. Since we already have this here at $work it seemed worthwhile to submit it. I have no doubt that Terry with his knowledge of the GC might come up with a more elegant approach. Until such a time however, this patch provides a couple of useful features with minimal changes. Using the ON_MEM_LIMIT option with a negative limit avoids a lot of GC overhead and at worst degenerates to the same performance as ALWAYS.

Marcel, the docs phrasing could be improved - I wrote that for internal consumption here at $work, so probably not as clear & friendly as it could be.

### DIFF
--- a/app/lua/lauxlib.c
+++ b/app/lua/lauxlib.c
@@ -14,6 +14,7 @@
 #include C_HEADER_STRING
 #ifndef LUA_CROSS_COMPILER
 #include "vfs.h"
+#include "user_interface.h"
 #else
 #endif
 
@@ -791,6 +792,11 @@ static void *l_alloc (void *ud, void *ptr, size_t osize, size_t nsize) {
   }
   if (L != NULL && (mode & EGC_ALWAYS)) /* always collect memory if requested */
     luaC_fullgc(L);
+#ifndef LUA_CROSS_COMPILER
+  if (L != NULL && (mode & EGC_ON_MEM_LIMIT) && G(L)->memlimit < 0 &&
+      (system_get_free_heap_size() < (-G(L)->memlimit)))
+    luaC_fullgc(L);
+#endif
   if(nsize > osize && L != NULL) {
 #if defined(LUA_STRESS_EMERGENCY_GC)
     luaC_fullgc(L);
--- a/app/lua/lstate.h
+++ b/app/lua/lstate.h
@@ -82,7 +82,7 @@ typedef struct global_State {
   Mbuffer buff;  /* temporary buffer for string concatentation */
   lu_mem GCthreshold;
   lu_mem totalbytes;  /* number of bytes currently allocated */
-  lu_mem memlimit;  /* maximum number of bytes that can be allocated, 0 = no limit. */
+  l_mem memlimit;  /* maximum number of bytes that can be allocated, 0 = no limit. <0 used with EGC_ON_MEM_LIMIT when free heap falls below -memlimit */
   lu_mem estimate;  /* an estimate of number of bytes actually in use */
   lu_mem gcdept;  /* how much GC is `behind schedule' */
   int gcpause;  /* size of pause between successive GCs */
--- a/app/lua/legc.c
+++ b/app/lua/legc.c
@@ -4,7 +4,7 @@
 #include "lstate.h"
 #include "c_types.h"
 
-void legc_set_mode(lua_State *L, int mode, unsigned limit) {
+void legc_set_mode(lua_State *L, int mode, int limit) {
    global_State *g = G(L); 
    
    g->egcmode = mode;
--- a/app/lua/legc.h
+++ b/app/lua/legc.h
@@ -11,7 +11,7 @@
 #define EGC_ON_MEM_LIMIT      2   // run EGC when an upper memory limit is hit
 #define EGC_ALWAYS            4   // always run EGC before an allocation
 
-void legc_set_mode(lua_State *L, int mode, unsigned limit);
+void legc_set_mode(lua_State *L, int mode, int limit);
 
 #endif
 
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -462,14 +462,23 @@ static int node_stripdebug (lua_State *L) {
 // See legc.h and lecg.c.
 static int node_egc_setmode(lua_State* L) {
   unsigned mode  = luaL_checkinteger(L, 1);
-  unsigned limit = luaL_optinteger (L, 2, 0);
+  int limit = luaL_optinteger (L, 2, 0);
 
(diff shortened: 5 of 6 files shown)


TerryE:
Johnny this is one aspect of the tweak that we discussed and thanks for this.  Of the three modes
-  `ALWAYS` is the most robust but causes your ESP to run like a dog.
-  `ON_ALLOC_FAILURE` is a leave it to the last minute option -- which would be OK if the libraries all used the Lua allocator or at least called the Lua GC before allocating memory.  But some don't and as a result it can't be safely used with some libraries.
-  `ON_MEM_LIMIT` has two failings, the first of which is addressed by this version of the patch, which is that the parameter refers to the memory used rather than the memory remaining, and you fix this here.  Can I suggest that we also change the limit from a hard one to a soft one?  That is at the moment, the GC only runs if the available heap crosses a threshold but then fails if it ca't free up across the threshold.  I see that a more robust mode would be switch into `ALWAYS` mode until we cross the threshold.

So by way of example the GC doesn't bother to collect garbage until we only have 4kb remaining, say.  Normally, this will free up lots of memory so the collector wil again back off for a period.  _However_, if the free is still less than 4K it continues to collect on every allocation event until the free level falls to below 4K or we get a hard memory exhaustion.   

There's another reason for the soft fail: because of the three-colour marking system, it takes two sweeps to guarantee recovering storage.

jmattsson:
@TerryE I honestly don't know how to do a "soft fail" here, that is beyond my ken. You're welcome to push amendments here, but I'll have to revoke my "thoroughly tested" guarantee on those parts =)

app/lua/lstate.h
```diff
@@ -82,7 +82,7 @@ typedef struct global_State {
   Mbuffer buff;  /* temporary buffer for string concatentation */
   lu_mem GCthreshold;
   lu_mem totalbytes;  /* number of bytes currently allocated */
-  lu_mem memlimit;  /* maximum number of bytes that can be allocated, 0 = no limit. */
+  l_mem memlimit;  /* maximum number of bytes that can be allocated, 0 = no limit. <0 used with EGC_ON_MEM_LIMIT when free heap falls below -memlimit */
```
TerryE: Johny, you can't overload the g->memlimit this way without going through all of the places in the core which use it.  For example, it is also used in `lapi.c` and `lvm.c`

app/lua/lauxlib.c
```diff
@@ -791,6 +792,11 @@ static void *l_alloc (void *ud, void *ptr, size_t osize, size_t nsize) {
   }
   if (L != NULL && (mode & EGC_ALWAYS)) /* always collect memory if requested */
     luaC_fullgc(L);
+#ifndef LUA_CROSS_COMPILER
+  if (L != NULL && (mode & EGC_ON_MEM_LIMIT) && G(L)->memlimit < 0 &&
+      (system_get_free_heap_size() < (-G(L)->memlimit)))
```
TerryE: I was going to ask expensive in terms of clock cycles is the `system_get_free_heap_size()` until I realised that I should answer this myself since I've got a remote debug session onto a running firmware.  The call is a wrapper around a getter in `libmain.a(mem_manager.o)` and the value itself is maintained by the memory manager in the 0th word of its `.bss`, so th tl;dr is that the call it dirt cheap.

TerryE:
The reason for my diatribe
> ...  Of the three modes ...

is that I've been thinking about this whole issue of the GC an who to configure it for optimum use.  I know the core and have done some playing, and understand a lot of it.  For most non-IoT Lua developers aggressive memory collection is just not an issue so the default trickle settings are fine.  However the default almost paranoid way that we run with might have been understandable in the 15Kb RAM avialable but is just crazy with current builds and even more so with LFS.  a full GC is a pretty expensive operation, so its should only be done as a matter of last resort.  If you have an app that runs in say 35K RAM  so you have ~10K freeboard, then the standard GC with sensible settings will do a pretty good job of keeping collection under control with minimal runtime cost. This all being said, I am happy with this patch so long as you cover off the uses of `G(L)->memlimit` in `lvm.c` and `lapi.c`.

app/lua/lstate.h
jmattsson: Through the wonders of int -> unsigned conversion, I can :)

In `lapi.c` the only use is through `collectgarbage("setmemlimit", newlimit)`. I did not change it because I see no reason to. Using a negative memlimit is an extension available through `node.egc.setmode()` _only_. Setting a regular *hard* memlimit via `collectgarbage()` is still possible (and if you want, you could force negative values through it, but no sane value on the ESP8266 would interfere).

In `lvm.c` I left it alone because a negative memlimit expressed as an unsigned is effectively the same as `MAX_SIZET` on our tiny architecture.

TerryE: @jmattsson Johny, IMO, you aren't fully correct in your assertion.  `G(L)->memlimit` is a single global `uint` variable that was introduced in the `eLua` patch but is used in more than just the EGC threshold calcs. 
-  In `lvm.c:luaV_concat()`, the maximum string length allowed is the lesser of `MAX_SIZET` and  `G(L)->memlimit`.  this this is negative then we will be checking the string length against a negative value and always throw a **"string length overflow"** error. This logic needs fixing or dropping.
-  the GC `setmemlimit` options is an eLua/EGC addition and is an alternative API for setting this ECG variable.  `memlimit`isn;t used by the regular GC, only the EGC add-on.

jmattsson: However, the negative `G(L)->memlimit` gets assigned into an `unsigned`, at which point it becomes a very large positive number, which on the ESP8266 is way above the available memory that the result is effectively the same as `MAX_SIZET`. Try it for yourself if you don't believe me :)

TerryE: OK, I agree. the size of a signed vs unsigned comparison or coerced to an unsigned comparison, but looking at this code I really have to wonder what the rationale for the EGC patch change was.  Why include a bounds check to ensure that the size of a concat string is < MAXUINT - a tiny bit?

However, that is your problem, so I accept your comment.

TerryE:
So this patch now has a clean bill of health from me, but I think that we want to leave merging it until after the 2.2 drop.

PR #2320, opened 2018-03-26
Fix Lua stack corruption problem

### LINKED ISSUE #2311: mysterious crash "already freed"
### Expected behavior
Lua program does not crash because a "table has been freed" mysteriously.

### Actual behavior
Lua program reports "3fff5168 already freed" (the address is arbitrary, of course) and crashes right afterwards when code tries to use a value of a table that has been freed already.

### Test code
``Provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) which will reproduce the problem.``
I know what you need but I cannot provide it, unfortunately. My program started crashing only recently when it grown to a certain size. Even though it's modularized I am afraid that the crash is related to the total size.

Some examples that do crash:
```Lua
local names = {}
dofile("names.lc").load(names)
local name = names[1]
```
The ```names``` is somehow freed right after the ```dofile()``` call so the subsequent ```names[1]``` access crashes. However, if I try to print the ```names``` address (to see if it matches the address reported before the crash) it no longer crashes and the code continues to run perfectly well!
```Lua
local names = {}
dofile("names.lc").load(names)
print(names)
local name = names[1]
```

Yes, one single ```print()``` added to the "right" place makes the crash go away.

Another example of the same issue:
```Lua
local ok, rules = pcall(sjson.decode, data)
local x = rules[1]
```
Again crashes because the ```rules``` has been freed. But if I insert any ```print()``` there it starts working correctly:
```Lua
local ok, rules = pcall(sjson.decode, data)
print("Please do not crash!")
local x = rules[1]
```

I must say this makes developing with NodeMCU kinda nightmare-ish.

BTW, if I replace the ```print("Please do not crash")``` with ```print(node.heap())``` it also "fixes" the issue and shows that there's about 23 kB free on the heap.

Any idea why a Lua table gets freed at unexpected places for no apparent reason? And why inserting a random ```print()``` fixes it?

FWIW, replacing the "right" ```print()``` with ```collectgarbage()``` makes the crash disappear as well.

### NodeMCU version
Current master.

### Hardware
Stock ESP8266-07.

TerryE:
Fixes #2311

-  [x] This PR is for the `dev` branch rather than for `master`.
-  [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well.
-  [x] I have thoroughly tested my contribution.

No documentation changes are required.  The rational and testing are discussed in the issue and referenced gist.

### DIFF
--- a/app/lua/ldo.c
+++ b/app/lua/ldo.c
@@ -144,8 +144,11 @@ static void correctstack (lua_State *L, TValue *oldstack) {
 void luaD_reallocstack (lua_State *L, int newsize) {
   TValue *oldstack = L->stack;
   int realsize = newsize + 1 + EXTRA_STACK;
+  int block_status = is_block_gc(L);
   lua_assert(L->stack_last - L->stack == L->stacksize - EXTRA_STACK - 1);
+  set_block_gc(L);       /* The GC MUST be blocked during stack reallocaiton */
   luaM_reallocvector(L, L->stack, L->stacksize, realsize, TValue);
+  if (!block_status) unset_block_gc(L);  /* Honour the previous block status */
   L->stacksize = realsize;
   L->stack_last = L->stack+newsize;
   correctstack(L, oldstack);


TerryE:
@marcelstoer, I feel that this is a patch that we should get in to the new 2.2 master so it is worth pulling into dev as soon as poss.

On the wider discussion point on whether there have been any changes to eLua since our fork that we should bring across, I did a git diff spanning 2015 to now and there have been minor tweaks, but no material bugfixes  What I will do is to roll them into my LFS patch.

marcelstoer:
> @marcelstoer, I feel that this is a patch that we should get in to the new 2.2 master so it is worth pulling into dev as soon as poss.

Sure, feel free...nobody needs my approval for this.

TerryE:
> Sure, feel free...nobody needs my approval for this.
Not so much approval as my respecting your role as the conductor of the process, so I just wanted to make sure that you were comfortable with this. :smile:

marcelstoer:
Thanks, I am...

Everybody who merges PRs can help "the process" by following https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md#for-collaborators, assigning labels to PRs and closing issues a PR fixes (did that now).

PR #2322, opened 2018-03-26
fix for ds18b20 negative decimals

### LINKED ISSUE #2313: DS18b20 integer readout: decimals wrong for negative temperature

peturdainn:
Fixes ds18b20 negative decimals bug by changing the way temp and temp_dec are calculated, taking care not to lose the sign of the decimals part

Fixes #2313

ds18b20 decimals do not take into account the sign bit. Since the original calculation was not so readable, rewritten in readable way that also fixes the bug. Same code as PR against master.

### DIFF
--- a/app/modules/ds18b20.c
+++ b/app/modules/ds18b20.c
@@ -192,6 +192,7 @@ static int ds18b20_lua_read(lua_State *L) {
 
 static int ds18b20_read_device(uint8_t *ds18b20_device_rom) {
 	lua_State *L = lua_getstate();
+	int16_t ds18b20_raw_temp;
 
 	if (onewire_crc8(ds18b20_device_rom,7) == ds18b20_device_rom[7]) {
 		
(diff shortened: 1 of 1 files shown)


marcelstoer:
@fetchbot you're the original author of this module, could you please verify this fix?

TerryE:
I've just taken a look at this code and some of the logic here is plain bizarre. I am talking about the original module, though @peturdainn's mods don't help. 

-   First, I would have expected some conditional logic based on `LUA_NUMBER_INTEGRAL`.  The CB should  only return a single value for floating point builds.
-   In the case of integer build, yes it might make sense to return an integral part and a decimal fraction part, but what these are is not well defined, and this confusion lies at the heart of this issue / PR. 
-  whether or not, the build is integral, internally within the module the calculations should be done in floating point. 

So taking an example what should a reading of - 1.4°C return on integer builds?
-   The number of decimal places isn't explicitly defined or documented. Should this be API configurable, or fixed at 3?  Let's assume 3 for the following.
-   - 2, 600 - valid and the maths convention but not easy to use for Lua programmers.
-   - 1, 400 - definitely not because 1000*(-1)+400 is the **wrong** answer.
-   - 1, - 400 - this is the one that makes the most sense to me. This is also easy to use in printing so long as you use `maths.abs() ` on the decimal component.

This PR doesn't do this so :-1:  from me.

peturdainn:
I don't see how **1, -400** is a good representation for -1.4

The patch makes it return **-1, -400** so you can do something like
t * 10 + tdec / 100
to generate a value in 0.1 units

before the patch it would return **-1, 600**

TerryE:
@peturdainn,  

> I don't have the time right now to go fix the other problems in this module (just because I offered this fix)

That's life, I guess, but what you fix is doing is to fix only a part of the problem, the other parts being the lack of complete integer vs float supprt and the confusion in the documentation.  

Is it worth a PR so solve 33% of the uses.  My feeling is **-½**, but I will defer to the other committers. @pjsg what are your thoughts, since you are tracking this?

peturdainn:
The doc will need edit anyway, without the fix the user should be warned that negative temps are wrong and particularly bad around 0 when trying to calibrate sensors with ice water, temps between 0 and -1 are reported positive.

I understand you want the code perfect, but it pushes away incremental patch contributors. I'm a big fan of Pieter Hintjens in that regard.

marcelstoer:
> Is it worth a PR so solve 33% of the related issues...I will defer to the other committers.

My answer is clear; I'm quite pragmatic. Any PR that solves a problem, or even a fragment thereof, is worth merging if 

a) it is in line with our guidelines
b) we can be reasonably certain that it doesn't introduce new problems

83% complete is better than 82%, isn't it? I see no reason why we would want to miss out on that 1%. Any developer who devotes time and energy to get from 82 to 83 does a lot more for our projects and all the others who couldn't bother.

peturdainn:
Thanks Marcel, you're actually motivating me to work on this in a later time-frame </ot>

TerryE:
As I said I will defer to the other committers, but once committed I will then raise a new issue to cover the other outstanding items so that they are logged and not forgotten.

marcelstoer:
> Thanks Marcel, you're actually motivating me to work on this in a later time-frame

I can't deny this being part of the motivation behind my comment 😜 You're welcome around here anytime.

TerryE:
@marcelstoer the fix does fix the bug. I can confirm this. The issue isn't of approval, but of policy: should we accept PR which makes a module less flawed rather then fixes all know related issues?

I suggest that we decouple the policy discussion from this point issue.

PR #2334, opened 2018-03-31
Put each ICACHE_RAM_ATTR function in its own section.

### LINKED ISSUE #2304: sigma_delta code is always included even when not needed
### Expected behavior
The sigma_delta code should only be included if needed

### Actual behavior
It always occupies space in the iram

### Root cause
`app/platform/platform.c` (which is always required) consumes IRAM for sigma_delta. Those few functions should be split out into their own C file.

In my environment, this allows the GDBSTUB to be included and the modules that I need.

pjsg:
Fixes #2304.

This implements putting each ICACHE_RAM_ATTR function in its own section. This allows the linker to throw away more functions and hence use less ram.

I checked the map and it looks good. The image boots and seems to work.

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -52,9 +52,9 @@ extern void luaL_assertfail(const char *file, int line, const char *message);
 
 #define ICACHE_STORE_TYPEDEF_ATTR __attribute__((aligned(4),packed))
 #define ICACHE_STORE_ATTR __attribute__((aligned(4)))
-#define STRINGIZE(x) STRINGIZE2(x)
-#define STRINGIZE2(x) #x
-#define ICACHE_RAM_ATTR     __attribute__((section(".iram0.text." __FILE__ "." STRINGIZE(__LINE__))))
+#define ICACHE_RAM_STRING(x) ICACHE_RAM_STRING2(x)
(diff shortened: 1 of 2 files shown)


app/include/user_config.h
```diff
@@ -52,7 +52,9 @@ extern void luaL_assertfail(const char *file, int line, const char *message);
 
 #define ICACHE_STORE_TYPEDEF_ATTR __attribute__((aligned(4),packed))
 #define ICACHE_STORE_ATTR __attribute__((aligned(4)))
-#define ICACHE_RAM_ATTR __attribute__((section(".iram0.text")))
+#define STRINGIZE(x) STRINGIZE2(x)
```
TerryE: Defines in `user_confgi.h` in practice end up getting included in most modules and platform files, so should these be prefixed to remove the risk of a collision with another use of `STRINGIZE`, e.g.
-  `STRINGIZE`->`ICACHE_STRINGIZE`
-  `STRINGIZE2`->`ICACHE_STRINGIZE2`

pjsg:
It turns out that the hw_timer is included by default as it is needed by the gpio.serout functionality that is part of the default build....

I fixed the macro names...

PR #2345, opened 2018-04-05
Create ftpserver.lua

### LINKED ISSUE #2343: Simple FTP server
### Missing feature
Simple FTP server

### Justification
Some times need upload files withuot rs232 connection. FTP is best solution.

### Workarounds
Just use lua for FTP Server
Lua example:
```lua
wifi.setmode(wifi.SOFTAP)
wifi.ap.config({ssid="test",pwd="12345678"})
-- a simple ftp server
USER = "test"
PASS = "12345"
local file,net,pairs,print,string,table = file,net,pairs,print,string,table
data_fnc = nil
ftp_data = net.createServer(net.TCP, 180)
ftp_data:listen(20, function (s)
  if data_fnc then data_fnc(s) end
end)
ftp_srv = net.createServer(net.TCP, 180)
ftp_srv:listen(21, function(socket)
  local s = 0
  local cwd = "/"
  local buf = ""
  local t = 0
  socket:on("receive", function(c, d)
    a = {}
    for i in string.gmatch(d, "([^ \r\n]+)") do
      table.insert(a,i)
    end
    if(a[1] == nil or a[1] == "")then return end
    if(s == 0 and a[1] == "USER")then
      if(a[2] ~= USER)then
        return c:send("530 user not found\r\n")
      end
      s = 1
      return c:send("331 OK. Password required\r\n")
    end
    if(s == 1 and a[1] == "PASS")then
      if(a[2] ~= PASS)then
        return c:send("530 \r\n")
      end
      s = 2
      return c:send("230 OK.\r\n")
    end
    if(a[1] == "CDUP")then
      return c:send("250 OK. Current directory is "..cwd.."\r\n")
    end
    if(a[1] == "CWD")then
      if(a[2] == ".")then
        return c:send("257 \""..cwd.."\" is your current directory\r\n")
      end
      cwd = a[2]
      return c:send("250 OK. Current directory is "..cwd.."\r\n")
    end
    if(a[1] == "PWD")then
      return c:send("257 \""..cwd.."\" is your current directory\r\n")
    end
    if(a[1] == "TYPE")then
      if(a[2] == "A")then
        t = 0
        return c:send("200 TYPE is now ASII\r\n")
      end
      if(a[2] == "I")then
        t = 1
        return c:send("200 TYPE is now 8-bit binary\r\n")
      end
      return c:send("504 Unknown TYPE")
    end
    if(a[1] == "MODE")then
      if(a[2] ~= "S")then
        return c:send("504 Only S(tream) is suported\r\n")
      end
      return c:send("200 S OK\r\n")
    end
    if(a[1] == "PASV")then
      local _,ip = socket:getaddr()
      local _,_,i1,i2,i3,i4 = string.find(ip,"(%d+).(%d+).(%d+).(%d+)")
      return c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..",0,20).\r\n")
    end
    if(a[1] == "LIST" or a[1] == "NLST")then
      c:send("150 Accepted data connection\r\n")
      data_fnc = function(sd)
        local l = file.list();
        for k,v in pairs(l) do
          if(a[1] == "NLST")then
            sd:send(k.."\r\n")
          else
            sd:send("+r,s"..v..",\t"..k.."\r\n")
          end
        end
        sd:close()
        data_fnc = nil
        c:send("226 Transfer complete.\r\n")
      end
      return
    end
    if(a[1] == "RETR")then
      f = file.open(a[2]:gsub("%/",""),"r")
      if(f == nil)then
        return c:send("550 File "..a[2].." not found\r\n")
      end
      c:send("150 Accepted data connection\r\n")
      data_fnc = function(sd)
        local b=f:read(1024)
        sd:on("sent", function(cd)
          if b then
            sd:send(b)
            b=f:read(1024)
          else
            sd:close()
            f:close()
            data_fnc = nil
            c:send("226 Transfer complete.\r\n")
          end
        end)
        if b then
          sd:send(b)
          b=f:read(1024)
        end
      end
      return
    end
    if(a[1] == "STOR")then
      f = file.open(a[2]:gsub("%/",""),"w")
      if(f == nil)then
        return c:send("451 Can't open/create "..a[2].."\r\n")
      end
      c:send("150 Accepted data connection\r\n")
      data_fnc = function(sd)
        sd:on("receive", function(cd, dd)
          f:write(dd)
        end)
        socket:on("disconnection", function(c)
          f:close()
          data_fnc = nil
        end)
        c:send("226 Transfer complete.\r\n")
      end
      return
    end
    if(a[1] == "RNFR")then
      buf = a[2]
      return c:send("350 RNFR accepted\r\n")
    end
    if(a[1] == "RNTO" and buf ~= "")then
      file.rename(buf, a[2])
      buf = ""
      return c:send("250 File renamed\r\n")
    end
    if(a[1] == "DELE")then
      if(a[2] == nil or a[2] == "")then
        return c:send("501 No file name\r\n")
      end
      file.remove(a[2]:gsub("%/",""))
      return c:send("250 Deleted "..a[2].."\r\n")
    end
    if(a[1] == "NOOP")then
      return c:send("200 OK\r\n")
    end
    if(a[1] == "QUIT")then
      return c:send("221 Goodbye\r\n",function (s) s:close() end)
    end
    c:send("500 Unknown error\r\n")
  end)
  socket:send("220--- Welcome to FTP for ESP8266/ESP32 ---\r\n220---   By NeiroN   ---\r\n220 --   Version 1.4   --\r\n");
end)
```

NeiroNx:
Fixes #2343 .

Some times need upload files withuot rs232 connection. FTP is best solution.

### DIFF
--- a/lua_modules/ftp/README.md
+++ b/lua_modules/ftp/README.md
@@ -1,0 +1,64 @@
+# FTPServer Module
+
+This is a Lua module for the access to SPIFFS via FTP protocol. 
+FTP server uses only one specified user and password to auth clients.
+All clients need authorization - anonumous do not supported yet.
+All files have RW access.
+Directory creation do not supported because SPIFFS do not have that.
+
+## Require
+```lua
+ftpserver = require("ftpserver")
+```
+## Release
+```lua
+ftpserver:closeServer()
+ftpserver = nil
+package.loaded["ftpserver"]=nil
+```
+
+# Methods
+
+## createServer()
+Starts listen on 20 and 21 ports for serve FTP clients. 
+
+The module requires active network connection.
+
+#### Syntax
+`createServer(username, password)`
+
+#### Parameters
+- `username` string username for 'USER username' ftp command.
+- `password` string password.
+
+#### Returns
+nil
+
+#### Example
+```lua
+require("ftpserver").createServer("test","12345")
+```
+
+## closeServer()
+CLoses all server sockets.
+
+#### Syntax
+`closeServer()`
+
+#### Returns
+nil
+
+#### Example
+```lua
+ftpserver = require("ftpserver")
+ftpserver:createServer("test","12345")
+-------------------------
+-- Some program code
+-------------------------
+if needStopFtp = true then
+  ftpserver:closeServer()
+  ftpserver = nil
+  package.loaded["ftpserver"] = nil
+end
+```
+
--- a/lua_modules/ftp/README.md
+++ b/lua_modules/ftp/README.md
@@ -1,8 +1,8 @@
 # FTPServer Module
 
-This is a Lua module for the access to SPIFFS via FTP protocol. 
-FTP server uses only one specified user and password to auth clients.
-All clients need authorization - anonumous do not supported yet.
+This is a Lua module to access the SPIFFS file system via FTP protocol. 
+FTP server uses only one specified user and password to authenticate clients.
+All clients need authentication - anonymous access is not supported yet.
 All files have RW access.
 Directory creation do not supported because SPIFFS do not have that.
 
@@ -40,7 +40,7 @@ require("ftpserver").createServer("test","12345")
 ```
 
 ## closeServer()
-CLoses all server sockets.
+Closes all server sockets.
 
 #### Syntax
 `closeServer()`
--- a/lua_modules/ftp/README.md
+++ b/lua_modules/ftp/README.md
@@ -1,6 +1,6 @@
 # FTPServer Module
 
-This is a Lua module to access the SPIFFS file system via FTP protocol. 
+This is a Lua module to access the [SPIFFS file system](../../docs/en/spiffs.md) via FTP protocol. 
 FTP server uses only one specified user and password to authenticate clients.
 All clients need authentication - anonymous access is not supported yet.
 All files have RW access.
--- a/lua_modules/ftp/ftpserver.lua
+++ b/lua_modules/ftp/ftpserver.lua
@@ -67,7 +67,7 @@ ftp_srv:listen(21, function(socket)
       return c:send("200 S OK\r\n")
     end
     if a1 == "PASV" then
-      local _,ip = socket:getaddr()
+      local _,ip = c:getaddr()
       local _,_,i1,i2,i3,i4 = string.find(ip,"(%d+).(%d+).(%d+).(%d+)")
       return c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..",0,20).\r\n")
     end
@@ -85,7 +85,7 @@ ftp_srv:listen(21, function(socket)
       return
     end
     if a1 == "RETR" then
-      f = file.open(a2:gsub("%/",""),"r")
+      local f = file.open(a2:gsub("%/",""),"r")
       if f == nil then
         return c:send("550 File "..a2.." not found\r\n")
       end
@@ -94,23 +94,23 @@ ftp_srv:listen(21, function(socket)
         sd:on("sent", function(cd)
           b=f:read(1024)
           if b then
-            sd:send(b)
+            cd:send(b)
             b=nil
           else
-            sd:close()
+            cd:close()
             f:close()
             data_fnc = nil
             c:send("226 Transfer complete.\r\n")
           end
         end)
         local b=f:read(1024)
-        sd:send(b)
+        cd:send(b)
         b=nil
       end
       return
     end
     if a1 == "STOR" then
-      f = file.open(a2:gsub("%/",""),"w")
+      local f = file.open(a2:gsub("%/",""),"w")
       if f == nil then
         return c:send("451 Can't open/create "..a2.."\r\n")
       end
--- a/lua_modules/ftp/ftpserver.lua
+++ b/lua_modules/ftp/ftpserver.lua
@@ -5,7 +5,7 @@ do
 local createServer = function (user, pass)
 local data_fnc, data_sock = nil, nil
 ftp_data = net.createServer(net.TCP, 180)
-ftp_data:listen(20, function (s) if data_fnc then data_fnc(s) else data_sock = s end end)
+ftp_data:listen(1024, function (s) if data_fnc then data_fnc(s) else data_sock = s end end)
 ftp_srv = net.createServer(net.TCP, 180)
 ftp_srv:listen(21, function(socket)
   local s = 0
@@ -18,6 +18,7 @@ ftp_srv:listen(21, function(socket)
       table.insert(a,i)
     end
     local a1,a2 = unpack(a)
+    print(d)
     if a1 == nil or a1 == "" then return end
     if s == 0 and a1 == "USER" then
       if a2 ~= user then
@@ -36,6 +37,9 @@ ftp_srv:listen(21, function(socket)
     if s ~= 2 then
       return c:send("530 Not logged in, authorization required\r\n")
     end
+    if a1 == "SYST" then
+      return c:send("250 UNIX Type: L8\r\n")
+    end
     if a1 == "CDUP" then
       return c:send("250 OK. Current directory is "..cwd.."\r\n")
     end
@@ -43,8 +47,11 @@ ftp_srv:listen(21, function(socket)
       if a2 == "." then
         return c:send("257 \""..cwd.."\" is your current directory\r\n")
       end
-      cwd = a2
-      return c:send("250 OK. Current directory is "..cwd.."\r\n")
+      if a2 == "/" then
+        cwd = a2
+        return c:send("250 OK. Current directory is "..cwd.."\r\n")
+      end
+      return c:send("550 Failed to change directory.\r\n")
     end
     if a1 == "PWD" then
       return c:send("257 \""..cwd.."\" is your current directory\r\n")
@@ -69,14 +76,14 @@ ftp_srv:listen(21, function(socket)
     if a1 == "PASV" then
       local _,ip = c:getaddr()
       local _,_,i1,i2,i3,i4 = string.find(ip,"(%d+).(%d+).(%d+).(%d+)")
-      return c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..",0,20).\r\n")
+      return c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..",4,0).\r\n")
     end
     if a1 == "LIST" or a1 == "NLST" then
       c:send("150 Accepted data connection\r\n")
       data_fnc = function(sd)
         local l = file.list();
         for k,v in pairs(l) do
-          sd:send( a1 == "NLST" and k.."\r\n" or "+r,s"..v..",\t"..k.."\r\n")
+          sd:send( a1 == "NLST" and k.."\r\n" or "-rwxrwxrwx 1 esp esp "..v.." Jan  1  2018 "..k.."\r\n")
         end
         sd:close()
         data_fnc = nil
@@ -151,6 +158,10 @@ ftp_srv:listen(21, function(socket)
       end
       file.remove(a2:gsub("%/",""))
       return c:send("250 Deleted "..a2.."\r\n")
+    end
+    if a1 == "SIZE" then
+      local f = file.stat(a2)
+      return c:send("213 "..(f and f.size or 1).."\r\n")
     end
     if a1 == "NOOP" then
       return c:send("200 OK\r\n")
--- a/lua_modules/ftp/ftpserver.lua
+++ b/lua_modules/ftp/ftpserver.lua
@@ -91,11 +91,11 @@ ftp_srv:listen(21, function(socket)
       end
       c:send("150 Accepted data connection\r\n")
       data_fnc = function(sd)
-        local b=f:read(1024)
         sd:on("sent", function(cd)
+          b=f:read(1024)
           if b then
             sd:send(b)
-            b=f:read(1024)
+            b=nil
           else
             sd:close()
             f:close()
@@ -103,10 +103,9 @@ ftp_srv:listen(21, function(socket)
             c:send("226 Transfer complete.\r\n")
           end
         end)
-        if b then
-          sd:send(b)
-          b=f:read(1024)
-        end
+        local b=f:read(1024)
+        sd:send(b)
+        b=nil
       end
       return
     end
--- a/lua_modules/ftp/ftpserver.lua
+++ b/lua_modules/ftp/ftpserver.lua
@@ -1,12 +1,11 @@
 -- a simple ftp server
-USER = "test"
-PASS = "12345"
 local file,net,pairs,print,string,table = file,net,pairs,print,string,table
-data_fnc = nil
+local ftp, ftp_data, ftp_srv
+do
+local createServer = function (user, pass)
+local data_fnc = nil
 ftp_data = net.createServer(net.TCP, 180)
-ftp_data:listen(20, function (s)
-  if data_fnc then data_fnc(s) end
-end)
+ftp_data:listen(20, function (s) if data_fnc then data_fnc(s) end end)
 ftp_srv = net.createServer(net.TCP, 180)
 ftp_srv:listen(21, function(socket)
   local s = 0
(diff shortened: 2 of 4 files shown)


lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
+-- a simple ftp server
+USER = "test"
+PASS = "12345"
```
TerryE: Keep these in locals, and they will be accesses upvals in nested closures.  See my FAQ for moy explanation:
```Lua
local USER, PASS = "test", "12345"
```

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+local file,net,pairs,print,string,table = file,net,pairs,print,string,table
+data_fnc = nil
```
TerryE:
```Lua
local data_fnc = function() end
```
and then the listener becomes `ftp_data:listen(20, function (s) data_fnc(s) end)`  You still need the extra closure because you overwrite data_fnc/

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+  if data_fnc then data_fnc(s) end
+end)
+ftp_srv = net.createServer(net.TCP, 180)
+ftp_srv:listen(21, function(socket)
+  local s = 0
+  local cwd = "/"
+  local buf = ""
+  local t = 0
+  socket:on("receive", function(c, d)
+    a = {}
```
TerryE: Should be a local.

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+    for i in string.gmatch(d, "([^ \r\n]+)") do
+      table.insert(a,i)
+    end
+    if(a[1] == nil or a[1] == "")then return end
```
TerryE:
```Lua
local a1,a2=unpack(a)
```
and use `a1` and `a2` below instead of `a[1]` and  `a[2]`; faster and generates less code.

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+    if(s == 0 and a[1] == "USER")then
```
TerryE: You know that the Lua syntax is `if cond then ...`.  This isn't C or Javascript :smiley:. IMO  `if a2 ~= USER then`is clearer, but even if you want to keep the () or even ((())) then add the spaces to make it readable.

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+      return c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..",0,20).\r\n")
+    end
+    if(a[1] == "LIST" or a[1] == "NLST")then
+      c:send("150 Accepted data connection\r\n")
+      data_fnc = function(sd)
+        local l = file.list();
+        for k,v in pairs(l) do
+          if(a[1] == "NLST")then
+            sd:send(k.."\r\n")
+          else
```
TerryE: As I said previously multiple sends in a single CB is a no-no. So you should do something like 
```Lua
  local resp={}
  for k,v in pairs(file.list()) do
    resp[#resp+1] = (a[1] == "NLST") and 
                      (k.."\r\n") or
                      ("+r,s"..v..",\t"..k.."\r\n")
  end
  multi_send(resp)
```
Where `multi_send(resp)` is a separate routine declared above to do the multi-send. I'll discuss this separately.   Note the `sd` used here is an upval which wll create a resource leak.

TerryE:
@NeiroNx, Read my [FAQ](https://nodemcu.readthedocs.io/en/dev/en/lua-developer-faq/) and particularly how to use `debug.getregistry()` to get a handle on registered upvals which can cause resource leaks.  Using upvals is in general good practice, but there are bearpits because these can create circular references through the Lua Registry that the LGC can't clear down unless you break the circular links.  

Re the `mutil_resp()` routine, there are examples around.  I can give you an example or you might prefer to do this yourself.  Its a pity that Lua doesn't have a `table.spilt()` , but in general the sum of the lines will be <1Kb so you can just do a `table.concat()` to form a response, _but_ if it is longer than 1024 then you want to split the string into the first bit (<1Kb) and the rest and send the first bit with an on("sent") CB to send the next 1Kb chunk, etc.

This non-blocking I/O can be a real PITA sometimes. :smile:

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+      f = file.open(a[2]:gsub("%/",""),"w")
+      if(f == nil)then
+        return c:send("451 Can't open/create "..a[2].."\r\n")
+      end
+      c:send("150 Accepted data connection\r\n")
+      data_fnc = function(sd)
+        sd:on("receive", function(cd, dd)
+          f:write(dd)
+        end)
+        socket:on("disconnection", function(c)
```
devsaurus: Should be `sd:on` instead of `socket:on`?

lua_modules/ftp/server.lua
NeiroNx: and owerwrite with blank function after STOR,RETR is comlete?

lua_modules/ftp/server.lua
NeiroNx: Yes - wery more JS and Cpp make me use ```If()``` automaticly

lua_modules/ftp/server.lua
```diff
@@ -0,0 +1,157 @@
...
+        return c:send("550 File "..a[2].." not found\r\n")
+      end
+      c:send("150 Accepted data connection\r\n")
+      data_fnc = function(sd)
+        local b=f:read(1024)
+        sd:on("sent", function(cd)
+          if b then
+            sd:send(b)
+            b=f:read(1024)
+          else
```
HHHartmann: Great module. This was on my wish list for some time.

Reading the next chunk of the file right before sending it and setting the buffer to nil right afterwards would save some memory.
```
            b=f:read(1024)
            sd:send(b)
            b=nil
```
The logic would have to be adapted also.

eku:
May I ask why ftp and not tftp?

NeiroNx:
TFTP:
1. is difficult to use from phone or tablet. 
2. have no directory listing.
3. do not have authorization.
4. ...

drawkula:
On some OSes you can mount FTP servers and many file managers can access FTP semi-transparently without mounting, so this will be a nice way to handle code on your favourite ESP-pets.

lua_modules/ftp/README.md
```diff
@@ -0,0 +1,64 @@
+# FTPServer Module
+
+This is a Lua module to access the SPIFFS file system via FTP protocol. 
+FTP server uses only one specified user and password to authenticate clients.
+All clients need authentication - anonymous access is not supported yet.
+All files have RW access.
+Directory creation do not supported because SPIFFS do not have that.
```
marcelstoer: While this is technically correct you might want to mention that you can mimic directories  by prefixing file names.

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,163 @@
+-- a simple ftp server
+local file,net,pairs,print,string,table = file,net,pairs,print,string,table
+local ftp, ftp_data, ftp_srv
+do
+local createServer = function (user, pass)
+local data_fnc = nil
+ftp_data = net.createServer(net.TCP, 180)
+ftp_data:listen(20, function (s) if data_fnc then data_fnc(s) end end)
```
devsaurus: Proposal to handle "early connect" scenario:
```lua
if data_fnc then data_fnc(s) else data_sock = s end
```

With a `local data_sock = nil` above.

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,163 @@
...
+      data_fnc = function(sd)
+        local l = file.list();
+        for k,v in pairs(l) do
+          sd:send( a1 == "NLST" and k.."\r\n" or "+r,s"..v..",\t"..k.."\r\n")
+        end
+        sd:close()
+        data_fnc = nil
+        c:send("226 Transfer complete.\r\n")
+      end
+      return
```
devsaurus: Proposal for "early connect" scenario:
```lua
if data_sock then
  node.task.post(function() data_fnc(data_sock); data_sock = nil end)
end
return
```
This starts `data_fnc` manually after the return.

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,163 @@
...
+            f:close()
+            data_fnc = nil
+            c:send("226 Transfer complete.\r\n")
+          end
+        end)
+        local b=f:read(1024)
+        sd:send(b)
+        b=nil
+      end
+      return
```
devsaurus: Also post `data_fnc` here for "early connect".

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,163 @@
...
+        sd:on("receive", function(cd, dd)
+          f:write(dd)
+        end)
+        sd:on("disconnection", function(c)
+          f:close()
+          data_fnc = nil
+        end)
+        c:send("226 Transfer complete.\r\n")
+      end
+      return
```
devsaurus: Also post `data_fnc` here for "early connect".

devsaurus:
Works very well in my environment after the latest changes - thanks a lot @NeiroNx for your contribution and dedication!

FrankX0:
The current implementation does not work in my environment (Windows and FileZilla).
This is because the ports used by this implementation are fixed to 20 and 21.
However, when the ftp client sends a PASV command, the server should respond with the actual port used by the client, and start listening on that port. So I changed your code from using a fixed port for the data connection (21) to a dynamic version, as dictated by the client:
~~~lua
    if a1 == "PASV" then
      local _,ip = c:getaddr()
      local _,_,i1,i2,i3,i4 = string.find(ip,"(%d+).(%d+).(%d+).(%d+)")
      return c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..",0,20).\r\n")
    end
~~~
Changed into:
~~~lua
        if(a[1] == "PASV")then
            local port, _ = c:getpeer()
            local _,_,i1,i2,i3,i4 = string.find(IP,"(%d+).(%d+).(%d+).(%d+)")
            c:send("227 Entering Passive Mode ("..i1..","..i2..","..i3..","..i4..","..(port / 256)..","..(port % 256)..").\r\n")
            if ftp_data ~= nil then
                ftp_data:close()
            end
            ftp_data = net.createServer(net.TCP, 180)
            ftp_data:listen(port, function (s)
                if data_fnc then
                    data_fnc(s)
                end
            end)
            return
        end
~~~
Then my connection with FileZilla in (passive mode) works correctly (on an integer build).
Looking at some found [information](http://slacksite.com/other/ftp.html), this seems in general necessary.

devsaurus:
> Then my connection with FileZilla in (passive mode) works correctly

@FrankX0 The only missing server constraint I can deduce from the linked information is that the server's data port should be >1023 in passive mode. This is achieved for sure by your dynamic port switching code.
Would FileZilla work for you as well if `ftp_data` listens statically e.g on port 2048 (and the `PASV` response announces this accordingly)?

FrankX0:
@devsaurus Yep! Then also FileZilla works!
So the (current) hardcoded ftp data port 20 should be either:

- configurable,
- dynamic, or
- hardcoded above 1023.

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,172 @@
...
+            b=nil
+          else
+            cd:close()
+            f:close()
+            data_fnc = nil
+            c:send("226 Transfer complete.\r\n")
+          end
+        end)
+        local b=f:read(1024)
+        sd:send(b)
```
FrankX0: Check for b not being nil, to prevent crashing reading an empty file.

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,172 @@
...
+      file.remove(a2:gsub("%/",""))
+      return c:send("250 Deleted "..a2.."\r\n")
+    end
+    if a1 == "NOOP" then
+      return c:send("200 OK\r\n")
+    end
+    if a1 == "QUIT" then
+      return c:send("221 Goodbye\r\n", function (s) s:close() end)
+    end
+    c:send("500 Unknown error\r\n")
```
FrankX0: Add support for SIZE command: this is generally used by web-browsers.
~~~lua
    if a1 == "SIZE" then
      local st,size
      st = file.stat(string.sub(a2, 2))
      if st then
        return c:send("213 "..st.size.."\r\n")
      else
        return c:send("550 Could not get file size.\r\n")
      end
    end
~~~

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,172 @@
...
+    end
+    if a1 == "CDUP" then
+      return c:send("250 OK. Current directory is "..cwd.."\r\n")
+    end
+    if a1 == "CWD" then
+      if a2 == "." then
+        return c:send("257 \""..cwd.."\" is your current directory\r\n")
+      end
+      cwd = a2
+      return c:send("250 OK. Current directory is "..cwd.."\r\n")
```
FrankX0: Suggest to change to:
~~~lua
      if a2 == "/" then
        cwd = a2
        return c:send("250 OK. Current directory is "..cwd.."\r\n")
      end
      return c:send("550 Failed to change directory.\r\n")
~~~
Needed to support web-browsers to retrieve the index of the root and prevent a CWD into a file.

NeiroNx: "cwd" in file is not main problem

NeiroNx:
need more tests, may be needs MLST,MLSD - but it needs more and more code and more memory for it

TerryE:
@NeiroNx, you are still breaking the SDK rules in the **LIST** processing.  You can't do 
```Lua
        for k,v in pairs(l) do
          sd:send( a1 == "NLST" and k.."\r\n" or "-rwxrwxrwx 1 esp esp "..v.." Jan  1  2018 "..k.."\r\n")
        end
```

Also I've done a rework your code in the latest of my [gists](https://gist.github.com/TerryE).  I don't have the time or desire to create an alternative to your implementation, as you have done all of the hard work and have the energy to see this to completion.  But copy this example and consider some of the Lua coding techniques in your own style :smile:.  

For example: 
-  [RFC 95](https://tools.ietf.org/html/rfc959) is the main specification for the FTP protocol and sec 5.3.1 summarises the list of commands that you are implementing.  These are all of the format
    -   \<CMD>\<CRLF>
    -  \<CMD>\<space>\<argument>\<CRLF>
-  Note that the command ignores case, rather than being UPPERCASE, and the the argument can include spaces. (e.g. filenames can include embedded spaces).
-  You should have a well defined and consistent approach to handling paths.  Yes, SPIFFS stores filenames in a single flat directory but "**/**" is still a vaild filename character and I still use it to create logical subdirectories or namespaces.  Also NodeMCU supports the [FatFS](https://nodemcu.readthedocs.io/en/dev/en/sdcard/) which very much uses a directory separator.
-  Lay your code out logically and for readability as per my example.  Comments are good.  Structure is good.
-  Also note my technique for splitting a ~20-long if cascade into 3 more manageable sections: bare commands, simple commands, complex commands.  The Lua pattern matching algo doesn't support PCRE-style alternatives, but using the `string.find()` function to search for `" PWD "` in `" CDUP NOOP PASV PWD QUIT "` is a fast way of doing this.
-  Always look at the  `luac.cross -l` output, and grepping this for `GLOBAL` is a good way to see if you've missed local declarations or are not local-caching some global that you should.  Grepping for `SETUPVAL` is a good way to validate where you are using upvals to store state, as is `SETTABLE`.
-  Also note my use of `send()` as a piece of peephole optimisation.  This 9 instruction function saves typically 4 instructions inline per send and also give a single point where you can add debug diagnostics if you need to.

Anyway have a look at this coding example and pick / choose anything that you feel might enhance your coding and implementation.   
 
Best regards Terry

HHHartmann:
Hi @NeiroNx first of all thx for the great work.
Just wanted to ask if you continue work on this. I have a fixe for the LIST commands and for STOR which only works for short files.

drawkula:
@HHHartmann ... please publish your changes.

TerryE:
I have been a little concerned about @NeiroNx silence.  I've got to get the LFS patch out and then I will be on a Greek island in May / early June.  If he hasn't progressed it then I will finish this off then.

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,182 @@
...
+          sd:send( a1 == "NLST" and k.."\r\n" or "-rwxrwxrwx 1 esp esp "..v.." Jan  1  2018 "..k.."\r\n")
+        end
+        sd:close()
+        data_fnc = nil
+        c:send("226 Transfer complete.\r\n")
+      end
+      if data_sock then
+        node.task.post(function() data_fnc(data_sock);data_sock=nil end)
+      end
+      return
```
HHHartmann: To make the list commands work use this data_fnc. tested with 70+ files

```
      data_fnc = function(sd)
        sd:on("sent", function(cd)
          if #files > 0 then
            local file = table.remove(files,1)
            local size = l and l[file]
            cd:send( a1 == "NLST" and file.."\r\n" or "-rwxrwxrwx 1 esp esp "..size.." Jan  1  2018 "..file.."\r\n")
          else
            l = nil
            files = nil
            cd:close()
            data_fnc = nil
            c:send("226 Transfer complete.\r\n")
            collectgarbage()
          end
        end)
        l = file.list();
        files = {}
        for file in pairs(l) do
          table.insert(files, file)
        end
        if a1 == "NLST" then l = nil end
        local file = table.remove(files,1)
        local size = l and l[file]
        sd:send( a1 == "NLST" and file.."\r\n" or "-rwxrwxrwx 1 esp esp "..size.." Jan  1  2018 "..file.."\r\n")
      end
```

lua_modules/ftp/ftpserver.lua
```diff
@@ -0,0 +1,182 @@
...
+      c:send("150 Accepted data connection\r\n")
+      data_fnc = function(sd)
+        sd:on("receive", function(cd, dd)
+          f:write(dd)
+        end)
+        sd:on("disconnection", function(c)
+          f:close()
+          data_fnc = nil
+        end)
+        c:send("226 Transfer complete.\r\n")
```
HHHartmann: The ``c:send("226 Transfer complete.\r\n")`` has to be executed after disconnect only. Else it stops sending the file after a couple of blocks which results in cut off files.

The ``on disconnection`` should look loke this: (also note the rename of the function param from ``c`` to ``cd``)

```
        sd:on("disconnection", function(cd)
          f:close()
          data_fnc = nil
          c:send("226 Transfer complete.\r\n")
        end)
```

HHHartmann:
@TerryE If you would not finish it, is there a way/best practice to hand over a PR to someone else (without commit rights)?
Btw: tested with Win10 file explorer.
Next I will lock into directory emulation to handle the files containing "/" in their names.
With LFS in place the size concern that @NeiroNx had does not matter that much either hopefully.

TerryE:
I think that @NeiroNx has timed out.  I will close this PR and open another based on my re-implementation.  No withstanding this, the impetus and original work has been done by NeiroNx so he or she deserves our thanks and full acknowledgment. :+1:

PR #2358, opened 2018-04-25
Update node.dsleep() to support longer deep sleep duration.

### LINKED ISSUE #2351: Upgrade node.dsleep to SDK >=2.1
SDK 2.1 changed the parameter for `system_deep_sleep` from uint32 to uint64. This allows deep sleep intervals beyond the current 71min boundary. It would also be useful to have a `node.dsleepmax` like Arduino has to get the most out those extended deep sleep capabilities: https://github.com/esp8266/Arduino/blob/2.4.1/cores/esp8266/Esp.cpp#L119

https://nodemcu.readthedocs.io/en/latest/en/modules/node/#nodedsleep needs updating as well.

More details at https://thingpulse.com/2018/04/10/max-deep-sleep-for-esp8266/ (disclaimer: it's from yours truly)

dnc40085:
Fixes #2351.

Beginning with SDK 2.1, the functions `system_deep_sleep()` and `system_deep_sleep_instant()` were changed to accept `uint64` values, allowing for longer sleep duration.
This PR updates `node.dsleep()` to accept these larger sleep times.

### DIFF
--- a/docs/en/modules/node.md
+++ b/docs/en/modules/node.md
@@ -87,9 +87,8 @@ dofile("hello.lc")
 
 Enters deep sleep mode, wakes up when timed out.
 
-The maximum sleep time is 4294967295us, ~71 minutes. This is an SDK limitation.
-Firmware from before 05 Jan 2016 have a maximum sleeptime of ~35 minutes.
-
+Theoretical maximum deep sleep duration can be found with [`node.dsleepMax()`](#nodedsleepmax)
+ 
 !!! caution
 
     This function can only be used in the condition that esp8266 PIN32(RST) and PIN8(XPD_DCDC aka GPIO16) are connected together. Using sleep(0) will set no wake up timer, connect a GPIO to pin RST, the chip will wake up by a falling-edge on pin RST.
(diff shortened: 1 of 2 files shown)


docs/en/modules/node.md
```diff
@@ -131,6 +127,38 @@ node.dsleep(nil,4)
...
 - [`node.sleep()`](#nodesleep)
+- [`node.dsleepMax()`](#nodedsleepmax)
+
+## node.dsleepMax()
+
+Returns the current theoretical maximum deep sleep duration.
+
+!!! caution
+
+	While it is possible to sleep longer than the theoretical maximum sleep duration, which can be found with `node.dsleepMax()`, it is not recommended to exceed this theoretical maximum.
```
marcelstoer: Did you test this? My [own tests](https://thingpulse.com/max-deep-sleep-for-esp8266/) conducted with Arduino core showed that the chip would sleep indefinitely if you went beyond. Even if you use dsleep max it will wake up before time is up (again in my tests).

dnc40085: I did... but I only got it to wake up once and it registered as an external reset(6) rather than a wake from deep sleep() and I made a mistake in the logging code, so times weren't recorded. I will have to do more testing.

PR #2367, opened 2018-05-08
Add message indicating that node.sleep() was disabled during build.

dnc40085:

I was browsing the NodeMCU forum on esp8266.com and found a [thread] (https://www.esp8266.com/viewtopic.php?f=23&t=17056) that reminded me that I forgot to include a message letting the developer know that `node.sleep()` is disabled, so here is the fix for that.


marcelstoer:
Thanks!

PR #2383, opened 2018-05-27
Fix for lack of nil return in file.read() when EOF is reached

### LINKED ISSUE #2376: file.read(): On end of file, rather than nil, nothing is returned
### Expected behavior
The functions [`file.read()`](http://nodemcu.readthedocs.io/en/dev/en/modules/file/#fileread-fileobjread) and [`file.readline()`](http://nodemcu.readthedocs.io/en/dev/en/modules/file/#filereadline-fileobjreadline) should [return `nil`](http://nodemcu.readthedocs.io/en/dev/en/modules/file/#returns_14) on subsequent calls after end of file has been reached.
### Actual behavior
After end of file is reached, nothing is returned on subsequent calls.
### Test code
```Lua
do
  file.open("init.lua", "r")
  while true do 
    retval=type(file.read())
    print(retval)
    if retval=="nil" then print("this never happens") break end
  end
  file.close()
  print("End (This also never happens)")
end
```
This code never reaches the end, it just errors out with:
`stdin:5: bad argument #1 to 'type' (value expected)`
### NodeMCU version
 dev( 914a4af)
### Hardware
ESP12-F

### Fix 
in `/app/modules/file.c` in the function `file_g_read()` on line [435](https://github.com/nodemcu/nodemcu-firmware/blob/914a4afc96136311db8ab923b8a6942f429d266f/app/modules/file.c#L435) replace:
```c
return 0;
```
with:
```c
lua_pushnil(L);
return 1;
```
Unless the documented return value is in error, I can submit a pull request with this fix.

dnc40085:
Fixes #2376.

This PR adds the missing `nil` return when a file is stepped through with `print(file.read())` and the end of file is reached.

### DIFF
--- a/app/modules/file.c
+++ b/app/modules/file.c
@@ -432,7 +432,8 @@ static int file_g_read( lua_State* L, int n, int16_t end_char, int fd )
       luaM_free(L, heap_mem);
       heap_mem = NULL;
     }
-    return 0;
+    lua_pushnil(L);
+    return 1;
   }
 
   vfs_lseek(fd, -(n - i), VFS_SEEK_CUR);


TerryE:
It's a pity that you didn't go through the rest of the modules and fix the other instances of this but I suppose a tiny step forward is still a step forward.

marcelstoer:
> a step forward is still a step forward

👍-> thanks @dnc40085

PR #2389, opened 2018-06-02
Update coap.c

hsj1104:

- [ ] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).

### DIFF
--- a/app/modules/coap.c
+++ b/app/modules/coap.c
@@ -1,6 +1,7 @@
 // Module for coapwork
 
 #include "module.h"
+#include <stdio.h>
 #include "lauxlib.h"
 #include "platform.h"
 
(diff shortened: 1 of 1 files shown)


TerryE:
@hsj1104 , this is a one line diagnostic that you might find useful, but this isn't something that we would want to add to our standard distro for all users. Also diagnostics should use the standard macros / API so that they can be correctly enabled and disabled using configuration options.

marcelstoer:
In addition to what Terry already said please stick to the template and have a look at the guidelines in references. There's a lot of useful information in there e.g. that we don't accept PRs against master.

PR #2399, opened 2018-06-14
Update README.md

nailujx86:
Fixes no issue.

Updated the percentage of C-Code used in the firmware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ### A Lua based firmware for ESP8266 WiFi SOC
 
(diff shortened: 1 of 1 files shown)


TerryE:
Why?  I am not sure why the original overview left out 1.9% as non-C what is this? the Makefiles? The loader files? All of the source in the firmware is C, and the small amount of assembler is embedded in C sources using the `asm()` directive.  We don't use any other high level language in the firmware, though we do have some host-side python.

marcelstoer:
>  I am not sure why the original overview left out 1.9% as non-C what is this? the Makefiles? The loader files?

It's just what GitHub reports if you click that color bar (which in our case isn't that colored at all).
screen shot 2018-06-15 at 13 58 27

TerryE:
Thanks @marcelstoer, but this is also misleading as the Lua is the Lua examples which are not part of the firmware, and the Python is the host-end `esptool.py`.  Not sure where the C++ is as we don't have any .cpp files in the app, though some of the headers are "bilingual" in that they are designed to be included in C++ source as sell as C and may have been therefore classified as C++; so the real % is 100% C.

marcelstoer:
Well, that whole sentence isn't particularly helpful

> The code repository consists of 98.6% C-code that glues the thin Lua veneer to the SDK.

TerryE:
Closing this; see #2418.   Thanks @nailujx86  Julian.  :smile:

Issue #2395, opened 2018-06-08
Upgrade to SDK 2.2.1


marcelstoer:
https://github.com/espressif/ESP8266_NONOS_SDK/releases/tag/v2.2.1

@devsaurus and @nwf you did this last time. Would you have spare time to work on this?

nwf:
From a quick glance through the deltas from v2.2.0, this should be real fast; almost all of the changes are in binary blobs rather than source.  I'll see if I have time this weekend.

Issue #2398, opened 2018-06-14
node.info() should return better information about the actual NodeMCU version


joysfera:
### Expected behavior
`node.info()` returns something concrete that identifies the version properly. At the very least the `devVer` would increment on each master release. Or a new `releaseDate` would be added to the `node.info()` table that would contain `20180608`, a number derived from date that would increase on each release.

### Actual behavior
`node.info()` returns `2, 2, 0` in its Ver fields for both the April and June releases, even though there were important bugfixes and even API changes in the latter.

I wish a Lua application was able to identify the NodeMCU build precisely enough so I'd know which nodes need updating the NodeMCU firmware when a new release with an important bugfix gets available.

### Test code
```Lua
=node.info()
```
### NodeMCU version
2.2.0-master_20180608

### Hardware
ESP-07

TerryE:
We already edit `user_config.h` in our Travis script. Maybe we should add `$(git rev-parse HEAD| cut -b 1-8)` to the version string or even just as a `-D` option in the `app/lua` Makefile.

joysfera:
I'm not sure if git commit ID is a number that could be compared with another number to find out whether a node is running an outdated NodeMCU or not. That's why I suggested the `20180608` number (currently part of the release "name") that is guaranteed to be increasing on each release.

Granted, it wouldn't identify git builds, but it would work for released ones.

Maybe we'd need both things - the exact commit of the build and also a version number that does increase on release.

TerryE:
All releases are published on the basis of a git commit, so any release has a unique git HEAD. What you seem to be wanting is an ordered value rather than a unique one.  What about the DTS of the commit:
```bash
git show -s --format=%ci HEAD
```

joysfera:
Well, that would work (more or less) if turned into an integer YYYYMMDD. I realize it doesn't allow for two releases in one day but I don't think that's an issue given that releases are at least two months apart.

But I'm surprised that we have `majorVer`, `minorVer` and `devVer` - seems almost like https://semver.org/ but it's not used that way. Any reason why?

marcelstoer:
> But I'm surprised that we have `majorVer`, `minorVer` and `devVer` - seems almost like https://semver.org/ but it's not used that way. Any reason why?

I didn't check the code but IIRC those values represent _the SDK_ version.

joysfera:
[...]

PR #2401, opened 2018-06-16
Espressif SDK v2.2.1

nwf:
Minimally build and boot tested, but the changes look pretty boring.
Fixes https://github.com/nodemcu/nodemcu-firmware/issues/2395

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,20 +3,20 @@
 .NOTPARALLEL:
 
 # SDK base version, as released by Espressif
-SDK_BASE_VER:=2.2.0
+SDK_BASE_VER:=2.2.1
 
 # no patch: SDK_VER equals SDK_BASE_VER and sdk dir depends on sdk_extracted
-#SDK_VER:=$(SDK_BASE_VER)
-#SDK_DIR_DEPENDS:=sdk_extracted
+SDK_VER:=$(SDK_BASE_VER)
+SDK_DIR_DEPENDS:=sdk_extracted
 
 # with patch: SDK_VER differs from SDK_BASE_VER and sdk dir depends on sdk_patched
-SDK_PATCH_VER:=f8f27ce
-SDK_VER:=$(SDK_BASE_VER)-$(SDK_PATCH_VER)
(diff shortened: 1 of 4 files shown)


devsaurus:
Looks good.

Espressif missed to update `ESP_SDK_VERSION_PATCH` to `1` in https://github.com/espressif/ESP8266_NONOS_SDK/blob/v2.2.1/include/version.h.
The firmware still reports
```
NodeMCU 2.2.0.0 build unspecified powered by Lua 5.1.4 on SDK 2.2.1(6ab97e9)
```

nwf:
Bah humbug re: Espressif.  We could leave it be or change `user_version.h` either to replace

    #define NODE_VERSION_REVISION	ESP_SDK_VERSION_PATCH

with

    #define NODE_VERSION_REVISION	1

or just bump `NODE_VERSION_INTERNAL` instead?  I think I'd be more in favor of the last or the first, but this might just be bikeshedding.

marcelstoer:
> We could leave it [...]

In light of the discussions in #2398 I don't think we should.

nwf:
@marcelstoer How do you feel about calling this 2.2.0.1 while we wait on Espressif?

marcelstoer:
@nwf I wouldn't do that. If Espressif fixes this we'd be shipping essentially the same SDK as before and, therefore, an change from 2.2.0.1 to 2.2.1.0 would feel odd to me.

jmd13391:
Why don't we simply wait on SDK 2.2.1 until Espressif releases a fix?
Does anyone know if Espressif is aware of the issue?

marcelstoer:
SDK is already released, that's what this PR is about. See my comment further up.

nwf:
@marcelstoer Admittedly, I was imagining we'd wait until 2.2.2 or 2.3.0 or 3.0.0 or whatever to re-merge with upstream, so we'd not do a 2.2.0.1 to 2.2.1.0 transition.  I'm content to wait a bit to see what Espressif does.

marcelstoer:
> I'm content to wait a bit to see what Espressif does.

They haven't even commented on my PR in a month...

PR #2416, opened 2018-07-01
New Telnet module

TerryE:
This is a network-efficient implementation of the Telnet server that can handle significantly larger output per command without exhausting memory.

- [x]  The example is documented in its own README.md

### DIFF
--- a/lua_examples/telnet/README.md
+++ b/lua_examples/telnet/README.md
@@ -1,0 +1,79 @@
+# Telnet Module
+
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2014-12-22 | [Zeroday](https://github.com/funshine) |  [Terry Ellison](https://github.com/TerryE) | [simple_telnet.lua](./simple_telnet.lua) |
+| 2018-05-24 | [Terry Ellison](https://github.com/TerryE) |  [Terry Ellison](https://github.com/TerryE) | [telnet.lua](./telnet.lua) |
+
+
+The Lua telnet example previously provided in our distro has been moved to this 
+file `simple_telnet.lua` in this folder. This README discusses the version complex 
+implementation at the Lua module `telnet.lua`.  The main reason for this complex
+alternative is that a single Lua command can produce a LOT of output, and the 
+telnet server has to work within four constraints:
+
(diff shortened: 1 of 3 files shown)


TerryE:
Some may wonder why I've put this in the _examples_ subdir rather than the _modules_ one.  It was 'toss a coin'.  I wanted to keep the old telnet example for reference and it isn't a module, and I wanted to keep them in the same directory, as two telnet examples in separate directories would just be too confusing.  If there is a strong view that it needs moved then we can do so.

Yes, the new version is a module, but it's not the sort of module that you'd integrate into a normal app, IMO.
  
Creating large arrays of small strings burns through RAM, and because the Lua RTS calls the `node.output()` CB on a once per field basis, then it's quite easy to type in a `for` expression that will exhaust memory and reboot the ESP, if you don't do this sort of marshaling in the code.

lua_examples/telnet/README.md
```diff
@@ -0,0 +1,79 @@
...
+| 2018-05-24 | [Terry Ellison](https://github.com/TerryE) |  [Terry Ellison](https://github.com/TerryE) | [telnet.lua](./telnet.lua) |
+
+
+The Lua telnet example previously provided in our distro has been moved to this 
+file `simple_telnet.lua` in this folder. This README discusses the version complex 
+implementation at the Lua module `telnet.lua`.  The main reason for this complex
+alternative is that a single Lua command can produce a LOT of output, and the 
+telnet server has to work within four constraints:
+
+-  The SDK rules are that you can only issue one send per task invocation, so any
```
pjsg: I'm not convinced that this is true. I think that the real rule is that you can call send as often as you like, but if it fails, then you must retry that send later (after the on:sent callback is called). 

I might try and hack on the net module to make this happen automatically.

TerryE: This was a strict rule, but you are probably correct in that open LwIP has probably relaxed this, however if you do a `print(a,b,c,d)` then this sends 5 separate fields and stick something like this in a for loop then you reboot the ESP with the old telnet. The new one is far more robust and you can happily print out 10k.

marcelstoer:
> Some may wonder why I've put this in the _examples_ subdir rather than the _modules_ one.

The PR title says "New Telnet module". You must have been really torn 😜

PR #2423, opened 2018-07-09
Add checking of lua sources to travis build

HHHartmann:

There are several *.lua files that are not compilable.
This PR adds checking lua files in travis builds.
Travis build will fail if there is a lua file in lua_modules or lua_samples that cannot be compiled.

### DIFF
--- a/app/pm/pmSleep.c
+++ b/app/pm/pmSleep.c
@@ -1,5 +1,5 @@
 #include <pm/pmSleep.h>
-#ifdeff  PMSLEEP_ENABLE
+#ifdef  PMSLEEP_ENABLE
 #define STRINGIFY_VAL(x) #x
 #define STRINGIFY(x) STRINGIFY_VAL(x)
 
--- a/app/pm/pmSleep.c
+++ b/app/pm/pmSleep.c
@@ -1,5 +1,5 @@
 #include <pm/pmSleep.h>
-#ifdef  PMSLEEP_ENABLE
+#ifdeff  PMSLEEP_ENABLE
 #define STRINGIFY_VAL(x) #x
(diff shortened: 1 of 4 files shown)


HHHartmann:
So found two more build errors in Lua code
Ready for integration now. (from my POV only of corse)

HHHartmann:
@TerryE Terry I do not quite understand what you ask me to do (or even not to do). Shouldn't I have merged the remote-tracking branch? I am not yet very familiar with git so I would be more than happy to be enlightened a bit more.
But most of all let me thank you for the great LFS patch. It is great to have telnet and ftp running and still have 40k of RAM left.

TerryE:
@HHHartmann Gregor, it is far more important that you feel comfortable contributing which is why Marcel and I were happy to merge your PR with multiple commits.  The rebasing stuff is covered [here](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) in the online Git book, but to be honest I get confused with the advanced features of git at times!  Sometime I'll get a round to a simple 1-2-3 guide.

HHHartmann:
@TerryE Terry, thanks for the link. I will read it and find out how to do it. But still I wonder _what_ I should have done? Should it have been only one commit? Puzzled and curious.

TerryE:
>   But still I wonder what I should have done
For a change of this size and scope I would had one or two commits, and if two then I would done the \*.lua changes as one commit.

HHHartmann:
Ah ok.
I had to do separate commits here because I could see no way to test beforehand as the travis build only gets triggered there. But I suppose I could have squashed them afterwards? Or reduce them by first thinking, then committing.
Thanks four your guidance anyways.

PR #2427, opened 2018-07-14
Disable stdout buffering to enable character echo

ryanplusplus:

On the `dev-esp32` branch, typing into the console/REPL does not echo characters typed until a newline is entered. This is because stdout is buffered and not flushed until a line is completed.

This change sets stdout to be unbuffered so that it is automatically flushed for every character typed by a user. I considered flushing immediately after each echo, but I found that there were several places in the code where a byte is echoed to the console and didn't want to litter `fflush` calls throughout the code.

I'm fairly certain that this is the wrong place to put this line, but I'm not sure what location would be preferred. Any feedback or help is welcome.

### DIFF
--- a/components/lua/lua.c
+++ b/components/lua/lua.c
@@ -39,7 +39,7 @@ static void lstop (lua_State *L, lua_Debug *ar) {
 
 
 static void laction (int i) {
-  // signal(i, SIG_DFL); 
+  // signal(i, SIG_DFL);
   /* if another SIGINT happens before lstop,
                               terminate process (default action) */
   lua_sethook(globalL, lstop, LUA_MASKCALL | LUA_MASKRET | LUA_MASKCOUNT, 1);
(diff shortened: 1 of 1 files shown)


components/lua/lua.c
```diff
@@ -39,7 +39,7 @@ static void lstop (lua_State *L, lua_Debug *ar) {
 
 
 static void laction (int i) {
-  // signal(i, SIG_DFL);
```
ryanplusplus: My editor is configured to strip trailing whitespace. I can rework this to have only the added line if that would be preferred.

components/lua/lua.c
```diff
@@ -384,6 +384,7 @@ struct Smain {
 
 
 static int pmain (lua_State *L) {
+  setbuf(stdout, NULL);
```
ryanplusplus: This is the line that was added to remove buffering for stdout.

jmattsson:
This sounds reasonable to me. @devsaurus what do you reckon?

ryanplusplus:
@devsaurus would you mind taking a look at this? It is a small change that makes a big difference when using the REPL.

devsaurus:
Wow, completely missed this one.
Is the buffer issue still present with #2508 which was just merged?

ryanplusplus:
Nope, looks like that covered it and IMO @tomsci had a cleaner fix. Thanks!

PR #2429, opened 2018-07-16
Add option to strip debug symbols (false by default)

ryanplusplus:

This makes stripping debug symbols during compilation optional. By default, it does not enable stripping (to match behavior for ESP8266). This makes Lua stack traces usable.

### DIFF
--- a/components/modules/node.c
+++ b/components/modules/node.c
@@ -175,6 +175,7 @@ static int node_compile( lua_State* L )
   int file_fd = 0;
   size_t len;
   const char *fname = luaL_checklstring( L, 1, &len );
+  int stripping = lua_toboolean( L, 2 );
   const char *basename = vfs_basename( fname );
   luaL_argcheck(L, strlen(basename) <= CONFIG_FS_OBJ_NAME_LEN && strlen(fname) == len, 1, "filename invalid");
 
@@ -196,8 +197,6 @@ static int node_compile( lua_State* L )
   }
 
   f = toproto(L, -1);
-
-  int stripping = 1;      /* strip debug information? */
 
   file_fd = vfs_open(output, "w+");
   if (!file_fd)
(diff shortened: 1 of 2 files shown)


ryanplusplus:
@devsaurus one more quick change for you to look at if you wouldn't mind

devsaurus:
I'm not too familiar with the Lua VM details, but there are probably already knobs and switches to achieve what you're looking for.
- Static: `make menuconfig` -> "Component config" -> "NodeMCU miscellaneous" -> "Discard debug..."
- Dynamic: [`node.stripdebug()`](https://nodemcu.readthedocs.io/en/dev-esp32/en/modules/node/#nodestripdebug)

devsaurus:
Trying yo summarize... Stripping of debug information with `node.compile()` is done unconditionally and independently from the settings with `node.stripdebug()`.
If this is correct, then the statement in the [doc](https://nodemcu.readthedocs.io/en/dev/en/modules/node/#nodestripdebug) is misleading, if not plain wrong:
> Controls the amount of debug information kept during `node.compile()`

I didn't get
> This is not how it works for mainline NodeMCU for ESP8266.

As far as I can tell, this part of the Lua core is identical on both platforms. What's the behavior of mainline NodeMCU for ESP8266 in this respect?

ryanplusplus:
I went back to reconfirm that this is a problem and you're absolutely right, everything is working as expected. I'm not sure if something changed since June or if I had inadvertently disabled debug info. Clearly I should have done this a few days ago before wasting your time. My bad.

HHHartmann:
@ryanplusplus I also experience the same behavior as you describe. when compiling with node.compile() the debug info is stripped unconditionally. On ESP8266 tjhat is.
If you found a way to circumvent this please let us know or are you stating that node.compile() now respects node.stripdebug() settings?

ryanplusplus:
@HHHartmann I just tried on an ESP8266 and I'm seeing what you are describing. No matter what the `node.stripdebug()` settings, I am unable to compile without stripping debug info. I am not seeing this behavior on ESP32 (at least not now).

ryanplusplus:
@devsaurus sorry, but I have one more update: my last test results were wrong. When I tested this I tested it on my fork's `dev-esp32` branch where I already applied this patch. I tested again on `dev-esp32` and it respects neither the static nor dynamic options. I have to either cross-compile off target or apply this change.

Issue #2431, opened 2018-07-16
Improving LFS documentation


TerryE:
This is a detailed spin-off of #2413.  We have now got LFS deployed in dev, but there are a number of rough edges, particularly in how it is documented and introduced to existing and new users

### What documentation current exists

-  ✅The [LFS whitepaper](http://nodemcu.readthedocs.io/en/dev/en/lfs/) is now included in the newly reorganised whitepapers section.  This goes into a lot of technical detail that is of no interest to Lua IoT developers who are only interested in what it can allow them to do, and not the magic behind it.
-  ✅[node.flashindex()](http://nodemcu.readthedocs.io/en/dev/en/modules/node/#nodeflashindex) and [node.flashreload()](http://nodemcu.readthedocs.io/en/dev/en/modules/node/#nodeflashreload) are properly documented in the **node** module but neither refer back to the LFS whitepaper :scream: 
- Some examples do exist, but more needs added here.

### Justification for extra work

LFS is the biggest change to NodeMCU since Johny's unaligned exception handler patch allowing most constant data to be moved into Flash. The use of LFS transforms the sort of applications that you can realistically develop for the ESP8266, yet it is in the "best kept secrets" category.   We should headline this somehow for users.
 
### Scope of documentation changes

-  ✅ The intro in the README.md should refer to the new LFS feature and link to the relevant documentation -> 4c6c9d3
-  ✅ The lead to the online documentation should do likewise.
-  ✅ The LFS whitepaper needs polishing.  I am probably to geeky and too close to the implementation to write this well for consumption by the average developer.  @vsky279 Lukáš has suggested a first cut in this draft commit c39891f9.  We need something like this.
-  The examples need expanding. 
-  (Out of scope but very important).  
    - ✅ Since a lot of developers use [esp8266.com](http://esp8266.com) and the [NodeMCU custom build service](https://nodemcu-build.com/), we should use these to advertise the new functionality, in the former by a sticky post, and the latter either by a news banner or a reference on the "your build is complete" email. 
    - ✅ Announce it on https://www.facebook.com/NodeMCU -> @marcelstoer

@vsky279 already has my email details;  unfortunately github doesn't offer PM, but @HHHartmann Gregor if you want to have a side chat then I still have an apache org mailbox from my committer days there (same userid); ping me an email and I'll reply on the mail address I usually use,

vsky279:
[...]

PR #2432, opened 2018-07-17
Adding "LFS Quick Start" section in the documentation

vsky279:
Relates #2431 (Improving LFS documentation).

Does not fix anything but is adding a "Quick Start" section into the LFS documentation.

### DIFF
--- a/docs/en/lfs.md
+++ b/docs/en/lfs.md
@@ -4,24 +4,30 @@
 
 LFS is a way how to execute your Lua code out of ESP8266 flash memory so more RAM is available for variables and data structures. This way large Lua files can be run on ESP8266. There would not be enough RAM to execute such files in a "normal" (out of SPIFFS) way.
 
+The tutorial assumes that you are able to build a nodemcu-firmware on a Windows 10 host.
+
 This is a simple step-by-step how to use the LFS feature:
 1. Get and flash LFS enabled firmware
 
-    Either you get it from https://nodemcu-build.com/. In the section "LFS options (currently just for dev)" choose the size of the LFS partition (64 KB should be fine). SPIFFS default settings should be fine.
+    Either you get it from [NodeMCU Build](https://nodemcu-build.com/). In the section "LFS options (currently just for dev)" choose the size of the LFS partition (64 KB should be fine). SPIFFS default settings should be fine.
     
-    Another possibility is to compile own firmware and enable LFS in [user_config.h](../../app/include/user_config.h), setting `#define LUA_FLASH_STORE 0x10000`.
-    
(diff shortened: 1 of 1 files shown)


vsky279:
@TerryE Thanks for your comments. I understand that you would prefer to have it in a separate file and LFS documentation can just cross-reference it. For the moment it is still in the main LFS doc but can be separated before it's merged.

Please have a look at the updates. I tried to reflect your comments.

What can be for a beginner a bit difficult is indeed the step with building the LFS image. He can download the LFS enabled firmware but still needs to be able the `luac.cross`. It works for me as I Docker but I imagine it would be quite difficult if I were to start from a scratch.

@HHHartmann Gregor, I think I am not referencing `init.lua` in Step 4. I've added a small note into Step 6 but I don't think the role of `init.lua` is to be discussed in this tutorial.

marcelstoer:
> I think that it merits its own separate section in the documentation tree.

Currently there _is_ no section anymore it could be moved to. You removed it white the latest reorg (introduction of "Whitepapers"). It wouldn't be logical for me to have two pages that talk about LFS. If we did that we'd have to do the same for the SPIFFS document. IMHO the new section belongs exactly to where it's currently sits.

However, the essential thing is that we've got the tutorial available _at all_. So, do as you prefer.

TerryE:
> Can we merge this?

As you say we can refine it later.  It's valuable as it.

I need to think about the LFS cross compiler bits.

marcelstoer:
@TerryE can I suggest you "Squash and merge" PRs unless there's a specific reason not to (rather than "Create a merge commit"). For this simple change we've now got 4 commits in our history. Not that we didn't have more important things to worry about but this one wouldn't cost anything extra 😉

TerryE:
Yup, point taken.

vsky279:
Gentlemen, there's some missformatting on http://nodemcu.readthedocs.io/en/dev/en/lfs/
I think there is a need for and CRLF after line 7.

marcelstoer:
I meant to verify it after the merge but got interrupted. Thanks, I'll fix that right away.

marcelstoer:
I was just about to add a link to [Compiling Lua on your PC for Uploading](http://nodemcu.readthedocs.io/en/dev/en/upload/#compiling-lua-on-your-pc-for-uploading) (for telling folks how to get a hold of `luac.cross`) when I noticed that this section is outdated. There's no `tools/cross-lua.lua` anymore.

@TerryE can you fix that?

jyh29:
Why not setup a web service under pay for users to cross compile the Nodemcu-firmware according to users uploading include/user_config.h ?

PR #2434, opened 2018-07-18
Improve documentation for wifi.monitor

### LINKED ISSUE #2433: WiFi Monitor is always set to channel 1 when started
I'm trying to monitor all different channels using my NodeMCU EPS8266 module. However, I quickly ran into the issue that whatever I tried, the monitored channel was always 1.

Looking into the source code, I found that upon starting the channel is set back to 1.

https://github.com/nodemcu/nodemcu-firmware/blob/8181c3be7aed9f0a0ceb73ac8137c1a519e8a8e9/app/modules/wifi_monitor.c#L651

A workaround is to set the channel after starting the monitor, but that can be quite confusing. Even the provided example in the docs would not work, since the channel there is also set before starting the monitor.

https://nodemcu.readthedocs.io/en/master/en/modules/wifi_monitor/#example

I don't know if the call to `wifi_set_channel` is needed after changing into monitor mode, otherwise, the line can simply be removed.

Other solutions would be:

- Clearly specify this behaviour in the documentation
- Add an argument to `wifi.monitor.start` which specifies the channel

HugoDaBosss:
Fixes #2433.

Better specifies the behaviour of `wifi.monitor.start` and `wifi.monitor.channel`


marcelstoer:
Never mind, I'll do it myself.

PR #2438, opened 2018-07-25
Enhanse file module for large number of files

nedoskiv:
First I'm new here (on github) and this may not be posted properly.

I'm developing on nodemcu lua, and my work produce large number of files.
problem is when they reach over 250-300 , file.list() function return a table and full all memory of device, causing it to reboot. Is it possible to be made function that return partial file list and take arguments like:

file.Elist (20,15)  and return 15 files starting from 20th? or some other way to get whole file list when got around 1000 files in file system?


marcelstoer:
Welcome to GitHub and NodeMCU!

> this may not be posted properly.

Indeed, you posted a pull request rather than an issue.

> Is it possible to be made function that return partial file list

Someone would have to implement that first. However, please take a look at #1010.

nedoskiv:
Sorry to bother you again, do I need to post a issue in order someone to pay attention to it. Or if I am  pointed in right direction may try to implement it too. Got limited C knowledge but may try.

Issue #2413, opened 2018-06-29
LFS follow-up issues and missing features.


TerryE:
Rather than following the Bugrep or Feature request templates, this issue is more of a post LFS merge follow up to #2292 (which really discussed trade-offs up to our LFS implementation and at 48 comments some large is becoming unwieldy to load).  I will use this issue to track LFS-related feedback and functional gaps.  We can use this for a catch-all for the next few weeks until our next LFS patch update, and after this we can use the normal Bugrep / FR templates to track any follow as individual issues.

TerryE:
#### Support for luac.cross and spiffsimg 

See #2315.  We now have WSL and Cygwin methods which cover all WinX versions.  Of course the other option is to buy yourself a $15 RRi Zero and load [DietPi](http://www.dietpi.com) and make luac.cross and use a small ssh script (either WSL native SSH or OpenSSH) from your PC to offload the compiles to the RPi0.

#### Cloud builder generating LFS firmware

In progress.  See marcelstoer/nodemcu-custom-build#29 

#### Additional LFS-enabled Lua Modules

In addition to `lua_examples/lfs` I have added PR #2416 (Telnet) and #2417 (FTP).

#### Compression Support

In progress.  See #2409.

#### Updates to FAQs and LFS whitepaper

Pending feedback from users.

HHHartmann:
I recently got my first LFS installation running. Here are some obstacles and suggestions I came across:

I used a docker build in standard configuration which built an integer and a float image.
I decided to flash the float image (which gets built before the int image)

I then found that aside the local/fs directory a local/lua directory is recognized by the tools make.

> 1. suggestion: add .gitignore file to local/lua to have it created automatically on checkout (just like in local/fs)

so I put some lua files in there and built an LFS image (make in tools directory, image stored in local/fs).

> 2. Suggestion: in the whitepaper you use LFS image once for a firmware image that is LFS enabled and once for an image that contains compiled lua and can be flashed in the firmware image. I had to read it several times to get it sorted out.

I then transferred the generated image to the ESP and called ''node.flashreload()''
No flashing took place and no reboot.
After a sleepless night it occurred to me that there could be a difference in int and float builds.
So building the int _after_ the float version had overwritten the float version of luac.cross.
So I rebuilt the float version, rebuilt the LFS lua image and could load my first modules as LFS. chacka!

> 3. Suggestion: have only one version of loac.cross with a switch to generate int or float images.

[...]

PR #2448, opened 2018-08-02
Add compression to LFS images

TerryE:
Relates to  #2413.

Sorry guys but another big patch.  What this does is to add a new library `app/uzlib` to provide ESP8266-compatible deflate/inflate support, and extend both `lflash.c` and `lflashimg.c` to use this.  

Compression drops the over-network and on-SPIFFS file size for the LFS image by ~3×.  The decompression process is two pass and the 1st is non-destructive and has a bunch of integrity checks in it (including the gzip CRC) so `node.flashreload()` will now return a meaningful error message if there is an LFS corruption or mismatch.

I've also added an extra lua example showing how to use this for Lua OTA and an extra documentation page on `luac.cross`.

Because of the size of this change I will have to do more pushes before the merge, I think.

### DIFF
--- a/app/lua/lflash.c
+++ b/app/lua/lflash.c
@@ -76,7 +76,6 @@ struct OUTPUT {
   int     (*fullBlkCB) (void);
   int       flashLen;
   int       flagsLen;
-  int       flagsEst;
   int       flagsNdx;
   uint32_t *flags;
   const char *error;
@@ -96,7 +95,7 @@ void dumpStrt(stringtable *tb, const char *type) {
       NODE_DBG("%5d %5d %08x %08x %5d %1s %s\n",
                i, j, (size_t) ts, ts->tsv.hash, ts->tsv.len,
                ts_isreadonly(ts) ? "R" : " ",  getstr(ts));
---    }
+    }
 }
 
 LUA_API void dumpStrings(lua_State *L) {
@@ -145,25 +144,25 @@ static void flashErase(uint32_t start, uint32_t end){
 
 /* =====================================================================================
  * luaN_init(), luaN_reload_reboot() and luaN_index() are exported via lflash.h.
- * The first is the startup hook used in lstate.c and the last two are 
- * implementations of the node.flash API calls. 
+ * The first is the startup hook used in lstate.c and the last two are
+ * implementations of the node.flash API calls.
  */
 
 /*
  * Hook in lstate.c:f_luaopen() to set up ROstrt and ROpvmain if needed
  */
 LUAI_FUNC void luaN_init (lua_State *L) {
-  curOffset      = 0;
-  flashAddr     = flash_region_base;
-  flashAddrPhys = platform_flash_mapped2phys((uint32_t)flashAddr);
-  flashSector   = platform_flash_get_sector_of_address(flashAddrPhys);
+  curOffset       = 0;
+  flashAddr       = flash_region_base;
+  flashAddrPhys   = platform_flash_mapped2phys((uint32_t)flashAddr);
+  flashSector     = platform_flash_get_sector_of_address(flashAddrPhys);
   FlashHeader *fh = cast(FlashHeader *, flashAddr);
 
   /*
-   * For the LFS to be valid, its signature has to be correct for this build variant,
-   * thr ROhash and main proto fields must be defined and the main proto address
-   * be within the LFS address bounds. (This last check is primarily to detect the
-   * direct imaging of an absolute LFS with the wrong base address.
+   * For the LFS to be valid, its signature has to be correct for this build
+   * variant, the ROhash and main proto fields must be defined and the main proto
+   * address be within the LFS address bounds. (This last check is primarily to
+   * detect the direct imaging of an absolute LFS with the wrong base address.
    */
 
   if (fh->flash_sig == 0 || fh->flash_sig == ~0 ) {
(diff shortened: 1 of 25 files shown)


mkdocs.yml
```diff
@@ -24,6 +24,7 @@ pages:
     - Home: 'en/index.md'
     - Building the firmware: 'en/build.md'
     - Flashing the firmware: 'en/flash.md'
+    - Compiling Lua:  'en/compiling.md'
```
devsaurus: I can't find this file in the commit.

TerryE: That's  because I forgot to add it to the commit :astonished:

app/lua/lflash.c
```diff
@@ -270,5 +298,261 @@ LUAI_FUNC int luaN_index (lua_State *L) {
...
+  if (out->ndx <= WRITE_BLOCKSIZE) {
+    uint32_t fl;
+    /* Process the flash header and cache the FlashHeader fields we need */
+    FlashHeader *fh = cast(FlashHeader *, out->block[0]);
+    out->flashLen   = fh->flash_size;                         /* in bytes */
+    out->flagsLen   = (out->len-fh->flash_size)/WORDSIZE;     /* in words */
+    out->flash_sig  = fh->flash_sig;
+
+    if ((fh->flash_sig & ~FLASH_SIG_ABSOLUTE) != FLASH_SIG)
+      flash_error("incorrect LFS header signature");
```
HHHartmann: I would really like to have different messages here for absolute addresses and integer/float missmatch and also stating this explicitly.

app/lua/luac_cross/luac.c
```diff
@@ -72,6 +73,7 @@ static void usage(const char* message)
  "  -f       output a flash image file\n"
  "  -a addr  generate an absolute, rather than position independent flash image file\n"
  "  -i       generate lookup combination master (default with option -f)\n" 
+ "  -m size  maximum LFS image in bytes\n"
  "  -p       parse only\n"
  "  -s       strip debug information\n"
  "  -v       show version information\n"
```
HHHartmann: Would it be possible to add integer/float information to the version command
Maybe in general stating that this is no normal luac.cross but LFS (and other) enhanced?

docs/en/upload.md
```diff
@@ -39,7 +39,7 @@ Supported platforms: macOS, Linux Windows, anything that runs Node.js
...
 
-Be very careful not to lock yourself out! If there's a bug in your `init.lua` you may be stuck in an infinite reboot loop. It is, therefore, advisable to build a small delay into your startup sequence that would allow you to interrupt the sequence by e.g. deleting or renaming `init.lua` (see also [FAQ](lua-developer-faq.md#how-do-i-avoid-a-panic-loop-in-initlua)). Your `init.lua` is most likely going to be different than the one below but it's a good starting point for customizations:
+Be very careful not to lock yourself out! If there's a bug in your `init.lua` you may be stuck in an infinite reboot loop. It is, therefore, advisable to build a small delay into your startup sequence that would allow you to interrupt the sequence by e.g. deleting or renaming `init.lua` (see also [FAQ](lua-developer-faq.md#how-do-i-avoid-a-panic-loop-in-initlua)). Your `init.lua` is most likely going to be different than the one below but it's a good starting point for customizations, unless you are using LFS in which case most of this could could be moved to LFS:
```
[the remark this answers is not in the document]
TerryE: Sorry Gregor, I banged this in after @marcelstoer pointed out that we didn't have such documentation.  At least it is at least a baseline for review and refinement. :smiley:

app/lua/lflash.c
TerryE: My urgency was to get an error returned.  You're right: maybe we should split these cases down.  Another change that I thought of making was to up and internal count in the flash_sig every time any format change is made to the LFS internal format.  Yes, it should be obvious that you use the `luac.cross` from the same build as the firmware, but I can see this as a bearpit that developers might fall into.

app/lua/luac_cross/luac.c
TerryE: Yes, we should say that this is a nodeMCU build with integer/float LFS/non-LFS support.

app/lua/lflash.c
HHHartmann: Yes, especially when there will be an option to download prebuilt luac.cross binaries.
Maybe the version command then should also tell the user what (internal) version it is building.
And the node.flashreload() also of course.

TerryE:
Please note that I've also done a non-trivial rework of the LFS paper in this update.  I would like to merge it in in the next few days, so if anyone has any comments then please feed them in sooner than later.  Thanks.

app/include/user_modules.h
```diff
@@ -26,7 +26,7 @@
 //#define LUA_USE_MODULES_ENCODER
 //#define LUA_USE_MODULES_ENDUSER_SETUP // USE_DNS in dhcpserver.h needs to be enabled for this module to work.
 #define LUA_USE_MODULES_FILE
-//#define LUA_USE_MODULES_GDBSTUB
+#define LUA_USE_MODULES_GDBSTUB
```
TerryE: This was added in error.  I will back it out before the merge.

app/modules/file.c
```diff
@@ -214,18 +214,53 @@ static int file_open( lua_State* L )
 static int file_list( lua_State* L )
 {
```
TerryE: Oops! another cock-up.  Now you know how I tested @nwf's file.list() patch!!!

docs/en/lfs.md
```diff
@@ -29,76 +56,33 @@ A nice example is to run Telnet and FTP servers from LFS. In order to do this pu
...
   
-Add the following lines:
-```lua
--- Execute the LFS init
-node.flashindex("_init")()
--- Start Telnet server
-require("telnet"):open()
--- Start FTP serer
-require("ftpserver").createServer('user', 'password')
+Once the LFS image file is on SPIFFS, you and execute the [node.flashreload()](/en/dev/en/modules/node/#nodeflashreload) command to load it into the LFS region.  If the file is invalid this call will return an error string, but it will normally not return, but instead immediately restart the ESP module with the new LFS loaded.
```
HHHartmann: you **can** execute

TerryE:
@marcelstoer. Marcel,  forgive me for being dumb, but I am assuming by "next release" that we are deferring merging this upgrade until _after_ the next cut from dev to master.  I think that this makes sense, since the addition of compression is quite a big change and I'd rather see this in `dev` for a full cycle.  I am also going to have to rebaseline my PR and do at least one more push.  This is going to be the easiest way to sort out the conflicts etc.

marcelstoer:
@TerryE I'm not sure if this is a misunderstanding or not, let me try to explain.

According to our 2-months release cycle we should have shipped at the end of July. That's why our [current release milestone "next release"](https://github.com/nodemcu/nodemcu-firmware/milestone/10) is past due by 12 days.

While LFS has been exposed to the (interested) public on `dev` long enough to be ready for prime I did not create one yet due to  #2431. As I stated there I think we are ready as soon as we 
- document how to build `luac.cross` and optionally
- can make the binaries available with every release (my proposal is lingering somewhere)

I initially expected you to wanting the compression shipped with the same release since it's delayed anyway (thus the assignment to "next release"). However, I would also prefer to ship sooner rather than later - even w/o compression.
-> I just removed it from the milestone

TerryE:
I would have thought that the two month cycle is indicative / aspirational and was set when the other committers were a lot more active.  At the moment, and other than my work, we are pretty quiet, so I see no harm in slipping this out to 3 months -- so long as it remains pretty regular.  However, if we are continuing with the guillotine cut method of moving `dev` to master then I think that we would agree that any changes which have a chance of impacting the stability of the firmware should be in `dev` for at least a month before being moved to `master`.   So changes to documentation or even the odd tweak to `luac.cross -h` text are fine but the luac.cross image compression, etc is not.

Most PRs are quite small in scope.  This one and the last luac.cross one are both as a single change bigger than the all other PRs over the last 6-12 months combined, and it is very difficult to decompose them into smaller sub-PRs.  So what I suggest is that I put this one on hold for the next week and focus on the smaller tidy-ups that we can commit and tag as next-release such as documentation.

I also see no problem in adding the luac.cross.elf and luac.cross.cygwin (that's the float builds) to the tools dir to make it easier for new developers (as these are only 100Kb & 150 Kb when zipped).

marcelstoer:
> I would have thought that the two month cycle is indicative / aspirational

It is; we don't ow it to anyone. However, we were confident enough talk about it right on our repo front page at https://github.com/nodemcu/nodemcu-firmware#releases.

> so I see no harm in slipping this out to 3 months

I wouldn't mind either. It's just that we haven't been talking about this so far.

All committers can edit the milestone(s) I create after every new master snap. So far I've been the only one actually assigning PRs to milestones. Feel free to add or remove PRs as you see fit. If we don't do this our minimal release planning slips back to the uncontrolled chaos we had three years ago.

TerryE:
> So far I've been the only one actually assigning PRs to milestones

Yup, not disputing this, but rather that you are doing this so efficiently, that IMO there's point in anythig other than letting you continue to take the lead here unless you want otherwise :smile:

marcelstoer:
I don't mind but I expect that whoever merges a PR sets the current milestone if not done beforehand while planning. I don't see why I should keep running after merged PRs to ensure they belong to a milestone. Without that PR-to-milestone correlation it's so much more time consuming to write effective release notes (I only care because it's _my_ time 😉). Over'n out -> back on topic.

TerryE:
Yup this PR is definitely outside "next release", but I will do a "lightwieght" PR for the bits (mainly documentation) to go into it tomorrow.

TerryE:
We've had the documentation bits split out and further enhanced on other PRs, as well as #2472.  

IMO, it makes some sense to rebaseline these changes and force push them here to resync this set against the current dev update, for a clean push into dev.  A job for tomorrow.

TerryE:
I have had a look at this merged version and it seems to be OK, so please feel free to pull my fork and try it out.  

The issue that I was having was because my `luac.cross` was also pathed into an old version and the reload didn't pick this up, so was forcd down a wierd error path.   I will need to fix this version detection before we merge into `dev`, but other than this it seems to work fine.

Issue #2443, opened 2018-07-30
trouble with enabling LUA_FLASH_STORE


mikep01:
8<------------------------ BUG REPORT -----------------------------------------
### Expected behavior
Proper startup calling init.lua

### Actual behavior
after a reset I see my own messages at 9600bps then the CPU appears to hang.  Changing to 115000 and another reset I see:

2018.07.29 21:03:23.948 lua: init.lua:1: unexpected symbol near '¼'
2018.07.29 21:03:23.891  powered by Lua 5.1.4 on SDK 2.2.1(6ab97e9)
2018.07.29 21:03:23.871 2.2.0.0 build created on 2018-07-29 04:28
2018.07.29 21:03:23.867 	SSL: false
2018.07.29 21:03:23.865 	commit: c6f6c54bc82e5868a45ebc44f73bec9674837c65
2018.07.29 21:03:23.862 	branch: dev
2018.07.29 21:03:23.859 NodeMCU  built with Docker provided by frightanic.com
2018.07.29 21:03:23.856 
2018.07.29 21:03:23.848 Flash sig not correct: 0 vs fafaaf52
2018.07.29 21:03:23.748 s$ ==> this is probably my messages from 9600bps...

### Test code
Works fine when built with LUA_FLASH_STORE not defined, calling init.lua properly.
When built with it defined I see the above messages.

### NodeMCU version
NodeMCU 2.2 but I've back ported the tmr functions (as per an earlier 

### Hardware
NodeMCU v0.9, nothing special.
8<------------------------ END BUG REPORT -------------------------------------

mikep01:
Enabling (defining) LUA_INIT_STRING seems to help, now things run to the point of showing the interactive prompt at my defined 9600bps.  The docs gave me the impression this was optional?  Running resume("init") fails:

2018.07.29 21:15:48.701 	stdin:1: in main chunk
2018.07.29 21:15:48.700 	[C]: in function 'require'
2018.07.29 21:15:48.680 	[C]: ?
2018.07.29 21:15:48.646 stack traceback:
2018.07.29 21:15:48.613 	init.lua:1: unexpected symbol near 'é'
2018.07.29 21:15:48.580 error loading module 'init' from file 'init.lua':
2018.07.29 21:15:48.299 > require("init")
2018.07.29 21:15:48.278 require("init")

[...]

PR #2449, opened 2018-08-04
Luac.cross.{integer/float}

HHHartmann:

As discusse in issue #2443 I changed the makefiles to create luac.cross.integer and luac.cross.float depending on the defined build type as defined for docker builds (and others)

I also added a .gitignore file to local/lua dictionry so that it appears on chekout/clone as local/fs does.


HHHartmann:
sorry, still not understanding git

Issue #2440, opened 2018-07-27
file_list enhancement


nedoskiv:
Hello 
I'm not familiar with github, but love nodemcu lua project, use it to make an RFID system, but that is out of the topic, point is that I need to handle a list of large number of files. There is nothing wrong with it unless you want to see list of all files. file_list function fails with out of memory error when file numbers goes above 280 (depends of file name length). So first I asked someone to make possible to get the list in other way, guess not posted it right. So I believe made it myself. Here is little modification of file_list function I made:

```
// Ven version of list()
static int file_list( lua_State* L )
{
  unsigned st = luaL_optinteger( L, 1, 1 );		// start offset
  unsigned tf = luaL_optinteger( L, 2, 100000 );	// how much files to list
  tf=tf+st;
  vfs_dir  *dir;

  if (dir = vfs_opendir("")) {
    lua_newtable( L );
    struct vfs_stat stat;
	int i=1;
    while (vfs_readdir(dir, &stat) == VFS_RES_OK) {
		if (i<st)
			{
			i++;
			continue;
			}
		if (i>=tf)
			{
			break;
			}
      lua_pushinteger(L, stat.size);
      lua_setfield(L, -2, stat.name);
	  i++;
    }
    vfs_closedir(dir);
    return 1;
  }
  return 0;
}
```

now it work as usual when called without arguments (list files from 1 to 100000), but if the function is called with an arguments - file_list(st,n) it gonna return    N number of  files, starting from st.
That way i can use it to retrieve file list on portion and avoid  out of memory errors&restart

I do not know how to post that for integration nodemcu, Also  my C knowledge is limited, may be someone can do it better. I do it my way, test it and it is working fine for me.

At the end, please someone answer to me - how many number of files SPIFFS nodemcu file system can handle?

TerryE:
Hi, first some simple answers.
> how many ... SPIFFS ...

The github repo is [here](/pellepl/spiffs); there is a FAQ and click on the wiki link for other SPIFFS documentation including a FAQ. 

[...]

PR #2452, opened 2018-08-09
file: list now takes optional pattern for filtering

nwf:
See #2440.

This adds an optional pattern to `file.list()` to filter results before handing them to Lua.  This optimizes the relatively common pattern of
```lua
for f,s in pairs(file.list()) do
  if s:match(pattern) then
    -- process matching file name
  end
end
```
into the more concise, faster, and less-memory-consuming
```lua
for f,s in pairs (file.list(pattern)) do
  -- process each matching file name
end
```
(neglecting the possibility that pattern might raise an error.  If that happens, `file.list(pattern)` will now return `nil, errorstr`, as typical).

This has been tested by running comprehensive test cases:
```
for k,v in pairs(file.list()) do print(k,v) end -- all files
for k,v in pairs(file.list("asdf")) do print(k,v) end -- pattern with no matches
for k,v in pairs(file.list("ini")) do print(k,v) end -- pattern some matches
for k,v in pairs(file.list({})) do print(k,v) end -- wrong type passed as argument
_, y = file.list("%fx"); print (y) -- pattern error passed up to caller
```

### DIFF
--- a/app/modules/file.c
+++ b/app/modules/file.c
@@ -214,18 +214,53 @@ static int file_open( lua_State* L )
 static int file_list( lua_State* L )
 {
   vfs_dir  *dir;
+  const char *pattern;
+  struct vfs_stat stat;
+  int pcres;
 
-  if (dir = vfs_opendir("")) {
-    lua_newtable( L );
-    struct vfs_stat stat;
-    while (vfs_readdir(dir, &stat) == VFS_RES_OK) {
-      lua_pushinteger(L, stat.size);
-      lua_setfield(L, -2, stat.name);
-    }
-    vfs_closedir(dir);
-    return 1;
+  lua_settop(L, 1);
+  pattern = luaL_optstring(L, 1, NULL);   /* Pattern (arg) or nil (not) at 1 */
+
+  dir = vfs_opendir("");
+  if (dir == NULL) {
+    return 0;
   }
-  return 0;
+
+  lua_newtable( L );                      /* Table at 2 */
+
+  if (pattern) {
+    /* 
+     * We know that pattern is a string, and so the "match" method will always
+     * exist.  No need to check return value here
+     */
+    luaL_getmetafield( L, 1, "match" );  /* Function at 3 */
+  }
+
+  while (vfs_readdir(dir, &stat) == VFS_RES_OK) {
+    if (pattern) {
+      lua_settop( L, 3 );                 /* Ensure nothing else on stack */
+
+      /* Construct and pcall(string.match,name,pattern) */
+      lua_pushvalue( L, 3 );
+      lua_pushstring( L, stat.name );
+      lua_pushvalue( L, 1 );
+      pcres = lua_pcall( L, 2, 1, 0 );
+      if (pcres != 0) {
+        vfs_closedir(dir);
+        lua_error( L );
+      }
+      if (lua_isnil( L, -1 )) {
+        continue;
+      }
+    }
+    lua_pushinteger( L, stat.size );
+    lua_setfield( L, 2, stat.name );
+  }
+
+  /* Shed everything back to Table */
+  lua_settop( L, 2 );
+  vfs_closedir(dir);
+  return 1;
 }
 
 static int get_file_obj( lua_State *L, int *argpos )
--- a/docs/en/modules/file.md
+++ b/docs/en/modules/file.md
@@ -144,13 +144,17 @@ print("\nFile system info:\nTotal : "..total.." (k)Bytes\nUsed : "..used.." (k)B
 Lists all files in the file system.
 
 #### Syntax
-`file.list()`
+`file.list([pattern])`
 
 #### Parameters
 none
 
 #### Returns
-a Lua table which contains the {file name: file size} pairs
+a Lua table which contains all {file name: file size} pairs, if no pattern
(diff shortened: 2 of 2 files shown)


docs/en/modules/file.md
```diff
@@ -144,13 +144,17 @@ print("\nFile system info:\nTotal : "..total.." (k)Bytes\nUsed : "..used.." (k)B
...
+`file.list([pattern])`
 
 #### Parameters
 none
 
 #### Returns
-a Lua table which contains the {file name: file size} pairs
+a Lua table which contains all {file name: file size} pairs, if no pattern
+given.  If a pattern is given, only those file names matching the pattern
+(interpreted as a traditional Lua pattern, not, say, a UNIX shell glob) will be
```
marcelstoer: Linking to https://www.lua.org/pil/20.2.html or http://lua-users.org/wiki/PatternsTutorial here?

nwf: Sure. :+1:

app/modules/file.c
```diff
@@ -214,18 +214,71 @@ static int file_open( lua_State* L )
...
-      lua_setfield(L, -2, stat.name);
+      /* Check and pop Result, restoring stack to T P F */
+      if (!lua_isnil( L, -1 )) {
+        lua_pushinteger( L, stat.size );
+        lua_setfield( L, -5, stat.name );
+      }
+      lua_pop( L, 1 );
+    } else {
+      /* Stack just holds result table at the top; add result */
+      lua_pushinteger( L, stat.size );
```
TerryE: these to push/setfield calls can be folded easily

nwf: I debated doing that.  I can, if you like.

app/modules/file.c
```diff
@@ -214,18 +214,71 @@ static int file_open( lua_State* L )
...
+    return 0;
+  }
+
+  /* Allocate result table T */
+  lua_newtable( L );
+
+  /* If we're doing a match, push the Pattern and Function */
+  if (pattern != emptystr) {
+    lua_pushstring( L, pattern );
+    /*
```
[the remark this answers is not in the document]
nwf: Fair enough.  I suppose `luaL_checkoptstring` just returns what's already on the stack.

app/modules/file.c
```diff
@@ -214,18 +214,71 @@ static int file_open( lua_State* L )
...
+  }
+
+  while (vfs_readdir(dir, &stat) == VFS_RES_OK) {
+    if (pattern != emptystr) {
+      /* Stack: T P F .  Duplicate F, push Name, duplicate P */
+      lua_pushvalue( L, -1 );
+      lua_pushstring( L, stat.name );
+      lua_pushvalue( L, -4 );
+      /* Call transforms stack from T P F F N P to T P F Result */
+      pcres = lua_pcall( L, 2, 1, 0 );
```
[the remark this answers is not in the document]
nwf: No, that's not sufficient.  Lua will `longjmp` over our activation frame and not give us the opportunity to call `vfs_closedir`, which might imply a leak of a directory iterator.  If not in SPIFFS or FATFS, some day, all the same.

nwf:
@TerryE Thanks for that.  I reworked mine in light of yours and re-ran the tests I enumerated above.

TerryE:
@nwf, it takes a bit to get your head around using the API, but once you do, it is really quite potent.  The more of us the merrier.  

Are you using the gdb stub?  It's really quite good. This is one time that I use esplorer, because I am cycling reboot/tweak and start script in explorer and then toogle into the xtensa gdb.  Also note the `.gdbinit` and `.gdbinitlua` in the `dev` branch  which contain useful macros.

app/modules/file.c
```diff
@@ -214,18 +214,60 @@ static int file_open( lua_State* L )
...
-      lua_setfield(L, -2, stat.name);
+  while (vfs_readdir(dir, &stat) == VFS_RES_OK) {
+    if (pattern) {
+      lua_settop( L, 3 );                 /* Ensure nothing else on stack */
+
+      /* Construct and pcall(string.match,name,pattern) */
+      lua_pushvalue( L, 3 );
+      lua_pushstring( L, stat.name );
+      lua_pushvalue( L, 1 );
+      pcres = lua_pcall( L, 2, 1, 0 );
```
TerryE: This is really the wrong thing to do.  You should just use a `lua_call()` and let the lua error system do its stuff.  So for example if you do something like:
 ```Lua
    for name, len in pairs(files.list('[a-f]%')) do`
```
you want it  to throw a `malformed pattern (ends with '%')` rather that a `bad argument #1 to 'pairs' (table expected, got number)`.  9 times out of 10 the pattern is programmed and on the 10th time, the programmer can always wrap the `file.list(filter)` in a `pcall()`

nwf: As I said above, I do not believe that `longjump`-ing out of `file.list()` is correct, as that will not call `vfs_closedir()` on its way out.  I'll rework the patch to use `lua_error` to re-throw the exception.

TerryE:
Sorry, missed that. In which case you don't need the goto. Just do the closedir and throw the error inline in the pcall status test code.

nwf:
@TerryE The goto's gone; mind taking another look?

PR #2453, opened 2018-08-10
app/lua/luac_cross: WRAPCC CC here, too

nwf:
Not that lua takes very long to compile, but allowing for ccache here
should still help the builder(s) consume a little less time and power.


HHHartmann:
Good idea. Maybe you could also change tools/spiffsimg while you are at it.
Any other places?

PR #2454, opened 2018-08-10
LFS: small tweaks to docs and examples

nwf:
A few typos fixed here and there, some rewording, and tweaked examples
of package.loader manipulation.

### DIFF
--- a/docs/en/lfs.md
+++ b/docs/en/lfs.md
@@ -22,7 +22,7 @@ For details see section [Selecting the firmware](#selecting-the-firmware) of the
 
 The easiest way is to maintain Lua files of your project in its own directory tree on your host. Project files will be compiled by `luac.cross` to build the LFS image in next step.
     
-A nice example is to run Telnet and FTP servers from LFS. In order to do this put the following files in 1 directory:
+A nice example is to run Telnet and FTP servers from LFS. In order to do this put the following files in one directory:
 	
 * [lua_examples/lfs/_init.lua](../../lua_examples/lfs/_init.lua)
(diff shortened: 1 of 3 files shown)


TerryE:
Nathaniel, thanks very much for these tweaks and corrections, however IMO the changes to the use of `package.loaders` aren't how I would do this, so I would have preferred to discuss these before committing them, but since  am doing some updates / fixes to the LFS paper myself in my #2448, we should at least reach consensus on how to do this.  The code for the 4 package.loaders is based on standard Lua use, however in the case of NodeMCU, we only use second, the Lua loader.  The 3rd and 4th are C (library) loaders, but since these ultimately use `ll_load()` and `ll_sym()` which just return `"dynamic libraries not enabled"` and so fail.  The 1st entry is for preloading packages (such as those statically linked like our modules, but Zeroday's initial approach bypasses this and instead modifies the core require code bypassing the requre if the name is that of a ROTable.  So slots 1,3 and 4 are free for reuse.  Slot 2 is always the Lua Loader, so inserting the LFS loader at slot 2 or 5 is _not_ what you should do.

nwf:
Well, perhaps we should just not install, from C, the things currently residing at indexes 1, 3, and 4, since they aren't of utility?  I had thought that `table.insert(package.loaders, 2, myfn)` would install `myfn` between `1` and `2`, which would have the effect of superseding the Lua loader?  If the thing at index 1 were not doomed to failure, this might be the desired behavior?

TerryE:
@nwf, what  you propose would work and it does move 2 up to 3 and put the LFS loader at 2.  It's not a bug per se, but neither was the original.   As you suggest in, some ways it would just be simpler to have C do the equivalent of `package.loaders={Lua_loader}`.

PR #2455, opened 2018-08-10
lua_examples/telnet changes

nwf:
Soliciting thoughts on these.


nwf:
I dropped the second, more contentious patch, so this PR is now just some tidying.  My hope was to replace my telnet server with the example, but instead I factored the example's double-fifo contribution out to its own module for my reuse (https://github.com/AcmeTensorToys/esp-lua_core/blob/master/net/fifosock.lua).

For whatever it's worth, I've had no trouble running multiple clients against my somewhat more generic `telnetd` (https://github.com/AcmeTensorToys/esp-lua_core/blob/master/telnetd/telnetd.lua) and have just tested having multiple listening `telnetd`s on multiple ports.  It all seems to go fine.

PR #2458, opened 2018-08-14
LFS documentation and example updates

TerryE:
See #2431.

Updates to LFS documentation, new LFS OTA example and backing out one unreviewed change to LFS _init file.

### DIFF
--- a/docs/en/lfs.md
+++ b/docs/en/lfs.md
@@ -2,19 +2,15 @@
 
 ## Background
 
-An IoT device such as the ESP8266 has very different processor characteristics from the CPU in a typical PC. ESP modules are postage stamp-sized and typically comprise one ESP [SoC](https://en.wikipedia.org/wiki/System_on_a_chip) and a flash memory chip used to store firmware and a limited file system.  Most IoT CPUs inluding the ESP variants use a [modified Harvard architecture](https://en.wikipedia.org/wiki/Modified_Harvard_architecture) where code can also be executed out of flash memory in a memory address region that is separate from the limited RAM.  Conventional processors typically more than 1 Gb RAM, that is used to store both code and data and use a motherboard which includes the RAM, a lot of support chips and supporting peripherals. 
+Lua was originally designed as a general purpose embedded extension language for use in applications run on a conventional computer such as a PC, where the processor is mounted on a motherboard together with multiple Gb of RAM and a lot of other chips providing CPU and I/O support to connect to other devices.
 
-Lua was originally designed as a general embedded extension language for applications that would typically run on conventional processors, such as PCs and the like, but its design goals of speed, portability, small kernel size, extensibility and ease-of-use also make Lua a good choice for embedded use on an IoT platform.  The NodeMCU firmware implementation was therefore constrained by the standard Lua core runtime system (**RTS**) that assumes a conventional CPU architecture with both Lua code and data in RAM, but ESP8266 modules only have some 48Kb RAM available for application use.
+ESP8266 modules are on a very different scale: they cost a few dollars; they are postage stamp-sized and only mount two main components, an ESP [SoC](https://en.wikipedia.org/wiki/System_on_a_chip) and a flash memory chip.  The SoC includes limited on-chip RAM, but also provides hardware support to map part of the external flash memory into a separate memory address region so that firmware can be executed directly out of this flash memory &mdash; a type of [modified Harvard architecture](https://en.wikipedia.org/wiki/Modified_Harvard_architecture) found on many [IoT](https://en.wikipedia.org/wiki/Internet_of_things) devices.  Even so, Lua's design goals of speed, portability, small kernel size, extensibility and ease-of-use have made it a good choice for embedded use on an IoT platform, but with one major limitaton: the standard Lua core runtime system (**RTS**) assumes that both Lua data _and_ code are stored in RAM; this isn't a material constraint with a conventional computer, but it can be if your system only has some 48Kb RAM available for application use.
 
-The Lua Flash Store (**LFS**) patch modifies the NodeMCU Lua RTS to allow Lua code and its associated constant data to be executed directly out of flash-memory, just as the firmware is itself executed. This enables NodeMCU Lua developers to create Lua applications with up to 256Kb Lua code and read-only (**RO**) constants executing out of flash, so that all of the RAM is available for read-write (**RW**) data.
+The Lua Flash Store (**LFS**) patch modifies the Lua RTS to support a modified Harvard architecture by allowing the Lua code and its associated constant data to be executed directly out of flash-memory (just as the NoceMCU firmware is itself executed). This now allows NodeMCU Lua developers to create Lua applications with up to 256Kb Lua code and read-only (**RO**) constants executing out of flash, with all of the RAM is available for read-write (**RW**) data.
 
-The ESP architecture does not allow normal write operations to flash memory, as writing to NAND flash memory involves block erasing complete 4Kb pages, and then overwriting each erased page with new content.  Whilst it is possible to develop R/W file systems within this constraint (such as SPIFFS), this makes it practically impossible to modify executable code pages on the fly. Hence the LFS patch must work within a reflash-and-restart paradigm for reloading the LFS, and it does this by adding two API new calls to the `node` module: one to reflash the LFS and restart the processor, and one to access the LFS store once loaded.  Under the hood, it also addresses all of the technical issues to make this magic happen.
+Unfortunately, the ESP architecture provides very restricted write operations to flash memory (writing to NAND flash involves bulk erasing complete 4Kb memory pages, before overwriting each erased page with any new content).  Whilst it is possible to develop a R/W file system within this constraint (as SPIFFS demonstrates), this makes impractical to modify Lua code pages on the fly. Hence the LFS patch works within a reflash-and-restart paradigm for reloading the LFS, and does this by adding two API new calls: one to reflash the LFS and restart the processor, and one to access LFS stored functions.  The patch also addresses all of the technical issues 'under the hood' to make this magic happen.
 
-The remainder of this paper is split into two parts:
-
--  The first provides an overview for Lua developer wanting to use LFS effectively at an application programming level. As most of our developers use a Windows platform, it first gives a quick start for such developers before covering some of application issues in more detail.
-
--  The second part is for those that want to understand a little of how this magic happens, and gives more details on the technical issues that were addressed in order to implement the patch.
+The remainder of this paper is split into two parts. The first provides an overview for Lua developers wanting to use LFS effectively at an application programming level, and as most of our developers use a Windows platform, it gives a quick start for these developers before covering some of application issues in more detail.  The second part is for those who want to understand a little of how this magic happens, and gives more details on the technical issues that were addressed in order to implement the patch.
 
 
 ## Using LFS
@@ -25,14 +21,12 @@ This is a simple step-by-step guide for ESP8266 Lua developers who want to start
 
 #### Get and flash LFS enabled firmware
 
-Use the [NodeMCU Build](https://nodemcu-build.com/) service to build the firmware. Selecting the `dev` branch enables the latest LFS options.  In the "LFS options" section choose the size of the LFS partition (64 KB should be fine). The SPIFFS default settings will be fine.  See section [selecting the firmware](#selecting-the-firmware) for further details.  
-
-Once you have received the build confirmation email and downloaded the flash image, you can then [flash the firmware](flash.md) to your ESP8266.
+Use the [NodeMCU Build](https://nodemcu-build.com/) service to build the firmware.  In the "LFS options" section choose the size of the LFS partition (64 KB should be fine). The SPIFFS default settings will be fine.  See section [selecting the firmware](#selecting-the-firmware) for further details.  Once you have received the build confirmation email and downloaded the flash image, you can then [flash the firmware](flash.md) to your ESP8266.
 
 #### Build or download your local copy of `luac.cross`
 
 Windows 10 users can install and use the Windows Subsystem for Linux (WSL).  Alternatively all Windows users can [install Cygwin](https://www.cygwin.com/install.html). (You will only need the Cygwin core). Either way, you will need a copy of the `luac.cross` compiler:
--  You can either download this from my fileserver.  The [ELF variant](http://files.ellisons.org.uk/esp8266/luac.cross) is used for all Linux and WSL flavours, or the [cygwin binary](http://files.ellisons.org.uk/esp8266/luac.cross.cygwin)) for the Cygwin environment.
+-  You can either download this from my fileserver.  The [ELF variant](http://files.ellisons.org.uk/esp8266/luac.cross) is used for all recent Linux and WSL flavours, or the [cygwin binary](http://files.ellisons.org.uk/esp8266/luac.cross.cygwin)) for the Cygwin environment.
 -  Or you can compile it yourself by downloading the current NodeMCU sources (this [ZIPfile](https://github.com/nodemcu/nodemcu-firmware/archive/master.zip)); edit the `app/includes/user_config.h` file and the `cd` to the `app/lua/luac_cross` and run make to build the compiler in the NodeMCU firmware root directory.  Note that the `luac.cross` make only needs the host toolchain which is installed by default in WSL, but in Cygwin you will need to tick the _gcc-core_ + _gnu make_ options during setup.
 
 #### Select Lua files to be run from LFS
@@ -46,7 +40,7 @@ For example to run the Telnet and FTP servers from LFS, put the following files 
 * [lua_examples/telnet/telnet.lua](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_examples/telnet/telnet.lua).  A simple **telnet** server.
 * [lua_modules/ftp/ftpserver.lua](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_modules/ftp/ftpserver.lua).  A simple **FTP** server.
 
-You should always include the first two modules, but the remaining files would normally be replaced by your own project files.
+You should always include the first two modules, but the remaining files would normally be replaced by your own project files.  Also remember that these are examples and that you are entirely free to modify or to replace them for your own application needs.
 
 #### Compile the LFS image
 
@@ -62,9 +56,9 @@ You might also want to add a simple one-line script file to your ~/bin directory
 
 #### Upload the LFS image and flash it to the LFS region
 
-Now upload the generate LFS image file (`lfs.img` in this case) as a normal SPIFFS file. Several tools are available to upload the files from host to SPIFFS, including [ESPlorer](https://github.com/4refr0nt/ESPlorer).  There is also a new example in `lua_examples` which can retrieve images from a standard web service.
+Now upload the compiled LFS image file (`lfs.img` in this case) as a normal SPIFFS file. Several tools are available to upload the files from host to SPIFFS, including [ESPlorer](https://github.com/4refr0nt/ESPlorer).  There is also a new example, [HTTP_OTA.lua](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_examples/lfs/HTTP_OTA.lua), in `lua_examples` that can retrieve images from a standard web service.
   
-Once the LFS image file is on SPIFFS, you can execute the [node.flashreload()](/en/dev/en/modules/node/#nodeflashreload) command and if the image file is valid, the loader will then load it into flash and immediately restart the ESP module with the new LFS loaded.  However, the call will return with an error _if_ the file is found to be invalid.
+Once the LFS image file is on SPIFFS, you can execute the [node.flashreload()](/en/dev/en/modules/node/#nodeflashreload) command and the loader will then load it into flash and immediately restart the ESP module with the new LFS loaded, if the image file is valid.  However, the call will return with an error _if_ the file is found to be invalid, so your reflash code should include logic to handle such an error return.
 
 #### Edit your `init.lua` file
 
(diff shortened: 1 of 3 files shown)


TerryE:
@HHHartmann, Gregor you might want to review this as well.

docs/en/lfs.md
```diff
@@ -1,205 +1,144 @@
...
 
--  Includes a 1 sec delay before connecting to the Wifi.  This is a "just in case" when I am developing.  This is enough to allow me to paste a `file.remove'init.lua'` into the UART if I want to do different development paths.
+Under rare circumstances, for example a power fail during the flashing process, the flash can be left in a part-written state following a `flashreload()`.  The Lua RTS start-up sequence will detect this and take the failsafe opton of resetting the LFS to empty, and if this happens then the LFS `_init` function will be unavailable.  Your `init.lua` should therefore not assume that the LFS contains any modules (such as `_init`), and should contain logic to detect if LFS reset has occurred and if necessary reload the LFS again. Calling `node.flashindex("_init")()` directly will result in a panic loop in these circumstances.  Therefore first check that `node.flashindex("_init")` returns a function or protect the call, `pcall(node.flashindex("_init"))`, and decode the error status to validate that initialisation was successful.
```
nwf: This is pretty far afield and NOT a blocker for the upcoming release, but I wanted to get the thought in your head: any possibility of a dual-slot LFS variant, where `node.flashreload()` can be told which slot to pave over and `node.flashindex()` either always uses the newest or always searches both?  This would give a pretty nice "application LFS" vs. "failsafe LFS" setup and permit an OTA dance where only one of the two was ever potentially invalid at any moment.

docs/en/lfs.md
```diff
@@ -1,205 +1,144 @@
...
-    return ba and "Module not in LFS" or fn
-  end
-  table.insert(package.loaders, loader_flash)
-
-end
-```
+See [lua_examples/lfs/_init.lua](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_examples/lfs/_init.lua) for the code that I use in my `_init` module to do create a simple access API for LFS.  There are two parts to this.
 
-This code is also conveniently packaged in [lua_examples/lfs/_init.lua](../../lua_examples/lfs/_init.lua).
+The first sets up an LFS table in `_G` with the `__index` and `__newindex` metamethods. The main purpose of the `__index()` is to resolve any names against the LFS using a `node.flashindex()` call, so that `LFS.someFunc(params)` does exactly what you would expect it to do: this will call `someFunc` with the specified parameters, if it exists in in the LFS.  The LFS properties `_time`, `_config` and `_list` can be used to access the other LFS metadata that you need.  See the code to understand what they do, but `LFS._list` is the array of all module names in the LFS.
```
nwf: "an `LFS` table in `_G`" ?

docs/en/lfs.md
```diff
@@ -1,205 +1,144 @@
...
 
-Of course, if you use Lua modules to build your application then `require "some_module"` will automatically path in and load your modules from LFS.  Note that SPIFFS is still ahead of LFS in this search list, so if you have a dev version in SPIFFS, say, then this will be loaded first. However, if you want to want to swap this search order so that the LFS is searched first, then run `table.insert(package.loaders,2,loader_flash)` in your `_init` code.  If you need to swap the search order temporarily for development or debugging, then do this after you've run the `_init` code:
+If you want to swap this search order so that the LFS is searched first, then SET `package.loaders[1] = loader_flash` in your `_init` code.  If you need to swap the search order temporarily for development or debugging, then do this after you've run the `_init` code:
 
 ```Lua
-do local pl = package.loaders; pl[2],pl[5] = pl[5],pl[2]; end
+do local pl = package.loaders; pl[1],pl[3] = pl[3],pl[1]; end
 ```
```
nwf: I remain somewhat unhappy about exposing these particular details to programmers, but for the moment it's fine, I suppose.

docs/en/lfs.md
```diff
@@ -1,205 +1,144 @@
...
 
-(The indexes 2 and 5 come from the value of the `loaders` array in [app/lua/loadlib.c](../../app/lua/loadlib.c)).
```
nwf: Maybe keep this explanation around?  Though the constants need updating.

docs/en/lfs.md
```diff
@@ -208,8 +147,12 @@ do
...
+
+You can then create a file, say `LFS_dummy_strings.lua`, and insert these `local preload` lines into it.  By including this file in your `luac.cross` compile, then the cross compiler will also include all strings referenced in this dummy module in the generated ROM string table.  Note that you don''t need to call this module; it's inclusion in the LFS build is enough to add the strings to the ROM table. Once in the ROM table, then you can use them subsequently in your application without incurring any RAM or LGC overhead.  A useful starting point may be found in [lua_examples/lfs/dummy_strings.lua](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_examples/lfs/dummy_strings.lua); this saves about 4Kb of RAM by moving a lot of common compiler and Lua VM strings into ROM.
+
+Another good use of this technique is when you have resources such as CSS, HTML and JS fragments that you want to output over the internet.  Instead of having lots of small resource files, you can just use string assignments in an LFS module and this will keep these constants in LFS instead.
+
```
nwf: Presumably these should not be `local` in LFS?

docs/en/lfs.md
```diff
@@ -208,8 +147,12 @@ do
...
-A useful starting point may be found in [lua_examples/lfs/dummy_strings.lua](../../lua_examples/lfs/dummy_strings.lua).
+You can do this at the interactive prompt or call it as a debug function during a running application in order to generate this string list, (but note that calling this still creates the overhead of an array in RAM, so you do need to have enough "head room" to do the call).
+
+You can then create a file, say `LFS_dummy_strings.lua`, and insert these `local preload` lines into it.  By including this file in your `luac.cross` compile, then the cross compiler will also include all strings referenced in this dummy module in the generated ROM string table.  Note that you don''t need to call this module; it's inclusion in the LFS build is enough to add the strings to the ROM table. Once in the ROM table, then you can use them subsequently in your application without incurring any RAM or LGC overhead.  A useful starting point may be found in [lua_examples/lfs/dummy_strings.lua](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_examples/lfs/dummy_strings.lua); this saves about 4Kb of RAM by moving a lot of common compiler and Lua VM strings into ROM.
+
```
nwf: It might be worth pointing out that the resulting LFS has a `LFS_dummy_strings` module in it that doesn't do anything.

lua_examples/lfs/HTTP_OTA.lua
```diff
@@ -0,0 +1,79 @@
...
+
+firstRec = function (sck,rec)
+  -- Process the headers; only interested in content length
+  local i      = rec:find('\r\n\r\n',1,true) or 1
+  local header = rec:sub(1,i+1):lower()
+  size         = tonumber(header:match('\ncontent%-length: *(%d+)\r') or 0)
+  print(rec:sub(1, i+1))
+  if size > 0 then
+    sck:on("receive",subsRec)
+    file.open('lfs.img', 'w')
```
HHHartmann: Did you happen to test this only with a file called "lfs.img"?
Here you open "lfs.img" but in ```finalize``` you use the variable image.
If they are not equal, finalize will fail (or burn some old stuff)

docs/en/lfs.md
HHHartmann: How is the magic that strings are in the rom table?
Every string that gets compiled by luac.cross is checked for duplicates and the ROM table and stored only once?
The node.compile also checks against the ROM table but cannot alter it (it's in ROM)
So it is only useful if I node.compile?
I would like to see some more explanation about that.

docs/en/lfs.md
TerryE: To quote the LFS paper: "The second uses standard Lua functionality to add the LFS to the require [package.loaders](http://pgl.yoyo.org/luai/i/package.loaders) list".  This is already openly documented, but we can refine it later.

nwf: I mean, I don't think we should be hiding that `package.loaders[1]` is an interesting thing, I just would wish for a way to not encourage that particular thing to end up in people's lua, since we might change `package.loaders` in the future.  But as you say, let's worry about refining it later, not now.

docs/en/lfs.md
```diff
@@ -1,205 +1,144 @@
...
+#### Upload the LFS image and flash it to the LFS region
 
-The LFS patch does this by adding two API new calls to the `node` module: one to reflash the LFS and restart the processor, and one to access the LFS store once loaded.  Under the hood, it also addresses all of the technical issues to make this magic happen.
+Now upload the generate LFS image file (`lfs.img` in this case) as a normal SPIFFS file. Several tools are available to upload the files from host to SPIFFS, including [ESPlorer](https://github.com/4refr0nt/ESPlorer).  There is also a new example in `lua_examples` which can retrieve images from a standard web service.
+
```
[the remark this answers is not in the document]
TerryE: Done

docs/en/lfs.md
```diff
@@ -1,205 +1,144 @@
...
 Power developers might want to use Docker or their own build environment as per our [Building the firmware](https://nodemcu.readthedocs.io/en/master/en/build/) documentation, and so `app/include/user_config.h` has now been updated to include the necessary documentation on how to select the configuration options to make an LFS firmware build.
 
-However, most Lua developers seem to prefer the convenience of our [Cloud Build Service](https://nodemcu-build.com/), so we will add two extra menu options to facilitate building LFS images:
+However, most Lua developers seem to prefer the convenience of our [Cloud Build Service](https://nodemcu-build.com/), so we have added extra LFS menu options to facilitate building LFS images:
 
 Variable | Option
 ---------|------------
-LFS size | (none, 32Kb, 64Kb, 94Kb) The default is none. Selecting a numeric value builds in the corresponding LFS.
-SPIFFS size | (default or a multiple of 64Kb) The cloud build will base the SPIFFS at 1Mb if an explicit size is specified.
+LFS size | (none, 32, 64, 96 or 128Kb) The default is none. Selecting a numeric value builds in the corresponding LFS.
```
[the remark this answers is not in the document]
TerryE: As per Johny's request, the LFS region is now statically allocated in `lflash.c`:
```C
char flash_region_base[FLASH_SIZE] ICACHE_FLASH_RESERVED_ATTR;
``` 
so how about "allocates the corresponding size of the LFS region".

docs/en/lfs.md
TerryE: How would you describe 
````Lua
local G=getfenv()
G.LFS = setmetatable(lfs_t,lfs_t)
```
How about "The first sets up a table in the global variable `LFS` with the `__index` and `__newindex` metamethods"?

docs/en/lfs.md
TerryE: I hooked the detailed online documentation above.  Either the apps programmers don't care or they can look up the master documentation.

docs/en/lfs.md
nwf: Ah ha.  So, I have just updated my LFS-flashing bit in `init.lua` to be

```lua
if node.flashindex() == nil then
  -- no LFS image.  Perhaps it failed to flash?
  if file.exists("luac.out.stage") and not file.exists("luac.out") then
    -- Looks like we tried once before, indeed.  Try again.
    node.flashreload("luac.out.stage")
  end
end
file.remove("luac.out.stage") -- remove stage file from last attempt
-- Do we have a new LFS blob?  If so, install it.
if file.exists("luac.out") then
  print("Updating LFS image.  Will reboot if things go well.")
  file.rename("luac.out", "luac.out.stage")
  node.flashreload("luac.out.stage")
  error("Failed to update LFS!")
end
```

If that seems sensible, it's perhaps worth including as an example in the docs?

docs/en/lfs.md
nwf: Oh, sorry for being overly short.  I just meant to put backticks around LFS: `LFS`.  I think the prose is fine.

docs/en/lfs.md
TerryE: Not sure that you mean, Nathaniel.  

The preload example just has one or more assignment of a single local to a list of strings.  Do a `luac.cross -l` to see what this generates.  The binary consists of a bunch of `LOADK` instructions, one per string and references a bunch of string constants which get put into the `ROstrt`.  

if you are referring to a resources file then the best way is to avoid creating global and avoid any unnecessary RAM hit by just having a load of:
```Lua
if arg == 'resourceA' then return [[
some long resource
]] end
```

docs/en/lfs.md
TerryE: >  it that doesn't do anything

Apart from freeing up about 4Kb RAM :wink:

@HHHartmann Gregor read on in the second half of the document.  Strings in Lua are [interned](https://en.wikipedia.org/wiki/String_interning).  The RTS will take any new string and resolve it first against the `ROstrt` and then the `RWstrt`.  If it already exists in one of those then this reference is used and so another copy is not created.  A new string is only created in the `RWstrt` in the case of a miss.  Also the Lua GC does mark and sweep on strings in the `RWstrt` so once dereferenced, they will be GCed again.

So if you compile or load an LC into RAM or even execute something like `dir..name..'.lua'`, this will still be checked against the `ROstrt` and created in the `RWstrt` if needed.  The strings in `dummy_strings.lua` are the ones that the Lua VM and the `file` and `net` modules create during startup.   I did think of adding the ones used in the compiler, as this would allow you to compiler larger files locally.

docs/en/lfs.md
nwf: I was thinking you meant that the developer should install into LFS a module that looked like this

```lua
resourceA = [[ ... ]]
resourceB = [[ ... ]]
```

rather than like this

```lua
local resourceA = [[ ... ]]
local resourceB = [[ ... ]]
```

since while the latter puts the strings in LFS, it doesn't make it easy to get at them.

lua_examples/lfs/HTTP_OTA.lua
```diff
@@ -0,0 +1,79 @@
...
+
+finalise = function(sck)
+  file.close()
+  sck:on("receive", nil)
+  sck:close()
+  local s = file.stat(image)
+  if (size == s.size) then
+    wifi.setmode(wifi.NULLMODE)
+    collectgarbage();collectgarbage()
+    if s and s.size == size then
```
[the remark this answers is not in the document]
TerryE: You are correct the second should be removed and the first add the `s and` predicate.  `file.stat()` returns `nil` if the file doesn't exist so evaluating `s.size` will throw an error.

lua_examples/lfs/HTTP_OTA.lua
```diff
@@ -0,0 +1,79 @@
...
+
+subsRec = function(sck,rec)
+  total, n = total + #rec, n + 1
+  if n % 4 == 1 then
+    sck:hold()
+    node.task.post(0, function() sck:unhold() end)
+  end
+  uart.write(0,('%u of %u, '):format(total, size))
+  file.write(rec)
+  if total == size then finalise(sck) end
```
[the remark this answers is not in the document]
TerryE: This code assumes that the server will conform to HTTP 1.1, so it ignores byte overrun.  
In terms of the early disconnect we would need to add a couple of extra `sock:on()` events to catch this, but the easiest way to handle this whole class of errors is to schedule a backstop alarm before calling `LFS.HTTP_LUA()`.  Given that the normal path is to do a `flashreload()` which restarts the ESP, if this timer fires then something has timed out.

lua_examples/lfs/HTTP_OTA.lua
```diff
@@ -0,0 +1,79 @@
...
+  end
+end 
+
+subsRec = function(sck,rec)
+  total, n = total + #rec, n + 1
+  if n % 4 == 1 then
+    sck:hold()
+    node.task.post(0, function() sck:unhold() end)
+  end
+  uart.write(0,('%u of %u, '):format(total, size))
```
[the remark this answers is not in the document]
TerryE: Print adds a `\n` (and also uses`\t` as a field separator).  `uart.write()` doesn't.

lua_examples/lfs/HTTP_OTA.lua
TerryE: > Did you happen to test this only with a file called "lfs.img"?

Oops :scream:

TerryE:
Thanks guys.  Really appreciate.

docs/en/lfs.md
TerryE: If your routine is `LFS.getResource`, then this approach would generate a load of globals whether you need them or not.  If you really want to save RAM then in my approach you would do something like:
```Lua
local resource =  LFS.getResource
-- ...
local js = resource('script1')
```
and you could avoid creating any globals.  Or if you have logic pools of resources then each  `getResource()` name could return a small table of resources. 

But the important point is that Lua developers are free to do this in whatever way they feel comfortable and suits their applications.

docs/en/lfs.md
HHHartmann: I would rather see problems in the content of the LFS image than in the flashing process. So there could be an early decision from which partition to load if used as alternatives or only to start the base LFS and not the app-LFS part.
This decisoin could be made based on the last run (or its failure) based on bootreason or some other startup markers (a file or RTC mem)

docs/en/lfs.md
HHHartmann: OK. being used to GB of ram this seemed so crazy to me that I didn't even consider this.
Great job done!

lua_examples/lfs/HTTP_OTA.lua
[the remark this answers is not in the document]
HHHartmann: agreed. After all it's an example. Maybe we can still improve it later on.

lua_examples/lfs/HTTP_OTA.lua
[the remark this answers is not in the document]
HHHartmann: But if I use node.output() I see only half of the show.
Besides this would print something like
``` 0 of 3000, 1024 of 3000, ```
I would prefer separate lines, but tastes vary.

File: docs/en/lfs.md (hunk 1 above)
Comment (TerryE) [reply docs/en/lfs.md]: > I would rather see problems in the content of the LFS image than in the flashing process. So there could be an early decision from which partition to load if used as alternatives or only to start the base LFS and not the app-LFS part.

The last two full words of an [RFC 1951](https://www.ietf.org/rfc/rfc1951.txt) deflate stream are the CRC and the unpacked len of the output stream.  Having done the first pass which involves reading the entire SPIFFS file, and validated the CRC and the length, we can have confidence that we can read and unpack the image. As I said, the only material risk of failure during reflash is one on power-fail and this will be detected on restart, so whilst having two flash regions seems comfortable, it is just a lot easier to keep any LFS variants that you need as LFS image files on SPIFFS.  As I said, the reflash takes about a second or two, though there is a further few seconds delay in the ROM restart routine.

 Can you give me a scenario where having hot swap LFS regions would add any value at all?

File: docs/en/lfs.md (hunk 6 above)
Comment (TerryE) [reply docs/en/lfs.md]: It's crazy what goes around comes around.  The first minis that I used were the PDP8 and PDP11 and the first micro the Intel 8080, and I did Motorola 6502 work and these sorts of tricks were standard, but discarded when VAXs and virtual memory came into vogue; they were all hauled out again with Win3.0 and the Intel 80286 protected mode, but abandoned again as 32-bit then 64-bit variants took over and VM became universal -- until IoTs turned up and they get rolled out again!

In essence we've gone from Lua code running in 48Kb RAM to running in 256Kb ROM + 48 Kb RAM, so you can now build apps 4× the size -- so long as you can work within the RAM constraint.

File: lua_examples/lfs/HTTP_OTA.lua [hunk 13]
```diff
@@ -0,0 +1,79 @@
...
+  file.close()
+  sck:on("receive", nil)
+  sck:close()
+  local s = file.stat(image)
+  if (size == s.size) then
+    wifi.setmode(wifi.NULLMODE)
+    collectgarbage();collectgarbage()
+    if s and s.size == size then
+      -- run as separate task to maximise RAM available
+      node.task.post(function() node.flashreload(image) end)
```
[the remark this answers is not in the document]
TerryE: As I said, this was just a starting point.  We can improve on it.  

The main thing was to show as an example how to do the GET request and what to process in the response, and also using flow control to ensure that server doesn't overrun the ESP because writing to SPIFFS is slower than the capacity of many WAN and LAN links.

TerryE:
@marcelstoer @vsky279, I was uncomfortable with the Windows quick start guide being the first section in the doc, as it just didn't fit well here.   I have therefore dropped this down below the background.   I have also trimmed the background section slightly and removed some of the duplication between the quick-start and the following section.  I hope that you are both comfortable with the result.  If not, then please add your comments.

marcelstoer:
I'll give this some though later but IMO the point of a quick start guide is that I _don't_ need to study the background first before I can start using a new feature. I can always come back later to learn more. What is the shortest path to success? I felt that placing it at the top would be very much in line with what @jmd13391 had requested in his issue.

docs/en/lfs.md
nwf: I see.  Thanks for clarification.

docs/en/lfs.md
HHHartmann: @TerryE Terry I did not mean any failure in the flashing process but false Lua code that fails to run for some reason.
I have a module setting a persistent flag at boot time and resetting it after some time.
If at boot it has not been reset (due to application failure) I would like to start a fallback or minimal system which is stored in a safe location (which is never destroyed by a faulty application LFS).

TerryE: Surely you can still do this by a little conditional logic in your `init.lua` and if necessary loading the fallback from SPIFFS?

lua_examples/lfs/HTTP_OTA.lua
```diff
@@ -0,0 +1,77 @@
...
+--
+
+local host, dir, image = ...
+
+local doRequest, firstRec, subsRec, finalise
+local n, total, size = 0, 0
+  
+doRequest = function(sk,hostIP)
+  if hostIP then 
+    local con = net.createConnection(net.TCP,0)
```
heythisisnate: @TerryE I recently tried modifying your HTTP_OTA example using `tls.createConnection` instead, so updated LFS images could be retrieved from a secure server such as AWS s3. It didn't work, however -- the connection was established and got a success back from the server, but it seems that the `on("receive",..)` callback was never called subsequently with the body of the file to download. I would have expected the tls submodule to be equivalent to the net module in retrieving a file.

Am I missing something obvious?

Issue #2332, opened 2018-03-30
RFC What is missing from the Lua Developer FAQ


TerryE:
### Background

We are about to implement the LFS and this fundamentally changes best practice for how Lua developers code their NodeMCU applications.  Incidentally, I have just updated an issue (#1119) with a view to issuing the PR for this immediately after we release LFS and I realised that it is 2 years since we created this issue.  This got me thinking about two questions that it is worth my posing, answering and our subsequenty discussing, because I want to reflect the conclusions of such discussion in our FAQ.

-  Two years ago, in what way did my Lua "best coding practice" differ from typical  coding style adopted by of most of our Lua developers?  Is there any material understanding / support for my techniques with the other committers?  If so, then how do we get the message across to our felllow developers?

-  In what way have my own practices evolved over thia last 2 years, and again the same agree/ diasgree / how do we get this message across set of qQs.

#### The lifecycle of GCOjects

Pretty much all of the RAM used by the LVM are what runtime referes to as Garbage Colllectable Objects (**GCO**bjects): Tables, Arrays, Functions, Strings, Userdata, ertc. The Lua Garbage Collector (LGC) handles all of these in a standard way: they are created during the program execution of the application. Once they have been fully dereferenced, then the LCG will collect and reuse the storage: birth, life and death.  

The developer has very limited RAM in ESP applications, and to get decent scalability, then some care has to be put into ensuring that oject has as short a life as possible; that they are created in as lazy a way as possible and that they are dereferenced immediately after their use is done.

Because the LGC does is its in-use scan from a number of roots: the **_G** table, the Lua registry, and the Lua stack, it is _very_ easy to leave dangling references to objects should be dead (for example storing the reference in a global or in the registry, or in a table which contains other live entries

Another aspect of GCOjects is that are all stored in variables by reference, so assigning a GCOject to anothe Lua variable is a very lightweight operation; referencing the variable doesn't dereference the underlying object if other references exit elsewhere.

#### Local and upval variables 

Lua applications should always make maximum use of local and upval variables in preference to global refences etc.  This is for two primary reasons:

[...]

PR #2461, opened 2018-08-17
Patch 1

timg11:
This is only a suggested page to improve the documentation. 


TerryE:
Leaving aside you've some done this PR of an old dev so it's messed up, I really think that you should raise this as an issue and we should address this in the FAQ as specific point on io and file.

> If you came here looking for the Lua `io` module that doesn't exist in NodeMCU, read on. 

> The only mention of the `io` module in NodeMCU is in the [NodeMCU Developer FAQ](https://nodemcu.readthedocs.io/en/master/en/lua-developer-faq/), where it says 
"For example, the standard `io` and `os` libraries don't work, but have been largely replaced by the NodeMCU `node` and `file` libraries."

> Unfortunately, if you are new to NodeMCU and Lua, there is little to go on to understand how to proceed.  If you have read the Lua documentation on [The Simple I/O Model](https://www.lua.org/pil/21.1.html),  you might be looking for io.write as a way to output text similar to print() but with more control. With NodeMCU, the `file` library is suggested as a replacement, but it will not write to stdout like io.write. `file.write(string)` will not work because there is no file open.

> It is not clear how to proceed without the `io` module. There is nothing analagous to stdout to open.  The solution is to use the uart module.  The Lua example `io.write("hello", "Lua")` would be written `uart.write(0,"hello", "Lua")`

> Note that the UART ID value of 0 has to be added. UART 0 is the same one the the print() function uses.

The ESP is an IoT device and does not have a POSIX OS or the like, so the full features offered by a typical OS aren't available.  The original authors chose to mirror the then features of the SPIFFS implementation and basic UART drivers.  The current developers here had all become familiar with this before they became committers so have never seen the need.  Perhaps we should spell this out, but the answer is not a "this module is not here" in dummy `io` and `os` slots in the modules section but as as a more expanded point in the FAQ.

marcelstoer:
@timg11 I have to agree with Terry here. I think you raised a valid point and kudos for offering this as PR. Unfortunately we can't accept it as-is due to the issues Terry mentioned.

-> I created #2462

timg11:
As a newcomer to NodeMCU and Lua, I found the existing FAQ (although well written) did not point me in the right direction to find my solution. The statement "The current developers here had all become familiar with this," while undoubtedly true, is still not helpful to a newcomer. If you would like to use any of my text to edit the FAQ, please do so. I can also make an attempt of another PR on the FAQ directly if you would consider it.

marcelstoer:
#2462 is one specific topic in the broader context of #2332.

PR #2463, opened 2018-08-18
FAQ update for io.write clarification

timg11:
\<Documentation update preferred in the FAQ instead of as a module readme\>


marcelstoer:
Grrr, this should have gone to `dev` rather than `master` - missed it, sorry. Will fix this now.

PR #2465, opened 2018-08-23
I2C sw driver with support of multiple buses, Slow, Fast, FastPlus, and user-defined speed selection

### LINKED ISSUE #2305: i2c.SLOW results in 35Kbit/s
The actual i2c.SLOW speed doesn't appear to be documented but empirically appears to be only around 35Kbits/s rather than the 100Kbits/s I was expecting for "slow". looking at the SCL line:

 A little digging finds **i2c_master.c** making frequent calls to ` i2c_master_wait() ` typically for 5us (presumably intended to acheive 100Kbits/s bit-banging). `i2c_master_wait()` is defined in **i2c_master.h** as `os_delay_us()` yet clearly the resulting delay is considerably greater than promised. The overheads of branching and GPIO bit manipulation account for (wasted) delays that could be subtracted from the timing values.

This code doesn't seem to have had much love since 2014 but strikes me as being worthy of improvement. Just messing around with a clone of the **dev** repository I was able to get ~100Kbits/s by tweaking the delays ('5us' -> '2us') and ~200Kbits/s by removing the delay calls altogether. I'm not sure if this is entirely bullet-proof but I have three different I2C peripherals all running happily with six times the throughput now.

Do we know if there is a good reason _not_ to tweak things in this area?

sonaux:
Fixes #2305.

- [x] This PR is compliant with the [other contributing guidelines]

Following @TerryE suggestions in #2305 for use of rsr.ccount instruction I was able to get good formula for calculation of delays. 
Old delays did not take in account that running code takes some time too, and function i2c_master_writeByte contained extra unneeded (i hope) calls for setDC and delays, so overall speed was 35-40 kHz, not 100.

I tested module with all I have: good old analog oscilloscope, couple of displays (u8g2.ssd1306_i2c_128x64_noname, u8g2.ssd1306_i2c_128x32_univision), sensor BMP180, chip EEPROM 24LC01B. 

SCL speed is independent of CPU clock speed until 250kHz, where delay reaches zero at CPU80MHz and ~400kHz clock achievable only at CPU160MHz.
Also I tried to replace gpio_output_set() with assembler code (s16i [gpio_mask], [gpio_addr]...), but only got 300 kHz SCL at CPU80MHz.

### DIFF
--- a/app/driver/i2c_master.c
+++ b/app/driver/i2c_master.c
@@ -51,12 +51,12 @@
 // enable use GPIO16 (D0) pin as SCL line
 #ifdef I2C_MASTER_GPIO16_ENABLE
 #define IS_PIN16(n) ((n)==16)
-// CPU_CYCLES_BETWEEN_DELAYS describes how much cpu cycles code runs 
+// CPU_CYCLES_BETWEEN_DELAYS describes how much cpu cycles code runs
 // between i2c_master_setDC() calls if delay is zero and i2c_master_set_DC_delay()
 // is not being called. This is not exact value, but proportional with length of code.
 // Increasing the value results in less delay and faster i2c clock speed.
 #define CPU_CYCLES_BETWEEN_DELAYS 80
-// CPU_CYCLES_GPIO16 is added to CPU_CYCLES_BETWEEN_DELAYS, 
+// CPU_CYCLES_GPIO16 is added to CPU_CYCLES_BETWEEN_DELAYS,
 // as RTC-related IO takes much more time than standard GPIOs.
 // Increasing the value results in less delay and faster i2c clock speed for GPIO16.
 #define CPU_CYCLES_GPIO16 90
@@ -64,7 +64,7 @@
 #else
 // If GPIO16 support is not enabled, remove GPIO16-related code during compile
 // and change timing constants.
-#define IS_PIN16(n) (0) 
+#define IS_PIN16(n) (0)
 #define CPU_CYCLES_BETWEEN_DELAYS 74
 #endif //I2C_MASTER_GPIO16_ENABLE
 
@@ -275,7 +275,7 @@ i2c_master_setup(uint16 id, uint8 sda, uint8 scl, uint32 speed)
     ETS_GPIO_INTR_DISABLE(); //disable gpio interrupts
 
     if (IS_PIN16(i2c[id]->pin_SCL)){ //if GPIO16
-        CLEAR_PERI_REG_MASK(PAD_XPD_DCDC_CONF, 0x43); //disable all functions for XPD_DCDC 
+        CLEAR_PERI_REG_MASK(PAD_XPD_DCDC_CONF, 0x43); //disable all functions for XPD_DCDC
         SET_PERI_REG_MASK(PAD_XPD_DCDC_CONF, 0x1); // select function RTC_GPIO0 for pin XPD_DCDC
         CLEAR_PERI_REG_MASK(RTC_GPIO_CONF, 0x1); //mux configuration for out enable
         SET_PERI_REG_MASK(RTC_GPIO_ENABLE, 0x1); //out enable
@@ -347,7 +347,7 @@ i2c_master_readByte(uint16 id, sint16 ack)
     sint8 i;
     //invert and clamp ACK to 0/1, because ACK == 1 for i2c means SDA in low state
     uint8 ackLevel = (ack ? 0 : 1);
-    
+
     i2c_master_setDC(id, i2c[id]->last_SDA, 0);
     i2c_master_setDC(id, 1, 0);
     for (i = 7; i >= 0; i--) {
@@ -448,7 +448,7 @@ i2c_master_setDC(uint8 SDA, uint8 SCL)
         I2C_MASTER_SDA_HIGH_SCL_LOW();
     } else {
         I2C_MASTER_SDA_HIGH_SCL_HIGH();
-    }    
+    }
 
     if(1 == SCL) {
         do {
@@ -517,7 +517,7 @@ i2c_master_configured(uint16 id)
  * FunctionName : i2c_master_setup
  * Description  : config SDA and SCL gpio to open-drain output mode,
  *                mux and gpio num defined in i2c_master.h
- * Parameters   : bus id, uint8 sda, uint8 scl, uint32 speed 
+ * Parameters   : bus id, uint8 sda, uint8 scl, uint32 speed
  * Returns      : configured speed
 *******************************************************************************/
 uint32 ICACHE_FLASH_ATTR
@@ -589,7 +589,7 @@ i2c_master_readByte(uint16 id, sint16 ack)
     uint8 retVal = 0;
     uint8 k, i;
     uint8 ackLevel = (ack ? 0 : 1);
-    
+
     i2c_master_wait(5);
     i2c_master_setDC(m_nLastSDA, 0);
 
@@ -632,7 +632,7 @@ i2c_master_writeByte(uint16 id, uint8 wrdata)
     uint8 dat;
     sint8 i;
     uint8 retVal;
-    
+
     i2c_master_wait(5);
 
     i2c_master_setDC(m_nLastSDA, 0);
--- a/app/driver/i2c_master.c
+++ b/app/driver/i2c_master.c
@@ -37,8 +37,12 @@
 
 #include "driver/i2c_master.h"
 
-//used only by old driver
-#define i2c_master_wait os_delay_us
+
+#ifndef I2C_MASTER_OLD_VERSION
+/******************************************************************************
+* NEW driver
+* Enabled if I2C_MASTER_OLD_VERSION is not defined in user_config.h
+*******************************************************************************/
 
 // enable use GPIO16 (D0) pin as SCL line. ~450 kHz maximum frequency
 #ifdef I2C_MASTER_GPIO16_ENABLED
@@ -136,31 +140,6 @@ static inline void i2c_master_wait_gpio_SCL_high(uint16 id)
 LOCAL void ICACHE_FLASH_ATTR
 i2c_master_setDC(uint16 id, uint8 SDA, uint8 SCL)
 {
-#ifdef I2C_MASTER_OLD_VERSION
-    // intentionally not optimized to keep timings of old version accurate
-    uint8 sclLevel;
-
-    SDA	&= 0x01;
-    SCL	&= 0x01;
-    i2c[id]->last_SDA = SDA;
-    i2c[id]->last_SCL = SCL;
-
-    if ((0 == SDA) && (0 == SCL)) {
-        gpio_output_set(0, 1<<i2c[id]->pin_SDA | 1<<i2c[id]->pin_SCL, 1<<i2c[id]->pin_SDA | 1<<i2c[id]->pin_SCL, 0);
-    } else if ((0 == SDA) && (1 == SCL)) {
-        gpio_output_set(1<<i2c[id]->pin_SCL, 1<<i2c[id]->pin_SDA, 1<<i2c[id]->pin_SDA | 1<<i2c[id]->pin_SCL, 0);
-    } else if ((1 == SDA) && (0 == SCL)) {
-        gpio_output_set(1<<i2c[id]->pin_SDA, 1<<i2c[id]->pin_SCL, 1<<i2c[id]->pin_SDA | 1<<i2c[id]->pin_SCL, 0);
-    } else {
-        gpio_output_set(1<<i2c[id]->pin_SDA | 1<<i2c[id]->pin_SCL, 0, 1<<i2c[id]->pin_SDA | 1<<i2c[id]->pin_SCL, 0);
-    }
-    if(1 == SCL) {
-        do {
-            sclLevel = GPIO_INPUT_GET(GPIO_ID_PIN(i2c[id]->pin_SCL));
-        } while(0 == sclLevel);
-    }
-    i2c_master_wait(5);
-#else
     uint32 this_SDA_SCL_set_mask;
     uint32 this_SDA_SCL_clear_mask;
     i2c[id]->last_SDA = SDA;
@@ -194,8 +173,6 @@ i2c_master_setDC(uint16 id, uint8 SDA, uint8 SCL)
             asm volatile("nop;nop;nop;"); // empty CPU cycles to maintain equal times for low and high state
         }
     }
-
-#endif
 }
 
 /******************************************************************************
@@ -209,9 +186,6 @@ static inline uint8 ICACHE_FLASH_ATTR
 i2c_master_getDC(uint16 id)
 {
     uint8 sda_out;
-#ifdef I2C_MASTER_OLD_VERSION
-    sda_out = GPIO_INPUT_GET(GPIO_ID_PIN(i2c[id]->pin_SDA));
-#else
     uint32 gpio_in_mask;
     //instead of: gpio_in_mask = gpio_input_get();
     asm volatile("l16ui %[gpio_in_mask], %[gpio_in_addr], 0;" //read gpio state
@@ -220,7 +194,6 @@ i2c_master_getDC(uint16 id)
                  :[gpio_in_addr] "r" (PERIPHS_GPIO_BASEADDR + GPIO_IN_ADDRESS)
     );
     sda_out = (gpio_in_mask >> i2c[id]->pin_SDA) & 1;
-#endif
     return sda_out;
 }
 
@@ -252,7 +225,7 @@ i2c_master_get_pin_SCL(uint16 id){
  * Parameters   : bus id
  * Returns      : i2c speed
 *******************************************************************************/
-uint16 ICACHE_FLASH_ATTR
+uint32 ICACHE_FLASH_ATTR
 i2c_master_get_speed(uint16 id){
     return (NULL == i2c[id]) ? 0 : i2c[id]->speed;
 }
@@ -325,11 +298,6 @@ i2c_master_gpio_init(uint16 id, uint8 sda, uint8 scl, uint32 speed)
     if(i2c[id]->speed < MIN_SPEED){
         i2c[id]->speed = MIN_SPEED;
     }
-
-#ifdef I2C_MASTER_OLD_VERSION
-    i2c[id]->speed = 100000; //dummy speed for backward compatibility
-#endif
-
     i2c_master_set_DC_delay(id); // recalibrate clock
 
     ETS_GPIO_INTR_DISABLE(); //disable gpio interrupts
@@ -389,9 +357,6 @@ i2c_master_start(uint16 id)
 void ICACHE_FLASH_ATTR
 i2c_master_stop(uint16 id)
 {
-#ifdef I2C_MASTER_OLD_VERSION
-    i2c_master_wait(5);
-#endif
     i2c_master_setDC(id, 0, i2c[id]->last_SCL);
     i2c_master_setDC(id, 0, 1);
     i2c_master_setDC(id, 1, 1);
@@ -409,9 +374,6 @@ i2c_master_setAck(uint16 id, uint8 level)
     i2c_master_setDC(id, i2c[id]->last_SDA, 0);
     i2c_master_setDC(id, level, 0);
     i2c_master_setDC(id, level, 1);
-#ifdef I2C_MASTER_OLD_VERSION
-    i2c_master_wait(3);
-#endif
     i2c_master_setDC(id, level, 0);
     i2c_master_setDC(id, 1, 0);
 }
@@ -486,19 +448,6 @@ i2c_master_readByte(uint16 id)
     uint8 k;
     sint8 i;
 
-#ifdef I2C_MASTER_OLD_VERSION
-    i2c_master_wait(5);
-    i2c_master_setDC(id, i2c[id]->last_SDA, 0);
-    for (i = 7; i >= 0; i--) {
-        i2c_master_setDC(id, 1, 0);
-        i2c_master_setDC(id, 1, 1);
-        k = i2c_master_getDC(id);
-        k <<= i;
-        retVal |= k;
-    }
-    i2c_master_wait(3);
-    i2c_master_setDC(id, 1, 0);
-#else
     i2c_master_setDC(id, i2c[id]->last_SDA, 0);
     i2c_master_setDC(id, 1, 0);
     for (i = 7; i >= 0; i--) {
@@ -508,7 +457,6 @@ i2c_master_readByte(uint16 id)
         k <<= i;
         retVal |= k;
     }
-#endif
     return retVal;
 }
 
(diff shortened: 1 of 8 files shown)


TerryE:
Thanks Natalia.  I've been meaning to do this, but too much on my plate.  I will do a proper review of this in the next day or so.

marcelstoer:
Is this in any way related to https://github.com/espressif/ESP8266_NONOS_SDK/issues/160?

sonaux:
Original driver in  ESP8266_NONOS_SDK does not support clock stretching. Nodemcu version supports it since 2016, thanks to https://github.com/nodemcu/nodemcu-firmware/pull/1589

TerryE:
@sonaux Natalia,  I haven't commented inline because my comments on this pass are general.  

-  **Precautionary principle**.  We need to be cautious about fundamental changes.  This is why major changes (for example my LCD and LFS patches) are enabled by `#define` constants and when the change if first released these defines are disabled by default.  That way developers can elect to use the patch, knowing that there might be some backwards compatibility issues.  On a later release once we have a body of developers who have used the patch without issue, then at that stage we might flip the default.  However individual developers still have the fallback of building the firmware with the patch disabled.  In this case the I2C driver has been stable for years, so we should also be very careful to make a binary equivalent of the old algo buildable as a compile option.  You don't do this.

There are various approaches to this, from simply having a large `#ifdef` / `#else` /`#endif` around the old and new implementations to using lots of inline macros (e.g. variant I2C_MASTER_WAIT(n) defines which either call the old `i2c_master_wait()` or generate nothing).  My instinct here is to keep-it-simple-stupid and do the former, so that we can really strip down the new code.

-   **Making sure we conform to the spec**.  Here NXP are the patent and licence holder so its [UM10204](https://www.nxp.com/docs/en/user-guide/UM10204.pdf) is the standard reference.

-  **Optimising the code**.  IMO, 19 times out of 20, the best thing to do it to write the code clearly and leave it to the compiler.  It will nearly always do a better job that we will, but this probably the 1 in 20 scenario certainly as far as the `i2c_master_setDC()` and wait calls are concerned.  So before I resort to assembly (which I do sometimes), then I always look at the output from the C preprocessor (the `-E` option) and the generated assembly  (the `-s` option) and also explore the impact of any rewrite on these -- just touch the c file, run the make and grab the `xtensa-lx106-elf-gcc` to tweak.

So here for example the approach to `i2c_master_setDC()` is just daft. `SCL` and `SDL` are in-file parameters and can be trust; long if chains generate lots of code and branches. So for example using the existing dev variant, the simplest I could get this was (here `pin_SDA_SCL_mask` is pre-computed in the init code):
```C
i2c_master_setDC(uint8 SDA, uint8 SCL) {
    uint32_t this_SDA_SCL_mask = (SDA<<pinSDA) | (SCL<<pinSCL);
    m_nLastSDA = SDA;
    m_nLastSCL = SCL;
    gpio_output_set(this_SDA_SCL_mask, pin_SDA_SCL_mask ^ this_SDA_SCL_mask, pin_SDA_SCL_mask, 0);
    if(1 == SCL) 
        while((!gpio_input_get()&(1<<pinSCL))) {}
}
```  
This a lot shorter and faster than the current version.  (This was a source only exercise and need validating).  And your rsr loop compiles down to
```s
        rsr a6, ccount
.L3:
        rsr a4, ccount
        sub     a4, a4, a6
        bltu    a4, a5, .L3
```
So big tick there. 

-  **Validating the edge cases**.  There are a could of places where the old code deliberately stretches timing and yours doesn't.  We need cross check against the standard to see if we are on safe ground.

So overall well done and thanks, but I do think that we need another iteration.

Terry :smile:

sonaux:
Thanks for a feedback! It's my first time coding in C since university classes 15 years ago, so i'm really new to this. Had to reread some textbooks at first. I've got the idea to #ifdef code to allow compile-time selection of old/new version.
I've read that pdf from NXP, though not to last page... 
About optimizing: my opinion too, there is too much complexity with macros and aliases for SDK functions,  so I thought, "Maybe there is some reason to code complex projects that way, I just do not understand it, so better leave it as it is". I used precalculated masks when I tried assembly version of gpio_output_set()
I looked at assembly code by launching xtensa-lx106-elf-objdump [-d/-S] i2c_master.o. C preprocessor output is interesting to look at.
Also, I finally got a cheap logic analyzer from Aliexpress, and now can look at samples of real transmissions, not a endlessly repeated piece of code on an analog oscilloscope.
So, I'm out to do next iteration.

TerryE:
If you want to work more informally, then you can use the Gitter IM for [nodemcu/nodemcu-firmware](https://gitter.im/nodemcu/nodemcu-firmware) or we can ping-pong emails. I am happy to help anyone get over the leaning hump -- if they are is here to contribute constructively.  

I am sorry to say that macros are part of Lua and the SDK so that's life.  If you do want to get a handle on the Xtensa architecture then something like this useful synopsis [An short guide to Xtensa assembly language](http://cholla.mmto.org/esp8266/xtensa.html) would help, or the biggy: [Xtensa Instruction Set Architecture (ISA) Reference Manual](https://0x04.net/~mwk/doc/xtensa.pdf).

Just look at the generated code before you think that macros hurt, for example you couldn't beat the 3 instruction wait loop.

nwf:
@sonaux Usually configuration knobs go in `./app/include/user_config.h`, I think.

sonaux:
new commit adds features:
- Multiple buses (up to 10) with different speeds on each bus
 - Standard(Slow, 100kHz), Fast(400kHz) and FastPlus(1MHz) modes or an arbitrary clock speed
 - Sharing SDA line over multiple I²C buses to save available pins
 - GPIO16 pin can be used as SCL pin, but it does not support clock stretching and selected bus will be limited to FAST speed.

My concern is memory allocation for array of i2c state structures. Now is defined as `static i2c_master_state i2c[NUM_I2C];` where NUM_I2C = 10 (that much you can squeeze out of ESP-12 module using shared SDA line and GPIO16.
Size of array is 10 * 22 bytes. Should I care about it?

sonaux:
New version is disabled by default by `#define I2C_MASTER_OLD_VERSION` in app/include/user_config.h

devsaurus:
> My concern is memory allocation for array of i2c state structures. Now is defined as static `i2c_master_state i2c[NUM_I2C];` where NUM_I2C = 10 

You could define it as an array of pointers and have the structures allocated dynamically on demand by the `i2c.setup()` function. With that you're getting closer to an OOP API style, similar to what we've done with the `tmr` module in #1416. But that would be out of scope for this PR IMO.

TerryE:
Some general comments that if we are going to do a major rework of this driver that we should also consider fixing.  And to this end, I've been doing a review of I²C use across all of our modules; none use the direct interface to the I²C driver.  Instead they all go through the `platform_i2C_XX` interface where `XX` is one of the following functions:

Module | setup  | send_start  | send_stop  | send_address  | send_byte  | recv_byte
-------|--------|-------------|------------|---------------|------------|----------
platform<br/>u8x8_nodemcu_hal  |   | 1  | 1  | 1  | 1  | 0
tsl2561<br/>tsl2561  | 1  | 5  | 5  | 5  | 4  | 2
i2c  | 1  | 1  | 1  | 1  | 3  | 1
si7021  |   | 6  | 6  | 6  | 6  | 2
adxl345  |   | 5  | 4  | 5  | 4  | 3
ads1115  |   | 4  | 4  | 4  | 5  | 2
bmp085  |   | 4  | 4  | 4  | 5  | 1
mcp4725  |   | 2  | 2  | 2  | 3  | 1
tcs34725  |   | 3  | 3  | 3  | 3  | 1
hmc5883l  |   | 5  | 5  | 5  | 4  | 3
hdc1080  |   | 5  | 5  | 5  | 5  | 2 
am2320  |   | 3  | 3  | 3  | 3  | 3
bme680  |   | 5  | 5  | 5  | 3  | 1
bme280  |   | 5  | 5  | 5  | 3  | 1
l3g4200d  |   | 5  | 4  | 5  | 4  | 3

This `platform_i2C` interface only invokes a subset of  the I²C calls: `i2c_master_gpio_init()`, `i2c_master_start()`, `i2c_master_stop()`, `i2c_master_readByte()`, `i2c_master_getAck()`, `i2c_master_setAck()`, `i2c_master_readByte()`, `i2c_master_writeByte()`.

The driver header `app/include/driver/i2c_master.h`  should only present the publicly callable interface, and macros that will be used publicly, but not declarations and macros that are really private to the `i2c_master.c` code.  The I2C_MASTER defines therefore belong in the C module. `i2c_master_wait()` is definitely internal to the driver and shouldn't be exposed.

We also have 5 other entries that aren't used anywhere: `init()`, `get_pinSCL()`, `get_pinSDA()`, `checkAck()`, `send_nack()`.  Of these, the one that surprises me the most is the `init()` code which seems to be a standard way of initialising the I²C bus to a known state, but is never used.

Lastly I did find an error in `int platform_i2c_send_address()`.  The `direction` check is optimised out and direction isn't validated anywhere  so `i2c.address(0,dev,27)` will not error, but produce bizarre results.  At a minimum direction should be clamped to 0/1, so the body of this function should be 
```C
  i2c_master_writeByte( (uint8_t) (
    (address << 1) + (direction == PLATFORM_I2C_DIRECTION_TRANSMITTER ? 0 : 1)
  );
  // Low-level returns nack (0=acked); we return ack (1=acked).
  return ! i2c_master_getAck(); 
```

app/driver/i2c_master.c
```diff
@@ -77,86 +66,170 @@ i2c_master_setDC(uint8 SDA, uint8 SCL)
...
+    i2c_master_wait(5);
+#else
+    uint32 cycles_start;
+    uint32 cycles_curr;
+    uint32 this_SDA_SCL_set_mask = (SDA << i2c[id].pinSDA) | (SCL << i2c[id].pinSCL);
+    uint32 this_SDA_SCL_clear_mask = i2c[id].pin_SDA_SCL_mask ^ this_SDA_SCL_set_mask;
+    i2c[id].m_nLastSDA = SDA;
+    i2c[id].m_nLastSCL = SCL;
+
+    if(i2c[id].cycles_delay > 0){//
```
TerryE: This asm fragment  and loop should be hoisted out of inline into a `static inline` function with a meaningful name and a proper explanation of what you are doing.  The compiler will end up inlining this in the code, so it doesn't have a runtime impact doing this. So you'd endup with
```C
  if(i2c[id].cycles_delay > 0) {
    wait_cycles(i2c[id].cycles_delay);
  }
```
as the inline element

app/driver/i2c_master.c
```diff
@@ -77,86 +66,170 @@ i2c_master_setDC(uint8 SDA, uint8 SCL)
...
+        WRITE_PERI_REG(RTC_GPIO_OUT,
+                       (READ_PERI_REG(RTC_GPIO_OUT) & (uint32)GPIO16_RTC_REG_MASK) | (uint32)SCL);
+        GPIO_REG_WRITE(GPIO_OUT_W1TC_ADDRESS, this_SDA_SCL_clear_mask); //SCL is shifted out of mask so no worry
+        GPIO_REG_WRITE(GPIO_OUT_W1TS_ADDRESS, this_SDA_SCL_set_mask);
+    }
+    else{
+        GPIO_REG_WRITE(GPIO_OUT_W1TC_ADDRESS, this_SDA_SCL_clear_mask);
+        GPIO_REG_WRITE(GPIO_OUT_W1TS_ADDRESS, this_SDA_SCL_set_mask);
+        if(1 == SCL) { //clock stretching
+            //instead of: while(!((gpio_input_get() >> pinSCL) & 1)) {}
```
TerryE: Again out-of-line this.  Also explain in the inline function comments that you've replicated the code of  gpio_input_get() to give a tighter timing loop.

app/driver/i2c_master.c
```diff
@@ -77,86 +66,170 @@ i2c_master_setDC(uint8 SDA, uint8 SCL)
...
  * FunctionName : i2c_master_getDC
  * Description  : Internal used function -
  *                    get i2c SDA bit value
- * Parameters   : NONE
+ * Parameters   : bus id
  * Returns      : uint8 - SDA bit value
 *******************************************************************************/
 LOCAL uint8 ICACHE_FLASH_ATTR
-i2c_master_getDC(void)
+i2c_master_getDC(uint16 id)
```
TerryE: Again have you thought of making this a static inline routine?  This is so short that the inline length will be as short as the overhead of  making the out of line call.

app/driver/i2c_master.c
```diff
@@ -12,61 +12,50 @@
...
-LOCAL uint8 pinSDA = 2;
-LOCAL uint8 pinSCL = 15;
+#define CPU_CYCLES_BETWEEN_DELAYS 80
+#define MIN_SPEED 25000
+#define GPIO16_SCL_MAX_SPEED 400000
+#define GPIO16_RTC_REG_MASK 0xfffffffe
 
-LOCAL uint32 i2c_speed = 0;
-LOCAL int cycles_delay = 0;
+static i2c_master_state i2c[NUM_I2C];
```
TerryE: The easiest way to avoid this ram overhead is to have 
```C
static i2c_master_state *i2c;
static uint32 i2c_max;
```
and realloc on init if the current pin is larger than the current `i2C_max`.

But remember to check for NULL return on realloc

app/driver/i2c_master.c
```diff
@@ -77,86 +66,170 @@ i2c_master_setDC(uint8 SDA, uint8 SCL)
...
+        do{
+            asm volatile("rsr %0, ccount":"=a"(cycles_curr));
+        } while (cycles_curr - cycles_start <  i2c[id].cycles_delay);
+    }
+    if(16 == i2c[id].pinSCL){ //if GPIO16
+        WRITE_PERI_REG(RTC_GPIO_OUT,
+                       (READ_PERI_REG(RTC_GPIO_OUT) & (uint32)GPIO16_RTC_REG_MASK) | (uint32)SCL);
+        GPIO_REG_WRITE(GPIO_OUT_W1TC_ADDRESS, this_SDA_SCL_clear_mask); //SCL is shifted out of mask so no worry
+        GPIO_REG_WRITE(GPIO_OUT_W1TS_ADDRESS, this_SDA_SCL_set_mask);
+    }
```
TerryE: This tells me that you don't support clock stretching on GPIO16.  Why?   Also isn't supporting this use of GPIO 16 going to screw up other use of the RTC bigtime or have I missed something here?  

I really thing that we need a compelling reason why we want to go down this route.  Is the ability to reuse this one pin for I2C that important?

app/include/driver/i2c_master.h
```diff
@@ -53,22 +24,37 @@
...
-void i2c_master_send_ack(void);
-void i2c_master_send_nack(void);
-
-uint8 i2c_master_get_pinSDA();
-uint8 i2c_master_get_pinSCL();
 
-#endif
+#endif //I2C_MASTER_OLD_VERSION
+
+typedef struct {
```
TerryE: Private to the C driver so why in the public header?

app/driver/i2c_master.c
sonaux: I agree, looks cleaner. Already done.

app/driver/i2c_master.c
sonaux: Done this.

app/driver/i2c_master.c
sonaux: Missed that comment, will do.

app/driver/i2c_master.c
sonaux: Already done with static array of pointers to struct. Now it takes 40 bytes initially and +32 bytes for each bus after calling `i2c_master_gpio_setup'. I thought static array would be convenient to check if bus is initialized.

app/include/driver/i2c_master.h
sonaux: My bad, fixed. I tempted to simplify "old version" by getting rid of macros like `I2C_MASTER_SDA_HIGH_SCL_HIGH` and such, but leaving `if-else` structure for replicating old timings, since there is no more binary equivalent of old code - only timings and behavior are simulated. It even supports multiple buses.

app/driver/i2c_master.c
TerryE: :+1:

app/include/driver/i2c_master.h
TerryE: :+1:

app/driver/i2c_master.c
TerryE: I must admit that usually when I am short of I/Os, I also run into 5V vs 3.3V issues so just solve both issues at the same time by using an MCP23008.  Personally, I could see the sense of using GPIO16 as a general IO pin, but whether this get so desperate that I'd need to use it for I²C is another matter. My real concern here is robustly testing this.

sonaux:
> We also have 5 other entries that aren't used anywhere: init(), get_pinSCL(), get_pinSDA(), checkAck(), send_nack(). Of these, the one that surprises me the most is the init() code which seems to be a standard way of initialising the I²C bus to a known state, but is never used.

init() is called from gpio_init() at the end of routine. 

> Lastly I did find an error in int platform_i2c_send_address(). The direction check is optimised out and direction isn't validated anywhere so i2c.address(0,dev,27) will not error, but produce bizarre results. At a minimum direction should be clamped to 0/1

I do not know yet how to tell compiler to leave code in place.

sonaux:
And I haven't looked at ESP32 code yet... I see i2c master software driver is the same and used same way, except that bus IDs 1 and 2 routed to hadrware i2c. It would be very easy to integrate new driver into dev-esp32 branch.

app/driver/i2c_master.c
```diff
@@ -317,23 +314,9 @@ i2c_master_writeByte(uint8 wrdata)
     uint8 dat;
     sint8 i;
 
-    i2c_master_wait(5);
-
-    i2c_master_setDC(m_nLastSDA, 0);
```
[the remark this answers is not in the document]
sonaux: Restored deleted lines.

app/driver/i2c_master.c
```diff
@@ -317,23 +314,9 @@ i2c_master_writeByte(uint8 wrdata)
...
         i2c_master_setDC(dat, 0);
-        i2c_master_wait(5);
         i2c_master_setDC(dat, 1);
-        i2c_master_wait(5);
-
-        if (i == 0) {
-            i2c_master_wait(3);   ////
-        }
-
-        i2c_master_setDC(dat, 0);
```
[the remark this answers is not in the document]
sonaux: Restored deleted lines.

devsaurus:
> except that bus IDs 1 and 2 routed to hadrware i2c. It would be very easy to integrate new driver into dev-esp32 branch.

I agree, and the constants for the hardware drivers could be mapped to higher IDs (e.g. 100, 101) to free up 0-9 for the SW driver.

TerryE:
@sonaux Natalia, I've pulled your last push version (which is suspect is now behind your local dev PC version) and had a play / looked at the code.   My main concern here is the code for GPIO16.  I understand and support your goals of (i) supporting fast or better I²C; and (ii) supporting multiple I²C buses.  A major issue here is the lack of usable IOs exposed on the ESP8266, and so your desire to have GPIO16 as an option is also understandable.

However, the constraints on using GPIO16 are poorly documented by Espressif.  GPIO0-15 all share a unified Hcan be set to open-drain and in this case

sonaux:
Sorry for delay, at this time new driver works in all configurations, but some TODOs and improvements I wanted is not done yet. For example:
Check SCL for low after pin configuration - if low, then throw error, otherwise will stuck in clock stretching loop
Check SDA for low after clearing bus by init - if low, then throw error for bus failure
Change if interface - move ack into read/write functions
Sharing of SDA line works for devices I tested, but it may not work with some, and I didn't test with shared SCL. If it works too, it would be cool :)

TerryE:
@sonaux Natalia, don't worry about the timescales.  We are all doing this pro-bono so other priorities can intrude.  The main thing is that you are working this, so others don't need to step in.  

Thanks Terry

marcelstoer:
Now as @devsaurus approved this PR I guess we're ready for merging once Terry approves it as well? The next planned master drop is still a month away -> no hurry here.

sonaux:
I think this is last commit for this PR. I have some code that improves driver further (combine clock stretching and reading SDA state, pass pointer instead of bus id to internal functions), but code becomes messy and function that calculates delay does not work properly anymore.

marcelstoer:
Terry, have you had a chance to test this in the meantime?

TerryE:
Jan is visiting her old StepMa for 4 days starting Thurs.  When the cats away, the mice will play.  Catch-up time on NodeMCU

marcelstoer:
If we wanted to have this ready for the next master drop we should merge it now. Ok to postpone it but I just don't like decaying PRs.

TerryE:
@marcelstoer Marcel, this is a pretty good improvement in capability, but I feel that it really needs to be in dev for pretty much a full release cycle before cutting over to master.  I know that I or one of the other committers could really do this trying this out on multiple I2C channels, but its just a Q of bandwidth for me.  I've been felling pretty shit because of a virus for over a week and am really really just starting to get on top of my backlog.  Maybe someone like @HHHartmann or @jpeletier could take this for a test drive?

marcelstoer:
Fine for me. I dropped it for the next cut. Ideally @sonaux would fix the conflicts in `i2c.md` (it's now in `docs/modules/`) so we can then merge it right after the next release.

TerryE:
If will be trailing spaces. You are going to see a few of these. Easy to resolve on the merge.

sonaux:
Rebased and changed path to source file in documentation

HHHartmann:
@TerryE I currently only have hardware that supports slow speed. Will see if I can try it on the weekend with several of these devices if it helps.

pjsg:
I just tried this on a 128x64 oled display and it ran at 400kHz just fine (the old code was at 48kHz). It isn't quite that much faster as the device itself inserts delays, but the display updates much better now. (I was using a logic analyzer to look at the speeds on the wire).

The only issue that I have with this PR is that you need to comment out a #define in user_config.h to enable the new driver. If this really is backwards compatible with the old code (except that it is much faster), then it probably ought to be the standard.

TerryE:
I've tried it on my current (slow) I²C devices and this doesn't seem to have broken anything, so I suggest that we merge this into dev. Any objections?

Michal78S:
I have tried all avaiable speeds: i2c.SLOW, i2c.FAST and i2c.FASTPLUS with firmware builded by cloud builder but always 47kHz clk speed appeared.

galjonsfigur:
@Michal78S That's because the default option in the firmware is to use old I2C driver. See https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/include/user_config.h#L175 .
I don't see an option to enable it in cloud builder so I think the only way to use this driver for now is to build the firmware yourself. But an option to enable this driver would be a nice feature to add to @marcelstoer cloud builder.

marcelstoer:
> an option to enable this driver would be a nice feature to add to the cloud builder

True, I can do that once it's available on master. Please ping me should I forget that.

mnpaslay:
This is working well for me with a ADS1115 and ADS1015.  The faster bus speeds allow me to sample at maximum rates.  Thanks!

Michal78S:
> 
> 
> > an option to enable this driver would be a nice feature to add to the cloud builder
> 
> True, I can do that once it's available on master. Please ping me should I forget that.

Did You add this option to the cloud builder? I'm still waiting for it ...

marcelstoer:
> I can do that once it's available on master

You need to keep an eye on that. It depends on https://github.com/nodemcu/nodemcu-firmware/milestone/13

Michal78S:
> 
> 
> > an option to enable this driver would be a nice feature to add to the cloud builder
> 
> True, I can do that once it's available on master. Please ping me should I forget that.

Ping, ping...

Michal78S:
Is there any chance to put this solution to the cloud builder?

HHHartmann:
Maybe we should undefine I2C_MASTER_OLD_VERSION by default. 
But that would be in another bug. 
Or @marcelstoer fixes it in his  builder.

Code1110:
I have just tried this new driver with 3x mcp23017 at `i2c.FASTPLUS` and it works fine and much faster. Just do not forget the pull-up resistors, which were not required in this case, when running with the old driver.

Issue #2467, opened 2018-08-24
Upgrade to SDK 3.0


marcelstoer:
https://github.com/espressif/ESP8266_NONOS_SDK/releases/tag/v3.0

> Memory optimization: iram can be used as memory, thus saving about 17KB memory for user application.

@nwf up for another SDK upgrade?

marcelstoer:
Seems they forgot to add the documentation about those partition tables before they tagged: https://github.com/espressif/ESP8266_NONOS_SDK/compare/v3.0...master#diff-b1d639d27fa3178a8d543ca3601dd027. They also forgot to update `version.h` - again. See https://github.com/espressif/ESP8266_NONOS_SDK/pull/164

nwf:
I am aware that this makes me be "that guy", but that's awfully thin documentation.  The examples all, for example, pass static, constant data in to the function.  I have no idea if this is required and can't look at the source, because hooray blobs.  I'll prod at it this weekend.

devsaurus:
> Opinions?

After a brief read-through of https://git.io/fAtbU (thanks Marcel!) I would see the following implications when porting our firmware to the 3.0 partition table concept:
1. Automatic flashing of the init data in `user_start_trampoline()` must consider the address range of the `SYSTEM_PARTITION_PHY_DATA` partition.
2. Same applies to the documentation at [SDK Init Data](https://nodemcu.readthedocs.io/en/dev/en/flash/#sdk-init-data).
3. The SPIFFS code also has to consider the partition layout. Currently it assumes that the available address range begins after the firmware image and ends `SYS_PARAM_SEC_NUM` before end of physical flash.
4. How to handle our SPIFFS region? Not sure whether this should go into a customer partition table.
5. I assume that the flash size byte correction in `nodemcu_init()` can remain as it is.

My first shot would be to construct a partition table which mimics the current flash layout as we had it up to now. With this, you can safely ignore the above items for the time being and focus on getting the table in shape.

nwf:
@devsaurus The partition table I attempt to register (but don't appear to make it far enough into boot to do so) is of the form you describe, I think?  At least, I meant it to be.

devsaurus:
Yes, I saw this as well when I read through your PR ;-)

TerryE:
> > https://github.com/espressif/ESP8266_NONOS_SDK/releases/tag/v3.0
>
> Memory optimization: iram can be used as memory, thus saving about 17KB memory for user application.

[...]

PR #2468, opened 2018-08-24
DON'T MERGE; first attempt at sdk 3.0

nwf:
See #2467.


nwf:
I attempted to infer the changes, but the resulting executable doesn't even make it to `user_pre_init`, as far as I can tell. :(

TerryE:
Even trying to debug this is a real sod at this stage, because you can't get anything out of the system and in terms diagnostics; you can't interact with it with the debugger, so it works or dies silently.

TerryE:
I would rather get LFS into the mainstream, that get involved with this one - especially as it does nothing for  Lua developers.

nwf:
Please do focus on LFS, yes.  This will be here when we get to it. :)

TerryE:
Did you see my [Cross compile service](https://blog.ellisons.org.uk/article/nodemcu/a-lua-cross-compile-web-service/) post? The `luac.cross` equivalent of @marcelstoer's Cloud Builder.
.

nwf:
I had not.  Very nice; thank you for building such a tool. :)

app/Makefile
```diff
@@ -92,7 +92,6 @@ LINKFLAGS_eagle.app.v6 = 			\
 	-T$(LD_FILE) 				\
 	-Wl,@../ld/defsym.rom			\
 	-Wl,--no-check-sections 		\
-	-Wl,--wrap=_xtos_set_exception_handler	\
```
devsaurus: Why do you remove the exception handler for unaligned loads?
I guess the firmware heavily relies on this and will fail miserably.

app/user/user_main.c
```diff
@@ -187,60 +184,73 @@ void user_rf_pre_init(void)
...
-        case FLASH_SIZE_128M_MAP_1024_1024:
-            rf_cal_sec = 4096 - 5;
+        case FLASH_SIZE_128M_MAP_1024_1024:  // 9
+            // XXX No examples given in docs; best guess
+            pt[2].addr = 0xffb000;
+            pt[3].addr = 0xffc000;
+            pt[4].addr = 0xffd000;
             break;
 
+        case FLASH_SIZE_4M_MAP_256_256: /* 0: Too small */
```
devsaurus: I think you should decode this case as well. Even if a 512kB flash might be too small, this `size_map` is the default at some points in the tool chain in the absence of a specific value.
Blocking the firmware startup in this case prevents error/diagnostic messages further down the startup code.

app/user/user_main.c
```diff
@@ -187,60 +184,73 @@ void user_rf_pre_init(void)
...
- * Parameters   : none
- * Returns      : rf cal sector
-*******************************************************************************/
-uint32
-user_rf_cal_sector_set(void)
+#define EAGLE_FLASH_BIN_ADDR      (SYSTEM_PARTITION_CUSTOMER_BEGIN + 1)
+#define EAGLE_IROM0TEXT_BIN_ADDR  (SYSTEM_PARTITION_CUSTOMER_BEGIN + 2)
+static const partition_item_t pt_template[] ICACHE_RODATA_ATTR STORE_ATTR = {
+    { EAGLE_FLASH_BIN_ADDR,     0x00000, 0x10000},
+    { EAGLE_IROM0TEXT_BIN_ADDR, 0x10000, 0x60000},
```
devsaurus: Will your current firmware fit into `0x60000`?
The partition length should be calculated from `platform_flash_get_first_free_block_address()` to be on the safe side.

app/Makefile
TerryE: > I can confirm that it is safe because Espressif have included unaligned access patch in their firmware core.

We did tweak our exception handler so that it plays nicely with the gdb stub, and this is essential IMO.  The _big_ problem with the nonOS stuff is that it is all closed source so you have to rely on disassembling library code to see what it really does.

nwf: @TerryE Another option, rather than just stripping the handler out of our code like this PR does, is to filter it out of the libmain.a that's shipped by the SDK.  Seem viable?

TerryE: I will need to disassemble the `libmain.a` code just to make sure that it doesn't have other side-effects that we would need to preserve

app/user/user_main.c
TerryE: Currently LFS and the TLS certificate allocate flash page-aligned areas.  Perhaps these need to be moved into the partition table?

app/user/user_main.c
```diff
@@ -187,60 +184,73 @@ void user_rf_pre_init(void)
...
     }
 
-    return rf_cal_sec;
+    if(!system_partition_table_regist(pt, sizeof(pt)/sizeof(pt[0]),size_map)) {
+        os_printf("system_partition_table_regist fail\r\n");
+        while(1);
+    }
 }
 
+uint32 user_iram_memory_is_enabled(void)
```
devsaurus: This one's missing an `ICACHE_RAM_ATTR` so the CPU fatals when libmain calls this function before flash is initialized.
Setting `ICACHE_RAM_ATTR` gives me an instant Lua prompt!

nwf: AH HA!  Thank you so much; I would not have seen that for ages!

nwf: Though making just that change doesn't seem to wake my board from unhappiness at boot, even if I completely erase the flash first.  What other changes have you made?  Is my build system in a bad state?

TerryE:
@nwf Nathaniel,  I don't want to merge this until I've had a chance to go through the trampoline code, 
and also sorted these potential exception handler conflicts. Hence my removal of the "Maybe" from the title for now.

TerryE:
@nwf Nathaniel,  I was wondering if we should take this opportunity to add OTA firmware upgrade to NodeMCU for the ESP8266?   I've been brooding about how to do this simply.  I think that the best compromise between simplicity and resource footprint would be an asymetric 2-partition system using `rBoot`.  Partition 1 is a fixed ~64 bootstrap upgrader.  Partition 2 is always the NodeMCU partition, so the NodeMCU firmware is always loaded against this.

-  Partition 1 includes a minimal bootstrap with SPIFFS RO support and my `uzlib` inflate.  When you boot this is looks for the SPIFFS file `nodemcu.img` and does a similar 2-pass inflate of the image where pass 1 is a dry run to check the integrity of the NodeMCU image and pass 2 does the actual inflate into the NodeMCU partition.
-  Partition 2 is the NodeMCU image
-  Partition 3 is LFS
-  Partition 4 (optional) the TLS data
-  Partition 5 is SPIFFS.

The advantage of this approach is the just like LFS, the Lua application programmer has full control over if and how she or he creates the `nodemcu.img` file.  The application can then execute a `node.fimewarereload()`; this will reboot into Partition 1 which will then overwrite Partition 2 or not; and then reboot again into either the old firmware image or the new.  Powerfail during the upgrade will result in the retry of this last step.

The typical image size is around 300Kb, so as long as the SPIFFS has room for this, then the upgrade is doable.

@jmattsson @pjsg have you any thoughts.

nwf:
That sounds viable and quite nice to me, though it's a great deal more low-level work than I am qualified to perform on this platform.  If you want to take over this PR, please feel free! :)

It sounds like you intend to journal the firmware reflashing operation (unlike LFS), so that on powerfail the bootloader tries to flash again.  Do I read that right?  How will the journaling work?  (Which bit of flash are you going to erase and program for the journal?)  Would it be simpler to have two NodeMCU firmware partitions with headers and footers that are verified before booting the latest valid image and leaving retries to the application?

Can we get a dry-run version of `node.flashreload()`, while we're thinking about OTAs?  I think the full update story (below) requires it to be safe.

* When building, bake a build timestamp into the build image like LFS.
* When downloading the firmware to the device, ensure that the SPIFFS also contains the build stamp of the new firmware and new LFS (or provide a mechanism for extracting it from the .img file).  Write out, explicitly, whether this is a just-firmware, just-LFS, or both update so that we can tell, at boot, whether the download is finished.
* At boot,
  1. If there is an LFS update pending, perform a dry-run check that it will pass all the pre-flight tests.  If not, remove it so the application retries the download and update.
  2. If there is a firmware update pending, compare the running firmware stamp to the stamp of the optional upgrade image in SPIFFS; if the loaded version is the lesser, perform an upgrade.  If this bounces (due to decompression failure, e.g.), remove it so the application retries the download.  Otherwise, if the stamps are equal, the update has succeeded already and we can carry on.  (This logic is compatible with the application needing to do its own retries of firmware updates in the two-partition story.)
  3. If there is a LFS update pending, and if (LFS is not loaded) or (LFS is loaded and its timestamp is less than the optional image in SPIFFS), carry out an LFS update.  This cannot bounce due to the pre-flight checks in step 1 above (and so the application cannot lose synchronization between firmware and LFS partitions).  This logic will retry to recover from power failures during LFS programming.
  4. Boot the application, entirely on the old firmware and LFS or entirely on the new firmware and LFS.

TerryE:
@nwf, you need to review the latest LFS code to understand how it works. IMO, the whole concept of journaling is inappropriate for something like the ESP IoT devices. What I attempt to realise is a form of transactional update integrity which is different. A reload operation can only have  one of three outcomes: fail/no change, successful update, fail/LFS cleared. 
The reload algorithm uses two passes:
-  The first is a dry run where the entire image is read from SPIFFS, gzip uncompressed and the CRC of the uncompressed image compared to the check CRC at the end of the image. Once we get to this point we know the image on the flash SPIFFS is clean. Any failure and the algo returns the first failure condition without the LFS region having been touched.
-  The second pass repeats the same process, but writing to the flash LFS region. Once complete, the CPU is rebooted. The integrity of this operation is ensured by using a characteristic of flash: you can overwrite flash memory so long as the update value only differs from the original by clearing bits. The first word of the LFS is it's signature but with an in-progress bit set. The last write to LFS overwrites this signature clearing the in-progress bit. This this is done we are good to go and can reboot and reload.
-  Because the two passes are data driven from the same data, as far as I can see the only realistic fail condition is if there is a power failure during the write to flash phase. In this case the firmware will detect the in-progress bit set and clear the LFS region.  This is the third outcome.

If you can identify any other paths then please let me know.

In this third fail condition,  the image is still on SPIFFS, so as long as `init.lua` includes a check "if LFS not loaded then reload LFS image" then the application will still recover after powerfail.

There is also nothing stopping the application programmer using an RTCmem byte to control LFS image choice, but IMO this is best left to the application as it is only a few lines of Lua.

So to all intents the LFS reload is a single step process, and there no need to convolve this with any OTA firmware upgrade.
------

I am suggesting exactly the same two pass approach for OTA upgrade, but using the small fixed second partition. So we use whatever upload means the application chooses to create a SPIFFS file which is  the new compressed firmware image.  The bootstrap opens SPIFFS read-only and does a similar two-pass validate / write of the image. This is a standalone from flash to flash operation. Again the only realistic failure mode that I can see is powerfail, because this is a standalone operation.  If there is a powerfail, then the CPU with reboot and the bootstrap will retry the operation. The very last step is toggle the active partition back to the firmware, and reboot the device.

There are a number of advantages to this approach:
- The firmware is always linked to the same base address.
- We don't need to allocate two firmware-sized partitions. If the application programmer wants to use OTA firmware update then the SPIFFS needs to be big enough to hold the image file, and if not then it doesn't.

Can you think of any failure modes that I haven't covered that could brick the ESP?

nwf:
I think I understand the LFS process, at least as described.  I had missed that the name of the file used in `node.fimewarereload()` is not a parameter but is hard-coded as `nodemcu.img` (unlike the LFS process to which I was comparing things), letting SPIFFS be the journaling mechanism.  (The application should be sure to remove `nodemcu.img` whenever it starts up to avoid repeated flashing, as the bootloader cannot after a successful flash, because it only has RO SPIFFS access, by supposition.)

------ 

The reason to entwine LFS and firwmare updates is that the application presumably wants these to move in lockstep: the new LFS image may require features of the new firmware, and the new firmware may not be compatible with the old LFS.  My desire to have a dry-run mode was to allow the application to defensively verify that it would be able to flash the new LFS image once it had flashed the new firmware, so as not to be stuck with a new firmware and old LFS.

TerryE:
In fact we can exploit the flash "all bits set" overwrite to allow the image name to be passed as a parameter, 
and to control state sequencing.  The the OTA process doesn't need to remove the image as it doesn't use its existence as a start flag.   This is achieved by the API call.  

Moving the firmware and LFS in lock step or not is surely an application choice, so whilst we might want to facilitate this, I don't think that we should hardwire fixed functionality into the firmware -- better to leave this to the Lua application.  

Ditto the mechanism for getting the firmware and LFS images onto SPIFFS in the first place.   Different environments might choose different methods: HTTP push or pull, ditto FTP or the secure variants or even some application-specific client-server protocol.  In my view we should stuff that can be best done by the developer in Lua in Lua.

nwf:
I think you mean, by the first paragraph, that you intend to reserve space in the firmware partition for the name of a replacement SPIFFS file, and if this area is not blank, to erase and reprogram the flash partition at boot.  This is not robust against power-failure unless the blank space is its own erase block (consider that the block must be erased to be reprogrammed, and so there is a window in which the space is blank and yet the firmware is not fully flashed).  It is possible that a variant of this involving multiple erase blocks is safe, but, honestly, depending on a reserved name in SPIFFS seems simpler.

The proposal I had above did not require baking anything in except extending the LFS mechanism with a call that only did the first phase (the pre-flight tests).

Yes, I made no assumptions about how the image files land in SPIFFS.  That obviously belongs in Lua and I do not believe I ever even hinted otherwise.

TerryE:
Almost but not quite in terms of the reserved space for the reserved space for the filename.  The reserved name area is set to all 0xff in the build. This can be overwritten without erasing at any stage. This is an intrinsic property of flash storage that both SPIFFS and LFS exploit. If this area is not all 0xff and zero terminated, then the filename has been validly written and can be used.

One of the advantages of the two passes over the compressed image is that the flash _always_ does a dry run, so there is little to be gained by adding a second dry run.

Nathaniel, sorry if this dialogue seems stilted, but I don't have WiFi in my cottage in Greece, so I am composing these replies  my mobile using Swype. 😱

nwf:
I still do not believe I am wrong on either point, but thank you for making sure my arguments are sound.

-------

For the file name, if it is stored within a flash block within the firmware, there is a race.  The first, and larger issue is not the *entry* to the "replace the firmware" state (the programming transition away from 0xff), it's the atomicity of the replacement itself.  During flashing, there is a window between erasure and programming for the block containing the 0xff/image name scratch region.  So:
* the boot code detects a need to upgrade the firmware, loads the image from SPIFFS, and carries out all preflight checks, which pass.
* the boot code erases and programs flash erase block after flash block
* eventually this reprogramming loop encounters the flash block containing the SPIFFS file name
* the boot code erases this flash block
* I pull power and programming does not take place
* I apply power and the boot code comes up and sees all 0xff, and decides that it does not need to retry flashing the firmware.
* The boot loader attempts to boot the firmware with at least one corrupted flash block
* Sadness ensues.

One answer is to move the file name into its own flash erase block and to erase that block last.  This way, the "corruption" above can't happen to meaningful bytes.  This has enormous space overheads, tho'.

Alternatively, you can hack around this by using *two* such buffers in different flash blocks and another byte in the one written by the firmware (a very tiny *journal*!), so that...
* the steady state is "buffer 1 contains anything, buffer 2 contains 0xff, and the byte is not 0xff",
* the "begin programming" state is "buffer 1 contains anything, buffer 2 contains a file name, and the byte is not 0xff", and
* during programming, the following states are transitioned through before landing in the steady state:
  * "buffer 1 contains 0xff, buffer 2 contains a file name, and the byte is not 0xff" (the buffer 1 erase block has been erased, but buffer 2+byte block has not)
  * "buffer 1 contains a file name, buffer 2 contains the same file name, and the byte is not 0xff" (buffer 1 erase block has been erased and programmed and had the file name duplicated, buffer 2+byte block has not been erased)
  * "buffer 1 contains a file name, buffer 2 is arbitrary and the byte is 0xff" (buffer 1 erase block programmed, buffer 2 erase block erased but not successfully programmed)
* The rule for the boot loader is replace the firmware partition if either
  * the byte is not 0xff, and buffer 2 is not all 0xff, in which case use the file name in buffer 2, or
  * the byte is 0xff, in which case, use the file name in buffer 1.

Another alternative, as stated above, is to reserve the name of the firmware image in SPIFFS and use its existence as a sign to reprogram the firmware.  Safety of this particular dance depends on the atomicity of spiffs' rename and remove operations, which are used to enter and exit the "replace the firmware" state of the boot loader. 

------

There is a small caveat to the "0xff to filename" transition, which is that it can only be done once, and if done partially, poses a potential problem.  That is, because it is a multi-byte programming sequence, it may not be atomic within the hardware, and so one may lose power and enter the bootloader with a partially written filename.  In the worst case, this results in the bootloader misprogramming the firmware.  In the best case, the application attempts to flash again using the same file name and it all goes well.  If the application attempts to use a different file name the second time, they may be unable to write it to the flash region, because some bits will be 0 after the first attempt.  This at least merits mention in the documentation: e.g., "once a `node.firmwarereload()` request has been issued, all subsequent `node.firmwarereload()` calls, until one succeeds, must use the same file name."

My understanding of flash technologies is that programming is only atomic at the level of individual bits, so extreme care must be taken to bootstrap this extremely narrow atomic primitive into larger atomic operations.  I *presume* SPIFFS does or could contain the appropriate magic, and I think it simpler to 
use SPIFFS to do the right thing for us.

None of these concerns apply to LFS, because we are assuming, correctly, that the application can direct the firmware to try again even if the LFS cannot be loaded, and that this is not a burden to the programmer (and I do not think it is).  The LFS update code, as you point out, does not need to persist its transactions across power failure and everything's fine.  I claim that this is manifestly untrue of OTA firmware updates.

------

The need is to add a *first* dry run of the LFS flash before updating the firmware, otherwise we might only discover that the LFS partition can't update until after we have *already* updated the firmware partition.  (Or vice-versa, if we try them in the other order.)

TerryE:
I agree that the algorithm that you present has a crater - which is why you shouldn't do it this way; you need to get the temporal logic correct, and the sequencing needs to safe.  I will read your reply and answer in detail, but I only have brief access to WiFi at the moment, so this will have to wait to later. Sorry.

liunanhao:
@TerryE I am trying to use rboot and lfs for my application,it can boot ,but not load into the application.If i comment the "#define LUA_FLASH_STORE" in user_configs.h or remove rboot,it can run  again.I dont understand how lfs impact the rom structure,and can you tell me how to solve this issue?

TerryE:
We've got a bit of work to do to integrate this,, and the priority is to bottom / resolve to two VM related issues, so yes, let us refer this one.

nwf:
Seems fine to defer to me, yes.

TerryE:
Hi all. sorting out this one with @nwf Nathaniel is the next major task that I want to work on, IMO.  I suggest that we just focus on rebaselining the firmware on SDK 3.0 for this PR and leave aside any issues around NodeMCU_OTA for a future PR.

On a purely source control aspect, this PR merge is now confusing because of the parallel work that has been going one since N first created his branch.  I am not a Git guru and this also hit me a couple of times on LFS and I ended up manually rebaselining the changes against a current dev to avoid this. IMO, however we get there, the PR diff should only reflect the changes needed to move from 2.2.1 -> 3.0,

However before we can move forward with the PR itself, there are a number of points that we need to reach consensus on and resolve in  #2467 first.

TerryE:
@nwf Nathaniel, I am going through all this in detail working out what differences which you've made in support of the SDK 3.0.

I still need to go through the code and see if the Espressif `load_non_32_wide_handler()` if debugger friendly.  If you recall our handler handles flash-mapped L8UI, L16UI and L16SI instructions but passes all other non-wide exceptions to the gdb stub.  Losing this gdb hook would be a major blow. It would also be good to benchmark the two handlers either statically or at runtime to see if we are taking a performance hit using the Espressif handler.

Did you investigate any of this?

TerryE:
Once I've committed #2665, we'll be in a position to update this and to put it into dev as well.

@nwf Nathaniel, should we just add to this or open a new issue, which can establish a clean slate for any more discussions.  Up to you :smile:

nwf:
@TerryE You're certainly welcome to clobber my branch, but given the length of this PR and (y)our more sophisticated understanding of the true impact of SDK 3 (certainly more sophisticated than mine at the top of the page), perhaps best to close it and open one with less of a backlog for would-be reviewers and other interested parties.

marcelstoer:
Superseded by #2692

PR #2483, opened 2018-09-10
Decode chunked responses case insensitve headers

### LINKED ISSUE #2482: http module does not always correctly decode chunked response
### Expected behavior
When decoding a HTTP response using the http module, the module should respect the _case insensitive_ `Transfer-Encoding: chunked` header in the response to determine if the response will be chunked.

[RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2) states that header fields are case insensitive.

### Actual behavior
Chunked responses are decoded correctly if the response includes the header `Transfer-Encoding: chunked`, but does _not_ decode the response correctly if the header is in all lowercase characters: `transfer-encoding: chunked`

### Test code
This is difficult to test in isolation because a server is required. In this case, I've encountered this issue with a server I do not control that for whatever reason returns a _raw_ response like this (notice the all lowercase transfer-encoding header):
```
% curl -i --raw -H "Accept: application/json" http://example.com/api/state

HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Date: Sun, 09 Sep 2018 23:19:22 GMT
transfer-encoding: chunked
Connection: keep-alive

13
{"pin":8,"state":0}
0
```

The numbers `13` and `0` here are the delimiters of the the first (and only) chunk. When this request is made using the HTTP module, the chunked response is not decoded correctly and these control characters erroneously end up in the response:

```Lua
http.get('http://example.com/api/state', "Accept: application/json\r\n", function(code, response)
  print(response)
end)
```
the above prints:
```
13
{"pin":8,"state":0}
0
```
### NodeMCU version
master

### Hardware
NodeMCU with ESP8266

### Proximate cause
After digging into the source, this appears to be a straightforward fix. I believe the problem lies here:
https://github.com/nodemcu/nodemcu-firmware/blob/8181c3be7aed9f0a0ceb73ac8137c1a519e8a8e9/app/http/httpclient.c#L417

Not being familiar or experienced with C programming, what's the best way to make this comparison case insensitive?

heythisisnate:
Fixes #2482 

HTTP Module: Correctly decode chunked responses if `Transfer-Encoding` header is all lowercase.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -414,7 +414,7 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
 						  body = body + 4;
 					}
 
-					if ( os_strstr( req->buffer, "Transfer-Encoding: chunked" ) )
+					if ( os_strstr( req->buffer, "Transfer-Encoding: chunked" ) || os_strstr( req->buffer, "transfer-encoding: chunked" ) )
 					{
 						int	body_size = req->buffer_size - (body - req->buffer);
 						char	chunked_decode_buffer[body_size];


marcelstoer:
Since we're already using a number of such `os_strstr` calls in that file I assume `strcasestr` is not available? Seems to be part of the SDK's `libc.a`, though.

TerryE:
@marcelstoer, isn't merging in this sort of PR into `dev` before the `master` cut over rather badly out of line with the whole idea of having a freeze pre dev->master merge.

marcelstoer:
Indeed, it's totally out of line. I took the liberty to bend (ok, break) the rules for the following reasons

- it fixes an obvious bug
- it's minimally invasive
- it's practically risk-free
- it's coming from a trusted author

TerryE:
Rules are there to be bent when appropriate.  I did check the change later. On reflection I didn't see it as an impediment to doing the matter drop. Sorry for sounding a bit ratty. 😊

marcelstoer:
> Sorry for sounding a bit ratty.

You didn't (at least for me) - not this time 😜

nwf:
Gack, why wasn't this pushed over to strcasestr?  Does that not work?

TerryE:
> why wasn't this pushed over?

Eh? See [master version:L417](/nodemcu/nodemcu-firmware/blob/3661b8d5eb5b42ed2d5ff51fa8e9628c17270973/app/http/httpclient.c#L417)

TerryE:
Sorry Nathaniel, I misunderstood your point: too terse.  This got pushed because it was in dev at the cut over to master.  It's the process.  I will reply on #2491

PR #2487, opened 2018-09-15
Add Getting Started page

### LINKED ISSUE #2481: Add more/better references for LFS infrastructure
### Missing feature
Better and more links to new LFS image build infrastructure (online service and Docker)
At least add/rework
- Quick start guide does not mention online service and docker.
- the "Compiling Lua on your PC for Uploading" in upload.md mentions LFS at the very end but even I over-read it the first time reading. The section should be restructured and clearly show LFS as favorized way to do it. Also mention the different ways to get an lfs image.
- In "Building the firmware" the LFS section should be more of a teaser and in short explain what to do, As it is in the other sections too.

### Justification
Start using LFS would be easier 

### Workarounds
Find the bits by digging deeper.

marcelstoer:
Fixes #2481, #2431

As discussed earlier in several other issues. Ideally this should land rather quickly and then immediately be pulled over to master to fill the (LFS) documentation gaps.


marcelstoer:
Thank you Terry for the detailed feedback!

TerryE:
@marcelstoer Marcel,  :+1:  to the updates.  The others have now had a couple of days to give feedback, so I suspect that we have had all of the review comments that we are going to get, and I will merge this now.  Feel free to ripple this over to master.

Thanks, this is a valuable addition to the doc set.

PR #2491, opened 2018-09-17
Better fix for HTTP header decode

nwf:
Use strcasecmp rather than multiple strstr tests in httpclient.  Revises #2482.


TerryE:
"Better fix" is a good description. IMO, this module it a very weak implementation with all sorts of issues. We had maybe 10 contributors patching up various issues.  

I personally wouldn't use it under pretty much any circumstances.  For example I would parse the response headers into a table and use that; then there would be no `strcasestr()` calls.

But again other Lua developers use it.  Each to his own, I guess.

PR #2492, opened 2018-09-17
Remove moribund C ds18b20 module

nwf:
Just use Lua speaking OW (via C) instead.


TerryE:
[Moribund](https://dictionary.cambridge.org/dictionary/english/moribund) is _dying_ rather than _dead_.  The problem with not being actually dead is that some apps developers out there are still using it and if we withdrawn and they then update their firmware -- for example to use LFS -- then they will suddenly discover that their applications no longer work without doing some coding effort.

I feel what we really need is a retirement process for modules if they have been superseded or don't work and are no longer being maintained.  This should include a warning / notice period of, say, 6 months between initial notice and any retirement that includes this caveat in the documentation.

In the case of the C version of the DS18B20 module, it is slow but it does work and if you are an application developer with a couple of thermos and it is fast enough, then why ask them to change their code?

nwf:
I am content to leave this PR open for whatever retirement period is wished, but I object, aesthetically, to arbitrarily preserving and continuing to advertise no-longer-recommended solutions, especially when they appear to be completely superseded by new alternatives.  Anyone updating to master to get LFS is already acknowledging that they will have to make changes to their code; I think the same is plausible for any update to master, but I suppose a cycle with deprecation warnings is fair.

Let me propose, then, that the next master drop include updated docs for the ds18b20 module pointing at this PR (I'll open another PR, I guess) and that the master drop after that include this removal of old code.

marcelstoer:
Now would be last opportunity to agree on whether to eventually remove the C module and, therefore, to add a note to the documentation before the upcoming master drop. At the same time task a) from my above list could be tackled.

👍 for deprecating & removing from my side

nwf:
Now that #2581 has been merged for some months, should I revisit this for this upcoming master drop or leave it for the next one?

marcelstoer:
I personally favor the latter.

marcelstoer:
@nwf could you please resolve the conflicts so we can add this to `dev`?

nwf:
Should be better now.  Yes?

Issue #2504, opened 2018-10-03
Optimising access to ROTables


TerryE:
### Background for Enhancement

One of the eLua memory optimizations is to move the list of base functions and  installed libraries out of `_G` and into ROM-based lists. An `__index` function  has been added to the `_G` metatables, so that any global misses are also resolved against the base function list then the library ROM list.

The upside is a RAM saving, but the downside is a material runtime hit when accessing ROtables. The average access time for a variable cached in a `local` is **3 µSec**, and for looking up one in `_G` is **8 µSec**. However resolving ROM entries is a lot slower: **43 µSec** in the case of the base functions such as `print()` and **132 µSec** for library entries.

Incidentally this **132 µSec** time is perhaps 5× better earlier builds than pre LFS versions since the string constants were byte-packed at `-Os` and `c_strcmp()`function compares the target key against each unaligned ROM table entry, thus constantly triggered unaligned exception handler. 

@pjsg  and I found that switching to `-O2` word-aligned the ROM-based string contents meant that the flash-aware optimised ROM-based `strcmp()` routine does proper word-aligned data fetches in this case.

Even so, the implementation is pretty dumb in that the code always does linear table scan using a pairwise `c_strcmp()`, and this annoyed me, I decided to spend a day optimising this code.

-  First the idea of having different structures -- a ROtable for the list of modules and a different one-off structure for the base functions seemed daft.  Why not just combine them into a unified master ROtable?

-  Unifying these into a single ROtable meant that I could drop the __index function and instead use Lua to do this by assigning `_G.__index = <master_ROtable>`

-  Doing this also made half of the lrotable.c functions redundant and I was left with one indexing by a lua string and one by an integer.

-  Since all keyed access to ROTables goes through a single access function, it seemed common sense to stick a small (16×4×2word) RAM lookaside cache in front of these accesses.  One my benchmarks, this gives better than 99.5% hit rate for keyed ROTable access.

[...]

PR #2505, opened 2018-10-04
Optimise ROTable accesses and interface

TerryE:
Addresses Issue #2504

See #2504 for detailed discussion of the background and scope of this patch.  There isn't any change at an application level, apart from making ROTable access times comparable to RAM Table access times.  The code changes therefore don't impact the documentation at `docs/en/*` excepting the following points:

-  We might want to mention the speedup in the FAQ, but this is due a major update anyway.
-  We might want to document how you can list off the installed ROM libraries by enumerating `pairs(_G.__index)`.

### DIFF
--- a/app/lua/lrotable.c
+++ b/app/lua/lrotable.c
@@ -9,77 +9,139 @@
 #include "lobject.h"
 #include "lapi.h"
 
-/* Local defines */
-#define LUAR_FINDFUNCTION     0
-#define LUAR_FINDVALUE        1
+#define ALIGNED_STRING (__attribute__((aligned(4))) char *)
+#define LA_LINES 16
+#define LA_SLOTS 4
+//#define COLLECT_STATS
 
-/* Externally defined read-only table array */
-extern const luaR_table *lua_rotable;
+/*
+ * All keyed ROtable access passes through luaR_findentry().  ROTables
+ * are simply a list of <key><TValue value> pairs.  The existing algo
+ * did a linear scan of this vector of pairs looking for a match.
+ *
+ * A N×M lookaside cache has been added, with a simple hash on the key's
+ * TString addr and the ROTable addr to identify one of N lines.  Each
+ * line has M slots which are scanned. This is all done in RAM and is
+ * perhaps 20x faster than the corresponding random Flash accesses which
+ * will cause flash faults.
+ *
+ * If a match is found and the table addresses match, then this entry is
+ * probed first. In practice the hit-rate here is over 99% so the code
+ * rarely fails back to doing the linear scan in ROM.
+ *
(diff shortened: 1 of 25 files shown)


app/lua/lrotable.c
```diff
@@ -9,77 +9,139 @@
...
+ * Note that this hash does a couple of prime multiples and a modulus 2^X
+ * with is all evaluated in H/W, and adequately randomizes the lookup.
+ */
+#define HASH(a,b) (519*((size_t)(a)>>4) + 17*((size_t)(b)>>4))
+
+static struct {
+  unsigned hash;
+  unsigned addr:24;
+  unsigned ndx:8;
+} cache[LA_LINES][LA_SLOTS];
```
jmattsson: Eeep! That's half a K right there, isn't it? :(
Could this be an optional feature, so those really tight on RAM can opt for the performance hit instead?

app/lua/lrotable.c
```diff
@@ -9,77 +9,139 @@
...
+        !c_strcmp(pentry[j].key.id.strkey, strkey)) {
+        if (ppos)
+          *ppos = i;
+        return &pentry[j].value;
+      }
+    }
+    /*
+     * The invariants for 1st word comparison are deferred to here since they
+     * aren't needed if there is a cache hit. Note that the termination null
+     * is included so a "on\0" has a mask of 0xFFFFFF and "a\0" has 0xFFFF.
```
jmattsson: I like this comment :)

app/lua/lrotable.c
TerryE: That was the ½K ☺️ I can make move it from static to dynamic, but we'll always need it - or at least the cache if we want fast ROTable access. I can half the line size with my next step on this journey. 

I'll also do the claw- back on DNS on this patch so we don't lose net RAM.

TerryE:
@jmattsson, the lookaside cache seems to work well. "How big should we make it" is a useful Q to ask.   8 x 4 seems to be a good compromise, and most real life Lua applications rarely use anything like 32 ROTable fields, so the hit rates should be pretty good -- in the 99% ballpark, which effectively eliminates the search overhead for normal field access.

We still have this issue of cases where the VM does existence checks for fields such as `__metatable` and `__index`.  I can add some extra instrumentation to my `luac.cross -e` build to see if this is an issue in practice.

app/lua/lrotable.c
jmattsson: No don't make it dynamic. This is something which if it's in use, it should be statically allocated to ensure it's there when needed. Plus it makes it easier to reason about other RAM usage.

If you can reclaim RAM from elsewhere so the net available stays the same, that would be Very Good :) And if half the size is enough for the common scenarios, that sounds good too!

Issue #2507, opened 2018-10-07
esp32 serial console doesn't show output until enter pressed


tomsci:
### Expected behavior

When using the interactive Lua console on the serial port, on ESP8266's you get the characters echoed while you type.

### Actual behavior

On the ESP32 port, you see nothing while typing, then the entire line is printed correctly when you press enter.

### Test code

N/A

### NodeMCU version

dev-esp32 branch, as of 2018-10-07, bf5808495ce5265916883b2e15f095241ea052bc

### Hardware

Adafruit HUZZAH32 ESP32 Feather (using SiLabs CP2104 USB-to-serial chip)

### Notes

I have a patch which fixes this, it seems to be an issue with the changed libc implementation defaulting to buffered I/O on stdout and the solution is to configure it to be unbuffered. PR to follow.

PR #2508, opened 2018-10-07
Set stdout to unbuffered

tomsci:
Fixes issue #2507. 

- _Actually for the dev-esp32 branch_
- _Tested on Adafruit Feather esp32_
- _No documentation changes required_

Not sure if `lua_main()` is the best place to put this code, happy to move it earlier/later in the setup sequence if desired.

### DIFF
--- a/components/base_nodemcu/user_main.c
+++ b/components/base_nodemcu/user_main.c
@@ -165,6 +165,7 @@ void app_main (void)
 #endif
 
   console_init (&cfg, input_task);
+  setvbuf(stdout, NULL, _IONBF, 0);
 
   nodemcu_init ();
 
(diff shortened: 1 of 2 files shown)


devsaurus:
Thanks for the fix - I wasn't aware of this behavior.

For the location I'd like to avoid platform specific code inside the Lua code tree. If possible, please add the fix to [`app_main()`](https://github.com/nodemcu/nodemcu-firmware/blob/80e6cc1e7e9eaf8fc3e2a8b250f89e4214cd7525/components/base_nodemcu/user_main.c#L147). Probably after the call to `start_lua()`.

tomsci:
Have moved it to app_main just after console_init, on the basis that the earlier the better is best practice for configuring the stdio buffers.

PR #2519, opened 2018-10-12
Integrate prebuilt toolchains

ryanplusplus:

When I first cloned `dev` and attempted to build the toolchain per the instructions I was not enthused about all of the setup and then managed to fail to build the toolchain. This was frustrating enough that I didn't bother to dig further. I was, however, able to simply clone and build `dev-esp32` because it includes @jmattsson's pre-built toolchains and this led to me poking around and making a few small contributions.

This change includes the pre-built toolchain as a submodule on `dev` at the same path as on the `dev-esp32` branch and updates the makefile to use the pre-built toolchains. Because the pre-built toolchains are added at the end of the path, users with their own toolchain on the path will continue to use this over the integrated toolchain. I hope that this will reduce the barrier to entry for users to build their own firmware locally and ultimately lead to more contributions.


Makefile
```diff
@@ -78,6 +78,7 @@ else
 	else
 		ESPPORT = $(COMPORT)
 	endif
+	export PATH := $(PATH):$(TOP_DIR)/tools/toolchains/esp8266/bin/
```
ryanplusplus: Note that this is exported only to shells that `make` creates to run commands; it does not escape and affect the shell that invoked `make`.

sonaux:
I used [esp-open-sdk](https://github.com/pfalcon/esp-open-sdk) before as a source of toolchain. For last 2 weeks I am using this patch and it works ok.

devsaurus:
@jmattsson and I had an off-line discussion quite some months ago about the same topic. Should we dump https://github.com/nodemcu/nodemcu-firmware/blob/master/tools/esp-open-sdk.tar.xz and  reference the toolchain as a submodule?
The (yet unsolved) issues were:
- The linked repo contains builds for both esp8266 and esp32. This doubles up bandwidth and storage space compared to the current binary blob.
- Updating binaries for either toolchain adds even more space&bandwidth. For both esp8266 and esp32.

If I recall correctly, the conclusion back then was that adding the toolchains repo as a submodule is no improvement over the current situation on esp8266 branch after all.
What's needed IMO:
* Separate branches or even separate repos for esp8266 and esp32
* Have shallow (`--depth 1`) submodule clones of the toolchains

Discussions got stuck at some point because github doesn't allow fetching a random single SHA1 commit which is implied by shallow clones (github only allows this for HEAD of master). It reported "error: Server does not allow request for unadvertised object".

sonaux:
"Today I learned" that esp-open-sdk always been in tools subdirectory and you do not need to clone esp-open-sdk repository... You just need to know files are there and you need to extract them manually and create some shell script which will put them in PATH and only then you will be able to build the firmware.
When I first tried 'clone and make', I failed. Tried 'apt search xtensa', failed; googled "download xtensa-lx106-elf-gcc", found something... typical junior steps :)
What I want to say, submodule or not, lack of ready-to-use toolchain rises the bar for every potential user. 
I think 90M of binaries is better approach than 570M git submodule. What we need is to add some lines to Makefile to extract esp-open-sdk.tar.xz like it is done for esp-iot-sdk.

ryanplusplus:
@devsaurus When I read your comment regarding the binaries in the tar I had the same thought as @sonaux. I'm not particularly concerned about how the toolchain is packaged and only went this route to match `dev-esp32`. If I close this pull request and create a new one that extracts the tar when building to enable the "clone and make" workflow, would that be an acceptable solution?

ryanplusplus:
@jmattsson if you avoid changing the submodule URL it's (relatively) painless to switch submodules. So instead of having a generically named `toolchain` submodule, you'd have the version added to the name of the submodule. To switch to a new toolchain you'd remove the old toolchain submodule, add the new one, and update the makefile to path to the new submodule. Then, when someone gets the update, it will add the new submodule but leave the old submodule on disk (but not in the git index) and the only real pain is needing to delete the old submodule.

jmattsson:
@ryanplusplus Right. Is it possible to change the submodule name but not the path? Then again, it might not be a terrible idea to have e.g. `toolchain-20180613` and `toolchain-20181031` sitting next to each other to allow easy hopping back and forth in revisions across a toolchain switch. So the main pain for this approach then is the need for individual repos for each toolchain version (unless github support has improved), am I reading this correctly?

ryanplusplus:
> Is it possible to change the submodule name but not the path?

If you are asking whether you can change the submodule URL but keep the path the same (something like `tools/toolchain-20180613`) then yes, you technically can, but it is a pain for users. As far as I know it requires digging into `.git/` to wipe out the cached state from the old URL or (realistically) re-cloning the repo. It's a painful enough process that we generally either have versioned submodule paths or have a submodule that dynamically fetches the correct toolchain repo via an internal `git clone` with some makefile magic.

> So the main pain for this approach then is the need for individual repos for each toolchain version (unless github support has improved), am I reading this correctly?

I think so, but this doesn't seem like it would be very painful because the current toolchain under `tools/` hasn't been changed for 2 years. It's entirely possible that we'd never need to update the toolchain anyway. Thus it's a good idea to have a plan, but I don't anticipate having to execute it very often.

devsaurus:
> I don't suppose github has gained the support for shallow clones

No, we can't fetch shallow clones from branches. I just retried that and again ended up at the same blocker that github doesn't allow fetching by commit id. Following the discussions in related issues, it all seems to be about server-side load (or the avoidance thereof).
Consequentially, the whole concept of shallow clones is kind of absurd if one can only reference the *current* commit at HEAD of master. Consider the long delays between drops on our `master` branch. If it had a shallow clone of toolchains, then building from `master` would break instantly when the referenced toolchains repo gets a new commit.

I get the impression that we're barking up the wrong tree when we use a system for revision control and have to work around its revision control "feature" at the same time. Maybe `make` should just download a .tar.gz blob from a particular release in the pre-built toolchains repo.

jmattsson:
> If you are asking whether you can change the submodule URL but keep the path the same (something like `tools/toolchain`) then yes, you technically can, but it is a pain for users. As far as I know it requires digging into `.git/` to wipe out the cached state from the old URL or (realistically) re-cloning the repo. It's a painful enough process that at work we generally either use versioned submodule paths or have a submodule that dynamically fetches the correct toolchain repo via an internal `git clone` with some makefile magic.

That sort of thing has been my past experience, so I'm not surprised it hasn't changed. Thanks for confirming.

> this doesn't seem like it would be very painful because the current toolchain under `tools/` hasn't been changed for 2 years. It's entirely possible that we'd never need to update the toolchain anyway. Thus it's a good idea to have a plan, but I don't anticipate having to execute it very often.

During my time with nodemcu I believe we've updated the esp8266 toolchain once or maybe twice. The esp32 toolchain is a bit more of a moving target, with actual compiler fixes and features being needed to bring in, hence why I wrote the script to build the toolchains easily.

> I get the impression that we're barking up the wrong tree when we use a system for revision control and have to work around its revision control "feature" at the same time. Maybe `make` should just download a .tar.gz blob from a particular release in the pre-built toolchains repo.

This is certainly also an option. We already do it for the SDK. Main question in this approach is "where do we host the .tar.gz (or .tar.xz) file?"

ryanplusplus:
> This is certainly also an option. We already do it for the SDK. Main question in this approach is "where do we host the .tar.gz (or .tar.xz) file?"

You can host binaries using releases in GitHub.

jmattsson:
Fair point, the file size would be within the limits easily these days.

So it sounds like the favoured approach then is for me to set up a new repo which only has the toolchain build script(s), augment said script to package up each toolchain in a .tar.xz (far better compression than .gz), and then manually (can it be automated?) upload those as a release.

On the nodemcu repo side, someone would then add the necessary Makefile magic to fetch the correct toolchain bundle, extract it, and set the path before continuing on its merry way with the usual stuff.

ryanplusplus:
@jmattsson That sounds like a plan to me. One thing we need to make sure we have covered is that there needs to be some way for the make magic to know whether or not the toolchain is current. At work we include a text file containing the version in the toolchain blob that make can key off of to know whether a new blob needs to be downloaded.

We could probably also do this just by looking at the filename of the archive if we 1) commit to leaving it on disk after unpacking and 2) assume that the unpacked toolchain matches the downloaded archive.

jmattsson:
I'd suggest including a datestamp in the toolchain directory name, and then have that datestamp as a variable in the Makefile, similar to we have for the SDK version. That way we can simply have a dependency/target like:
```
tools/toolchain-$(TOOLCHAIN_VER)/esp8266/bin/xtensa-lx106-elf-gcc:
  # fetch into cache/
  # extract & touch
```

Then a toolchain upgrade from the nodemcu repo's point of view is simply changing the TOOLCHAIN_VER variable to the new value.

ryanplusplus:
That's fair. It makes it more challenging to remove other toolchains from the disk but that may not be a real concern.

jmattsson:
If we want we could also provide a `clean-old-toolchains` target that does it. Over all I see it as a benefit to not need to re-extract the toolchain if switching between branches/commits using different toolchain versions.

jmattsson:
Right... so is the consensus then that I should find some time to do the toolchain side of this?

jmattsson:
Alright, does someone want to take [this](https://github.com/jmattsson/esp-toolchains/releases/tag/20181105.0) for a spin? @ryanplusplus I'm looking at you ;)

ryanplusplus:
@jmattsson try this out: https://github.com/nodemcu/nodemcu-firmware/pull/2545

jmattsson:
Closing PR due to being superseded by #2545

PR #2526, opened 2018-10-18
u8g2 i2c multibus support

sonaux:

u8g2 HAL have hardcoded i2c bus id=0. The patch removes this constraint.
I used this patch while developing new i2c driver.  Works with original driver too.
No documentation changes needed.

### DIFF
--- a/app/platform/u8x8_nodemcu_hal.c
+++ b/app/platform/u8x8_nodemcu_hal.c
@@ -9,6 +9,7 @@
 #include "c_stdlib.h"
 
 #include "platform.h"
+#include "cpu_esp8266.h"
 
 #define U8X8_USE_PINS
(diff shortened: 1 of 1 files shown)


app/platform/u8x8_nodemcu_hal.c
```diff
@@ -9,6 +9,7 @@
 #include "c_stdlib.h"
 
 #include "platform.h"
+#include "cpu_esp8266.h"
```
devsaurus:
```suggestion
```
We get this by including `platform.h`, don't we?

sonaux: You're right, I didn't check that platform.h already has this include.

marcelstoer:
@devsaurus thanks for adding the labels

PR #2528, opened 2018-10-22
Add examples to the "bit" module documentation

adrianomelo:

I missed having some examples for the `bit` module when I was checking the available operators. This PR request adds examples to this module.

### DIFF
--- a/docs/en/modules/bit.md
+++ b/docs/en/modules/bit.md
@@ -19,6 +19,12 @@ Arithmetic right shift a number equivalent to `value >> shift` in C.
 #### Returns
 the number shifted right (arithmetically)
 
+#### Example
+```lua
+bit.arshift(3, 1) -- returns 1
+-- Using a 4 bits representation: 0011 >> 1 == 0001
+```
+
 ## bit.band()
 
 Bitwise AND, equivalent to `val1 & val2 & ... & valn` in C.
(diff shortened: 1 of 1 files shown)


marcelstoer:
Looks good, thanks. However, I would also confirm in a comment what the expected result is. Pattern:

```lua
print(bit.xxx(a, b)) -- prints 'n'
```

adrianomelo:
Hey @devsaurus and @marcelstoer, thanks for reviewing the pull request and accepting the contribution. You guys are doing a fantastic job here. I started using the project last week, and everything works out-of-the-box. 

I agree with you @marcelstoer. The examples can have a better format. Maybe we can list some of the options and choose which one is better.

```lua
-- #1 (current)
print(bit.xxx(a,b) == c)

-- #2
print(bit.xxx(a,b)) -- prints c

-- #3
bit.xxx(a,b) -- returns c

-- #4
-- 1 is 0001, 2 is 0010 and it returns 0100 
bit.xxx(1, 2) -- returns 4
```

I like using a simplified binary representation of numbers (fourth option). However, I am not sure if it helps.

marcelstoer:
Thanks for the 💐I like the 4th option best as it's the most expressive.

marcelstoer:
@adrianomelo just to be sure (expectation management on my side), _you_ will update the documentation wrt Lua comments, right?

adrianomelo:
Hi @marcelstoer, yes, I will update it. Sorry for the delay. I should be able to work on it later today.

PR #2533, opened 2018-10-28
esp32: Add node.uptime() function using esp_timer_get_time()

tomsci:
Also added larger limit before wrapping if Lua's number type supports it.

_no github issue for this as far as I could see_.

    - dev-esp32 branch actually
- [X] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well
    - Ran `tmr.create():alarm(60000, tmr.ALARM_AUTO, function() print(tmr.now()) end)` on both integer and floating-point builds for > 35 mins, watched the integer one wrap and the fp one not

`tmr.now()` is absent in the esp32 branch, presumably because it used to call `system_get_time()` which is now deprecated? `esp_timer_get_time()` works just as well now, in fact better since it's 64-bit.

### DIFF
--- a/components/modules/node.c
+++ b/components/modules/node.c
@@ -279,6 +279,23 @@ static int node_osprint (lua_State *L)
 }
 
 
+static int node_uptime(lua_State *L)
+{
+  uint64_t now = esp_timer_get_time();
+#ifdef LUA_NUMBER_INTEGRAL
+  lua_pushinteger(L, (lua_Integer)(now & 0x7FFFFFFF));
+  lua_pushinteger(L, (lua_Integer)((now >> 31) & 0x7FFFFFFF));
+#else
+  // The largest double that doesn't lose whole-number precision is 2^53, so the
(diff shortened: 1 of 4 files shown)


tomsci:
I agree having gettimeofday() and friends exposed in some fashion would be useful, but that's not what I trying to solve with this PR. I use the old `tmr.now()` specifically as a measure of _uptime_ and for calculating elapsed time when debugging. Full RTC/wallclock support is beyond what I was aiming for. (Although having said that, it'll probably end up being the very next thing I need...)

If it doesn't fit in `tmr`, can we perhaps rename it `uptime()` and move it to the `node` module?

Thinking about it some more, it should probably also return the top 32 bits as a second result, on `LUA_NUMBER_INTEGRAL` 32-bit builds, to make wrapping and debouncing easier. (Man roll on Lua 5.3 support, using 5.1 with no integer subtype seems so backwards now!).

PR #2540, opened 2018-11-04
Add ESP32 http module

### LINKED ISSUE #2529: ESP32: http module
**Rationale:**

The IDF has a full HTTP stack included in `esp_http_client.h` and it would be nice to expose this instead of the old http.c/httpclient.c version. Currently there's no `http` module and I assume everyone is using `net` directly.

At the same time, I'd like to reconcile the `http` vs `net/tls` dichotomy, that is to say that the APIs in `http` stop being usable if your body data is large, because only the `net` module offers a multiple-callback model, meaning you have to rewrite your code using `net` and parse the HTTP manually if you hit memory limits using the old `http` module. I'd like to keep the existing simple `http.get()` API but also have a more net-like `on("receive")` style API available as well to gain the benefits of HTTP header parsing etc even when the payload is large.

Does this sound sensible? It may sound like a laundry-list of requests but I'm planning to do the work myself and submit a PR for it (I really need HTTPS support!) so I'm keen to get feedback before I disappear too far down the rabbit hole of implementation.

(I'm aware of #1622 but I don't really want to have to keep doing HTTP header parsing myself given there are APIs in the SDK just waiting to be used for this.)

tomsci:
Fixes #2529.

This PR adds a Lua module for making HTTP and HTTPS requests using the `esp_http_client.h` APIs already present in the Espressif SDK. Enabling this module pulls in a lot of code into the ROM, about 164KB. But since it doesn't (quite) exceed the default 1MB partition size, and the esp8266 equivalent module was included by default, I've made it enabled by default in the menuconfig.

### DIFF
--- a/docs/en/modules/http.md
+++ b/docs/en/modules/http.md
@@ -162,6 +162,15 @@ None
 ## connection:close()
 Closes the connection if it is still open. Note this does not reset any configured URL or callbacks on the connection object, it just closes the underlying TCP/IP socket. If called from within a callback, this aborts the connection, but the connection object must not be reused until the `complete` callback occurs (waiting until after `request()` returns is also allowed, if the request was synchronous). In the case of a connection in asynchronous mode, will error if called from outside of a callback while a request is ongoing, _unless_ the last callback returned `http.DELAYACK` _and_ `connection:ack()` has not yet been called. In such a situation, `connection:ack()` must still be called, after the call to `connection:close()`. An asynchronous connection that is still open but not ongoing (ie the `complete` callback has taken place) may be closed without restriction. 
 
+#### Syntax
+`connection:close()`
+
+#### Parameters
+None
+
+#### Returns
+`nil`
+
 # One shot HTTP APIs
 These APIs are wrappers around the connection object based APIs above. They allow simpler code to be written for one-off requests where complex connection object management is not required. Note however that they buffer the incoming data and therefore are limited by free memory in how large a request they can handle. The EGC can also artificially limit this - try setting `node.egc.setmode(node.egc.ON_ALLOC_FAILURE)` for more optimal memory management behavior. If however there is physically not enough RAM to buffer the entire request, then the connection object based APIs must be used instead so each individual `data` callback can be processed separately.
 


marcelstoer:
> What's the correct thing to do for esp32-only PRs?

I simply wouldn't tick it. One of us committers then usually adds the ESP32 label (right column).

> happy to make the http module set something by default instead?

That's what we (have to) do for ESP8266: https://github.com/nodemcu/nodemcu-firmware/blob/master/app/http/httpclient.c#L215. I just think distinguishing our firmware from all the other ESP-IDF firmware wouldn't harm even though the actual HTTP stack is the same. Curious to find out what Arnim prefers.

tomsci:
It appears that according to [RFC2616](http://www.ietf.org/rfc/rfc2616.txt) setting it to `NodeMCU ESP32` with a space in would actually declare two products (ie this is a "NodeMCU" agent with no version specified, and a "ESP32" agent with no version specified). Which isn't quite right. It also means that the `esp_http_client` default of `ESP32 HTTP Client/1.0` is just completely wrong (it should probably have been `ESP_HTTP_Client/1.0`).

Short answer, `NodeMCU_ESP32` or `NodeMCU (ESP32)` would probably be a better choice :)

(The latter has the benefit that if we ever wanted to rationalise the 8266 user agent it could be share the same product and be `NodeMCU (ESP8266)`.)

tomsci:
Thinking about it some more, `ESP32 (NodeMCU)` might be an option too, given that the (weird and wrong) default in `esp_http_client` does at least include a product `ESP32`. Although if Espressif bring out a new chip called the ESP33 is their SDK's default still going to say ESP32? Perhaps it's best not to put ESP32 in the product, and stick with `NodeMCU (ESP32)`.

devsaurus:
I've been hunting a memory leak in my application for several days now and it *seems* to be related to the `http` module. The suspected leak is reproducible with the following code:
```lua
print(node.heap())
c = http.createConnection("http://server/binary", {async=true,bufsz=2048})
c:on("connect", function() print("connected") end)
c:on("headers", function(status, headers) print("headers received") end)
c:on("data", function(status, data) print(" data "..#data) end)
c:on("complete", function(status) print("finished") end)
c:request()
-- wait until "finished"
c = nil

-- repeat until heap stabilizes after some seconds
print(node.heap())
```

Looping this code a couple of times shows that head is decreased by ~90&nbsp;bytes each time. Synchronous or asynchronous mode is not relevant as far as I can tell

After executing this code, the Lua registry contains strange entries.
```lua
function pt(t)
    for k,v in pairs(t) do print(k,v) end
end

-- debug module has to be enabled in "Core Lua modules" menu
pt(debug.getregistry())
```
results in
```
> pt(debug.getregistry())
1	table: 0x3ffcdbd8
2	function: 0x3ffcd8c4
3	function: 0x3ffcdeb4
4	function: 0x3ffce0b4
5	function: 0x3ffcdd9c
6	function: 0x3ffcddb4
7	8
8	9
tmr.timer	romtable: 0x3f40bb88
ucg.display	romtable: 0x3f406c88
0	7
spi.device	romtable: 0x3f40c778
spi.master	romtable: 0x3f40c700
_LOADED	table: 0x3ffdad34
ledc.channel	romtable: 0x3f40a638
net.tcpsocket	romtable: 0x3f40b440
9	10
http.context	romtable: 0x3f40c0f8
10	11
```

Note the numbers at indices 7, 8, 0, etc. They didn't exist prior to executing `c:request()` and I can't track them down to any code in `http.c`.
They appear to follow a pattern:
1. A single number is visible after `c:request()` finished
2. More numbers follow after `c = nil` is executed
3. Lowest number is always `7`, although it's not the *first* number from 1.
4. Further numbers are increments by 1
5. The max number can be different. in the example above it's 11 - I've seen series up to 17

Can you reproduce this as well?

tomsci:
That's how the Lua registry handles freeing refs isn't it? From memory I think the refs you `luaL_unref` get turned into a linked list of free entries (so ref 7 now contains an integer 8, being the next free ref), so that subsequent `luaL_ref` calls don't need to search through the table for a free ref. I'm not sure when entries get completely removed from the registry table (if ever) but they should get reused next time something needs one. So I think this is probably a red herring? I'll have closer look tomorrow.

devsaurus:
> That's how the Lua registry handles freeing refs isn't it?
> So I think this is probably a red herring?

Maybe, yes. At least I couldn't find any obvious issue.

tomsci:
Before seeing your most recent comment I ran some tests to ensure that a connection didn't leak any memory that Lua's bookkeeping can see, and it doesn't for the given test code. Which means that if there was a leak anywhere, it'd probably be deep in the esp_http_client code.

tomsci:
Great, thanks for merging. I didn't change the default user agent though, do you think we should set it to eg `NodeMCU (ESP32)` rather than `ESP32 HTTP Client/1.0`? I have the change ready to go but was waiting feedback from you or @marcelstoer as to the best thing to set it to.

marcelstoer:
I had given [your respective comment/question](https://github.com/nodemcu/nodemcu-firmware/pull/2540#issuecomment-435706736) a 👍 but I agree it's a non obvious way of conveying support, sorry.

tomsci:
Ah, missed that. No problem, will open a new PR for the user agent change.

devsaurus:
Sorry guys for rushing this. I had the feeling that this PR started to gather dust but obviously missed that open topic.

marcelstoer:
> PR started to gather dust

There are a [few more candidates](https://github.com/nodemcu/nodemcu-firmware/pulls?q=is%3Apr+is%3Aopen+label%3AESP32) that have since accumulated a lot of that 😉

PR #2543, opened 2018-11-05
Adding qrcodegen module for generating QR Codes

tomsci:
    - _Raised against dev-esp32_
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well

I'm not sure if this is a module of any general interest to the NodeMCU community, but I've been using it for the last little while on an e-ink display project so I thought I might as well share. Feel free to close the PR if you don't want it :)

It adds the ability to generate QR Codes suitable for rendering on a display and being read (eg) by a smartphone.

This is a wrapper around the C API in https://github.com/nayuki/QR-Code-generator.git, which copying the example of the `u8g2` component I've added as a submodule. Far as I can tell, it's an excellent implementation (QR Codes are _complicated_!) and the C API uses very little RAM. The code isn't ESP32 specific, but that's what I've been using and testing with, so that's what I'm raising this against in the first instance at least.

Edit: Updated PR with the missing KConfig info!


docs/en/modules/qrcodegen.md
```diff
@@ -0,0 +1,70 @@
...
+| :----- | :-------------------- | :---------- | :------ |
+| 2018-10-20 | https://github.com/nayuki/QR-Code-generator.git integrated by [Tom Sutcliffe](https://github.com/tomsci) | [Tom Sutcliffe](https://github.com/tomsci) | [qrcodegen.c](../../../components/modules/qrcodegen.c)|
+
+This module wraps the [QR Code](https://en.wikipedia.org/wiki/QR_code) Generator API written in C by https://github.com/nayuki/QR-Code-generator.git for producing a QR Code.
+
+## qrcodegen.encodeText()
+Generates a QR Code from a text string. In the most optimistic case, a QR Code at version 40 with low ECC can hold any UTF-8 string up to 2953 bytes, or any alphanumeric string up to 4296 characters, or any digit string up to 7089 characters.
+
+#### Syntax
+`qrcodegen.encodeText(text, minVer, maxVer, ecl, mask, boostEcl)`
```
devsaurus: The high degree of freedom with optional positional parameters would call for a rather complex formal syntax of the API. E.g.:
```
`qrcodegen.encodeText(text[, [minVer][, [maxVer][, [ecl][, [mask][, [boostEcl]]]]]])`
```
How about an optional table?

components/modules/qrcodegen.c
```diff
@@ -0,0 +1,67 @@
...
+
+#include "qrcodegen.h"
+
+// qrcodegen.encodeText(text, minVer, maxVer, ecl, mask, boostEcl)
+// all params except text are optional
+static int encodeText(lua_State *L)
+{
+  const char *text = luaL_checkstring(L, 1);
+
+  int min_ver = luaL_optint(L, 2, 1);
```
devsaurus: `luaL_optint(L, 2, qrcodegen_VERSION_MIN)`? More below.

components/modules/qrcodegen.c
```diff
@@ -0,0 +1,67 @@
...
+  int mask = luaL_optint(L, 5, qrcodegen_Mask_AUTO);
+  bool boost = lua_toboolean(L, 6);
+
+  const size_t buf_len = qrcodegen_BUFFER_LEN_FOR_VERSION(max_ver);
+  uint8_t *result = (uint8_t *)luaM_malloc(L, buf_len * 2); 
+  uint8_t *tempbuf = result + buf_len;
+
+  bool ok = qrcodegen_encodeText(text, tempbuf, result, ecl, min_ver, max_ver, mask, boost);
+  if (ok) {
+    lua_pushlstring(L, (const char *)result, buf_len);
```
devsaurus: This assumes that qrcodegen's result doesn't exceed the maximum string length limit of the Lua VM (4095). Is this correct?

tomsci: The max size of a QRCode is `qrcodegen_BUFFER_LEN_MAX` which is 3918 so we just squeeze in under that limit, although personally I always unset that limit in my init.lua script since I find it's not helpful on the ESP32.

tomsci:
> I struggled a bit to understand the use case in the first place. Would be nice to have a small lua_example to demonstrate how this module interacts with u8g2.

The idea is to generate a QR Code object from a URL or text string, then there are functions to pull out the (x, y) pixel data from that to do whatever with. I guess I could have made `qrcodegen.encodeText()` return a userdata, and have `getPixel()` and `getSize()` methods on that, it might make the API clearer? It was such a thin wrapper with so few functions it didn't occur to me, but in hindsight having encodeText return a string is a bit odd since there's nothing you can do with that data other than pass it to one of the `get*()` functions.

I've never used `u8g2` (it doesn't support the panel I'm using) so I don't have any examples of how to integrate with it I'm afraid.

docs/en/modules/qrcodegen.md
tomsci: I think it'd probably be `qrcodegen.encodeText(text[, minVer[, maxVer[, ecl[, mask[, boostEcl]]]]])` wouldn't it? But yes, it's a bit clunky to write. I was on the fence as to whether to make an options table but you end up having to write a bit more boilerplate without the `luaL_opt*` fns to help so I didn't bother. Maybe we should add some more helper fns for things like options tables to some shared file? Otherwise I'll be duplicating the `check_optint/check_optbool` code from `http.c` in order to do thorough argument validation.

devsaurus:
> `qrcodegen.encodeText()` return a userdata

I'm kind of unbiased here. The module doesn't contain additional attributes or internal state, thus representing the QR code with a plain string is good enough and doesn't restrict any functionality IMO. Even when thinking of conversion to png or xbm.
I consider string vs. object a cosmetic decision at the moment. Might be different if future use cases call for more complex things.

> I've never used `u8g2`

Ok, I mixed that up somehow.

docs/en/modules/qrcodegen.md
devsaurus: :smile: I understood the code the way that `qrcodegen.encodeText("lorem ipsum", , 4)` is accepted.

> Maybe we should add some more helper fns for things like options tables to some shared file?

Yes, agree to that. I too felt to reinvent the wheel more than once in the past. Each time with a different look of course. Though I can't really propose a good solution atm. Maybe just a simple `components/modules/common.c`?

tomsci:
I've updated the PR changing the syntax to `qrcodegen.encodeText(text[, options])` and added a common.h/.c to components/modules with some options helper fns in. Also updated http.c to use that since that's where I took the helpers from. There are probably other modules that could benefit too (?) but I didn't want to drag too much other feature creep into this PR.

PR #2544, opened 2018-11-05
Increase MQTT recv buffer size to support MTU 1500  (#2308)

### LINKED ISSUE #2308: MQTT subscribe - large messages get ignored quietly
### Expected behavior
NodeMCU should receive every message that has been published with a MQTT topic it is subscribed to.

### Actual behavior
If the message is longer than 984 bytes (in my tests with 35 bytes long topic) then such message is not received, it gets ignored quietly.
Maybe the real limit is actually sum of topic length and message length, and it seems like it's less than 1024 bytes.
FYI, the topic max size is 64 kB and message max size is 256 MB per MQTT spec. Of course I don't want an ESP device with 20 kB free heap to process 256 MB message but a hidden/undocumented limit to less than 1 kB is not fun either.

The worst thing here is that no message is received at all, there's no indication that a message got lost. If would be far better (though still not ideal) if it at least returned first part of the message that fit into an internal buffer...

### Test code
```Lua
mqtt_client = mqtt.Client(1, 120, name, passwd)
mqtt_client:on("message", function(client, topic, data) print(#data) end)
mqtt_client:connect(broker, 1883, 0, 0, function(client)  mqtt_client:subscribe({["#"]=1}) end)
```
```bash
$ printf '%*s\n' 1024 | tr ' ' 'a' |  mosquitto_pub -t x -l
```
### NodeMCU version
current master

### Hardware
plain ESP8266-07

stromnet:
Any TCP packet with more than 1024 bytes of payload was silently
dropped. With MTU of 1500 the TCP payload can be up to 1460 bytes
(1500 - 20(IP hdr) - 20(TCP hdr))

Fixes #2308

- [x] I have thoroughly tested my contribution... **But I'm quite a new NodeMCU & MQTT user so I might be unaware of some weird circumstances**

(Not sure what documentation would be appropriate to update, as this is more of a "fix what was expected to work" than anything else)

I've done some experiments and tcpdumps, based on what I wrote here: https://github.com/nodemcu/nodemcu-firmware/issues/2308#issuecomment-435518706

I had some issues with small but rapidly sent messages getting dropped by my ESP12-F on NodeMCU 2.2.1.
The messages where just 52 bytes long (33 byte topic), but sent with 15 consecutive calls to mosquitto_pub (so at most some shell & connect overhead between msgs). Problem was that I was just getting the first one or two messages.

Doing some tcpdump revealed that  rapidly sending small messages to a standard mosquitto broker, to a topic which a NodeMCU is subscribed to, the broker will send the first msg in a separate TCP packet, then (I suppose) it is waiting for the TCP ACK before it goes on to send the next packet/message. By that time, the broker has received all the messages, and the rest of them are sent in one TCP packet with payload length of 1126b. This will not fit in MQTT_BUF_SIZE of 1024, and the full payload is dropped.

Another test, when sending a single message of 990bytes to a 29 bytes topic (=1020) bytes, the TCP payload length equals 1024, and this is properly delivered to the LUA callback. Adding 1 bytes to the message and it is dropped silently.

Sending a message of let's say 2000b, on my network it makes one IP packet of 1500 bytes (max MTU), with a payload size of 1460.
The IP header is at least 20 bytes, and TCP header is at least 20 bytes (=40b), giving us 1460 bytes for payload data in a MTU 1500 network.

Given my understanding of the code, the `MQTT_BUF_SIZE` must be at least 1460 to support a normal local network with a MTU of 1500.

Setting this to 1460 and running the above same tests yields expected results (messages arriving to LUA-land).
---

A sane default for this should thus be 1460, rather than 1024 which is quite non-standard and creates broken behaviour in (what I would expect to be) normal situation

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -17,7 +17,7 @@
 
 #include "user_interface.h"
 
-#define MQTT_BUF_SIZE 1024
+#define MQTT_BUF_SIZE 1460
 #define MQTT_DEFAULT_KEEPALIVE 60
 #define MQTT_MAX_CLIENT_LEN   64
 #define MQTT_MAX_USER_LEN     64


nwf:
It's alarming to see Ethernet MTU entering into a discussion of MQTT, which rides on top of TCP and so has no legitimate excuse to care about such things.  That's not to say that I don't believe you, just that it is a sad reflection of the state of our MQTT implementation if this matters.  Patch seems reasonable enough, tho', so :+1: from me.

stromnet:
One thing which confuses me is how it was able to deliver a message longer than 1460b, which i seemed to see yesterday.. There doesn't look like there is any inter-packet buffering going on so.. I'll take a closer look tonight to see if it actually delivers the whole thing, or just claims to give full msg which then is truncated..

Please wait with the MR :)

stromnet:
Ok, some more testing:
* A MQTT PUB message (29b topic, 1426b data) which is 1460 bytes long, i.e. one TCP packet, will fit into the buffer and is given properly to LUA code.
* A MQTT PUB message (29b topic, 1427b data) which is 1461 bytes long, two TCP packets: In lua this is a string where string.len() reports 1427 bytes, but the last character is wrong.
* A MQTT PUB message (29b topic, 1450b data) which is 1461 bytes long, two TCP packets. LUA string.len() reports 1450 here too.
* Two MQTT PUB message, where the first one fits in one TCP packet, and the second one is partially sent in the same TCP packet: the first packet is given to LUA callback in full, the second is partial (but size reported full)
* Multiple MQTT PUB messages split over multiple TCP packets, where one TCP packet begins with partial continued message, with more messages following: No messages at all are passed to LUA.


So, yeah.. The state of the MQTT implementation isn't quite perfect.

The proper solution should of course be to, in case of a non-finished packet, keep the data until next recv callback, and append the new data to previous data and continue processing if possible. This will of course make code a bit more complex, and memory hungry. The alternative would be to state in docs explicitly "we do not support MQTT messages over XXX and they will be, in full, discarded... and messages which are partially in a continuation TCP frame will be dropped too". Not preferable!

It seems that the websocketclient solves this: https://github.com/nodemcu/nodemcu-firmware/blob/8181c3be7aed9f0a0ceb73ac8137c1a519e8a8e9/app/websocket/websocketclient.c#L357
Is that the proper way to go in NodeMCU module code, using z_calloc and try to fit the data there?
If so, I can give it a try for the mqtt module.

nwf:
The "partially in a continuation TCP frame" is really, *really*, **really** bad; TCP is byte-oriented and will break your messages wherever it pleases.  It'd be one thing to discard the overlong message but resume parsing the TCP stream thereafter; that's merely really, *really* bad.  Unfortunately, it seems like we might be forced into this situation anyway because MQTT says "messages up to 256MB are OK!", there's no mechanism to inform the broker of smaller tolerances, and no amount of cleverness on the nodemcu platform will do anything for us there.  So, I propose:

* MQTT messages up to ~1500 bytes, regardless of how they fragment in TCP, are delivered to the application via callback.
* MQTT messages larger than ~1500 bytes, regardless of how they are fragmented in TCP, will be dropped and trigger a "stream resynchronization" behavior, wherein we continue to pull and discard bytes from the broker until we are caught up with the end of this overlong message, at which point we continue receiving messages.  We may want a Lua callback (e.g. `on("overlong", ...)`) to inform the code that this has happened; it may be of debugging utility to provide the first ~1500 bytes of the message to the callback, too.
* We amend the documentation to discuss this limitation, noting that there are ~10 (ish, haven't checked exactly) bytes of MQTT framing overhead and that the remaining ~1490 bytes are split between the topic and message.

stromnet:
Sounds good!

How about providing a option to the mqtt.Client() constructor with max message length (which defaults to ~1500). It will have to keep a heap buffer between each recv in any way, this way it at least allows the application developer to control max msg len.

I'll see what I can do, this is not really a high prio task for me (actually just found it while doing some Halloween blinky-light-foolaround with my previously unused ESP12F's), but an interesting one nonetheless, and learning some low-level MQTT/NodeMCU seems fun :)

marcelstoer:
I propose we merge this as-is and handle any further improvements @stromnet comes up with in subsequent PRs.

PR #2545, opened 2018-11-06
Download pre-built toolchains

ryanplusplus:
See discussion in #2519 

This integrates the pre-built toolchains from @jmattsson by downloading and extracting the appropriate `tar.xz` archive.


Makefile
```diff
@@ -204,16 +207,32 @@ $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
...
 .PHONY: sdk_patched
 .PHONY: sdk_pruned
+.PHONY: toolchain
 
 sdk_extracted: $(TOP_DIR)/sdk/.extracted-$(SDK_BASE_VER)
 sdk_patched: sdk_extracted $(TOP_DIR)/sdk/.patched-$(SDK_VER)
 sdk_pruned: $(SDK_DIR_DEPENDS) $(TOP_DIR)/sdk/.pruned-$(SDK_VER)
 
+ifeq ($(OS),Windows_NT)
+toolchain:
```
ryanplusplus: For Windows just ignore the pre-built toolchain.

Makefile
```diff
@@ -204,16 +207,32 @@ $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
...
+endif
+
+$(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc: $(TOP_DIR)/tools/toolchains/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz
+	mkdir -p $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/
+	tar -xJf $< -C $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/
+	touch $@
+
+$(TOP_DIR)/tools/toolchains/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz:
+	mkdir -p $(TOP_DIR)/tools/toolchains
+	wget --tries=10 --timeout=15 --waitretry=30 --read-timeout=20 --retry-connrefused https://github.com/jmattsson/esp-toolchains/releases/download/$(TOOLCHAIN_VERSION)/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz -O $@ || { rm -f "$@"; exit 1; }
```
ryanplusplus: This uses all of the same options that are used in `wget`s below. In the long run t would probably be a good idea to extract this as a function.

Makefile
```diff
@@ -204,16 +207,32 @@ $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
...
 
+ifeq ($(OS),Windows_NT)
+toolchain:
+else
+toolchain: $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc
+endif
+
+$(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc: $(TOP_DIR)/tools/toolchains/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz
+	mkdir -p $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/
+	tar -xJf $< -C $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/
```
ryanplusplus: Note that the archive has `esp8266/` at the root so we can apply this to `dev-esp32` and have the toolchains sit nicely side-by-side.

Makefile
```diff
@@ -204,16 +207,32 @@ $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
...
 sdk_patched: sdk_extracted $(TOP_DIR)/sdk/.patched-$(SDK_VER)
 sdk_pruned: $(SDK_DIR_DEPENDS) $(TOP_DIR)/sdk/.pruned-$(SDK_VER)
 
+ifeq ($(OS),Windows_NT)
+toolchain:
+else
+toolchain: $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc
+endif
+
+$(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc: $(TOP_DIR)/tools/toolchains/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz
```
ryanplusplus: The gist here is that `toolchain` depends upon the extracted `*-gcc` which in turn depends upon the archive itself. This assumes that we will never be half-extracted.

Makefile
```diff
@@ -2,6 +2,8 @@
 #
 .NOTPARALLEL:
 
+TOOLCHAIN_VERSION:=20181105.0
```
ryanplusplus: Select the toolchain version here

Makefile
```diff
@@ -78,6 +80,7 @@ else
 	else
 		ESPPORT = $(COMPORT)
 	endif
+	export PATH := $(PATH):$(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/
```
ryanplusplus: If on Linux, add the pre-built toolchain to the path

Makefile
```diff
@@ -161,7 +164,7 @@ CCFLAGS += 			\
 	-nostdlib       \
 	-mlongcalls	\
 	-mtext-section-literals
-#	-Wall
```
ryanplusplus: My editor removes trailing whitespace. I figure we'd rather not have trailing whitespace so I didn't put it back.

Makefile
jmattsson: Ah, it occurs to me that I forgot to include something like `linux-x86_64` in the release bundle name. I had meant to do it, but then I was getting it all done in the background while working on other things and it slipped my mind. What's your view? Should I add it?

jmattsson: I'm not aware of anyone building on Windows without using a Linux docker or such for the build. Maybe @marcelstoer knows more?

I was largely thinking whether there may be a case made for providing e.g. linux-aarch64, linux-armhf or even linux-x86 in the future. I could make it part of the directory name inside the archive easily enough in such cases (as well as the tar.xz name, naturally).

ryanplusplus: Ah, that makes sense.

ryanplusplus:
I am a smidge concerned that not all distributions will support untarring a `tar.xz` without installing something like `xz-utils`. At some point in the past this seems to have been the case for Ubuntu but I am not sure when that changed. If this becomes an actual issue we can use `zip` since we're already using `unzip` to extract the SDK.

Makefile
jmattsson: I've never actually seen a Windows toolchain in the wild. Personally I wouldn't be sad if the Windows_NT ifeq was dropped. It's hinting at supporting something which I don't think has ever been tested. Should we ever need it back it's in the repo history anyway. Just my $.02, I'm not claiming to speak for the other core committers :)

jmattsson:
Even using `zip -9`, it's still a terrible option compared to a `J` on tar:
```
$ ls -lh esp8266.zip  build/toolchain-esp8266-20181105.0.tar.xz 
-rw-r--r-- 1 johny johny 27M Nov  5 12:41 build/toolchain-esp8266-20181105.0.tar.xz
-rw-r--r-- 1 johny johny 93M Nov  6 12:39 esp8266.zip
```

Makefile
jmattsson: Alright, I'm testing a change that will change the filename to `toolchain-esp8266-linux-x86_64-<releasename>.tar.xz` and the internal directory from `esp8266` to `esp8266-linux-x86_64-<releasename>`. I think that should provide full co-existence support.

I wonder if I could get $work to get me a Threadripper 2990WX build box... compiling compilers is so tedious >.>

jmattsson: Okay, it is [done](https://github.com/jmattsson/esp-toolchains/releases/tag/linux-x86_64-20181106.0).

ryanplusplus: Everything should be up-to-date with your changes.

By the way it made things much nicer to have the directory in the tar be `esp8266-$(PLATFORM)-$(VERSION)/` instead of `esp8266/` so that `make` doesn't have to insert an extra layer :+1:

Makefile
```diff
@@ -204,16 +207,32 @@ $(BINODIR)/%.bin: $(IMAGEODIR)/%.out
...
+toolchain:
+else
+toolchain: $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc
+
+$(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/esp8266/bin/xtensa-lx106-elf-gcc: $(TOP_DIR)/tools/toolchains/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz
+	mkdir -p $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/
+	tar -xJf $< -C $(TOP_DIR)/tools/toolchains/$(TOOLCHAIN_VERSION)/
+	touch $@
+
+$(TOP_DIR)/tools/toolchains/toolchain-esp8266-$(TOOLCHAIN_VERSION).tar.xz:
```
jmattsson: Actually, could I get you to download this into the top-level `cache/` directory instead, right next to the SDK zip? That way one only has to `rm cache/*` to ensure one gets a fresh copy of everything downloaded.

ryanplusplus: Sure, didn't know that one existed.

jmattsson:
Ah, I just realised our Travis CI builds would still be using the in-repo toolchain rather than this. In theory everything should work if we remove the `install` block in the `.travis.yml` file now. And thanks to the toolchain being downloaded into `cache/`, it won't need to be re-downloaded all the time. Anyone have time to check out the Travis side?

ryanplusplus:
@jmattsson I updated `.travis.yml` and removed the the old toolchain tarball and everything appears to be working smoothly.

jmattsson:
Oh sweet, thanks!

ryanplusplus:
Thanks @devsaurus and @jmattsson for your help and time

marcelstoer:
I'm a bit late to the party, sorry. Cool "the logic" is now in the `Makefile` rather than `.travis.yml` and a relief to see another bin dumped from the repo - at the cost of another external dependency, @jmattsson don't let us down 😉.

Can you hold back with the merge for a few more days? I need to adjust and release https://github.com/marcelstoer/docker-nodemcu-build/blob/master/build#L57-L66 and  https://github.com/marcelstoer/nodemcu-custom-build/blob/master/.travis.yml#L61-L62 first I guess.

HHHartmann:
Mybe it would also be possible to leave the toolchain tarball in place for some time to assure a smoth transition in docker, so not everybody has to update their images immediately.

marcelstoer:
Ok, one down...Docker is ready, my smoke tests passed: https://github.com/marcelstoer/docker-nodemcu-build/commit/90301adaf5767b1dc68edc6d10fb92828e625af0

marcelstoer:
Two down...it's not 100% clean but good enough: https://github.com/marcelstoer/nodemcu-custom-build/commit/3e809cc105d1aa16997ac7b4280b66627714181a

ryanplusplus:
@marcelstoer are there more updates needed or is this ready to go?

ryanplusplus:
@marcelstoer @jmattsson @devsaurus any more blockers on this?

PR #2571, opened 2018-11-25
MQTT: handle large/chunked/fragmented messages properly

stromnet:
If a message spans multiple TCP packets it must be buffered before delivered to LUA. Prior code did not do this at all, so this "patch" really adds proper handling of fragmented MQTT packets.
This could also occur if multiple small messages was sent in a single TCP packet, and the last message did not completely fit in that packet.

For in-depth description of the problem, please see discussion in #2544 (avoiding repeating it here)

Introduces a new option to the mqtt.Client constructor: max_publish_length which defaults to 1024    
Also introduces a new 'overflow' callback.

Fixes issue #2308, and is a proper fix for PR #2544 (ping @nwf and @marcelstoer who was involved in that PR).

I opted to straighten up the use of uint16_t instead of int and some similar changes. I deliberately selected uint16_t as max buffer length, with the assumption that no platform where NodeMCU runs on will be able to handle anything above 64k. Let me know if that is an bad assumption :)

All in all, this is quite a big commit, and as a first-time commiter on NodeMCU, you may very well find some valid concerns and input on how to improve it. Feel free to give pointers :)

### DIFF
--- a/app/modules/mqtt.c
+++ b/app/modules/mqtt.c
@@ -46,16 +46,31 @@ typedef struct mqtt_event_data_t
 #define RECONNECT_POSSIBLE 1
 #define RECONNECT_ON    2
 
+typedef enum {
+    MQTT_RECV_NORMAL,
+    MQTT_RECV_BUFFERING_SHORT,
+    MQTT_RECV_BUFFERING,
+    MQTT_RECV_SKIPPING,
+} tReceiveState;
+
 typedef struct mqtt_state_t
 {
   uint16_t port;
   uint8_t auto_reconnect;   // 0 is not auto_reconnect. 1 is auto reconnect, but never connected. 2 is auto reconnect, but once connected
   mqtt_connect_info_t* connect_info;
-  uint16_t message_length;
-  uint16_t message_length_read;
   mqtt_connection_t mqtt_connection;
   msg_queue_t* pending_msg_q;
+
+  uint8_t * recv_buffer; // heap buffer for multi-packet rx
+  uint8_t * recv_buffer_wp; // write pointer in multi-packet rx
+  union {
+      uint16_t recv_buffer_size; // size of recv_buffer
+      uint32_t recv_buffer_skip; // number of bytes left to skip, in skipping state
+  };
+  tReceiveState recv_buffer_state;
+
 } mqtt_state_t;
+
 
 typedef struct lmqtt_userdata
 {
@@ -65,6 +80,7 @@ typedef struct lmqtt_userdata
   int cb_connect_fail_ref;
   int cb_disconnect_ref;
   int cb_message_ref;
+  int cb_overflow_ref;
   int cb_suback_ref;
   int cb_unsuback_ref;
   int cb_puback_ref;
@@ -80,6 +96,9 @@ typedef struct lmqtt_userdata
   ETSTimer mqttTimer;
   tConnState connState;
 }lmqtt_userdata;
+
+// How large MQTT messages to accept by default
+#define DEFAULT_MAX_MESSAGE_LENGTH 1024
 
 static sint8 socket_connect(struct espconn *pesp_conn);
 static void mqtt_socket_reconnected(void *arg, sint8_t err);
@@ -109,6 +128,13 @@ static void mqtt_socket_disconnected(void *arg)    // tcp only
       call_back = true;
     }
   }
+
+  if(mud->mqtt_state.recv_buffer) {
+    c_free(mud->mqtt_state.recv_buffer);
+    mud->mqtt_state.recv_buffer = NULL;
+  }
+  mud->mqtt_state.recv_buffer_size = 0;
+  mud->mqtt_state.recv_buffer_state = MQTT_RECV_NORMAL;
 
   if(mud->mqtt_state.auto_reconnect == RECONNECT_ON) {
     mud->pesp_conn->reverse = mud;
@@ -178,9 +204,9 @@ static void mqtt_socket_reconnected(void *arg, sint8_t err)
   NODE_DBG("leave mqtt_socket_reconnected.\n");
 }
 
-static void deliver_publish(lmqtt_userdata * mud, uint8_t* message, int length)
+static void deliver_publish(lmqtt_userdata * mud, uint8_t* message, uint16_t length, uint8_t is_overflow)
 {
-  NODE_DBG("enter deliver_publish.\n");
+  NODE_DBG("enter deliver_publish (len=%d, overflow=%d).\n", length, is_overflow);
   if(mud == NULL)
     return;
   mqtt_event_data_t event_data;
@@ -191,13 +217,15 @@ static void deliver_publish(lmqtt_userdata * mud, uint8_t* message, int length)
   event_data.data_length = length;
   event_data.data = mqtt_get_publish_data(message, &event_data.data_length);
 
-  if(mud->cb_message_ref == LUA_NOREF)
+  int cb_ref = !is_overflow ? mud->cb_message_ref : mud->cb_overflow_ref;
+
+  if(cb_ref == LUA_NOREF)
     return;
   if(mud->self_ref == LUA_NOREF)
     return;
   lua_State *L = lua_getstate();
   if(event_data.topic && (event_data.topic_length > 0)){
-    lua_rawgeti(L, LUA_REGISTRYINDEX, mud->cb_message_ref);
+    lua_rawgeti(L, LUA_REGISTRYINDEX, cb_ref);
     lua_rawgeti(L, LUA_REGISTRYINDEX, mud->self_ref);  // pass the userdata to callback func in lua
     lua_pushlstring(L, event_data.topic, event_data.topic_length);
   } else {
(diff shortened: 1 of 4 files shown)


app/modules/mqtt.c
```diff
@@ -281,11 +310,109 @@ static void mqtt_socket_received(void *arg, char *pdata, unsigned short len)
...
+    case MQTT_RECV_NORMAL:
+      // No previous buffer.
+      break;
+    case MQTT_RECV_BUFFERING_SHORT:
+      // Last buffer had so few byte that we could not determine message length.
+      // Store in a local heap buffer and operate on this, as if was the regular pdata buffer.
+      // Avoids having to repeat message size/overflow logic.
+      temp_pdata = c_zalloc(mud->mqtt_state.recv_buffer_size + len);
+      if(temp_pdata == NULL) {
+        NODE_DBG("MQTT[buffering-short]: Failed to allocate %u bytes, disconnecting...\n", mud->mqtt_state.recv_buffer_size + len);
```
nwf: Sorry, I'm not very familiar with this module; will this disconnection be exposed to the Lua callbacks?  (Does the espconn layer call us back on disconnection, for example?)

app/modules/mqtt.c
```diff
@@ -355,19 +482,81 @@ static void mqtt_socket_received(void *arg, char *pdata, unsigned short len)
...
+
+      NODE_DBG("MQTT_DATA: msg length: %u, buffer length: %u\r\n",
+           message_length,
+           in_buffer_length);
+
+      if(message_length > mud->connect_info.max_message_length) {
+        // The pending message length is larger than we was configured to allow
+        NODE_DBG("MQTT: msg too long: total=%u, deliver=%u\r\n", message_length, in_buffer_length);
+        if (msg_type == MQTT_MSG_TYPE_PUBLISH) {
+          // In practice we should never get any other types..
```
nwf: Even for QoS 2 messages?

app/modules/mqtt.c
```diff
@@ -355,19 +482,81 @@ static void mqtt_socket_received(void *arg, char *pdata, unsigned short len)
...
+          // Just put what we have on heap, and place in state BUFFERING_SHORT.
+          NODE_DBG("MQTT: Partial message received (%u of %d). Buffering\r\n",
+              in_buffer_length,
+              message_length);
+
+          // although message_length is 32bit, it should never go above 16bit since
+          // max_message_length is 16bit.
+          uint16_t alloc_size = message_length > 0 ? (uint16_t)message_length : in_buffer_length;
+
+          mud->mqtt_state.recv_buffer = c_zalloc(alloc_size);
```
nwf: Can I get here with an existing recv_buffer allocated?

docs/en/modules/mqtt.md
```diff
@@ -11,18 +11,39 @@ The client adheres to version 3.1.1 of the [MQTT](https://en.wikipedia.org/wiki/
...
+
+According to MQTT specification the max PUBLISH length is 256Mb. This is too large for NodeMCU to realistically handle. To avoid
+an out-of-memory situation, there is a limit on how big messages to accept. This is controlled by the `max_message_length` parameter.
+In practice, this only affects incoming PUBLISH messages since all regular control packets are small.
+The default 1024 was chosen as this was the implicit limit in NodeMCU 2.2.1 and older (where this was not handled at all).
+
+Any message *larger* than `max_message_length` will be (partially) delivered to the `overflow` callback, if defined. The rest
+of the message will be discarded. Any following messages should be handled properly.
+
+If raising this, heap memory will be used o buffer the received message data between each incoming TCP packet.
```
nwf: "used to buffer"

But it's not just if the limit is raised, right?  It's whenever there might be a misalignment between the TCP and MQTT packetizations of the byte stream.  Ah, I see you address that below.  Maybe reword for clarity?

app/modules/mqtt.c
stromnet: Yes, as soon as the MQTT socket is connected, it registers an disconcb (in `mqtt_socket_connected`) which will trigger the `offline` callback (as bound via `on("offline", ..)`.

app/modules/mqtt.c
stromnet: We could ignore the max_message_length for qos 2 if we want to, but question that follows: at what size should we crash instead of handling it gracefully?

app/modules/mqtt.c
stromnet: Should not be possible (but I may have overseen something of course). If it was in BUFFERING before, and recv_buffer was already allocate, then we should never get down here until  message_length == in_buffer_length (line 361, `in_buffer_length < mud->mqtt_state.recv_buffer_size` check which stops further processing).

docs/en/modules/mqtt.md
stromnet: Indeed! rewrote those two sentences, clearer now?

app/modules/mqtt.c
nwf: Rad.

app/modules/mqtt.c
nwf: I think we should treat QoS 2 just like we treat QoS 1: delivery to "overflow" counts as delivery, even if we had to truncate the message.

docs/en/modules/mqtt.md
nwf: Yes, thank you.

nwf:
That seems like a perfectly fine strategy to me, though it might be even better to just allocate the buffer once and hold on to it for the duration of the connection.  While that does increase heap utilization, it decreases dynamism and the opportunities for surprise OOM due to asynchronous behavior.  (That is, because MQTT messages can arrive at any moment, one has to plan as if the buffer were almost always allocated anyway, I think?)

stromnet:
Yes, MQTT messages could arrive at any moment once subscribed, so definitely adds some unpredictability in heap usage. Allocating max_message_size at Client construction would make heap usage more or less stable regardless of received message sizes. The exception is whatever `espconn` might do, and a few bytes required for ACK/REC/REL/COMP/PINGRESP which are stored temporarily at heap.

If you think that is a stabler solution, I can adjust the code for that too while adding QOS handling on dropped packets.

nwf:
That'd be my preference, but others may have arguments in the other direction.  (In particular, it's not quite true that the allocation can happen at any moment, I think; it can only happen when we are not running a Lua callback, so maybe deferring it would let the Lua heap burst higher?)

Thank you again for doing this; the mqtt module has needed attention for a long time.

stromnet:
Not sure how TCP RX is scheduled vs LUA execution, but yes this only happens when in recv-callback from the espconn, before calling any LUA callback.

A counter-argument: most clients will receive at most 1024b messages (as this is the default), and most likely not receive messages so quick that they will be crammed into same TCP segment and risking being fragmented. For those applications, we would waste 1024b of heap for no reason at all (it would never be used).

app/modules/mqtt.c
nwf: Yeah, not ACKing is going to be unfortunate.  It should be sufficient to just buffer the partial header and recombine it with the rest of the partial header that comes in a later packet, yes?

nwf:
That's fair, and given that it's the current strategy, probably best left as is and we can revisit later.

app/modules/mqtt.c
stromnet: Yes, as far as I know we can ACK as soon as we have the fixed+variable headers (topic + packet ID).
I guess that it should be done after we have "received" (skipped) the full message, not when we have the header(s), so we need to keep it until finished "receiving" the message.

docs/en/modules/mqtt.md
```diff
@@ -11,18 +11,42 @@ The client adheres to version 3.1.1 of the [MQTT](https://en.wikipedia.org/wiki/
...
 - `keepalive` keepalive seconds
 - `username` user name
 - `password` user password
 - `cleansession` 0/1 for `false`/`true`. Default is 1 (`true`).
+- `max_message_length`, how large messages to accept. Default is 1024.
 
 #### Returns
 MQTT client
 
+#### Notes
```
marcelstoer: Thanks for the nice write-up! Well done.

app/modules/mqtt.c
nwf: That sounds right.  I suspect some brokers would be... surprised if they got the ACK back before they were done sending the message!

stromnet:
I'll see if I can spend some time on it this evening.

Additional testing is welcome though.. I've only tested with mosquitto on local WLAN with very basic code.. :)

app/modules/mqtt.c
stromnet: Pushed a new commit which ACKs overflowed messages, but doing this as soon as packet ID is known (if needed going into "short-buffer" mode first).
It does not wait for the final TCP segment to arrive, to keep things simple. Avoids having to keep more buffers & refs & states etc, and in most cases the sending broker will most likely have sent all of the PUBLISH to the OS socket anyway, being ready for an ACK.

nwf:
Been doing some testing this evening.  Thus far it's handled well!  I do have what might be a silly question, tho'.  Not adjusting the default settings, messages up to 996 bytes in size come through the on("message") callback, while larger messages up to 1432 bytes come through the on("overflow") callback without truncation.  Larger messages are truncated to 1432 (or 1430, I've also seen, when there's nonzero QoS involved, or 1429 with really big messages; I suspect the exact numbers don't matter).  I kind of expected there to be one threshold for behavioral changes, not two?

But from a brief poking, it looks like QoS 1 and 2 are both handled fine, too! :smile: :tada:

stromnet:
Good to hear! :) 

Ah, message length should of course be clarified in docs.. It's the size of the full message, aka header + variable header + payload. For a Publish[1], the header is 1 bytes + 1-4 bytes (Remaining Length, size depends on length [2]). Then we have the variable header which is 2 bytes + N bytes of topic, and then if QoS is non-zero, 2 bytes of packet ID. Followed by payload bytes.

So, the exact numbers should matter, but you have to take into account the length of topic, and also qos vs no qos.

I'll try to add something in docs later today!

1. http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718037
2. http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023

marcelstoer:
Gentlemen, thanks for your effort!

Can I invite you to briefly scan the [open MQTT issues](https://github.com/nodemcu/nodemcu-firmware/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+mqtt) and tag those (by leaving a comment pointing here) which might now be fixed?

PR #2577, opened 2018-11-29
Port i2s, spi_master, and ledc to helper functions for option tables

devsaurus:

#2543 added helper functions to get values from option tables. In a first round I've ported modules `i2s`, `spi_master`, and `ledc`.
This brought up 2 new use cases:
* Check for a string option
* Check for a mandatory option table entry

This PR adds function `opt_checklstring()` to support the first use case and tweaks `opt_checkint_range()` to support the latter.

There are several ways to implement mandatory table items and my proposal is as follows:
1. `opt_checkint_range()` checks range on the effective result. I.e. not only the table entry (if existing) but also the default value if the entry doesn't exist.
1. The module code controls optional vs. mandatory:
    - Optional: default value is within range
    - Mandatory: default value is out of range
1. In the "default value is out of range" case, the range check throws an errors when the table entry is missing. It's mandatory for the user to supply a value.

@tomsci may I ask you for your feedback on these additions?

### DIFF
--- a/components/modules/ledc.c
+++ b/components/modules/ledc.c
@@ -6,6 +6,8 @@
 
 #include "driver/ledc.h"
 
+#include "common.h"
+
 typedef struct {
   int timer;
   int channel;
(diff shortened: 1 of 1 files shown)


tomsci:
No other comments, 👍 to the changes to opt_checkint_range for optional vs mandatory.

devsaurus:
> I did have some other comments on this PR

Sorry, I didn't mean to ignore them :disappointed: But nothing popped up in the PR here and no notification mails hit my inbox. Please post them again.

tomsci:
How odd, they show up for me when viewing the PR... I may have failed my modern github 101, I will try and untangle tomorrow.

Cheers,

Tom

PR #2587, opened 2018-12-08
Update TLS protocol support

nwf:

I would like to adjust the default TLS settings of NodeMCU as per this commit.  While I believe these are all sane changes, I would appreciate some checking. :)


NicolSpies:
- I have been testing TLS operation against various ReST API servers using the http module. 
- The tweaking, as discussed in another thread, to have successful/valid handshaking is linked to right-sizing the `SSL_BUFFER_SIZE` setting the `MBEDTLS_SSL_MAX_CONTENT_LEN` to cater for the amount of data returned from the server and prevent the dreadful `-0x7200` `MBEDTLS_ERR_SSL_INVALID_RECORD` error message.
- The buffer size implication is that the specified size uses double the memory amount to cater for the transmit and receive buffers.
- The TLS Specification RFC5246 allows for the returned message size up to 16KB requiring an impractical 32KB of RAM from the ESP8266.
- Increasing the `SSL_BUFFER_SIZE` to `6000` resulted in successful secure Telegram API communication. Google web services require less while Amazon Web Services seems to be not possible even with LFS.
- There appears to be options other than dumping the ESP8266 by setting `MBEDTLS_SSL_MAX_CONTENT_LEN`  dynamically in combination with  `MBEDTLS_SSL_MAX_FRAGMENT_LENGTH`. Servers adhering to RFC6066 will then play nice with the ESP8266 and return managable TLS message sizes.
- If this can be done we will have a way to interface to secure cloud services currently out of reach of the ESP8266.
- Any thoughts ?

nwf:
My only thought on the matter of `SSL_BUFFER_SIZE` and friends is that the EC support should enable us to talk to more servers because we'll use smaller keys.  The bottleneck is, I am pretty sure, almost always the certificates in the handshake, and smaller keys should help.  I'm not particularly interested in otherwise trying to solve the `SSL_BUFFER_SIZE` problem in this PR, but perhaps the documentation should have a table of known-working sizes for popular endpoints.

That said, I would appreciate if you could report the key exchanges supported by the sites you care about, so I could quantify my assumptions that ECDSA support will help.

jmd13391:
@TerryE, Spot on. IMO, Using an ESP for this type of application is like trying to hammer a nail with a screwdriver.

NicolSpies:
@TerryE Terry, In principle I agree with your reply which is still in line and similar to your response about two years ago related to running a web server on the ESP8266. :smile: 

I agree with the balanced approach to focus on detail while retaining the perspective of the big picture. 
For me the quest is to attempt to optimize and maximize what the ESP8266 provides similar to what has been achieved with the LFS. I also accept that there are limits to what should be attempted with the ESP8266.

I believe it will be beneficial if the TLS module can be optimized by removing unused functionality and having smaller certificates enabling interaction with more servers. It might however be an overreach to use dynamically asymmetric SSL buffers, but I leave that for the experts. 

I like the "separation of powers" approach you implemented using a RPI gateway based system between the ESP8266s and the outside world.

nwf:
I have taken the opportunity of being neck-deep in SSL anyway to update us to mbedtls 2.7.7 as well.

NicolSpies:
@nwf with reference to : 
> I would appreciate if you could report the key exchanges supported by the sites you care about

I am currently testing against a commercial mock server using Apiary with OAuth 2.0 client certification using bearer tokens. Fails on SSL handshake due to buffer requirements (-0x7200). It Might be helpful for future testing if required.

nwf:
@NicolSpies please clarify if that is with or without this PR applied?

TerryE:
@NicolSpies, I absolutely agree that if we are going to have an option to include TLS stacks then they should be correctly tailored to match the resources available on the ESP, so we should remove configuration options that will fail with RAM available, etc.  If you have the bandwidth to take this on then you very much have my support and I suspect the full support of the other active contributors.  As Nathaniel suggests, what would be very useful if we had some some of brief discussion paper (e.g. in  a gist MD file) covering the tailoring options that you propose, and that you briefly mention in `tls.md`. 

Both the client and server configurations have a rich spectrum of options that compile in a _lot_ of code.  If we can strip this back to a usable subset then this would be a good step IMO.

I suspect that we have a couple of usecases that we would want to support as options.  Again defining these and understanding them would be a good thing.

NicolSpies:
@TerryE, I will gladly do it, we are moving house with the moving company coming on Saturday. So I am swamped by boxes and Polyfilla. I keep my sanity by visiting here in my breaks. :smile: 
We will be unpacking and settling in throughout the rest of the month but I will try to start working on it.

joysfera:
Would it be helpful if I mentioned this excerpt from Mongoose documentation (AFAIK Mongoose does allow successful TLS usage on ESP8266, even though I'm going to test it myself, for sure)?

> Mongoose OS provides memory-optimised mbedTLS library which reduces consumed RAM from ~32k down to ~2k per single TLS connection.

Also, this could perhaps be interesting - even though the above mentioned optimization might not be there (just guessing from the commit log):
https://github.com/mongoose-os-libs/mbedtls

nwf:
@joysfera Oh!  Those patches really should go upstream, but I will take a look.  Also apparently the higher-order ECs are too slow on 8266; maybe I will need to pull us up to mbedTLS 2.14 with resume-able EC support if we want those. :\

dtran123:
@joysfera thanks for the excerpt that explains how Mongoose OS does it to optimize RAM usage in TLS negotiations. I've always wondered how they did it. It would be nice if we could implement some of their optimization into ours.  As I said in another thread, Mongoose OS supports AWS IoT on the ESP8266 (I've tested it and it works flawlessly). Solving this problem will definitely increase the value of the ESP8266 Lua by leaps and bounds.

nwf:
It appears that MbedTLS upstream is not excited about buffer-reduction patches; see https://github.com/ARMmbed/mbedtls/pull/1208 .  I am particularly unexcited by the prospect of maintaining a divergent TLS implementation.

If we wish to pursue better TLS support, it may behoove us to consider moving from mbedTLS to either wolfSSL or BearTLS or another small TLS library.  This would require rewriting `espconn_mbedtls.c`, supplying the equivalent of `app/mbedtls/platform/esp_hardware.c`, and moving `app/modules/tls.c` over to the new library.

However, I am not volunteering to do the work, as I am not facing problems with TLS on NodeMCU beyond things tweaked in this PR.  Someone else can step up, or I can be contacted out of band about contractual arrangements.

marcelstoer:
@nwf thanks for digging out that mbedTLS link. I just finished reading (nearly) all the comments and I'm now really concerned to be dependent on a project that is (un)maintained like that. ACK for the issues raised by @lws-team.

> ...or BearTLS or another small TLS library

I guess you know that ESP8266 Arduino-core switched to BearSSL with the soon-GA release.

> I am not volunteering to do the work, as I am not facing problems with TLS on NodeMCU 

:+1: the issue is not severe enough for most users/developers

Looking forward to seeing this PR land.

joysfera:
I read all the comments. It sounds like mbedTLS is basically unmaintained, or at least not accepting any patches from outside for more than 2 years no matter what. Scary.

I also read a bit about BearSSL. This is interesting: "_No dynamic allocation whatsoever. There is not a single malloc() call in all the library_". This thing itself sounds like it doesn't actually help to decrease memory usage as it needs to have static buffers, doesn't it? Only if it had the buffers very small and somehow managed to work with them. At the same time it mentions **Maximum Fragment Length extension** and reduction of memory usage, so it apparently does some magic there. Interesting.

BTW, I am not sure that Marcel is 100% correct with "_the issue is not severe enough for most users_". Things have changed dramatically over the last year - all public servers moved to https exclusively, it's not possible to use almost any public service over http or other plain text protocol. SSL/TLS is required everywhere. @TerryE  invested so much time and energy into LFS to free a bit more RAM. If the mbedTLS will not allow communicating with others even with all the RAM the 8266 can offer then I am afraid NodeMCU usage will get very limited (basically for WLANs only). Yes, I know about Terry's suggestion of using a computer for proxying the requests to the public internet but that's not always a viable option. NodeMCU on ESP32 will be the only way of using Lua on internet then.

Just my 0.00 cents.

joysfera:
> @joysfera The "No dynamic allocation" thing means that the BearSSL API requires that the caller allocate buffers, not that all buffers are static.

Makes sense. I apparently know nothing about it, just read few words on the main page. The important thing is that BearSSL includes the RFC fragmentation support, I believe. 
Anyway, if Arduino core moved from mbedTLS to BearSSL it's a strong sign for others using 8266, I'd think.

marcelstoer:
@joysfera I didn't mean to suggest that we didn't need SSL/TLS support. We do and I'm really grateful Nathaniel & Co. continue to tweak and tune mbedTLS.

@nwf 
> that's probably a good sign that maybe we should as well, at some point.

Considering that some of the same Espressif folks who work on ESP-IDF also contribute to ESP8266 Arduino-core I was honestly quite surprised seeing them swapping TLS libraries. I wondered back then if this indicated in any way they might be doing the same for ESP-IDF or NON-OS SDK.

But sorry, we got off track again. Anyone wishing to discuss _that_ topic should open a new issue. Let's focus on your PR again.

lws-team:
To be clear I wouldn't say mbeltls is not well-maintained... as a project it makes a lot of forward progress and they know what they're doing.  It's just not maintained as a FOSS project, it's run as a proprietary project where the source happens to be open.  Unfortunately I also had experience with mbed itself in the early days and Arm was the same with the Not Invented Here, making me jump through a CLA and then as a policy ignoring my patches.  For whatever reason (fierce competition, demand from customers... contact with Enlightened Beings, I dunno) mbed itself reformed and last time I looked that had become something more normal.  mbedtls seem to be stuck with the policy to ignore nontrivial external contribution, and they appear completely comfortable with that.

At any rate the issue is dynamic sizing, from a quick look at beartls they have the approach to ban dynamic alloc in their library and as mentioned here to make buffers a caller problem.  That's a more reasonable way than mbedtls, since unlike the mbedtls build-time global define for buffer limit disgusting hack it allows you to choose your limit connection by connection, but it doesn't address the issue you must pass it fixed 16KiB +16KiB for a tls channel that may operate under all conditions (ie, to operate to tls spec).  For http/2, with multiple streams on one tcp + tls, and RFC8441 ws-over-h2 now, that is not a theoretical problem, since one stream on the single network connection may do bulk GET and simultaneously or later another decide to do bulk PUSH, each mandating a 16K buffer for rx + tx according to dynamic conditions.  If you unwisely used mbedtls' dumb allocation ceiling define hack, your build of the library can simply no longer do that, ALL connections are so limited as decided at build-time, the connections will drop if you try to support browser-based file PUSH firmware upgrade or anything else that doesn't fit the hack profile on any connection.  It's really a garbage hack that breaks tls and if replicated on beartls by simply providing smaller parent-allocated buffers, you will be left with the same problem.  Dynamic sizing (which brings with it the possibility of OOM during a connection) is the only solution where you can optimize the memory requirement while still being able to do anything allowed by the tls spec, memory permitting.

One way if beartls is anyway attractive to you is discuss their adding build-optional callbacks at points the buffer size is mandated to be something in their code, and perhaps a callback for when a buffer is empty, so you can also dynamically deallocate according to parent policy.  Beartls users that don't care don't enable it and continue to always provide 16KiB + 16KiB buffers.  If you are stuck on something like 8266 though you can implement the dynamic part then above beartls in the callback.

I also feel the same way as was mentioned above, I tracked mainline mbedtls with my patches for a while, and I got some help from others later, but strugging with that for the rest of my life isn't why I wrote the patches... I wrote them to go upstream and did the outreach work for that.  But mbedtls are simply unreasonable FOSS partners.  It doesn't generate any revenue to offset the time and blood pressure, so after the first few times... meh.

lws-team:
@joysfera just FYI, the fragmentation stuff is defined to only operate in one direction.  TLS simply doesn't have a way for both sides to logically agree a <16KiB buffer.

marcelstoer:
Regardless of the other potential knobs one could turn wrt TLS config, can we merge this as "definitely an improvement"?

NicolSpies:
Agreed, I will start testing against it regardless of being merged in a day or so.

NicolSpies:
@nwf, as the largest EC (512 bits) is now disabled, will it make sense to uncomment `//#define MBEDTLS_ECP_MAX_BITS  521 /**< Maximum bit size of groups */` in `user_mbedtls.h` and change to `#define MBEDTLS_ECP_MAX_BITS 384 /**< Maximum bit size of groups */` ?

galjonsfigur:
I tried to test this patch and found interesting thing: when `MBEDTLS_DEBUG_C` is defined in `user_mbedtls.h` (this is used in `tls.setDebug()` and generally useful when testing TLS connections) then build fails with:
```
mbedtls/.output/eagle/debug/lib/libmbedtls.a(debug.o): In function `debug_send_line':
nodemcu-firmware/app/mbedtls/library/debug.c:77: undefined reference to `mbedtls_vsnprintf'
...
mbedtls/.output/eagle/debug/lib/libmbedtls.a(debug.o): In function `mbedtls_debug_print_msg':
nodemcu-firmware/app/mbedtls/library/debug.c:92: undefined reference to `mbedtls_vsnprintf'
```

nwf:
@NicolSpies That seems perfectly reasonable and hopefully reduces RAM requirements further.  I'll try to test things as time permits.

@galjonsfigur Thanks, I guess I paved over my previous `vsnprintf` fix in this one.  I'll try to push a fix some time later today, assuming time permits.

galjonsfigur:
@nwf I'm testing the patch right now and restoring  `vsnprintf` macros in `app/include/mbedtls/platform.h`does the trick, but now only if `DEVELOP_VERSION` is defined `tls.setDebug` works (by that I mean getting output from the handshake process) and I'm not sure why. I think this is supposed to work regardless of  `DEVELOP_VERSION` macro and I'm still looking in code to find out what is the problem.

nwf:
I think it was always intended that `tls.setDebug` would only work with `DEVELOP_VERSION`, since it substantially enlarges the executable to have all those debug strings around.

marcelstoer:
>  `tls.setDebug` would only work with `DEVELOP_VERSION`

Yep, that flag is prerequisite for any debug output AFAIU (`NODE_DEBUG` actually).

joysfera:
Just a little note: these memory requirement numbers should be documented somewhere in bold as they are so unbelievably high. 30+ kB of free heap is never available in any decent Lua application (unless the LFS is used, I guess) so it basically means _do not waste your time trying to use TLS without LFS to connect anywhere - it will not work_.

nwf:
@joysfera I added a warning to the tls module docs; if that is insufficient, please propose a patch.

docs/en/modules/tls.md
```diff
@@ -8,11 +8,42 @@
...
+(RIPEMD-160, SHA1, SHA2), signature (RSA, deterministic ECDSA), and key
+exchange (EC)DHE algorithms as well as a reasonable set of elliptic curves
+(secp{256,384}r1, secp256k1, bp{256,384}) supported.
+
+!!! tip
+	If possible, you will likely be much better served by using the ECDSA
+	key exchange algorithms than by using RSA.  An increasingly large
+	fraction of the Internet understands ECDSA, and most server software
+	can speak it as well.  The much smaller key size (at equivalent
+	security!) is beneficial for NodeMCU's limited RAM.
```
marcelstoer: A brief note or link to how to achieve that would be helpful I guess.

marcelstoer:
> [on HTTP]... the module should just go away because the experience it offers is so rough around the edges

I agree, but
- 85% of all the users of my cloud builder use NodeMCU without TLS
- that module has always been the most favourite non-essential module

source: https://nodemcu-build.com/stats.php

NicolSpies:
@joysfera, @nwf; I have been trying to improve my understanding of IoT requirements and trends related to encryption from the endpoint and cloud server perspective. My thoughts so far:

- By 2020 the estimate is that there will be 50 billion IoT devices worldwide.
- The requirement to have secure connected IoT devices are gaining general acceptance.
- This is not new as per the Internet Engineering Task Force (IETF) RFC historic documents on constrained devices.
- TLS is here to stay and authenticated access to cloud APIs using OAuth 2 (dependent on TLS) is becoming the standard where authenticated access to third party cloud services are required.
- The HTTP module will make the OAuth 2 process easier but then even more memory will be required on top of the required TLS. The workaround is to do it inline using only TLS, but it is a bit more tricky.  
- Our Class 2 constrained device are better off compared to other more constrained devices where we at least have cases where TLS can be used utilizing LFS.
- With more constrained devices or the ESP8266 where TLS is not always feasible, CoAP using Datagram TLS (DTLS) might be an option.

With this all said, and no silver bullets, the we will most probably settle with a combination of a number of suggested, good practices like the suggested doing the TLS early in the app when resources are still plenty, removing all the ciphers except the smallest one the server will be happy with, trying to convince the server owner not to use the default allowed 16KB frame buffer size, hope SNI becomes available, hope dynamic buffer sizing and server renegotiation becomes possible, etc.

And last but not least, because there will not be a "one size fits all" recipe, clear documentation guidelines will most probably determine the adoption use. It might however be so involved that general Lua users will find it to challenging/tricky.

joysfera:
I am glad the ECDSA is mentioned in the updated documentation. I asked for ECDSA support about 10 months ago because I was hoping the smaller certificates would help with the limited memory we have on 8266. At that time the ECDSA was optional and not compiled in by default (cloud build service). 
Do I understand it correctly that starting with the next NodeMCU release the ECDSA will be available by default? Then we could start updating our servers to use ECDSA as well and hopefully make it easier to connect to them from NodeMCU devices.
BTW, I failed to update my own MQTT broker with ECDSA those 10 months ago for some reason (don't remember details anymore). I would try it again if it was really a saver for the NodeMCU.

nwf:
@NicolSpies The ESP8266, while a lovely machine, is, IMHO, not suitable as a generic IoT substrate.  It works in many constrained situations, but the lack of heap RAM -- such that the ~20KiB occupied ephemerally by mbedTLS connection bringup is a substantial fraction -- means that it is, IMHO, not suitable for generic applications making connections to generic endpoints on the open Internet.

DTLS will not save us; the costs are in the complicated session and certificate handling phase, which is common across TLS and DTLS.  Once the session is established, things are much lighter weight, but we will always require at least the RX buffer for asynchronous packets.

Server renegotiation is, for similar reasons, unlikely to be a good idea, even if it'd be trivial to not disable it in this patch series.

mbedTLS will probably never support dynamic buffers; they are not acting as though they care about such things.

SNI would require changes to the Espressif SDK layer upon which NodeMCU is built, and Espressif has indicated that the only changes they wish to make are to abandon the particular version we sit on in favor of a much more memory intensive RTOS version.

It is possible that changing TLS providers would address some of these issues, but would also be a huge amount of work, requiring rewriting the Espressif-provided secure connection layer, with relatively unknown payoff.  Interested parties are welcome to contact me out of band for contracting rates.

@joysfera The little testing I did suggests that switching to ECDSA keys shaves off about 5KiB of heap during session bringup.  While that's substantial, it is probably not enough to actually matter, because it's still going to be hard to keep the 15KiB bursted by TLS bringup free during an application's run.  I think I stand by my suggestion that TLS bringup happen as early as possible and that connection failures transmute into device reboots even with ECC keys.

jmd13391:
What about creating an open/free secure connection proxy service in the cloud that all ESP devices can use?

nwf:
@jmd13391 Be my guest.  That, I don't think, factors in to this PR, tho'. :)

jmd13391:
@nwf 

> Be my guest. That, I don't think, factors in to this PR, tho'. :)

Sometimes my "big picture" Architect side surfaces at the strangest times. That being said, I respectfully disagree.. At the highest level, I see this topic (and several other topics here) as an attempt to provide secure communications with the ESP device... i.e. make it safe and secure to use the ESP in an IoT solution. TLS, HTTPS, etc. support is a means to that end. As @TerryE so elegantly put it years ago (paraphrasing) "If you are trying to make the ESP8266 a web server, you're doing it wrong".

It's awesome to see firmware developers here squeeze so much blood out of the ESP device rock. However, IMO the ESP platform is just too resource constrained to keep up with the security world. It seems to me that there are one or two minimally secure "good enough" communication protocols that the ESP device can support and still have enough RAM left over to actually run a user application. So why not use those to communicate with a cloud proxy host and have the cloud host interface with the Internet using any/all secure protocols currently available. i.e. a compromise. These were thoughts when I suggested a cloud proxy solution to the higher level problem.

As for the "be my guest", I'd love to... Unfortunately, I'm too busy at the moment trying to produce actual products using NodeMCU Lua and the ESP platform. I merely offered the suggestion to others in the hopes it might be helpful and spawn additional thought on the topic.

nwf:
@jmd13391 Going via a third party is often antithetical to security.  For better or worse, TLS is what we have, and it's probably how we'd reach said third party's proxy anyway!  Moreover, running such a service requires money and proposing it to volunteers has a very "bear my externalities' costs for me" vibe to it if you're unwilling to offer it yourself.

I am getting sick of this PR (and, frankly, the demands that Someone Else should Do Something and ideally do that thing For Free to help others using the NodeMCU project to make money, but I digress) and so request that it either be merged or that some constructive feedback (thank you @marcelstoer) as to why it hasn't yet be given.

jmd13391:
@nwf I feel your pain. For the record, I dont use TLS or HTTPS or any connection to the Internet whatsoever in my NodeMCU Lua / ESP device products. So, I have no conflict of interest here... nor am I looking for a handout... Good luck with this effort. ;-)

TerryE:
> The much smaller key size (at equivalent security!) is beneficial for NodeMCU's limited RAM.

> I was hoping the smaller certificates would help with the limited memory we have on 8266.

(I've still got the McKean & Moll book on ECs on my bookshelf; I got interested in them because of Wley's proof of the Fermat conjecture.)  Us of Modular ECs does reduce the key size for a given cryptographic strength, and this in turn reduces the size of the large precision integers manipulated so as @nwf says this does yield some small improvements, but the underlying issue is that establishing a trusted end-to-end channel is complicated, and so I fully support Nathaniel's summary the ESP8266 is not suitable as a generic IoT substrate for _direct and secure interaction over the cloud_.  IMO, mbedTLS implements the session communications poorly in terms of memory footprint, but it is establishing the sessions in the first place that is the real killer.  So IMO, ESP32 yes, ESP8266 no.

I run NodeRED on an RPi-class SBC and use this as my hub for all cloud based connection.  I use a simple signed JSON protocol between my ESPs and NodeRED with my own simple custom node to simplify the ESPs integration in to the flows.  This give me update security but not privacy for minimal runtime RAM impact at the IoT device.

nwf:
@marcelstoer I believe we are good to go.

nwf:
@galjonsfigur You're probably right.  Sorry for dropping the ball on that delta.  It's now part of this PR.

nwf:
@marcelstoer Sorry about the thinko caught by the CI.  *Now* I think we're good to go.

NicolSpies:
@FrankX0 Hi Frank, I found you had the same "Minor version mismatch" problem last year in March 2018 as detailed on the armMBED forum. 
Your feedback how you solved it by sending the server name is interesting. Is that related to Server Name Identification from the tls module ? I would really appreciate info on how you solved it.

nwf:
@NicolSpies Please take your troubleshooting off this PR.

PR #2592, opened 2018-12-14
Recreate and unify documentation for Lua modules

### LINKED ISSUE #2588: Unify documentation for Lua modules in lua_modules/ folder
Documentation for NodeMCU and it's modules is written very well and it's really helpful. However, for lua_modules there are only `README.md` files and sometimes examples for modules in separate `.lua` files. 

The idea is to unify this documentation in docs subfolder (like `/docs/en/lua_modules/`) and rewrite .md files to work with ReadTheDocs + restructure them to look like documentation for C modules. This will be more friendly approach and will encourage to use these modules, especially after LFS patch. 

I would happily do this and make PR, but first I want to know if this would be desired approach.

galjonsfigur:
Fixes #2588.

The complete rewrite of documentation for Lua Modules as mentioned in #2588. For some modules like `ftpserver` or `ds18b20` documentation rewriting was pretty straightforward. For other modules like `imap` some of the original documentation had to be rewritten because it was using Doxygen style of documentation and it couldn't fit into current style documentation. But there was modules like `redis` or `yeelink` without documentation, so I created it basing on code. I couldn't really test most of the examples attached to original documentation so the modifications in snippets are minimal. 

Also Lua module `http` will have to be renamed because of name clash with C module but before I could submit that change (maybe new PR?) I want to be aware of any issues with this commit.


galjonsfigur:
Thanks for pointing this out. Also found and fixed some typos. There might be more, but this is all I could find.

TerryE:
@galjonsfigur :+1: :smile:

PR #2605, opened 2019-01-09
fix number2integer conversion in ESP32

### LINKED ISSUE #2604: bitwise `bit` module producing incorrect results for large integers
### Expected behavior
`bit.bnot(0x80000000) == -2147483648` (`0x7FFFFFFF`)
### Actual behavior
`bit.bnot(0x80000000) == 0x80000000`
### Test code
Run this on a ESP8266 `master` branch
```lua
print(string.format("%x", bit.bnot(0x80000000)))
```
You'll get the correct result: `7fffffff`.
Now run the same in an ESP32 using the `dev-esp32` branch.
You'll get this: `80000000`, which is incorrect.

### NodeMCU version
`dev-esp32`

### Hardware
ESP32, ESP-WROOM-32

jpeletier:
Fixes #2604 for ESP32.

ESP8266 PR: #2609.

This PR fixes a quirk in floating point builds whereupon conversion of a large integer (>= 0x80000000) contained in a lua value is not properly done, resulting in the most significant bit set to zero. This manifests itself visibly when using the `bit` module. Notably, `bit.lshift(0xFFFFFFFF,0)` (shifting **zero** bits) does not result in the same number.

I tracked this down to the `lua_number2integer` macro, so now it casts it to a `long long` (at least 64 bits) before truncating it. This fixes the issue.

### DIFF
--- a/components/lua/luaconf.h
+++ b/components/lua/luaconf.h
@@ -702,8 +702,18 @@ union luai_Cast { double l_d; long l_l; };
 
 /* this option always works, but may be slow */
 #else
-#define lua_number2int(i,d)	((i)=(int)(d))
-#define lua_number2integer(i,d)	((i)=(lua_Integer)(d))
+
+#ifdef LUA_NUMBER_INTEGRAL
+
+#define lua_number2int(i, d) ((i) = (int)(d))
+#define lua_number2integer(i, d) ((i) = (lua_Integer)(d))
+
+#else // for floating-point builds, cast to a larger integer first to avoid undefined behavior on overflows.
+
+#define lua_number2int(i, d) ((i) = (int)(long long)(d))
+#define lua_number2integer(i, d) ((i) = (lua_Integer)(long long)(d))
+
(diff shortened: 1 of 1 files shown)


jpeletier:
@devsaurus Espruino had the same issue and they used `long long`. I guess the compiler wouldn't care. 

My reasoning is that `lua_Integer` is a signed type, so wouldn't make sense to cast double (signed) - > unsigned -> signed

devsaurus:
Ok, I mixed the signedness - that should have been `int64_t` instead in my previous post.
My point is: If you require a 64&nbsp;bit representation then why not explicitly state that in the cast? Relying on `long long`&nbsp;==&nbsp;64&nbsp;bit is a platform specific implementation.

jpeletier:
I just need it to be bigger than 32 bits. `long long` is guaranteed to be at least 64 bits. I can change it if you want, but I see `long long` all over the place in the code.

devsaurus:
This one should follow once #2609 is merged.

PR #2607, opened 2019-01-13
ESP32 crypto module

jpeletier:
- [x] This PR is for the `esp32-dev` branch rather than for `master`.

This PR is a minimal implementation for ESP32 of a `crypto` module using mbedtls. For now, it contains the SHA1, SHA256, SHA224, SHA512, SHA384 and MD5 hashing functions, as well as the `toHex()` function to convert binary strings to hexadecimal.

If hardware acceleration is enabled for mbedtls, this module will make use of it as well.

The interface of this module aims to be backward-compatible to the one in the `master` branch (ESP8266). Not all functions have been implemented, but those who are, are backward compatible. The updated documentation shows only the functions that have been implemented.

To enable/disable this module, use `make menuconfig` and select/deselect it in the NodeMCU modules section. You can also disable hashing algorithms there in order to reduce the binary image size

### DIFF
--- a/components/modules/Kconfig
+++ b/components/modules/Kconfig
@@ -74,6 +74,12 @@ config LUA_MODULE_CAN
   help
       Includes the can module.
 
+config LUA_MODULE_CRYPTO
+  bool "Crypto module"
+  default "n"
+  help
+      Includes the crypto module.
+
 config LUA_MODULE_DAC
   bool "DAC module"
   default "n"
--- a/components/modules/Kconfig
+++ b/components/modules/Kconfig
@@ -80,6 +80,30 @@ config LUA_MODULE_CRYPTO
   help
       Includes the crypto module.
 
+menu "Crypto module hashing algorithms"
+  config CRYPTO_HASH_SHA1
+    bool "SHA1"
+    default "y"
+    help
+      Includes the SHA1 hashing algorithm
+  config CRYPTO_HASH_SHA256
+    bool "SHA256 and SHA224"
+    default "y"
+    help
+      Includes the SHA256 and SHA224 hashing algorithms
+  config CRYPTO_HASH_SHA512
+    bool "SHA512 and SHA384"
+    default "y"
+    help
+      Includes the SHA256 and SHA384 hashing algorithms
+  config CRYPTO_HASH_MD5
+    bool "MD5"
+    default "y"
+    help
+      Includes the MD5 hashing algorithm
+endmenu
+
+
 config LUA_MODULE_DAC
   bool "DAC module"
   default "n"
--- a/components/modules/Kconfig
+++ b/components/modules/Kconfig
@@ -81,6 +81,8 @@ config LUA_MODULE_CRYPTO
       Includes the crypto module.
 
 menu "Crypto module hashing algorithms"
+  depends on LUA_MODULE_CRYPTO
+
   config CRYPTO_HASH_SHA1
     bool "SHA1"
     default "y"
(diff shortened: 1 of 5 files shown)


components/modules/crypto.c
```diff
@@ -0,0 +1,272 @@
...
+    size_t len;
+    const char* msg = luaL_checklstring(L, 1, &len);
+    char* out = (char*)luaM_malloc(L, len * 2);
+    int i, j = 0;
+    for (i = 0; i < len; i++) {
+        out[j++] = crypto_hexbytes[msg[i] >> 4];
+        out[j++] = crypto_hexbytes[msg[i] & 0x0F];
+    }
+    lua_pushlstring(L, out, len * 2);
+    luaM_free(L, out);
```
devsaurus: Please change to `luaM_freemem()`.

components/modules/crypto.c
```diff
@@ -0,0 +1,272 @@
...
+
+    // if the mbedtls context is NULL, it means allocation failed in new_hash(), so nothing to do.
+    if (phctx->mbedtls_context == NULL)
+        return 0;
+
+    // free mbedtls-related resources for this hash operation:
+    phctx->ainfo->free(phctx->mbedtls_context);
+
+    // free the memory allocated to store the mbedtls context:
+    luaM_free(L, phctx->mbedtls_context);
```
devsaurus: Please change to `luaM_freemem()`.

components/modules/Kconfig
```diff
@@ -74,6 +74,36 @@ config LUA_MODULE_CAN
   help
       Includes the can module.
 
+config LUA_MODULE_CRYPTO
+  bool "Crypto module"
+  default "n"
+  help
+      Includes the crypto module.
+
+menu "Crypto module hashing algorithms"
```
devsaurus: Cosmetic: Hide this menu when the crypto module is not selected:
```suggestion
menu "Crypto module hashing algorithms"
  depends on LUA_MODULE_CRYPTO

```

components/modules/crypto.c
jpeletier: > `mkdocs.yml` is missing a reference to `crypto.md`.

ah, so that is where it was!. Fixed.

components/modules/crypto.c
jpeletier: fixed.

jpeletier:
@devsaurus All comments resolved.

I don't get why GitHub says there is a conflict in `mkdocs.yml` when only one line is added and this branch is not behind `dev-esp32`. Do you know?

marcelstoer:
> this branch is not behind `dev-esp32`

Sorry to say but in fact it is. I committed https://github.com/nodemcu/nodemcu-firmware/commit/5750da121c5bdf03b7cff12de8c0574226db6474#diff-45ee620855dad847bea76b49932ad7db a little over an hour ago. You need to rebase.

jpeletier:
hah! :-) @marcelstoer fixed!

devsaurus:
Looks good, thanks a lot!

devsaurus:
@jmattsson Johny, any objections or additions?

jmattsson:
@devsaurus Overall this looks great!

I do think the `toHex()` is superfluous here though - it's already available in the `encoder` module. While originally we had `crypto.toHex()` on the ESP8266, that was superseded and deprecated when the `encoder` module was introduced by Terry(?). Re-introducing it here would feel like an unnecessary design regression.

marcelstoer:
Good point, your memory serves you right 👍

jpeletier:
@jmattsson Yes, I agree. I stumbled upon the `toHex()` in the encoder module the other day, I just didn't see it before this PR and happily ported the one from crypto@ESP8266

I'll remove this function and we should be good to go. :-)

jpeletier:
@jmattsson @devsaurus @marcelstoer I removed `toHex()`, so this can be merged if there are no further comments. Thanks!

jmattsson:
Thanks for the PR :)

Issue #2377, opened 2018-05-20
The LGC usage count is incorrectly calculated if luaM_free() is used to free arrays in our modules


b-a-9-0:
### Expected behavior
Read line, no permanent memory consumption.

### Actual behavior
Readline works, but causes permanent memory consumption.

Console output directly after startup:
>=collectgarbage("count")
4

Function is run via console:
> increment_job_counter()

Memory usage increases:
> =collectgarbage("count")
5

> increment_job_counter()
> =collectgarbage("count")
6

> increment_job_counter()
> =collectgarbage("count")
7

> collectgarbage()
> =collectgarbage("count")
7

This example can be continued until memory overflow error occurrs.

### Test code
This is my minimalistic init.lua
```Lua
function increment_job_counter()
	if file.exists("job.counter") == true then
		local handle = file.open("job.counter", "r")
		local val = tonumber(file.readline()) + 1
		handle = file.open("job.counter", "w+")
		handle:writeline(tostring(val))
		handle:close()
		handle = nil
		val = nil
	else
		local handle = file.open("job.counter", "w+")
		handle:writeline(tostring(1))
		handle:close()
		handle = nil
	end
end
```
### NodeMCU version
branch: master (latest)
commit: 67027c0d05f7e8d1b97104e05a3715f6ebc8d07f

### Hardware
Wemos D1 mini pro

TerryE:
OK, I had a binary chop on this and the simplest code which demonstrates the leak is
```Lua
do 
     file.open("job.counter", "r")
     for i = 1,10,1 do
          file.readline()
--          file.seek "set"
          print("used memory: ",collectgarbage("count")*1024)
     end
     file.close()
end
```
It doesn't matter whether you reseek to pos 0 or not; it's the readline that leaks.  Replacing this with `file.read(1)` works fine.  I'll take a look at the C code.

BTW your `tonumber(file.readline())` should be `tonumber(handle:readline())`

TerryE:
Another update.  I added `node.heap()` to the "memory used" diagnostic and ran this for 100 iterations.  The last few are:
```
used memory:    105421  Heap:   42552
used memory:    106444  Heap:   42552
used memory:    107467  Heap:   42552
````
We only have ~48Kb heap on the ESPs and `node.heap()` is reporting a solid 42552; it's the LGC reported value that is crap.

Now the file read routine does a:
[...]

PR #2610, opened 2019-01-14
ESP32: fix memory leak in encoder.fromHex and avoid use luaM_free

jpeletier:
Works toward fixing #2377. Also fixes a memory leak.

While reviewing the encoder module I stumbled upon 2 occasions in `encoder.fromHex()` where code would exit without freeing allocated memory if an invalid hex string was passed (odd length or with non-hex chars). Also it would allocate 4 times more memory than required. I fixed this and also modified the code so that it does not use `luaM_free()`.

If this is approved, it should be backported to ESP8266 in `dev`.


jpeletier:
Cleaned up implementation and pushed changes.

PR #2615, opened 2019-01-15
fix memory leak in http one-shot requests

### LINKED ISSUE #2614: ESP32: huge memory leak in http one-shot requests
### Expected behavior
All memory is released when using `http.get(url, callback)`
### Actual behavior
Around 2k of memory is retained every time you call `http.get(url, callback)`
### Test code
```Lua
collectgarbage()
a=node.heap()
count=0
function finish()
    collectgarbage()
    b=node.heap()
    print(string.format("\n\n\n\n\n\ndiff %d-%d=%d\n\n\n", a,b,a-b))
    a=b
    count=count+1
    if count < 200 then
       tmr.create():alarm(50, tmr.ALARM_SINGLE, herewegoagain)
    end 
end


function herewegoagain()
    http.get("SOME_URL", function(code, body, headers)
        tmr.create():alarm(100, tmr.ALARM_SINGLE, finish)
    end)
end

herewegoagain()
```
Executing the above will quickly exhaust your heap. At some point you'll get an error saying "cannot create rtos task", or other memory related issues.

### NodeMCU version
`dev-esp32`, `1adbc791062563aa99a6a010dfff384b6299b0d9`

### Hardware
ESP-32

jpeletier:
Fixes #2614

When working in "one shot" mode, the http module does not free resources appropriately and loses around 2k ram on each call (!!). The referenced issue has some code you can run to see your precious heap evaporate very quickly.

After much pain, I traced this to the `http_accumulate_complete()` function. This function is a closure and had the request context as an upvalue, thus creating a circular reference that prevented the context userdatum to be garbage collected, along with all esp_http-related resources.

This fix explicitly clears this closure after it is finished executing, thus triggering gc.

### DIFF
--- a/components/modules/http.c
+++ b/components/modules/http.c
@@ -648,10 +648,9 @@ static int http_accumulate_complete(lua_State *L)
     lua_rawseti(L, cache_table, i);
   }
   luaL_pushresult(&b); // data now pushed
+  lhttp_context_t *context = (lhttp_context_t *)luaL_checkudata(L, context_idx, http_context_mt);
   if (lua_isnoneornil(L, 1)) {
     // No callback fn so must be sync, meaning just need to stash headers and data in the context
-    lhttp_context_t *context = (lhttp_context_t *)luaL_checkudata(L, context_idx, http_context_mt);
-    
     // steal some completion refs, nothing's going to need them again in a one-shot
     context_setref(L, context, DataCallback); // pops data
     lua_rawgeti(L, cache_table, 2); // headers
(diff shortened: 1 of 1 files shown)


TerryE:
I haven't had the opportunity to test this but this issue and fix is as Javier describes. :)

jpeletier:
Pinging @tomsci (`http.c` author) to see if he can take a look to the fix as well.

tomsci:
Sorry about that @jpeletier, I'm embarrassed to have committed that! This PR certainly fixes the cyclic reference in the normal case, by breaking the cycle in the complete fn, so from that point of view it's all good. I think there are some still some potential (rare) corner cases where the context could still get leaked (eg if http_lapi_request errored and thus the completion fn never got called) so I think I'd like to rework that fn to not require that cyclic reference in the place. But I'm happy to raise a separate issue for that and not hold up this fix.

PR #2623, opened 2019-01-22
Compilation fix

kevinresol:
This fixes an obvious typo which happens in debug builds only. No tests have been conducted, sorry.


devsaurus:
Thanks a lot!

PR #2628, opened 2019-01-24
Mqtts

### LINKED ISSUE #2626: [esp32] MQTT mutual auth
Trying to get MQTT mutual authentication (client cert) work on esp32.

[Here is my attempt](https://github.com/kevinresol/nodemcu-firmware/commit/bea2c4813306d3c4ebe5a025df66a981120a12fb), trying to pass the cert/key PEM's to esp-mqtt.

And the lua call is:

```lua
mqttClient:connect(host, port, 1, 0, cacert, client_cert, private_key, on_ok, on_nok);
```

However, I am getting this error when connecting:

```
I (11446) MQTT_CLIENT: Sending MQTT CONNECT message, type: 1, id: 0000
E (11546) MQTT_CLIENT: Invalid MSG_TYPE response: 9, read_len: 0
I (11546) MQTT_CLIENT: Error MQTT Connected
I (11546) MQTT_CLIENT: Reconnect after 10000 ms
E (11556) MQTT_CLIENT: Client has not connected
```

`read_len` is zero meaning the response (9 in this case) is basically rubbish data in the buffer, we could ignore that.

[`select()`](https://github.com/espressif/esp-idf/blob/e2ae69f/components/tcp_transport/transport_ssl.c#L89) tells us the fd is ready to be read but [actually reading](https://github.com/espressif/esp-idf/blob/e2ae69f/components/tcp_transport/transport_ssl.c#L129) it yields `0` length. I guess that means the socket is closed. But I am not sure why.

I am not a C expert, I would like to have some guide on this problem. Thanks.

kevinresol:
Fixes #2626.

Adds some options for MQTTS (secure MQTT).
- CA Cert (to verify server)
- Client Cert and Key (to let server verify client)

### DIFF
--- a/components/modules/mqtt.c
+++ b/components/modules/mqtt.c
@@ -446,6 +446,9 @@ static int mqtt_connect( lua_State* L )
 	const char * host = luaL_checkstring( L, 2 );
 	int port = 1883;
 	int n = 3;
+	const char * cert_pem = NULL;
+	const char * client_cert_pem = NULL;
+	const char * client_key_pem = NULL;
 
 	if( lua_isnumber( L, n ) )
 	{
@@ -462,6 +465,24 @@ static int mqtt_connect( lua_State* L )
 	if( lua_isnumber( L, n ) )
 	{
 		reconnect = !!luaL_checkinteger( L, n );
+		n++;
+	}
+
+	if( lua_isstring( L, n ) )
+	{
+		cert_pem = luaL_checkstring( L, n );
+		n++;
+	}
+
+	if( lua_isstring( L, n ) )
+	{
+		client_cert_pem = luaL_checkstring( L, n );
+		n++;
+	}
+	
+	if( lua_isstring( L, n ) )
+	{
+		client_key_pem  = luaL_checkstring( L, n );
 		n++;
 	}
 
@@ -482,6 +503,9 @@ static int mqtt_connect( lua_State* L )
 	lua_pop( L, n - 2 ); //pop parameters
 
 	strncpy(mqtt_cfg->host, host, MQTT_MAX_HOST_LEN );
+	if(cert_pem) mqtt_cfg->config.cert_pem = cert_pem;
+	if(client_cert_pem) mqtt_cfg->config.client_cert_pem = client_cert_pem;
+	if(client_key_pem) mqtt_cfg->config.client_key_pem = client_key_pem;
 	mqtt_cfg->config.port = port;
 
 	mqtt_cfg->config.disable_auto_reconnect = (reconnect == 0);
--- a/docs/modules/mqtt.md
+++ b/docs/modules/mqtt.md
@@ -92,13 +92,16 @@ none
 Connects to the broker specified by the given host, port, and secure options.
 
 #### Syntax
-`mqtt:connect(host[, port[, secure[, autoreconnect]]][, function(client)[, function(client, reason)]])`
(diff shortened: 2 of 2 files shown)


kevinresol:
tbh I am don't like this API very much. There are too many arguments and I think using a table would be more appropriate.

devsaurus:
Thanks a lot for your submission @kevinresol!

> There are too many arguments and I think using a table would be more appropriate.

I share this impression, but I don't have a proposal at the moment.
* Just the options related to the certificate in a table?
    `mqtt:connect(host[, port[, secure[, autoreconnect]]][, cert_table]][, function(client)[, function(client, reason)]])`
* Move everything into a table?
    `mqtt:connect(parameter_table)`
* Keep `mqtt:connect()` as it is and use a dedicated function to set up this before calling `connect()`?
    `mqtt:config(config_table)`

Any other ideas?

jpeletier:
Hello @kevinresol & @devsaurus:

I just submitted a rewrite of the MQTT module... I was having a hard time using it. This means we got now a (hopefully) much more stable implementation and I know the details very well. The new source is heavily documented, so anyone can get up to speed in no time.

Regarding the API, I also think there are too many config parameters. Also, some parameters are configured in `mqtt.Client()` and some when you call `client:connect()`. I'd rather have only `client:connect(parameter_table)` (@devsaurus's 3rd option above); this is what maps the best to the underlying `esp_mqtt` library and saves memory (we don't have to remember config parameters submitted to `mqtt.Client()` until we actually connect).

I can refactor this very quickly to whatever we agree on. My only requirement is that any API change is backported to ESP8266 as well. Being able to share lua code among these two platforms is what makes NodeMCU a great choice for IoT projects.

devsaurus:
> My only requirement is that any API change is backported to ESP8266 as well

An important aspect, especially when considering backward compatibility with existing Lua code.
With this in mind, the `client:connect(parameter_table)` would be an alternative to the existing variant with positional parameters. Both variants would need to be maintained in parallel, I'm afraid.

> Any other ideas?

Can we simplify bus also extend the existing API?
- `autoreconnect` is on the [path to deprecation](https://github.com/nodemcu/nodemcu-firmware/blob/11592951b90707cdcb6d751876170bf4da82850d/app/modules/mqtt.c#L1333)
- overload the current `secure` parameter with a table for certificate handling
    0 or nil: plain connection
    1: secure connection, no certs
    table: secure connection, certificates according table entries

jpeletier:
@devsaurus I like the overload `secure` parameter idea to simplify things.
@kevinresol what do you think?. Would you take a stab at this? This change would need two PRs: ESP32 and ESP8266 so we maintain the implementations in sync.

kevinresol:
@jpeletier thanks for your efforts! I could help testing the stuff. But I think I just don't have enough lua/c knowledge to implement anything more complex than this PR. Especially the memory management part.

I agree overloading the `secure` parameter is a good start if changing the API completely to a table-based one is too much work (involve backporting)

devsaurus:
Kevin, I'd like to get this enhancement into the code base.
I could set up a new PR based on your code and the changes to the API according to above conclusion. I can't test it without an mqtt setup including certificates, so it would be helpful if you could test the PR before merge.

Are you ok with such an approach?

kevinresol:
Yes I can help testing it.

Btw, I am using AWS IoT. It has some free quotas for a newly registered account for one year. If you wish to know how to set up one. Let me know.

devsaurus:
Ok, great. I'll see how far I get.

devsaurus:
Code was re-used in #2657 and merged to `dev-esp32`.
Thanks again, @kevinresol!

PR #2631, opened 2019-01-28
ESP32: remove use of luaM_free in file module

jpeletier:
Works toward fixing #2377 for ESP32

modifications in `file_g_read()` to avoid use of `luaM_free()`

### DIFF
--- a/components/modules/file.c
+++ b/components/modules/file.c
@@ -2,6 +2,7 @@
 
 #include "module.h"
 #include "lauxlib.h"
+#include "lnodeaux.h"
 #include "lmem.h"
 #include "platform.h"
 
@@ -140,7 +141,6 @@ static int file_obj_free( lua_State *L )
   if (ud->fd) {
     // close file if it's still open
     vfs_close(ud->fd);
-    ud->fd = 0;
   }
 
   return 0;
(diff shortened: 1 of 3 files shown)


components/modules/file.c
```diff
@@ -383,32 +383,27 @@ static int file_stat( lua_State* L )
 // g_read()
 static int file_g_read( lua_State* L, int n, int16_t end_char, int fd )
 {
-  static char *heap_mem = NULL;
-  // free leftover memory
-  if (heap_mem) {
```
devsaurus: This code was added to free memory which was left allocated by a previous call. Standard execution flow doesn't need that, but `heap_mem` isn't freed in case `lua_pushlstring()` throws a memory exception as noted in https://github.com/nodemcu/nodemcu-firmware/pull/1541#issuecomment-254654541.

If we keep considering this corner case then `bufsize` would also need to be declared `static`.

jpeletier: @devsaurus I added a new function to make a protected call to `lua_pushlstring()` and avoid this issue. The new `luaX_pushlstring()` returns nonzero instead of throwing an exception, allowing you to free memory. This function could be useful elsewhere.

Please check it out and let me know :-)

devsaurus: Excellent solution!

PR #2638, opened 2019-01-31
ESP32/mqtt: allow daisy-chaining subscriptions

### LINKED ISSUE #2635: ESP32: Daisy chaining mqtt subscription callbacks not possible
### Expected behavior
The Lua code can daisy chain `client:subscribe()` calls, so that each successful subscription will trigger the next one.

### Actual behavior
The first callback for `client:subscribe()` gets called back, but the callback for the second subscription is never triggered. The second topic is subscribed nonetheless - messages are received,

### Test code
```Lua
m = mqtt.Client("esp32", 120)
m:connect(ip, port, 0, function(client)
  print("connected, schedule topic subscriptions")
  client:subscribe("topic1", 0,
    function (client)
      client:subscribe("topic2", 0, function () print("subscriptions done") end)
    end)
end,
function(client, reason)
  print("failed reason: " .. reason)
end)

```
### NodeMCU version
Current `dev-esp32` branch.

### Hardware
Any esp32 module.

### Background
I stumbled upon this while investigating hints for the `mqtt` docs how the user should handle multiple subscriptions in a row. There's dedicated support on esp8266 where `subscribe()` and `unsubscribe()` accept tables, while the esp32 module doesn't offer this (and probably never will due to restrictions in the esp-mqtt API).

@jpeletier I suspect that [unreferencing `subscribed_ok_cb`](https://github.com/nodemcu/nodemcu-firmware/blob/56f19e44c35f019b6b56d04ec5b9a764e6e0413c/components/modules/mqtt.c#L187) *after* executing the user callback function immediately invalidates a new callback that was set during the `pcall()`. `luaX_unset_ref()` for the current cb should be done *before* executing the user callback.
This also affects `unsubscribe()` and `publish()`.

jpeletier:
Fixes #2635 

### DIFF
--- a/components/modules/mqtt.c
+++ b/components/modules/mqtt.c
@@ -180,11 +180,10 @@ static void event_subscribed(lua_State* L, mqtt_context_t* mqtt_context, esp_mqt
     ESP_LOGD(TAG, "CB:subscribe: calling registered one-shot subscribe callback");
     lua_rawgeti(L, LUA_REGISTRYINDEX, mqtt_context->subscribed_ok_cb);  // push the function reference on the stack
     luaX_push_weak_ref(L, mqtt_context->self);                          // push the client object on the stack
+    luaX_unset_ref(L, &mqtt_context->subscribed_ok_cb);                 // forget the callback since it is one-shot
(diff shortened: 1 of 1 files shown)


devsaurus:
Looks good, thanks for the quick response!

PR #2641, opened 2019-02-01
ESP8266: add flash baudrate environment variable to Makefile and preset for ESP8285

jpeletier:

This PR adds a new environment variable `BAUDRATE` for `make` to be able to specify the upload baud rate:
```bash
$ export BAUDRATE=460800
$ make flash4m
``` 
The default is `115200`

It also adds a preset to flash ESP8265 devices, such as the sonoff:

```bash
$ make flash1m-dout
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,9 @@ ifeq ($(OS),Windows_NT)
 	else
 		ESPPORT = $(COMPORT)
 	endif
+	ifndef BAUDRATE
+		BAUDRATE=115200
+	endif
     ifeq ($(PROCESSOR_ARCHITECTURE),AMD64)
 # ->AMD64
     endif
@@ -299,8 +302,9 @@ clobber: $(SPECIAL_CLOBBER)
 
 flash:
 	@echo "use one of the following targets to flash the firmware"
-	@echo "  make flash512k - for ESP with 512kB flash size"
-	@echo "  make flash4m   - for ESP with   4MB flash size"
+	@echo "  make flash512k     - for ESP with 512kB flash size"
+	@echo "  make flash1m-dout  - for ESP with   1MB flash size and flash mode = dout (Sonoff, ESP8285)"
+	@echo "  make flash4m       - for ESP with   4MB flash size"
 
 flash512k:
 	$(MAKE) -e FLASHOPTIONS="-fm qio -fs  4m -ff 40m" flashinternal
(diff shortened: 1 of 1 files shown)


marcelstoer:
The only _potential_ reason for not merging, and it's not even a good one, is that we decided a long time ago to eventually drop the flash target from `make`.

jpeletier:
> we decided a long time ago to eventually drop the flash target from make

Really? I find it very convenient and use it every day.

galjonsfigur:
I was using simillar change in makefile for a long time, but with new `esptool` this wouldn't be needed anymore - current versions can detect flash size without problems so only `make flash` target with removed flash size parameters would be sufficient for all ESP chips.

`make flash` is very useful for developing firmware and I think it should be kept, especially when SDK 3.0 update by @TerryE will bring upstream version of `esptool`.

marcelstoer:
Not having to maintain our own - always outdated - copy of `esptool.py` was exactly the reason why we want**ed** to drop `make flash`.

TerryE:
I think that we can do everything with the Espressif esptool. I'll incorporate this into the SDK 3.0 patch if I can.

TerryE:
> Really? I find it very convenient and use it every day.

I find it a lot easier to use multiple binaries on the esptool command line, but I accept that many will find this target convenient.

devsaurus:
Also ok for me to merge this.
With the outlook that we will drop our local modifications and sync with upstream esptool.py, a generic `make flash` should be finally feasible.

Makefile
```diff
@@ -299,20 +302,25 @@ clobber: $(SPECIAL_CLOBBER)
 
 flash:
 	@echo "use one of the following targets to flash the firmware"
-	@echo "  make flash512k - for ESP with 512kB flash size"
-	@echo "  make flash4m   - for ESP with   4MB flash size"
+	@echo "  make flash512k     - for ESP with 512kB flash size"
+	@echo "  make flash1m-dout  - for ESP with   1MB flash size and flash mode = dout (Sonoff, ESP8265)"
```
marcelstoer: > ESP8265

What kind of SoC is that? Did you mean to write ESP8285 here?

marcelstoer: @jpeletier we can merge the PR once this comment has been addressed.

jpeletier: @marcelstoer Yes, it was a typo. Fixed.

PR #2654, opened 2019-02-10
Minor fix to telnet Lua example and ftpserver Lua module

vsky279:
Fixes telnet lua example and ftpserver lua module in consequence of Expunge of integer timers (PR #2603)

### DIFF
--- a/lua_examples/telnet/telnet.lua
+++ b/lua_examples/telnet/telnet.lua
@@ -55,7 +55,7 @@ local function telnet_listener(socket)
  -- debug("entering sendLine")  
     if not s then return end
 
-      if fifo2l + fifo1l == 0 then -- both FIFOs empty, so clear down s
+    if fifo2l + fifo1l == 0 then -- both FIFOs empty, so clear down s
         s = nil     
      -- debug("Q cleared")        
         return
@@ -67,7 +67,7 @@ local function telnet_listener(socket)
       insert(fifo2,concat(fifo1))
    -- debug("flushing %u bytes / %u recs of FIFO1 into FIFO2[%u]", fifo1l, #fifo1, #fifo2) 
       fifo2l, fifo1, fifo1l = fifo2l + fifo1l, {}, 0
-	  end
+    end
 
     -- send out first 4 FIFO2 recs (or all if #fifo2<5)  
     local rec =  remove(fifo2,1)        .. (remove(fifo2,1) or '') ..
@@ -134,6 +134,7 @@ local function telnet_listener(socket)
   socket:on("disconnection", disconnect)
   socket:on("sent",          sendLine)
   node.output(queueLine, 0)
+  print(("Welcome to NodeMCU world (%d mem free, %s)"):format(node.heap(), wifi.sta.getip()))
 end
 
 local listenerSocket
@@ -143,11 +144,13 @@ return {
       wifi.setmode(wifi.STATION, false)
       wifi.sta.config { ssid = ssid, pwd  = pwd, save = false }
     end
-    tmr.alarm(0, 500, tmr.ALARM_AUTO, function()
+    local t = tmr.create()
+    t:alarm(500, tmr.ALARM_AUTO, function()
       if (wifi.sta.status() == wifi.STA_GOTIP) then
-        tmr.unregister(0)
-        print("Welcome to NodeMCU world", node.heap(), wifi.sta.getip())
-        net.createServer(net.TCP, 180):listen(port or 2323, telnet_listener)
+        t:unregister()
+        t=nil
+        print(("Telnet server started (%d mem free, %s)"):format(node.heap(), wifi.sta.getip()))
+        net.createServer(net.TCP, 180):listen(port or 23, telnet_listener)
       else
         uwrite(0,".")
       end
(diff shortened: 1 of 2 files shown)


lua_examples/telnet/telnet.lua
```diff
@@ -55,7 +55,7 @@ local function telnet_listener(socket)
  -- debug("entering sendLine")  
     if not s then return end
 
-      if fifo2l + fifo1l == 0 then -- both FIFOs empty, so clear down s
+	  if fifo2l + fifo1l == 0 then -- both FIFOs empty, so clear down s
```
marcelstoer: Indentation

lua_examples/telnet/telnet.lua
```diff
@@ -125,15 +125,16 @@ local function telnet_listener(socket)
     node.input(line)
   end
 
-  local function disconnect(s)
+  local function discontect(s)
```
marcelstoer: Spelling

lua_examples/telnet/telnet.lua
```diff
@@ -125,15 +125,16 @@ local function telnet_listener(socket)
...
     fifo1, fifo1l, fifo2, fifo2l, s = {}, 0, {}, 0, nil
     node.output(nil)
   end
 
   socket:on("receive",       receiveLine)
-  socket:on("disconnection", disconnect)
+  socket:on("disconnection", discontect)
```
marcelstoer: Spelling

lua_examples/telnet/telnet.lua
```diff
@@ -143,11 +144,11 @@ return {
       wifi.setmode(wifi.STATION, false)
       wifi.sta.config { ssid = ssid, pwd  = pwd, save = false }
     end
-    tmr.alarm(0, 500, tmr.ALARM_AUTO, function()
+    local tmr_wait = tmr.create()
```
marcelstoer: A nit pick but I find `tmr_wait` a subpar name.

nwf:
Be advised that the telnet changes will conflict with #2650 (which also dealt with the fallout of #2603).  I don't mind merging this one first, if desired.

lua_examples/telnet/telnet.lua
vsky279: fixed

lua_examples/telnet/telnet.lua
vsky279: Oups, I thought I corrected this one form the initial comit.

lua_examples/telnet/telnet.lua
vsky279: renamed to `t` so it's compatible with #2650

vsky279:
@nwf Up to you. I think the sooner it is corrected the better. It can be quite misleading when the shipped lua examples do not work.

nwf:
@vsky279 Agreed re: timeline.  Probably best to commit this one and let #2650 linger a little bit for review.

lua_examples/telnet/telnet.lua
```diff
@@ -134,6 +134,7 @@ local function telnet_listener(socket)
   socket:on("disconnection", disconnect)
   socket:on("sent",          sendLine)
   node.output(queueLine, 0)
+  print(("Welcome to NodeMCU world (%d mem free, %s)"):format(node.heap(), wifi.sta.getip()))
```
nwf: I think this should be left where it is, below; the IP address is important to print out before the first connection, right?

lua_examples/telnet/telnet.lua
```diff
@@ -143,11 +144,11 @@ return {
...
       wifi.sta.config { ssid = ssid, pwd  = pwd, save = false }
     end
-    tmr.alarm(0, 500, tmr.ALARM_AUTO, function()
+    local t = tmr.create()
+    t:alarm(500, tmr.ALARM_AUTO, function()
       if (wifi.sta.status() == wifi.STA_GOTIP) then
-        tmr.unregister(0)
-        print("Welcome to NodeMCU world", node.heap(), wifi.sta.getip())
-        net.createServer(net.TCP, 180):listen(port or 2323, telnet_listener)
+        t:unregister()
```
nwf: It may be good practice, as @HHHartmann suggested for #2650, to add `t = nil` both here and in `ftpserver`.

marcelstoer:
@vsky279 could you please address Nathaniel's review comments so we can merge this afterwards?

lua_examples/telnet/telnet.lua
vsky279: Added.
However I don't understand why is it not redundant. I thought that when the timer is unregistered, `t` will be gced at first opportunity.

lua_examples/telnet/telnet.lua
vsky279: I like to see this message when I connect to the telnet server. This is the way it works in the simple telnet example.
At the initial location I would expect message like:
```Lua
print(("Telnet server started (%d mem free, %s)"):format(node.heap(), wifi.sta.getip()))
```

nwf: OK, so add something like that.

vsky279: Done

PR #2659, opened 2019-02-11
trailing spaces cleanup

### LINKED ISSUE #2648: Please cleanup trailing spaces
There are a lot of trailing spaces that reduces readability and should be cleaned up.

532910:
Fixes #2648

There are a lot of trailing spaces that reduces readability and should be cleaned up.


galjonsfigur:
It's great PR, but I don't think removing trailing spaces in the whole project is a good idea. For example the `fatfs` folder, where most of the files were just copied from the upstream version - this could cause some problems if someone would want to update that code.

Also It would be great to add something like pre-commit git hook to detect trailing whitespaces - without that all this work would be worthless.

marcelstoer:
> Also It would be great to add something like pre-commit git hook to detect trailing whitespaces

True, could be used for a number of other static checks as well. The major problem: hooks are local to the repository (clone) and are thus not distributed automatically. Hence, every developer would have to set up the pre-commit hook manually. GitHub does not support pre-receive hooks to do such things upstream.

devsaurus:
> where most of the files were just copied from the upstream version

I agree with @galjonsfigur that such sources should be left untouched. As consequence, this would apply to a number of sub-trees where we potentially have. E.g.
- app/fatfs
- app/lwip
- app/mbedtls
- app/spiffs
- app/sqlite3

> Hence, every developer would have to set up the pre-commit hook manually.

Can this be scripted - like a target in the Makefile?

marcelstoer:
>> Hence, every developer would have to set up the pre-commit hook manually.
>
> Can this be scripted - like a target in the Makefile?

Sure, a Git hook is just a shell script that becomes a Git hook by having a specific name and by residing in a specific folder (`make` could copy it there).

TerryE:
See my https://github.com/nodemcu/nodemcu-firmware/issues/2648#issuecomment-464124432.  We do need to review all of these changes and if we are going to do this, then we need a separate commit to recode any cases where trailing spaces are material first so that we can be confident by convention that we can do a bulk trailing space removal.  

Also once we've done this, then adding a tab and trailing whitespace check could also be implemented.

TerryE:
I did the comparison.  All looks OK.

PR #2660, opened 2019-02-11
support for dhcp ntp server (option 42)

### LINKED ISSUE #2651: Add support for DHCP NTP server
Please add support for ntp server over dhcp.

532910:
Fixes #2651

support for ntp server offered via dhcp (option 42)

### DIFF
--- a/app/modules/sntp.c
+++ b/app/modules/sntp.c
@@ -38,6 +38,8 @@
 #include "os_type.h"
 #include "osapi.h"
 #include "lwip/udp.h"
+#include "lwip/inet.h"
+#include "lwip/dhcp.h"
 #include "c_stdlib.h"
 #include "user_modules.h"
 #include "lwip/dns.h"
@@ -47,6 +49,8 @@
 #ifdef LUA_USE_MODULES_RTCTIME
 #include "rtc/rtctime.h"
 #endif
+
+struct netif * eagle_lwip_getif(uint8 index);
 
 #define max(a,b) ((a < b) ? b : a)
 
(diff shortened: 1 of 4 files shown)


c-e-p-x-u-o:
I'm not sure if ```struct netif * eagle_lwip_getif(uint8 index);``` is correct in sntp.c, but I've not found any other way to define it there.

app/modules/sntp.c
```diff
@@ -805,15 +809,23 @@ static int sntp_sync (lua_State *L)
...
     lua_newtable(L);
-    int i;
-    for (i = 0; i < 4; i++) {
-      lua_pushnumber(L, i + 1);
-      char buf[64];
-      c_sprintf(buf, "%d.nodemcu.pool.ntp.org", i);
-      lua_pushstring(L, buf);
-      lua_settable(L, -3);
+    struct netif *iface = (struct netif *)eagle_lwip_getif(0x00);
+    ip_addr_t ntp_addr = iface->dhcp->offered_ntp_addr;
```
pjsg: Is this always guaranteed to be set? What happens if the interface was manually configured an didn't use dhcp?

app/modules/sntp.c
```diff
@@ -48,6 +50,8 @@
 #include "rtc/rtctime.h"
 #endif
 
+struct netif * eagle_lwip_getif(uint8 index);
```
c-e-p-x-u-o: I'm not sure this is correct, but I haven't found another way to define it here.

app/modules/sntp.c
c-e-p-x-u-o: Is this a trick question?
The real answer is "I don't know and havn't tested this case". looks like it should be: 
```
ip_addr_t ntp_addr = iface->dhcp ? iface->dhcp->offered_ntp_addr : NULL;
```

marcelstoer:
Superseded by #2709

PR #2661, opened 2019-02-11
Bugfix/2632 luaccrash

### LINKED ISSUE #2632: dangling pointer bug in uzlib_deflate.c causes lua.cross to sometimes crash
### Expected behavior
not crash, generate a flash image

### Actual behavior
crashes, leaving a 0-length flash image file

### Test code
`lua.cross.exe -f -m 0x20000 -o lfs.img _init.lua dummy_strings.lua inspect.lua kosbo_lfs.lua`

I have already debugged the code and have a resolution posted below in the Details section

### NodeMCU version
master
HEAD:  11592951b90707cdcb6d751876170bf4da82850d
Merge pull request #2582 from nodemcu/dev
Arnim Läuger
Friday, December 07, 2018 3:47:49 PM

### Hardware
I am using a NodeMCU-dev board, rev 0.9, but that is not relevant for lua.cross

### Details
app\uzlib\uzlib_deflate.c:571, a buffer is freed:
`	FREE(dynamicTables);`

However, there are pointers into this buffer that are left dangling.  Specifically, oBuf, which is used at line 575.

You can see the memory layout as it was allocated at line 221, set into dynamicTables at line 217, and where oBuf was set to point within that allocated block at line 224.

To resolve this, I would move the FREE at line 571 to after the control block at 574-582.  (i.e., just before returning from the function.)

Since the Uzlib seems to be third party code, it might be useful to see if it was already fixed upstream.

This bug may be unnoticed because the freed memory block is being accessed quickly after freeing, and it's contents have likely not been damaged yet.  However it was crashing reliably for me in debug builds, because the debug malloc on MSVC blitzes freed blocks with a pattern.  In release builds, it usually ran OK, but occasionally did not.

ziggurat29:
Fixes #2632 .

- [ ] The code changes are reflected in the documentation at `docs/en/*`.  ??? I see no such path.

A block of memory is accessed after having been freed.  This was obscured by the fact that 'oBuf' is a pointer into the middle of the block 'dynamicTables', so when dynamicTables is freed, oBuf is pointing to freed memory.  Occasionally, luac.cross would crash because of this.

### DIFF
--- a/app/uzlib/uzlib_deflate.c
+++ b/app/uzlib/uzlib_deflate.c
@@ -568,7 +568,6 @@ int uzlib_compress (uchar **dest, uint *destLen, const uchar *src, uint srcLen) 
     status = UZLIB_OK;
   }
 
-  FREE(dynamicTables);
   for (i=0; i<20;i++) DBG_PRINT("count %u = %u\n",i,debugCounts[i]);
 
   if (status == UZLIB_OK) {
@@ -580,6 +579,8 @@ int uzlib_compress (uchar **dest, uint *destLen, const uchar *src, uint srcLen) 
     *destLen = 0;
     FREE(oBuf->buffer);
   }
+
+  FREE(dynamicTables);
 
   return status;
 }


TerryE:
> Looking at the code, you can see that at about line 224:
```c
  oBuf          = (struct outputBuf *)(dt+1);
```
so, oBuf is pointing /into/ the dt chunk. Also note, dt is an alias of dynamicTables.

Uuuuaaaargghhh @ziggurat29.  Mea cupla; mea culpa.

PR #2669, opened 2019-02-14
Pwm2

### LINKED ISSUE #2667: new pwm module (pwm2)
### Justification

Feedback is welcome! 

### Workarounds
Are there any? ;)

fikin:
Fixes #2667.

Lets review the code and think of changes if needed.


pjsg:
Can you restage this to the dev branch please?

galjonsfigur:
The only problem I can see is that `ESP8266_new_pwm` project is GPL licensed and `nodemcu-firmware` is MIT. I don't know much about licenses but this could be a potential problem. Other than that this PR is very good idea - better PWM can be used to make more advanced things.

fikin:
huh, that would certainly become a bother.
i ignored it until now but looking at some previous discussions for license change (https://github.com/StefanBruens/ESP8266_new_pwm/issues/16), feel kind of not very optimistic...
is rewrite the only way out of this?

marcelstoer:
> `ESP8266_new_pwm` project is GPL licensed

Yep, that's a no-go :(

nwf:
IANAL, IANYL, TINLA, &c, but: because `nodemcu` is licensed MIT, it can always be mixed in to GPL code, and the result shared under the GPL (just as it can be used as part of proprietary products).  There is nothing stopping a developer or end-user from mixing `ESP8266_new_pwm` and `nodemcu`, as this PR does, but the result will be that the whole thing is GPL and must be distributed as such.

None of this is meant to say that we should merge this PR -- indeed, I think the project will only accept such a module if it can also be MIT licensed -- merely to say that if the lack of interesting PWM support is a problem for down-stream users right now, they are free to do the integration themselves and take on the different licensing regime.

fikin:
i'll re-write it under mit license so that it can make it to the master. would that be ok with you guys?

nwf:
Be careful when rewriting that you do not implicitly create a derivative work; just typing in new code with the old code on screen is probably not legally sufficient.  I think past efforts on these lines have involved two people, one of whom documented the old thing's behavior as pertained to the wider system, and the other looked only at that documentation to produce the new software.  Again, though, I am not an IP lawyer.

fikin:
fair enough. i've seen the code but i't rather do it my own way.
btw, just to confirm, is using hw_timer1 for pwm purposes ok from nodemcu pov?

PR #2671, opened 2019-02-16
Revise fifo{,sock}

nwf:

@TerryE raises a number of good points on the `fifo` and `fifosock` PR (#2650) which was perhaps prematurely committed.  This PR is to address those comments.


nwf:
@marcelstoer If @TerryE signs off on these small fixups, can they be scheduled to go before the next drop?

TerryE:
I haven't had the time to dynamic test this, but in general this is an excellent example.  No doubt we can refine our coding and implementation styles as LFS etc. evolves. 

My only real quibble is that I feel that @nwf Nathaniel is doing himself a disservice by listing my as the principle author given that the bulk of the implementation is his.  He should take the main credit here not me.

nwf:
@TerryE All I did was copy code from the telnet server example and introduce bugs. ;)

PR #2695, opened 2019-03-15
Next master drop

marcelstoer:
https://github.com/nodemcu/nodemcu-firmware/milestone/12

PR #2697, opened 2019-03-16
Minor fix to allow building of luac.cross.int

### LINKED ISSUE #2696: Building an Integer firmware does not create luac.cross.int
### Expected behavior

If you build an Integer build, then luac.cross.int ought to be built so that it can be used by the other tools.

### Actual behavior

It always buiold luac.cross

### NodeMCU version
dev

pjsg:
Fixes #2696

This just changes the Makefile to pick the right target name.

### DIFF
--- a/app/lua/luac_cross/Makefile
+++ b/app/lua/luac_cross/Makefile
@@ -1,7 +1,7 @@
 #
-# This Make file is called from the core Makefile hierarchy with is a hierarchical
-# make wwhich uses parent callbacks to implement inheritance.  However is luac_cross
-# build stands outside this and uses the host toolchain to implement a separate
+# This Makefile is called from the core Makefile hierarchy which is a hierarchical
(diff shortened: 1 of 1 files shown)


HHHartmann:
OK that change was in there before and reverted by "Add compression to LFS images (#2448)"
There were some other things removed from the that Makefile in that commit. Care to also fix these?

pjsg:
I think that the remaining differences were just the commands and the commands in the 'test' target. All the rest looked plausible.

TerryE:
Seems fine by me 😊👍

devsaurus:
Should this go into the pending master drop #2695?

HHHartmann:
Sure would save some time not to have to explain why in an int build it is not called luac.cross.int and why a docker int build generates Integer compatible LFS images that are called *float*.

Confused? Then add it to the pending master drop.

HHHartmann:
Its also code that was in before and worked. So the risk should be minimal.

PR #2707, opened 2019-03-29
LFS: add ressources to LFS image

### LINKED ISSUE #2691: LFS Add storing arbitrary resource files in LFS
### Missing feature
There is no way to store ressource files in LFS images. They have to be stored separately in SPIFFS

### Justification
Better provisioning and possible saving RAM for smaller files, not having to read them into strings.

### Workarounds
Use SPIFFS

I created the possibility to add files to LFS images and read them back as dtrings.

I did this by adding a switch -r to be used after all lua files, marking all further files as ressources.
I created a spechial reader, which wraps the file content in

```lua
return [=====[<File content here>]=====]
```

reading can be done like accessing the lua files.
I would also create a new method in _init.lua
``` lua
getressource("basename")
```
or whatever it should be called.

HHHartmann:
Fixes #2691.

This PR adds the possibility to store entire files as resources in LFS images

Added a switch -r which declares all files after it to be stored as resources
Use luac.cross like this
``luac.cross -f  start.lua -r res1.txt``

Access it like this:

``` lua
print(LFS.getresource("res1.txt"))
print(loadfile("getresource.lua")("res1.txt"))
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ sdk/
 cache/
 .ccache/
 local/
+luac.cross
+luac.cross.int
 user_config.h
 server-ca.crt
 luac.cross
@@ -14,3 +16,6 @@ tools/toolchains/
 .cproject
 .project
 .settings/
+
+#ignore VisualStudio files
+.vs/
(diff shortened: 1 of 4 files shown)


TerryE:
See #2691 for my my reasons for my :-1:

PR #2709, opened 2019-03-29
support for dhcp ntp server (option 42) reloaded

### LINKED ISSUE #2651: Add support for DHCP NTP server

HHHartmann:
Fixes #2651.

- [x] The code changes are reflected in the documentation at `docs/en/*`. (why do I get an old template here with wrong path /en)

support for ntp server offered via dhcp (option 42)

since #2660 is unchanged for over one month I try this one.

I took the changes of #2660 and added some fixes

Tested with DHCP server with and without this feature


app/lwip/core/dhcp.c
```diff
@@ -1463,6 +1472,10 @@ dhcp_parse_reply(struct dhcp *dhcp, struct pbuf *p)
         LWIP_ASSERT("len >= decode_len", len >= decode_len);
         decode_idx = DHCP_OPTION_IDX_DNS_SERVER;
         break;
+      case(DHCP_OPTION_NTP):
+        LWIP_ASSERT("len == 4", len == 4);
```
pjsg: What happens if the DHCP server puts two (or more) NTP addresses in that field?

app/modules/sntp.c
```diff
@@ -878,4 +890,4 @@ static const LUA_REG_TYPE sntp_map[] = {
   { LNILKEY, LNILVAL }
 };
 
-NODEMCU_MODULE(SNTP, "sntp", sntp_map, sntp_open);
+NODEMCU_MODULE(SNTP, "sntp", sntp_map, sntp_open);
```
pjsg: Can you add the newline at end of the file please?

app/lwip/core/dhcp.c
HHHartmann: Hmm just checked the documentation and saw this option.
Currently I guess it just ignores all but the first. :-(

pjsg: What does the assert do if there is more than one?

pjsg: I guess that the code ought to be
```
        LWIP_ASSERT("len % 4 == 0", len % 4 == 0);
        decode_len = LWIP_MIN(len, 4 * DNS_MAX_SERVERS);
        LWIP_ASSERT("len >= decode_len", len >= decode_len);
```

HHHartmann: OK found the template in DHCP_OPTION_DNS_SERVER
Will try but my C is not that strong.
Should I add a 
``` C
#define NTP_MAX_SERVERS 2
```
to ``lwipopts.h`` then?`

Or could we just fix NTP_MAX_SERVERS to 1 and be done?

marcelstoer: @pjsg could you guide the author to finish this PR?

pjsg: Sorry about the delay -- I think that I'm nervous about the whole use of LWIP_ASSERT -- I don't think that it does what anybody expects. I.e. in the production firmware it does nothing. Thus using it to validate (??) data received over wifi is not a good idea. You could argue that we don't need to worry about (in this case) malicious DHCP servers (or other malicious devices that can send DHCP responses) -- but I think that that would be a dangerous argument.

Note that I think that the existing code is dangerous and I'm not sure what the best path forward is.

marcelstoer: my2cents: We have had those asserts in the code since its inception in 2014. It has seen very few changes since, none of them substantially changed anything and it largely just seems to work. That all doesn't mean you didn't raise a valid point but that I think these concerns shouldn't hold up this PR.
@HHHartmann WDYT?

marcelstoer:
Needs to resolve the conflict in `sntp.c` (`"c_stdlib.h"` vs `<stdlib.h>`).

app/include/user_version.h
```diff
@@ -11,10 +11,10 @@
 #define NODE_VERSION_STR(x)	#x
 #define NODE_VERSION_XSTR(x)	NODE_VERSION_STR(x)
 
-#define NODE_VERSION		"NodeMCU " ESP_SDK_VERSION_STRING "." NODE_VERSION_XSTR(NODE_VERSION_INTERNAL)
```
marcelstoer: The changes in this file need to be reverted.

HHHartmann: Ups. done now.

PR #2710, opened 2019-04-03
ESP32: Joystick Lua library and example code

chilipeppr:

This pull request is to contribute code for a pure Lua Joystick controller.

The library can be used to control analog joysticks that are implemented using variable resistors where you feed in 3.3v and get out a variable voltage based on position of the joystick. Joysticks like the Playstation portable or Xbox joysticks work this way and are readily available on Amazon, Ebay, or Aliexpress.

Working example of using this library:
https://www.youtube.com/watch?v=ITgh5epyPRk

There is no license needed to use/modify this software. It is given freely
to the open source community. Modify at will.


devsaurus:
Thanks @chilipeppr

PR #2712, opened 2019-04-05
Crypto fix EM:0 error, and other memory leaks, plus minor typos in doc

### LINKED ISSUE #2703: crypto.mask on a zero length string says 'e:m 0'
### Expected behavior
It should just return an empty string
### Actual behavior
It prints out 'E:M 0'
### Test code

```Lua
> =crypto.mask("", 'bar')
E:M 0

```
### NodeMCU version
dev

This is because the code looks like:

```c
  int len, mask_len;
  const char* msg = luaL_checklstring(L, 1, &len);
  const char* mask = luaL_checklstring(L, 2, &mask_len);

  if(mask_len <= 0)
    return luaL_error(L, "invalid argument: mask");

  int i;
  char* copy = (char*)c_malloc(len);

  for (i = 0; i < len; i++) {
    copy[i] = msg[i] ^ mask[i % mask_len];
  }
  lua_pushlstring(L, copy, len);
  c_free(copy);

  return 1;

```
and calling c_malloc(0) causes this to be printed....

TerryE:
Fixes #2703

Bugfix and tidy up as discussed in #2703

### DIFF
--- a/app/modules/crypto.c
+++ b/app/modules/crypto.c
@@ -75,22 +75,21 @@ static const char* bytes64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwx
   */
 static int crypto_base64_encode( lua_State* L )
 {
-  int len;
+  int len, i;
   const char* msg = luaL_checklstring(L, 1, &len);
-  int blen = (len + 2) / 3 * 4;
-  char* out = (char*)c_malloc(blen);
-  int j = 0, i;
+  luaL_Buffer out;
+
+  luaL_buffinit(L, &out);
   for (i = 0; i < len; i += 3) {
     int a = msg[i];
     int b = (i + 1 < len) ? msg[i + 1] : 0;
     int c = (i + 2 < len) ? msg[i + 2] : 0;
-    out[j++] = bytes64[a >> 2];
-    out[j++] = bytes64[((a & 3) << 4) | (b >> 4)];
-    out[j++] = (i + 1 < len) ? bytes64[((b & 15) << 2) | (c >> 6)] : 61;
-    out[j++] = (i + 2 < len) ? bytes64[(c & 63)] : 61;
+    luaL_addchar(&out, bytes64[a >> 2]);
+    luaL_addchar(&out, bytes64[((a & 3) << 4) | (b >> 4)]);
+    luaL_addchar(&out, (i + 1 < len) ? bytes64[((b & 15) << 2) | (c >> 6)] : 61);
+    luaL_addchar(&out, (i + 2 < len) ? bytes64[(c & 63)] : 61);
   }
-  lua_pushlstring(L, out, j);
-  c_free(out);
+  luaL_pushresult(&out);
   return 1;
 }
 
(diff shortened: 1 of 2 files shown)


jmattsson:
Sounds good to me. Code cleanups ftw! Could you drag this over to the esp32 branch too when ready please?

TerryE:
I was having a WhatsApp aside with @HHHartmann on this and there are two separate change strategies that I am implementing here:
-  The use if `luaL_Buffer` handling for dynamic strings.  I feel this one is pretty non-contentious because this is how the rest of the Lua runtime does it and by following this template we get the ecosystem 'for free'.
-  Using a single userdata block to replace three separate mallocs.

It is this second item that Gregor raised concerns about:

> I think allocating one chunk of memory for 3 separate items is not a good Idea in terms of readability of the code.  It is just an extra load to keep in mind that they technically belong together but not logically and not clearly visible in the code.

My respond is that the old code does 3 allocations for three separate structures that are a transactional set: in that all three must be allocated or non, and ditto for freeing. If you don't maintain this integrity then the code will experience resource leakage and so it (partially) implements backout and recovery logic.  Putting all three in a single userdata means that we get this integrity as a free side-effect and because the resource being used is Lua userdata which is a Lua collectable object, then again GC is also implemented by the Lua ecosystem.  Incidentally this approach is also used in `app/lua/lstring.c` where the `TString` header and the string itself are allocated and freed as a single unit.

If anyone has any suggestions how to code this in a more clear way then please make them.

nwf:
At a quick skim, I think your approach is fine, @TerryE, though I would be tempted to hide things like `void *ctx = (char *)(dudat + 1);` behind macros, because this relies on knowing that the type of `dudat` is a `digest_user_datum_t` and that the `ctx` is next.  Maybe use something like
```
#define DUDAT_SIZE(ctxsz, opadsz) ((sizeof digest_user_datum_t) + ctxsz + opadsz)
#define DUDAT_CTX(d) ((void *)(((digest_user_datum_t *)d) + 1))
#define DUDAT_OPAD(d, ctxsz) ((char *)(((char *)(DUDAT_CTX(d)))+ctxsz))
```
just to centralize the layout?  (It's a pity that C doesn't have support for structures with multiple flexible members.)

jmattsson:
I think the main way to make it easier to read would be to turn the user data into an actual struct, with three members (header, ctx, hmac), but that would mean we'd need to work out the min/max size needed for the ctx and hmac at compile time. Possibly worth doing, but honestly, this is embedded, and doing aggregate allocations is a common enough strategy imo.

Maybe a middle way would be to move the allocation into the mechs, and only expose an opaque pointer which is handed back to the mech to obtain the header / ctx / mac. A common implementation across all mechs should suffice I'd imagine. Slightly larger code size I'd expect, but could ease readability.

TerryE:
> `void *ctx = (char *)(dudat + 1);`

Or even just:
```C
`void *ctx = dudat + 1;. // The ctx block immediately follows the user data header
```
😊 Terry

nwf:
@TerryE Whatever style works best for you.

PR #2732, opened 2019-04-20
1st Tranch of SDK 3.0 follow up changes

### LINKED ISSUE #2689: Upgrade to SDK 3.0 -- update 1
### Key points from of previous discussions

-  Espressif currently state that [Espressif SDK 3.0](https://github.com/espressif/ESP8266_NONOS_SDK/releases/tag/v3.0) will be the final release of the non-OS variant of the SDK, and therefore further planned NodeMCU releases will be based on this.
-  As well various bugfixes and enhancements SDK 3.0 has 2 major changes that impact on NodeMCU firmware.
    -  It now integrates a version of @jmattsson Johny's unaligned exception handler which enables a lot of the SDK constant data to be moved in flash mapped memory and thus freeing up some 12Kb extra RAM for application use.  Because of this, I have now decommissioned our original version.
    -  Note that 3.0 also allows the ICACHE size to reduced from 32Kb to 16Kb in theory making another 16Kb RAM available for application use. However, in practice such a small cache only works effectively for small footprint firmware applications, and  NodeMCU performs _terribly_ with option enabled so _**we will not support builds with a reduced 16 Kb cache**_.
     -  It exposes a partition table (PT) interface similar to that implemented for the ESP32, and this release of the firmware will make full use of this partition table.
-  We will have dedicated LFS and SPIFFS partitions that are configurable within the PT.  NodeMCU keeps a static copy of the PT at a fixed flash offset 0x10000, enabling it to be easily accessed and updated during module provisioning.
-  The firmware addresses these partitions through the PT. 
-  Our standard PT configuration moves the 3 small SDK System Partitions (`RF_CAL`, `PHY_DATA` and `SYSTEM_PARAMETER`) into unused flash between the 0x0000 segments and 0x10000 segment)
-  The associated PR covers the above scope.  The following are out of scope but will be covered by subsequent PRs, so discuss of these is currently outside the scope of this issue.
    -  Move the TLS certificate store into its own PT partition.
    -  Implement Firmware OTA upgrade.
-  The SDK 3.0 version will continue to support the GDB stub.  This now seems to work fine but there might be some interoperability issues because our version of the non-aligned exception handler did do exception handler chaining.
-   The SDK 3.0 `user_pre_init()` entry-point has the ICACHE_FLASH_ATTR qualifier -- that is unlike the SDK 2.0 `user_trampoline` code, ICACHE has already been enabled.
-  The SDK 3.0 firmware drops support for `flash_detect_size_byte()` as this can now be easily set during configuration using `esptool.py` or any wrapper. 

### Superseded issues
-   #2467.  (Previous SDK 3.0 issue) I've agreed with @nwf Nathaniel that it will probably be easier to committers and members to follow discussions is that we wrap up the old issue and restart with a clean baseline.  Notwithstanding this, this is still worth a browse for background discussion.

TerryE:
Fixes #2689

- [x] I have ~~thoroughly~~ tested my contribution.

This is a fairly large change, with a summary of the details below 
-  **`app/platform/platform.c`**. Add support for R/W of Reboot Config Records (RCRs) to store configuration changes that persist across reboots using flash NAND writing rules.  See inline comments for documentation. Flash erase is only carried out when GC of the RCR is required.  <br>&#x1F4D9; _Note: IMO, it would be well worthwhile using this API to store the TLS certificates._  
-  **`app/platform/platform.h`**. Header changes to support the above.
-  **`app/user/user_main.c`**. This declares a static initial partition table RCR.  A fairly major rework of the `user_pre_init()` code; this reads the current PT RTR and if this is the compiled one does calculation to work out the actual PT in RAM for the SDK to use.  It then writes this updated RCR back for use on subsequent boots, so this calculation is one-time only.  <br>Note that this is designed to interoperate with the `tools/nodemcu-partition.py` utility and the `node` API so that you can configure the LFS or SPIFFS partitions at compile time, over UART configuration, or at runtime.   
-  **`app/include/user_config.h`**. Following review feedback, I have changed this so that you can still set the compile-time default partition sizes for LFS and SPIFFS here. 
-  **`tools/nodemcu-partition.py`**. Supports simple configuration of the LFS or SPIFFS partitions and downloading of LFS and SPIFFS images via this script.  For example the following commands download a firmware image, then sets the LFS partition to 128KB and loads it with an LFS image over the UART:
```bash
  esptool -p /dev/ttyUSB0 -b 460800 --after no_reset write_flash \
    -fs 4MB --flash_mode=dio 0x00000  bin/0x00000.bin 0x10000 bin/0x10000.bin
  nodemcu-partition.py -ls 128K -lf bin/lfs.img
```
-  **`app/modules/node.c`**. Addition of `node.readrcr()` and `node.writercr()` if the macro DEVELOPMENT_TOOLS is defined.  <br> &#x1F534; Not documented since these functions are only available for development. <br> &#x1F534; the extra functions to query and to update the PT need to be added. 
-  **`app/libc/c_string.h`**. Minor change to add `c_memmove` mapping to `ets_memmove`.
-  **`ld/nodemcu.ld`**. Force KEEP() on the RCR page.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ else
   CCFLAGS += -O2
 endif
 
+
 # Handling of V=1/VERBOSE=1 flag
 #
 # if V=1, $(summary) does nothing
@@ -75,7 +76,16 @@ endif
 #  If any developers wish to develop, test and support alternative environments 
 #  then please raise a GitHub issue on this work.
 #
-ifeq (,$(findstring linux,$(MAKE_HOST)))
+
+ifndef $(OS)
+  ifneq (,$(findstring linux,$(MAKE_HOST)))
+    export OS := linux
+  else
+    export OS := windows
+  endif
+endif
+
+ifneq ($(OS),linux)
   #------------ BEGIN UNTESTED ------------ We are not under Linux, e.g.under windows.
 	ifeq ($(XTENSA_CORE),lx106)
 		# It is xcc
@@ -116,7 +126,7 @@ else
 
   UNAME_S := $(shell uname -s)
   UNAME_P := $(shell uname -p)
-  ifeq ($(MAKE_HOST),x86_64-pc-linux-gnu)
+  ifeq ($(OS),linux)
     ifndef TOOLCHAIN_ROOT
       TOOLCHAIN_VERSION = 20181106.0
       GCCTOOLCHAIN      = linux-x86_64-$(TOOLCHAIN_VERSION)
@@ -142,7 +152,6 @@ else
 	FIRMWAREDIR = ../bin/
   WGET = wget --tries=10 --timeout=15 --waitretry=30 --read-timeout=20 --retry-connrefused
 endif
-#############################################################
 
 GITHUB_SDK       = https://github.com/espressif/ESP8266_NONOS_SDK
 GITHUB_ESPTOOL   = https://github.com/espressif/esptool
@@ -301,7 +310,6 @@ $(TOP_DIR)/cache/esptool/v$(ESPTOOL_VER).tar.gz:
 $(TOP_DIR)/sdk/.extracted-$(SDK_VER): $(TOP_DIR)/cache/$(SDK_FILE_VER).zip
 	mkdir -p "$(dir $@)"
 	$(summary) UNZIP $(patsubst $(TOP_DIR)/%,%,$<)
-	$(summary) HOST $(MAKE_HOST)
 	(cd "$(dir $@)" && \
 	 rm -fr esp_iot_sdk_v$(SDK_VER) ESP8266_NONOS_SDK-* && \
 	 unzip $(TOP_DIR)/cache/$(SDK_FILE_VER).zip \
--- a/Makefile
+++ b/Makefile
@@ -2,21 +2,32 @@
 #
 .NOTPARALLEL:
 
-TOOLCHAIN_VERSION:=20181106.0
-
-# SDK base version, as released by Espressif
-SDK_BASE_VER:=3.0
-SDK_VER:=$(SDK_BASE_VER)
-SDK_DIR_DEPENDS:=sdk_extracted
-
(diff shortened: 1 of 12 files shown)


TerryE:
The issue with the Makefile is that the `sdk_pruned` target uses `xtensa-lx106-elf-ar` but is not explicitly dependent on toolchain.  I needed to add this. Something bizarre seems to be going on since the make works fine under Docker on my PC but fails in Travis.

TerryE:
OK, the penny has dropped re the make failure.  I dropped the use of `$(OS)` because this isn't defined in standalone make, and used GNU make builtin $(MAKE_HOST) instead -- except that the Travis make uses OS and MAKE_HOST is blank. Ooops.  Some conditional magic is needed to use OS if defined and MAKE_HOST otherwise so the conditional logic works properly.

TerryE:
Damn, `ci-build-linux.sh` doesn't pass in the OS environment variable either.  So I will default to linux if neither $(OS) or $(MAKE_HOST) are set.

HHHartmann:
Maybe it makes sense to export the OS variable in travis. yml then. 
Travis uses $TRAVIS_OS_NAME which should be set in the script too. 
In fact I only set $OS as shorthand in the first place to make the code clearer. 
So setting $MAKE_HOST in travis also Sounds like a way to go.

TerryE:
@HHHartmann Gregor, to be honest it doesn't really matter which way.  I find debugging the predicate logic in Makefiles a bit of a PITA, but the real lightbulb moment was just that there is something subtly different about the Travis make environment, and that I needed to do a couple of commit / pushes just to work out exactly what.  They'll get rolled up in the  "squash and merge" anyway.  I'll add the node API calls today and merge this into dev.

TerryE:
I have now added the get / set partition table calls to the node library, and documented these in `node.md`.  This tranche is ready for pushing to dev.

PR #2742, opened 2019-04-29
Updated ROTables 1st tranch

### LINKED ISSUE #2726: Lua 5.3 Compatible ROTables
### New feature

The current NodeMCU modules use a ROTable implementation taken directly from the **eLua** implementation.  This is a design compromise that does allow a form of RO Table to be implemented in a Flash memory compatible data structure, but in a way that gives poor runtime performance and uses an address-range test to determine whether a table pointer points to a `Table` record in RAM or a vector of ROtable entries in Flash.  

The serial search algorithm for ROTables gives poor access performance especially for table-miss accesses.

It is going to be very difficult to implement the same approach for Lua 5.3 which has already established an implementation architecture of using record variants for such subtypes.

### Justification

So it makes sense to move to a Lua 5.3 compatible ROTable implementation as this will give performance benefits and this is the main aspect where our current Lua C API used in our modules will need reworking to be compatible with Lua 5.3.  We can therefore prepare all of modules as a separate PR step to the Lua 5.3 migration itself.   There are also performance benefits in doing this now.

### Scope and impacts

In our `app/modules` we have the following numbers of ROTables in our C files: 45 × 1, 15 × 2, 4 × 3, 3 × 4, 1 × 5, 1 × 6.  (e.g one module `net.c` has 6 ROTable declarations.).  We have another 6 in `app/lua` and 1 in `app/pm` These all need updating to the new format.   

The way to do this is to use a special to purpose Python conversion script that will automate the source conversion.  So for example this will convert:
```C
static const LUA_REG_TYPE wifi_ap_dhcp_map[] = {
  { LSTRKEY( "config" ),  LFUNCVAL( wifi_ap_dhcp_config ) },
  { LSTRKEY( "start" ),   LFUNCVAL( wifi_ap_dhcp_start ) },
  { LSTRKEY( "stop" ),    LFUNCVAL( wifi_ap_dhcp_stop ) },
  { LNILKEY, LNILVAL }
};
``` 
into the Lua 5.3 format:
```C
LROT_BEGIN(wifi_ap_dhcp)
  LROT_FUNCENTRY(config, wifi_ap_dhcp_config)
  LROT_FUNCENTRY(start,  wifi_ap_dhcp_start)
  LROT_FUNCENTRY(stop,   wifi_ap_dhcp_stop)
LROT_END(wifi_ap_dhcp, NULL, 0)
```

### Implementation notes and wrinkles:

-  The table entries will always be in ascending order to allow the access routine to do a binary chop access.
-  One of the main performance issues is that the Lua VM needs to check any table access miss for the presence of a meta table and meta methods such as `__index` this involves a table scan which can fail and as a result is currently a full table scan.  With this PR ROTables are now implemented as a variant Table record pointing to an ordered `luaR_entry` vector.  The Table record includes a pointer its metatable and if this is blank then no metatable search is required.
-  The aim with the batch conversion is that the old to new syntax will result in a total clean compile or not.  So whilst the size of the PR is large, the actual risks are largely fixed.
-  We probably will need a temporary separate `dev-LROTABLE` branch to trial this one -- its not one that I would want to push into `dev` without multiple testers checking it out.

TerryE:
Fixes #2726 

### Details

This is a `large` change.  I am going to need to rebaseline against dev at the point of merge, but I wanted to make this available for review comment before this.  The scope of the changes call into the following broad categories:

-  **Changes to the rotable.h macros to support the new declarative style**.  These are largely in `modules.h` and `rotable.h`.  Note that the eLua patch used two separate headers, `rotable.h`and `rodefs.h`.  This second file has been dropped and the residual defines moved into  `rotable.h`.

-  **Retiring dead eLua options and code**.  For at least the last 4 years we have had to build the firmware (and later luac.cross) with a couple of hardwired eLua optimisation settings, and the Lua VM includes dead code that we can't use, so I've removed these options and stripped out this dead code.
  
-  **Minor optimisations to the rotable lookup code**.   Since we only support character keys, there is no point in having a rotable key type as this doubles the number of flash work accesses in a ROTable lookup scan.  Since the new declarative style hides these implementation details we do this doubling largely "for free".

-  **Change to the Lua RTS to support the new macros.** The base Lua libraries such as `string` (and `os` in the case of `luac.cross`) ave been updated to use the new macros.  

-  **Bulk largely automated edits of all of the module C files to use the new format ROTable declarations**.  In terms of files this was the bulk of the change, but this was largely automated though a bunch of ah-hoc `sed` scripts.  Not that none of the functionality has been changed, though lookup is faster due to the optimisations discussed above.


app/lua/lrotable.h
```diff
@@ -7,39 +7,46 @@
 #include "luaconf.h"
 #include "lobject.h"
 #include "llimits.h"
+#include "lrotable.h"
```
jmattsson: Including `lrotable.h` in `lrotable.h`? Seems a bit too pious ;)

app/lua/lrotable.h
```diff
@@ -7,39 +7,46 @@
...
+#define LROT_BEGIN(t)        static const LUA_REG_TYPE t ## _map [] = {
+#define LROT_PUBLIC_BEGIN(t) const LUA_REG_TYPE t ## _map[] = {
+#define LROT_EXTERN(t)       extern const LUA_REG_TYPE t ## _map[]
+#define LROT_TABENTRY(n,t)   {LRO_STRKEY(#n), LRO_ROVAL(t ## _map)},
+#define LROT_FUNCENTRY(n,f)  {LRO_STRKEY(#n), LRO_FUNCVAL(f)},
+#define LROT_NUMENTRY(n,x)   {LRO_STRKEY(#n), LRO_NUMVAL(x)},
+#define LROT_LUDENTRY(n,x)   {LRO_STRKEY(#n), LRO_LUDATA((void *) x)},
+#define LROT_END(t,mt, f)    {NULL, LRO_NILVAL} };
+#define LROT_BREAK(t)         };
+
```
jmattsson: Umm.. this looks like duplication from `lrodefs.h` - why not just `#include` that file?

app/lua/lstrlib.c
```diff
@@ -824,67 +824,37 @@ static int str_format (lua_State *L) {
...
+  LROT_FUNCENTRY( gsub, str_gsub )
+  LROT_FUNCENTRY( len, str_len )
+  LROT_FUNCENTRY( lower, str_lower )
+  LROT_FUNCENTRY( match, str_match )
+  LROT_FUNCENTRY( rep, str_rep )
+  LROT_FUNCENTRY( reverse, str_reverse )
+  LROT_FUNCENTRY( sub, str_sub )
+  LROT_FUNCENTRY( upper, str_upper )
+  LROT_TABENTRY( __index, strlib )
+LROT_END(strlib, NULL, 0)  // OR DO WE NEED LRTO_MASK_INDEX    **TODO**
```
jmattsson: There's a `TODO` left here :)

app/lua/ltable.c
```diff
@@ -580,7 +580,7 @@ const TValue *luaH_getnum (Table *t, int key) {
 
 /* same thing for rotables */
 const TValue *luaH_getnum_ro (void *t, int key) {
-  const TValue *res = luaR_findentryN(t, key, NULL);
+  const TValue *res = NULL;  // integer values not supported: luaR_findentryN(t, key, NULL);
```
jmattsson: Do we get a compile-time error if someone does try to register a numeric key in an RO table? If not, can we?

app/lua/ltable.c
```diff
@@ -739,11 +739,14 @@ int luaH_getn (Table *t) {
 
 /* same thing for rotables */
 int luaH_getn_ro (void *t) {
-  int i = 1, len=0;
-
+  int len = 0;
+#if 0
```
jmattsson: Some love wanted here, I think.

app/lua/lbaselib.c
```diff
@@ -475,51 +475,52 @@ extern const luaR_entry lua_rotable_base[];
  * ROTables, with the latter chained from the former using its __index meta-method.
  * In this case all library ROTables are defined in linit.c.
  */
+#if 0
```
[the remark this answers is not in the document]
TerryE: It turns out that this variant code isn't needed, so I removed it -- however I decided to leave it in for the review with a `#if 0` / `#endif` just to expose the change.  I am happy to cut it now that we have had the review.

app/lua/lbaselib.c
```diff
@@ -676,14 +674,12 @@ static void base_open (lua_State *L) {
...
+
   /* open lib into global table */
-  luaL_register_light(L, "_G", base_funcs);
-#if LUA_OPTIMIZE_MEMORY > 0
-  lua_pushvalue(L, -1);
-  lua_setmetatable(L, -2);
-  lua_pushrotable(L, (void *)BASE_ROTABLE);
-  lua_setglobal(L, "__index");
-#endif
+  luaL_register_light(L, "_G", &((luaL_Reg) {0}));
```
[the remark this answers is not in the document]
TerryE: `luaL_register_light()` does multiple things.  It registers the library in the global table (`_G` itself in this case) and also registers a list of functions.  There aren't any functions to be registered here since these are in a ROTable for the baselib, so we either call `luaL_register_light()` with an empty list or clone and copy the subset of the functionality that we used inline.  Doing the call is simpler than cut and paste.

app/lua/lrodefs.h
```diff
@@ -38,10 +37,17 @@
...
+#define LROT_TABLE(t)        static const LUA_REG_TYPE t ## _map[];
+#define LROT_TABLEREF(t)     ((void *) t ## _map)
+#define LROT_BEGIN(t)        static const LUA_REG_TYPE t ## _map [] = {
+#define LROT_PUBLIC_BEGIN(t) const LUA_REG_TYPE t ## _map[] = {
+#define LROT_EXTERN(t)       extern const LUA_REG_TYPE t ## _map[]
+#define LROT_TABENTRY(n,t)   {LRO_STRKEY(#n), LRO_ROVAL(t ## _map)},
+#define LROT_FUNCENTRY(n,f)  {LRO_STRKEY(#n), LRO_FUNCVAL(f)},
+#define LROT_NUMENTRY(n,x)   {LRO_STRKEY(#n), LRO_NUMVAL(x)},
+#define LROT_LUDENTRY(n,x)   {LRO_STRKEY(#n), LRO_LUDATA((void *) x)},
+#define LROT_END(t,mt, f)    {LRO_NILKEY, LRO_NILVAL} };
```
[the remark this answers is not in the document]
TerryE: Future proofing.  A ROTable is currently a `luaR_entry` vector, so the Table code need to include conditional logic `(is the address in Flash) ? do a ROTable reference : do a Table reference`  On the next version the `LROT_END()` will generate a `luaR_entry` terminator followed by a readonly `Table` rec which points to the vector.  Generating the extra fields in the module source now means that we can implement this change without needing to change the modules again.

The main advantage of ROTables having a `Table` record is that we can push the RO vs RW abstraction down a level an dump a lot of conditional testing in the Lua VM where it only needs to reference and manipulate the table as an entity.  The Table record contains useful stuff like the size of the table, and flags to indicate whether metas such as `__index`, `_newindex` and  `__gc` exist.

The [ROTable lookup cache](https://github.com/nodemcu/nodemcu-firmware/blob/aa2e29d3dd6deb982df52da05834441558668f2a/app/lua/lrotable.c#L54] now give a better than 95% hit rate for key-found searches of ROTables.  (I've tweaked the lookup algo to give more uniform coverage of the cache and to improve performance, BTW.)  There is still a residual performance issue with key misses (even though this code runs about 4× faster than the eLua version).  Having the abitity use the flabs byte to short-circuit frequent meta misses, and the length to allow you to sort the key list into ascending order and tag that you can do a binary search on the longer ROTables such as the baselib one, will pretty much  remove this penalty all together.
    So that in future we can re

app/lua/linit.c
```diff
@@ -15,89 +15,100 @@
...
- * registration.  This uses a PSECT naming convention to allow the lib and rotab
- * entries to be collected by the linker into consoliated tables.  The linker
- * defines lua_libs_base and lua_rotable_base.
+ * registration.  This uses a PSECT naming convention to allow the ROTable and
+ * initialisation function entries to be collected by the linker into two
+ * consoliated ROTables.  This significantly simplifies adding new modules and
+ * configuring builds with a small subset of the total modules.
+ *
+ * This linker-based approach is not practical for cross compiler builds which
+ * must link on a range of platforms, and were we down have control of PSECT
```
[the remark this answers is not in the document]
TerryE: Yup :smile: and also **were** -> **where**

app/lua/linit.c
```diff
@@ -15,89 +15,100 @@
...
 
+  lua_pushrotable(L, LROT_TABLEREF(lua_libs));
+  lua_pushnil(L);  /* first key */
   /* loop round and open libraries */
-  for (; lib->name; lib++) {
-    if (lib->func) {
-      lua_pushcfunction(L, lib->func);
-      lua_pushstring(L, lib->name);
-      lua_call(L, 1, 0);
+#ifndef LUA_CROSS_COMPILER
```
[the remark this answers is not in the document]
TerryE: Yup, but I use this test point so frequently for debugging startup that having left as a commented out hook is simpler.  I would prefer to leave this in as it is only a comment for production use.

app/lua/loadlib.c
```diff
@@ -480,7 +485,9 @@ static int ll_require (lua_State *L) {
     lua_pop(L, 1);
   }
   /* else must load it; iterate over available loaders */
-  lua_getfield(L, LUA_ENVIRONINDEX, "loaders");
+  lua_getfield(L, LUA_GLOBALSINDEX, "package");
```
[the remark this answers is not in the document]
TerryE: Oh this was a real doozy that took something for me to get my head around.  I still haven't worked out why the original version worked.  C closures can have an environment associated with then and in the case standard Lua of the init hook for a given library has the packages table as its environment. This also allow it to have a `__gc` meta entry.  This is all to support dynamic loaded libraries so that when the library is no longer used, then the GC can fire to unload it.  

None of this is relevant to our firmware where libraries are statically linked and referenced by a ROTable assembled during the link.  Also we use lightweight C closures which _can't_ have a separate environment associated.  The easiested work around  if just to reference the package table directly in _G.

app/lua/lrodefs.h
```diff
@@ -38,10 +37,17 @@
...
-#define LROT_END            {LNILKEY, LNILVAL}
+#define LROT_TABLE(t)        static const LUA_REG_TYPE t ## _map[];
+#define LROT_TABLEREF(t)     ((void *) t ## _map)
+#define LROT_BEGIN(t)        static const LUA_REG_TYPE t ## _map [] = {
+#define LROT_PUBLIC_BEGIN(t) const LUA_REG_TYPE t ## _map[] = {
+#define LROT_EXTERN(t)       extern const LUA_REG_TYPE t ## _map[]
+#define LROT_TABENTRY(n,t)   {LRO_STRKEY(#n), LRO_ROVAL(t ## _map)},
+#define LROT_FUNCENTRY(n,f)  {LRO_STRKEY(#n), LRO_FUNCVAL(f)},
+#define LROT_NUMENTRY(n,x)   {LRO_STRKEY(#n), LRO_NUMVAL(x)},
+#define LROT_LUDENTRY(n,x)   {LRO_STRKEY(#n), LRO_LUDATA((void *) x)},
```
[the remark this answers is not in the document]
TerryE: Yes we could add one but currently none of our modules use a key = <string value>, so I hadn't bothered.

app/lua/lrotable.h
TerryE: Sorry, `lrodefs.h` had got so slim that there was no point in keeping it since it did a dozen defines then included `lrotable.h`.  Combining the two actually shortened `lrotable.h`.  So one of my bulk edits was to replace any references  `lrodefs.h` by  `lrotable.h` and remove the duplicate where we ended up with a double occurrence.   I forgot this self-reference case. :astonished:

app/lua/lrotable.h
TerryE: See previous comment. `lrodefs.h` is now dead.

app/lua/lstrlib.c
TerryE: That's because it is TODO when I add proper Table support.  Can't do everything in one go.

app/lua/ltable.c
TerryE: There is no macro to do this, so they can't.  None of our current code uses numerically indexed ROTables, and there is roughly a 2× performance hit in leaving in this feature that we don't use, and so I removed it.

If there we want to add it back we should think about how we could do this but we should have this as a discussion on the issue, as there are trade-offs here.

app/lua/ltable.c
TerryE: So previous comments about whether to retain numeric indexes for ROTables.

jmattsson:
Thanks for the explanations & clarifications. I am satisfied you know what you're doing here :D

Happy for you to merge once you've done the discussed tidying up.

marcelstoer:
Not sure if you or Johny noticed but the CI build failed.

```
gpio.c:323:1: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'extern'
 extern int gpio_pulse_init(lua_State *);
 ^
In file included from ../include/module.h:5:0,
                 from gpio.c:4:
gpio.c:333:25: error: 'gpio_pulse_map' undeclared here (not in a function)
   LROT_TABENTRY( pulse, gpio_pulse )
                         ^
../lua/lrotable.h:16:39: note: in definition of macro 'LRO_ROVAL'
 #define LRO_ROVAL(v)    {{.p = (void*)v}, LUA_TROTABLE}
                                       ^
gpio.c:333:3: note: in expansion of macro 'LROT_TABENTRY'
   LROT_TABENTRY( pulse, gpio_pulse )
   ^
gpio.c: In function 'luaopen_gpio':
gpio.c:351:3: error: implicit declaration of function 'gpio_pulse_init' [-Werror=implicit-function-declaration]
   gpio_pulse_init(L);
```

TerryE:
@marcelstoer, thanks Marcel, yes I did.  There are a few edge cases with table references that barfed on the CI build.  I'll clean these up tomorrow before the merge.

TerryE:
If we don't have any  more issues, then I'll merge this into `dev` tomorrow.

PR #2745, opened 2019-05-05
new pwm2 module

### LINKED ISSUE #2667: new pwm module (pwm2)

fikin:
Fixes #2667, follow-up to #2669 

this is an alternative to pwm module, capable of working with many frequencies and many pins at the same time. and it works with cpu160mhz too. module's doc should be covering most of the aspects.


marcelstoer:
>  [x] This PR is for the `dev` branch rather than for `master`.

Nope, it's not. I tried changing the base to `dev` here but they have diverged too much. Please rebase to `dev` and force-push your branch. We can then try to change the base again here.

fikin:
of course, i'm totally brain dead. will submit new commit altogether.

PR #2747, opened 2019-05-06
new pwm2 module

### LINKED ISSUE #2667: new pwm module (pwm2)

fikin:
Fixes #2667, follow-up to #2669

an implementation capable of running up to few x100kHz with varying periods, sub-Hz and fraction of frequencies, support for all gpio pins and compatible with CPU160MHz.


app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+{
+  return system_get_cpu_freq() * 1000000;
+}
+
+static uint8_t getCpuTimerTicksDivisor()
+{
+  return system_get_cpu_freq() == 80 ? PWM2_TMR_MAGIC_80MHZ : PWM2_TMR_MAGIC_160MHZ;
+}
+
+static uint32_t findGCD(uint32_t n1, uint32_t n2)
```
TerryE: Yuk.  even Euclid uses remainder here.  Time for a couple of % operators. :smile:

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+{
+  sPin->resolutionInterruptCounterMultiplier = sPin->resolutionCPUTicks / timerCPUTicks;
+}
+
+static void updatePinPulseToInterruptsCounter(pwm2_pin_interrupt_t *iPin, pwm2_pin_setup_t *sPin)
+{
+  iPin->pulseInterruptCcounter = (sPin->pulseResolutions + 1) * sPin->resolutionInterruptCounterMultiplier;
+}
+
+static uint8_t getDutyAdjustment(const uint32_t duty, const uint32_t pulse)
```
TerryE: Conditional operator considered unhealthy :wink:

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+static void ICACHE_RAM_ATTR timerInterruptHandler(os_param_t arg)
+{
+  pwm2_interrupt_handler_data_t *data = (pwm2_interrupt_handler_data_t *)arg;
+  setGpioPins(data->enabledGpioMask, findAllPinOns(data));
+}
+
+//############################
+// lua bindings
+
+static int lpwm2_open(lua_State *L)
```
TerryE: Please read PiL 28.1 and review userdata use.  This should using `lua_newuserdata()` etc. instead of malloc.  See `app/lua/luac_cross/liolib.c` for example.  The meta method `__GC` allows you to integrate close down and cleanup into the Lua disposal system.

If you are using an OO paradigm then the MD is maintained by the calling application as an inscope variable and the methods called with the `:` syntax.  If you want to use a static singleton then you need to store a reference in the Registry to ensure that it is referenced for GC purposes.

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+    moduleData->setupData.pin[i].pulseResolutions = 0;
+    moduleData->setupData.pin[i].resolutionCPUTicks = 0;
+    moduleData->setupData.pin[i].resolutionInterruptCounterMultiplier = 0;
+  }
+  return 0;
+}
+
+static int lpwm2_print_setup(lua_State *L)
+{
+  ets_printf("pwm2 : isStarted %u\n", moduleData->setupData.isStarted);
```
TerryE: In general it is a bad idea to include any form of `printf()` in NodeMCU modules, as most production IoTs will not have the UART connected.  Better to return this as an array or even strings so that you leave it to the calling app to print this or return it over the network as, say, a JSON buffer.

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+}
+
+static int lpwm2_setup_pin_common(lua_State *L, const bool isFreqHz)
+{
+  if (moduleData->setupData.isStarted)
+  {
+    return luaL_error(L, "pwm2 : already started, stop it before setting up pins.");
+  }
+  int pos = 0;
+  while (true)
```
TerryE: This code does not match the documentation for `pwm2.setup_pin_hz(pin,frequencyAsHz,pulsePeriod,initialDuty [,frequencyDivisor])` (and ditto the sec variant).  Eg. why the `while` loop and why are these mandatory parameters optional?  I would have expected something more like:
```C
  int pin           = lua_tointeger(L, 1);
  int freq          = lua_tointeger(L, 2);
  int resolution    = lua_tointeger(L, 3);
  int freqFractions = lua_tointeger(L, 4);
  int initDuty      = lua_tointeger(L, 5);
  
  luaL_argcheck(L, pin > 0 && pin <= GPIO_PIN_NUM, 1, "invalid pin number");
  luaL_argcheck(L, freq > 0, 2, "invalid frequency");
  luaL_argcheck(L, resolution > 0, 2, "invalid resolution");
  // ...
```
or whatever is consistent with the documentation.  The Lua RTS rolls up error reporting as standard.  There's little to be gained by doubling up on this.

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+
+static int lpwm2_start(lua_State *L)
+{
+  if (moduleData->setupData.isStarted)
+  {
+    return 0;
+  }
+  if (!platform_hw_timer_init(PWM2_TMR_OWNER, FRC1_SOURCE, TRUE))
+  {
+    return luaL_error(L, "pwm2: currently platform timer1 is being used by another module.\n");
```
pjsg: In fact the platform_hw_timer stuff now supports multiple users and this will only fail if it fails to allocate memory.

docs/modules/pwm2.md
```diff
@@ -0,0 +1,195 @@
+# PWM Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2019-02-12 | [fikin](https://github.com/fikin) | [fikin](https://github.com/fikin) | [pwm2.c](../../../app/modules/pwm2.c)|
+
+Module to generate PWM impulses to any of the GPIO pins. 
+
+PWM is being generated by software using soft-interrupt TIMER1 FRC1. This timer is used by other modules like ws2812 and pwm, but only one of them can use it at given time. Also some parts of OS might mask the interrupt for awhile.
```
pjsg: Multiple users can use it at the same time....

pjsg: The note that the OS may mask interrupts for a while is a point that I think ought to be made more prominent. When I was working on the original PWM code, there was a fair amount of complexity to try and deal with interrupts being masked off (and then deciding what to do about it). What does the actual pulse train of a 50% 200kHz signal look like in the presence of (say) running an http server which serves up small blobs of data?

app/modules/pwm2.c
fikin: right, i see. i used the older version and didn't realized this was cooked up. 
i'd have to re-test and benchmark against this version before i can comment further on the timings.
with older version i struggled to compensate interrupt handler overhead when working in non-autoload mode. that's why i opted for autoloaded mode, which seemed to interrupt rather accurate.
with the new code, i gather hw_timer_isr_cb is the actual orchestrator of interrup delays, right? how is the compensation from varying interrupt handler execution handled?

app/modules/pwm2.c
fikin: i think i'm getting what you're saying. and i thought of same at some point of time but then i opted towards more procedural way.
if you look at the lua api of this module, do you really see a need to introduce client object? 
to me at least current interaction pattern seems consistent enough without it.

TerryE: @fikin Nikolay, Even if you adopt a more procedural way and use the `pwm2.` notation, it is still better to stay in the Lua ecosystem in terms of memory allocation and error handling.  For example, the Lua GC is lazy and an allocation failure will kick off a GC and retry.  If you do a plain malloc, then this will just fail.

app/modules/pwm2.c
fikin: @pjsg : i've ran same measurements against old and new platform_hw_timer code.
internal scheduling overhead for one interrupt, in average in cpu ticks, cpu80mhz: 
- previous version : for autoload mode=78, for non-autoload mode=157
- this version : for autoload mode=227, for non-autoload mode=508

internal pwm2 interrupt handler overhead is ~161+12*pins.
before total interrupt cost for 1pin was 78+161+12=251 cpu ticks, or ~318kHz max interrupt freq. 
with current code it drops to ~200kHz. 

i guess it is to be expected but frankly i'm not feeling excited. 
would you be ok if i bypass platform_hw_timer and hook directly to the interrupt (as it used to be before)?

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+} pwm2_pin_interrupt_t;
+
+typedef struct
+{
+  pwm2_pin_interrupt_t pin[GPIO_PIN_NUM];
+  uint16_t enabledGpioMask;
+} pwm2_interrupt_handler_data_t;
+
+typedef struct
+{
```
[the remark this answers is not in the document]
fikin: sure, is the formatting fitting now better?

app/modules/pwm2.c
fikin: he is being giant even millenniums old. i yet have to reach even half of century ;) anyhow, would now even euclidus be satisfied with this version?

app/modules/pwm2.c
fikin: i'm not catching this up, what do you mean?

app/modules/pwm2.c
fikin: nice spotting, the loop was from a past idea not relevant anymore. and thanks for the examples, now should be ok.

app/modules/pwm2.c
pjsg: You can certainly do that -- can you make sure that the build fails if PWM2 is included and any of the other timer related code is included? E.g. maybe add a #define HW_TIMER_EXCLUSIVE 1 if PWM2 is included and then don't compile in the platform_hw_timer functions if HW_TIMER_EXCLUSIVE is set. [Need to think about how to make this work with the travis build which turns on all modules]. 

We had to add support for multiple users of the hw_timer as people didn't read the documentation closely enough and would complain that using module A and module B at the same time would not work. 

Does this make sense?

jmd13391:
@pjsg "We had to add support for multiple users of the hw_timer as people didn't read the documentation closely enough and would complain that using module A and module B at the same time would not work."

If I recall correctly (don't have the PR numbers handy at the moment), multiple users functionality was added to allow gpio.pulse and pwm modules to work at the same time (this simultaneous functionality is critical to several of my projects). During the course of you providing this functionality, you discovered and corrected at least one bug and added clarification to the gpio.pulse docs. IMO, this had little to nothing to do with "not reading the closely enough".

Losing the multi-user hardware timer functionality (especially with gpio.pulse / pwm) would be detrimental to several of my ESP8266 based products already out in the field.

app/modules/pwm2.c
fikin: absolutely. let me think a bit and propose something for review with one of next updates.

fikin:
@jmd13391 : i think no one is thinking of loosing current capabilities to share timer1 between modules. not me at least. 
it is more about offering ability to either share it or reserve it exclusively for given module.

app/modules/pwm2.c
pjsg: thank you....

app/modules/pwm2.c
```diff
@@ -0,0 +1,553 @@
...
+  pwm2_interrupt_handler_data_t interruptData;
+  pwm2_setup_data_t setupData;
+} pwm2_module_data_t;
+
+//############################
+// module vars, lazy initialized, allocated only if pwm2 is being used
+
+static pwm2_module_data_t *moduleData = NULL;
```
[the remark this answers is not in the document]
fikin: @TerryE : perhaps the var name is giving a bit of wrong impression here.
strictly speaking pwm2.c contains driver and lua module code. this var and its associated malloc belong most certainly to the driver side.
given that, would you still see a need to use lua gc?

app/modules/pwm2.c
fikin: could you pls take a look at the latest version and let me know if you like the approach i've taken?

docs/modules/pwm2.md
fikin: ok, take a look at latest upload and lets see if the situation is improved.

fikin: i haven't tested it that far yet ;)  on relatively idling env i've checked up to 100kHz and haven't spotted side effects.

TerryE:
This has failed Travis as a consequence of #2742. Have a look at the other modules, e.g. `app/modules/file.c` for how to declare ROTables.

docs/modules/pwm2.md
```diff
@@ -0,0 +1,202 @@
...
+
+When using different frequencies, one has to pay close attention at their greates common divisor when expressed as CPU ticks. For example, mixing 100kHz with period 2 and 0.5Hz with period 2 results in underlying interrupt period of 800CPU ticks. But changing to 100kHz+1 will easily result to divisor of 1. This is clearly non-working combination. For the moment best would be to [pwm.print_setup()](#print_setup) and observe how `interruptTimerCPUTicks` and `interruptTimerTicks` change with given input.
+
+## Understanding timer use
+
+This module is using soft-interrupt TIMER1 FRC1 to generate PWM signal. Since its interrupts can be masked, as some part of OS are doing it, it is possible to impact quality of generated PWM impulse. As a general principle, one should not expect high precision signal with this module.
+Also note that interrupt masking is dependent on other activities happening within the ESP besides pwm2 module.
+
+Additionally this timer is used by other modules like pwm, pcm, ws2812 and etc. This requires exclusive lock on the timer in order to ensure support for high end frequencies. Because of that simultenious use of other modules would not be possible.
```
pjsg: spelling mitake in simultaneous

app/modules/pwm2.c
```diff
@@ -0,0 +1,503 @@
...
+}
+
+static uint8_t getCpuTimerTicksDivisor() {
+  return system_get_cpu_freq() == 80 ? PWM2_TMR_MAGIC_80MHZ : PWM2_TMR_MAGIC_160MHZ;
+}
+
+static uint32_t findGCD(uint32_t n1, uint32_t n2) {
+  if (n2 == 0)
+    return n1;
+  return findGCD(n2, n1 % n2);
```
pjsg: Does the compile do tail-end recursion removal? I suspect that the stack frame may be small enough and the depth is probably limited to under 40 (if the inputs are limited to 2^24). Might be safer to code it as a loop.

app/modules/pwm2.c
pjsg: Yes -- your new approach is fine. With a bit of luck, if there is a conflict then something will fail hard!!

docs/modules/pwm2.md
fikin: fixed

app/modules/pwm2.c
fikin: alright, changed to loop.

app/modules/pwm2.c
fikin: ok, take a look at latest commit and lets discuss again if not good enough.

app/modules/pwm2.c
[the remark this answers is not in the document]
fikin: if not of much trouble, i'd say lets stick with current static, for now at least?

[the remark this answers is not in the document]
TerryE: > strictly speaking pwm2.c contains driver and lua module code

This is an accurate observation and also bad practice.  Drivers belong in `app/driver`, and the interface to the Lua environment in `app/modules`.  You would be better organising this with all of the driver related code in `app/driver/pwm2.c`and strip `app/modules/pwm2.c` down to just the wrapper stuff.

Sorry for not raising this point earlier.

HHHartmann:
Just to be sure, this Module is not supposed to replace the PWM Module, right?
Maybe it would be a good idea then to rename it to something like ```PWM-fast``` or some other fitting name to distinguish it from the original PWM Module.
Also a common section to PWM and PWM2 documentation to help the user decide which one to pick would make sense.
Sorry for being so late ...

app/modules/pwm2.c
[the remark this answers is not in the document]
fikin: and i was assuming that single .c file was more neat to deal with ;) 
i guess what i don't appreciate in driver space is the lack of optional compilation, similar to modules space.
but no problem, i'll split the code.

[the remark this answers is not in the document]
fikin: split in driver and lua module.

app/modules/pwm2.c
fikin: i guess now that there is driver is split from the lua module, this should be ok.

docs/modules/pwm2.md
fikin: if you don't object, i'll close this until some more real world feedback is gathered.

fikin:
@HHHartmann : i've added a small section in pwm2 doc about comparison with pwm. let me know how useful you're finding it.

fikin:
@TerryE : apparently i'm not getting how to return values from c to lua. what i am missing in get_timer_data?

TerryE:
> apparently i'm not getting how to return values from c to lua. what i am missing in get_timer_data?

The return value to any C method called from the Lua VM is not a status; it's a return argument count.  Typically this matches the number of responses that your routine has pushed onto the Lua stack, but the VM will ignore / push extra nils if needed.  So `lpwm2_get_timer_data()` should `return 3` and so on.

fikin:
ok, now to module is more or less ready.

marcelstoer:
Nit pick: you're missing newlines at the end of most files. GitHub reports this if you look at https://github.com/nodemcu/nodemcu-firmware/pull/2747/files

fikin:
@marcelstoer : ok, i've added missing new lines.
btw how documentation is tested?

galjonsfigur:
Documentation can be tested using [mkdocs](https://github.com/mkdocs/mkdocs/). You can install it using pip (for example`pip3 install --user mkdocs`) then just use `mkdocs serve` command in project directory. But to show up in module list you will have to add link to pwm2 documentation in [mkdocs.yml](https://github.com/nodemcu/nodemcu-firmware/blob/master/mkdocs.yml#L91) file and it should work.

fikin:
@galjonsfigur : thanks, i've opted using "docker run -i --rm -v $(pwd):/docs -p 8000:8000 melopt/mkdocs serve" ...

marcelstoer:
Documentation was correctly published at https://nodemcu.readthedocs.io/en/latest/modules/pwm2/ and the module is available in the cloud builder (given you select the dev branch).

PR #2752, opened 2019-05-08
luaOTA updated to include object form timers

pienjo:

The `luaOTA` example still used static timers, which are no longer supported. I converted these to dynamic timers. 

Secondly, the example server was missing documentation, making it harder than really necessary to get a proof-of-concept going.

Finally, there were two bugs preventing the system from working correctly in the first place.


TerryE:
Hi @pienjo Matjin.  I've pretty much stopped using luaOTA since I introduced LFS, since I find compiling and downloading LFS  images faster and simpler, and I've got a _lot_ more RAM available.  Nonetheless, I understand that some developers will prefer not moving from a working provisioning system. I am out of country  at the moment, but I'l l take look at this on my return.

Issue #2754, opened 2019-05-11
FIrmware fails to build with #define PMSLEEP_ENABLE


pjsg:
### Expected behavior
The firmware should build
### Actual behavior
It fails building app/modules/node.c

### NodeMCU version
Today's dev branch

The solution is to remove the extra comma......

Even then, I thought that the LROT stuff had to have the keys in alphabetical order -- which is vry much not the case for node.c

TerryE:
> Even then, I thought that the LROT stuff had to have the keys in alphabetical order -- which is vry much not the case for node.c

There is a trade-off here: you need a minimum of maybe 6 entries in your ROTable before a binary chop is on average any faster than a simple linear search -- and the search only takes place now anyway where there is a cache miss -- for example where `__index` meta is used to chain ROTables.  So my current thinking is to make the binary chop optional, and only use it initially on the higher frequency long tables where this is the case -- for example the baselib ROTable.

As far as the comma goes -- Philip Is can do this on my next PR if you want.  Your call.

pjsg:
Do it on your next pr

TerryE:
Philip, thanks for the feedback and will do :smile:

TerryE:
#2757 fixes this.

Issue #2751, opened 2019-05-08
Infinite loop in os_malloc() on out-of-memory causes WDT in SKD 3.0


TerryE:
### Expected behavior

On SK 2.2.1 builds, the following script produces the output shown:
```
> function f() a[#a+1]=string.rep('string'..#a,32) print(#a, node.heap()) end
> a={};for i=1,200 do f() end
1	41640
2	41368
...
127	1656
128	1336
E:M 320
not enough memory
```
That is allocation slowly consumes heap and on exhaustion malloc returns 0 so that Lua GC correctly recovers the RAM.

### Actual behavior on SDK 3.0
```
> function f() a[#a+1]=string.rep('string'..#a,32) print(#a, node.heap()) end
> a={};for i=1,200 do f() end
1	57176
2	56904
3	56608
...
117	19544
118	19216
119	18888

 ets Jan  8 2013,rst cause:4, boot mode:(3,6)

wdt reset
... 
```
That is attempting to allocate the last 18Kb of heap hangs malloc and causes a WDT reset.  The heap() should correctly report available RAM and out-of-memory should be correctly reported by the allocator.

### Test code

as above

### Hardware

Wemos D1 Mini

EgorOA:
Is it my case?

TerryE:
#2743 is definitely related to this and I was meaning to add this link later.  Certainly exploring the reasons for this issue led me to isolate this more specific and yet more widely applicable issue -- hence the separate number

TerryE:
I am a bit embarrassed that I didn't pick this up in my SDK 3.0 testing.  I have to do some testing and binary chop here.  We probably want to force DRAM malloc mode \(`-DMEM_DEFAULT_USE_DRAM`), and I need to consider all of the other random residual malloc use, though this test case is creating the issue with the default Lua allocator.

TerryE:
See espressif/ESP8266_NONOS_SDK#243

TerryE:
I've just done builds with the following SDK versions which report the following free heap at boot on this test harness and also don't exhibit this bug:
-  2.2.1(release): 2.2.1(6ab97e9)  41608 free
-  3.0.0 (latest on branch): SDK 3.0.1-dev(fce080e) 40712 free
-  master (current dev): SDK 3.1.0-dev(3b41fcf) 40664 free

Any thoughts one which SDK version we should bind to `dev``as the controlled 3.0.0. release version isn't usable in my view.

[...]

Issue #2743, opened 2019-05-03
reboots with wdt reason. what does it mean?


EgorOA:
I have a code for a home meteo which worked fine on march dev build (march 10), but since april releases (apr 13 and above) it reboots on wdt when using ucg with ili9341_18x240x320 display. I have changed firmware back and forth several times.

Output on new builds is:
> ets Jan  8 2013,rst cause:4, boot mode:(3,6)
>wdt reset
>load 0x40100000, len 28600, room 16 
>tail 8
>chksum 0x22
>load 0x3ffe8000, len 2592, room 0 
>tail 0
>chksum 0x4a
>load 0x3ffe8a20, len 8, room 8 
>tail 0
>chksum 0x3b
>csum 0x3b

What can this message mean?

I cant reproduce situation with simple example, sorry

devsaurus:
The message means that your Lua code keeps the CPU busy for too long so the watchdog timer resets the esp.

EgorOA:
But this code is working well in previous builds (based on sdk 2.2.1).

TerryE:
SDK 3.0 moved a lot of constant data from RAM to Flash ROM.  The upside was that this move has freed about 18Kb RAM, but one consequence is that we had reported some runtime performance issues with the extra unaligned exception handler overhead within the SDK code. (See espressif/ESP8266_NONOS_SDK#233).  You may be being impacted by this.  See our [FAQ](https://nodemcu.readthedocs.io/en/master/lua-developer-faq/#so-how-does-the-sdk-event-tasking-system-work-in-lua) for more discussion.

EgorOA:
Can wd timer be adjusted to avoid cases like mine? Q

TerryE:
No. It's an SDK function. You can make your individual Lua tasks shorter by breaking up the processing, or issue the `tmr.wdclr()` API function. Read the FAQ.

EgorOA:
Something wrong is happening.
I have a module 'dispDraw.lua' https://yadi.sk/d/fsK1UjiXstDBjw in my project. 
wdt fires in my code 2 times:
> node.compile('dispDraw.lua')
and 
> local d = require('dispDraw')
I have put tmr.wdclr() before each of these instructions. Nothing happend.
In previous builds all goes well.

TerryE:
[...]

PR #2757, opened 2019-05-16
SDK-3.0 tranche updates

### LINKED ISSUE #2729: Build against recent SDK version rather than original release
Follow-up to https://github.com/nodemcu/nodemcu-firmware/issues/2689#issuecomment-484037820

We used to pull in the latest revision of an SDK IIRC. With 3.0 we seem to have switched to using the initially released version.

Unless there are really good reasons that speak against it we should be using the latest revision to profit from any upstream fixes.

> so that the builder can decide whether to select the Release SDK or the latest (as at the Makefile)

I disagree. We should stick to a version we are happy with. That way we can provide a known to be working combination to our users. I don't think it help contribute to our effectiveness if we had to handle issues here that really are SDK issues. Whatever breaks will be reported here.

TerryE:
Fixes #2729, #2743,  #2751 and #2754 

- [ ] The code changes are reflected in the documentation at `docs/*`.  _Not applicable_

### Changes in this PR

-  `Makefile` now supports different SDK versions as per #2729.  The default is the latest SDK 3.0.1-dev (e4434aa) version.
-  `app/Makefile` now includes the `-DMEM_DEFAULT_USE_DRAM` and `app/lua/lauxlib.c` explicitly allocates dram.
-  `app/include/user_config.h` adds IRAM_DATA_ATTR to map static data into iRAM. `ld/nodemcu.ld` adds the `.iram0.data.` psects to iRAM.
-  `app/esp-gdbstub/gdbstub.c`, `app/lua/lrotable.c`, `app/lwip/core/dns.c`, `app/user/user_main.c`, `app/spiffs/spiffs.c` add IRAM_DATA_ATTR to some static data elements.
- `node.c` one line fix for #2754
- `tools/nodemcu-partition.py` fix pathing issue.

We probably need to have some 3.0 SDK overview, but I will do this as a separate PR.


TerryE:
I see that the [Travis build for Windows luac.cross](https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/533552759) failed.  I need to investigate why.

EgorOA:
Sorry, but todays build (built by https://nodemcu-build.com/)
[nodemcu-dev-19-modules-2019-05-17-20-20-16-integer.zip](https://github.com/nodemcu/nodemcu-firmware/files/3193128/nodemcu-dev-19-modules-2019-05-17-20-20-16-integer.zip)
is not running properly. It does cycle reboot with such message
>Formatting file system. Please wait...
>
> ets Jan  8 2013,rst cause:2, boot mode:(3,6)
>
>load 0x40100000, len 17248, room 16 
tail 0
chksum 0x4b
load 0x3ffe8000, len 2592, room 8 
tail 8
chksum 0x92
load 0x3ffe8a20, len 8, room 0 
tail 8
chksum 0xb6
csum 0xb6

and so on...

NicolSpies:
@TerryE, after flashing the image built using Docker with the following feedback:

`SDK ver: 3.0.1-dev(fce080e) compiled @ Apr 16 2019 15:02:05
phy ver: 1145_0, pp ver: 10.2`
`Formatting file system. Please wait...` 
`boot not set`
`ota1 not set`
`ota2 not set`
`rf cal sector: 11`
`freq trace enable 0`
`rf[112] : 00`
`rf[113] : 00`
`rf[114] : 01`
A reboot follows and the cycle repeats.

I have used `#define LUA_FLASH_STORE` to define the size during compile time. Where should I set the parameters indicated as not set ?

TerryE:
@EgorOA  @NicolSpies. Yup, I get the same if I don't use `nodemcu-partition.py` to set up the partition table.  I will track down the issue in the debugger and push a fix today.

NicolSpies:
@TerryE , the max LFS size is 256Kb (`#define LUA_FLASH_STORE 0x40000`), I am already need 128Kb (`#define LUA_FLASH_STORE 0x20000`). What about setting it to max as default and using `#define FLASH_AUTOSIZE` as default. The first format will then take long. After that if required users can use `node.setpartitiontable()` to set the `lfs_size` and `spiffs_size` respectively to requirements. After that formatting and access times will be faster according to the new settings.

TerryE:
> The first format will then take long

The problem with this is that there is no visual feedback of progress or that the formating is underway if you connect to the UART after the initial `Formatting file system. Please wait..`  It would be better if the formatter did a `...` progress bar, writing a dot every 100Kb formatted, say.  Anyway, I am doing a bot of NodeMCU wokr later.  I'll take a look.

TerryE:
We are crashing randomly inside the Espressif SDK 3.0 version of the `flash_erase()` function.  More investigation needed.

TerryE:
I don't think that we should be discussing an open issue on a closed PR, so I have opened this as a separate issue to track it properly.

PR #2761, opened 2019-05-19
Sdk 3.0 tranche3

### LINKED ISSUE #2760: SDK 3.0 crashes during file.format()
@EgorOA @NicolSpies reported in PR https://github.com/nodemcu/nodemcu-firmware/pull/2757#issuecomment-493603965 that the build craters when trying to format the SPIFFS.  

### Test code
Boot image with preformatted SPIFFS and then execute `file.format()`.  I've added some diagnostic code to (temporarily) remove the start-up automatic format and entry /exit bookends around the call to `spi_flash_erase_sector(sector_id)` in the `platform_flash_erase_sector()` wrapper. This shows that the cpu resets during the `spi_flash_erase_sector()` call typically after 10-30 erase calls.

### NodeMCU version
Current `dev`

### Hardware
Wemos D1 Mini

### Discussion
I have reverse compiled the `spi_flash_erase_sector()` in `libmain.a(user_interface.o)` and this does what I would expect:
```C
int spi_flash_erase_sector(int n) {
  if (!protect_flag || !spi_flash_erase_sector_check(n))
    return SPI_FLASH_RESULT_ERR;
  spi_flash_check_wr_protect();
  system_soft_wdt_feed();
  Cache_Read_Disable_2();
  int s = SPIEraseSector(n);
  Cache_Read_Enable_2();
  return s;
} 
```
except that it uses the `libmain.a(spi_flash.o)` `Cache_Read_Disable_2()` and `Cache_Read_Enable_2()` instead of the equivalent BootROM  `Cache_Read_Disable()` and `Cache_Read_Enable()`.  These seem to be more limited in scope in that they apparently disable/enable the Icache without flushing it.

The stochastic nature of the crash strongly suggests to me that we have an ISR firing which includes Icached code. More investigation is needed

TerryE:
Fixes #2760.

The main change here is to force libpp.a into iRAM which fixes the issue.  As discussed in #2760, we've so little iRAM left over that there is no point in complicating the code with  the `IRAM_DATA_ATTR` so this has been backed out.

### DIFF
--- a/app/esp-gdbstub/gdbstub.c
+++ b/app/esp-gdbstub/gdbstub.c
@@ -107,7 +107,7 @@ extern void xthal_set_intenable(int);
 #define OBUFLEN 32
 
 //The asm stub saves the Xtensa registers here when a debugging exception happens.
-struct XTensa_exception_frame_s IRAM_DATA_ATTR gdbstub_savedRegs;
+struct XTensa_exception_frame_s gdbstub_savedRegs;
 #if GDBSTUB_USE_OWN_STACK
 //This is the debugging exception stack.
 int exceptionStack[256];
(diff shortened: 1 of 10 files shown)


EgorOA:
https://github.com/nodemcu/nodemcu-firmware/issues/2760 not fixed for me. I have a continuous reboots  while board tries to format flash after firmware flash :(

PR #2763, opened 2019-05-20
Use dynamic memory for cached DNS names

### LINKED ISSUE #2511: Optimising DNS memory usage
### Expected behavior

The DNS library should make minimal use of `.bss` RAM to avoid exhausting valuable RAM

### Actual behavior

The library uses an allocation system design for high-RAM general purpose and as a result `.bss.dns_table` takes up 1120 for no good reason.  (See `app/mapfile` of any build).

### NodeMCU version

All post LWiP

### Comments

Current build defaults to the following DNS configuration entries

```C
#define DNS_LOCAL_HOSTLIST            0
#define DNS_LOCAL_HOSTLIST_IS_DYNAMIC 0
#define DNS_TABLE_SIZE                4
#define DNS_MAX_NAME_LENGTH           256
#define DNS_MAX_RETRIES               4
#define DNS_MAX_TTL                   604800
#define DNS_MSG_SIZE                  512
#define DNS_SERVER_PORT               53
#define DNS_TMR_INTERVAL              1000
#define DNS_DEBUG                     LWIP_DBG_OFF
#define DNS_DOES_NAME_CHECK           1
```

We should define `DNS_LOCAL_HOSTLIST_STORAGE_PRE` to `static const` so that if a static host list is defined then it goes into flash rather than RAM.  We should also expose the following DNS defines to `user_config.h` to allow developers the option to overide these and to define their own static hosts enties:  
```C
/*
 * Use the following setting to include a static hosts list in your build.
 */ 
//#define DNS_LOCAL_HOSTLIST 1
//#define DNS_LOCAL_HOSTLIST_INIT = {{"entry1",0x123}, {"entry2",0x456}}
```
  
DNS_TABLE_SIZE is the size of the local DNS cache.  This cache allows you to resolve the same hosts multiple times without going back to the DNS server.  Resolving more hosts than this will result in some repeat resolutions going back to DNS.  So if you only refer to 2 external hosts or you cache your resolved IPs as Lua variables in your application, then set this to 2 as your application won't really make use of this cache.
```C
/*
 * By default 4 DNS entries are cached; this can be varied by setting the following
 */
//#define DNS_TABLE_SIZE                4
```

The current DNS use a fixed hostname length of `DNS_MAX_NAME_LENGTH` (256) so the cache allocates 4×256 bytes in  valuable `.bss` for host names.  Given that DNS resolution is very low frequency operation then it makes sense to replace the fixed `char name[DNS_MAX_NAME_LENGTH]` in `struct dns_table_entry` by `char *name;` and updating the assignment at `dns.c:930,933` 

```C
  pEntry->arg   = callback_arg;
  namelen = LWIP_MIN(os_strlen(name), DNS_MAX_NAME_LENGTH-1);
  MEMCPY(pEntry->name, name, namelen);
  pEntry->name[namelen] = 0;
```
by
```
  pEntry->arg   = NULL;
  namelen = os_strlen(name);
  if (namelen > DNS_MAX_NAME_LENGTH-1)
    return ERR_VAL;
  char *p = os_realloc(pEntry->name, namelen+1);
  if (!p)
    return ERR_MEM;
  else
    pEntry->name = p;
  os_strcpy(pEntry->name, name);
  pEntry->arg   = callback_arg;
```
and this will allocate string storage JiT, and use a lot less than the current preallocated 1K space.

If there is general support then I will raise the PR.

TerryE:
Fixes #2511

- [ ] ~The code changes are reflected in the documentation at `docs/*`.~  N/A

Adds about 1K RAM to the free heap.

### DIFF
--- a/app/lwip/core/dns.c
+++ b/app/lwip/core/dns.c
@@ -177,7 +177,7 @@ struct dns_table_entry {
   u8_t  seqno;
   u8_t  err;
   u32_t ttl;
-  char name[DNS_MAX_NAME_LENGTH];
+  char *name;
   ip_addr_t ipaddr;
   /* pointer to callback on DNS query done */
   dns_found_callback found;
@@ -929,6 +929,9 @@ dns_enqueue(const char *name, dns_found_callback found, void *callback_arg)
   pEntry->found = found;
   pEntry->arg   = callback_arg;
   namelen = LWIP_MIN(os_strlen(name), DNS_MAX_NAME_LENGTH-1);
+  if ((pEntry->name = (char *) mem_realloc(pEntry->name, namelen+1)) == NULL) {
+    return ERR_MEM;
+  }
   MEMCPY(pEntry->name, name, namelen);
   pEntry->name[namelen] = 0;
 
--- a/app/lwip/core/dns.c
+++ b/app/lwip/core/dns.c
@@ -924,14 +924,14 @@ dns_enqueue(const char *name, dns_found_callback found, void *callback_arg)
   LWIP_DEBUGF(DNS_DEBUG, ("dns_enqueue: \"%s\": use DNS entry %"U16_F"\n", name, (u16_t)(i)));
 
   /* fill the entry */
-  pEntry->state = DNS_STATE_NEW;
-  pEntry->seqno = dns_seqno++;
-  pEntry->found = found;
-  pEntry->arg   = callback_arg;
   namelen = LWIP_MIN(os_strlen(name), DNS_MAX_NAME_LENGTH-1);
   if ((pEntry->name = (char *) mem_realloc(pEntry->name, namelen+1)) == NULL) {
     return ERR_MEM;
   }
+  pEntry->state = DNS_STATE_NEW;
+  pEntry->seqno = dns_seqno++;
+  pEntry->found = found;
+  pEntry->arg   = callback_arg;
   MEMCPY(pEntry->name, name, namelen);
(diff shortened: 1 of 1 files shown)


app/lwip/core/dns.c
```diff
@@ -929,6 +929,9 @@ dns_enqueue(const char *name, dns_found_callback found, void *callback_arg)
   pEntry->found = found;
   pEntry->arg   = callback_arg;
   namelen = LWIP_MIN(os_strlen(name), DNS_MAX_NAME_LENGTH-1);
+  if ((pEntry->name = (char *) mem_realloc(pEntry->name, namelen+1)) == NULL) {
+    return ERR_MEM;
```
nwf: Perhaps worth cleaning out the rest of `pEntry` on this path (e.g., setting `found` and `callback_arg` to `NULL`)?

TerryE:
This is a bit of low hanging fruit that I identified a few months back.  I am doing it now to mitigate the memory impact of SDK 2.2.1 -> 3.0 migration.    Note that the DNS has four cache slots which previously had 4×256 bytes reserved for host names.  Now 4 byte pointers are used and the space for the name is allocated on a just-in-time basis.  So if you don't use DNS lookup then you have an extra 1,008 bytes of heap; if you just have one DNS lookup, then you loose the actual amount of heap needed to allocate this string say 20 bytes or so, and so on.

@nwf, I've reordered realloc so that the entry isn't left in an inconsistent state.

PR #2768, opened 2019-05-25
new benchmark module

fikin:

this module offers methods which are timing various esp operations. there are timing methods from other people and as well as some new ones. right now it has methods to cover cpu and memory use, some os functions, some gpio operations and even methods to test timer1 performance in all possible combinations.

the module is not an operations module per se but rather something of testing and learning nature. one would use it mainly when wants to evaluate performance aspects of esp. being compiled as lua module makes its use rather easy and convenient.

note: right now the code is based on pwm2 branch as i'm timing exclusive and shared use of timer1 (exclusive api comes with pwm2 commits).


marcelstoer:
You need to rebase this on `dev` once #2747 (pwm2) is merged to get a clean commit history.

marcelstoer:
I don't know what you based your implementation on but the commit history is (still) a jungle. I tried to rebase it on the current `dev` but I get merge conflicts all over the place. I suggest you extract the relevant changes, run `git reset --hard dev`, then re-apply them, and then force push this branch. The PR will update here accordingly.

fikin:
sorry for that hustle. now i think it should be ok.

docs/modules/benchmark.md
```diff
@@ -0,0 +1,565 @@
+# PWM Module
```
marcelstoer: Copy/paste mishap.

fikin: spot on, fixed now.

devsaurus:
Build fails now since #2841 was merged:
```
In file included from benchmark.c:9:0:
../../sdk-overrides/include/c_types.h:5:2: error: #error "Please do not use c_types.h, use <stdint.h> instead"
 #error "Please do not use c_types.h, use <stdint.h> instead"
```

Please rebase onto current `dev` and adapt.

TerryE:
@fikin Nikolay in principle this a great idea, and I personally thank you for doing this work.  Even so, our process is to raise an issue to discuss, scope and agree a broad implementation strategy, the whats and hows before and in parallel with raising the PR. 

I missed this when first raised -- my apologies -- I'll take a look over the weekend and then give you my review comments.

fikin:
@TerryE : no problem to reshape it on my side. This module was simply an attempt to salvage some otherwise tooling code.

fikin:
@devsaurus : right now I'm traveling abroad. First feasible timeslot to look into it is next Tuesday.

devsaurus:
No problem.

fikin:
now import errors are fixed.

fikin:
@TerryE : would you find some time to post some feedback about this module?

fikin:
> what should we benchmark

my working assumption for the moment is that benchmarking is serving module developer rather than casual lua-user. hence this module is mainly platform functionality.
this is also reflected on the request notes.

right now if i have to project the future, i'd imagine this module as dev-branch only available, where people needing cross reference or leaning can compile and use it. and if new functionality comes along extend it.

but now that you mentioned about larger scope lua-space testing, i'm kind of intrigued. 
personally i don't have any idea popping out of my mind, but if have some pls do share it?

> current platform tests

you're right that most of these tests are not truly informative for experienced developers, except interrupts perhaps (nodemcu platform seems to be changing enough to want to track this behavior). but by having them provided, a reference is established to another published benchmark. this way comparison and outcomes become much more reliable to interpret. at least that was my thinking when opted for adding them.

> about xtensa

sure, i'm learning it as it goes ;) and thanks for the help. i'll spend some time in next days to review how memory hints are being used (i largely copied the code rather than analyse it) and then open up some more on it.

TerryE:
Nikolay, you'll find my email addr in commits on the git log.  If you wish then ping me and we can set an email channel.  Here are a few useful references that are quite hard to find: 
-  [the Xtensa ISA](https://0x04.net/~mwk/doc/xtensa.pdf)
-  [A short guide to Xtensa assembly language](http://cholla.mmto.org/esp8266/xtensa.html)
-  [A No Frills Introduction to Lua 5.1 VM Instructions](A No Frills Introduction to Lua 5.1 VM Instructions)

Even if we are targeting the C module developer, then the benchmarks as proposed are misleading because of cached vs non-cached issues.  Espressif don't give details of the cache design and Cadence (like ARM) provide a range of selection options for their licensed SoC manufacturers (see the ISA), so for example we don't even know whether it is direct mapped or two-way associative, but certainly IROM1 cache-miss execution is at least an order slower than cache-hit or ROM or IRAM1 execution.

Hence when the instruction timing is absolutely essential, then you really need to run the code out of IRAM1, but this is such a scarce resource that we should only put _absolutely_ essential code here.  For example, I2C master is sloppy cock tolerant so its drivers don't need to run in IRAM1; caching works well enough in practice.

When it comes to C module development, one of the trade-offs that you have to make is when to do stuff in C and when to use Lua.  Changing C code requires use of a complete xtensa toolchain and this is a major hurdle  for most developers.  Lua code it easily tractable by Lua developers so you will see a trend to only coding functionality in C if there is a compelling reason to.  Without an understanding of comparative timing, it is difficult for developers to amke this trade-off.

fikin:
@TerryE thanks for the links above

i've been sitting on this pull request for some time now, without clear idea where to take it to from here. basically i can't find a good enough way to make it general-purpose beneficial.

so my thinking is to close it and keep its codebase as dedicated branch in my repo. 
if someone needs ready infrastructure to add/do some timing c-tests, he can pull it on top of dev-branch from there.

PR #2547, opened 2018-11-07
MPU6050 driver

hiperiondev:
Fixes #\<GitHub-issue-number\>.

Make sure all boxes are checked (add x inside the brackets) when you submit your contribution, remove this sentence before doing so.

- [x] This PR is for the `dev` branch rather than for `master`.
- [ ] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [ ] I have thoroughly tested my contribution.
- [ ] The code changes are reflected in the documentation at `docs/en/*`.

\<Description of and rationale behind this PR\>
This contribution add MPU6050 driver. Is not fully tested (only mpu functions)
Need more tests in dmp functions and documentation

PR #2769, opened 2019-05-26
[wip] add mpu6050 sensor

feilongfl:


nwf:
Given that this module appears to be only I2C and needs no sophisticated timing handling &c, it may be better off as a Lua module than a C one.

feilongfl:
@nwf 
This sensor requires very frequent read and write operations, and requires a lot of calculations when parsing data(todo). I think that making a c language module will have better performance.

HHHartmann:
Just for reference: there was also #2547

devsaurus:
Build fails now since #2838 was merged:
```
CC app/modules/mpu6050.c
mpu6050.c:10:22: fatal error: c_stdlib.h: No such file or directory
 #include "c_stdlib.h"
                      ^
compilation terminated.
```

You need to rebase onto current `dev` and adapt.

vsky279:
I've missed this PR. I have a Lua version of the MPU9250 driver - see https://github.com/vsky279/nodemcu-firmware/commit/232f61d38463728ed36d94ab20fbd8ec17b29e0d
MPU9250 is actually MPU6050 + AK8975 magnetometer + BMP280.

But I could not make it work reasonably so I dit not file any PR. It was not working even on Arduino with existing libraries.

@feilongfl if you can try this Lua driver?

You are right that there are quite complex calculations to be performed. Maybe C library just handling complex calculations would be the right way to implement this. I wanted to do it in a second step. But I froze on the first step :-).

We've taken similar approach for the new BME280 driver (https://nodemcu.readthedocs.io/en/dev/lua-modules/bme280/) where calculations are in the C module `bme280_math` (https://nodemcu.readthedocs.io/en/dev/modules/bme280_math/).

PR #2774, opened 2019-05-30
add serial method to table

feilongfl:

I use lua function like this.
```
function ByteTableToString(t)
    local tt = {}
    for _,v in pairs(t) do 
        tt[#tt + 1] = string.char(v)
    end
    return table.concat(tt)
end
```
but it's too slow.
so I make a c one.
eg:
```
table.serial({0x41,0x42,0x43,0x44,"Hello",0x45})
ABCDHelloE
```

### DIFF
--- a/app/lua/lauxlib.c
+++ b/app/lua/lauxlib.c
@@ -662,6 +662,22 @@ LUALIB_API void luaL_addvalue (luaL_Buffer *B) {
   }
 }
 
+LUALIB_API void luaL_addrawvalue (luaL_Buffer *B) {
+  lua_State *L = B->L;
+  unsigned char bytes = lua_tonumber(L, -1);
+
+  if (1 <= bufffree(B)) {  /* fit into buffer? */
+    *(B->p) = bytes & 0xff;
+    B->p ++;
+    lua_pop(L, 1);  /* remove from stack */
+  }
+  else {
+    if (emptybuffer(B))
+      lua_insert(L, -2);  /* put buffer before new value */
+    B->lvl++;  /* add new value into B stack */
+    adjuststack(B);
+  }
+}
 
 LUALIB_API void luaL_buffinit (lua_State *L, luaL_Buffer *B) {
   B->L = L;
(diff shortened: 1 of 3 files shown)


feilongfl:
Because I copied it according to the concat method, this problem will occur.

https://github.com/nodemcu/nodemcu-firmware/issues/2775

nwf:
`table` is a Lua-language library and I'm loathe to add custom nodemcu elements to it.

Try `function ByteTableToString(t) return string.char(table.unpack(t)) end` instead?  (Though it will not work on your example involving `"Hello"`, but I don't think a function named `ByteTableToString` should accept string elements in the table.)

feilongfl:
@nwf 
Thank you very much for your solution. I have very little experience with the lua language, so I searched the lua language byte table to convert the string method, but the method on the stackoverflow performance is very poor. 
Under test, your method is similar to this pr method execution time, so I close this pr.

feilongfl:
sorry for my careless,
I had an error in the project while testing, which caused the download to be unsuccessful.

I found that `table` has no member called `unpack`

```
for k,v in pairs(table) do print(k,v) end
concat	lightfunction: 0x4024debc
foreach	lightfunction: 0x4024e1ac
foreachi	lightfunction: 0x4024e130
getn	lightfunction: 0x4024e104
maxn	lightfunction: 0x4024e084
insert	lightfunction: 0x4024dffc
remove	lightfunction: 0x4024df6c
setn	lightfunction: 0x4024dce4
serial	lightfunction: 0x4024ddb4
sort	lightfunction: 0x4024e458
```

galjonsfigur:
Try `unpack` instead of `table.unpack` - This function has been moved to `table` in Lua 5.2 if I remember correctly. NodeMCU is still using modified Lua 5.1 so this is why the code is not working.

feilongfl:
my table size is 100,and only change bytetabletostring function,I got single

> Try `unpack` instead of `table.unpack` - This function has been moved to `table` in Lua 5.2 if I remember correctly. NodeMCU is still using modified Lua 5.1 so this is why the code is not working.

thanks

Issue #2772, opened 2019-05-28
the lua module httpservr.lua said not enough memory 


hanfengcan:
### Test code
I use the  this [code](https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_modules/http/http-example.lua)(from the lua_modules) to test the httpserver. I found the heap decrease when fresh the browser many time until reboot

I tried to read the code to fix this problem, but failed. the `collectgarbage` seem invalid.

BTW, how do you debug lua in the nodemcu? do you have a good way like jtag step a step ?

### NodeMCU version
last release build by docker 

### Actual behavior
+R	GET	/	22464
+H	host	192.168.0.101
+H	connection	keep-alive
+H	cache-control	max-age=0
+H	upgrade-insecure-requests	1
+H	user-agent	Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36
+H	accept	text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3
+H	accept-encoding	gzip, deflate
+H	accept-language	zh-CN,zh;q=0.9
+B	0	22536
**+B	nil	22696**

+R	GET	/favicon.ico	21888
+H	host	192.168.0.101
+H	connection	keep-alive
+H	pragma	no-cache
+H	cache-control	no-cache
+H	user-agent	Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36
+H	accept	image/webp,image/apng,image/*,*/*;q=0.8
+H	referer	http://192.168.0.101/
+H	accept-encoding	gzip, deflate
+H	accept-language	zh-CN,zh;q=0.9
+B	0	21912
**+B	nil	22056**

+R	GET	/	17808
+H	host	192.168.0.101
+H	connection	keep-alive
+H	cache-control	max-age=0
+H	upgrade-insecure-requests	1
+H	user-agent	Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36
[...]

PR #2785, opened 2019-06-08
add a new httpserver module

hanfengcan:

I use the lua module `httpserver` to do a project. but this module seem memory leak. I try to fix, but failed. 

so, I write this new module. it is a similar module to [http/httpserver](../../lua_modules/http/httpserver.lua), but more stable memory consumption.

In fact, I copy some code from old module to new module.

### DIFF
--- a/docs/lua-modules/httpserver2.md
+++ b/docs/lua-modules/httpserver2.md
@@ -1,0 +1,104 @@
+# HTTP Server Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2015-01-19 | [hanfengcan](https://github.com/hanfengcan) | [hanfengcan](https://github.com/hanfengcan) | [httpserver2.lua](../../lua_modules/http2/httpserver2.lua) |
+
+A similar module to [http/httpserver](../../lua_modules/http/httpserver.lua), more stable memory consumption
+
+
+!!! note
+	Copy this module and upload to your nodemcu by yourself.  
+
+### Require
+```lua
+httpserver = require("httpserver2")
+```
+
+### Release
+```lua
+httpserver = nil
+package.loaded["httpserver"] = nil
+```
+
+### httpserver.createServer()
+Function to start HTTP server.
+
+#### Syntax
+`httpserver.createServer(port, handler(req, res))`
+
+#### Parameters
+- `port`: Port number for HTTP server. Most HTTP servers listen at port 80.
+- `handler`: callback function for when HTTP request was made. It has 2 arguments: `req` (request) and `res` (response).  
+	The first object holds values:
+
+	- `method`: Request method that was used (e.g.`POST` or `GET`).
+	- `url`: Requested URL.
+	- `onheader`: value to setup handler function for HTTP headers like `content-type`. Handler function has 3 parameters:
+
+		- `self`: `req` object.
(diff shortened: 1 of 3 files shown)


nwf:
Please don't.  Rewriting from scratch is fun, but not a great approach.  If there's an issue with the existing `httpserver`, just fix it in place.

That aside, this is clearly derivative from the existing `httpserver` and so should not strip the existing copyright and license notice.

The use of `xpcall` in `send_header` seems incredibly dubious; this code is squashing any possibility of error handling in favor of printing to the console.

Can you articulate why `cfini` needs to explicitly `nil` out `req` and `res`?  I didn't think we'd be holding on to those in closure after the existing `cfini` was done.

Also note that `http2` is quite confusing, given the existence of https://en.wikipedia.org/wiki/HTTP/2

hanfengcan:
sorry, I am not add the license notice.

the asset will make the nodemcu reboot, so I use the xpcall

maybe we are able to fix the memory leak of the httpserver together and then I cancel this PR

btw, the name of httpserver2 just to describe it fork from httpserver2😊

hanfengcan:
Today, I try to fix the memory leak again. I almost modified all over the code to make it look like the new module, include closure. 

 But, no use.

nwf:
To respond only to "the asset will make the nodemcu reboot, so I use the xpcall": the *point* of an assertion is to throw an error, which can be caught by either wrapping code or by the top-level handler, which reboots.  If you don't want that behavior, don't use an assert!

Please attempt to characterize the memory leak: which object is being retained, and how, before mutating the code in an effort to fix it.

hanfengcan:
thanks

I just check the type of header and header.headers. hint like wifi.sta.config `config table not found` and stop the program while the type are wrong

About the memory leak, there are some information in #2772. the memory decrease until `not enough memory`

Possible I can try to modified the new module to make it look like the old module.  maybe to find the problem and fix it

marcelstoer:
I'm definitely against maintaining another HTTP Server module ourselves. If it's got issues (memory or otherwise) then those should be fixed.

However, instead we might even decided to remove the current module and tell users to use https://github.com/marcoskirsch/nodemcu-httpserver instead. I never used it, I didn't analyse its code but it has been around for a long time and seems quite popular (and @HHHartmann contributed to it).

hanfengcan:
I know. let me try to fix it

HHHartmann:
The module is quite good. Unfortunately it is not maintained anymore for no apparent reason.

mjg14:
@HHHartmann -- which module? This one, or the Marcos Kirsch one? In general: If I wanted to serve a single, simple page (with variable data on it), and react to a few buttons or links on that page by making the nodeMCU do things, which one would you recommend?

HHHartmann:
The Marcos Kirsch one is a complete project including network configuration and uploader. Recently it also got adapted to work with LFS. You only need to add your one lua file which generates the desired webpage.
It has problems however if your (POST) request has a largish body as it first receives the while body and then calls your Lua code. However return ing largish HTML pages is not limited, as it uses coroutines and can send the page as you build it.
Here or in httpserver.lua as a matter of fact it is vice versa (if i understood the code correctly). The body of a request can be as large as needed but the reply is limited by memory size.
Also you will need to add some code for startup/network config and to match the web URL to the Lua file to execute.
The Marcos Kirsch variant can also do authentication for you. It has lots of examples which will get you started quickly.

Issue #2783, opened 2019-06-04
RFC - Emergency GC setting for Lua 5.3


TerryE:
I am currently doing my Lua 5.3 port and there are a number of issues on which I would like feedback from the contributors.  The first relates to garbage collection.  Our current Lua 5.1 VM includes the `eLua` Emergency GC, plus the various EGC tuning parameters that seem to be rarely used and don't seem to be stable.  

The default setting EGC (which most users use) is `node.egc.ALWAYS` and this triggers a full GC before every memory allocation.  I find that VM spends maybe 90% of its time doing full GC sweeps with this default setting!  

My 'production' apps were initially developed pre-LFS, so I rarely bump against memory exhaustion.  The setting that I now use in  `node.egc.setmode()` is `ON_MEM_LIMIT=-8096` (note the negative value) which defaults to the standard Lua incremental GC unless there is less than 8Kb heap left; under this threshold it kicks into a full GC before every memory allocation.  This setting drops the time in the GC by maybe 10× so overall the Lua code runs maybe 5× faster.

The issue is that many of the modules including `net` and `LwIP` don't use the Lua allocator so you need headroom in the heap to allow these to operate as these mallocs will not trigger a GC if needed.  As a consequence using `ON_ALLOC_FAILURE` will invariably cause your application to crater.  Likewise there are bugs in how the GC counts memory, so relying on ON_MEM_LIMIT with a positive limit will also tend to crash your application.

(As an aside, the ECG uses multiples of 1Kb internally for its parameters but the `node` methods work in exact bytes which are then rounded to multiple Kb internally.)

**_So now what do we do for Lua 5.3?_**  This now already includes an EGC mode, but this defaults to `ON_ALLOC_FAILURE` without the `ALWAYS` and `ON_MEM_LIMIT` options.

-  I see little point of re-implementing the options which crater the app, but a headroom-based trigger does seem sensible.  

-  I propose we have a _backwards compatibility break_ on the `node.egc` options (I don't see this as a material issue as most developers don't go here and the ones that do are power users)  and have `node.egc.setheap(n)` where `n` is the Kb of headroom required for other mallocs, above which the EGC kicks in.  This would be synonymous to the current `node.egc.setmode( node.egc.ON_MEM_LIMIT, -1024*n )` with a default of `n=12`, and setting the limit to `nil` would cause it to behave like the current default `ALWAYS` option.

[...]

PR #2790, opened 2019-06-11
Add luacheck config and configuration for Travis CI

### LINKED ISSUE #772: Polish the Lua examples
Go through most of the examples an refresh them using the API and decent coding style.

@TerryE you once stated you had this on your TODO list. Feel free to remove the assignee.

Also, I don't see why we should have both `/examples` and `/lua_examples`.

galjonsfigur:

This PR is 1st PR to fix #772. 

It adds configuration for a tool called [luacheck](https://github.com/mpeterv/luacheck) to help with checking NodeMCU Lua scripts for various issues (more on GitHub project page). I also added script for Travis CI that installs `luacheck` and checks all files in `lua_examples` folder. For now it's commented but after fixing all the existing examples it can be turned on and from there it would be much easier to maintain quality of those examples. There is also a note in `CONTRIBUTING.md` file on how to use `luacheck` tool in local environment.


tools/travis/run-luacheck.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+LUA_FILES_TO_CHECK=`find lua_examples -iname "*.lua"`
```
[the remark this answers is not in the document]
galjonsfigur: Added - but it will take more time to fix all the warnings to be able to enable this feature. But now everyone can use `luacheck` and I now in summer have more free time to fix them one by one :smile:

tools/travis/localua.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+
+# Downloads and installs a self-contained Lua and LuaRocks.
+# Supports Linux, macOS and MSYS2.
+# Copyright (c) 2015-2019 Pierre Chapuis, MIT Licensed.
+# Latest stable version available at: https://loadk.com/localua.sh
+# Maintained at: https://github.com/oploadk/localua
+
+DEFAULT_LUA_V="5.3.5"
```
marcelstoer: We're actually still using 5.1.4 but 5.3 is coming (#2783).

galjonsfigur: This Lua version is only to run `luarocks`+ `luacheck` and in `luacheck_config` [here](https://github.com/nodemcu/nodemcu-firmware/pull/2790/files#diff-e9472bcdeadb819510dd8e9e8ae1869cR883) code to check is set to support Lua 5.1 (change for 5.3 will only require change `51` to `53`). I used 5.3 version because I'm not 100% sure if `luacheck` will work properly on older (according to docs it should) but that setting only matters for Travis. AFAIK most of linux environments allow to install multiple Lua versions but defaults to 5.3.

tools/travis/run-luacheck.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+LUA_FILES_TO_CHECK=`find lua_modules lua_examples -iname "*.lua"`
```
HHHartmann: This will fail if file names contain spaces or other funny characters. But then who needs spaces in filenames.

galjonsfigur:
Many thanks for feedback!

> @galjonsfigur maybe it would be good to also add the script to create the configuration file if you like. If it runs reliably the configuration could even be created on the fly.

My script is not robust enough because LROT table names often don't correspond with module names and to extract it properly it would require much more effort to recognize functions available for created objects, modules with own submodules like `wifi` or things like `ucg` or `u8g2` where the all global names would have to be acquired from their source files. (for now globals from those 2 modules are not checked). But it should be helpful when adding a new module or adding new functions to existing ones.

I also modified bash script and it should be good now. I tested it in my [testing repository](https://github.com/galjonsfigur/travis-lua-nodemcu-tests) and everything looks fine.

I think it would be a wise option to also change the script for `luac_cross` in [travis.yml](https://github.com/nodemcu/nodemcu-firmware/blob/dev/.travis.yml#L33) the same way so it will work with names with spaces.

CONTRIBUTING.md
```diff
@@ -26,6 +26,13 @@ Use the platform and tools you feel most comfortable with. There are no constrai
 - [Full-fledged Linux environment](http://www.esp8266.com/wiki/doku.php?id=toolchain#how_to_setup_a_vm_to_host_your_toolchain), either physical or virtual.
 - [Docker image](https://hub.docker.com/r/marcelstoer/nodemcu-build/) which allows running the build inside the container as if you were running a build script on your local machine.
 
+## Writing Lua Code
+A great resource about writing Lua for NodeMCU can be found in [Lua Developer FAQ](https://nodemcu.readthedocs.io/en/latest/lua-developer-faq/) - make sure to read it! When you're writing your Lua code and it's not working as it should you can test it with `luacheck` tool that can help you find various types of bugs. To install it you have to install [luarocks](https://luarocks.org/) and use command `sudo luarocks install luacheck` to install the tool. Now you're ready to go! By using this command (assuming you're in `nodemcu-firmware` directory):
+
+`luacheck --config luacheck_config.lua <your file to check>`
```
HHHartmann: this needs to be adapted to the new location of ```luacheck_config.lua```

CONTRIBUTING.md
```diff
@@ -26,6 +26,13 @@ Use the platform and tools you feel most comfortable with. There are no constrai
...
+
+you can look for bugs and problems with the code!
```
HHHartmann: "within the code"

HHHartmann:
And please feel free to also implement the xargs variant in travis.yml.
This PR will definitely bring the project forward one more step as we now prefer lua modules over c modules.

tools/travis/run-luacheck.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+echo "Installing Lua 5.3, LuaRocks and Luacheck"
+(
```
HHHartmann: Travis persists the cache folder over builds, so Installing just once would be enough. Maybe give it a version number to be able to update to a newer version if needed.
I am not sure if this has to go in this PR, so only a comment.

HHHartmann:
> Can't really say anything about localua but seems to be sane.
> Will give it a spin These days.

I cloned @galjonsfigur s branch and started ```tools/travis/run-luacheck.sh``` which failed compiling.

```
gcc -std=gnu99 -O2 -Wall -Wextra -DLUA_COMPAT_5_2 -DLUA_USE_LINUX    -c -o lua.o lua.c
lua.c:82:31: fatal error: readline/readline.h: No such file or directory
 #include <readline/readline.h>
                               ^
```
Could you have a look?

galjonsfigur:
@HHHartmann Thanks for more feedback - now it should be good.

 `localua` builds Lua from source, so it requires additional packages. If I remember correctly they are: `libreadline-dev`(for Lua itself), `unzip`(for `tar`), and `build-essential`(for make and compiler). I checked and Travis already has those packages installed, so I didn't bothered with adding those to package list.

For now `localua` don't have ability to cache build, but I will try to fix it someday and send PR to the author of the script. Lua builds quite fast (~10 seconds on Travis) so this will probably another PR after polishing the examples and modules.

.travis.yml
```diff
@@ -30,6 +30,8 @@ script:
 - if [ "$OS" = "windows" ]; then bash "$TRAVIS_BUILD_DIR"/tools/travis/ci-build-windows-ms.sh; fi
 - if [ "$OS" = "linux" -a "$TRAVIS_PULL_REQUEST" != "false" ]; then bash "$TRAVIS_BUILD_DIR"/tools/travis/pr-build.sh; fi
 - cd "$TRAVIS_BUILD_DIR"
-- LUA_FILES=`find lua_modules lua_examples -iname "*.lua"`
-- echo checking $LUA_FILES
-- $LUACC -p $LUA_FILES
+- echo "checking:"
+- find lua_modules lua_examples -iname "*.lua" -print0 | xargs -0 echo
+- find lua_modules lua_examples -iname "*.lua" -print0 | xargs -0 $LUACC -p
+#- if [ "$OS" = "linux" ]; then bash "$TRAVIS_BUILD_DIR"/tools/travis/run-luacheck.sh; fi
```
HHHartmann: In my opinion it would make sense to change this line to
```- if [ "$OS" = "linux" ]; then bash "$TRAVIS_BUILD_DIR"/tools/travis/run-luacheck.sh || true ; fi```
for a start until all lua files are fixed. That way everybody can see the progress of fixing and what can be done.
It also makes sure that breaking the script can be detected early.

galjonsfigur: This is a good idea, but I don't know if Travis will handle this many warnings:
`Total: 1177 warnings / 2 errors in 67 files`
I will try and revert this in case it fails.

HHHartmann:
It does. I tried

.travis.yml
galjonsfigur: @HHHartmann 
>It does. I tried

Indeed! Build went fine and the warning list is visible in build log. Thanks again for reviewing this PR!

marcelstoer:
@galjonsfigur good job! Thanks for hanging in there and reacting so quickly to feedback.

TerryE:
@marcelstoer, this is great in principle but is proving to be a total **PITA** in practice.  Why? because it runs every travis check, but this decision was made without any contributor committing to do the actual clean-up.  As result for example when my latest PR failed its Travis check [here](https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/562059763) with 4258 lines listed of which the last 40% are the install of luacheck and time and again repeated error report.  This makes it very difficult for me to try and see _why_ the Travis build failed.

Can we please disable this _as a default_ for now until we can find a resource with the time available to clean up the Lua scripts.  Especially as anyone can run this interactively if they are working on scripts, as I have started to do with mine.

galjonsfigur:
@TerryE 
>Can we please disable this as a default for now 

Sure - I have no problem with this as I'm still testing the examples and there are more important things to be done like all the C cleanup and Lua 5.3 stuff. It would be OK if @marcelstoer or You could push a commit to comment out [this line](https://github.com/nodemcu/nodemcu-firmware/blob/dev/.travis.yml#L36).

TerryE:
Sure I will do this one my next PR if @marcelstoer doesn't mind that shortcut.

The `run-luacheck.sh` is unusable anywhere except for in a virgin Travis container, since it assumes that lua and luacheck aren't already installed on the target.  In my case I just did a `sudo apt-get install` to get them, so I just do a `luacheck $(find wherever -name \*.lua)` at the command line.

marcelstoer:
I'm not completely against disabling it for the CI build but

- "this decision was made without any contributor committing to do the actual clean-up" - true, but @galjonsfigur did IIRC.
- when Travis builds fail all I usually have to do is open the build log and search for "exited with " [1 or 2] to find the culprit
- if you do then please do so in a dedicate commit straight to `dev` rather than burying it in a larger, likely unrelated PR.

TerryE:
@marcelstoer I was seeking your blessing, not criticising. I'll do the separate PR as you suggest.

galjonsfigur:
@marcelstoer 
> true, but @galjonsfigur did IIRC.

Yes, the results are [here](https://github.com/galjonsfigur/nodemcu-firmware/tree/dev-lua-polish) but there are many things to test. I create PR soon when I will be done with them. Those are only little fixes to make `luacheck` happy and there is still lots of work and ideas to be tested and discussed, so for now I think it can be disabled if it slows down Travis and development down.

marcelstoer:
>  I was seeking your blessing

You know you don't need that 😜 but thanks for asking anyhow.

PR #2791, opened 2019-06-14
fixed the httpserver memory leak

### LINKED ISSUE #2772: the lua module httpservr.lua said not enough memory

hanfengcan:
Fixes #2772 .

finally, I fixed the httpserver module memory leak.

It's strange that defined the function variable must be after other type of variable. And the res and req must be set nil


lua_modules/http/httpserver.lua
```diff
@@ -23,54 +23,49 @@ do
...
+      end
     end
-    if data then
-      -- NB: no headers allowed after response body started
-      if self.send_header then
-        self.send_header = nil
-        -- end response headers
+    local send_header = function(header)
+      if type(header) ~= "table" then
+        print("httpserver.lua: header need a table! eg: {status = 200, headers = {Connection = \"close\"}}")
```
nwf: Please throw an error here.  This is not the kind of mistake that should just merit printing.

If you are unfamiliar with the "kinds" of errors I mean, consider the taxonomy given in http://www.cs.cmu.edu/~410/lectures/L08a_Errors.pdf

lua_modules/http/httpserver.lua
```diff
@@ -79,17 +79,19 @@ do
...
+        csend(function() conn:on("sent", nil) conn:close() res = nil req = nil end)
       end
-      local req, res
-      local buf = ""
-      local method, url
       local ondisconnect = function(conn)
-	conn.on("sent", nil)
+        conn.on("sent", nil)
+        res = nil
+        req = nil
```
nwf: 1. Does `ondisconnect` really need to take the `conn` as an argument?
2. I think `conn.on()` is an error and it should be `conn:on()`?
3. Assuming that the answer to 1 is no and that 2 is true, then move `ondisconnect` above and have `cfini` call `csend(ondisconnect)` rather than duplicating code.

lua_modules/http/httpserver.lua
```diff
@@ -23,54 +23,49 @@ do
...
-      if self.send_header then
-        self.send_header = nil
-        -- end response headers
+    local send_header = function(header)
+      if type(header) ~= "table" then
+        print("httpserver.lua: header need a table! eg: {status = 200, headers = {Connection = \"close\"}}")
+        return
+      end
+      if type(header.headers) ~= "table" then
+        print("httpserver.lua: header.headers need a table!")
```
nwf: Here, too, throw an error.

docs/lua-modules/httpserver.md
```diff
@@ -5,6 +5,10 @@
 
 This Lua module provides a simple callback implementation of a [HTTP 1.1](https://www.w3.org/Protocols/rfc2616/rfc2616.html) server.
 
+!!!note
+	copy this module to nodemcu by yourself 
+
+
```
nwf: This is true of all `lua_modules`, in general.

lua_modules/http/httpserver.lua
```diff
@@ -116,13 +116,12 @@ do
...
-        -- NB: once length of seen chunks equals Content-Length:
-        --   onend(conn) is called
         body_len = body_len + #chunk
-        -- print("-B", #chunk, body_len, cnt_len, node.heap())
         if body_len >= cnt_len then
-          req:ondata()
+          -- last chunk
+          req:ondata(chunk, true)
+        else
+          req:ondata(chunk, false)
```
nwf: What's wrong with calling `req:ondata()` with a `nil` `chunk` to indicate the end of the data?  (I realize that `req:ondata()` is perhaps confusing, but it is equivalent to `req:ondata(nil)`.)

docs/lua-modules/httpserver.md
```diff
@@ -42,14 +46,15 @@ Callback function has 2 arguments: `req` (request) and `res` (response). The fir
 	- `name`: Header name
 	- `value`: Header value
 
-- `ondata`: value to setup handler function HTTP data. Handler function has 2 parameters:
+- `ondata`: value to setup handler function HTTP data. Handler function has 3 parameters:
 	- `self`: `req` object
 	- `chunk`: Request data
+	- `last`: set true while the last chunk
```
nwf: As mentioned earlier, I think I'd prefer that `chunk` be `nil` for the end of the stream.  The documentation could still stand an update, even so.

lua_modules/http/httpserver.lua
```diff
@@ -23,54 +23,49 @@ do
...
-    -- close connection
-    cfini()
-   end
-   --
+
     local res = { }
     res.send_header = send_header
     res.send = send
-    res.finish = finish
+    -- res.finish = finish -- no need, use send()
```
nwf: Then remove it, don't just comment it out.

lua_modules/http/httpserver.lua
```diff
@@ -23,54 +23,49 @@ do
...
-      csend("HTTP/1.1 ")
-      csend(tostring(status or 200))
-      -- TODO: real HTTP status code/name table
-      csend(" OK\r\n")
-      -- we use chunked transfer encoding, to not deal with Content-Length:
-      --   response header
-      self:send_header("Transfer-Encoding", "chunked")
-      -- TODO: send standard response headers, such as Server:, Date:
+   -- send("hello world\r\n") will send the data
+    -- send() will close connection
```
nwf: Try something like "`send(data)` sends the string `data` to the client as a HTTP chunk.  If `data` is `nil`, the stream will be terminated and the connection closed."

lua_modules/http/httpserver.lua
```diff
@@ -23,54 +23,49 @@ do
...
+        return
+      end
+      if type(header.headers) ~= "table" then
+        print("httpserver.lua: header.headers need a table!")
+        return
+      end
+      -- status
+      local check = status_code[header.status]
+      if check == nil then check = 200 else check = header.status end
+      csend(("HTTP/1.1 %s\r\n"):format(status_code[check]))
```
nwf: That's not terribly idiomatic Lua and defaulting to success is worrying.  I'd prefer, I think,
`csend(("HTTP/1.1 %s\r\n"):format(status_code[status] or status))`
which replaces `header.status` only if it's found in the table and otherwise passes exactly what's been given.  If you want the API to default to 200, which... well, maybe... then add a `status = status or 200` earlier in the function.

lua_modules/http/http-example.lua
```diff
@@ -21,17 +23,16 @@ require("httpserver").createServer(80, function(req, res)
     -- end
   end
   -- setup handler of body, if any
-  req.ondata = function(self, chunk)
+  req.ondata = function(self, chunk, last)
     print("+B", chunk and #chunk, node.heap())
-    if not chunk then
+    if last then
```
nwf: As said elsewhere, I'd prefer the original API.

lua_modules/http/httpserver.lua
```diff
@@ -23,54 +23,49 @@ do
...
+        -- close connection
+        cfini()
+      end
     end
-    if data then
-      -- NB: no headers allowed after response body started
-      if self.send_header then
-        self.send_header = nil
-        -- end response headers
+    local send_header = function(header)
```
nwf: I think a better API would be for `send_header` to take the `status` and `headers` separately, rather than packaged together into a table.

lua_modules/http/httpserver.lua
```diff
@@ -7,6 +7,14 @@
...
 
 local http
+
+-- add some status code
+local status_code = {}
+status_code[200] = "200 OK"
+status_code[304] = "304 Not Modified"
+status_code[404] = "404 Not Found"
+status_code[500] = "500 Internal Server Error"
+
```
nwf: While I anticipate that `httpserver.lua` will usually end up in LFS, is there any benefit for these?  Does the standard require them, and if so, shouldn't we have the full set or require the user to supply it?  If we keep them, replace "add some status code" with "translate numeric HTTP codes to required descriptive strings" or such.

docs/lua-modules/httpserver.md
```diff
@@ -42,14 +46,15 @@ Callback function has 2 arguments: `req` (request) and `res` (response). The fir
...
 
 The second object holds functions:
 
-- `send(self, data, [response_code])`: Function to send data to client. `self` is `req` object, `data` is data to send and `response_code` is HTTP response code like `200` or `404` (for example)
-- `send_header(self, header_name, header_data)`: Function to send HTTP headers to client. `self` is `req` object, `header_name` is HTTP header name and `header_data` is HTTP header data for client.
-- `finish([data])`: Function to finalize connection, optionally sending data. `data` is optional data to send on connection finalizing.
+- `send(data)`: Function to send data to client. `data` is data to send if the `data` is `nil` will close the connection
```
nwf: Yeah, see, here you use `nil` to signal the end of the stream.  Please use that for `ondata`, too, for consistency.

lua_modules/http/httpserver.lua
hanfengcan: I saw that has a todo( add status code table). so I add this. all the code maybe unnecessary I think 😂

docs/lua-modules/httpserver.md
hanfengcan: at first time, I don't know that must be copy by myself. so add a note to tell others who don't know

lua_modules/http/httpserver.lua
hanfengcan: I am not sure. in fact, when the disconnect event came, call this `ondisconnect` will reboot. 

I think that  should set the `res` and `req` to nil to avoid memory leak in this disconnect callback.

This `ondisconnect` had before. so, as few changes as possible

lua_modules/http/httpserver.lua
hanfengcan: the new `send` already look like this. when use `send()` will close the connection.

lua_modules/http/httpserver.lua
hanfengcan: Is it 3 function?
one is the `send` just send the response body data.
other is `send_status` just send the first line `HTTP/1.1 200 OK\r\n`
and `send_header` just send the headers like `Connection: close`

lua_modules/http/httpserver.lua
hanfengcan: if res has 3 function. Throw error will needless

lua_modules/http/httpserver.lua
hanfengcan: OK

lua_modules/http/httpserver.lua
hanfengcan: Call once  maybe batter then call twice when the last chunk was came

lua_modules/http/httpserver.lua
hanfengcan: like this ?
```lua
-- status maybe `404` or `Not Found`
function send_status(status)
    status = status or 200
    csend(("HTTP/1.1 %s\r\n"):format(status_code[status] or status))
end
```

lua_modules/http/httpserver.lua
nwf: I think the reason for the reboot is that `conn.on` should be `conn:on`.  Please test with that change in place, and if it solves the problem, please add it to the commit stack for this PR.

docs/lua-modules/httpserver.md
nwf: That's fair, but I think this disclaimer should be higher in the document hierarchy, perhaps where we introduce the `lua_modules`.

lua_modules/http/httpserver.lua
nwf: Oh, that's an interesting counter-point... hm.  I *think* I still like `nil` better, but would appreciate hearing opinions of more proficient Lua programmers (@TerryE or @pjsg or @djphoenix spring to mind).

lua_modules/http/httpserver.lua
nwf: Sorry to be unclear; I was just meaning to replace the comment with something a little more descriptive and less example-centric.

lua_modules/http/httpserver.lua
nwf: I think that's correct code, though I don't think that "Not Found" is an acceptable `status` value?  AFAICT the acceptable values are of the form `404` (the integer), `"404"` (the string), or `"404 Not Found"` (the string), yes?

lua_modules/http/httpserver.lua
nwf: Oh I see.  Well, either way's OK by me, I guess. :)

lua_modules/http/httpserver.lua
nwf: The old API, AFAICT, indeed had `status` passed to a different function, didn't it?

Maybe the three-function API is a better plan; would you mind seeing what it makes the example look like?

lua_modules/http/httpserver.lua
hanfengcan: OK, let me try

docs/lua-modules/httpserver.md
hanfengcan: agree

lua_modules/http/httpserver.lua
hanfengcan: -- send(nil) ?

lua_modules/http/httpserver.lua
hanfengcan: the `send` is current version

```lua
-- status maybe `404` or `404 Not Found`
function send_status(status)
    status = status or 200
    csend(("HTTP/1.1 %s\r\n"):format(status_code[status] or status))
    csend("Transfer-Encoding: chunked\r\n")
end

function send_headers(headers)
    -- header
    for k, v in pairs(header.headers) do
      csend(k)
      csend(": ")
      csend(v)
      csend("\r\n")
    end
    csend("\r\n")
end
```

because `csend("\r\n")` in `send_headers`. so, a necessary operation is 
```lua
send_status()
send_headers()
send("hello world")
send()
```

nwf: I think I would prefer a two-function API, but with a twist:

```lua
sendbody = res:send_headers(status, {foo = bar, ...})
sendbody(data)
-- ...
sendbody(nil)
```

That is, I would like `send_headers` to return the body-sending function; this way, there's no (or at least, less) ambiguity about which one the user is to call when, I think?

hanfengcan: great

HHHartmann: But what if the user does not need to send any headers?

hanfengcan: I think `res:send_headers()` send the 200 without any headers.

hanfengcan:
@nwf  Hi, I update some change. But except the response handler

lua_modules/http/httpserver.lua
nwf: I think that points 1 and 3 still stand here?

lua_modules/http/httpserver.lua
nwf: I would still like a 2nd opinion on the API here.

lua_modules/http/httpserver.lua
HHHartmann: but ```status_code``` defines "404 not found", so this seems to be correct to me.

nwf: @HHHartmann The comment above the function suggests that "Not Found" is an acceptable `status` value.  That, I think, is not true.  The comment should be something like
```
-- status may be any of: an int code (`404`), stringified int code (`"404"`), or code and text (`"404 Not Found"`)
```
(As an aside, English is a terrible language; "maybe" and "may be" mean very different things).

HHHartmann: Ok I agree. Didn't get that you meant the comment.
I hope that @hanfengcan will pick up on this PR. Would really be a pitty it his work would get lost.

docs/lua-modules/httpserver.md
```diff
@@ -42,14 +46,15 @@ Callback function has 2 arguments: `req` (request) and `res` (response). The fir
...
+and `response_code` is HTTP response code like `200` or `404` (for example)
```
HHHartmann: the response code line has to be removed. The status moved to the send_header method.

docs/lua-modules/httpserver.md
```diff
@@ -42,14 +46,15 @@ Callback function has 2 arguments: `req` (request) and `res` (response). The fir
...
+- `send_header(header)`: Function to send the response header. `header` must be a table. likes `{status = 200, headers={Connection = "close"}}`. status will be set 200 if the status omitted. the built-in code is `200, 304, 404, 500`
```
HHHartmann: I would prefer the header array to be given separately. It is an unnecessary creation of an object which contains the status_code and the headers which don't really belong together and can as well be given separately:

- send_header(status, headers)`: Function to send the response header. status will be set 200 if the status is omitted. The built-in codes are `200, 304, 404, 500`. `headers={Connection = "close", "Content-Type" = "text/plain"}}`.

lua_modules/http/http-example.lua
HHHartmann: Me too, It would be more consistent with `send` as you said (elsewhere?)

docs/lua-modules/httpserver.md
nwf: The API I proposed above (`sendbody = res:send_headers(status, {foo = bar, ...})`) is as you describe, @HHHartmann, right?

HHHartmann: Ah yes the discussion is there, sorry.

lua_modules/http/httpserver.lua
HHHartmann: I would prefer to have 3 methods right from the start.
`send_status` and `send_header`  (or combined in one function)
`send(body)`

I would prefer to be able to omit any of the calls so to just send an 200 OK response calling `send(nil)` would be sufficient.

The server would need to have a status to know when it has to internally send status and headers and also to prevent sending status or headers multiple times or in a wrong order.

nwf: @HHHartmann: How often does one want to send an empty 200 response, and is `res:send_headers()()` really that bad if that's what you want?

docs/lua-modules/httpserver.md
nwf: No apology needed; I was asking to make sure that the proposal matched what you wanted.  The docs still need to be updated.

lua_modules/http/httpserver.lua
hanfengcan: there has `\r\n` between response header and response data. if just send an 200. the `\r\n` need send alone.

so, I think that send_status and send_header need combined

docs/lua-modules/httpserver.md
hanfengcan: OK

lua_modules/http/httpserver.lua
hanfengcan: The `ondisconnect(conn)` is need. lost the `conn`, the menory will leak.

nwf: OK.

galjonsfigur:
@hanfengcan @nwf @HHHartmann I'm not involved in reviewing this fix, but because `luacheck` now checks every .lua file in this repository, including this module, there is a report with some warnings: https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/558453269#L4764 - those can be only nitpicks that can be ignored using comment line `-- luacheck: ignore` in the beginning of code, but I think it may be helpful to see it.

marcelstoer:
@nwf I see two ways forward from this point (there may be more): you approve and merge, we postpone this and we drop it from the upcoming master snap in which case we'd remove the `Next release` label. WDYT?

nwf:
I do not have time to fix this myself, and we haven't heard from @hanfengcan in a while, I think.  I would suggest removing it from the next drop; I don't think this httpserver gets a lot of use anyway?

hanfengcan:
what 🤣  I think that all change is finish

the `req:ondata()`  call once, I insit

the `res:send_headers` return function doesn't seem to do much good

nwf:
Oh, OK then.  You might fix the `luacheck` warnings reported in @galjonsfigur's link (https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/558453269#L4764) but I guess I'm otherwise OK with this PR landing.

hanfengcan:
how to fix this `setting non-standard global variable _`

galjonsfigur:
@hanfengcan You can try to add `--luacheck: push ignore` before the line and `--luacheck: pop` after it - something like this:
```lua
-- luacheck: push ignore
_, _, someVar = someFunct()
-- luacheck: pop
```
This is probably a bug because `luacheck` shouldn't issue any warning but I encountered this problem and this should be the easiest fix.

HHHartmann:
Adding
'local _'
Might be a better fix. Havn't tried though.

TerryE:
`_` is just another Lua variable that will be created as a global if not local, and you definitely don't want this to happen.

galjonsfigur:
> _ is just another Lua variable that will be created as a global if not local, and you definitely don't want this to happen.

Yes - I don't know why I wrote about luacheck having a bug - sorry, my bad. Luacheck ignores `_` but only when it's used for returning values you don't want from a function. So that one warning [here](https://github.com/nodemcu/nodemcu-firmware/blob/45d7a9cc2bdb073e5625fa2ae328fdade05d28d7/lua_modules/http/httpserver.lua#L145) shouldn't be ignored and instead local  `_` declared somewhere earlier (for example [here](https://github.com/nodemcu/nodemcu-firmware/blob/45d7a9cc2bdb073e5625fa2ae328fdade05d28d7/lua_modules/http/httpserver.lua#L85))

Also in commit https://github.com/nodemcu/nodemcu-firmware/pull/2791/commits/21690fa4adbb2c86e112995ab9f5d64cb95db5da you removed unused function arguments - I think that it could break existing code. Instead replace them with `_` and `luacheck` will be happy.

To fix `conn` warnings about shadowing upvalues you can just rename the function arguments and it's  occurrences in functions `ondisconnect` and `onreceive`  to `connection`  for example.

TerryE:
>  Instead replace them with `_` and `luacheck` will be happy.

Nah, bad practice precisely because `_` is just regular variable and is used as a bit bucket by convention.  Try:
```Lua
do local _ ="oops"
  print ((function(p) return _,_,p end)'hello')
end
``` 
in any Lua interpreter and see what happens.

marcelstoer:
Pity we didn't merge this in due time 😞@hanfengcan could you please resolve the conflicts? And then @nwf (or @galjonsfigur) could take a final look even though it's already approved. Would be great if we could land this on `dev` before the next release.

hanfengcan:
It's my frist time do this. I hope I'm not mistaken 😂

lua_modules/http/httpserver.lua
HHHartmann: Still SendHeader needs to be called with a complex structure, Why do we change the interface in such a way without good reason? It will break everybodies code. Users will even have to adapt their logic if unlucky and collect all header data themselves to pass them at once. Aside from this multiple headers with the same key are not possible that way.

docs/lua-modules/httpserver.md
HHHartmann: Please don't just overwrite on merge conflicts.

marcelstoer:
Superseded by #3307

PR #2795, opened 2019-06-18
Docsupd

cwrseck:

- [x ] This PR is for the `dev` branch rather than for `master`.

To document the initial setup of LFS for new users.


nwf:
This is not targeting the `dev` branch, despite the box being ticked.

cwrseck:
I spent 15 minutes writing the patch, and two days fighting Markdown, git  and GitHub to try and generate a pull request.  Obviously, as @nwf says, I got it wrong.  Can this pull request just be deleted?  Synchronising three separate repositories isn't easy; I think changes had better come from someone who actually understands git.

Sorry for the waste of time.

C W Rose

galjonsfigur:
@cwrseck Take it easy - Git can be confusing but often the best way to use it is to use GUI Git client (even some IDEs like Eclipse have Git plugin) and that way you will see exactly what you are doing. And this branch targets `dev` but it seems like locally you worked in `master` branch and that's the reason for all those additional commits. 

I managed to merge your commit into `dev` branch, so you are still visible as the author. If you want I can make the PR as the results are in [my branch](https://github.com/nodemcu/nodemcu-firmware/compare/dev...galjonsfigur:dev-doscupd?expand=1) (or even you can make a PR but I never tested if GitHub allows to do a PR from other user branch)

TerryE:
Honestly, I work on fairly complex changes that take time to develop.  What I always do now is to develop the patch, trash my local copy of `dev`,  fetch the latest version, and then rebaseline my patch against it before pushing it to my fork then submitting it. This takes less than 5 mins once you've got the hang of it.   Google and StackOverflow are your friends if you get stuck.

cwrseck:
@galjonsfigur - if you could go ahead and submit a pull request I'd be grateful; I'm not really bothered about who is credited.  One of the problems was that I never really clarified my understanding of the workflow between the repositories GitHub original / GitHub personal / local personal.  A day or so's reading has got me mostly there, and a couple more trials should fix it.

Thanks for the help.

C W Rose

Issue #2787, opened 2019-06-08
RFC - Sorting out the input handling


TerryE:
Another case of "I am currently doing my Lua 5.3 port and there are a number of issues on which I would like to give a heads-up / get feedback from the contributors." This one relates to how we integrate input handing into `lua.c`.  

I have never really taken a close look at this bit of `lua.c` + `user_main.c` `uart.c` before, but it is really all a bit of a botch -- the sort of thing that might have evolved if you were hacking an application together against a deadline rather than designing it properly -- which was the case for the early versions 1.x of Lua51, but now I am trying to srt out Lua53 for the long term so it's time to design it properly, IMO.

Architecturally, the standard `src/lua.c` assumes that an underlying OS will provide some form of API for a synchronous `readline()` function that handles stuff like terminal echo; the Lua interpreter sits in the `doREPL()` function looping around this readline call.  This approach doesn't map onto an event-driven task-orientated system, where the uart driver or posts a receive events for each line received which must then be handled and an atomic task.  Hence this `doREPL()` logic must be inverted, so that the interactive processing is all triggered from a line-in callback.

Note that the way the Lua interpreter does chunking is a bit like a sledge hammer.  Each source chunk is assembled from sequence of successive input lines.  The entire chunk is recompiled each time an extra line is added .
-  If it is successful then compiled chunk is executed and the line list flushed.
-  If the compile raises an `<eof>` error then the chunk is assumed incomplete and the interpreter waits for another line.
-  Any other error is reported and the line list flushed.

The means that if a 20 line chunk is received then the 1st line compiles a 1-line chunk, the 2nd a 2-line chunk and so on so the 20th line compiles a 20 line chunk.  Hence the per-line the compiler overhead goes uniformly as the chunk size increases.  This isn't really an issue if the lines are really being typed interactively. However, if some IDE such as Esplorer or even if you are just pasting a chunk of code into PuTTY, then it is quite easy to get to the point where the input overruns the ability of the interpreter to flush input buffer.

Another complication in the case of NodeMCU is that there is nothing stopping other application callbacks firing whilst you are typing and thus threading other Lua processing tasks in between the lines in a chunk.

Given then everything outside the `app/lua` or `app/lua53` directories is going to be shared then its going to be simpler for me to packport some of this tidy-up into `app/lua/lua.c` so that we can keep the rest common.

[...]

PR #2797, opened 2019-06-19
Baseline version of Pipe  Library

TerryE:
As discussed in #2787.

As I discussed in this https://github.com/nodemcu/nodemcu-firmware/issues/2787#issuecomment-502491598, this Pipe library provides a very runtime-efficient method of piping a records stream from one Lua task to another -- a C library version of @nwf's FIFO Lua module.  It stores the queued pipe data as a Lua array of 256 byte Userdata pages with the empty pipe having an empty array.  

As well as the obvious `pipe.create()`, `p:read()` and `p;write()` methods, it also has a `p:reader()` which can be used in a for loop: `for l in p:reader('\n') do print(l) end` and a  `p:unread()`.

I needed this for my Lua53 implementation so it made sense to release it now.


docs/modules/pipe.md
```diff
@@ -0,0 +1,131 @@
...
+#### Syntax
+`pobj:read([size/end_char])`
+
+#### Parameters
+- `size/end_char`
+	- If numeric then a string of `size` length will be returned from the pipe.
+	- If a string then this is a single character delimiter, followed by an optional "+" flag.  The delimiter is used as an end-of-record to split the character stream into separate records.  If the flag "+" is  specified then the delimiter is also returned at the end of the record, otherwise it is discarded.
+  - If omitted, then this defaults to `"\n+"`
+
+Note that if the last record in the pipe is missing a delimiter or is too short, then it is still returned, emptying the pipe.
```
HHHartmann: It would be helpful to know this condition to be able to recreate the original content. Especially if a delimiter without + was given. 
Maybe a second return value ```incomplete```.

TerryE: Gregor, I disagree here.  If you want the delim then use the '+' option.  If you don't, then you don't.  I want to keep this interface simple.  If you want to test for it's existence then use `+` and 
`if rec:sub(-1) == delim`

TerryE: Sorry the default is `"\n"`, not `"\n+"`.  I will update this on the next update.

marcelstoer:
Terry, I don't want to spoil the party and I try to be constructive. However, there are few governance issues:
- self-approval of PR plus
- merged after less than 24h exposure -> very little time for others to review
- missing `new module` label
- missing link to `next release` milestone

The documentation is missing in `mkdocs.yml`.

TerryE:
@marcelstoer, the rules are there as a framework, not to be abused.  The root of why I did this is based on my experience of the other development that I do to the core Lua engine.  I find that very few _if any_ reviewers pull the fork copy and try the code out and give suggestions and feedback based on proposed feature / functionality.  I wish this wasn't the case, but it is what it is.  At best I seem to get static eyeballing on the source and not a discussion about what it does and why.  

I needed this commit into `dev` to do my core `lua.c` alignment between `lua`\(51) and `lua53`, and the only other alternatives were to include `pipe` in this `lua.c`  PR or to stretch the whole process by a week or two.

This is a new module and the "do nothing" case is that it doesn't get linked into the firmware, so this is a zero-risk merge.  At least now our committers can fetch the latest `dev`, enable the module in `user_modules.h` and build to have a play.  

Now that the baseline is there, I have a draft issue in other open window in my browser, so we can discuss the feature / functionality.

The missing  `new module` label and link to `next release` milestone were deliberate, correct or not.  As I said, I want to get the `lua.c` core stuff out there and I don't want to expose this to general developers downloading `dev` until its interface is stable.

I am not sure that another efficient alternative is -- perhaps another `dev-core` branch for doing this work -- but I think that we would both agree that the odd abuse of the letter of the governance rules rather than the principles is an easier alternative.

app/modules/pipe.c
```diff
@@ -0,0 +1,309 @@
...
+** Read the docs/modules/pipe.md documentation for a functional description.
+*/
+
+#include "module.h"
+#include "lauxlib.h"
+#include <string.h>
+
+#define INVALID_LEN ((unsigned)-1)
+
+#define LUA_USE_MODULES_PIPE
```
galjonsfigur: I just checked this patch and it seems that this module is enabled by default by this `define`. If its supposed to be enabled by default I think `#define LUA_USE_MODULES_PIPE` in `user_modules.h` should be uncommented and this one removed but I could be wrong.

TerryE: uuUUaaarggh.  Sorry.  I've just added about 2½-3Kb to `dev` builds.  Other than this, no harm done -- unless of course you start to use it.  Yup it should have been disabled.  Though it will need to be enabled by default on my next PR as it will used by `app/lua/lua.c'

PR #2805, opened 2019-06-21
Reworked linker magic to work with IDF 3.3

jmattsson:

With the IDF asserting full control over the linker scripts and insisting on
the application description being the first entry in the .flash.rodata
section, or previous method of doing link-time arrays stopped working.
Why? Because the build patched in a SHA256 digest straight into our arrays.

With the limited language of the gcc linker scripts I could find no other
way of getting our arrays in anywhere else.

The IDF "linker fragments" support also can not be made to work for our needs:
  - no support for setting alignment before including objects
  - no support for declaring symbols
  - no support for adding our terminating zeros
  - insists on grouping objects by lib rather than by declared grouping,
    which means we could at most have a single link-time-array using
    the IDF mechanism
  - also does not like underscores in section names, but that's just an
    annoyance

So, the least bad option that I could come up with was to use a project-wide
makefile snippet to add a target in-between the IDF's generation of the
esp32.project.ld file, and the linking of our NodeMCU.elf. In this target
we read in the esp32.project.ld linker script, check whether we have our
arrays in there, and if not rewrites the linker script.

The resulting change is quite small, but it's taken a full day to get here *sigh*.

Oh, and the esp32.project.ld file only came into existence on the IDF 3.3
branch, so I had to change up the IDF to the latest release/3.3 as well.
I would've preferred to not do it just yet, or at least with a stable tag,
but the v3.3-beta3 had a really nasty regression for us (can't add partition
entry), so that was a no-go.

### DIFF
--- a/components/base_nodemcu/Makefile.projbuild
+++ b/components/base_nodemcu/Makefile.projbuild
@@ -1,0 +1,8 @@
+BASE_NODEMCU_DIR:=$(dir $(lastword $(MAKEFILE_LIST)))
+BASE_NODEMCU_BUILD_DIR:=$(BUILD_DIR_BASE)/base_nodemcu
+
+$(BUILD_DIR_BASE)/$(PROJECT_NAME).elf: $(BASE_NODEMCU_BUILD_DIR)/ld_patched
+
+$(BASE_NODEMCU_BUILD_DIR)/ld_patched: $(BUILD_DIR_BASE)/esp32/esp32.project.ld
+	"$(BASE_NODEMCU_DIR)/add_rodata_ld.sh" "$<" "$(BASE_NODEMCU_DIR)/ld/nodemcu_rodata.ld"
+	touch $@
--- a/components/base_nodemcu/add_rodata_ld.sh
+++ b/components/base_nodemcu/add_rodata_ld.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# syntax: add_rodata_ld.sh /path/to/esp32.project.ld /path/to/snippet_file
+set -eu
+ldfile="$1"
+partial="$2"
+out="${ldfile}.new"
+IFS='
+'
+msg="/* NodeMCU patched */"
+if [[ $(head -n 1 "$ldfile") =~ "$msg" ]]
+then
+  echo "NodeMCU rodata already patched into $(basename $ldfile)"
+  exit 0
+else
+  echo "Patching in NodeMCU rodata into $(basename $ldfile)"
+  echo "$msg" > "$out"
+fi
+
+cat "$ldfile" | while read line
+do
+  if [[ $line =~ "drom0_0_seg" ]]
+  then
+    cat "$partial" >> "$out"
+  fi
+  echo $line >> "$out"
+done
+mv "$out" "$ldfile"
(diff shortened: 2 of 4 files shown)


devsaurus:
Wow, thanks for digging into this Johny! I'll give it a try soon and report back here.

jmattsson:
Haha, in the end it was a lot less magic and considerably more *"Well fine then - I'll do it the ugly way!"*. I do love `make` though, it's so powerful :)

chilipeppr:
I get this error when building due to this PR. When digging into this error it seems the shell script add_rodata_ld.sh is erroring out thus it says "not found." I'm building on Ubuntu 19. When I revert this PR it builds fine.

Generating esp32.project.ld
/bin/sh: 1: /home/john/nodemcu-firmware-cp2/components/base_nodemcu//add_rodata_ld.sh: not found
make[1]: *** [/home/john/nodemcu-firmware-cp2/components/base_nodemcu/Makefile.projbuild:13: /home/john/nodemcu-firmware-cp2/build/base_nodemcu/ld_patched] Error 127
make[1]: Leaving directory '/home/john/nodemcu-firmware-cp2'
make: *** [Makefile:19: all] Error 2

jmattsson:
@chilipeppr That's suggesting you don't have `/bin/bash` or that it's not executable. Has Ubuntu changed to only install `zsh` by default or some such?

chilipeppr:
/bin/sh is there. /bin/bash is default. /bin/zsh is not there.

jmattsson:
How odd. What happens if you try to invoke it manually with `bash -x /home/john/nodemcu-firmware-cp2/components/base_nodemcu/add_rodata_ld.sh`? It should complain about missing $1 at that point, but I'd be curious to see if it gets that far.

jmattsson:
Ah, and check to make sure you haven't somehow ended up with a Windows style line ending - that has certainly been known to trigger those unhelpful `not found` errors. You can use something like `sed -i s'/\r//' components/base_nodemcu/add_rodata_ld.sh ` to strip our any carriage returns from the file.

chilipeppr:
Hmm. Ok. Let me check the CRLF vs LF issues. I do use Windows and the Github Desktop app on Windows with a Samba share to Ubuntu 19 for the build, so perhaps that's it. I also use my own fork on Github and have to do reverse pull requests to update my repo on Github from dev-esp32 branch.

I'm trying to wrap up my awesome Touch module for ESP32 so was trying to do the pull request tonight, but this build issue is causing me hours of pain as I've re-checked out clone after clone to debug what's wrong.

chilipeppr:
Interesting. When I checkout a new clone of Github dev-esp32 repo directly from Ubuntu 19 it builds fine. When I checkout a new clone via Github Desktop on Windows via Samba share to Ubuntu 19 it fails to build with that add_rodata_ld.sh issue.

jmattsson:
Right, git is trying to be clever and is outsmarting itself :(

I'll see if I can add something to force the correct line endings regardless, at least for scripts.

jmattsson:
@chilipeppr if you have a few minutes, could you try cloning https://github.com/DiUS/nodemcu-firmware/tree/esp32-no-crlf from Windows and see if it still messes up the line endings on the Ubuntu side? I've added a `.gitattributes` file which I'm hoping will prevent git from messing with the line endings of shell scripts. If it works, I'll PR it to NodeMCU proper.

chilipeppr:
Ooh, headed to bed. Have to get on flight in morn for 2 weeks in Europe, so not sure I can get to this anytime soon. Thanks for the quick help though. It really saved me.

jmattsson:
No worries, we can get back to it some time later. Enjoy your trip!

PR #2493, opened 2018-09-20
End User Setup: Save the post data in a file on the filesystem

Gui13:
Possibly Fixes #1302, and is a follow up on https://github.com/nodemcu/nodemcu-firmware/pull/2132 .

- [x] This PR is for the `dev` branch rather than for `master`.
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [x] I have thoroughly tested my contribution.
- [x] The code changes are reflected in the documentation at `docs/en/*`.

Several PR have tried to implement this, and I took to myself to try and finish this.

The idea is to save the parameters that the user sent in his EndUser setup into a file on the ESP filesystem. 
Once we have that, you can modify the eus HTML file and add several fields that will be sent along with the wifi credentials. These fields and values will be saved in a LUA file on the filesystem, that you can now load when needed to access it.

I implemented several things:
- The /setwifi API now is a standard form POST, which is way easier to handle (the body contains the key/values) and more correct if you want to send lots of data.
- I parse the body and save the keypairs in a specialized struct, which I later serialize on the ESP filesystem

I have tested the code in a separate standalon project with Valgrind for the parsing/fill up of the struct part, and I'm fairly confident that it should be leak free.

As an MVP2, we could also pass the onConnected callback a structured LUA table which contains all the variables set in the POST.

Please, feel free to suggest code enhancements, my C is a bit rusty.

Issue #2799, opened 2019-06-19
Add the nodemcu-firmware version hash to the signon.


cwrseck:
Running the current (20190405) NodeMCU build on an Adafruit HUZZAH I have:
```
> 
> print(node.info()) 
2	2	0	1389798	1458400	4096	2	40000000
>
```
and a reset gives:
```
NodeMCU 2.2.0.0 build April 2019 powered by Lua 5.1.4 on SDK 2.2.1(6ab97e9)
```
I've been updating a number of systems, and a more precise indication of
the running NodeMCU version would be helpful.  The easiest way would seem
to be the use of the Git checksum, perhaps via 'git log -1 --pretty="%H"'
or "%h"; if a Git server hook were used to write that information to a
project header file, and the result added to the signon, a firmware binary
could be easily identified.

Does GitHub have server hooks available to the project?

### Justification
It would make checking and updating firmware easier..

### Workarounds
No, but editing the header files directly would be possible.

C W Rose

HHHartmann:
Maybe we could move the mechanism used in docker to the build system. 
This should then append the given message from the header.

nwf:
I'd prefer that the docker system ultimately move to something else; rewriting the header file is kind of rude and leaves the build tree dirty when the builds fail.  The usual dance here is to have `make` build a file from the output of `git describe`; something like:

```
$(shell echo 'static const char git_rev_bytes[] = { $(shell git rev-parse HEAD | sed -e "s/\(..\)/0x\1,/g") };' >> $(ROOT_DIR)/build-version.h)
```

A slightly better version of the above generates a temp file and only replaces the `build-version.h` if the temp file differs from the existing one, so that `make`'s extremely limited change detection logic doesn't cause rebuilds all the way up the dependency tree.

cwrseck:
@nwf the Makefile might well be a better approach; I did something similar with Mercurial and an (Autotools) makefile.  Given a separate version.h, updated only on success, excessive rebuilds should be avoidable.

C W Rose

marcelstoer:
This is more or less a dup of #2398 

> I'd prefer that the docker system ultimately move to something else; rewriting the header file is kind of rude and leaves the build tree dirty when the builds fail.

👍fully agree

[...]

PR #2810, opened 2019-06-22
[Replacement for 2493] EUS: Save the post data in a file on the filesystem 

marcelstoer:
This replaces #2493 which has been laying around for so long that it now has a couple of merge conflicts. I pulled in that branch and rebased it on `dev` (rather than bothering the author with this).

Tagging everyone who's been involved in the other PR: @Gui13, @davidcbittner, @Jcrash29, @Pirngruber, @hanfengcan, @IDispose. I invite you to take the module for one last test ride before we merge these changes.

Here's a binary with the following properties for testing:

```
	branch: eus_params_post
	commit: efe1103d9c9c7cde62cffcef9de16c371803e498
	SSL: true
	Build type: float
	LFS: disabled
	modules: enduser_setup,file,gpio,http,i2c,mdns,mqtt,net,node,spi,tmr,uart,wifi
 build created on 2019-06-22 21:53
 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
```
[nodemcu_float_eus_params_post_20190622-2153.bin.zip](https://github.com/nodemcu/nodemcu-firmware/files/3317326/nodemcu_float_eus_params_post_20190622-2153.bin.zip)


HHHartmann:
Just reading the docs I see two weaknesses. 
As i recall it has been mentioned in the original issue that writing a file and then having to read it afterwards wears SPIFFS and is unnecessarily slow. 
It is also not desirable to generate global variables when reading the file via dofile.
Would be nice here to return an array here.
The file could look like this:
``` Lua
local p = {}
p.wifi_ssid="ssid"
p.wifi_password="password"
-- your own parameters:
p.timeout_delay="xxx"
p.device_name="yyy"

return p
```

But alltogether it seems to be better to return the array right away.
Saving as json might also make sense, but then the sjson module would be required.

But this could also be done in a separate PR so this one will be integrated finally.

docs/modules/enduser-setup.md
```diff
@@ -30,8 +65,8 @@ The following HTTP endpoints exist:
...
 |/status|Returns plaintext status description, used by the web page|
 |/status.json|Returns a JSON payload containing the ESP8266's chip id in hexadecimal format and the status code: 0=Idle, 1=Connecting, 2=Wrong Password, 3=Network not Found, 4=Failed, 5=Success|
-|/setwifi|Endpoint intended for services to use for setting the wifi credentials. Identical to `/update` except returns the same payload as `/status.json` instead of redirecting to `/`.|
-|/update|Form submission target. Example: `http://example.com/update?wifi_ssid=foobar&wifi_password=CorrectHorseBatteryStaple`. Must be a GET request. Will redirect to `/` when complete. |
+|/setwifi|Endpoint intended for services to use for setting the wifi credentials. Identical to `/update` except must be used as a form POST and returns the same payload as `/status.json` instead of redirecting to `/`.|
+|/update|Form submission target. Example: `http://example.com/update?wifi_ssid=foobar&wifi_password=CorrectHorseBatteryStaple`. Must be a GET request. Will redirect to `/` when complete. Note: won't update the eus_params.lua file. |
```
HHHartmann: Then when is eus_params.lua being updated? I'm a bit puzzled?
Or is this meant for the /setwifi route?
It should be updated in both cases.

app/modules/eus/enduser_setup.html
```diff
@@ -153,14 +153,17 @@
...
-				<input id=wifi_password type=text autocorrect=off autocapitalize=none autocomplete=off placeholder=Password />
-				<button id=submit type=button>Save</button>
+				<form action="/setwifi" method="POST">
+					<button id=networks type=button class=utility></button>
+					<div id=dropdown>
+						<span id=arrow>&#x25bc;</span>
+						<select id=aplist name=aplist></select>
+					</div>
+					<input id=ssid name=wifi_ssid type=text autocorrect=off autocapitalize=none placeholder='Wi-Fi Name' />
+					<input name=wifi_password type=text autocorrect=off autocapitalize=none autocomplete=off placeholder=Password />
```
marcelstoer: I know it's been like that before but this should really be `type=password`.

docs/modules/enduser-setup.md
marcelstoer: The choice of HTTP API is somewhat of an acquired taste, I agree. However, I don't mind `/update` and `/setwifi` being distinctively different. I overhauled that table to make things a little clearer. See https://github.com/nodemcu/nodemcu-firmware/blob/e98a1042b05037c00e85151fcf8c28d4ea968b6b/docs/modules/enduser-setup.md#http-endpoints

marcelstoer:
> It is also not desirable to generate global variables when reading the file via dofile.
> Would be nice here to return an array here.

👍 definitely. I think we shouldn't merge it without that fix as changing the file format later would mean a breaking change. Are you up for it?

HHHartmann:
I'm not sure what you mean. No benefit in json or no benefit in CB with array?

I would like to implement returning the data in a CB and not writing a file at all. 

Having to dofile leaves a dirty heap. I don't know how bad sjson is in this respect. But probably better.

marcelstoer:
IMO C modules shouldn't depend on each other. Requiring SJSON to use EUS is a no-go for me. Even more so as I see no benefit in storing the data as JSON. After all it's a simple key-value structure.

> I would like to implement returning the data in a CB and not writing a file at all.

From firmware perspective this would be nice indeed. However, you're likely only *delegating* the "dirty" work - to the application developer. For all the use cases _I_ can imagine (I am sure there are many more) you will want to keep the custom settings around in a persistent manner. Hence, I expect most developers will write them to a file in the callback.

> Having to dofile leaves a dirty heap.

No way around that?

app/modules/enduser_setup.c
```diff
@@ -755,12 +755,20 @@ static int enduser_setup_write_file_with_extra_configuration_data(char *form_bod
...
   }
 
-  // write all key pairs as KEY="VALUE"\n
+  // write all key pairs as KEY="VALUE"\n into a Lua table, example:
+  // local p = {}
+  // p.wifi_ssid="ssid"
+  // p.wifi_password="password"
+  // p.device_name="foo-node"
+  // return p
+  vfs_write(p_file, "local p = {}\n", 13);
```
HHHartmann: Ah darn, I prepared exactly the same fix (except for the comment)
I then removed the spaces around '=' to have a common coding style throughout the file. :-)
Didn`t have a chance to check it though and so did not dare to push it.

marcelstoer: Oh, and I thought you had lost interest 😉 Good point about the spaces; now fixed.

app/modules/eus/enduser_setup.html
HHHartmann: It really should

app/modules/enduser_setup.c
HHHartmann: I was busy with #2799 and just had seen that the branch was in the main repo and thus shared.

marcelstoer: That's great, thanks. Could you please assign the issue to yourself then? Helps in trying to avoid that several people work on stuff in parallel.

app/modules/eus/enduser_setup.html
marcelstoer: I will also fix this. And we should mention `app/modules/eus/prepare.sh` in the `.md`. I only just found it when I wondered how to prepare the backup `.def` file.

HHHartmann: I will also have a look to get it checked out under Windows with Unix linens. Unless you beat

marcelstoer: Don't worry. I don't work on Windows.

HHHartmann:
Looking at this I have some more ideas :-)
The file ```enduser_setup.html.gz``` does not need to be checked in, it is only needed to create ```http_html_backup.def```
It should not be named "backup" but rather "default" or maybe "fallback". The wording should be changed in documentation also.
And why not ```enduser_setup.html.gz.def.c``` or whatever. to better see that it is connected to the html file.
Last the folder ```eus``` should be renamed to ```enduser_setup```. Whoever is not involved in enduser_setup cannot easily see the connection to ```eus```.

@marcelstoer what do you think about that? Should I change it while we are at it?

marcelstoer:
> The file `enduser_setup.html.gz` does not need to be checked in, it is only needed to create `http_html_backup.def`

True. Strictly speaking neither does the `.html`. We could reverse the `xxd` and then uncompress the data (provide a script in the repo for that) if we need to make changes.

> It should not be named "backup" but rather "default" or maybe "fallback". The wording should be changed in documentation also.

Both "default" and "fallback" sound good to me.

> And why not `enduser_setup.html.gz.def.c` or whatever. to better see that it is connected to the html file.

Yeah, it's odd to include a `.def` file: `#include "eus/http_html_backup.def"`

> Last the folder `eus` should be renamed to `enduser_setup`. Whoever is not involved in enduser_setup cannot easily see the connection to `eus`.

👍

app/modules/eus/enduser_setup.html
HHHartmann: all done

HHHartmann:
There is also a breaking change which should be mentioned in the release notes:
The /setwifi route has changed from GET to POST. So any individual adaption of ```enduser_setup.html``` will break.

marcelstoer:
> There is also a breaking change which should be mentioned in the release notes:

True, thanks for mentioning it again. I added a `breaking change` label we can add to PRs for such cases.

PR #2812, opened 2019-06-24
OTA support for ESP32

jmattsson:

Alright, here's functional OTA support for the ESP32. This module is agnostic to the method of obtaining the firmware - you can use http, tftp or even something running over [rfc1149](https://tools.ietf.org/html/rfc1149) if you so wish. This way, you can implement whichever authentication, authorization and verification scheme you desire.

Thanks to boot loader upgrades in IDF 3.3, the ESP32 now has proper rollback support, allowing for safe and convenient upgrade-and-test before committing to the new version. The old version is only a reset away until you've accepted the new application :)  Or, if you prefer you can live on the wild side without rollback support. The choice is yours!

This module is quite lightweight, as the IDF has already dealt with the heavy lifting.

### DIFF
--- a/docs/modules/otaupgrade.md
+++ b/docs/modules/otaupgrade.md
@@ -8,7 +8,10 @@ support, enabling new application firmware to be applied and booted into.
 
 This module is not concerned with where the new application comes from.
 The choice of download source and method (e.g. https, tftp) is left to
-the user.
+the user, as is the trigger to start an upgrade. A common approach is
+to have the device periodically check in with a central server and
+compare a provided version number with the currently running version,
+and if necessary kick off an upgrade.
 
 In order to use the `otaupgrade` module, there must exist at least two
 OTA partitions (type `app`, subtype `ota_0` / `ota_1`), as well as the
@@ -30,8 +33,8 @@ ota_1,    app,  ota_1, 0x140000,0x130000
 ```
 
 Depending on whether the installed boot loader has been built with or
-without rollback support, the upgrade process has four or three steps.
-Without rollback support, the steps are:
+without rollback support, the upgrade process itself has four or three
+steps. Without rollback support, the steps are:
 
 - `otaupgrade.commence()`
 - feed the new application image into `otaupgrade.write(data)` in chunks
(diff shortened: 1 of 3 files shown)


docs/modules/otaupgrade.md
```diff
@@ -0,0 +1,176 @@
...
+- feed the new application image into `otaupgrade.write(data)` in chunks
+- `otaupgrade.complete(1)` to finalise and reboot into the new application
+
+If the boot loader is built with rollback support, an extra step is needed
+after the new application has booted (and been tested to be "good", by
+whatever metric(s) the user chooses):
+
+- `otaupgrade.accept()` to mark this image as valid, and allow it to be
+  booted into again.
+
```
nwf: Is that to say that if the module reboots from the new firmware without accepting (and without paving over the old firmware) it will load the old version?  Does the IDF layer prevent a not-yet-accepted firmware from OTAing over the old firmware?

docs/modules/otaupgrade.md
```diff
@@ -0,0 +1,176 @@
...
+#### Returns
+`nil`
+
+A Lua error may be raised if the image does not pass validation, or no data
+has been written to the image at all.
+
+
+# otaupgrade.accept()
+
+When the installed boot loader is built with rollback support, a new
```
nwf: Is there a way to test if the bootloader supports rollback and if we're booting a non-accepted image?

docs/modules/otaupgrade.md
```diff
@@ -0,0 +1,176 @@
...
+Never. Either the system is rebooted, or a Lua error is raised (e.g. due
+to there being no other firmware to roll back to).
+
+
+# otaupgrade.info()
+
+The boot info and application state and version info can be queried with
+this function. Typically it will be used to check the version of the
+running application.
+
```
nwf: I was unfamiliar with the IDF's OTA support and so had a number of questions while reading the doc; perhaps it would make sense to front discussion of `otaupgrade.info()` and/or an example of what a typical OTA-ing application would do at boot?

docs/modules/otaupgrade.md
```diff
@@ -0,0 +1,176 @@
...
+#### Parameters
+None.
+
+#### Returns
+A list of three values:
+- the name of the partition of the running application
+- the name of the partition currently marked for boot next (typically the
+  same as the running application, but after `otaupgrade.complete()` it
+  may point to a new application partition.
+- a table with entries for each OTA partition, containing:
```
nwf: "a table whose keys are the names of OTA partitions and corresponding values are tables containing:" or such?

nwf:
Some nits in commentary but generally looks good to me. :)

docs/modules/otaupgrade.md
jmattsson: Yes, and yes. If you try to blatter over the old image before you've marked the new, running one as accepted, the `commence()` will fail.

docs/modules/otaupgrade.md
jmattsson: Good idea, can do!

docs/modules/otaupgrade.md
jmattsson: As in, move some of the `info()` notes up into the main discussion, and expand on that section?

nwf: I think I like what you've done in the latest commit. :)

jmattsson:
Alright, who wants to hit that gorgeous green "merge" button then? :)

devsaurus:
Oops, `mkdocs.yml` is missing a link.

jmattsson:
Sorry about that missing link - I keep doing that... :/

PR #2815, opened 2019-06-28
Improved crypto module for ESP32, now with HMAC

jmattsson:

This adds the `new_hmac()` feature as per the 8266 version to the crypto module, and optimises the crypto module a bit in the process.

Based on my testing, mbedtls pulls in all its algorithm regardless of whether the NodeMCU crypto module was using them or not. As such, the space savings from omitting algorithms were only in the tens of bytes (i.e. the table entries in the crypto module itself).

By switching to using the mbedtls generic message digest interface, the crypto module itself could be shrunk in size and complexity. Despite adding support for HMAC on all algorithms (plus including RIPEMD160), this version is 330 bytes smaller in my build.

Tagging @jpeletier as the module maintainer.

### DIFF
--- a/docs/modules/crypto.md
+++ b/docs/modules/crypto.md
@@ -1,30 +1,32 @@
 # crypto Module
 | Since  | Origin / Contributor  | Maintainer  | Source  |
 | :----- | :-------------------- | :---------- | :------ |
-| 2019-01-13 | [Javier Peletier](https://github.com/jpeletier) | [Javier Peletier](https://github.com/jpeletier) | [crypto.c](../../../components/modules/crypto.c)|
+| 2019-01-13 | [Javier Peletier](https://github.com/jpeletier), [Johny Mattsson](https://github.com/jmattsson) | [Javier Peletier](https://github.com/jpeletier) | [crypto.c](../../../components/modules/crypto.c)|
 
 The crypto module provides various functions for working with cryptographic algorithms.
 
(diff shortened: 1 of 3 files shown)


docs/modules/crypto.md
```diff
@@ -36,4 +38,34 @@ hashobj:update("FirstString"))
...
+- `algo` the hash algorithm to use, case insensitive string
+- `key` the signing key (may be a binary string)
+
+#### Returns
+Hasher object with `update` and `finalize` functions available.
+
+#### Example
+```lua
+hmac = crypto.new_hmac("SHA1", "top s3cr3t key")
+hmac:update("Some data"))
```
devsaurus: Extra `)` here and below.
Same for `crypto.new_hash()` btw.

jmattsson: Good catch. Fixed up both of them. That's what I get for copy/pasting stuff!

Feel free to merge when happy :)

jpeletier:
@jmattsson Sorry for not being able to chime in earlier. Thank you for this contribution. It is great you leveraged the multidigest capability of mbedtls. I see in its `config.h` that it enables all digest algorithms. Would it be possible to have this configurable as we had before?

jmattsson:
@jpeletier That `config.h` is under the control of the IDF, and it exposes no settings for the individual digests. Until/unless it does, there's no point in NodeMCU limiting the available digest mechanisms artificially - the code is linked in regardless.

PR #2817, opened 2019-07-03
Fix uart regressions & bugs

jmattsson:

Using `uart.on()` with a search character was broken in that it did not invoke the callback on a full UART buffer as documented. Logic reworked to match docs again.

Fixed memory leak on `task_post()` failure (eep!).

Improved logic to attempt to coalesce input bytes to reduce the number of `task_post()` slots used up by the platform uart.

Finally, added a semaphore to prevent the platform uart from overrunning the `task_post()` slots all the time on high baud rates (e.g. 1mbit). With the semaphore in there, the LVM RTOS task gets a chance to actually process the received data and free up a `task_post()` slot or two. The above mentioned read coalescing then allows the platform uart to immediately catch up.

Also added an error log message if the `task_post()` actually does fail.

### DIFF
--- a/components/platform/platform.c
+++ b/components/platform/platform.c
@@ -53,14 +53,13 @@ void uart_event_task( task_param_t param, task_prio_t prio ) {
   us = & uart_status[id];
   xSemaphoreGive(sem);
   if(post->type == PLATFORM_UART_EVENT_DATA) {
-    need_len = us->need_len;
-    end_char = us->end_char;
-
     for(size_t p = 0; p < post->size; p++) {
       ch = post->data[p];
       us->line_buffer[us->line_position] = ch;
       us->line_position++;
 
+      need_len = us->need_len;
+      end_char = us->end_char;
       size_t max_wanted =
         (end_char >= 0 && need_len == 0) ? LUA_MAXINPUT : need_len;
       bool at_end = (us->line_position >= max_wanted);
--- a/components/platform/platform.c
+++ b/components/platform/platform.c
@@ -5,7 +5,7 @@
 #include "driver/uart.h"
 #include <stdio.h>
 #include <string.h>
-
+#include <freertos/semphr.h>
 #include "lua.h"
 
 #include "esp_log.h"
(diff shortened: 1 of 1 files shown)


TerryE:
@jmattsson Johny, another case of planes passing in the night.  Sorry.   I have just finished implementing equivalent functionality and about to issue a `dev` PR for this.  I feel that part of your challenge here is that you are doing malloc / free the classic C / POSIX way and this just doesn't play nicely with the Lua VM, error handling and allocation system which are all nicely integrated.  Adopting the Lua way usually ends up with cleaner and more robust code. 

In my case I've implemented an object type of **Pipe** which has the normal pipe access methods but uses 256 byte Userdata chunks in a Lua array to implement the pipe.  Because everything is a Lua collectable object, once referenced either by normal descoping or by descoping through a thrown error, then the Lua GC handles all of the clean up automatically "for free".  I'll poke you through an @ prompt when I post this.

jmattsson:
I didn't think you were working on the ESP32 yet @TerryE?

TerryE:
> I didn't think you were working on the ESP32 yet @TerryE?

Johny, not directly.  My PR is for the ESP8266 and not the ESP32 so there is no issue of commit conflict here.  However, I am trying to lay the groundwork to do the unified lua53 code and so I am ending up tacking issues that I realise that you've already tacked in the RTOS implementation.   It just that with the last few years of Lua VM work I am approaching my implementation from a more Lua VM aligned perspective.

This doesn't impact what you are doing today to feed into your DiUS project timescales.

jmattsson:
As nice as the Lua memory handling is, it's not a good idea to attempt to use it from two separate RTOS threads, let alone two separate CPU cores at the same time. With the 32, a chunk of the work is simply marshalling things back into the LVM thread/context.

TerryE:
> As nice as the Lua memory handling is, it's not a good idea to attempt to use it from two separate RTOS threads, let alone two separate CPU cores at the same time.

I suppose the same argument could be used for the node.js kernel as this has pretty much the exact same non-blocking event-driven task architecture as NodeMCU -- except that node.js happily runs on a multicore CPU and the node.js webserver blows the socks off Apache / PHP in terms of performance / throughput.   But this is no reason to change current direction for now.  I am just saying that the issues that concern you could be fairly easily addressed, IMO.  However, this is for a longer discussion for another day, I think, and definitely part of or post lua53.

jmattsson:
@devsaurus It'll work even better after the second patch here. Changing one's mind about delimiters from within the callback now works even with data already in the buffer.

devsaurus:
Oh, I missed the update completely :+1:
@pjsg good to merge?

jmattsson:
I'm gonna merge this RSN unless someone objects.

TerryE:
> I'm gonna merge this RSN unless someone objects.

No objections from me.  This is the ESP32 branch and we've got a bunch of realignment work to do anyway when we do realign the two branches, so this makes no difference but does help ESP32 users in the meantime.

PR #2819, opened 2019-07-05
WIP: Rewrite sntp in Lua with only a little C

nwf:

This PR contains a new `sntp` module with almost all the state machine done in Lua and only the packet processing and fiddly large-integer math done in C.  It is intended to supplant #2700 and the in-C in-progress rewrite I mentioned therein.

<s>At present, it is exceptionally naive about time itself, even if its state machine endeavors to be more robustly implemented; @jmattsson's existing C displays a much more thorough understanding of time keeping than I possess.  It seems likely that we will need, and I am not opposed to adding, additional `userdata` structures for persistent state in the `sntppkt` C half.</s>

### DIFF
--- a/app/modules/sntppkt.c
+++ b/app/modules/sntppkt.c
@@ -1,0 +1,403 @@
+/*
+ * Copyright 2015 Dius Computing Pty Ltd. All rights reserved.
+ * Copyright 2020 Nathaniel Wesley Filardo. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * - Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the
+ *   distribution.
+ * - Neither the name of the copyright holders nor the names of
+ *   its contributors may be used to endorse or promote products derived
+ *   from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @author Johny Mattsson <jmattsson@dius.com.au>
+ * @author Nathaniel Wesley Filardo <nwfilardo@gmail.com>
+ */
+
+// Module for Simple Network Time Protocol (SNTP) packet processing;
+// see lua_modules/sntp/sntp.lua for the user-friendly bits of this.
+
+#include "module.h"
+#include "lauxlib.h"
+#include "os_type.h"
+#include "osapi.h"
+#include "lwip/udp.h"
+#include <stdlib.h>
+#include "user_modules.h"
+#include "lwip/dns.h"
+
+#define max(a,b) ((a < b) ? b : a)
+
+#define NTP_PORT 123
+#define NTP_ANYCAST_ADDR(dst)  IP4_ADDR(dst, 224, 0, 1, 1)
+
+#if 0
+# define sntppkt_dbg(...) printf(__VA_ARGS__)
+#else
+# define sntppkt_dbg(...)
+#endif
+
+typedef struct
+{
+  uint32_t sec;
+  uint32_t frac;
+} ntp_timestamp_t;
+
+static const uint32_t NTP_TO_UNIX_EPOCH = 2208988800ul;
+
+typedef struct
+{
+  uint8_t mode : 3;
+  uint8_t ver : 3;
+  uint8_t LI : 2;
+  uint8_t stratum;
+  uint8_t poll;
+  uint8_t precision;
+  uint32_t delta_r;
+  uint32_t epsilon_r;
+  uint32_t refid;
+  ntp_timestamp_t ref;
+  ntp_timestamp_t origin;
+  ntp_timestamp_t recv;
+  ntp_timestamp_t xmit;
+} __attribute__((packed)) ntp_frame_t;
+
+#define NTP_RESPONSE_METATABLE  "sntppkt.resp"
+typedef struct {
+    /* Copied from incoming packet */
+    uint32_t  delta_r;
+    uint32_t  epsilon_r;
+    uint8_t   LI;
+    uint8_t   stratum;
+
+    /* Computed as per RFC 5905; units are 2^(-32) seconds */
+    int64_t   theta;
+    int64_t   delta;
+
+    /* Local computation */
+    uint64_t  t_rx;
+    uint32_t  score;
+} ntp_response_t;
+
+static const uint32_t MICROSECONDS      = 1000000;
+
+static uint64_t
+sntppkt_div1m(uint64_t n) {
+  uint64_t q1 = (n >> 5) + (n >> 10);
+  uint64_t q2 = (n >> 12) + (q1 >> 1);
+  uint64_t q3 = (q2 >> 11) - (q2 >> 23);
+
+  uint64_t q = n + q1 + q2 - q3;
+
+  q = q >> 20;
+
+  // Ignore the error term -- it is measured in pico seconds
+  return q;
+}
+
+static uint32_t
+sntppkt_us_to_frac(uint64_t us) {
+  return sntppkt_div1m(us << 32);
+}
+
+static uint32_t
+sntppkt_long_s(uint64_t time) {
+  return time >> 32;
+}
+
+static uint32_t
+sntppkt_long_us(uint64_t time) {
+  return ((time & 0xFFFFFFFF) * MICROSECONDS) >> 32;
+}
+
+/*
+ * Convert sec/usec to a Lua string suitable for depositing into a SNTP packet
+ * buffer.  This is a little gross, but it's not the worst thing a C
+ * programmer's ever done, I'm sure.
+ */
+static int
+sntppkt_make_ts(lua_State *L) {
+  ntp_timestamp_t ts;
+
+  ts.sec  = htonl(luaL_checkinteger(L, 1) + NTP_TO_UNIX_EPOCH) ;
+
+  uint32_t usec = luaL_checkinteger(L, 2)  ;
+  ts.frac = htonl(sntppkt_us_to_frac(usec));
+
+  lua_pushlstring(L, (const char *)&ts, sizeof(ts));
+  return 1;
+}
+
+/*
+ * Convert ntp_timestamp_t to uint64_t
+ */
+static uint64_t
+sntppkt_ts2uint64(ntp_timestamp_t ts) {
+  return (((uint64_t)(ts.sec)) << 32) + (uint64_t)(ts.frac);
+}
+
+/*
+ * Process a SNTP packet as contained in a Lua string, given a cookie timestamp
+ * and local clock second*usecond pair.  Generates a ntp_response_t userdata
+ * for later processing or a string if the server is telling us to go away.
+ *
+ * :: string (packet)
+ * -> string (cookie)
+ * -> int (local clock, sec component)
+ * -> int (local clock, usec component)
+ * -> sntppkt.resp
+ *
+ */
+static int
+sntppkt_proc_pkt(lua_State *L) {
+  size_t       pkts_len;
+  ntp_frame_t  pktb;
+  const char  *pkts = lua_tolstring(L, 1, &pkts_len);
+
+  luaL_argcheck(L, pkts && pkts_len == sizeof(pktb), 1, "Bad packet");
+
+  // make sure we have an aligned copy to work from
+  memcpy (&pktb, pkts, sizeof(pktb));
+
+  uint32_t now_sec  = luaL_checkinteger(L, 3);
+  uint32_t now_usec = luaL_checkinteger(L, 4);
+
+  size_t           cookie_len;
+  ntp_timestamp_t *cookie = (ntp_timestamp_t*) lua_tolstring(L, 2, &cookie_len);
+
+  /* Bad *length* is clearly something bogus */
+  luaL_argcheck(L, cookie && cookie_len == sizeof(*cookie), 2, "Bad cookie");
+
+  /* But mismatching value might just be a packet caught in the crossfire */
+  if (memcmp((const char *)cookie, (const char *)&pktb.origin, sizeof (*cookie))) {
+    /* bad cookie; return nil */
+    return 0;
+  }
+
+  /* KOD?  Do this test *after* checking the cookie */
+  if (pktb.LI == 3) {
+    lua_pushlstring(L, (const char *)&pktb.refid, 4);
+    return 1;
+  }
+
+  ntp_response_t *ntpr = lua_newuserdata(L, sizeof(ntp_response_t));
+  luaL_getmetatable(L, NTP_RESPONSE_METATABLE);
+  lua_setmetatable(L, -2);
+
+  ntpr->LI        = pktb.LI;
+  ntpr->stratum   = pktb.stratum;
+
+  // NTP Short Format: 16 bit seconds, 16 bit fraction
+  ntpr->delta_r   = ntohl(pktb.delta_r);
+  ntpr->epsilon_r = ntohl(pktb.epsilon_r);
+
+  /* Heavy time lifting time */
+
+  // NTP Long Format: 32 bit seconds, 32 bit fraction
+  pktb.origin.sec  = ntohl(pktb.origin.sec);
+  pktb.origin.frac = ntohl(pktb.origin.frac);
+  pktb.recv.sec    = ntohl(pktb.recv.sec);
+  pktb.recv.frac   = ntohl(pktb.recv.frac);
+  pktb.xmit.sec    = ntohl(pktb.xmit.sec);
+  pktb.xmit.frac   = ntohl(pktb.xmit.frac);
+
+  uint64_t ntp_origin = sntppkt_ts2uint64(pktb.origin); // We sent it
+  uint64_t ntp_recv   = sntppkt_ts2uint64(pktb.recv);   // They got it
+  uint64_t ntp_xmit   = sntppkt_ts2uint64(pktb.xmit);   // They replied
+
+  // When we got it back (our clock)
+  uint64_t ntp_dest   = (((uint64_t) now_sec + NTP_TO_UNIX_EPOCH ) << 32)
+                        + sntppkt_us_to_frac(now_usec);
+
+  ntpr->t_rx = ntp_dest;
+
+              // |       outgoing offset           |   |        incoming offset         |
+  ntpr->theta = (int64_t)(ntp_recv - ntp_origin) / 2 + (int64_t)(ntp_xmit - ntp_dest) / 2;
+
+              // |       our clock delta           |   |       their clock delta        |
+  ntpr->delta = (int64_t)(ntp_dest - ntp_origin) / 2 + (int64_t)(ntp_xmit - ntp_recv) / 2;
+
+  /* Used by sntppkt_resp_pick; bias towards closer clocks */
+  ntpr->score = ntpr->delta_r * 2 + ntpr->delta;
+
+  sntppkt_dbg("SNTPPKT n_o=%llx n_r=%llx n_x=%llx n_d=%llx th=%llx d=%llx\n",
+    ntp_origin, ntp_recv, ntp_xmit, ntp_dest, ntpr->theta, ntpr->delta);
+
+  return 1;
+}
+
+/*
+ * Left-biased selector of a "preferred" NTP response.  Note that preference
+ * is rather subjective!
+ *
+ * Returns true iff we'd prefer the second response to the first.
+ *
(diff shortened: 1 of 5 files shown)


jmattsson:
I'll try to have a look late this week / next week!

TerryE:
Nathaniel,  If you don't mind, I'll give a general review rather than one anchored to PR lines, Though I will add a few specific anchored comments. 

-  I find the overall concept of using hybrid modules an extremely interesting one and one that is really built around utilising LFS.  This allows us to keep the C part minimalist and really private to the public Lua module which keeps the bulk of the login in a form that is accessible and extensible to the application programmer.  With LFS this can incur to RAM penalty and also allow the developer to move decisions about whether SNTP is needed to LFS image load.  I think that this is well worth exploring.

-  Is it worth making the C module `sntppkt` truly private (or not easily accessible from general Lua) because if it is only used by the "trusted" SNTP module then we can slim down on some of the validation and error checking.  One way of doing this would not to publicly document this C helper and also give it a non-compliant Lua variable name, e.g. `sntp-helper,c`.  The SNTP constructor could still look up this up using the  `ROM['sntp-helper,c']` access path and stash this in a shared upval, and also throw an explicit error "You must include the SNTP_HELPER module" in your build to use SNTP" if this is `nil`.

-  BTW, I strongly recommend recording all upvals used in each Lua function in its declaration e.g. with `-- upvals: x, y, z` as I do in my code, especially when said upvals are used RW within the routine and can therefore create side-effects.  You can find these by mining the `-l` listing.

- Is the SNTP object intended to be a singleton or not? (I see that it is currently written as mutli-instance -- that is the application can create multiple SNTP servers).   The reason that I ask this is that if it makes more sense to make it a singleton then the locals used as upvals / internal functions can be hoisted one level rather than including them inside `new()`.

lua_modules/sntp/sntp.lua
```diff
@@ -0,0 +1,226 @@
...
+
+  function doserver(err)
+    if _sat == MAX_SERVER_ATTEMPTS then
+      if fk then fk(err, serv[_six]) end
+      return nextServer()
+    end
+    _sat = _sat + 1
+    local tok = _tok
+    return _udp:dns(serv[_six], function(skt, ip)
+      if tok ~= _tok then
```
TerryE: This sort of multl-level _upval_ use  can be really inscrutable.  The `tok`, `_tok` and `_udp` are all upvals to the DBS CB function, and in the case of `_udp` this is already passed as a local parameter `skt` to the function and so this should be used instead.

lua_modules/sntp/sntp.lua
```diff
@@ -0,0 +1,226 @@
+local MAX_SERVER_ATTEMPTS = 2
+local SNTP_TIMEOUT        = 5000
+
+local defserv = {
```
TerryE: There is a good argument for making this JiT to avoid having the array permanently allocated and bound to the upval `defserv`.  E.g. 
```Lua
local function defserv()
 return {"0.nodemcu.pool.ntp.org",  "1.nodemcu.pool.ntp.org",  "2.nodemcu.pool.ntp.org", "3.nodemcu.pool.ntp.org"}
end
```
and using `serv = defserv()` below

app/modules/sntppkt.c
```diff
@@ -0,0 +1,315 @@
...
+static int
+sntppkt_read_resp(lua_State *L) {
+  ntp_response_t *r = luaL_checkudata(L, 1, "sntppkt.resp");
+
+  lua_createtable(L, 0, 6);
+
+  /* For large corrections, don't bother exposing fine values */
+  int d40 = r->delta >> 40;
+  if (d40 != 0 && d40 != -1) {
+    lua_pushnumber(L, r->delta >> 32);
```
TerryE: You use this pattern a lot.  Why not just have function which adds an integer field to the array at ToS?
```C
static void numuber_tofield(lua_State *L, const char * field_name, lua_Number value) {
  lua_pushnumber(L, value);
  lua_setfield(L, -2, field_name);
}
```
This will generate a little less code.

lua_modules/sntp/sntp.lua
nwf: Sorry, can you explain that to me in more detail?  In this case, why is it better to point at a thunk that will build the table than at the table itself?

lua_modules/sntp/sntp.lua
nwf: Sorry, I'll call out what's going on with `tok` and `_tok` and will replace `_udp` with `skt`.

nwf:
> Nathaniel, If you don't mind, I'll give a general review rather than one anchored to PR lines, Though I will add a few specific anchored comments.

Any and all commentary is more than welcome. :)

> * I find the overall concept of using hybrid modules an extremely interesting one and one that is really built around utilising LFS.  This allows us to keep the C part minimalist and really private to the public Lua module which keeps the bulk of the login in a form that is accessible and extensible to the application programmer.  With LFS this can incur to RAM penalty and also allow the developer to move decisions about whether SNTP is needed to LFS image load.  I think that this is well worth exploring.

I confess the intent was not so much to take advantage of LFS as it was to get rid of some hairy C, but yes, LFS certainly makes this kind of design much more attractive.  LFS really does change the programming landscape for the project.

> * Is it worth making the C module `sntppkt` truly private (or not easily accessible from general Lua) because if it is only used by the "trusted" SNTP module then we can slim down on some of the validation and error checking.  One way of doing this would not to publicly document this C helper and also give it a non-compliant Lua variable name, e.g. `sntp-helper,c`.  The SNTP constructor could still look up this up using the  `ROM['sntp-helper,c']` access path and stash this in a shared upval, and also throw an explicit error "You must include the SNTP_HELPER module" in your build to use SNTP" if this is `nil`.

Oh, that's an intriguing idea.  The `sntppkt` module is, at the moment, stateless, so there's presently no harm in exposing it, but I'm not planning on documenting it (heavily).  I'm glad to know that private-ish/name-mangled is possible; that might well be the right answer here.

> * BTW, I strongly recommend recording all upvals used in each Lua function in its declaration e.g. with `-- upvals: x, y, z` as I do in my code, especially when said upvals are used RW within the routine and can therefore create side-effects.  You can find these by mining the `-l` listing.

Good plan; will do.

> * Is the SNTP object intended to be a singleton or not? (I see that it is currently written as mutli-instance -- that is the application can create multiple SNTP servers).   The reason that I ask this is that if it makes more sense to make it a singleton then the locals used as upvals / internal functions can be hoisted one level rather than including them inside `new()`.

I think it is slightly simpler to have the application construct a new one and discard the old one if servers change, so I had not intended it to be a singleton.  That said, maybe there's not that much gain in simplicity and nobody's ever going to want more than one sntp client instance anyway.  It can go back to being a singleton like the existing one if there's desire.

lua_modules/sntp/sntp.lua
TerryE: What the original for does is to build a 4-entry table during the require evaluation and store it in the local `defserv` which is kept as a closed upval in the constructor.  The function version keeps a closure `defserv`  as a closed upval in the constructor. The table is built JiT. In practice, I suspect that most uses will use this default, so this isn't really a saving.  In terms of saving RAM, the best approach would be to move the default array inline.

lua_modules/sntp/sntp.lua
TerryE: No apology needed 😊

lua_modules/sntp/sntp.lua
nwf: Oh, I see.  OK, inline it shall go.

lua_modules/sntp/sntp.lua
nwf: Mm, I think `_tok` can go; it dates back to a time when I thought it might be worth holding on to the UDP socket between scans.  I no longer think that's likely worth doing, so it's gone, replaced with a comment.

marcelstoer:
> DNM: WIP: Rewrite

OT but still, what does "DNM" in this context stand for?

HHHartmann:
> * Is it worth making the C module `sntppkt` truly private (or not easily accessible from general Lua) because if it is only used by the "trusted" SNTP module then we can slim down on some of the validation and error checking.  ...

If we want to allow the Lua part to be extensible by the application programmer, we should not slim down on error checking too much. Also there is no guarantee that the user packs the matching version of the Lua module. So this should be able fail gracefully with a meaningful error message and not just "not work".

nwf:
The intent was to expose callbacks from the Lua layer much as is done now but to limit the direct extensibility of `sntp.lua` itself, again, just as we currently discourage modification of (or at least, require acknowlegement of being on one's own when modifying) `sntp.c`.

nwf:
I've spent a while more hacking on this and gotten it passing some basic functionality tests.  It would be good if someone (esp. @jmattsson) could review it, but I think I continue to like the feel of "hard bits in C, easy bits in Lua" development rather than "self-contained, complete C modules".

marcelstoer:
I don't feel I can add anything meaningful to the discussion here so maybe I better keep my mouth shut but I'm still tempted...sorry.

> I continue to like the feel of "hard bits in C, easy bits in Lua" development rather than "self-contained, complete C modules".

Is that the perspective of the firmware/module maintainer or that of the module user? As an end user having a self-contained C module available feels more convenient to me personally (unless one uses LFS maybe). However, I accept that there are different kinds of "end-users". Terry stated 

> ...private to the public Lua module which keeps the bulk of the login in a form that is accessible and extensible to the application programmer

and that's certainly a valid point. How many would want/need to extend that module, though?

nwf:
From the LFS-capable end-user perspective, the two are interchangeable.  Without LFS, it's hard to imagine NodeMCU as anything but a toy, and so I am not particularly worried about the heap bloat from the "easy bits in Lua" code.

I'm sorry to be harsh, but the quality of existing C modules suggests that very few developers, and I exclude myself, are capable of producing production-quality code.

pjsg:
I think the this project suffers from only having a few significant contributors and there is a lot of legacy code from the past. I'd like to get the sntp module cleaned up, but it does have quite a lot of functionality today, some of which I added to make the module relatively foolproof (i.e. call it once at startup and then it does the right thing).

nwf:
I will make the default success callback in this one DTRT, and will add a simple "just go" function to the module.

nwf:
OK, I've done a lot more polishing on this and I believe I'm ready to propose that it land.

In terms of longer future, I suggest that we leave the existing `sntp` C module around unaltered for the next master drop and only mark it deprecated in documentation, the drop after that can add the deprecation warning to source, and the drop after that can remove the module.  This should give us ample time to suss out any bugs with the fallback still around.  Does that seem fair?

TerryE:
@nwf, out of interest, if there any specific reason for us deciding not to use the SDK's built-in SNTP functions (`SDK/include/sntp.h`) and instead doing our own implementation?

nwf:
It looks like the SDK's `sntp` implementation is closed-source and buried inside `liblwip_536.a` which we don't link in because we build our own `liblwip.a` from source.

nwf:
[Previous incorrect comment deleted]

It looks like I removed the LwIP `sntp.c` in c972d86ea9f86794c0d43250f4bc99c3814e0e4c since we had the existing one, which was more aware of the native hardware RTC.  And I've replicated that RTC awareness here.  We could try adding Lua callbacks to the LwIP `sntp`?

app/modules/sntppkt.c
```diff
@@ -0,0 +1,421 @@
...
+ *
+ * :: string (packet)
+ * -> string (cookie)
+ * -> int (local clock, sec component)
+ * -> int (local clock, usec component)
+ * -> sntppkt.resp
+ *
+ */
+static int
+sntppkt_proc_pkt(lua_State *L) {
```
TerryE, line 174: The Lua module style tends to be a lot more terse.  For example:
```C
  size_t       pkts_len, cookie_len;

  const char  *pkts       = lua_tolstring(L, 1, &pkts_len);
  ntp_timestamp_t *cookie = (ntp_timestamp_t*) lua_tolstring(L, 2, &cookie_len);
  uint32_t     now_sec    = luaL_checkinteger(L, 3);
  uint32_t     now_usec   = luaL_checkinteger(L, 4)
  ntp_frame_t  pktb;
  
  luaL_argcheck(L, pkts && pkts_len == sizeof(*pkts)), 1, "Bad packet");
  memcpy(&pktb, pkts, sizeof(*pkts));
  
  luaL_argcheck(L, cookie && pkts_len == sizeof(*cookie))
                   !memcmp(cookie, &pktb.origin, sizeof (*cookie)),
                 , 1, "Bad cookie");
  
  ntp_response_t *ntpr = lua_newuserdata(L, sizeof(ntp_response_t));;
...
```
Also note:
-  Try to use `luaL_argcheck()` to check args.
-  You can hoist the userdata create to the declaration preamble.  This works because its being returned to the caller as a result is what prevents is GC.  If an error is thrown or it isn't returned then it is collected.
-  Use the standard library synonyms when possible, e.g. `memcpy()` instead of `os_memcpy()`
-  Pre SDK 3, the runtime included an eLua "optimisation" where string constants in Flash where stored as a `const char *` reference in the `TString`  instead of a byte sequence.  Hence something like `('12345'):sub(2)` would not be word aligned.  I removed this because in practice it rarely saved much memory and there was a per access cost of handling the variant.

app/modules/sntppkt.c
```diff
@@ -0,0 +1,421 @@
...
+ * @author Johny Mattsson <jmattsson@dius.com.au>
+ * @author Nathaniel Wesley Filardo <nwfilardo@gmail.com>
+ */
+
+// Module for Simple Network Time Protocol (SNTP) packet processing;
+// see lua_modules/sntp/sntp.lua for the user-friendly bits of this.
+
+#include "module.h"
+#include "lauxlib.h"
+#include "lmem.h"
```
TerryE, line 41: In general, I try to be absolutely minimal when it comes to `#include` directives. We should avoid any "l\*.h" except "lauxlib.h" (which includes "lau.h"), so no "lmem.h" and no "task/task.h".  "modules.h" already includes  "user_interface.h"

lua_modules/sntp/sntp.lua
```diff
@@ -0,0 +1,290 @@
...
+      "1.nodemcu.pool.ntp.org",
+      "2.nodemcu.pool.ntp.org",
+    }
+    local ni = net.ifinfo(0)
+    ni = ni and ni.dhcp
+    serv[1] = ni.ntp_server or "0.nodemcu.pool.ntp.org"
+  elseif type(serv) ~= "table"  then error "Bad server table"
+  end
+
+  if type(sc) ~= "function" and type(sc) ~= "lightfunction" then
```
TerryE, line 21: The type "lightfunction" no longer exists, so all of these clauses are redundant.  All functions are type "function". (This is for compatibility with Lua53)

lua_modules/sntp/sntp.lua
```diff
@@ -0,0 +1,290 @@
...
+  if now ~= nil and type(now) ~= "function" and type(now) ~= "lightfunction" then
+    error "Bad clock type"
+  end
+  now = now or (rtctime and rtctime.get)
+  if now == nil then error "Need clock function" end
+
+  local self = {}
+  self.servers = serv
+
+  local _tmr -- contains the currently running timer, if any
```
TerryE, line 36: You clearly are following some convention so prefixing some -- but not all -- local names with "_".  What is this convention?

lua_modules/sntp/sntp.lua
```diff
@@ -0,0 +1,290 @@
...
+  local _res -- the best result we've got so far this pass
+  local _best -- best server this pass, for updating _pbest
+
+  local _six -- index of the server in serv to whom we are speaking
+  local _sat -- number of times we've tried to reach this server
+
+  -- Shut down the state machine
+  --
+  -- upvals: _tmr, _udp, _six, _sat, _res, _best
+  local function _stop()
```
TerryE, line 51: The destructor in general the cleanup needed.  Also list assignments are useful.  This routine body could have been encoded
```Lua
    _tmr = _tmr and _tmr:unregister()
    _udp = _udp and _udp:close()
    _six,_sat,_res, _best = nil
    for k,v in pairs(_kod) do _kod[k] = (v > 0) and (v - 1) or nil end
```

lua_modules/sntp/sntp.lua
```diff
@@ -0,0 +1,290 @@
...
+  if fc ~= nil and type(fc) ~= "function" and type(fc) ~= "lightfunction" then
+    error "Bad failure callback type"
+  end
+  if now ~= nil and type(now) ~= "function" and type(now) ~= "lightfunction" then
+    error "Bad clock type"
+  end
+  now = now or (rtctime and rtctime.get)
+  if now == nil then error "Need clock function" end
+
+  local self = {}
```
TerryE, line 33: Incidentally why not `local self = {servers = serv}` ?

I more general point given that self is a module local, this is a singleton class  when I am not sure this is the intent.  Shouldn't self be local to `new()` so 
```Lua
local sntp = require'sntp'.open(a,b,c,d) works as expected
```
Also in general the object should be the first param so the ":" form of method invocation works as expected.

TerryE (commented):
See also my comment about why we are reimplementing built-in SDK functionality.

TerryE:
> It looks like I removed the LwIP `sntp.c` in [c972d86](https://github.com/nodemcu/nodemcu-firmware/commit/c972d86ea9f86794c0d43250f4bc99c3814e0e4c) since we had the existing one, which was more aware of the native hardware RTC. And I've replicated that RTC awareness here. We could try adding Lua callbacks to the LwIP `sntp`?

We might have a perfectly good reason, but it was still worth asking the Q, I think.  I think that Johny's original SMTP module predated Espressif's addition of its own.

lua_modules/sntp/sntp.lua
nwf: It is not particularly meaningful any more; the code's been through a lot of refactoring.

lua_modules/sntp/sntp.lua
nwf: `self` *is* local to `new`?  The only module locals are `MAX_SERVER_ATTEMPTS`, `SNTP_TIMEOUT`, `new`, `update_rtc`, and `go`.  Or am I missing something?

While I generally understand the `:` calling convention, I don't see its relevance here?  Neither of the exported functions on the "SNTP object" created by `new` (i.e., `sync()` and `stop()`) take arguments, as all their associated state is in their upvals.  The callbacks are held as upvals not as table members.

app/modules/sntppkt.c
nwf: AIUI

> This works because its being returned to the caller as a result is what prevents is GC.

is not quite right.  What's preventing the UD handle from getting GC'd is the existence of a reference in the registry.

By contrast, were `sntppkt_proc_pkt` to call up into Lua after its call to `lua_newuserdata` (and lack of `luaL_ref`) and encounter an exception that `longjmp()`-ed back over `sntppkt_proc_pkt`, say, just before returning back to Lua... the reason the UD would be eventually GC'd, despite having been returned to C and had $DEITY-knows-what done to it, is that there are no more references from the Lua side.

lua_modules/sntp/sntp.lua
TerryE: > `_udp:close()` does not unregister callbacks

No but _udp:__gc() does and as `_udp:close()` returns `nil` this clears derefences the socket which triggers the proper GC including clearing down the registry entries.

But there is really a separate `net` tidy up issue that closing a socket should automatically unref any CBs in it.

nwf: Right, but we won't get to `_udp:__gc()` if there are knots tied through the registry.  Again, I don't mean for there to be, here, but it seemed generally like a belt&suspenders kind of approach.

And yes, as per #3062, I think `net` needs some tidying.  If you think the contract is that `:close()` should rip all callbacks out, it would be good to a note to that effect that there.  I have generally shied away from that particular option because the object outlives `:close()` and can, in principle, be pressed back into service.  At least, the TCP objects can; UDP is a little harder to justify.

lua_modules/sntp/sntp.lua
TerryE: Sorry, yes self is as you say, but I got a bit lost in the depth of semantic nesting.  For others, a call `v:name(args)` is syntactic sugar for `v.name(v,args)`, except that this takes less VM instructions and v is evaluated only once.

Here is the size / upvalue analysis of the module (hope you find this useful).
```
main     <L0,0>     (15 ins, 5 const)     
function <L6,201>   (145 ins, 25 const) -- upvals:  SNTP_TIMEOUT, MAX_SERVER_ATTEMPTS
function <L45,63>   (50 ins, 9 const)   -- upvals:  _tmr, _six, _sat, _res, _best, _udp, _kod
function <L71,143>  (42 ins, 10 const)  -- upvals: _tmr, SNTP_TIMEOUT, _udp, doserver, now, _kod, fc, serv, _six, _self, nextServer,
                                                    _pbest, _res, _best
function <L72,76>   (15 ins, 4 const)   -- upvals: _udp, doserver
function <L82,137>  (142 ins, 21 const) -- upvals: ip, txts, now, _kod, fc, serv, _six, _self, nextServer, _pbest, _res, _best, _tmr
function <L146,157> (31 ins, 2 const)   -- upvals: _sat, MAX_SERVER_ATTEMPTS, fc, serv, _six, _self, nextServer, _udp, doserver, hail
function <L153,156> (16 ins, 3 const)   -- upvals: doserver, hail
function <L162,179> (45 ins, 4 const)   -- upvals: _six, serv, _res, _pbest, _best, _stop, sc, _self, fc, _sat, doserver
function <L184,191> (18 ins, 6 const)   -- upvals: _stop, _udp, _tmr, _six, nextServer
function <L193,197> (11 ins, 1 const)   -- upvals: _res, _best, _stop
function <L204,246> (58 ins, 16 const)             
function <L251,273> (26 ins, 6 const)   -- upvals: new, update_rtc
function <L252,252> (1 ins, 0 const) 
function <L254,266> (40 ins, 9 const)   -- upvals: update_rtc, freq, sc
function <L270,270> (6 ins, 2 const)    -- upvals: sntpobj
```
This is a pretty chunky piece of code that takes 10kB RAM to load if you are using LFS.  That's a big enough chunk to think about how we could optimise this.  Have a look at how I've done this for the FTP server.

nwf: I assume that's supposed to be "10kB RAM if you're **not** using LFS"?  If this is 10K even with LFS then I really am in trouble.  I'll see what I can do to drive it down all the same.  (I continue to vaguely wish that Lua had a `const` keyword.)

Again, I know what `:` does.  I know it's syntactic sugar and I understand the VM instructions it expands into.  What I don't understand is why you bring it up.  Am I `foo.bar(foo,...)`-ing somewhere in here?  Am I failing to provide an interface that looks like it should support that?

TerryE: > I assume that's supposed to be "10kB RAM if you're **not** using LFS"?

Correct.  The static RAM usage if loaded into LFS is zero, though executing code does create runtime variables.  Lua constants are just that and are stored as TValues in the code space (and which is why I listed them above).   Each instruction takes 4 bytes, each TValue takes 12 bytes in Lua51 and 8 bytes in Lua53.  `Proto` structures and (debug) `lineinfo` vectors also take code space, though the NodeMCU lineinfo is typically 10-20× more dense than in standard Lua).  String constants are a TValue pointing to a ROstrt entry.  

Unlike PHP or Javascript, Lua has no concept of an array constant; arrays are constructed in RAM at runtime, though the opcodes which do this construction are instruction that can be in LFS.  The only exception to this was introduced by eLua and built upon in NodeMCU; this is the ROTable construct, but these are declared statically in C code as `const` structures and mapped into Flash memory.

HHHartmann:
Sorry, trying to trigger a CI build

nwf:
CI builds are great, but could you please use a new PR that won't send me email? :)

HHHartmann:
Sure, sorry to annoy you, I was just on the way to give several PRs the chance to run against the new tests.

PR #2826, opened 2019-07-10
Video walk through of build process

chilipeppr:
I thought this would be super helpful to the community to link to this video since there are so many dependencies you need in Linux to get this going. It was very confusing to me initially and this video summarizes the whole process to get a greater community going.

- [x ] This PR is for the `dev` branch rather than for `master`.
- [x ] I have thoroughly tested my contribution.
- [x ] The code changes are reflected in the documentation at `docs/*`.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -37,6 +37,9 @@ git submodule init #only if repo was cloned w/o submodules init
 git submodule update --recursive
 ```
 
+Here is a video walk through by John Lauer (ChiliPeppr) of building the firmware in Linux from scratch with a fresh install of Ubuntu 19 so you can see all of the dependencies needed to get your build completed and flashed to your ESP32 device.
+[![Video walk through for Linux Build Environment](https://img.youtube.com/vi/x6CGECsioYg/0.jpg)](https://www.youtube.com/watch?v=x6CGECsioYg "Video walk through for Linux Build Environment")
+
 ## Build Options
 
 All configuration options are accessed from the file `sdkconfig`. It's advisable to set it up with the interactive `make menuconfig` - on a fresh checkout you're prompted to run through it by default.
--- a/docs/build.md
+++ b/docs/build.md
@@ -38,6 +38,7 @@ git submodule update --recursive
 ```
 
 Here is a video walk through by John Lauer (ChiliPeppr) of building the firmware in Linux from scratch with a fresh install of Ubuntu 19 so you can see all of the dependencies needed to get your build completed and flashed to your ESP32 device.
+
 [![Video walk through for Linux Build Environment](https://img.youtube.com/vi/x6CGECsioYg/0.jpg)](https://www.youtube.com/watch?v=x6CGECsioYg "Video walk through for Linux Build Environment")
 
 ## Build Options


TerryE:
John, I really applaud the idea of making YouTube walkthroughs and tutorials to make the whole process more accessible. So a definite :+1:  for this. 

- I found the impromptu process quite informative but I understand all the steps, however I suspect that the typical app developer would get lost in this. Perhaps it would have been better to do a dress rehearsal and the just say on the video you need the following components to run the IDE and do a single `sudo apt-get install` and `sudo pip` before doing the make?

-  I haven't really used WinX for over 10 years (just Linux and Android these days), so the Win10 interoperability side is weak for me, but AFAIK, you've got a few options for the POSIX build environment, e.g. Virtualization through VBox, VMware or M$; WSL; and Docker.  Perhaps a brief overview on the fact that you have to build the firmware in Linux, there are these options, but your preference is VBox so this tutorial is based on it.
-  Ditto the whole USB side could be a little more polished.

marcelstoer:
GitHub renders this a bit oddly, haven't tested MkDocs/RTD though. I think you need a blank line after the paragraph.

Screenshot 2019-07-10 at 10 14 38

chilipeppr:
Just added the blank line after the paragraph to fix that dangling text/image issue.

On the impromptu nature of the video, I agree that it would have been nice to have a full summary "apt install" command, but depending on your version of Linux you could have quite the alternate list of dependencies. So, for example, when I did this on Ubuntu 18 around 6 months ago, I recall having to uninstall/reinstall make because it was too old of a version and it being painful to find the right version because it was not in the apt-get repo yet.  So, I think the impromptu nature is how folks do it in the real world like I do. In a way, the errors built into the code, are the docs of the dependencies required, you just have to find them one by one. I really don't know if other flavors of Linux will have totally different dependencies. I recall recently following somebody's massive "apt-get install" list for opencv and it got outdated fast and I ended up iterating through 10 more errors of other dependencies beyond what the docs had recorded. 

On your other environments suggestion, I have tried Docker and found the USB serial port issue to not work so you can't use esptool from it. I found a bunch of other stuff to not work either so I gave up on Docker. If somebody had posted a Youtube vid on how they do it, I may have found success using that approach instead because I do think full virtual machines are heavy these days. So I suppose I'm advocating for the ad-hoc nature of my video because it'll help others not feel intimidated of how to get the build to succeed, like I was intimidated initially.

I would love to make some more Youtube videos because another area I was intimidated was building a new component/module and doing a pull request for it. I'd like to show folks how to do it, but I think the only way I'd get to it is to do it impromptu and then just try to edit out the slow parts like I did in this one.

marcelstoer:
> I have tried Docker and found the USB serial port issue to not work so you can't use esptool from it. 

That shouldn't actually be an issue. The image is built with the intention to be (only) used for compiling the firmware. The container is normally started ad-hoc and then stopped and destroyed after the compilation has finished. So you would use esptool from your host OS. Of course you _can_ keep the container around if you like. I didn't document that though and I never checked whether the image contains all the USB stuff.

> I found a bunch of other stuff to not work either so I gave up on Docker.

I, and possibly @HHHartmann (who contributed greatly), would definitely be interested in learning about that other stuff. Please raise any issues you find over at https://github.com/marcelstoer/docker-nodemcu-build/. The only one that really bothers me that I'm aware of is https://github.com/marcelstoer/docker-nodemcu-build/issues/64.

marcelstoer:
Oh sorry Terry, I only just saw your comments now :frow Feel free to push that 'Revert' button if you're unhappy enough with the video.

chilipeppr:
Ahh. Yeah, that makes sense to not actually do esptool from Docker. Hmm,
that means an even more complex install because I usually don't run Python
on my Windows box. I suppose another video on running the Docker approach
could be in the future here. Maybe I'll give that a go at some point and
see if I can change my workflow to Docker. Now that I understand the
NodeMCU build process more, it may give me enough to figure out Docker.

BTW, partly why I did this video is that a user asked in the ChiliPeppr
forums how to compile in MQTT. I didn't have that in any of my prebuilt
binaries. They tried to find a cloud build for ESP32 like you have for
ESP8266 and couldn't find it. I tried finding a cloud builder a year ago
too because the ESP8266 one is so bad-ass, so I understand his pain.
Anyway, because he was so frustrated he was considering moving to
MicroPython instead of NodeMCU, so I figured a video could help him get
through his frustration.

marcelstoer:
> that means an even more complex install because I usually don't run Python
on my Windows box.

If you don't mind not having everything automated (i.e. scripted) you can run my NodeMCU PyFlasher GUI application. It's self-contained, no installation and no Python required.

jmd13391:
> I found a bunch of other stuff to not work either so I gave up on Docker.

I must be missing something here... Maybe it's that you are talking ESP32 and I am using ESP8266...

In any case, my 2-cents: I am using Marcel's Docker environment under Windows and find the app build process to be extremely easy, quick and error free.

TerryE:
The docker use with NodeMCU is really an incremental adaption of its use on Travis, but this isn't Docker's main target, which is to act as a component provisioning system that totally decouples components.  For windows users, so long as they have docker install in a VM or WSL then this gives a fairly pain-free  way of installing standard components in a Lux environment on Windows.  There is absolutely no reason why the docker container needs to be run in the batch amnesia mode that is what Travis wants.  Containers can be interactive, can include persistent directory hierarchies and can connect to host devices such as a USB connected UART.  So there is absolutely nothing stopping us developing an IDF container or a esptool container or whatever, other than having a suitable resource with the time and skills to do the work.  I could do this but I am already booked out with other NodeMCU commitments for the next six months at least.

TerryE:
@chilipeppr John, this is now a pure community project so nearly all contributions here are pro-bono from individual contributors (with the notable exception of DiUS who use NodeMCU for their business and in return freely return their contributions to the community).  We are always willing to explore ways of aligning our goals to other active developers who can contribute to the commonwealth.  So thanks. :smile:

chilipeppr:
Hmm. Based on this thread I tried to give Docker a go again this morning because I don't totally recall why I gave up on it. After jumping through lots of hoops getting Docker going again on Windows this morning, here's the error I get during build and I can't seem to figure out why.

Here's the build command I'm running:

image

Here's the error:

image

I have the latest repo, etc.

PR #2829, opened 2019-07-12
Fix net module data loss & RTOS task unsafety

### LINKED ISSUE #2824: esp32 net module loses TCP data at times
### Expected behavior
Data in a TCP connection should not be lost.

### Actual behavior
Occasionally, when receiving lots of data, some data goes missing from within the TCP session. This can be seen for example when using the OTA example for the `otaupgrade` module. At times the image validation fails, and the image size reported does not match the size of the image actually sent to the device.

Looking at the `net` module, there appears to be a possibility of losing events due to failing `task_post()` calls. Additionally, we have a bunch of thread-unsafe accesses to the userdata from the lwIP RTOS thread - any lwIP callback runs in the lwIP thread, and ideally it should not touch a Lua userdata what-so-ever, but if it absolutely has to, that would need to be properly mutex guarded.

I'll have a look at it and see if I can fix it similarly to the recent uart fixes.

### NodeMCU version
dev-esp32

### Hardware
n/a

jmattsson:
Fixes #2824 (second, larger half of the fix).

To avoid races between the lwIP callbacks (lwIP RTOS task) and the Lua
handlers (LVM RTOS task), the data flow and ownership has been simplified
and cleaned up.

lwIP callbacks now have no visibility of the userdata struct. They are
limited to creating small event objects and task_post()ing them over
to the LVM "thread", passing ownership in doing so. The shared identifier
then becomes the struct netconn*.

On the LVM side, we keep a linked list of active userdata objects. This
allows us to retrieve the correct userdata when we get an event with
a netconn pointer. Because this list is only ever used within the LVM
task, no locking is necessary.

The old approach of stashing a userdata pointer into the 'socket' field
on the netconn has been removed entirely, as this was both not
thread/RTOS-task safe, and also interfered with the IDFs internal use
of the socket field (even when using only the netconn layer). As an
added benefit, this removed the need for all the SYS_ARCH_PROTECT()
locking stuff.

The need to track receive events before the corresponding userdata object
has been established has been removed by virtue of not reordering the
"accept" and the "recv" events any more (previously accepts were posted
with medium priority, while the receives where high priority, leading
to the observed reordering and associated headaches).

The workaround for IDF issue 784 has been removed as it is now not needed
and is in fact directly harmful as it results in a double-free. Yay for
getting rid of old workarounds!

DNS resolution code paths were merged for the two instances of "socket"
initiated resolves (connect/dns functions).

Also fixed an instance of using a stack variable for receiving the resolved
IP address, with said variable going out of scope before the DNS resolution
necessarily completed (hello, memory corruption!).

Where possible, moved to use the Lua allocator rather than plain malloc.

Finally, the NodeMCU task posting mechanism got a polish and an adjustment.
Given all the Bad(tm) that tends to happen if something fails task posting,
I went through a couple of iterations on how to avoid that. Alas, the
preferred solution of blocking non-LVM RTOS tasks until a slot is free
turned out to not be viable, as this easily resulted in deadlocks with the
lwIP stack. After much deliberation I settled on increasing the number of
available queue slots for the task_post() mechanism, but in the interest
of user control also now made it user configurable via Kconfig.

### DIFF
--- a/components/base_nodemcu/user_main.c
+++ b/components/base_nodemcu/user_main.c
@@ -146,6 +146,8 @@ void nodemcu_init(void)
 
 void app_main (void)
 {
+  task_init();
+
   esp_event_queue =
     xQueueCreate (CONFIG_SYSTEM_EVENT_QUEUE_SIZE, sizeof (system_event_t));
   esp_event_task = task_get_id (handle_esp_event);
--- a/components/modules/net.c
+++ b/components/modules/net.c
@@ -14,9 +14,12 @@
 #include "lwip/api.h"
 #include "lwip/err.h"
 #include "lwip/ip_addr.h"
-#include "lwip/dns.h" 
+#include "lwip/dns.h"
 #include "lwip/tcp.h"
+#include "lwip/inet.h"
+#include "lwip/igmp.h"
 
+#include <esp_log.h>
 
 // Some LWIP macros cause complaints with ptr NULL checks, so shut them off :(
 #pragma GCC diagnostic ignored "-Waddress"
@@ -27,9 +30,7 @@ typedef enum net_type {
   TYPE_UDP_SOCKET
 } net_type;
 
-typedef const char net_table_name[14];
-
-static const net_table_name NET_TABLES[] = {
+static const char *NET_TABLES[] = {
   "net.tcpserver",
   "net.tcpsocket",
   "net.udpsocket"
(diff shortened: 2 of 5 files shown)


jmattsson:
Unless I screwed up somewhere, this should be the last of the fixes to the net module. As is, I'm not aware of any further outstanding issues with it that need fixing.

Issue #2823, opened 2019-07-07
Code logic flaw in `app/driver/rotary,c


TerryE:
As part of my pre lua53 cleanup , I am adding `-Wall` to the lua and drivers directories.  In the case of the latter this has thrown up a bug in `rotary.c`.  The macro at L40
```C
#define STATUS_IS_PRESSED(x) ((x & 0x80000000) != 0)
``` 
has the parentheses around `x` in the expansion missing which means that the use in L175  generates
```C
|| (last_status ^ new_status & 0x80000000) != 0)
```
which is not the same as the intended
```C
|| ((last_status ^ new_status) & 0x80000000) != 0)
```
because the **&** operator has a higher precedence than **^**.  Ditto L176.

I've added the extra guard parenthesis for my cleanup so my next PR will fix this, but perhaps @pjsg could confirm that this is correct.

pjsg:
You are entirely correct -- there should be parens around the x. However, it doesn't actually make much difference. In the driver, this is used to determine whether it is worth waking up the task to handle a message. The module layer gets it correct (I think). Thus there is no functional impact, but there is a performance impact.

Thanks for fixing.

TerryE:
I picked up a couple of these in my own code when I turned on -Wall :stuck_out_tongue_closed_eyes:

Issue #2811, opened 2019-06-23
RFC Why have different coding standards for ESP32 and EP8266 modules?


TerryE:
This is really one where I am looking for views from Arnim and Johny.  As you should be aware, I am going through the process of preparing for the Lua53 migration, and one of my main derisking strategies is that I an _not_ going to add new `dev` and `master`  branches for Lua 5.3 because everything apart from the `lua` and `lua53` sub-components will be common.

I am also review how this will carry across to porting Lua53 to the ESP32 environment, and really find myself asking the Q:

- _Why are the coding standard / APIs different for the two variant platforms?_

I can see that it was easier just to fork and patch when working to a deadline.  I can also understand why modules might have to be different when they depend on different hardware features.  With a little bit of effort we could have aligned the coding standards and used a little bit of define magic to hide virtually all of the API and coding differences.

The fork and patch lowers the upfront effort, but though-life it just causes the project to bleed because we are having to divide or limited resources between the two code variants: there is a conversion overhead every time we have to migrate a module across platforms, and we can't just cherry-pick fixes from one platform to the other.

For example:
-  ESP32 code follows the `<stdint.h>` conventions for the various integer definitions (`uint32_t` etc.) wheres the ESP8266 code follows the SDK "c_types.h" conventions (`uint32` _without the `_t` suffix_)
- ESP32 code follows the `<sting.h>` conventions for the string and memory operation (`stcpy()` etc.) whereas the ESP8266 code again follows the SDK convention of using the `c_` prefix.
-  There are other minor API changes that have been made in ESP32.

Let me be clear: I support the choice of using POSIX-compliant + C99 coding standards whenever practical, so I don't disagree with what you've done for the `dev-esp32` branch.  However, my assertion is the it would be better in the long run if all of code including that for the ESP8266 adopted this strategy.
    
I had to solve most of these issues in order to build the cross compiler on the host platform.  Surely, it is worth the effort to converge all of this?

As a first step, I can take on the job of reworking the modules.  I also suggest that 
-  I rename `app` to `components`.
-  We spilt the `modules` into three sub-directories:
   -  **core_modules**. Those modules that are in essence mandatory for any firmware build and their omission isn't really supported.
   -  **common_modules**.  Those modules that should (in principle) work on both platforms, and where any platform-specific variations are handled through `#ifdef` logic.
   -  **esp8266_modules**.  Those modules that have ESP8266 architectural dependencies that means that the ESP32 implementation must be different.

[...]

Issue #2808, opened 2019-06-21
Impacts of migrating to Lua 5.3


TerryE:
I have done a roll up of the changes in moving from Lua 5.1 (referred to as **Lua51**) to Lua 5.3 (**Lua53**). (This combines any changes introduced in Lua 5.2 and also removes the points that aren't relevant to NodeMCU implementation (e.g. `os` and `io` stuff and unimplemented `math` and `debug` components.)  Broadly these changes fall into three broad categories:

- **Changes in the Lua language itself**.  The biggy here  is the introduction of an integer subtype for numbers which makes the need to separate integer and float builds redundant.

- **Changes in the Lua libraries**.  These might break existing examples and user applications, but for the majority of users these won't impact their applications.  In general they are pretty easy to code around.  For example `loadstring()` functionality has been moved into `load()` so replacing `loadstring` by `(loadstring or load)` means the Lua code works under 5.1 and 5.3.

-  **Changes in the C API**.  These might break existing C library modules.  An example here is that all of the different signed and unsigned integer variants of the API calls have been dropped in favour of the core set which return `lua_Integer`.  Type casting and conversion is left to the application program.  So no more `luaL_checkint()` or `luaL_checklong()` only `luaL_checkinteger()` which returns a  `lua_Integer`.

Clearly I will deal with all of the core stuff (the `app/lua` to `app/lua53` directories and core NodeMCU modules such as `node`, `file`, etc.) as part of the PR.

Having done a scan of the `lua_examples` and `lua_modules` directories there are very few cases where moving to Lua5.3 will break the application, and I can put through a bulk edit so that the source is Lua53 compatible (such as the above `loadstring` use).  _However, I can't commit to testing all of these changes,_ but they should just work out of the box.

I can also do a similar bulk edit so that the `app/modules` directly compiles cleanly on both build variants, but again I just don't have the bandwidth to test every module.

So as I see it, given the reality that the other committers will spend little if any effort in supporting this upgrade, we have two alternatives:

1. **Give up now because it involves just too much effort which we can't resource.**
2. **Take an optimistic migration strategy**.  In this one, we just make the changes after a review window, accepting the risk that module X might just exhibit a bug that we need to fix downstream. (As an example we had a previous example where the integer 0x80000000 generated different code for `luac.cross` and `node.compile()` because of these integer edge effects.  We might get similar issues moving from 5.1 to 5.3.)

[...]

Issue #2803, opened 2019-06-20
RFC on Tasking Inside our NodeMCU Framwork


TerryE:
### The NodeMCU architecture in essence.

NodeMCU works broadly the same as `Node.js` \([here](https://www.journaldev.com/7462/node-js-architecture-single-threaded-event-loop) is a good overview).  On the ESP variants (RTOS for the ESP32 or the non-OS SDK), a Lua application is composed of a set of tasks organised and scheduled through a Single Threaded Event Loop Scheduler.  

Each task typically has a thin C initiator which then calls a Lua function that may call other Lua functions in turn, but then the whole runs to completion.  The event scheduler will then start the next task ready to run based on FIFO within priority.  The whole framework is based on the rule that individual tasks run to completion and are not interrupted by other tasks, so the system as a whole can be implemented in a single processing thread.  For this to work, tasks should be short, sharp and non-blocking.  Each tasks is typically initiated based on an external event: a timer has fired, a GPIO has been set, a network packet has arrived, and so these are referred to as callbacks in SDK terminology.  Because the Lua VM only executes one task at a time we don't need mutexes or other fancy task synchronisation mechanisms.  Multi-tasking is cooperative: a task yields control by terminating.

In a typical well-written ESP Lua application, most task are short and execute within a few milliseconds so the ESP processor can complete 100s of tasks a second with minimal overhead, making it and NodeMCU really well suited to embedded IoT applications.

So a typical implementation pattern for a task is that is comprises:

- A initiator coded in C which is scheduled in response to an external event.  For example a network socket event, such as receiving a TCP packet, invokes the routine `net_recv_cb()` which then decodes the event and decides which Lua action function to execute.  

- There is typically a 1-1 association with a Lua-callable booking function which can book such events and associate the correct Lua function with the event occurring, in this case `net_on('receive',func)`.

Because each task exits from a Lua VM perspective, that is the Lua call stack unrolls entirely, the only Lua variables that are preserved from task-to-task are stored in the Lua environment (`_G`) and in the Lua Registry or their direct children.  The Lua GC will collect all local variables created and released during the task execution.

Because Lua task functions must persist from task to task, this are all stored in the Lua Registry and referenced using an integer handle. The booking function will use the `luaL_ref()` API to allocate this registry slot and obtain the handle, and then the event routine will retrieve the task function using this handle and then call `luaL_unref()` to return the used slot to the pool, before executing a `lua_call()` to execute the Lua task.

[...]

PR #2836, opened 2019-07-18
Lua 5.1 to 5.3 realignement phase 1

TerryE:
See #2033, #2787 #2802 #2803 #2808 and #2823

### Overview

My aim here is to make the `app/lua53` interface with the platform and other libraries as clean as possible, and since all of the other directories are going to be shared, it is just simpler to backport this tidying into the `app/lua` code hierarchy.  This PR makes the first tranche of these changes to achieve this alignment.

This has ended up  pretty large change and it does change the feel of the Lua interpreter:
-  You can now paste 100s of lines into either the UART or telnet interface without data overrun.  (The 4Kb string limit hasn't been changed.)
-  Error reporting is more robust and works with telnet.  The hook to remove CB panics is include (but not yet deployed in all modules)
-  Telnet works as expected with the VM facilitating the necessary field marshalling

### Detailed changes

-  `lua/lua.c`.  Strip out all dead code left over from the eLua and core versions but not used in NodeMCU (e.g standard Lua supports command line arguments but we don't use them so there's no point in having all of these argument decoding functions.  Tidy up how startup and how the input loop is handled using the task interface.  (Pasting bulk content into Putty etc. not longer overruns the input buffer).

- `app/lua/lnodemcu.c`. The NodeMCU extensions to the Lua API are now located in 3 API files: `lflash.c` for LFS (`lauN_*()` calls),  `lrotables.c` for ROTables (`lauF_*()` calls) and `lnodemcu.c` for other extension (for other `lauN_*()` calls).  The current items in this last category include Lua task handling and catch-all panic handling.  The new  `luaN_call()` variant establishes a PANIC call handler that uses `luaN_posttask()` to send the error traceback to the stdout pipe.  _Note that the error reporter defaults to using the base `print` function to send the error to stdout, but I plan add a `node.setatpanic()` API call in a subsuquent PR to allow production system to establish an alternative logging function for logging such panic errors._

-  `lua/lua.h`.  Add entries for `luaN_posttask()` and `luaN_call()` as these functions are used as part of the core Lua VM. New any variants `lua_isanyfunction(L,n)` and `lua_isanytable(L,n)` to hide the type test differences between the Lua 5.1 and 5.3 APIs.  (Note that these are used in `pipe.c` and `uart.c`, but will also be adopted in a later PR for all modules.) 

-  `lua/lbaselib.c`. Stderr-type errors now use the base `print` function and this calls `c_puts()` rather than `dbg_printf()` as the latter is really only for debugging and bypasses `node.output()` redirection.  (It was this change to use `dbg_printf()` that broke telnet).  

-  **Task Library**.  This functionality has now been moved into the platform core.  See https://github.com/nodemcu/nodemcu-firmware/issues/2811#issuecomment-505913336 for background to this.  Note that I have temproarily retained a `task.h` header that remaps the old interface to the platform calling conventions so that modules using the task interface don't need recoding except that the modules that use the taksk inferface have the `#include "task.h"` hoisted in the include order. The `luaX_` functions now include the Lua call interface.

-  `app/module/pipe.c` add the ability to bind a Lua CB function at `pipe.create()` to read and empty the pipe.

-  `app/module/node.c` the `node.output()` function now takes a piper reader as an argument.  The field output writes to a `stdout` pipe and the output reader is now executed as a separate task.  This means that `node.output()` works fine from the interactive session and the `stdin` and `stdout` pipes handle of the field marshalling automatically.

-  Make for `lua` now includes `-Wall`  plus minor fixes to lua files to remove `-Wall` warnings.  The one non-trivial change here is that `app/platform/vfs.h` previously used `static` declarations in this header which genrated compiler warnings when these should should have been `static inline`.  Plus the `lflash.c` changes also picked up by Johny on the `esp32` branch.  Also includes fixing some operator precedence issues because of lack of extra guard parentheses in some define macros in `driver/rotary.c` and `driver/spi.c`

-  Move the driver file `receive.c`to `input.c` plus enchancements, and ditto for its header.  The old file didn't actually do the UART receive handling as this had been moved into `lua.c`.  I've moved the receive handling back and since it does all of the line input handling for the UART0, I decided to rename it to `input.c` as this better reflects its purpose.

-  The driver file `uart.c` has had its initialisation binding split into two parts: `uart_init()` which is called from `user_main.c` and `uart_init_task()` which is called from the `input.c` driver as part of Lua startup to enable input delivery.
 
-  **.gdbinit**  some minor enhancements to this remote GDB macro set fro API developers

### Notes

This is a large change because of all of the interdependencies.  I suggest we flush out the pending PRs that we want to do in the next master drop, and merge this immediately after.


TerryE:
And the Pipe version of the telnet server.  This one is actually usable.  I was pasting the following to test out the marshalling
```Lua
r=debug.getregistry() 
function list(name, t)
  if #name < 15 and (name == '_G' or t ~= _G) then
    print(name,t)
    for k,v in pairs(t) do
      local name = name..'.'..k
      if type(v):sub(-5) == 'table' and name:sub(1,8) ~= '_G.r.std' then
        list(name, v)
      else
        print(name,v)
      end
    end
  end
end
list("_G",_G)
```
It kept running out of memory, until I twigged was also enumerating `r.stdin` and `r.stdout` and the pipes support the tostring intrinsic for debugging so r.stdout[2] contains the string representation of the first stdout pipe UData -- recursive chain reaction, hence the `name:sub(1,8) ~= '_G.r.std'` guard.  Phew!!

app/lua/lnodemcu.c
```diff
@@ -0,0 +1,116 @@
...
+** Use in CBs and other C functions to call a Lua function. This includes
+** an error handler which will catch any error and then post this to the 
+** registered reporter function as a separate follow-on task.
+*/
+int luaN_call (lua_State *L, int narg, int res, int doGC) { // [-narg, +0, v]
+  int status;
+  int base = lua_gettop(L) - narg;
+  lua_pushcfunction(L, luaN_traceback);
+  lua_insert(L, base);                                      /* put under args */
+  status = lua_pcall(L, narg, (res < 0 ? LUA_MULTRET : res), base);
```
jmattsson: Q: Would we not be better off having `luaN_call()` take the potential error message off the stack here? I know from experience that it's something quite easily forgotten by the caller since normally things don't fail and there's nothing left on the stack. Alternatively, push a nil on success to force the caller to balance the stack (less clean, but would provide the error message to the caller).

app/lua/lnodemcu.c
```diff
@@ -0,0 +1,116 @@
...
+}
+
+static platform_task_handle_t task_handle = 0;
+
+/*
+** Task callback handler. Uses luaN_call to do a protected call with full traceback
+*/
+static void do_task (platform_task_param_t task_fn_ref, uint8_t prio) {
+  lua_State* L = lua_getstate();
+  if (prio < 0|| prio > 2)
```
jmattsson: Symbolic names, please. And `platform_task_prio_t` in the prototype.

app/lua/lua.c
```diff
@@ -22,498 +19,284 @@
...
   }
   return status;
 }
 
+static void l_print(lua_State *L, int n) {
+  lua_getglobal(L, "print");
+  lua_insert(L, -n-1);
+  if (lua_pcall(L, n, 0, 0) != 0)
+    l_message(lua_pushfstring(L, "error calling " LUA_QL("print") " (%s)",
+                              lua_tostring(L, -1)));
```
jmattsson: I think this leaves the original error message on the stack here? Is that intentional?

app/lua/lua.c
```diff
@@ -22,498 +19,284 @@
...
+static int dojob (lua_State *L) {
+  size_t l;
+  int status;
+  const char *prompt;
+
+//dbg_printf("dojob entered\n");
+  lua_settop(L, 1);                                       /* pipe obj at S[1] */
+  lua_pushlightfunction(L, &pipe_read);                  /* pobj:read at S[2] */
+  lua_pushvalue(L, 1);                                    /* dup pobj to S[3] */
+  lua_pushliteral(L, "\n+");                                  /* S[4] = "/n+" */
```
jmattsson: Comment has / instead of \

app/lua/lua.c
```diff
@@ -22,498 +19,284 @@
...
+    if (MLref == LUA_NOREF) {
+      MLref = luaL_ref(L, LUA_REGISTRYINDEX);
+    } else {
+      lua_rawseti(L, LUA_REGISTRYINDEX, MLref);
     }
+  } else {
+    /* compile finished OK or with hard error */
+    lua_remove(L, 2);                    /* remove line because now redundant */
+    if (MLref!= LUA_NOREF) {       /* also remove multiline if it exists */
+      luaL_unref(L, LUA_REGISTRYINDEX, MLref);
```
jmattsson: I wish `luaL_unref()` returned `LUA_NOREF` so the unrefs could be one-liners easily.

app/modules/net.c
```diff
@@ -735,6 +735,14 @@ int net_getaddr( lua_State *L ) {
   lua_pushstring(L, addr_str);
   return 2;
 }
+#if 0
```
jmattsson: Tidy up various added debug things, please.

app/modules/net.c
```diff
@@ -763,11 +771,14 @@ int net_close( lua_State *L ) {
   }
   if (ud->type == TYPE_TCP_SERVER ||
      (ud->pcb == NULL && ud->client.wait_dns == 0)) {
-    lua_gc(L, LUA_GCSTOP, 0);
+//    lua_gc(L, LUA_GCSTOP, 0);
```
jmattsson: I believe these are intentional, to guard against the GC kicking in on the unref of `ud`, rendering the `ud->self_ref = LU_NOREF;` as undefined behaviour due to a use-after-free scenario.

app/modules/net.c
```diff
@@ -812,10 +823,13 @@ int net_delete( lua_State *L ) {
       ud->server.cb_accept_ref = LUA_NOREF;
       break;
   }
-  lua_gc(L, LUA_GCSTOP, 0);
+//  lua_gc(L, LUA_GCSTOP, 0);
```
jmattsson: Same as above

app/modules/node.c
```diff
@@ -171,69 +171,59 @@ static int node_heap( lua_State* L )
...
-  // }
-
-  if (output_redir_ref == LUA_NOREF) {
-    uart0_sendStr(str);
-    return;
-  }
+  int n = lua_gettop(L);
+  lua_pushliteral(L, "stdout");
+  lua_rawget(L, LUA_REGISTRYINDEX);                       /* fetch reg.stdout */
+  if (lua_istable(L, -1)) { /* reg.stdout is pipe */
```
jmattsson: Any old table suffices as a pipe? Not even want to check a metatable here?

app/modules/node.c
```diff
@@ -171,69 +171,59 @@ static int node_heap( lua_State* L )
...
     uart0_sendStr(str);
   }
-
-  lua_rawgeti(L, LUA_REGISTRYINDEX, output_redir_ref);
-  lua_pushstring(L, str);
-  lua_call(L, 1, 0);   // this call back function should never user output.
+  lua_settop(L, n);         /* Make sure all code paths leave stack unchanged */
 }
 
+extern int pipe_create(lua_State *L);
```
jmattsson: I've seen a few of these now. Could we get them into a proper header instead please? :)

app/modules/node.c
```diff
@@ -171,69 +171,59 @@ static int node_heap( lua_State* L )
...
-  if (lua_type(L, 1) == LUA_TFUNCTION || lua_type(L, 1) == LUA_TLIGHTFUNCTION) {
-    lua_pushvalue(L, 1);  // copy argument (func) to the top of stack
-    if (output_redir_ref != LUA_NOREF)
-      luaL_unref(L, LUA_REGISTRYINDEX, output_redir_ref);
-    output_redir_ref = luaL_ref(L, LUA_REGISTRYINDEX);
-  } else {    // unref the key press function
-    if (output_redir_ref != LUA_NOREF)
-      luaL_unref(L, LUA_REGISTRYINDEX, output_redir_ref);
-    output_redir_ref = LUA_NOREF;
+  serial_debug = (lua_isnumber(L, 2) && lua_tointeger(L, 2) == 0) ? 0 : 1;
```
jmattsson: `luaL_optint()` instead maybe? Or is it acceptable to have non-numbers as 2?

app/modules/node.c
```diff
@@ -171,69 +171,59 @@ static int node_heap( lua_State* L )
...
+  lua_settop(L, 1);
+  if (lua_isanyfunction(L, 1)) {
+    lua_pushlightfunction(L, &pipe_create);
+    lua_insert(L, 1); 
+    lua_pushinteger(L, LUA_TASK_MEDIUM);
+    lua_call(L, 2, 1);             /* T[1] = pipe.create(dojob, low_priority) */
```
jmattsson: Comment says "do_job", though in this case it's a user-provided function.

app/modules/pipe.c
```diff
@@ -3,22 +3,54 @@
...
 **
 ** Reads and writes may span multiple UD buffers and if the read spans multiple UDs
 ** then the parts are collected as strings on the Lua stack and then concatenated
-** with a `lua_concat()`.
+** with a lua_concat().
 **
-** Note that pipes also support the undocumented length and tostring operators
-** for debugging puposes, so if p is a pipe then #p[1] gives the effective
-** length of pipe slot 1 and printing p[1] gives its contents
+** Note that pipe tables also support the undocumented length and tostring
```
jmattsson: Looks like it's no longer undocumented! ;)

app/modules/pipe.c
```diff
@@ -104,22 +141,115 @@ static char getsize_delim (lua_State *L, int ndx, int *len) {     // [-0, +0, v]
...
+#define UVfunc  lua_upvalueindex(2)  // The CB's Lua function
+#define UVprio  lua_upvalueindex(3)  // The task priority
+#define UVstate lua_upvalueindex(4)  // Pipe state;
+#define CB_NOT_USED       0
+#define CB_ACTIVE         1
+#define CB_WRITE_UPDATED  2
+#define CB_QUIESCENT      4
+/*
+** Note that nothing precludes the Lua CB function from itself writing to the 
+** pipe and in this case this routine will call itself recursively.
```
jmattsson: Recursively, or merely retasking itself?

app/modules/uart.c
```diff
@@ -7,91 +7,65 @@
...
+
+void uart_on_data_cb(const char *buf, size_t len){
   lua_State *L = lua_getstate();
-  if(!L)
-    return false;
   lua_rawgeti(L, LUA_REGISTRYINDEX, uart_receive_rf);
   lua_pushlstring(L, buf, len);
-  lua_call(L, 1, 0);
-  return !run_input;
+  luaN_call(L, 1, 0, 0);
```
jmattsson: So here for example, I believe if the call fails we've just littered the stack with an error message.

app/modules/uart.c
```diff
@@ -7,91 +7,65 @@
...
+  char end_char = 0;
+  const char *method = lua_tostring( L, 1);
+  bool run_input = true;
+  luaL_argcheck(L, method && !c_strcmp(method, "data"), 1, "method not supported");
+
+  if (lua_type( L, stack ) == LUA_TNUMBER)
   {
-    need_len = ( uint16_t )luaL_checkinteger( L, stack );
+    data_len = luaL_checkinteger( L, stack );
+    luaL_argcheck(L, data_len >= 0 && data_len <= 255, stack, "wrong arg range");
```
jmattsson: Not `LUA_MAXINPUT`?

app/modules/uart.c
```diff
@@ -7,91 +7,65 @@
...
     }
-    end_char = (int16_t)end[0];
-    need_len = 0;
   }
 
-  // luaL_checkanyfunction(L, stack);
-  if (lua_type(L, stack) == LUA_TFUNCTION || lua_type(L, stack) == LUA_TLIGHTFUNCTION){
-    if ( lua_isnumber(L, stack+1) ){
-      run = lua_tointeger(L, stack+1);
+  if (lua_isfunction(L, stack) || lua_islightfunction(L, stack)) {
```
jmattsson: I believe you added a nice `lua_isanyfunction(L, stack)` macro above.

jmattsson:
Oh, and Travis is unhappy with you. :)

TerryE:
> Oh, and Travis is unhappy with you.

@jmattsson, you can blame @marcelstoer and @galjonsfigur and  #2790 for this one :smile: This does a `lint`- style check of all our Lua examples, This is a great idea, IMO _but we haven't as yet done the hard bit of going through all of the examples and fixing the issues, so it is supposed to be optional for now, but for some reason after I rebaselined against `dev`it is running in this PR.  The Travis fail arises from these check failing.

As to your other comments, I will go through them and do a separate consolidated reply for most as this will probably be the most readable for you and others.

galjonsfigur:
@TerryE I just looked at the Travis log and lint check does not break the build - it only spills lots of warnings, but because of `true` [in this line](https://github.com/nodemcu/nodemcu-firmware/blob/dev/.travis.yml#L36) it won't break the build. The real error is [here](https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/560814634#L3867) in the build log:
```
lflash.c: In function 'flashBlock':
lflash.c:124:3: error: format '%x' expects argument of type 'unsigned int', but argument 3 has type 'const void *' [-Werror=format=]
   NODE_DBG("flashBlock((%04x),%08x,%04x)\n", curOffset,b,size);
```

TerryE:
Reasoning behind some of these changes:
-  **Tasking**.  We added the task interface to Lua quite late on, but the more that I use it and `node.js` I realise that NodeMCU Lua should fully embrace the tasking paradigm. As `node.js` demonstrates this isn't really an SDK vs RTOS issue, but is core to the NodeMCU execution model.  The SDK vs RTOS aspect simply dictates whether the next event scheduler is provided as an external (SDK) service or is part of the platform environment (under RTOS).  
We need the tasking interface to be available before the Lua environment has had a chance to initialise itself and any ROM-based libraries, so it makes sense to hoist this code into the `app/platform` code set, hence the rename to the `platform_xxx` "namespace", though since we have other namespaces such as `vfs_` already as part of `platform` there would be no harm in retaining the `task_` but simply moving the service into the hierachy.  This includes decisions like renaming types like `platform_task_param_t`.
One thing that I want the platform headers to do is to fully encapsulate any links to SDK or RTOS headers to that we handle these withing the two platform layers and not `#ifdef` bracketed include sets in our code.  _**Reviewers thoughts, please**_
-  `stdin` and `stdout` pipes are an example of how we now use tasking.  The `uart` ISR posts a high priority task to request the UART buffer to be emptied.  This task is handled by the `input` driver code which buffer whole input lines and then does a `stdin:write()` to stash them in the stdin pipe. The pipe is read by a low priority pipe reader task internal to `lua.c` which processes one (extra) Lua line per invocation.  Since the stashing is done at a higher priority then this will be done preferentially to starting processing the next line.  What this means in practice is that it is really difficult to overflow the Lua input: you can just cut and paste large blocks of code into the UART or telnet without source data drop.  For me this makes using tools like Esplorer obsolete.  I also don't bother with SPIFFSimg any more since it is just easy to have a block of source Lua containing `file.putcontents('file1.lua [=[ ... ]=]) chunks and pasting this into the interpreter to set up the FS.
Likewise the core `print` function / `output_redirect` writes to the stdout pipe and this is emptied by a CB task which reads the pipe.
-   `luaN_call()` uses this  task interface as well.  Like `lua_pcall()` it doesn't throw an error and so always returns.  Internally it established a traceback error handler which catches any error traceback as a string and then posts a separate task to print this out / send it to the stdout pipe ( or in future to a application selectable handler if the app wants to log these errors separately).  So no more PANIC reboots for modules with use `luaN_call()` to call their Lua CBs, just a proper error traceback or the ability to log this separately.

I'll fix the minor review points that you also picked up.  Thanks :)

app/modules/uart.c
TerryE: > The real error is [here](https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/560814634#L3867) in the build log:

Yup, I'd already noticed that one. :cry:

app/driver/input.c
```diff
@@ -0,0 +1,201 @@
...
+int lua_main (void);
+static bool input_readline(void);
+
+static void input_handler(platform_task_param_t flag, uint8 priority) {
+  (void) priority;
+  if (!ins.data) {
+    lua_main();
+    return;
+  }
+  ins.input_sig_flag = flag & 0x1;
```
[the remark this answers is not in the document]
TerryE: It's now a 0/1 field so this is once where I think that a symbolic isn't really needed.

app/driver/input.c
```diff
@@ -0,0 +1,201 @@
...
+        ins.data[ins.line_pos] = 0;
+        ins.last_char = NUL;
+        continue;
+      }
+      ins.last_char = ch;
+
+      /* end of data */
+      if (ch == CR || ch == LF) {
+        if (ins.uart_echo) putc(LF);
+        if (ins.uart_cb) ins.uart_cb(ins.data, ins.line_pos);
```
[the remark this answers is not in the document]
TerryE: We have control over output but not necessarily input so we want to make input LF/CR tolerant.

app/driver/pwm.c
```diff
@@ -253,7 +253,7 @@ pwm_set_freq(uint16 freq, uint8 channel)
 
     pwm.period = PWM_1S / pwm.freq;
 }
-
+#if 0
```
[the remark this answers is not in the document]
TerryE: This was throwing an unused static warning, so I added the `#if 0` `#endif` guard to kill the error, but I didn't want to remove the code altogether because I am not the maintainer.  We can live with this one for now.

app/driver/spi.c
```diff
@@ -440,7 +433,7 @@ void spi_byte_write_espslave(uint8 spi_no,uint8 data)
 *******************************************************************************/
 void spi_slave_init(uint8 spi_no)
 {
-    uint32 regvalue;
+//  uint32 regvalue;
```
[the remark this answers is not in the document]
TerryE: Same argument as above:  I was doing the minimum change to get a clean `-Wall` compile.

app/driver/uart.c
```diff
@@ -165,7 +165,7 @@ uart_tx_one_char(uint8 uart, uint8 TxChar)
     WRITE_PERI_REG(UART_FIFO(uart) , TxChar);
     return OK;
 }
-
+#if 0
```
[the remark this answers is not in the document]
TerryE: Done

app/driver/uart.c
```diff
@@ -300,13 +300,15 @@ uart0_rx_intr_handler(void *para)
     }
 
     if (got_input && sig) {
+      // Only post a new handler request once the handler has fired clearing the last post
       if (isr_flag == *sig_flag) {
         isr_flag ^= 0x01;
```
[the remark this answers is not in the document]
TerryE: See previous comment

app/lua/Makefile
```diff
@@ -16,7 +16,7 @@ SUBDIRS = luac_cross
 GEN_LIBS = liblua.a
 endif
 
-STD_CFLAGS=-std=gnu11 -Wimplicit
+STD_CFLAGS=-std=gnu11 -Wimplicit -Wall
```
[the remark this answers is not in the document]
TerryE: It took me a few hours to get `-Wall` to compile.  There are some circumstances where we might validly need to add pragmas to suppress `-Wextra` warnings, so a job for another day please.

app/lua/lbaselib.c
```diff
@@ -40,20 +38,11 @@ static int luaB_print (lua_State *L) {
     if (s == NULL)
       return luaL_error(L, LUA_QL("tostring") " must return a string to "
                            LUA_QL("print"));
-#if defined(LUA_USE_STDIO)
-    if (i>1) c_fputs("\t", c_stdout);
-    c_fputs(s, c_stdout);
-#else
-    if (i>1)  luai_writestring("\t", 1);
-    luai_writestring(s, c_strlen(s));
-#endif
```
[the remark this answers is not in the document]
TerryE: This is the existing `lua`, so we will never merge `dev` and `esp32` on this branch.  I agree that we need to harmonise `lua53` properly.

app/lua/lflash.c
```diff
@@ -456,7 +451,7 @@ int procSecondPass (void) {
   int i, len = (out->ndx > out->flashLen) ?
                   (out->flashLen % WRITE_BLOCKSIZE) / WORDSIZE :
                   WRITE_BLOCKSIZE / WORDSIZE;
-  uint32_t *buf = (uint32_t *) out->buffer.byte, flags;
+  uint32_t *buf = (uint32_t *) out->buffer.byte, flags = 0;
```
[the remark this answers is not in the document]
TerryE: Done

app/lua/lnodemcu.c
```diff
@@ -0,0 +1,116 @@
+/*
+** $Id: ltm.c,v 2.8.1.1 2007/12/27 13:02:25 roberto Exp $
+** Tag methods
+** See Copyright Notice in lua.h
```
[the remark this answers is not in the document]
TerryE: Yup but there's only so much I can do in one PR and this one is maybe already 4× the aggregate of all other PRs this drop cycle.

app/lua/lnodemcu.c
```diff
@@ -0,0 +1,116 @@
...
+#include "lauxlib.h"
+#include "lgc.h"
+#include "lstring.h"
+#include "ltable.h"
+#include "ltm.h"
+#include "lrotable.h"
+#include "platform.h"
+
+extern int debug_errorfb (lua_State *L);
+#if 0
```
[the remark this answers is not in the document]
TerryE: Done

app/lua/lnodemcu.c
```diff
@@ -0,0 +1,116 @@
...
+extern int pipe_unread(lua_State *L);
+extern int pipe_write(lua_State *L);
+#endif
+/*
+** Error Reporting Task.  We can't pass a string parameter to the error reporter
+** directly through the task interface the call is wrapped in a C closure with
+** the error string as an Upval and this is posted to call the Lua reporter.
+*/
+static int report_traceback (lua_State *L) {
+// **Temp**  lua_rawgeti(L, LUA_REGISTRYINDEX, G(L)->error_reporter);
```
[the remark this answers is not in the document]
TerryE: Nope placeholder because on the next API, I am adding an extra `node` API call to allow the app to set the error reporter CB.  What these changes do is to send PANIC tracebacks to stdout, but on a prod system the app developer might want to log them or send them back to a central syslog.

app/lua/lnodemcu.c
```diff
@@ -0,0 +1,116 @@
...
+  lua_getglobal(L, "print");
+  lua_pushvalue(L, lua_upvalueindex(1));
+  lua_call(L, 1, 0);  /* Using an error handler would cause an infinite loop! */
```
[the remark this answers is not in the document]
TerryE: Can't use pcall because this would just end up in an infinite loop here.

app/lua/lnodemcu.c
TerryE: See general response below.  `luaN_call()` does take the error off the task (see `luaN_traceback()` and `report_traceback()`.  Ultimately the app programmer will have the option of establishing an error reporter.   If the library routine wants to manage the error itself then it will already be using a `lua_pcall()`.  IIRC there is only one instance of this in the existing modules.  

This is designed to be a one-line swap-out in such library CB executions, and they don't have or have never tested error paths, since the current runtime simply throws a PANIC error and reboots the system.  Can I suggest if we want to fine tune this functionality then let us raise a separate specific issue.

app/lua/lua.c
TerryE: Not sure, but I immediately do a settop on return so the stack does balance -- I checked this one.

app/lua/lua.c
TerryE: Done . Nit picker.  :smile:

app/lua/lua.c
TerryE: So do I but this sort of rationalisation is for a downstream PR.

app/modules/net.c
TerryE: Sorry but I think that we should leave this one in.  The GC is net is still not robust and this debug diagnostic can be really useful to sort out what is going on here.  For now this is all commented out, but maybe we should leave it like this IMO

app/modules/net.c
TerryE: The UD is still on the Lua stack so the `luaL_unref()` _cannot_ trigger its GC;  there is no point in stopping GC around this statement.

app/modules/net.c
TerryE: Same as above

app/modules/node.c
TerryE: I can't think of any circumstance where the Lua runtime performs integrity checks to see if the app have been dicking around with the Lua registry.  The rule is: if you are dumb enough to screw around with the registry your app dies in horrible ways.  Why have this one exception here?

app/modules/node.c
TerryE: This is just a compaction of the existing coded logic.  I just didn't have the bandwidth to rewrite everything that I touched.

app/modules/node.c
TerryE: Yes there are 2 explicit externs in `lua.c` and `node.c` but these are internal to VM and  I **_don't_** want to expose this extern for general application use in a general header, which is why it isn't in one.

We can do it this way or I can have a header with one line in it called, say `pipe_for_lua_runtime_only.h`.  I chose the pragmatic option.

app/modules/node.c
TerryE: Fixed.  Sorry cut and paste residue.

app/modules/pipe.c
TerryE: Documented as in readthedocs general documentation.  Anyone reading this is digging around in the pipe internals anyway.

app/modules/pipe.c
TerryE: Recursively vs retasking.  I leave it to you to propose alternative wording. 

The real point here was that node.output() used to be a minefield as it was really breaking the VM architectural assumptions in that the print base function was never intended to embed a Lua CB, so doing a `print` inside this Lua CB code could trash the whole runtime environment and PANIC the processor.  

Because the output interface now uses pipes and the tasking mechanism we now comply with the architecture and there is actually nothing stopping the output CB doing a print.

app/modules/uart.c
TerryE: This actually touches on a more systemic issue in that all CB action routines should always balance the stack from entry to return but we don't have any guidance / quality check / coding pattern to ensure this and failure to do so leads to resource leakage that will ultimately exhaust memory.   We need to address this wider issue, IMO.

However as to this specific point, we need to decide the stack rules for `luaN_call()`.  IMO we have two options:
1.  It follows the `lua_pcall()`convention of returning 1 argument and an error status if the called routine throws an error, or
2.  It dummies the `lua_call()` of returning `nresults` arguments (which might be nil in the case of an error) and an error status.

IIRC, there only one use in the modules library where the library routine accepts _any_ result from the Lua CB, and all that typically happens is that the wrapper function returns control to the SDK on return from the CB.  I think that we should go for easy of migration to `luaN_call()` and that adopting option 2 will best achieve this.

app/modules/uart.c
TerryE: Dunno.  Ask Zeroday why he used 255.  All I did was to recode the existing validation logic to use `luaL_argcheck()` whilst getting rid of some warnings. 

The actual UART input buffer is allocated with `input_setup()` and this is called in `lua.c` using `LUA_MAXINPUT` so I agree that this would be a better symbolic value.  So updated.

app/modules/uart.c
TerryE: Yup, but I'll be doing all of these as a separate sweep.

app/lua/Makefile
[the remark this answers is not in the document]
jmattsson: Can't blame a guy for trying, eh? ;)

app/lua/lnodemcu.c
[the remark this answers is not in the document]
jmattsson: Only if you give an error handler for it, I thought? Without an error handler you should only end up with a message on the stack. I'm happy to leave as-is though, this is already such a big improvement in potential stability.

app/lua/lnodemcu.c
jmattsson: Agreed that we can tweak specifics later. I just want to make sure we don't accidentally leave the result of an errored pcall on the stack and thus unbalancing it.

app/lua/lua.c
jmattsson: Ah, all good then!

app/modules/net.c
jmattsson: Alright, if there's a reasonable justification for it it's all good.

app/modules/net.c
jmattsson: Ah, of course, yes, you're right. Let's get rid of those (commented) calls altogether then, they're plain misleading and encouraging bad copy/pasta :)  I mean, they had me fooled...

app/modules/node.c
jmattsson: Fair point.

app/modules/node.c
jmattsson: \<insert clever retort here\> (because I can't come up with one - it's not like that particular prototype is ever likely to change)

app/modules/node.c
jmattsson: Acknowledged.

app/modules/uart.c
jmattsson: My vote would be for #2 as well. If there are any instances where a module is actually interested in the error message, they can use `lua_pcall()` for that. For everyone else, I think the simplicity of `lua_call()` behaviour will give us the most robust outcome.

app/modules/uart.c
jmattsson: Noted. And thumbs up.

jmattsson:
When Travis is happy, I'll be happy. This is a very nice (and large) improvement - thanks for all the work Terry!

TerryE:
Thanks Jonny for all of the feedback. I've tried to action most of your comments except for a few where I've given my reasons.  As I say in the title, this is only phase 1. We've got a few more steps to go.  However now that you've taken off #2838, I'll be pausing further work here until #2838 is landed and then I can rebase this against the new `dev`.

devsaurus:
These changes look good and are a significant step forward. Thanks a lot Terry!
Haven't had the time to actually test drive this PR though.

TerryE:
I've just rebaselined against the current `dev`, hence the forced push.  No ready for dev merge yet.  Need to do master drop anyway.

TerryE:
@jmattsson @devsaurus @nwf, I have another big batch of changes pending tidying up the modules.  I suspect that its going to be another couple of weeks before we eventually do the next master drops and are positioned to merge this.  That will mean another month before we also review and do the module changes.  Would there be any advantage also to rolling in the module changes into this PR?  That way we can drop a couple of weeks from the timeline.

marcelstoer:
> I suspect that it is going to be another couple of weeks before we eventually do the next master drops 

Why? Most issues in https://github.com/nodemcu/nodemcu-firmware/milestone/13 are either completed or close to completion. If it's any help we can drop some PRs from the milestone and snap to master next week. What's your preference?

TerryE:
Marcel whenever you, Gregor and Nathaniel can make the cut ... but if these are joined, then I can roll on the prep work.  It's really up to the reviews as to whether they think this rolled up patch will be too big to review.

jmattsson:
I'd prefer a second PR for the next round of clean-up, it'd be a lot easier to review. Nothing stopping you from just branching off from this PR branch and continuing work though.

TerryE:
OK will do it that way. Just have too rebase twice.

jmattsson:
Might not need any more rebasing. This one is looking all green at the moment, so unless there's something conflicting going in soon it should just be a couple of squash-n-merges :)

TerryE:
I've got my `dev-rotable-opts` rebased against this PR and everything seems to be working fine, though there is little point in raising this as a PR until this one is merged, as I will want to rebase against the current `dev` before pushing the new PR.

marcelstoer:
@TerryE can you please update or close all the referenced issues if necessary.

Issue #2839, opened 2019-07-21
Tidying up the make hierarchy 


TerryE:
This is a component of #2816 and relates to #2811 

The current make hierarchy is a sloppy mess which I want to clean up but staying within the current make framework.  If we look at a typical compile line generated in the make

Type | Option
----|--------
Compiler Warnings | `-Wimplicit` `-Wpointer-arith` `-Wundef` `-Werror`
Optimisation stuff | `-O2` `-ffunction-sections` `-fno-jump-tables` `-fdata-sections` `-fno-inline-functions` `-mlongcalls` `-mtext-section-literals`
Language Spec | `-std=gnu11`
Linker stuff | `-Wl,-EL` `-nostdlib`
Preprocesssor defines | `` `__ets__` `EBUF_LWIP` `ICACHE_FLASH` `LWIP_OPEN_SRC` `MBEDTLS_USER_CONFIG_FILE="user_mbedtls.h"` `` `MEM_DEFAULT_USE_DRAM` `PBUF_RSV_FOR_WLAN` `USE_OPTIMIZE_PRINTF` ``

Type | Include List (not nec in this order)
----|--------
Dirs don't exist / contain headers | `<root>/` `<root>/include` `../smart` `../../include/eagle` 
Defaults | `./` `include` `../include``../../include`| `../spiffs` `../fatfs`
Core stuff | `../libc` `../lua` `../platform` 
Why expose? | `../spiffs` `../fatfs`
Packages | `../coap ` `../dht` `../http` `../mqtt` `../pcm` `../pm` `../sjson` `../sqlite3` `../u8g2lib/u8g2/src/clib` `../ucglib/ucg/src/clib` `../websocket`

This all need cleaning up:
-  The warnings and optimisation stuff are fine, except that we will be moving directories to `-Wall` `-Wextra` over time. 
-  Defines should be moved into a config header file or down into the sub make that needs them
-  Pathing in directories that don't exist or don't validly contain includes should be removed
-  We have some standard roots: `ROOTDIR` `APPDIR` `LUADIR` `INCDIR`
-  Most of the package includes are used in very specific instances so there is no point in including `app/http` (for example) in the search list; the one reference in  `app/http/httpclient.c` should simply refer to `#include "http/httpclient.h"` instead.

This is all pretty mechanistic tidying.  I just don't want any hard wired reference to `app` or `app/lua` in any Makefiles.

jmattsson:
[...]

PR #2841, opened 2019-07-22
Minor cleanup: removal of c_types.h, other c_prefix tidy-up

jmattsson:
Alright, due to popular request (@galjonsfigur), here is a follow-up to #2838 that removes uses of `c_types.h` as well, and prevents it from going back in.

We're still left with inherited junk in our overridden `stdint.h`, but that can be whittled down over time hopefully. Some of that stuff should definitely be moved into some platform abstraction layer or other. Other bits yet should be search'n'replaced and removed in my opinion (e.g. a bunch of the integer type variations).

Also included is the tidy-up of a few missed other c_stdxxx replacements.

This is a far smaller cleanup than #2838, but I believe this now leaves us with being able to refer only to the standard headers within the NodeMCU code. A second sweep for `os_`prefixes should probably also be done though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,9 @@ SDK_DIR          := $(TOP_DIR)/$(SDK_REL_DIR)
 ESPTOOL_VER := 2.6
 
 # Ensure that the Espresif SDK is searched before the tool-chain's SDK (if any)
-CCFLAGS:= -I$(TOP_DIR)/sdk-overrides/include -I$(TOP_DIR)/app/include/lwip/app -I$(SDK_DIR)/include
+# Also, prevent the SDK's c_types.h from being included from anywhere, by
+# predefining its include-guard.
+CCFLAGS:= -I$(TOP_DIR)/sdk-overrides/include -I$(TOP_DIR)/app/include/lwip/app -I$(SDK_DIR)/include -D_C_TYPES_H_
 LDFLAGS:= -L$(SDK_DIR)/lib -L$(SDK_DIR)/ld $(LDFLAGS)
 
 ifdef DEBUG
--- a/app/coap/coap_client.c
+++ b/app/coap/coap_client.c
@@ -1,5 +1,5 @@
 #include "user_config.h"
-#include "c_types.h"
+#include <stdint.h>
 
 #include "coap.h"
 #include "hash.h"
--- a/app/coap/coap_io.h
+++ b/app/coap/coap_io.h
@@ -5,7 +5,6 @@
 extern "C" {
 #endif
 
-#include "c_types.h"
 #include "lwip/ip_addr.h"
 #include "espconn.h"
 #include "pdu.h"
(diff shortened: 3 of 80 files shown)


nwf:
Could we move the compatibility cruft from `stdint,h` to a `sdk_compat,h` or something to make it more obvious what's going on?

What's with the guard against ever including `c_types.h`?  Do we end up transitively including it somewhere and need to prevent that?

TerryE:
`c_types.h` is pretty incompatible with the newlib headers. So including it will tend to break compiles.  I pulled a late nighter #2839. This is ready to submit, but since it's mostly Makefiles, rebasing will be easy.

nwf:
Hm; rephrasing: is the `-D_C_TYPES_H_` intended to be permanent or just something to help us weed it out of our tree?

jmattsson:
What Terry said. And as mentioned, much of the cruft should be possible to get rid of with a search'n'replace sweep, but I wanted to keep that separate. The ICACHE stuff needs another cleanup sweep. We removed the need for most of that when we went to using a custom linker script and the exception handler for unaligned loads, but we (I?) never got around to nuking the now-unnecessary ICACHE macro uses that are littered in the older code. The other macros in that area are used by the lwIP code, and since that's something that won't get shared between the 8266/32 branches (in a hurry at least), those macros should probably be stuffed into an internal lwIP header somewhere.

And to answer the question that popped up while I was typing this, I'd be inclined to keep the `-D_C_TYPES_H_` indefinitely; accidentally getting `c_types.h` in can break the build in not-entirely-obvious-what-you-did-wrong ways. Since there are no more expected non-OS SDK versions, there should be no risk of accidental breakage due to the SDK header changing in unexpected ways.

TerryE:
@jmattsson, this is brilliant, this makes la53 even closer to "out of the box".  I've got  to rebaseline a could of PRs now, but that the cost of progress.  I would merge this but just in case I pre-empt some further tweaks, given that you have a double :+1:, I think we can leave you to do the squash / merge.

TerryE:
@jmattsson your <stdio.h> override does an include_next on the newlib one in the xtensa toolchain with a could of redefines that allows the code to compile, but these calls do not map onto the vfs equivalents and hence any code which uses the stdio API might compile clean but won't run under the SDK.  More work needed on this one.

marcelstoer:
@TerryE @jmattsson do you see this as a blocker for the upcoming master drop? If so (or even if not) does this need tracking in an issue of its own?

jmattsson:
I do not. This is largely a gotcha for module developers. Probably worth tracking separately, unless someone has a great idea on the best way forward. I'd like to see something that gets us closer to a standard C library, but I don't have much to offer in the ways of getting us towards that (what little time I have atm will be going toward the branch merging goal).

TerryE:
I've already coded around it in one module and they all compile clean under both lua51 and lua53 on my test setup.  This is just a longer term cleanup note so we don't forget this in the later ESP variant merge.

PR #2842, opened 2019-07-22
Dev make cleanup

### LINKED ISSUE #2839: Tidying up the make hierarchy

TerryE:
Fixes #2839

This change doesn't affect the compiled binary, just the make process itself.  The best way to see this is with a `make V=1` before and after.

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -15,8 +15,6 @@ TARGET = eagle
 #FLAVOR = release
 FLAVOR = debug
 
-#EXTRA_CCFLAGS += -u
-
 ifndef PDIR # {
 GEN_IMAGES= eagle.app.v6.out
 GEN_BINS= eagle.app.v6.bin
@@ -42,10 +40,10 @@ SUBDIRS= 					\
 	smart 					\
 	modules 				\
 	spiffs 					\
-	net					\
+	net						\
 	fatfs					\
 	esp-gdbstub				\
-	pm					\
+	pm						\
 	uzlib					\
 	$(OPT_SEL_MKTARGETS)
 
@@ -54,19 +52,19 @@ endif # } PDIR
 APPDIR = .
 LDDIR = ../ld
 
-TARGET_LDFLAGS = 				\
+TARGET_LDFLAGS = 			\
 	-nostdlib 				\
 	-Wl,-EL 				\
-	--longcalls 				\
+	--longcalls 			\
 	--text-section-literals
 
 LD_FILE = $(LDDIR)/nodemcu.ld
 
-COMPONENTS_eagle.app.v6 = 			\
+COMPONENTS_eagle.app.v6 = 		\
 	user/libuser.a 				\
 	crypto/libcrypto.a			\
 	driver/libdriver.a 			\
-	platform/libplatform.a 			\
+	platform/libplatform.a		\
 	task/libtask.a 				\
 	libc/liblibc.a 				\
 	lua/liblua.a 				\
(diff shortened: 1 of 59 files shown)


nwf:
These changes all seem sensible.  While you're at it, can the `** HEALTH WARNING **` section of the main `Makefile` get tidied, too?  (Perhaps that's a bigger thing and should wait for a later PR.)

TerryE:
OK I've now found why Travis failed.  The build process specifies EXTRA_CCFLAGS as a commandline parameter and doing this overrides all EXTRA_CCFLAGS setting inside the make so I need to make some adjustments to accommodate this.

TerryE:
It might be good to get #2841 and this in before we do the master drop, but I think that they both touch some common files, so whoever goes second will need to do some fixups or rebaseline.  We should let #2841 go first and soon, I think.

app/Makefile
```diff
@@ -54,35 +52,36 @@ endif # } PDIR
...
+	pm/libpm.a					\
+	esp-gdbstub/libgdbstub.a	\
+	net/libnodemcu_net.a		\
+	mbedtls/libmbedtls.a    	\
+	modules/libmodules.a		\
 	smart/smart.a 				\
 	uzlib/libuzlib.a 			\
 	$(OPT_SEL_COMPONENTS)
-
+# Ensure that the Espresif SDK is searched before the tool-chain's SDK (if any)
```
jmattsson: I don't think this comment is accurate here. Copy paste slip?

app/user/Makefile
```diff
@@ -38,13 +38,7 @@ DEFINES += -DESP_INIT_DATA_DEFAULT="\"$(SDK_DIR)/bin/esp_init_data_default_v05.b
...
 #
 
-INCLUDES := $(INCLUDES) -I $(PDIR)include
-INCLUDES += -I ./
-INCLUDES += -I ../../include/ets
-INCLUDES += -I ../libc
-INCLUDES += -I ../platform
-INCLUDES += -I ../lua
-INCLUDES += -I ../spiffs
+INCLUDES += -I ../../include/ets -I ../spiffs
```
jmattsson: I don't think we have an `include/ets` either, we should be able to evict that too.

sdk-overrides/include/c_types.h
```diff
@@ -0,0 +1,2 @@
+#define ICACHE_FLASH
+#include_next "c_types.h"
```
jmattsson: Objection, your honour! :)
When I hit that other shiny green button just before, our `c_types.h` got decimated and should stay that way.

app/user/Makefile
TerryE: Will do

sdk-overrides/include/c_types.h
TerryE: Yup, agreed. All part of the rebaseline TBD today. The reason for the wrapper is that ICACHE_FLASH was only used in this include.

app/Makefile
TerryE: I will check and remove

TerryE:
I will do the rebaseline my late PM. I've got other commitments before then.

TerryE:
I am note sure of the procedural niceties here. @nwf has reviewed this but not used the github review function do to this.  @jmattsson has also reviewed this and I've addressed these comments but not I am not sure how to mark this off as completed.  Anyway, if this Travis make completes then this is a verification of the make options as reviewed by two other committers, so I will then merge and squash this.

PR #2851, opened 2019-07-24
Example Lua module for coroutining

### LINKED ISSUE #2848: Adding a coroutining example and explaining how it can be really useful
Now that I am really getting to grips with the NodeMCU Lua execution model in my bones I've realised that we can use Lua coroutining for a whole class of long running applications and still stay within the SDK strictures.  

In essence the dilemma of the SDK is that tasks are recommended to be no more than 15 mSec in length and so the standard way of tackling this is by creating CB event chains, but another simple way is to use coroutining instead.   For example if I do a simple recursive walk of _G, ROM and the Registry then this will crash-out with a WDT error unless I smatter the code with `tmr.wdclr()` that could in turn cause the network stack to fail.  So here is how to do it using coroutining:
```Lua
do
  local function batch_tree_walk(taskYield, list)  
    local s, n, nCBs = {}, 0, 0
     
    local function list_entry (name, v) -- upval: taskYield, nCBs
      print(name, v)
      n = n + 1
      if n % 20 == 0 then nCBs = taskYield(nCBs) end
      if type(v):sub(-5) ~= 'table' or s[v] then return end
      if name == 'Reg.stdout' then return end  -- needed for telnet !!
      s[v]=true
      for k,tv in pairs(v) do
        list_entry(name..'.'..k, tv)
      end
      s[v] = nil
     end
     
     for k,v in pairs(list) do
       list_entry(k, v)
    end
    print ('Total lines, print batches = ', n, nCBs)
  end

  local co = coroutine.create(batch_tree_walk)

  local function taskYield(nCBs)  -- upval: co
    node.task.post(function () -- upval: co, nCBs 
      coroutine.resume(co, nCBs)
    end)   
    return coroutine.yield() + 1
  end 

  coroutine.resume(co, taskYield, {_G = _G, Reg = debug.getregistry(), ROM= ROM})
end
```
For my minimal test build this prints out just under 400 lines with each task printing out 20; and it works fine over telnet as well.

So do you understand how this works, and should we add it as an example / FAQ discussion point?  

At the end of this the Lua GC collects the lot (since the coroutine state is in an upvalue that gets descoped).  You can also have other Lua and SDK CBs transparently slotting in between the slices.

 _Really neat!_

**PS**: and a bonus point for anyone who can tell me what the array `s` is doing here.

TerryE:
Fixes #2848 

See #2848 and supplied docs for explanation.

### DIFF
--- a/docs/lua-modules/cohelper.md
+++ b/docs/lua-modules/cohelper.md
@@ -1,0 +1,69 @@
+# cohelper Module
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2019-07-24 | [TerryE](https://github.com/TerryE) | [TerryE](https://github.com/TerryE)  | [cohelper.lua](../../lua_modules/cohelper/cohelper.lua) |
+
+This module provides a simple wrapper around long running functions to allow 
+these to execute within the SDK and its advised limit of 15 mSec per individual
+task execution.  It does this by exploiting the standard Lua coroutine 
+functionality as described in the [Lua RM §2.11](https://www.lua.org/manual/5.1/manual.html#2.11) and [PiL Chapter 9](https://www.lua.org/pil/9.html).
+
+The NodeMCU Lua VM fully supports the standard coroutine functionality. Any
+interactive or callback tasks are executed in the default thread, and the coroutine
+itself runs in a second Lua thread. The coroutine can call any library functions,
+but any subsequent callbacks will, of course, execute in the default thread.
+
+Interaction between the coroutine and the parent is through yield and resume 
+statements, and since the order of SDK tasks is indeterminate, the application 
+must take care to handle any ordering issues.  This particular example uses
+the `node.task.post()` API withi the `taskYield()`function to resume itself, 
+so the running code can simple call `taskYield()` at regular points in the 
+processing to spilt the work into separate SDK tasks.
+
+A similar approach could be based on timer or on a socket or pipe CB.  If you 
+want to develop such a variant then start by reviewing the source and understanding
+what it does. 
+
+## Use
+```Lua
+do
+  local function longRunningFunc(yield, ...)
+  -- <content of function to be split into tasks>
+  end
+  require "cohelper".exec(longRunningFunc, params)
+end
+```
+
+## Full Example
+
+Here is a function which recursively walks the globals environment, the ROM table
+and the Registry. Without coroutining, this walk terminate with a PANIC following
+a watchdog timout. I don't want to sprinkle the code with `tmr.wdclr(`) that could
+in turn cause the network stack to fail. Here is how to do it using coroutining:
+
+```Lua
+require "cohelper".exec(
+  function(taskYield, list)
+    local s, n, nCBs = {}, 0, 0
+
+    local function list_entry (name, v) -- upval: taskYield, nCBs
+      print(name, v)
+      n = n + 1
+      if n % 20 == 0 then nCBs = taskYield(nCBs) end
+      if type(v):sub(-5) ~= 'table' or s[v] or name == 'Reg.stdout' then return end
+      s[v]=true
+      for k,tv in pairs(v) do
+        list_entry(name..'.'..k, tv)
+      end
+      s[v] = nil
+    end
+
+    for k,v in pairs(list) do
+      list_entry(k, v)
+    end
+    print ('Total lines, print batches = ', n, nCBs)
+  end,
+  {_G = _G, Reg = debug.getregistry(), ROM = ROM}
+)
+```
+
--- a/docs/lua-modules/cohelper.md
+++ b/docs/lua-modules/cohelper.md
@@ -16,7 +16,7 @@ but any subsequent callbacks will, of course, execute in the default thread.
 Interaction between the coroutine and the parent is through yield and resume 
 statements, and since the order of SDK tasks is indeterminate, the application 
 must take care to handle any ordering issues.  This particular example uses
-the `node.task.post()` API withi the `taskYield()`function to resume itself, 
+the `node.task.post()` API with the `taskYield()`function to resume itself, 
 so the running code can simple call `taskYield()` at regular points in the 
 processing to spilt the work into separate SDK tasks.
 
@@ -24,17 +24,35 @@ A similar approach could be based on timer or on a socket or pipe CB.  If you
 want to develop such a variant then start by reviewing the source and understanding
 what it does. 
 
-## Use
-```Lua
-do
-  local function longRunningFunc(yield, ...)
-  -- <content of function to be split into tasks>
-  end
-  require "cohelper".exec(longRunningFunc, params)
-end
+### Require
+```lua
+local cohelper = require("cohelper")
+-- or linked directly with the  `exec()` method
+require("cohelper").exec(func, <params>)
 ```
 
-## Full Example
+### Release
+
+Not required.  All resoruces are release on completion of the `exec()` method
+
+## `cohelper.exec()`
+Execute a function which is wrapper by a coroutine handler.
+
+#### Syntax
+`require("cohelper").exec(func, <params>)`
+
+#### Parameters
+- `func`: Port number for HTTP server. Most HTTP servers listen at port 80.
+- `<params>`: list of 0 or more parameters used to initialise func.  the number and types must be matched to the funct declaration  
+
+#### Returns
+Return result of first yield.
+
+#### Notes
+1.  The coroutine function `func()` has 1+_n_ arguments The first is the supplied task yield function. Calling this yield function within `func()` will temporarily break execution and cause an SDK reschedule which migh allow other executinng tasks to be executed before is resumed. The remaining arguments are passed to the `func()` on first call.
+2.  The current implementation passes a single integer parameter across `resume()` /   `yield()` interface.  This acts to count the number of yields that occur.  Depending on your appplication requirements, you might wish to amend this.
+
+### Full Example
 
 Here is a function which recursively walks the globals environment, the ROM table
 and the Registry. Without coroutining, this walk terminate with a PANIC following
--- a/docs/lua-modules/cohelper.md
+++ b/docs/lua-modules/cohelper.md
@@ -3,26 +3,27 @@
 | :----- | :-------------------- | :---------- | :------ |
 | 2019-07-24 | [TerryE](https://github.com/TerryE) | [TerryE](https://github.com/TerryE)  | [cohelper.lua](../../lua_modules/cohelper/cohelper.lua) |
 
-This module provides a simple wrapper around long running functions to allow 
+This module provides a simple wrapper around long running functions to allow
 these to execute within the SDK and its advised limit of 15 mSec per individual
-task execution.  It does this by exploiting the standard Lua coroutine 
+task execution.  It does this by exploiting the standard Lua coroutine
 functionality as described in the [Lua RM §2.11](https://www.lua.org/manual/5.1/manual.html#2.11) and [PiL Chapter 9](https://www.lua.org/pil/9.html).
 
 The NodeMCU Lua VM fully supports the standard coroutine functionality. Any
 interactive or callback tasks are executed in the default thread, and the coroutine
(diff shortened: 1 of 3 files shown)


TerryE:
@galjonsfigur, perhaps you might to review this as well :smile:

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,69 @@
...
+
+The NodeMCU Lua VM fully supports the standard coroutine functionality. Any
+interactive or callback tasks are executed in the default thread, and the coroutine
+itself runs in a second Lua thread. The coroutine can call any library functions,
+but any subsequent callbacks will, of course, execute in the default thread.
+
+Interaction between the coroutine and the parent is through yield and resume 
+statements, and since the order of SDK tasks is indeterminate, the application 
+must take care to handle any ordering issues.  This particular example uses
+the `node.task.post()` API withi the `taskYield()`function to resume itself,
```
nwf: typo `with`

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,69 @@
...
+itself runs in a second Lua thread. The coroutine can call any library functions,
+but any subsequent callbacks will, of course, execute in the default thread.
+
+Interaction between the coroutine and the parent is through yield and resume 
+statements, and since the order of SDK tasks is indeterminate, the application 
+must take care to handle any ordering issues.  This particular example uses
+the `node.task.post()` API withi the `taskYield()`function to resume itself, 
+so the running code can simple call `taskYield()` at regular points in the 
+processing to spilt the work into separate SDK tasks.
+
```
nwf: Other than this prose, there doesn't appear to be our usual kind of documentation for `taskYield()`?

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,69 @@
...
+```Lua
+require "cohelper".exec(
+  function(taskYield, list)
+    local s, n, nCBs = {}, 0, 0
+
+    local function list_entry (name, v) -- upval: taskYield, nCBs
+      print(name, v)
+      n = n + 1
+      if n % 20 == 0 then nCBs = taskYield(nCBs) end
```
nwf: This will want to be adjusted in light of my suggestion below, I think.  `... then taskYield(nil); nCBs = nCBs + 1 end`, I think?

lua_modules/cohelper/README.md
```diff
@@ -0,0 +1,3 @@
+# Coroutine Helper Module
+
+Documentation for this Lua module is available in the [ftpserver.md](../../docs/lua-modules/cohelper.md) file and in the [Official NodeMCU Documentation](https://nodemcu.readthedocs.io/) in `Lua Modules` section.
```
nwf: `ftpserver.md` is a copy-paste-o.

lua_modules/cohelper/cohelper.lua
```diff
@@ -0,0 +1,27 @@
...
+
+local modname = ...
+
+local function taskYieldFactory(co)
+  local post = node.task.post
+  return function(nCBs)  -- upval: co,post
+    post(function () -- upval: co, nCBs
+      coroutine.resume(co, nCBs or 0)
+    end)
+    return coroutine.yield() + 1
```
nwf: I think this should be
```lua
  return function(arg)  -- upval: co,post
    post(function () -- upval: co, arg
      coroutine.resume(co, arg)
    end)
    return coroutine.yield()
```
so as to not be specific to passing around numbers. (i.e., renamed `nCBs` to `arg` and removed the `or 0` and `+ 1`)

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,69 @@
...
+
+This module provides a simple wrapper around long running functions to allow 
+these to execute within the SDK and its advised limit of 15 mSec per individual
+task execution.  It does this by exploiting the standard Lua coroutine 
+functionality as described in the [Lua RM §2.11](https://www.lua.org/manual/5.1/manual.html#2.11) and [PiL Chapter 9](https://www.lua.org/pil/9.html).
+
+The NodeMCU Lua VM fully supports the standard coroutine functionality. Any
+interactive or callback tasks are executed in the default thread, and the coroutine
+itself runs in a second Lua thread. The coroutine can call any library functions,
+but any subsequent callbacks will, of course, execute in the default thread.
```
nwf: There's still only actually one thread in Lua at a time, right?  I'm not sure that this discussion of threading deserves to be so fronted; it's worth keeping, but perhaps later in the document under `## Threading details` and with a disclaimer that it doesn't usually matter?

docs/lua-modules/cohelper.md
TerryE: Yes, but usual documentation is in the case of C modules where in practical terms they are "black boxes" for most developers who never look at the code.  This is really a different kind.  it is a Lua example that we want our developers to look at and understand before using or adapting for their specific needs.  The purpose of `taskYield()` is pretty self evident: you call it to break-off execution and allow in other SDK tasks before resuming.

docs/lua-modules/cohelper.md
TerryE: Yes this would work just as well, but the reason that I do the yield count inside `taskYield()` is just to demonstrate than upvals work across threads. 

So taking you as an example of an advanced NodeMCU Lua developer: you now know that such techniques work fine and this is another tool to add to your toolbox.

lua_modules/cohelper/README.md
TerryE: Yes

lua_modules/cohelper/cohelper.lua
TerryE: Try running it :smile:  You approach works fine, but there are circumstances when you _will_ wantto pass context between the two threads to keep synchronisation.

docs/lua-modules/cohelper.md
TerryE: No there are two definitely two separate Lua threads.  `L` is different for API calls from the these.  But this is standard Lua and explained in the Lua references that I provided.  Not withstanding this just like `node.js` the basic model is not pre-emptive. Execution of a thread is determined by its life or up to a yield / resume exchange in the case of coroutines

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,69 @@
...
+must take care to handle any ordering issues.  This particular example uses
+the `node.task.post()` API withi the `taskYield()`function to resume itself, 
+so the running code can simple call `taskYield()` at regular points in the 
+processing to spilt the work into separate SDK tasks.
+
+A similar approach could be based on timer or on a socket or pipe CB.  If you 
+want to develop such a variant then start by reviewing the source and understanding
+what it does. 
+
+## Use
```
galjonsfigur: I think it would be a good idea to rename this to Usage and add `Syntax`, `Parameters` and `Returns` sections - just to be uniform with the rest of the documentation.

lua_modules/cohelper/cohelper.lua
```diff
@@ -0,0 +1,27 @@
+--[[  A coroutine Helper   T. Ellison,  June 2019
+
+This version of couroutine helper demonstrates the use of corouting within
+NodeMCU execution to split structured Lua code into smaller tasks
+
+]]
+--luacheck: read globals node
```
galjonsfigur: This won't be needed - I forgot about the `task` functions when creating luacheck config and I fixed it in  https://github.com/nodemcu/nodemcu-firmware/pull/2846/commits/990515a2f29afb50fd038f7e49304b7b3ef569c6 in #2846

TerryE: Done

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,87 @@
...
+| :----- | :-------------------- | :---------- | :------ |
+| 2019-07-24 | [TerryE](https://github.com/TerryE) | [TerryE](https://github.com/TerryE)  | [cohelper.lua](../../lua_modules/cohelper/cohelper.lua) |
+
+This module provides a simple wrapper around long running functions to allow 
+these to execute within the SDK and its advised limit of 15 mSec per individual
+task execution.  It does this by exploiting the standard Lua coroutine 
+functionality as described in the [Lua RM §2.11](https://www.lua.org/manual/5.1/manual.html#2.11) and [PiL Chapter 9](https://www.lua.org/pil/9.html).
+
+The NodeMCU Lua VM fully supports the standard coroutine functionality. Any
+interactive or callback tasks are executed in the default thread, and the coroutine
```
HHHartmann: I think it is no good calling this a thread. As the PiL also points out it is only similar. A thread is executed concurrently. So taffeta will get confused. Maybe use "execution context" or whatever.

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,87 @@
...
+itself runs in a second Lua thread. The coroutine can call any library functions,
+but any subsequent callbacks will, of course, execute in the default thread.
+
+Interaction between the coroutine and the parent is through yield and resume 
+statements, and since the order of SDK tasks is indeterminate, the application 
+must take care to handle any ordering issues.  This particular example uses
+the `node.task.post()` API with the `taskYield()`function to resume itself, 
+so the running code can simple call `taskYield()` at regular points in the
```
HHHartmann: Not sure about my English but should it be"simply"

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,87 @@
...
+### Require
+```lua
+local cohelper = require("cohelper")
+-- or linked directly with the  `exec()` method
+require("cohelper").exec(func, <params>)
+```
+
+### Release
+
+Not required.  All resoruces are release on completion of the `exec()` method
```
HHHartmann: released

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,87 @@
...
+
+## `cohelper.exec()`
+Execute a function which is wrapper by a coroutine handler.
```
HHHartmann: wrapped

docs/lua-modules/cohelper.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Syntax
+`require("cohelper").exec(func, <params>)`
+
+#### Parameters
+- `func`: Port number for HTTP server. Most HTTP servers listen at port 80.
```
HHHartmann: What?

docs/lua-modules/cohelper.md
TerryE: > A thread is executed concurrently.

Gregor, I must disagree on this one.  Until we had multi-processor CPUs. threads would never be executed concurrently, since there was only one processing unit.  The WP [thread article](https://en.wikipedia.org/wiki/Thread_(computing)) has a nice definition "a thread of execution is the smallest sequence of programmed instructions that can be managed independently by a scheduler".  Features such as concurrence and pre-emption are more emergent benefits of this independence.   In our Lua world the _machine_ is the Lua VM and this must, like node.js, run in a single OS thread.  

What I am trying to do is to explain how this works to the average IoT developer.  If you can suggest better wording then I'll adopt this.

docs/lua-modules/cohelper.md
TerryE: And resources

docs/lua-modules/cohelper.md
TerryE: Sorry, cut and past crud.  Needs removing.

TerryE:
I will leave it for another day or for any other feedback before pushing the updates.

docs/lua-modules/cohelper.md
nwf: The term to use, I think, is "cooperative multi-threading", assuming I am correct that, on our C substrate, a Lua task will run to completion, without preemption and without concurrent access to the Lua VM and its heap, even if tasks are being dispatched by multiple threads.  I'm not sure that *this* module's documentation is the right place to spell it out; perhaps `node.task.post()` is the right place and this should simply cite that.

TerryE: I've used stack instead of thread.

marcelstoer:
I just aded it to the milestone...

PR #2852, opened 2019-07-25
Fix enduser_setup default POST request

### LINKED ISSUE #2847: enduser_setup not working
# Expected behavior
the `POST /setwifi` call shpould parse the params

### Actual behavior
1. The end of the wifi_password param is searched to be & or space. In the post it is not anymore but the end of the message
2. if the headers and body are fragmented in several packets only the first one is analyzed and thus data is incomplete. This applies to the GET request as well but likely will nerver hit there.
At least fiddler (a windows web analyzer tool) does send headers and body in two packets.

(1.) works well on params in the URL as used in the GET request. They are terminated by '&' or 'space' 
``` http
GET /url?wifi_ssid=ssid&wifi_password=pass HTTP/1.1
```
The new POST locks loke this:
``` http
POST http://aaa.com/setwifi HTTP/1.1
Host: aaa.com
Content-Type: application/x-www-form-urlencoded
Accept-Encoding: gzip, deflate
Content-Length: 30

wifi_ssid=xxx&wifi_password=aa
```
So the wifi_password is terminated by the end of data here.

### Test code
compile with 
``` c
#define ENDUSER_SETUP_DEBUG_ENABLE 1
#define ENDUSER_SETUP_DEBUG_SHOW_HTTP_REQUEST 1
```
Example from the documentation
```Lua
enduser_setup.start(
  function()
    print("Connected to WiFi as:" .. wifi.sta.getip())
  end,
  function(err, str)
  end,
  print -- Lua print function can serve as the debug callback
)
```

### NodeMCU version
Current dev build with the brand new enduser_setup

### Hardware
ESP8266 dev board

HHHartmann:
partially fixes #2847.

**Rational**
the out of the box enduser_setup does not work as the param parsing fails.
It assumes to have a ' ' or '&' to terminate the params.
While this is true for the GET request (where the params are part of the URL which is followed by "&nbsp;HTTP/1.1" it does not hold for POST requests, where the message ends with the last char of the params.
```http
GET /url?wifi_ssid=ssid&wifi_password=pass HTTP/1.1
```
This fix returns a false positive if the request is malformed and does not have a space to terminate the URL.
But since message parsing is not very strict anyhow that should be OK (for now)

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -397,14 +397,20 @@ static err_t close_once_sent (void *arg, struct tcp_pcb *pcb, u16_t len)
 /* ------------------------------------------------------------------------- */
 
 /**
- * Get length of param
+ * Get length of param value
+ * 
+ * This is being called with a fragment of the parameters passed in the 
+ * URL for GET requests or part of the body of a POST request.
+ * The string will look like one of these
+ * "SecretPassword HTTP/1.1"
+ * "SecretPassword&wifi_ssid=..."
+ * "SecretPassword"
+ * The string is searched for the first occurence of deliemiter '&' or ' '.
+ * If found return the length up to that position.
+ * If not found return the length of the string.
  *
- * Search string for first occurence of deliemiter '&' or ' '.
- * if found that terminates the param, if not end of string does.
- *
- * @return -1 if no occurence of char was found.
  */
-static int enduser_setup_get_len(const char *str)
+static int enduser_setup_get_lenth_of_param_value(const char *str)
 {
   char *found = strpbrk (str, "& ");
   if (!found)
@@ -820,8 +826,8 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
   char *name_str_start = name_str + name_field_len;
   char *pwd_str_start = pwd_str + pwd_field_len;
 
-  int name_str_len = enduser_setup_get_len(name_str_start);
-  int pwd_str_len = enduser_setup_get_len(pwd_str_start);
+  int name_str_len = enduser_setup_get_lenth_of_param_value(name_str_start);
+  int pwd_str_len = enduser_setup_get_lenth_of_param_value(pwd_str_start);
 
 
   struct station_config *cnf = luaM_malloc(lua_getstate(), sizeof(struct station_config));
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -397,15 +397,16 @@ static err_t close_once_sent (void *arg, struct tcp_pcb *pcb, u16_t len)
 /* ------------------------------------------------------------------------- */
 
 /**
- * Search String
+ * Get length of param
  *
- * Search string for first occurence of any char in srch_str.
+ * Search string for first occurence of deliemiter '&' or ' '.
+ * if found that terminates the param, if not end of string does.
  *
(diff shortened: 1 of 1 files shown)


HHHartmann:
@marcelstoer Marcel, I agree that the fix ist not perfect. That's the problem with comments, they always are not adapted to code changes. But that's a different discussion.

A Proper fix would IMHO be to extract the params in the respective GET and POST handler and pass only this substring `wifi_ssid=ssid&wifi_password=pass` to `enduser_setup_http_handle_credentials`.

I could also leave `enduser_setup_srch_str` as is and add some "not found" handling to the calling function.
In any case this should go in the next release together with #2827 and #2810 or none of them.

marcelstoer:
> A Proper fix would IMHO be to extract the params in the respective GET and POST handler and pass only this substring `wifi_ssid=ssid&wifi_password=pass` to `enduser_setup_http_handle_credentials`.

True, that sounds like a good plan - as long as we keep in mind that the string may contain other `key=value` pairs in arbitrary order.

> In any case this should go in the next release together with #2827 and #2810 or none of them.

👍 IMHO rolling back is not an option, forward-fixing is.

HHHartmann:
This should be fixed in a better way now.
Also found and fixed a memory leak closing a file (in rare circumstances)

app/modules/enduser_setup.c
```diff
@@ -397,18 +397,19 @@ static err_t close_once_sent (void *arg, struct tcp_pcb *pcb, u16_t len)
 /* ------------------------------------------------------------------------- */
 
 /**
- * Search String
+ * Get length of param
  *
- * Search string for first occurence of any char in srch_str.
+ * Search string for first occurence of deliemiter '&' or ' '.
```
marcelstoer: I think there are few naming/linguistic issues. All of them tiny but in combination this isn't easy to comprehend:

- I don't understand the second sentence in this comment.
- `@return -1 if...` - no longer true
-  `get_len(const char *str)` is extremely generic, can the function and the param be named more precisely?
-- `*str` is expected to be a "value&key=..." string, no?

HHHartmann: ok. To avoid more checkins:

``` c
/**
 * Get length of param value
 * 
 * This is being called with a fragment of the parameters passed in the 
 * URL for GET requests or part of the body of a POST request.
 * The string will look like one of these
 * "SecretPassword HTTP/1.1"
 * "SecretPassword&wifi_ssid=..."
 * "SecretPassword"
 * The string is searched for the first occurence of deliemiter '&' or ' '.
 * If found return the length up to that position.
 * If not found return the length of the string.
 *
 */
static int enduser_setup_get_lenth_of_param_value(const char *str)
```

KT819GM:
Can anyone confirm, that fix is applied to master branch as of (310faf7)? As seems it migrated to SDK3.0.0 (sorry! missed that somehow till today) but EUS does not save credentials. Also 'smartconfig' does not work anymore with that branch.
t.y.

HHHartmann:
Are you using an enduser_setup.html file on the SPIFFS Filesystem or the built in version? 
The call to store the data has changed.

KT819GM:
Omg, answered so fast,before I deleted question by seeing that I'm posting into pull request. I'm using builtin version, will try to read how to use it properly, thought so, that I did not saw all the changes. Thank you

Issue #2853, opened 2019-07-25
Can't compile in Docker after PR #2845


vsky279:
### Expected behavior
After PR #2845 and PR #2844 (probably only one of them being toxic) I am not able to compile `dev` version with Docker (Windows 10).

### Actual behavior
```
CC app/user/user_main.c
In file included from user_main.c:25:0:
../../sdk-overrides/include/espconn.h:1:1: error: expected identifier or '(' before '.' token
 ../../app/include/lwip/app/espconn.h
 ^
user_main.c: In function 'user_process_input':
user_main.c:298:32: error: 'input_sig' undeclared (first use in this function)
     return task_post_low(input_sig, force);
                                ^
user_main.c:298:32: note: each undeclared identifier is reported only once for each function it appears in
user_main.c: In function 'user_init':
user_main.c:340:5: error: 'input_sig' undeclared (first use in this function)
     input_sig = task_get_id(handle_input);
     ^
make[2]: *** [.output/eagle/debug/obj/user_main.o] Error 1
make[2]: Leaving directory `/opt/nodemcu-firmware/app/user'
make[1]: *** [.subdirs] Error 2
make[1]: Leaving directory `/opt/nodemcu-firmware/app'
make: *** [.subdirs] Error 2
make: Leaving directory `/opt/nodemcu-firmware'
```

### Test code
No test code is needed. I just can't compile clean pull of `dev` version with the error above.

### NodeMCU version
`dev`

### Hardware
N/A

HHHartmann:
Can confirm

marcelstoer:
> PR #2844

How is that one related? It only changed documentation for the ESP32 branch.

marcelstoer:
@HHHartmann since the NodeMCU CI build is obviously ok do you think this may be related to trusty vs xenial (and therefore related to https://github.com/marcelstoer/docker-nodemcu-build/pull/69)?

vsky279:
> > PR #2844
> 
> How is that one related? It only changed documentation for the ESP32 branch.

Then it is not related. Of course. 
Sorry I did not go into detail. I just saw that after these two PRs compile fails.

HHHartmann:
seems to be that sdk-overrides/include/espconn.h is a link which does not get checked correctly out under windows as mentioned elsewhere. It contains 
```
../../app/include/lwip/app/espconn.h
```

HHHartmann:
[...]

PR #2854, opened 2019-07-25
net_info module - ping function

### LINKED ISSUE #1555: expose ping function to lua

vsky279:
Fixes #1555.

This PR is created based on initial module by @wolfgangr with only few fixes. It seems to be pity to drop the work that has been done by @wolfgangr.
This version is working reliably in one of my applications so I think it is eligible to be merged into `dev` version.
Once reviewed I will squash these commits into one.
Actually can I do it just for the PR and have them separated in my branch?


marcelstoer:
Regardless of whether this will ever be merged...the CI build failed. Not sure if you have access to this https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/563711700

```
net_info.c:14:22: fatal error: c_stdlib.h: No such file or directory
 #include "c_stdlib.h"
                      ^
compilation terminated.
../../Makefile:417: recipe for target '.output/eagle/debug/obj/net_info.o' failed
make[2]: *** [.output/eagle/debug/obj/net_info.o] Error 1
make[2]: Leaving directory '/home/travis/build/nodemcu/nodemcu-firmware/app/modules'
../Makefile:380: recipe for target '.subdirs' failed
make[1]: *** [.subdirs] Error 2
make[1]: Leaving directory '/home/travis/build/nodemcu/nodemcu-firmware/app'
Makefile:380: recipe for target '.subdirs' failed
make: *** [.subdirs] Error 2
make: Leaving directory '/home/travis/build/nodemcu/nodemcu-firmware'
The command "if [ "$OS" = "linux" ]; then bash "$TRAVIS_BUILD_DIR"/tools/travis/ci-build-linux.sh; fi" exited with 1.
```

That is due to #2838.

app/modules/net_info.c
```diff
@@ -53,6 +53,7 @@
 static int ping_callback_ref; 
 static int ping_host_count;
 static ip_addr_t ping_host_ip;
+struct ping_option *ping_opt = NULL;
```
nwf: This is a pretty unpleasant workaround for object lifetime issues and it implies that only one `ping` can be going at once.

Looking at the LwIP code, it appears to me that you should be able to properly manage the object lifetime by hooking both the *sent and received* callbacks in the `struct ping_option` you pass to `ping_start`.

vsky279:
I will fix it when #2853 is solved. I can't compile now :-(

app/modules/net_info.c
```diff
@@ -0,0 +1,201 @@
...
+      lua_State *L = lua_getstate();
+      lua_rawgeti(L, LUA_REGISTRYINDEX, ping_callback_ref);
+      lua_pushinteger(L, pingresp->bytes);
+      lua_pushstring(L, ipaddrstr);
+      lua_pushinteger(L, pingresp->seqno);
+      lua_pushinteger(L, pingresp->ttl);
+      lua_pushinteger(L, pingresp->resp_time);
+      lua_call(L, 5, 0);
+    } else {
+      c_printf("%d bytes from %s, icmp_seq=%d ttl=%d time=%dms\n",
```
nwf: While I appreciate this as a form of debugging, I don't think it's an appropriate API for ping with no callback to spew to the console.  I think the callback should be mandatory and you can provide an example in the docs of a Lua function that prints to the console.

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+function ping_recv(bytes, ipaddr, seqno, ttl, ping)
+```
+* length of message
+* ip-address of pinged host
+* icmp_seq number
+* time-to-live-value
+* ping time in ms
+
+!!! Caution
+    The callback functionality is still untested. Use at even more your own risk!
```
nwf: This must be tested before merge.

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+
+For further reference to callback functionality, see smcl origin link provided on top of this page.
```
nwf: I think we would prefer to have documentation in-tree rather than on a blog that can go away at any moment.

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+
+The callback receives the following arguments:  
+```
+function ping_recv(bytes, ipaddr, seqno, ttl, ping)
+```
+* length of message
+* ip-address of pinged host
+* icmp_seq number
+* time-to-live-value
+* ping time in ms
```
nwf: Please name it something other than `ping`; `rtt` perhaps.

docs/en/modules/net_info.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+Ping **callback function**:
+```
+net_info.ping(host, count, callback)
+```
+Instead of printing ping results on the console, the given callback function ist called each time a ECHO_RESPONSE is received.
+
+The callback receives the following arguments:  
+```
```
nwf: The callback should also be called for timeouts, which are perhaps best signaled by `icmp_seq` and subsequent fields being `nil`.

vsky279:
> Regardless of whether this will ever be merged...the CI build failed. Not sure if you have access to this https://travis-ci.org/nodemcu/nodemcu-firmware/jobs/563711700

Ok, so at least it compiles now.

Nathan @nwf  thanks for all suggestions. Progressively I will try to implement them. It will be pending for some while as I'm on another project now.

Stay tuned.

app/modules/net_info.c
```diff
@@ -0,0 +1,197 @@
...
+#else
+  #define NET_INFO_ERR(...)
+#endif
+
+
+#define NET_INFO_PRIORITY_OUTPUT   TASK_PRIORITY_MEDIUM
+#define NET_INFO_PRIORITY_ERROR    TASK_PRIORITY_MEDIUM
+
+// these statics should go away for reentrancy and maybe other reasons!
+static int ping_callback_ref;
```
nwf: You can use the oddly-named `reverse` field of the `struct ping_option` for this, rather than it being global, so that there can be multiple pings outstanding at once.

app/modules/net_info.c
```diff
@@ -0,0 +1,197 @@
...
+static int ping_host_count;
+static ip_addr_t ping_host_ip;
+struct ping_option *ping_opt = NULL;
+
+void ping_received(void *arg, void *data) {
+    // this would require change of the interface 
+    // struct ping_msg *pingmsg = (struct ping_msg*)arg;
+    // struct ping_option *pingopt = pingmsg->ping_opt;
+    struct ping_option *pingopt = (struct ping_option*)arg;
+    struct ping_msg *pingmsg = pingopt->parent_msg ;
```
nwf: There doesn't appear to be such a field in our LwIP (https://github.com/nodemcu/nodemcu-firmware/blob/dev/app/include/lwip/app/ping.h), so I'm not sure what's going on here.

app/modules/net_info.c
```diff
@@ -0,0 +1,197 @@
...
+
+    struct ping_resp *pingresp = (struct ping_resp*)data;
+
+    char ipaddrstr[16];
+    ip_addr_t source_ip;
+    
+    source_ip.addr = pingopt->ip;
+    ipaddr_ntoa_r(&source_ip, ipaddrstr, sizeof(ipaddrstr));
+
+    if (ping_callback_ref != LUA_NOREF) {
```
nwf: If this is stashed in `pingopt->reverse` (or in something that that points at, if you need more state, but I don't think you do), then there's no need for this check here, just down below when the ping is constructed.

app/include/lwip/app/ping.h
```diff
@@ -54,6 +54,7 @@ struct ping_option{
 	ping_recv_function recv_function;
 	ping_sent_function sent_function;
 	void* reverse;
+	struct ping_msg *parent_msg;
```
nwf: I don't believe this is necessary; the `reverse` field right above it should serve admirably for the same purpose?

docs/en/modules/net_info.md
vsky279: `lwip/ping.c` calls the callback also for timeouts with `bytes` (received) = 0. Other fields are 0 too (except for `ipaddr`). This seems reasonable to me.

app/modules/net_info.c
```diff
@@ -0,0 +1,228 @@
...
+	ping_t nip = (ping_t)luaL_checkudata(L, stack, "net_info.ping");
+	if (nip == NULL)
+		return (ping_t)luaL_error(L, "ping object expected");
+	return nip;
+}
+  
+void ping_received(void *arg, void *data) {
+    NODE_DBG("[ping_received] ");
+    // struct ping_option *pingopt = (struct ping_option*)arg;
+    ping_t nip  = (ping_t)arg;
```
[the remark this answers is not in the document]
vsky279: Yep. More detailed comment added.

vsky279:
Ok. I have a version that is working somehow.

Testing code:
```Lua
    local i = 0
    tmr.create():alarm(4000, tmr.ALARM_AUTO, function() 
        local ni = net_info.create()
        ni:ping("192.168.1." .. i%255+1, 4, function (b, ip, sq, tm) 
            print(string.format("%d bytes from %s, icmp_seq=%d time=%dms heap=%d", b, ip, sq, tm, node.heap())) 
        end)
        i = i + 1
    end)
```

It appears that when the host does not respond then random WDT resets occur. More study is needed - this is probably somewhere in LWIP code.

When all hosts are reachable the code is stable and I don't notice any memory leaks.

When trying to shorten the interval to have concurrent pings. I get WDT reset too. But also when 
ping to 192.168.1.2 is started, I don't get any good response for previously started ping (192.168.1.1). Response is the same as if the destination is unreachable.

I wonder whether the LWIP/ping is able to service concurrent requests.

marcelstoer:
@nwf was your input adequately addressed?

nwf:
I think this one shouldn't be merged until we understand what's causing

> It appears that when the host does not respond then random WDT resets occur. More study is needed - this is probably somewhere in LWIP code.

I'll see if I can scare up some time for it.

marcelstoer:
Arggg, sorry I missed this one 😡. Thanks for being more diligent than I was.

app/modules/net_info.c
```diff
@@ -0,0 +1,189 @@
...
+    if (lua_isnumber(L, 3)) {
+        count = luaL_checkinteger(L, 3);
+    }
+   
+    // retrieve callback arg
+    NODE_DBG("[net_info_ping] nip->ping_callback_ref = %p\n", nip->ping_callback_ref);
+    if (nip->ping_callback_ref != LUA_NOREF) {
+      luaL_unref(L, LUA_REGISTRYINDEX, nip->ping_callback_ref);
+    }
+    nip->ping_callback_ref = LUA_NOREF;
```
HHHartmann: This could go inside the if brackets if I am not mistaken.

app/modules/net_info.c
```diff
@@ -0,0 +1,189 @@
...
+    NODE_DBG("[net_info_unregister] luaL_unref nip->self_ref (%p, %p)\n", nip, nip->self_ref);
+    luaL_unref(L, LUA_REGISTRYINDEX, nip->self_ref);
+    nip->self_ref = LUA_NOREF;
+  }
+
+  if(nip->ping_callback_ref != LUA_NOREF) {
+    NODE_DBG("[net_info_unregister] luaL_unref nip->ping_callback_ref (%p)\n", nip->ping_callback_ref);
+    luaL_unref(L, LUA_REGISTRYINDEX, nip->ping_callback_ref);
+  }
+  nip->ping_callback_ref = LUA_NOREF;
```
HHHartmann: Should also go inside the if-body

app/modules/net_info.c
```diff
@@ -0,0 +1,189 @@
...
+    lua_State *L = lua_getstate();
+    lua_rawgeti(L, LUA_REGISTRYINDEX, nip->ping_callback_ref);
+    lua_pushinteger(L, pingresp->bytes);
+    lua_pushstring(L, ipaddrstr);
+    lua_pushinteger(L, pingresp->seqno);
+    lua_pushinteger(L, pingresp->resp_time);
+    
+    (nip->count)--;
+    if (!nip->count) { // is this a last call?
+      if (nip->self_ref != LUA_REFNIL) {
```
HHHartmann: These two if blocks repeat in `net_info_unregister` and could be refactored in a separate function.

vsky279:
Sorry, I missed your comments Gregor @HHHartmann.
I like your proposal. I will deal with it in coming days.

vsky279:
Sorry, I missed your comments Gregor @HHHartmann.
I like your proposal. I will deal with it in coming days.

nwf:
@HHHartmann As per the comments above, I understood that the WDT resets do not appear to be limited to the case of only one ping at a time.  That is pretty strong evidence for something broken with our use of LwIP (and as you may have seen, I think Espressif really @#(*&*&^#*&!@ up more or less everything related, as per #3004 and #3006 and friends).

HHHartmann:
@nwf yes indeed when writing the suggestion the watchdog had escaped my mind.

app/include/user_modules.h
```diff
@@ -39,6 +39,7 @@
 //#define LUA_USE_MODULES_MDNS
 #define LUA_USE_MODULES_MQTT
 #define LUA_USE_MODULES_NET
+#define LUA_USE_MODULES_NET_INFO
```
marcelstoer: By default all new modules are disabled.

app/modules/net_info.c
vsky279: done

app/modules/net_info.c
vsky279: done

app/modules/net_info.c
vsky279: done

app/include/user_modules.h
vsky279: Yep. This will be fixed. I'm afraid we are still far from merging this module...

clearisoft:
Random crash when ping.

Cause:
Call unregister_refs() too early.

Solution:
static void ping_received(void *arg, void *data) {
...
(nip->count)--;
---if (!nip->count) { // is this a last call?
--- unregister_refs(L, nip);
---}

lua_call(L, 4, 0);
}

+++static void ping_sent(void *arg, void *data) {
+++ ping_t nip = (ping_t)arg;
+++ lua_State *L = lua_getstate();
+++ unregister_refs(L, nip);
+++}

....
nip->ping_opt.recv_function = &ping_received;
+++nip->ping_opt.sent_function = &ping_sent;
....

nwf:
@clearisoft Please reformat that as an actual unified diff and a markdown code block.

nwf:
@vsky279 Looking at this patch again... why the name `net_info`?  Why a persistent object with a `:ping()` method?  Is that better than just

    net.ping("ip address", function(size, ip, seqno, time) ... end)

Somewhat separately, the callback might want to take a self-ref so that we could add a `stop` method, and `net.ping` might return something so that someone else could call `stop`.  I'll work up a draft proposal the next time I have a minute if you don't beat me to it:

    pinger = net.ping("ip address", function(self, size, seqno, time) ... --[[ use self.ip if desired ]]-- end)
    if ... then pinger:stop() end

HHHartmann:
> @vsky279 Looking at this patch again... why the name `net_info`? Why a persistent object with a `:ping()` method? Is that better than just
> 
> ```
> net.ping("ip address", function(size, ip, seqno, time) ... end)
> ```

I was also thinking of combining it with the `net.ifinfo` somehow.

Maybe have `net.info.ping` and `net.info.interface`. But your suggestion is fine too. Just wanted to group it in case more info stuff appears (traceroute or whatever).
But it sure feels like it belongs to the `net` module.

vsky279:
Gentlemen, thanks for bringing this issue back to life.
I will test the bug fix suggested by @clearisoft.

> Why a persistent object with a :ping() method?

I though that was the right way to cope with your (reasonable) requirement:
> I would like to be able to have multiple ping requests going at once (as LwIP supports) and see proper C object life-cycle management.

Then
```Lua
net_info.create():ping("ip address", 4, function(size, ip, seqno, time) ... end)
```
seemed to be the right way how to cope with multiple ping requests.
I also thought the object is not persistent but rather dynamic when its destroyed by the `net_info_unregister` method.

>```stop``` method

Do we need this approach? To me typical usage is to ping some server with fixed number of packets. I have no use case where the ```stop``` method would be needed.

> why the name net_info

It's kinda legacy as the module is based on initial work of @wolfgangr (#1555). But it seemed ok to me I `net` is a basic package  and it seemed to be reasonable to keep it compact.

HHHartmann:
> Gentlemen, thanks for bringing this issue back to life.
> I will test the bug fix suggested by @clearisoft.

> 
> > why the name net_info
> 
> It's kinda legacy as the module is based on initial work of @wolfgangr (#1555). But it seemed ok to me I `net` is a basic package and it seemed to be reasonable to keep it compact.

The module could be a submodule with separate configuration. Just like `net.monitor`.

But I would vote for bringing this to dev if it works and not have it sleep for another couple of years.
We might later on move it to net.

nwf:
There's no need to create an object like that to have multiple pings in flight at once.  It should work to

    net.ping("4.4.4.4", function() ... end)
    net.ping("4.4.8.8", function() ... end)

Behind the scenes there will be userdata created for each ping, but that's not exactly the same as needing a `.create()` call in Lua.

nwf:
I am skeptical that 4f1e7dd is really the right fix... surely `ping_sent` gets called before `ping_received`?

nwf:
Also ick, ew... LwIP's `ping_start` calls into `malloc` and `ping_send` just silently does nothing if it can't allocate a packet buffer.  This is no way to write good code; I think we should replace `app/lwip/app/ping.c` before adding this module into our tree or we're just inviting trouble.

vsky279:
Okey doke. So there's a revamp of the `net_info` module.

I agree that the object approach was a bit overkill in the case where there was only one method. So now we have only `net_info.ping()` function. No random crashes anymore.

Multiple pings can start in short sequence thought if the new ping overlaps with the previous one the first stops receiving answers, i.e.
```Lua
function ping_resp(b, ip, sq, tm)
  print(string.format("%d bytes from %s, icmp_seq=%d time=%dms", b, ip, sq, tm))
end

net_info.ping("8.8.8.8", 4, ping_resp)
tmr.create():alarm(1000, tmr.ALARM_SINGLE, function() net_info.ping("8.8.4.4", 4, ping_resp) end)
```
gives
```
32 bytes from 8.8.8.8, icmp_seq=9 time=14ms
32 bytes from 8.8.8.8, icmp_seq=10 time=9ms
32 bytes from 8.8.4.4, icmp_seq=11 time=6ms
32 bytes from 8.8.4.4, icmp_seq=13 time=12ms
0 bytes from 8.8.8.8, icmp_seq=0 time=0ms
32 bytes from 8.8.4.4, icmp_seq=15 time=16ms
0 bytes from 8.8.8.8, icmp_seq=0 time=0ms
32 bytes from 8.8.4.4, icmp_seq=16 time=7ms
```

I don't observe any memory leaks so I'm not sure we need to replace `app/lwip/app/ping.c` before merging it. This seems to work.

The question is whether this should be implemented in a same was as `net.monitor` in the `net` module. It's up to you to decide.

Feel free to test and comment.

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+
+#include "lwip/ip_addr.h"
+#include "espconn.h"
+#include "lwip/dns.h" 
+#include "lwip/app/ping.h"
+#include "lwip/raw.h"
+
+/* 
+ping_opt needs to be the first element of the structure. It is a workaround to pass the 
+calllback reference to the ping_received function. Pointer the ping_option structure is
```
nwf, line 24: "callback" (one too many l-s).

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+equal to the pointer to net_info_ping_t strcture. Reference to callback function is assumed
+to be just after the ping_option structure.
```
nwf, line 26: The "is assumed" sentence is quite worrying and suggests that there's some pretty evil pointer math going on?

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+*/
+typedef struct {
+  struct ping_option ping_opt; 
+  sint32_t ping_callback_ref; 
+  uint32 count;
```
nwf, line 31: This field appears unused except in debugging.  The callback gets the `pingresp->seqno` not `count`.

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+      luaL_unref(L, LUA_REGISTRYINDEX, nip->ping_callback_ref);
+      nip->ping_callback_ref = LUA_NOREF;
+    }
+    NODE_DBG("[ping_sent] free nip (%p)\n", nip);
+    free(nip);
+}
+
+static int net_info_ping(lua_State *L)
+{
+    ping_t nip = malloc( sizeof(net_info_ping_t));
```
nwf, line 74: I recognize that you're not exposing the `ping_t` structures to Lua, but please allocate a userdata for it all the same.  Give it a `self_ref` in the registry and a `__gc` method that releases the callback function ref.  That way, we can rely on Lua's GC to do memory management for us.

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+    unsigned count = 4;
+     
+    // retrieve address arg (mandatory)
+    if (lua_isstring(L, 1)) {
+        ping_target = luaL_checkstring(L, 1);
+    } else {
+        return luaL_error(L, "no address specified");
+    }
+
+    // retrieve count arg (optional)
```
nwf, line 90: Usually "optional" arguments means "isn't required to be specified" not "might be nil".  That is, it makes it sound like I could write

   net_info.ping("4.2.2.2", function(...) ... end)

but as written, the module requires that I write 

   net_info.ping("4.2.2.2", nil, function(...) ... end)

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+       
+    if (lua_type(L, 3) == LUA_TFUNCTION || lua_type(L, 3) == LUA_TLIGHTFUNCTION) {
+      nip->ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+    } else {
+        return luaL_error(L, "no callback specified");
+    }
+    NODE_DBG("[net_info_ping] nip->ping_callback_ref = %p\n", nip->ping_callback_ref);
+    
+    // attempt to parse ping target as IP
+    uint32 ip = ipaddr_addr(ping_target);
```
nwf, line 110: If it's not *too* much hassle, it'd be nice if this module understood DNS, too.  #3063 was rejected on usability grounds, so I suppose the same arguments apply here.

app/modules/net_info.c
```diff
@@ -0,0 +1,133 @@
...
+    lua_pushstring(L, ipaddrstr);
+    lua_pushinteger(L, pingresp->seqno);
+    lua_pushinteger(L, pingresp->resp_time);
+    
+    (nip->count)--;
+    
+    lua_call(L, 4, 0);
+}
+
+static void ping_sent(void *arg, void *data) {
```
nwf, line 59: I am *astounded* that `ping_sent` is called after all the `ping_received` but so it goes.  In which case, it looks like the reference counting issue is indeed fixed.

app/modules/net_info.c
vsky279: This was not precise. I have modified the comment to
```c
/* 
ping_opt needs to be the first element of the structure. It is a workaround to pass the 
calllback reference and self_ref to the ping_received function. Pointer the ping_option 
structure is equal to the pointer to net_info_ping_t structure.
*/
```
The order of other fields is unimportant. `ping_option` just needs to be the first element.

app/modules/net_info.c
vsky279: You're right. The field is not needed unless someone needed to know the counter of current ping. `seqno` increments across all performed pings. It can be reintroduced if needed but is removed for now.

app/modules/net_info.c
vsky279: This is for me the most tricky part as I'm still newbie in working with Lua in c code. I'm rather reusing what's in other modules. The unusual part is not to expose the userdata/object to Lua but to keep the reference only in module. So I hope I manage properly the lifetime of the userdata structure:
```c
static int net_info_ping(lua_State *L)
{
    // ping_t nip = malloc( sizeof(net_info_ping_t));
    ping_t nip = (ping_t)lua_newuserdata(L, sizeof(net_info_ping_t));
    if (!nip) return luaL_error(L, "not enough memory");
    luaL_getmetatable(L, "net_info.ping");
    lua_setmetatable(L, -2);
    nip->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
...
```
The `luaL_unref` and `__gc` implementation is quite clear to me then. This does not seem to raise any exception nor mem leaks.

app/modules/net_info.c
vsky279: I'll handle this in a second round.

TerryE:
@vsky279 Lukàš, I am apologise for coming to this review late -- a combination of my own health problems and trying to get my head down to do the Lua 5.3 port.  

### General observations

My overall feeling about this module is that it exposes a systemic issue that we have with most of our modules and that relates to how we develop them.  Being a systemic issue, this is really a subject for a separate issue, but perhaps I can use this implementation to discuss what concerns me here.

IMO, the 'Lua approach' to developing its VM runtime and library modules is quite different to typical C development.  
-  The Lua implementation is very orthogonal in that in general there is only a single approach to address any architectural feature and all of the code uses that approach for that feature.  
-  There is little attempt to do peephole optimisation of the implementation (such as manual code unrolling).  It is assumed that leaving such issues to the compiler as this will do a better job in most cases.
-  The overall error strategy is one of 'throw and catch' error handling.  Very few routines return an error status that assumes that the wrapping routine will test for this and handle the error correctly, simply because:
    -  This error test and pass up generates large volumes of error handling source code that confuses the reader with noise and is difficult to cover in unit testing.
    -  The major bug with with strategy is that at some level the programmer doesn't bother to do the testing and simply assumes that the lower level routine works, resulting in downstream bizarre and difficult to track errors (ironically your code includes a good example of this which I discuss below).
-  In terms of modules interfacing to the Lua runtime there is a single `lua.h` interface.  There is also a rich set of helper routines which are declared in `lauxlib.h` and which are layered on the Lua API and provide macro operations intended to reduce the code volume of module code.

If anyone wants to see this implementation strategy in action look at these headers, plus the `lapi.c` implementation, plus the `lauxlib.c` rapper routines and the various `lXXXlib.c` Lua core libraries. Anyone comparing and contrasting this last set to our C modules will see that our modules are a example of following this strategy.  The main reason for this was that the initial project implementers were really IoT C developers who decided on the Lua language and runtime, but because they were working against tight deadline they didn't learn the Lua approach before implementing the first core set of modules, so these are poor example for any new modules developers to follow.

One of my TODO items is to go through these and tidy up their implementations, but this has been a TODO for the last two years.  Understandably but also unfortunately your module here looks more like it has understandably followed some of these poor practices rather than those used in the `l*lib.c` sources.

Some some specific examples.  These use your code as a reference because this is the scope of this PR, but to be honest I could also give examples littered across the core modules that the project maintain.

## Lua API calls nearly always throw an error if an error is detected

See [LRM §3.6](https://www.lua.org/manual/5.1/manual.html#3.6) and the following §3.7 in particular the use of the `[-o, +p, x]` designator: very few access abnd check function have an **X == -** error designator.

> The third field, x, tells whether the function may throw errors: '-' means the function never throws any error; 'm' means the function may throw an error only due to not enough memory; 'e' means the function may throw other kinds of errors; 'v' means the function may throw an error on purpose. 

So there is absolutely no point in testing for an error since the code will never reach this path if an error was raised so:
```C
    ping_t nip = (ping_t)lua_newuserdata(L, sizeof(net_info_ping_t));
    if (!nip) return luaL_error(L, "not enough memory");
```
`nip` can only point to the Userdata.  An error is thrown if one (such as out of memory) is detected so the `ìf(!nip` branch will never be taken.  Get rid of this type of dead error branch code.

## Use the `lauxlib.c` wrapper API when you can and don't reimplement inline
So  with:
```C
if (lua_isstring(L, 1)) {
        ping_target = luaL_checkstring(L, 1);
    } else {
        return luaL_error(L, "no address specified");
    }

    unsigned cbpar = 2;
    if (lua_isnumber(L, 2)) {
        count = luaL_checkinteger(L, 2);
        cbpar++;
    }
   
    // retrieve callback arg
    if (lua_type(L, cbpar) == LUA_TFUNCTION || lua_type(L, cbpar) == LUA_TLIGHTFUNCTION) {
      nip->ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
    } else {
        return luaL_error(L, "no callback specified");
    }
```
What happens with `net_info.pig(f,23)`?  KISS, please.
```C
   const char *ping_target = luaL_checkstring(L, 1);
   lua_settop(L, lua_isfunction(L, 2) ? 2 : 3);
   lua_Integer count = luaL_optinteger(L, 2, 4);
   luaL_argcheck(L, lua_isfunction(L, -1), -1, "no callback specified");
   nip->ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```

## Understand the broad memory allocation strategy

Lua API calls (and Lua modules) should throw a memory error on OOM.
-  So always check any system call which can return a mallow error status and throw the error.
-  Never catch a OOM and return as a status.  Lua code should use a `pcall()` to catch these.

```C
  if (!ping_start(&(nip->ping_opt)))
    return luaL_error(L, "memory allocation error: cannot start ping");
```

BTW, the `ping_*` routines malloc and free the `ping_t` buffer so this module should _never_ free this.

I see that you've cross-pushed an update which covers some of these points, but the correct method of allocating persistent resources for use in C modules is to use Userdata and not to use malloc.  Userdata is a Lua collectible class and so the Lua GC will automatically handle any memory clean-up with needing extra application coding.  

We also need to design in the proper scope and life of Userdata resources to prevent their premature Lua GC.  So in this case, I believe that the `ping__sent()` CB is fired once per ping and the response (if any) is processed by the `ping_rcv()` (which we don't hook into for some reason).  These will be called `count` times so these routines certainly shouldn't do any resource clean up _per call_.  We need to do this scope by design, so we need to consider the usecase variants.

In this case the net_info_ping() function create the Userdata on the stack but it isn't hooked into the GC sweeplist anywhere and so will be GC'ed.  We can make it non-colectible in use either by:
-  Adding the Userdata as upvalue to the registered sweep function in which case we need some form of ping timeout / count to zero fuunction so the CB can remove the upval reference to allow the LGC to collect it, or
-  the `net_info.ping()` returns the UD as a "ping socket" and the app can issue a `sk:close()`, but again we need to work through the temporal dependencies.

## Miscellaneous

I know that the standard pattern is to register Userdata metatables at startup, but another approach is to drop the module open entry and do this lazily on first use, for example:

```C
    if (!luaL_getmetatable(L, "net_info.ping"))
      luaL_rometatable(L, "net_info.ping", LROT_TABLEREF(net_info_dyn));
```
Smaller code size and the `net_info.ping` entry is only registered on first use.

## Conclusions

More work need.

TerryE:
@vsky279 Lukàš, again, my apologies for coming late into this review.  Most of the comments above can be fixed quite easily.  The main issue IMO which requires some careful design though together with a re-engineerig analysis of  which resources are created, used and destroyed by which controller and when together with use cases.  For example, exactly what is the CB and when is it called, because the LWiP ping interface allows 3 types of callback: sent, recv and timeout.  Which are you exposing?  ...

vsky279:
@TerryE Terry, I'm happy to see you back. Thanks for reviewing the code.

> Lua API calls nearly always throw an error if an error is detected

Clear. This useless error check is removed.

> Use the `lauxlib.c` wrapper API when you can and don't reimplement inline

Wow. This is very elegant. :+1:

> Understand the broad memory allocation strategy

I hope I dealt with this with this already in reaction to Nathaniel's comments. Now I allocate the memory by
```c
ping_t nip = (ping_t)lua_newuserdata(L, sizeof(net_info_ping_t));
```
I hoped I hooked the userdata into the GC sweeplist by keeping a `nip->self_ref` retrieved via 
```c
nip->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```
Is the correct approach?
No freeing of resources occurs in `ping_received`. The clean-up is done in `ping_sent`. The `__gc` is also implemented.

> BTW, the ping_* routines malloc and free the ping_t buffer so this module should never free this.

It seems to me that ping_* routines malloc `ping_msg` buffer and not the `ping_opt`. So this implementation seems ok to me.

> Miscellaneous

It seems that `luaL_getmetatable` is void function. So the code does not compile:
```c
  if (!luaL_getmetatable(L, "net_info.ping"))
      luaL_rometatable(L, "net_info.ping", LROT_TABLEREF(net_info_dyn));
```

> LWiP ping interface allows 3 types of callback: sent, recv and timeout.

I am exposing the `recv` callback. The `sent` callback is used for freeing up resources.

vsky279:
I have implemented DNS too. Looking forward for your comments.

app/modules/net_info.c
vsky279: This is fixed in reaction to Terry's comments.

app/modules/net_info.c
vsky279: DNS is implemented in latest commit.

TerryE:
@vsky279 Lukáš,  I can't (or at haven't figured out how to) do github style inline review comments on a forced push hence here are my next batch.  But my general comment is that you have address only part of what I was suggesting so there is still more work needed. 

##  Argument processing

-  **Defaults**. `ping.c` treats 0 / NULL in the opt array as "use the default".  It's better not to double default so use 0 rather than 4 for the default `count`.

-  **Optional arguments**.  Your template is `net_info.ping(ip_address, [count], callback)`  BTW **_not_** as given in your [net_info.md](https://github.com/nodemcu/nodemcu-firmware/blob/127cebe73877a7c8fa7dfc84855827ea8370548b/docs/modules/net_info.md).  What happens if the user adds an extra parameter? e.g. `net_info.ping(ip, callback, 10)`  **Your code will bugout**.  That's why you must use `lua_settop()` if you want to use ToS relative addressing of the parameter list.  

So if you don't understand why I suggest doing it this way, then either you or I might have missed something.  Either way it is better to come back and ask why I suggest an approach rather than skip over the suggestion.:
```C
   lua_settop(L, lua_isfunction(L, 2) ? 2 : 3);
   lua_Integer count = luaL_optinteger(L, 2, 0); /* use ping_start() default */
   luaL_argcheck(L, lua_isfunction(L, -1), -1, "no callback specified");
   nip->ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```


## Re-egineering the ping_start interface

Still to do IMO, but here is my reading of this.  

The `ping_out` struct has these params: 
-  `count` No of pings to be done. 0=Default Max (4)
-  `ìp` IP4 address to be pinged. 
-  `course_time` time between sends in mSecs (0 = 1 sec).  See below for further discussion.  Your ping code uses the default 1 sec ping interval.
-  `recv_function` is called for each `send()` response.  A fixed one second timer is used to catch timeout in which case a zeroed response is returned. (Any actual response is ignored if it arrives after this timeout).
-  `sent_function` CB is fired after the Nth response is received or timesout.  _There is no codepath which does not call this CB._ 

LWiP has its own timer wrapper `sys_timeout()` which calls a CB after the specified elapsed mSec.  A linked list of outstanding timers is maintained so another call to the same timer cancels the previous instance.  `sys_untimeout()` clears down a pending timer.  `ping.c` uses two timers: one to trigger "receive timeout" after 1 sec, and one to schedule the next send after n mSec timeout.  (The timers internally use uSec so there is the usual × or / scalling by 1000 as a conversion.) 

`ping.c` allocates its own `ping_msg` struct to maintain its internal state and this is then freed immediately before calling the sent CB.  The caller must allocate (and free) its own `ping_option` structure.

All CBs are also passed a reference to a stack-allocated  `ping_resp` structure with the following fields set:

Field | On recv | On recv Timeout | On sent
----|----|----|----
`total_count` | 0 | 0 | Y
`resp_time` | Y | 0 | 0 
`seqno` | Y | 0 | 0
`timeout_count` | 0 | 0 | Y
`bytes` | Y | 0 | 0
`total_bytes` | 0 | 0 | Y
`total_time` | 0 | 0 | Y
`ping_err` | 0 | -1 | 0 

Note that this can be use to select on recv / sent as `total_count` > 0 if and only if on sent.

## Strong recommendations

1.  Use a Userdata for the `ping_option` structure (plus any addendum needed).
2.  This _must be reference (possibly indirectly) via the Lua registry to prevent its GC during the life of the ping sequence, so:
3.  Use a single C CB for both recv / sent and bind the Userdata to it as an upval. This makes the Userdata content available to the CB,
4.  Register the CB in the Lua registry, this will indirectly reference the Userdata during the life of the registered CB.
5.  The Userdata needs a self reference field for the CB entry.  It doesn't need a separate registry entry.
6.  The CB on the sent path should unregister itself from the registry.  This is enough to trigger complete GC.

`pipe.c` gives some coding examples of how to implement this.  If you are unsure or want me to give coding fragments then email me direct.

TerryE:
Incidentally the reason for my suggesting that you register the CB function and not the UData is that the way that the LGC works is quite straight forward:  it start sweeps from the Lua stack, the registry and the global environment, tracing down all references to mark any directly or indirectly referenced collectable objects.  Any `GCObject`s that are mark are then moved to the "to be GCed" list.

Thus is you create a UD on the stack and then don't  hook into one of the above then the stack will be trimmed back when to return from the function, so the UD is no longer referenced and will then be collected on the next GC sweep -- which can crash the application if there is still C callbacks still referencing it.

If you do something like `net_info.ping(ip, 10, func(n,ip,i,t) ... end)` then there is no reference to the UserData or the function over the next 10sec unless they are somehow lined into the registry.  You also want the code to somehow clean itself up, and the way to do this is as part of calling the sent CB.  Hence whilst the ping is running we have slot **N**, say in the registry -> CB function -> (upval 1) Userdata -> (callback ref) N. 
 
On the sent call the function can use the upval -> callback ref to get **N** and unref this. Once execution returns, the function and the Userdata are no longer hooked into a sweep chain and will be CGed.

vsky279:
@TerryE Terry, thanks again for your comments. This will be the most fine tuned module ever 😄 

>I can't (or at haven't figured out how to) do github style inline review comments on a forced push hence here are my next batch.

Is there a better way to submit commits in existing PR not to pile up more and more commits? I'm not strong in git either 😄.

>### Argument processing
>* Defaults. 

I agree. It makes no sense to define another default value. I missed that.

>* Optional arguments. 

Initially I took the code you proposed. But it does not work for `net_info.ping("www.google.com", function (...) print(...) end)`. It sets the top to 2. But `luaL_optinteger(L, 2, 0);` fails as parameter 2 is function and not an integer nor nil,

So to deal with `net_info.ping(ip, callback, 10)` (a case I did not think of) I suggest the following compromise:
```c
    bool isf2 = lua_isfunction(L, 2);
    nip->ping_opt.count  = isf2 ? 0: luaL_optinteger(L, 2, 0);  /* use ping_start() default */
    lua_settop(L, isf2 ? 2 : 3);
    luaL_argcheck(L, lua_isfunction(L, -1), -1, "no callback specified");
    nip->ping_callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);    
```

>### Strong recommendations

I think I'm starting to understand the issue around UD, CB and GC. Let me take some time to deal with it.

TerryE:
The main thing is to keep itt simple and clear.  You suggestion sounds good.

Sorry about seeming to set a high bar, but to be honest I am sort of using you and your module as a guinea pig / test case.  I want to raise the bar on our module coding and this means doing for new modules as well as a sweep of existing ones.  My dialogue with you will help me get the level right as your are already a good C dev, but more just new to the Lua C API foibles :smile:  As I said, if you want to IRC or ping stuff off me 1-1, then just drop me an email.  Terry

marcelstoer:
> I am sort of using you and your module as a guinea pig / test case. I want to raise the bar on our module coding

👍 😄 would be great of that experience eventually led to an updated (or "finally released" actually) version of this page: https://github.com/nodemcu/nodemcu-firmware/wiki/%5BDRAFT%5D-How-to-write-a-C-module

</OT>

vsky279:
I'm squashing all commits and will dig into depths of Lua C coding.

TerryE:
>  would be great of that experience eventually led to an updated (or "finally released" actually) version of this page: https://github.com/nodemcu/nodemcu-firmware/wiki/%5BDRAFT%5D-How-to-write-a-C-module

@marcelstoer  my thoughts entirely.

vsky279:
Ok, so now we have very clean implementation of the `net_info.ping()` function.
It was redesigned with great help of @TerryE Terry (in fact Terry re-wrote it 😄).
That was great experience that improved substantially my understanding of C Lua programming. Thank you.

TerryE:
> (in fact Terry re-wrote it smile).

That's too much attribution, IMO. We wouldn't be here without Lukáš doing the heavy lifting.  Doing this exercise has really help me clarify some of my own thoughts.  Thanks for your work.

TerryE:
@nwf  I think that this code in good to go now.  My only Q is whether this one function merits its own module or should be just move it into `net` possibly with a `NET_INFO_ENABLE` option in `user_config.h` just in we don't always want to compile in the ping code.  What is your view?

nwf:
I'd like it to be merged with `net`, available as `net.ping()`, possibly gated with `NET_PING_ENABLE`.

marcelstoer:
> My only Q is whether this one function merits its own module or should be just move it into net

That has come up here and in the original issue #1555 a couple of times. I personally favor the latter with no extra option. The current `net` module is quite a large beast (server sockets, client sockets, TCP, UDP, DNS) that adding another function won't make much difference. If at some point someone feels motivated to split it into more manageable chunks then yes, go for it.

nwf:
@marcelstoer Well, merging with `net` and gating behind a default-on `#define` would be fine for web builders and leave the option for people trying to reduce firmware size.  Still, I think you're right that just landing it in `net` is fine.

TerryE:
@vsky279 can you drop the net_info module, ditto the separate doc file and modules define. Add your code and doc to net and the enabled  `NET_INFO_ENABLE` option in `user_config.h`. 

I'd like to merge this before I do my next tranche of fixes. Thanks.

vsky279:
ping is a part of `net` module now. By default it is not included. `NET_INFO_ENABLE` needs to be defined in `user_config.h`.
I preserved ping implementation in separate `net_info.c` and `net_info.h` files. It seems cleaner to me then to have it directly in `net.c`. But let me know if you prefer latter option.

I was thinking about implementing a callback to the `sent` event as @TerryE suggested once. A second callback function as the fourth parameter of `net.ping` seems to me appropriate. What do you think?

nwf:
I dislike the use of the name `info` for `ping`; can it be `net_ping.c` and `NET_PING_ENABLE`?  Sorry for the long tail of cosmetic changes.

app/include/user_config.h
```diff
@@ -136,6 +136,9 @@
 //#define TIMER_SUSPEND_ENABLE
 //#define PMSLEEP_ENABLE
 
+// The net module optionally offers net info functionnality. Uncomment the following
+// to enable the functionnality.
+//#define NET_INFO_ENABLE
```
marcelstoer, line 141:
```suggestion
#define NET_INFO_ENABLE 
```

I was under the impression that we had consensus, albeit a vague one, between @nwf, @TerryE  and myself that it should be enabled by default? If not then I'll adjust the cloud builder to enable it.

TerryE: If we are dropping this down to `net.ping()`then it makes sense to make this define `NET_PING_ENABLE`.

TerryE: If we are dropping this down to `net.ping()`then it makes sense to make this define `NET_PING_ENABLE`.

TerryE: If we are dropping this down to `net.ping()`then it makes sense to make this define `NET_PING_ENABLE`.

TerryE: If we are dropping this down to `net.ping()`then it makes sense to make this define `NET_PING_ENABLE`.

vsky279:
All implemented and tested. Should be ready to be merged.

PR #2859, opened 2019-07-26
More ROTable optimisations 

### LINKED ISSUE #2858: Some more ROTable preparation / optimisations pre Lua 5.3
### ROTable performance changes

-  ROTable accesses use a lookaside cache (currently 32×4) to prime any ROTable lookup.  The (4) entries in the appropriate cache line are checked for a table/key hit and if so then this specific table entry in Flash is matched against the requested key.  In practice this achieves a better than 99% hit rate enabling direct reads of the ROTable entry in Flash without any ROTable scan.
-  In the case of a miss, a scan of the ROTable is required, and because of the high cache hit rate, nearly all such ROTable scans are for a missing key.  The highest source of key-misses (over 90%) are the Lua VM probes for various metafields such as `__metatable` or `__index`.  We have therefore added an ordering constraint on all ROTables: _fields with keys starting with the character "\_" **must** be located at the head of the table in ascending order._
-  A separate scan algo is used for fields starting with "_": this breaks if `strcmp(pentry->key, strkey)>=0` which means that typically only one or two fields are checked before the scan is aborted. 
-  A key-miss for all other key will result in a full scan of the table.  However we only support string keys for rotables, the `luaR_entry` `key` field now direclty points to the CString key name which are now word-aligned in Flash.  The scan now does a word-read from flash of the  `pentry->key` and `*pentry->key` and does a masked compare against the 1st word of the search key. (The mask is to accommodate keys shorter than 3 chars in length.)  This means that in practice the scan is a small inline loop with no unaligned exceptions or calls to the  ROM `strcmp()` (other than the last hit validation if there is a key match), and this executes fast.  

In practice this means that ROTable access is now perhaps 3× slower than equivalent RAM Table accesses, rather eLua case of perhaps 100× slower. 

### Preparing for `lua53`

In the Lua53 implementation, all tables will have a `Table` header, (though this is a reduced variant which omits unused fields in the case of ROTables).  The `LROT_` macros have been designed to hide these implementation differences at a source level in our C modules.

-  How a table is bound to its metatable is different in the two version.  This is a minor issue in our code, since with one exception we declare ROTables for use as UData metatables _not_ ROTable metatables.  
   -  In the case of Lua51, the eLua patch overloads the `__metatable` field.  The core Lua VM only uses the `__*` metafields in a metatable and ignores them in a normal table, but here any normal table is scanned for the  `__metatable` entry to determine if it uses a metatable.  This entry has been encapsulated in the `LROT_METATABENTRY( mt )` entry which expands to a `LROT_TABENTRY( __metatable, mt )`.
   -  In the case of Lua53,  The mt value is in the second parameter of the `LRO_END()` macro and this is used in the generation of the `ROTable`'s header.  Note that this field is ignored in lua51 builds and similarly the `LROT_METATABENTRY()` declarations are treated as empty expansons in lua53 builds.  This enable the same module source to be used for both targets.
-  The default Lua VM uses the table header `flags` as an existence check for the common meta fields such as `__index` to avoid unnecessary table accesses.  This optimisation is bypassed for the lua51 eLua patch.  However it is honoured in the lua53 implementation and the 3rd parameter of the `LRO_END()` macro contains this static flag setting.

### Suggestions for reviewers reviewing these meta fields

Try using the command:
```bash
grep -RIPn '\bLROT_(\w+?ENTRY\s*\(\s*_|(PUBLIC_)?BEGIN|END)' app | grep _ -C 2 | vi -
```

-  The `__*` fields must be in ascending order and consecutive immediately following the `LROT_BEGIN()` statement.
-  The second field of the `LRO_END()` macro such be `NULL` unless this is a ROTable that uses another ROTable as its metatable.
-  The third field of the `LRO_END()` macro should be 0 unless the table is used as a metatable and in this cae should match the corresponding metafield entries.

### Comment:

I realise that this change denomalises the source which is a bad thing, and if anyone can think of a better approach to us keeping this ROTable code base common to lua51 and lu53 then I am open to suggestions.

TerryE:
Fixes #2858

- [ ] ~~The code changes are reflected in the documentation at `docs/*`.~~  N/A

As discussed in #2858

This PR is not ready to commit but has been made available for review.

**PS**. The first 5 commits are an echo of `dev-new-lua.c` (#2836) and can be ignored for the purposes of any review.  The commit "ROTable dev optimisations" contains the changes to be reviewed.  I will rebaseline against dev after the next master drop and #2836 has been merged into dev and I will force push the version only containing the ROTable optimisations.


TerryE:
Apologies, I squashed a number of commits into 39f9d40 and one of these was separately labelled "sweep to remove trailing blanks that have crept into recent (mostly my) changes."  So you will see 20 files where the only changes are removing trailing blanks in this.

TerryE:
I have just added another commit.  Some of the modules use Lua API calls that aren't compatible with `lua53`, e.g. switching to the `anyfunction`variants; `lua_typeerror()` has been withdrawn; `lua53` requires you to use `luaL_ref()`  instead of the eLua macro `lua_unref()`, etc.  These module changes allow them to compile on both lua and lua53.

TerryE:
@jmattsson @nwf  I have updated my [Lua 5.3 Port](https://gist.github.com/TerryE/646407100785fb00c2cbddbb0e8a2b61) paper in the light of work to date.  Can you guys give this a careful read and get back to me with any issues / concerns that you might have because I will be using this as my bible to scope the further Lua53 migration work.

TerryE:
Just managed to compile `app/modules` under lua53 cleanly, albeit with a few botches which need to be moved to the right locations and excepting `node.c` which needs more work and #2865

TerryE:
I am closing this review copy and I will be reissuing an updated version as per #2890.

Issue #2893, opened 2019-08-21
TLS: don't KEEP_PEER_CERTIFICATE


nwf:
mbedTLS is holding on to a copy of the peer certificate for us, just in case we want it later.  But none of our APIs expose it to the Lua application.  We should flip `MBEDTLS_SSL_KEEP_PEER_CERTIFICATE` off.  This should have no functional consequence.

nwf:
Sigh.  This is a 2.17 knob and that isn't even a stable release yet.

nwf:
Doubly sigh: esconn_mbedtls manually sheds a great deal of TLS state.  This is effectively already implemented.

Issue #2890, opened 2019-08-15
Progress on testing Lua 5.3


TerryE:
**Apologies for my bad discipline but this issue has morphed into a status update on my testing and to avoid confusion, I have decided to rename it accordingly.**

------

Now that some developers are putting a _lot_ of functions in LFS we are starting to see a scaling issue which I feel that we need to address in Lua5.3.  See the relevant [Handling LFS-based functions](https://gist.github.com/TerryE/646407100785fb00c2cbddbb0e8a2b61#handling-lfs-based-functions) section in my migration working paper for more details.

The tl;dr is that I will be adding a ROTable index to LFS, in a slightly botched way as doing the Right Way (as Johny would describe it) involves too much work at this stage, IMO.  This will be transparent at a `LFS.module` level with a few tweaks to `_init`, but module resolution will be faster and scale a bit better.

If the committers are happy then I'll close this.

HHHartmann:
Maybe some metatable magic could also help to convert the `UData` to a `function` on the fly.
From Lua 5.2 on there is also __pairs and __ipairs.

TerryE:
> Maybe some metatable magic could also help to convert the `UData` to a `function` on the fly.

I think that we are saying the same thing.  :smile:   The  `__index`, `__call` and `__pairs` meta-methods  as described in the Lua RM are relevant, but the complication here is address resolution.  In a C module you statically declare a metatable or a function and the GCC compiler / linker will correctly resolve these references these in another ROTable, but this doesn't work between ROTables in LFS images and the firmware because the `luac.cross` compiler generated image isn't linked to a specific image, and so the meta wrapping needs to be resolved at runtime.  

We can do this in Lua using a table as we do currently with our `_init` LFS implementation -- which is what I was talking about.  We could also implement this in C using an empty Udata which adds a little more work but would be more runtime efficient.

The advantage of adding a lightweight Lua function as a proper function subtype is that none of this would be necessary as the TValue in the ROTable could directly reference this Proto,  so the table would just work.  But adding this subtype support involves a _lot_ more work than tweaking `_init` which is why I am suggesting going for this KISS approach for now.

HHHartmann:
Ok I think I got it. 
Couldn't you then create a new table in node.flashindex which accesses the ROTable in its metamethods?
That way it would be a snap in solution and not everyone had to update their version of _init.lua

TerryE:
[...]

Issue #2881, opened 2019-08-07
Stack integrity during EGC


TerryE:
### The issue

If a new resource is allocated, then the VM adds an extra step in the malloc process.  If the malloc fails returning NULL, then the  VM carries out a GC sweep to free up garbage memory and retries the malloc before aborting the allocation request with an out of memory error thrown.

### Expected behaviour

This process should maintain internal integrity and either succeed or correctly throw an out-of-memory error which can be handled and reported by the application.   It should also be economic as practical -- that is avoid throwing an error when GC could make sufficient available.

### Actual behaviour

The complication arise here if collectable objects have an associated `_gc` methamethod.  This means the the GC can end up calling Lua functions during the GC sweep.  Such calls can trigger a Lua stack resize.  This occurs very rarely and when it does the Lua VM internally handles all of the housekeeping to update the stack references in the `lua_State`, etc. preserving data integrity.  Issues can occur if the API code does something like:  
```C
  StkId o = index2addr(L, idx);          /* Cache a TValue pointer to the Lua stack */
  funcWhichAllocatesResource(L, ...);    /* this can result in a stack resize */
  /* any use of (o) now points to the wrong location and can crater the VM */
```
We have hit and tracked down one example of this in a live DiUS application.  The manifestation is that the VM will throw a H/W exception when this happens.  This is very occasionally and seemingly randomly, and so is very hard to reproduce systematically and therefore is hard to identify.

A specific example is the use of the `seterrorobj()` function in the `lua53/ldo.c`:
```C
  StkId oldtop = restorestack(L, ci->extra);
  seterrorobj(L, status, oldtop);
```
Here `seterrobj()` can push and error message as a TString to the stack when can trigger a GC and if it does then `oldtop` is still pointing to the old stack and not the new one.  If seterrobj used a `ptrdiff_t` as its third argument `seterrorobj(L, status, oldtop - L->stack)` then this offset reference is stack relative and won't cause an issue if the stack is resized and relocated.

### So what am I asking here?

Standard Lua doesn't experience this because it is working in a large RAM heap and recreating this issue is almost impossible, IMO.  This isn't the case when you've got 10s of Kb RAM available.  The eLua patch addressed on of the cases but missed the one which tripped up DiUS. So what our approach to this architectural issue.  I think that we have three broad responses:

1.  Ignore it and hope that it goes away.
[...]

PR #2912, opened 2019-09-12
Lua 5.1 to 5.3 Realignment Phase 2 and baseline 5.3

TerryE:
Addresses: #2808, #2839 (partial), #2858, #2890

To be addressed before merge: #2783

- [x] This PR is for th e `dev` branch rather than for `master`.

### Overview

This is a huge PR.  Sorry for this, but by nature, I can't practically release this in chunks.  As per https://github.com/nodemcu/nodemcu-firmware/issues/2890#issuecomment-529037946, I have reorganised this into a series of commits:

1. **Updates to app/lua to realign APIs**.  This is a reasonably large path to the Lua 5.1 codebase, implementing the Lua 5.3 version improved ROTables and the introduction of Function subtypes so that the modules can use the same API for building within both 5.1 and 5.3.
 
2.  **Updates to app/modules and supporting directories to realign APIs**.  A general cleanup of the modules to use the stricter Lua API implemented in (1)

3.  **Add the standard Lua 5.3.5 as a reference baseline**.  This is just the standard Lua 5.3.5 source without any changes reorganised into our current directory hierarchy.  This is purely for baselining purposes.  Note that the Lua Test Suite is included.

4.   **First pass of Lua53 changes to make runable Lua53**.  This includes the core changes needed to be able to build `LUA=53` and download a runnable Lua 5.3 firmware image onto the ESP, albeit with some limitations as discussed below.  This has been tested against the Lua test suite within the `luac.cross -e` environment.

In this form, the build works for the default `LUA=51` target, and as a minimum prerequisite for merge, we should have solidly tested this against both the firmware and `luac.cross` variants.

We should also have a stable and usable (albeit with agreed deficiencies) `LUA=53` target. 

I will create the initial PR with only (1) above and then add (2) - (4).  

### Still outstanding to be added before merge

- (#2783) Finalise and test ECG algo
- Int / Float variants
- (#2839) has been done upto handover to Johny's migration to IDF.
- (#2881) Stack integrity during EGC
- #2893 

**Special Request**. _Please review holistically at the first instance.  If necessary make changes and email me your patch.  Given the 1000s of individual line changes here, if we use the Github review API and commenting system, this PR will become rapidly unworkable._


TerryE:
Gosh this `hx711.c` is hassle :smile:  

Anyway what I've on on this last commit is to go through all of the Lua Test Suite in detail removing any temporary changes which I did to get lua53 to compile and run the tests, debugging issues as I went along.  The tests run clean with minimal and explained differences.  

There were some genuine bugs, and also some test assumptions that were invalid: e.g. the various base library functions and modules are in `ROM` and not directly in `_G`.  So looking good overall.  Lua 5.3 now includes almost complete `math` and `debug` libraries.  If you want a sort of lua interface on the host then create a file containing `debug.debug()` and `luac.cross -e 'debug.lua'` gives you a command prompt.

Still got to port LFS -- my next commit.

TerryE:
This latest commit includes the LFS functionality and various cascade fixes to allow the lua test suite to run will only known variations from standard to accommodate LFS running and in particular running the test suite as LFS images so that we hammer the core against execution from LFS.  This all runs within the `tests.c` wrapper which does paranoid dynamic memory tracking (e.g. checking overrun marker fields and failing if _any_ mallocs are not paired with frees.

This last feature both makes LFS on `luac.cross` possible since we can `lua_newstate()`, load LFS, `lua_close()`, `lua_newstate()` (with LFS loaded. Bouncing the Lua runtime like this saves a restart.

The next tranche of work is hammering the on-ESP firmware build.

TerryE:
OK, this latest push largely adds LFS support.  But in more detail
-  `luac.cross -o name -f <lua file list> generates an LFS image file.  The `-a` option isn't yet implemented.
-  LFS images can be loaded into `luac.cross` for execution in the `-e script` environment by using the `-F LFSimage` option.  They can also be loaded into the ESP from SPIFFS using `node.flashreload('imagename') which will reboot the ESP with the new LFS loaded.
-  The LFS file format is different from Lua 5.1 but is of comparable size. The compiled image (`.lc`) format is also different but about ½ the size.
-  Lua 5.3 is faster and uses less RAM than Lua 5.1.
-  I've added an extra `pbrk()` method to `gdbstub`.  Calling `gdbstub.init() is optional, and this method also enables UART output capture.  The `gdbstub` documentation has been updated. @pjsg Philip could you take a look at this please.
-  I've added the `%` operator to strings.  This works in a similar fashion  to Python.

Yet TODO:
-  Documentation
-  Review SPIFFS loading. 
-  Update LFS image loading in `nodemcu-parttion.py`
-  Lua test suite.  I'll raise a separate issue to discuss this one.  I've done a couple but this is quite a job.  Note that all of the failures so far have been issues with the scripts working in a NodeMCU environment, rather than the runtime itself failing. (e.g. references to `io` library or needing more than 44Kb RAM to run.) 

**You must include `LUA=53` in your make to build a Lua 5.3 runtime.**  Omitting this will get you a Lua 5.1 build.

I am getting to the point where I desperately need other testers to get involved, and in particular regression test the Lua 5.1 option.  There are quite a few changes here to align the Lua 5.1 and Lua 5.3 runtimes so that the modules work with both.  Once we are happy that at least Lua 5.1 is stable, then we can merge this into dev, since at the moment you need to do an explicit  `LUA=53` make to include the Lua 5.3 runtime.

fikin:
compiling this branch using marcel's docker nodemcu-build img is getting me to ld error :
```
HOSTCC /opt/nodemcu-firmware/app/lua/luac_cross/../lstrlib.c
../lstrlib.c: In function 'str_format2':
../lstrlib.c:828:5: warning: implicit declaration of function 'lua_rawlen' [-Wimplicit-function-declaration]
     int i,n=lua_rawlen(L,2);
     ^
HOSTCC /opt/nodemcu-firmware/app/lua/luac_cross/../ltable.c
...
HOSTLD ../../../luac.cross
.output/host/release/obj/lstrlib.o: In function `str_format2':
lstrlib.c:(.text+0x10b3): undefined reference to `lua_rawlen'

``` 
what am i missing here?

TerryE:
> what am i missing here?

An offline PM:
> Sorry guys
>
> If you want to do a LUA=51 (default) build then patch
>
>> sed -i -e '828s/lua_rawlen/lua_objlen/' app/lua/lstrlib.c
>
>This was one of the backported features to lua 5.1.  I copied across the code forgetting that lua_rawlen() in 5.3 is lua_objlen() in 5.1.  BTW this adds

If you are getting this message then you are testing the LUA=51 variant.  Try `make LUA=53 DEBUG=1` if you want to look at Lua 5.3

fikin:
ok, no worries. using docker run .. -e LUA=53 ... does the trick, compilation is ok ;)

fikin:
test code is:
```
file.remove("test.lua")
assert(file.open("test.lua","w+"), "file open failed")
file.writeline("print(1)")
file.close()
require("test")
```

first problem: running following as esplorer block-command results in reset:
```
> file.remove("test.lua")
> assert(file.open("test.lua","w+"), "file open failed")

 ets Jan  8 2013,rst cause:1, boot mode:(3,7)
load 0x40100000, len 28484, room 16 
tail 4
chksum 0xb7
load 0x3ffe8000, len 2468, room 4 
tail 0
chksum 0x1b
load 0x3ffe89a4, len 8, room 8 
tail 0
chksum 0xf6
csum 0xf6
```

second problem: running these line per line results in require error:
```
> file.remove("test.lua")
> assert(file.open("test.lua","w+"), "file open failed")
> file.writeline("print(1)")
> file.close()
> require("test")
error loading module 'test' from file 'test.lc':
	cannot open test.lc: 
stack traceback:
	[C]: in ?
	[C]: in function 'require'
	stdin:1: in main chunk
	[C]: in ?
	[C]: in ?
>
```

running as block-command all file operations is fine. reset comes only when require is being added.
btw i'm using d1.

so far haven't encountered other problems with this code.

TerryE:
@fikin  Nikolay,  thanks; that is brilliant.  I will track these down tomorrow.

HHHartmann:
When building with
```c
#define LUA_USE_MODULES_ENCODER
```
the build breaks
```
CC app/modules/crypto.c
crypto.c: In function 'call_encoder':
crypto.c:53:3: error: implicit declaration of function 'luaL_checktable' [-Werror=implicit-function-declaration]
   luaL_checktable(L, -1);
   ^
```

HHHartmann:
Hi again Terry,
this is strange behaviour:
```
NodeMCU 3.0.0.0 
	branch: pr_2912
	commit: 630c2903bf7561e3a9b8b539bff70b50bdf6bc98
	release: 2.0.0-master_20170202 +455
	release DTS: 
	SSL: false
	build type: float
	LFS: 0x0
 build 2019-11-05 21:02 powered by Lua 5.3.5 on SDK 3.0.1-dev(fce080e)
cannot open init.lua: 
> 
----------------------------
mispec.lc       : 3543 bytes
mispec.lua      : 4762 bytes
mispec_file.lua : 5283 bytes
mispec_ws2812.lua : 5741 bytes
mispec_ws2812_2.lua : 4781 bytes
mispec_ws2812_effects.lua : 808 bytes
----------------------------
Total file(s)   : 6
Total size      : 24918 bytes

Total : 3486139 bytes
Used  : 27359 bytes
Remain: 3458780 bytes

> require ("mispec")
stdin:1: module 'mispec' not found:
	no field package.preload['mispec']
	no file 'mispec.lc'
	no file 'mispec.lua'
stack traceback:
	[C]: in function 'require'
	stdin:1: in main chunk
	[C]: in ?
	[C]: in ?
> =node.heap()
41184
> require ("mispec")
error loading module 'mispec' from file 'mispec.lc':
	cannot open mispec.lc: 
stack traceback:
	[C]: in ?
	[C]: in function 'require'
	stdin:1: in main chunk
	[C]: in ?
	[C]: in ?
> 
```
The file `mispec.lua` and `mispec.lc` exist on SPIFFS.
Note the different error messages between the two require calls.
For the content of `mispec.lua` please see https://github.com/nodemcu/nodemcu-firmware/pull/2953
The lc file is compiled on chip.

TerryE:
@fikin @HHHartmann sorry for not updating on this.  Job for tomorrow.  I've been adding another feature requested by @jmattsson: support for the `luac.cross -a` option so now as well as
```bash
luac.cross -F myLFS.img -e init.lua
```
which will load a compiled LFS image into the `luac.cross` execution environment and make it available to the `init.lua` script (this is how I do a lot of my testing); you can now also do: 
```bash
luac.cross -F myLFS.img -o myLFS-a8000.bin -a 0x402a8000
```
will create an ESP binary LFS based at 0x402a8000 that you can use `esptool` to write directly to flash or to integrate it into a provision image.  I just need to add this to nodemcu-partition.

TerryE:
I've tracked down and fixed a couple of these. 

- The require bug, is because an error in vfs/stdio abstraction.  I missed a ! test.  The require handling has subtly changed 5.1 to 5.3, I need to do a bit more digging.  
-  `luaL_checktable()` is a macro in 5.1 `lauxlib.h` for `luaL_checktype(L, (n), LUA_TTABLE)` that was dropped in 5.3.  I've just substituted the check type expression in crypto.c. 

Will post another update tomorrow.

TerryE:
OK, yet more enhancements and fixes from feedback:
-  The big change is the addition of the `-a` option. I had to rework the LFS header to support this.  It no longer uses `size_t` pointers which can vary for target and host 64bit implementations, but now uses `lu_int32` offsets into the LFS region.  Hence `luac.cross` output formats have changed and any LFS images will need to be recompiled.
-  Implementation of #2950.  Simple and easy. Works well.
-  Fix LC loader issue in #2948 
-  Fix #2959.
-  Fix Crypto compile with Lua 5.1.

We really need some volunteers to hammer the standard make (with no `LUA=53` option) as this wi;ll be the default dev after merge.  I would like to consider merging this version in this next week or so.  Normal cloud builder users will continue to make the Lua 5.1 version, but advanced users who use Docker or Linux builds will be free to make the Lua 5.3 version.  Other than the issues that Gregor picked up with the `crypto`  library and which this latest commit fixes, all of the libraries that I've tested so far work with both the Lua 5.1 and 5.3 VMs.

HHHartmann:
You also need to replace the link in `nodemcu-firmware\sdk-overrides\include\espconn.h` with a file like in dev to not break windows checkouts
``` C
#ifndef _SDK_OVERRIDE_ESPCONN_H_
#define _SDK_OVERRIDE_ESPCONN_H_
// Pull in the correct lwIP header
#include "../../app/include/lwip/app/espconn.h"
#endif
```

HHHartmann:
@TerryE Not sure about this, but did you address the `lua_rawlen` issue in `lstrlib.c` when building LUA51?

TerryE:
> did you address the `lua_rawlen` issue in `lstrlib.c` when building LUA51

Yes, I am using the Lua 5.1 variant at this moment on this ESPlorer issue.

TerryE:
Changes in  32e161b 

### Interface consistency across the Lua 5.1 and 5.3 versions

-  As discussed previously all firmware components other than the Lua directories are shared, so these components must use a common API interface to the two different Lua runtimes.

-  The startup and interactive interface for Lua is primarily maintained in the `lua.c` file.  The standard distribution version has had quite a bit of restructuring moving from Lua 5.1 to 5.3.  The NodeMCU version is driven by the needs of LFS startup and that the runtime executes within a non-preemptive event driven task-oriented architecture, so we can't have an interactive poll `stdin` loop, so our two version were similar without good reasons for the differences, so I have decided to converge our NodeMCU `lua.c` variants completely, so that they will now use the _same_ source file (`app/lua/lua.c` will actually be a symlink to `app/lua53/lua.c`) with `#if LUA_VERSION_NUM` conditionals used where a few source changes are needed to reflect the internal runtime API introduced in 5.2 and 5.3.

-  I've had to do a little realignment to `lauxlib.c` etc, but with a common `lua.c` main module, keeping the Lua 5.1 and 5.3 versions in sync and ditto with the module libraries should be a lot more straightforward.

### UART fixes

The bugs breaking the uploaders have been fixed.

HHHartmann:
Thanks Terry for this commit, I hope it will fix several issues.
Let me remind you that symlinks will break checkouts under windows however. This is ugly but could we use short includes instead like 
```C
#ifndef _SDK_OVERRIDE_ESPCONN_H_
#define _SDK_OVERRIDE_ESPCONN_H_
// Pull in the correct lwIP header
#include "../../app/include/lwip/app/espconn.h"
#endif
```

I don't even know if the include guard is needed here. Maybe a oneliner is sufficient here.

jmattsson:
@HHHartmann You're right, for this use case we don't even need the include guard. I'd be "happy" with adopting the one-line-#include as a poor man's symlink.

HHHartmann:
@TerryE  Terry I am having problems loading an LFS image.
I am using docker, 
changed the build script to build LUS=53, 
flashed with complete clean,
built LFS image with new luac.cross
uploaded the image (yes it works now)
Executed flashreload
which gave me this
```
=node.flashreload("LFS_float_20191215-2052.img")
=node.flashreload("LFS_float_20191215-2052.img")
true
> 

load 0x40100000, len 28820, room 16 
tail 4
chksum 0x07
load 0x3ffe8000, len 2480, room 4 
tail 12
chksum 0xf5
ho 0 tail 12 room 4
load 0x3ffe89b0, len 8, room 12 
tail 8
chksum 0xf0
csum 0xf0
‡;ƒ›’nì’d„˜Œ <...>

Erasing LFS from flash addr 0x08e000 to 0x0cdfff
(error object is a boolean value)
stack traceback:
	[C]: in ?
	[C]: in ?
```
The device hangs here until reset
I configured LFS as
``` C
#define LUA_FLASH_STORE                   0x40000
```
and added several modules I commonly use

HHHartmann:
@TerryE  Terry I built from a clean clone only changing `#define LUA_FLASH_STORE                   0x40000` and using LUS=53.

 I get the following error building luac.cross

```
HOSTCC /opt/nodemcu-firmware/app/lua53/host/../lundump.c
../lundump.c: In function 'addTS':
../lundump.c:462:10: warning: unused variable 'totl' [-Wunused-variable]
   size_t totl     = sizelstring(l);
          ^
```
There is also a symbolic link in `sdk-overrides/include/espconn.h` which I resolved by copying the target.

Loading an LFS image built with this build yields the same results as mentioned above.

Going back to pre 53 firmware.

TerryE:
IIRC, total is used in a `lua_assert()` which generates the unused warning if debug isn't enabled. I can change the define for the non-debug path to generate `UNUSED(e)` which will suppress the warning.

The symbolic link issue is really an interoperability problem of docker on WinX using samba SMB which doesn't support symlinks across mount points.  You are the expert on docker over Windows. What is the best practice workaround for this flaw in docker?

I can't debug your LFS reload because I am not psychic; I don't have your test case. Email me a Zip of your LFS code and your config so I can reproduce it. Thanks.

HHHartmann:
For the symlink please see https://github.com/nodemcu/nodemcu-firmware/pull/2912#issuecomment-564284413 and @jmattsson s comment thereafter. 
There is also another symlink in app/lua/lua.c.
I will send an email this afternoon when I am back home.

TerryE:
Gregor, I understand _what_ the symlink issue is. My point is that the _why_ is nothing directly to do with with NodeMCU but rather this is a bug or functional limitation of how docker operates over Windows: if you want your source root to be a winX file tree in NTFS, then Docker uses Samba to expose this to the container. NTFS and Linux FS all support some form of symbolic linking. Samba doesn't, because the two implementations aren't fully consistent and it is difficult to map this functionality.

HHHartmann:
Terry as I recall this is an issue of git under Windows. Marcel pointed out somewhere that a plugin or an extension (not sure of the terminology) for git world have to be installed on the client machine. 

See https://github.com/nodemcu/nodemcu-firmware/pull/2856 for more info.

TerryE:
AFAIK, using git on Windows works with fixes as per the referenced issue; ditto git within a container so long as the root is within a container FS.  The symlink problem arises when using git within the container with the root referencing a Samba mapped mount point.  This is annoying, but as you point out we can use a `#include` as a workaround to avoid duplicating files unnecessarily.

fikin:
rebase to latest encounters one compile problem, namely gregor's findings about unused "size_t totl     = sizelstring(l);".
commenting it out makes compilation ok.
i haven't got any softlink problems as mentioned above (using linux docker -e LUA=53). 

i'll begin experimenting with it now.

@marcelstoer , @TerryE how do you see merging it to dev branch? is there something to be helped with? for 51 vs 53 building?
imho it would be totally awesome to get it mainstream ;)

TerryE:
> @TerryE Terry I am having problems loading an LFS image.
> I am using docker,
> changed the build script to build LUS=53,
> ...
> The device hangs here until reset

@Gregor sorry for the illness related sabbatical.  Back in the loop now.  The load seems to be working, but the issue is do with an optimisation to remove a reboot.  If you externally reset the ESP then so should find that the LFS has actually loaded.  I will do a fix for this.

HHHartmann:
@TerryE Terry,  so glad to hear that you are back in the saddle and please don't excuse yourself for being ill.
So I will give it another try this weekend. I suspect that my current project will not run with lua 5.3. So I will see what I can do about that. 
Btw I am HHHartmann. "gregor" belongs to Gregor Herdmann. Guess he is wondering why he was mentioned ;-)

TerryE:
@gregor, I tried loading `./luac.cross -f -o /tmp/lfs.img  $(find lua_examples -name \*.lua)` which loaded fine, other than (1) the above sticking reboot bug, and (2) I had to exclude one file which has lua 5.1 only syntax (see #3028\)

This second triggers EM: errors when doing an `LFS._config`.

TerryE:
@HHHartmann, Sorry about the delay.  I had a mind fart during merging the 51 and 53 variants of `lua.c`.  The RTS uses slightly different methods of signalling LFS reload to lua.c needs to have variant code.  The 5.3 restart was hanging but I've fixed this so here is an example where I load pretty much all of the Lua files in `lua_examples`:
```
>  node.flashreload'lfs.img'
> 
 ...
Erasing LFS from flash addr 0x096000................................. to 0x0d5fff
LFS image loaded

NodeMCU 3.0.0.0 
	branch: dev-lua53-2
	commit: 32e161b131400a0cd215fddd50733d26dd98ff93
	release: 2.0.0-master_20170202 +457
	release DTS: 201912102127
	SSL: false
	build type: float
	LFS: 0x0
	modules: adc,bit,file,gdbstub,gpio,net,node,tmr,uart,wifi
 build 2020-02-09 17:29 powered by Lua 5.3.5 on SDK 3.0.1-dev(fce080e)
cannot open init.lua: 
> =table.concat(LFS._list,' ')
dummy_strings HTTP_OTA _init lfs_fragments graphics_test sjson-streaming send_text_message play_file play_network irsend GT_cross GT_graphics_test GT_clip GT_text GT_gradient GT_box GT_pixel_and_lines GT_fonts GraphicsTest GT_color_test UcgLogo GT_triangle HelloWorld telnet_pipe telnet_fifosock simple_telnet telnet default luaOTAserver _provision _doTick check adc_rgb send_email_smtp myfile tz mqtt_file mqtt2cloud somfy mcp23008_leds mcp23008_buttons webap_toggle_pin tcp2uart make_phone_call
> =node.heap()
42536
> 
```
I just want to check out LFS reload for 5.1 and rebaseline against current `dev` so we can do the merge into `dev`.

marcelstoer:
Superseded by #3075 -> closing

PR #2938, opened 2019-10-05
SSL rampage

nwf:

Update us to mbedtls 2.17 and then angrily rampage around Espressif's goo.  It turns out that they basically have a second lwip just for SSL.  This is completely bogus and should go away.  Some of it can go away right now because it's unused; the rest is used and will require, among other things, making our primary lwIP build with NO_SYS==0, which requires some understanding of mutexes and such.

Anyway, this is not really a PR per se so much as a request for review and guidance.


marcelstoer:
@TerryE would it be a problem for you if we merged this ahead of your large #2912? The changes are confined to mbedTLS.

nwf:
If it's good to go, let me make sure that we're sync'd up with mbedtls 2.17 before landing this one.

pjsg:
Does this mean that we can now configure the incoming/outgoing ssl buffers independently? This will significantly improve interoperability and memory consumption!!

marcelstoer:
>...2.17 seems still-born? Or we could stick with 2.16.3...

I wasn't able to work out their release strategy. However, as per https://tls.mbed.org/tech-updates it seems 2.16 is their newest LTS branch. That maybe explains why 2.17, 2.18, etc. aren't announced and the only exist as tags (but not releases) on GitHub. The latest in this line currently is 2.19.1.

TerryE:
@marcelstoer,  we're going to have to do an intelligent rebaseline /  merge anyway, and we're not getting the testing effort that we really need so let's unblock some of these critical issues for now.

nwf:
@marcelstoer In that case, I guess the "go big or go home" is 2.19.1.  I'll see if jumping up breaks anything for us when I have a minute.

@pjsg Oh, that would be kind of nice, but I think we should keep the 4K buffer sizes as the default.

@TerryE I apologize for having not had time to do any real work on NodeMCU in months.  Day job significantly ramped up its time consumption.

pjsg:
I would leave them as 4k each for now, but we could change the incoming one to 6k and the outgoing one to 2k and significantly improve interoperability without consuming extra memory.

dtran123:
@pjsg , did you mean MBEDTLS_SSL_MAX_IN_CONTENT_LEN and MBEDTLS_SSL_MAX_OUT_CONTENT_LEN. I am not sure if this is supported on the latest mbedtls ?

pjsg:
I checked some time ago (maybe a year or more ago), and it was a feature in the current version of mbedtls, but not in the version that was in our codebase. This PR is adding support for the IN/OUT_CONTENT_LEN and we can use that fact to improve our interoperability without increasing the total RAM demands.

dtran123:
Super. That opens open up new possibilities for sure to control/tune the in and out buffers!

nwf:
This is a very slightly less aggressive rampage now, in that it lands the latest commit on the 2.16.3 series, which is the latest one they seem to believe in supporting.  However, importantly, this series now includes a test program (`lua_tests/tls-test.expect`) that performs some basic sanity checking of the TLS stack.  I intend to expand this functionality, and I think this kind of test infrastructure is what we want for #2145.

nwf:
I have moved the test infrastructure out to its own PR, #2983, as I do not think it should be blocked by TLS changes.

marcelstoer:
@TerryE do you mind if we land this on `dev` now or would it cause too much (merge) trouble?

nwf:
Given TerryE's thumbs up, I'm going to go ahead and click merge.  Sorry if this breaks anything for anyone.

PR #2942, opened 2019-10-11
lua module ds18b20: fixed missing forward declaration

glumanda:

- [- ] This PR is for the `dev` branch rather than for `master`.
- [x ] The code changes are reflected in the documentation at `docs/*`.

bugfixing

### DIFF
--- a/lua_modules/ds18b20/ds18b20.lua
+++ b/lua_modules/ds18b20/ds18b20.lua
@@ -50,6 +50,8 @@ local function to_string(addr, esc)
   end
 end
 
+local conversion
+
 local function readout(self)
   local next = false
   local sens = self.sens
@@ -114,7 +116,7 @@ local function readout(self)
   end
 end
 
-local function conversion(self)
+conversion = function (self)
   local sens = self.sens
   local powered_only = true
   for _, s in ipairs(sens) do powered_only = powered_only and s:byte(9) ~= 1 end


TerryE:
Hi @glumanda Andreas, 
```Lua
local function conversion(self)
```
is just syntactic sugar for 
```Lua
local  conversion;  conversion = function(self)
```
(This is _not_ the same as `local  conversion = function(self)` as the former makes the local `conversion` available and the latter is evaluated right to left and will use whatever was in scope before the `local  conversion` declaration.)

Since the preceding `readout()` makes to reference to `conversion`, moving the local declaration above this has _zero_ effect.  So sorry, there is no point in making this null change.  Also on  a procedural note changes go into `dev` first.

HHHartmann:
@TerryE Terry, I don't understand your comment.

> Since the preceding readout() makes **to** reference to conversion, moving the local declaration above this has zero effect. ...

Did you mean **no** reference?

In this case your assumption is incorrect as `readout()` does reference `conversion` in line 108.

@glumanda The PR would still have to be prepared against dev.

TerryE:
Gregor, you are right: I did look through the source but missed this reference :scream: so my _zero effect_ conclusion was wrong.

vsky279:
Hi, I don't get what is the implication of this missing forward declaration.
I thought the module worked fine. Or not?

TerryE:
Lukáš, do a `luac.cross -l -l` to get a Lua VM code listing for the functions in the module.  The parser has to compile the function `function() return conversion(self) end` within `readout()`.  If `conversion` isn't local which it isn't here then it

-  creates an upval reference (in both this anon function and in `readout()` _if_ there is a `local conversion` in one or more of its parents, or
-  it generates a global reference.

Referencing only resolves backwards so this changes the generated code and the runtime behaviour.

vsky279:
Got it. Thanks Terry.

Issue #2921, opened 2019-09-21
Bounds check for ws2812.buffer:replace does not work


HHHartmann:
I found this error while writing tests for this module.
So the effort was worth it right away.

### Expected behavior
buffer:replace with invalid replace position throw an error.

### Actual behavior
Does not abort and possibly corrupts memory.

### Test code
```Lua
strip_buffer = ws2812.newBuffer(5, 3)
strip_buffer:replace(string.char(3,255,165,33,0,244,12,87,255), -7)
strip_buffer:dump()
```
the last commad reboots the mcu after some seconds

### NodeMCU version
NodeMCU 3.0.0.0 
	branch: dev
	commit: 0b90495517c8e2b34dbb3a2a99455ac3c4f05f1c

### Hardware
standard dev board

HHHartmann:
Need to fix 
``` C
static ptrdiff_t posrelat (ptrdiff_t pos, size_t len)
```
and adapt all occurrences if needed

PR #2953, opened 2019-11-02
Fixes for `ws2812` and `ws2812_effects`

### LINKED ISSUE #2946: shifting more pixels than available for ws2812.buffer:shift reboots
I found this error while writing tests for this module.
So the effort was worth it right away.
This Bug was introduced a while ago, so if we had had tests then it would not be there.

Expected behavior
buffer:shift with more pixels than selected(available) should throw an error.

Actual behavior
Reboots.

Test code
```Lua
buffer = ws2812.newBuffer(4, 4)
buffer:shift(-10)
```
the last commad reboots the mcu

NodeMCU version
NodeMCU 3.0.0.0
branch: dev
commit: 0b90495

Hardware
standard dev board

HHHartmann:
Fixes #2946 #2921 and more.

Cleanup code and fix issues in `ws2812` and `ws2812_effects`
Besides the afore mentioned ones I fixed a false bounds check in `ws2812_effects.set_brightness` and `ws2812_effects.set_speed`

Also add some automated tests to show the differences between current and this implementation.
You will need to precompile at least the mispec.lua file to not run out of memory.
If you don't like them I will remove them from the PR.

### DIFF
--- a/app/modules/ws2812_effects.c
+++ b/app/modules/ws2812_effects.c
@@ -58,6 +58,7 @@ typedef struct {
   uint8_t effect_type;
   uint8_t color[4];
   int effect_int_param1;
+  ws2812_buffer_shift_prepare* prepare;
 } ws2812_effects;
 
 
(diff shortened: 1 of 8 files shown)


app/modules/ws2812_effects.c
```diff
@@ -1005,7 +1001,7 @@ static int ws2812_effects_start(lua_State* L) {
     state->counter_mode_call = 0;
     state->counter_mode_step = 0;
     // set the timer
-    os_timer_setfn(&(state->os_t), ws2812_effects_loop, NULL);
+    os_timer_setfn(&(state->os_t), ws2812_effects_loop, (void*)L);
```
HHHartmann: Some more testing revealed that passing the lua_State this way does not work.
Trying to find a clean solution.
Any hints welcome!

marcelstoer:
@TerryE would it be a problem for you if we merged this ahead of your large #2912?

Issue #2956, opened 2019-11-03
httpd server module in c


NiklausS:
8<------------------------ FEATURE REQUEST ------------------------------------
### I am using this httpd server now for quite some time, and thought it might be a good idea to let others also use it if they like so.

it provides quite robust upload / download / execute capabilities.

### Justification
I've read it in the posts every now and then that this is something others would also use.

### Workarounds
i know there are some httpd servers written in lua. I don't want to judge on which one is better, as this one is quite simple and works for me ...

I have provided a git diff to the master branch, including documentation and some samples. I'd be happy if some of you can have a look at it, and let me know what i have to change to have this in the master.

Thank you
 Nik
[httpd.patch.gz](https://github.com/nodemcu/nodemcu-firmware/files/3802099/httpd.patch.gz)

8<------------------------ END FEATURE REQUEST --------------------------------

TerryE:
@NiklausS It would be far better to comply with out contribution guideline by forking the repo and issuing a PR rather than a bare patch.  Also I see that you are making a lot of changes to the net module.  We would be reluctant to do this without compelling reasons and decent backwards compatibility.

marcelstoer:
> i know there are some httpd servers written in lua. I don't want to judge on which one is better

We normally state that anything that doesn't have C-level timing requirements (e.g. hardware protocol) should be done in Lua. Therefore, we also started to deprecate and remove C modules for which there is a suitable Lua alternative.

We have a [Lua HTTP server module](https://nodemcu.readthedocs.io/en/master/lua-modules/httpserver/) in this repo and https://github.com/marcoskirsch/nodemcu-httpserver is also quite popular.

seregaxvm:
> We have a Lua HTTP server module in this repo and https://github.com/marcoskirsch/nodemcu-httpserver is also quite popular.

It seems abandoned. It would be nice to have a community fork.

marcoskirsch:
> It seems abandoned. It would be nice to have a community fork.

I don't work on this anymore but would be happy to transfer ownership to somebody interested and willing to keep working on it.

PR #2957, opened 2019-11-05
ESP8266 add webserver module

NiklausS:

#2956

This is a simple webserver. The reason i have written it initially was to allow for a web based develop environment. It is written as a c module as i run to often into memory shortages (this was before LFS). Furthermore especially the upload feature required some small additions in the net.c to achieve a decent performance and stability.

The main changes are the additions of the functions from httpd.c, the changes in net.c are basically the move from the local typedefs into a net.h, the addition of the hooks to the net_recv_cb, net_err_cb, net_sent_cb, net_listen and net_close. By intercepting earlier to received data, it can be written faster and expecially without furher memcopy to flash.When receiving a 500k file there is usually never more then 6k mem used.


marcelstoer:
For the reasons stated in #2956 I vote to close this.

HHHartmann:
+1 from me.

NiklausS:
Well, that's sad.
It's was so super convenient and fast to upload and test lfs.images while i had some trouble lately with ESPlorer.
i really got used to this:

```
curl -H Expect: --data-binary @luac.out.1574191913 http://192.168.1.204/up\?filen=lfs.img
echo "node.flashreload ('lfs.img')" | curl -m 1 -d @ - 'http://192.168.1.204/cmd'
```

and other automated testcases.

In this case, it probably makes no sense if I also provide the patch for the dev-esp32.

pjsg:
For easy upload, I run a tftpd server on the esp8266 and then when it gets the complete lfs.img, it automatically flashes it and restarts.

NiklausS:
i know the tftpd, but when you want to upload from a browser out of javascript this does not work. Furthermore is the tftpd not for interactive commands in (REST).

marcelstoer:
Do the Lua HTTP servers we mentioned not support POSTing/uploading binary data?

PR #2967, opened 2019-11-20
improve mqtt documentation

stromnet:
- [x] This PR is for the dev branch rather than for master.
- [x] This PR is compliant with the other contributing guidelines as well (if not, please describe why).
- [x] have thoroughly tested my contribution.
- [x] The code changes are reflected in the documentation at docs/*.

The existing documentation invalidly states that the offline callback is same as the connection-failure callback. I had reconnect code only in my 'offline' callback, which made the system fail to reconnect when mqtt connection was not immediately possible after it was disconnected (i.e. timeout did not trigger offline callback)

### DIFF
--- a/docs/modules/mqtt.md
+++ b/docs/modules/mqtt.md
@@ -157,8 +157,8 @@ end
 
 In reality, the connected function should do something useful!
 
-The two callbacks to `:connect()` alias with the "connect" and "offline"
-callbacks available through `:on()`.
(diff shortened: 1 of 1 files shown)


marcelstoer:
If the documentation is inconsistent with the code you can always ask yourself which of the two describes the _intended_ behavior. I'm merging this so they're at least consistent.

stromnet:
Yes, I was contemplating that.. But the current behaviour makes sense I think, if we where never connected, then we cannot go offline (and thus should not call the "offline" callback).

PR #2969, opened 2019-11-23
Sensor driver for BME680 sensor - update to version 3.5.9

### LINKED ISSUE #2944: Temperature influence to the BME680 air pressure and humidity readout
### Expected behavior:
The air pressure, humidity and gas resistance are compensated during ambient temperature changes.
### Actual behavior:
The air pressure drops or grown when temperature increase or decrease. Please look at attached screen. I have checked two sensor. Both behaved the same. Then I have checked it in Arduino with ClosedCube library problem still existed, but when checked with Adafruit library the pressure was stable, and changed as weather forecast shows. I suposse that NodeMCU and ClosedCube library comes from the same source with bug.

### Test code
ClosedCube lib: https://github.com/closedcube/ClosedCube_BME680_Arduino
Adafruit lib: https://github.com/adafruit/Adafruit_BME680 - works OK
### Hardware
ESP12

vsky279:
Fixes #2944.

Solves the issue where the air pressure, humidity and gas resistance are not compensated during ambient temperature changes.

Updates the Bosh BME680 driver from version 3.5.1 to current version 3.5.9.

### DIFF
--- a/app/modules/bme680.c
+++ b/app/modules/bme680.c
@@ -68,7 +68,10 @@ static uint8_t r8u(uint8_t reg) {
 	return ret[0];
 }
 
-/* This part of code is coming from the original bme680.c driver by Bosch.
+// replace 'dev->calib.' with 'bme680_data.'
+// replace 'dev->amb_temp' with 'amb_temp'
+
+/**\mainpage
  * Copyright (C) 2017 - 2018 Bosch Sensortec GmbH
  *
  * Redistribution and use in source and binary forms, with or without
@@ -108,19 +111,12 @@ static uint8_t r8u(uint8_t reg) {
  * other rights of third parties which may result from its use.
  * No license is granted by implication or otherwise under any patent or
  * patent rights of the copyright holder.
+ *
+ * File		bme680.c
+ * @date	19 Jun 2018
+ * @version	3.5.9
+ *
  */
-
-/**static variables */
-/**Look up table for the possible gas range values */
-uint32_t lookupTable1[16] = { UINT32_C(2147483647), UINT32_C(2147483647), UINT32_C(2147483647), UINT32_C(2147483647),
-        UINT32_C(2147483647), UINT32_C(2126008810), UINT32_C(2147483647), UINT32_C(2130303777), UINT32_C(2147483647),
-        UINT32_C(2147483647), UINT32_C(2143188679), UINT32_C(2136746228), UINT32_C(2147483647), UINT32_C(2126008810),
-        UINT32_C(2147483647), UINT32_C(2147483647) };
-/**Look up table for the possible gas range values */
-uint32_t lookupTable2[16] = { UINT32_C(4096000000), UINT32_C(2048000000), UINT32_C(1024000000), UINT32_C(512000000),
-        UINT32_C(255744255), UINT32_C(127110228), UINT32_C(64000000), UINT32_C(32258064), UINT32_C(16016016), UINT32_C(
-                8000000), UINT32_C(4000000), UINT32_C(2000000), UINT32_C(1000000), UINT32_C(500000), UINT32_C(250000),
-        UINT32_C(125000) };
 
 static uint8_t calc_heater_res(uint16_t temp)
 {
@@ -132,9 +128,7 @@ static uint8_t calc_heater_res(uint16_t temp)
 	int32_t var5;
 	int32_t heatr_res_x100;
 
-	if (temp < 200) /* Cap temperature */
-		temp = 200;
-	else if (temp > 400)
+	if (temp > 400) /* Cap temperature */
 		temp = 400;
 
 	var1 = (((int32_t) amb_temp * bme680_data.par_gh3) / 1000) * 256;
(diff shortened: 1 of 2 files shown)


app/modules/bme680.c
```diff
@@ -244,21 +248,34 @@ static uint32_t calc_humidity(uint16_t hum_adc)
...
 	uint64_t var2;
 	int64_t var3;
 	uint32_t calc_gas_res;
-
-	var1 = (int64_t) ((1340 + (5 * (int64_t) bme680_data.range_sw_err)) * ((int64_t) lookupTable1[gas_range])) / 65536;
-	var2 = (((int64_t) ((int64_t) gas_res_adc * 32768) - (int64_t) (16777216)) + var1);
-	var3 = (((int64_t) lookupTable2[gas_range] * (int64_t) var1) / 512);
-	calc_gas_res = (uint32_t) ((var3 + ((int64_t) var2 / 2)) / (int64_t) var2);
+	/**Look up table 1 for the possible gas range values */
+	uint32_t lookupTable1[16] = { UINT32_C(2147483647), UINT32_C(2147483647), UINT32_C(2147483647), UINT32_C(2147483647),
```
pjsg: probably want to make these arrays static... reduces initialization cost.

docs/modules/bme680.md
```diff
@@ -143,15 +143,13 @@ bme680.setup()
 bme680.startreadout(150, function ()
     T, P, H, G, QNH = bme680.read(alt)
     if T then
-        local Tsgn = (T < 0 and -1 or 1); T = Tsgn*T
-        print(string.format("T=%s%d.%02d", Tsgn<0 and "-" or "", T/100, T%100))
-        print(string.format("QFE=%d.%03d", P/100, P%100))
-        print(string.format("QNH=%d.%03d", QNH/100, QNH%100))
-        print(string.format("humidity=%d.%03d%%", H/1000, H%1000))
+        print(string.format("T=%.2f", T/100))
```
pjsg: I would prefer to leave this the way that it was as it works for both integer and float builds.

app/modules/bme680.c
vsky279: This is coming from original Bosh code. I don't know why they made it dynamic. Changed back to static.

docs/modules/bme680.md
vsky279: I though integer version is more like a history. But ok, I revert the change.

marcelstoer:
@Michal78S it would be great to get your confirmation that this fix solves the issue. Can you build a firmware from @vsky279's branch testing?

vsky279:
I agree that it is good idea to know whether it solves @Michal78S's issue.
But even if it doesn't we should merge as it upgrades the driver and solves the same issue in my case.

Michal78S:
It is an honor for me, but since I have started nodemcu i build firmware witch cloud builder...

vsky279:
@Michal78S What modules do you need? I will compile the fw for you. Floating version?

Michal78S:
The standard modules: GPIO, net, node, timer, uart, wifi and of course i2c, BME680.  
The open issue is IAQ. Is it possible to add it to BME680 module? There is on the market one solution which uses STM32 and Bme680 and provides both: IAQ index and gas resistance.

vsky279:
Please find the compiled module here: [https://github.com/vsky279/bme680_bin/blob/master/nodemcu_float_bme680_20191203-0946.bin](https://github.com/vsky279/bme680_bin/blob/master/nodemcu_float_bme680_20191203-0946.bin).

The issue with IAQ calculation is that the algorithm has not been published by Bosh. Please see module documentation:
> This module is able to measure the gas resistance (see Bosch's datasheet). The gas resistance is not the IAQ (Indoor Air Quality) Index. But apparently it can be used as some proxy. The value still should somehow reflect the air quality. It seems that the higher value the air quality is better.
>
> The algorithm for IAQ calculation from the gas restistances (probably measured at different temperatures) is not publicly available. Bosch says that at this point of time the calculations for the Indoor Air Quality index are offered only as a pre-compiled library (see discussion here: BoschSensortec/BME680_driver#6). It is available as the BSEC Library. The algorithm is implemented in the library bsec/algo/bin/ESP8266/libalgobsec.a. Unfortunately I did not even manage to run the Bosch BSEC example on ESP8266 using this library.

If you find somewhere module implementing IAQ calculation then we can replicate it here.

shimosaurus:
I made the Lua code for reading data from module GY-MCU680V1 with sensor BME680 and STM32. It reads IAQ and gas resistance values: https://github.com/shimosaurus/mcu680

marcelstoer:
@shimosaurus Woah, you even used our documentation template in your README 😄 Great job.

vsky279:
> I made the Lua code for reading data from module GY-MCU680V1 with sensor BME680 and STM32. It reads IAQ and gas resistance values: https://github.com/shimosaurus/mcu680

Have you tested it? I can't imagine this can be the way to get the IAQ value. Bosch's documentations says the IAQ calculation is made available through the BSEC library which is not open-sourced.

The code you provide is a bit weak as the commands sent to the sensor are not documented.

Michal78S:
> 
> 
> I made the Lua code for reading data from module GY-MCU680V1 with sensor BME680 and STM32. It reads IAQ and gas resistance values: https://github.com/shimosaurus/mcu680

Nice job :) but it works only with module GY-MCU680 where are STM32 and BME680. Main job and calculations provides STM32. This solution has groving number of ussers because is one of few available so adding Lua code for it is good idea.

shimosaurus:
> @shimosaurus Woah, you even used our documentation template in your README smile Great job. Is that really all the code it takes? Were your inspired by this gist https://gist.github.com/xxlukas42/a878f5c44f1ec108a460e9b9500b79f5 ?

I was inspired here: https://forum.mysensors.org/post/94736 but @xxlukas42 used probably the same source.

> Have you tested it? I can't imagine this can be the way to get the IAQ value. Bosch's documentations says the IAQ calculation is made available through the BSEC library which is not open-sourced.
The code you provide is a bit weak as the commands sent to the sensor are not documented.

As @Michal78S wrote - my code doesn't count IAQ, but reads "only" the STM32 output on the GY-MCU680 board where BSEC is running. But as long as the BSEC library is closed-source, this solution is a way to obtain IAQ values ​​into the NodeMCU Lua.

My comment was not directed mainly to this pull request but to query @Michal78S how to get IAQ values.

vsky279:
> My comment was not directed mainly to this pull request but to query @Michal78S how to get IAQ values.

Oh, I see that puzzled me.

Michal78S:
I can confirm, new BME680 driver works better. I got the same readings of pressure from BME680 and BME280. During weekend I'm going to make outdoor tests (wider temp. range).

Other importatnt thing is @shimosaurus solution. Many NodeMCU users don't konow about it, so how about add it to the cloud builder?

marcelstoer:
> ...@shimosaurus solution. Many NodeMCU users don't konow about it

Jan could make a PR to have the script hosted at https://github.com/nodemcu/nodemcu-firmware/tree/master/lua_modules - that's up to him. It would then also show up under the Lua modules at https://nodemcu.readthedocs.io/en/latest/.

> so how about add it to the cloud builder?

Contrary to C modules Lua modules are not built into the firmware binary. You need to provision them separately via SPIFFS or LFS.

Michal78S:
Checked. New driver gives stable and reliable readings air pressure from BME680. Yesterday in the afternoon pressure rapidly dropped. It was registered by BME680, and values are near identical with the forecast and readings from BME280 and amateur weather station.
pogoda

PR #2974, opened 2019-11-30
add LiquidCrystal lua module

seregaxvm:
This module provides functionality similar to Arduino [LiquidCrystal](https://www.arduino.cc/en/Reference/LiquidCrystal) library - control LCD text screens based on the Hitachi HD44780 chip. It supports 4bit and 8bit GPIO; 4bit I2C interfaces (via GPIO expander). I2C version was tested in `i2c.SLOW` and `i2c.FAST` speed modes.

Some functionality of the LCD chip was not implemented: reading text from internal memory and adding custom characters. These functions are exotic and are not needed for most applications.


nwf:
I have done limited testing with a 4x20 4-bit/I2C LCD module I have laying around and it seems workable.  I've been using a forked version of https://github.com/dvv/nodemcu-thingies/blob/master/lcd1602.lua in my projects, but having something in the NodeMCU tree itself seems like a grand plan.  The things I think would be welcome improvements are...

- Some notion of a four line display (even if just a table of the offsets for column 0).
- Implementing `LCD_SETCGRAMADDR` functionality (custom characters)
- The ability to specify which bit of the I/O port is actually `LCD_RS`, `LCD_RW`, `LCD_EN`, and `LCD_BACKLIGHT` in the `bus_args` table.

But I don't think any of those are blockers for this going into the tree.

nwf:
Oh, actually, one thing that might be important before merging: documenting exactly what the `busyloop` parameter influences, so that callers can choose to instead provide their own timing delays rather than using `tmr.delay`.

seregaxvm:
- Added `customChar` function that defines custom characters.
- Changed `write` so that it would allow printing characters by char code.
- Added optional `row` argument to `cursorMove`.

> The ability to specify which bit of the I/O port is actually `LCD_RS`, `LCD_RW`, `LCD_EN`, and `LCD_BACKLIGHT` in the `bus_args` table.

@nwf  Do you mean that someone might want to use I2C GPIO expander other then [this one](https://ae01.alicdn.com/kf/H3f3a1ec8320a4e32b850619c2e0a5e4d0.jpg)?

nwf:
Some people apparently use the same PCF8574P I2C GPIO expander with different wiring for the LCD itself; the suggested schematic in http://lcdproc.sourceforge.net/docs/lcdproc-0-5-5-user.html#hd44780-i2c, for example, uses different pinouts.  The `lcdproc` driver thus exports, and I have had to use, configuration knobs for which pin is which; see https://github.com/lcdproc/lcdproc/blob/5ea8501240d00f3e9744f6a78c088a3446eea75f/server/drivers/hd44780-i2c.c#L150, for example.  If it's straightforward, it'd be great if `LiquidCrystal` were similarly capable.

seregaxvm:
Added the following definition to `i2c4bit.lua`:
```lua
local LCD_RS = 0
local LCD_RW = 1
local LCD_EN = 2
local LCD_BACKLIGHT = 3
local LCD_D4 = 4
local LCD_D5 = 5
local LCD_D6 = 6
local LCD_D7 = 7
```

This way it is easy enough to tune for different pinouts and `init` function would not be too messy.

nwf:
I was just about to commit this, because it's been languishing too long, but then I realized... `img/HD44780_char_codes.png` is probably not redistributable; it looks like an excerpt from a manual, and so probably copyrighted.  @seregaxvm, if you still care about this patch series after all this time (sorry), can you please amend it to not include that file?  Failing that, I'll get around to it myself at some point, because I really would like to have a LCD module in NodeMCU.

seregaxvm:
> can you please amend it to not include that file

@nwf Done

marcelstoer:
Thanks for paying attention to copyright! Instead of redistributing an image one could either link to one (e.g. https://www.google.ch/search?q=HD44780+char+codes&tbm=isch&source=univ) or link to a specific page in the PDF like so `HD44780.pdf#page=35`

seregaxvm:
Had some static analyzer warnings there...
Added link to specific datasheet page.

lua_modules/liquidcrystal/i2c4bit.lua
```diff
@@ -0,0 +1,55 @@
...
+   send4bitI2C(value, bit.lshift(1, LCD_RS), self._bus_args, self._backlight)
+end
+
+I2C4bit.backlight = function(self, on)
+   self._backlight = on
+   self._backend.command(self, 0x0)
+end
+
+I2C4bit.init = function(self)
+   i2c.setup(self._bus_args.id, self._bus_args.sda, self._bus_args.scl, self._bus_args.speed)
```
nwf: Since there may be other peripherals on the bus, I think this should perhaps be conditional on `self._bus_args.sda` being not-`nil`?

lua_modules/liquidcrystal/i2c4bit.lua
```diff
@@ -0,0 +1,55 @@
+local I2C4bit = {}
+
+-- I2C GPIO expander bitmap
```
nwf: On second thought, could these be shoved into `_bus_args` instead of being fixed like this?

lua_modules/liquidcrystal/i2c4bit.lua
seregaxvm: [`self._bus_args.sda`](https://nodemcu.readthedocs.io/en/master/modules/i2c/#i2csetup) is a pin number, so it's pretty safe to assume that it's not nil. This module barely fits into RAM, additional error checks would make it LFS-only (not sure, maybe it should be)

lua_modules/liquidcrystal/i2c4bit.lua
seregaxvm: The reason behind this it that most of the [LCD I2C modules](https://aliexpress.com/wholesale?SearchText=i2c+lcd+arduino) have this pinout. Of course, it could be shoved into `_bus_args`, but wouldn't it make initialization function excessively large?

lua_modules/liquidcrystal/i2c4bit.lua
nwf: I was meaning just making this `if self._bus_args.sda then i2c.setup(...) end` so that there's a mechanism to forgo bus initialization here.

seregaxvm: Added this feature

lua_modules/liquidcrystal/liquidcrystal.lua
```diff
@@ -0,0 +1,213 @@
...
+LiquidCrystal._displayfunction = 0
+LiquidCrystal._displaycontrol = 0
+LiquidCrystal._displaymode = 0
+LiquidCrystal._backlight = false
+
+LiquidCrystal.init = function(self, bus, bus_args, fourbitmode, onelinemode, eightdotsmode, column_width, insert_delay)
+   self._bus_args = bus_args
+   if bus == "I2C" then
+      assert(fourbitmode, "8bit mode not supported")
+      self._backend = dofile "i2c4bit.lua"
```
nwf: Creating backends like this is going to result in a new object per LCD, even if they all use the same backend.  That's fine, and could even be what you want, except that the backends here are coded to be immutable collections of functions.  So I'd suggest either:

* switch to `require`-ing the backends, so that they only get loaded once (and are subsequently cached)
* redesign the backends to be factories that produce functions that close over their configuration parameters and state, rather than look them up in dictionaries.  These factories can return (or directly set, if that's more your thing) functions that end up being `self._command`, `self._backlight`, `self._write` and friends.

Either approach should help with memory usage.  I think doing both will have the maximum impact, as then the function bodies will be shared and only the closures will vary per LCD.

lua_modules/liquidcrystal/i2c4bit.lua
```diff
@@ -0,0 +1,59 @@
...
+end
+
+I2C4bit.backlight = function(self, on)
+   self._backlight = on
+   self._backend.command(self, 0x0)
+end
+
+I2C4bit.init = function(self)
+   if self._bus_args.pinout == nil then
+      self._bus_args.pinout = {
```
nwf: Thanks, but I think making this a second table is bloating memory, which you indicated was tight.  Maybe something like
```
  self._bus_args.rs = self._bus_args.rs or 0
  self._bus_args.rw = self._bus_args.rw or 1
  -- ...
```

lua_modules/liquidcrystal/liquidcrystal.lua
seregaxvm: Current code works with both SPIFFS and LFS. `require` will not find module if it is stored on LFS

nwf: We really should just make `package.loaders` search LFS by default at this point, but thus far the guidance has been for programmers to enable that themselves: https://nodemcu.readthedocs.io/en/master/lfs/#accessing-lfs-functions-and-loading-lfs-modules

Anyway, `require` can be taught to search LFS and it's the recommended configuration.

Failing that, you could push the onus of finding and constructing the backend to the user.  So rather than `:init(bus, bus_args, ...)`, it'd be more like `:init((require "backend")(args), ...)`.

lua_modules/liquidcrystal/liquidcrystal.lua
```diff
@@ -0,0 +1,213 @@
...
+
+   self._backend.init(self)
+   self._backend.command(self, bit.bor(LCD_FUNCTIONSET, self._displayfunction))
+   self._backend.command(self, bit.bor(LCD_ENTRYMODESET, self._displaymode))
+
+   self:display(true)
+   self:clear()
+end
+
+LiquidCrystal.clear = function(self)
```
nwf: A belated style note.  Rather than writing `foo.bar = function(self, args) ... end`, you can write `function foo:bar (args) ... end`; the name `self` is implicit from the `:`; see https://www.lua.org/pil/16.html .

seregaxvm: Changed function style in `liquidcrystal.lua`. Also, renamed `self` to `screen` in other files.

seregaxvm:
> I've made some time to draft up what I had in mind in hopes of tamping down on memory requirements by using constructors and closures rather than tables. I've also found out that it is possible to read back from these LCDs and so gone and written a first pass at an automated test harness for LiquidCrystal.
> 
> Give https://github.com/nwf/nodemcu-firmware/tree/lcd-draft a look over and take anything that's useful. Please note that it the draft there is _incomplete_.

@nwf I'm not very familiar with Lua internals. Does it really require less memory to have such constructors with local functions? Don't all functions internally have metatables associated with them? Or `require` somehow caches all that?

nwf:
I will not pretend to be an expert at Lua internals, by any means.  But, as I understand it, the function bodies themselves are constant and are shared values across all constructor invocations.  The bits that aren't shared are the "closure" entries that point at the code blocks and the captured "upval" variables.  These are not generic (meta)tables, but slightly specialized storage structures for each function.  Section 3 of https://pdfs.semanticscholar.org/73a2/e3c03f799956aa5a3188e4eb35c90977a471.pdf seems to do an OK job describing what's going on.

Even with the juggling I did, instantiating a LCD still seems to occupy 3-4KB of RAM, which seems... high.  I mean to investigate further as time permits.

seregaxvm:
I've got these results for (keep in mind that closure version has extra feature read from lcd implemented):

| `node.heap()` | table | closure |
|:-:|:-:|:-:|
| initial | 43056 | 43056 |
| instantiated | 28232 | 26496 |
| garbage collected | 28232 | 27016 |
| Δ | 14824 | 16040 |

Maybe this ought to be a C module?

nwf:
I'd rather not push this down to C, since that brings a lot of maintenance burden.  I've been re-running your experiment with modules in LFS and I found a missing `local` in my refactoring.  Fixing that (it's now pushed to the same place), and re-running, I see (Δs from initial conditions in parens)...

|                         | table        | constructor  |
|-------------------------|--------------|--------------|
| init loaded             | 37320        | 37456        |
| LC loaded               | 34480 (2840) | 35848 (1608) |
| LC, backend loaded      | n/a          | 35760 (1696) |
| backend instantiated    | n/a          | 34344 (3112) |
| LC+backend instantiated | 33312 (4008) | 34312 (3144) |
| GC'd                    | 33392 (3928) | 35600 (1856) |

Both were run after `node.reboot()` and waiting for the system to settle, with the `lfs` example `_init.lua` present on the module as `init.lua`; I'm not quite sure whence the slight difference in starting conditions.  But even allowing that the `constructor` numbers might be off by 136 bytes, I think the savings are significant, but we/I should dig into where those 3KB are going in more detail.

Some preliminary comments:
* The extra read-back feature appears to occupy mostly code space, and that lives entirely in LFS.
* I am surprised by the `table` GC'd number being so high, but have not attempted to analyse it.
* The backend load is as light as I expected it would be; the LC load is *heavy* and I'm not sure why.
* The backend instantiation is heavy, but some of that is because it has built a table we expect to be ephemeral and released after full instantiation.
* Nevertheless, the full instantiation overhead (3144 - 1696 = 1448 bytes) is also heavier than I was expecting.

To ease repetition, the code for the `constructor` tests was
```
print("INIT", node.heap())

metalcd = require "liquidcrystal"
print("LC LOAD", node.heap())

metabe = require "lc-i2c4bit"
print("BE LOAD", node.heap())

backend = metabe({address = 0x27, id  = 0, speed = i2c.SLOW, sda = 4, scl = 5})
print("BE INST", node.heap())

lcd = metalcd(backend, false, true, 20)
metalcd=nil
metabe=nil
backend=nil
print("FULL INST", node.heap())

lcd=nil
print("GC", node.heap())
```
and for the `table` tests was
```
print("INIT", node.heap())

metalcd = require "liquidcrystal"
print("LC LOAD", node.heap())

lcd = metalcd:init("I2C", {address = 0x27, id  = 0, speed = i2c.SLOW, sda = 4, scl = 5}, true, false, true, 20, false)
metalcd = nil
print("FULL INIT", node.heap())

lcd = nil
print("GC", node.heap())
```

seregaxvm:
Replaced tables with closures for I2C. Code is close to <https://github.com/nwf/nodemcu-firmware/tree/lcd-draft> but I don't think that we should replace all constants (`LCD_CLEARDISPLAY` etc) with their values in the code. This makes code too obscure; replacing some logical functions using this constants with the result hides the code logic.  I think we can secrefice some memory and calculation cycles for readability.

Because now user has access to `busy` flag, we can remove all `tmr` functions. If user wishes, he can use loop `while lc:busy() do end`.

Not sure if we should defer initialization of the screen. Maybe `backend.init` should be removed and initialization should be done upon backend instantiating?

Had to hack around status reading.
When reading like this

EN | RS | RW | I2C
:--:|:--:|:--:|:--:
1|1|0|write
1|1|0|read
0|1|0|write

had to 

EN | RS | RW | I2C
:--:|:--:|:--:|:--:
1|1|0|write
1|1|0|read
1|1|1|write
0|1|1|write

This increments address counter which has be corrected by saving/restoring cursor position.

seregaxvm:
TODO:

`liquidcrystal.lua`:
- [x] rewrite code
- [ ] update docs
- [ ] write tests
`i2c4bit.lua`:
- [x] rewrite code
- [ ] update docs
- [ ] write tests
`gpio4bit.lua`:
- [ ] rewrite code
- [ ] update docs
- [ ] write tests
`gpio8bit.lua`:
- [ ] rewrite code
- [ ] update docs
- [ ] write tests

nwf:
If there's desire to keep the names around, rather than inline the constants (a position I fully understand for code clarity's sake, to be clear!), then it becomes important to pay attention to space overheads.

Consider:

```Lua
local FOO = 0
local BAR = 1

function locals() return FOO + BAR end
```

creates a closure with two upvals, so two pointers in its closure:

```
function <tmp/foo.lua:4,6> (5 instructions at 0x55ba8c3d8e90)
0 params, 2 slots, 2 upvalues, 0 locals, 0 constants, 0 functions
	1	[5]	GETUPVAL 	0 0	; FOO
	2	[5]	GETUPVAL 	1 1	; BAR
	3	[5]	ADD      	0 0 1
	4	[5]	RETURN   	0 2
	5	[6]	RETURN   	0 1
```

while

```
local t = {
  FOO = 0,
  BAR = 1
}
function tabular() return t.FOO + t.BAR end
```

creates one with one upvalue and two constants.  This is important, because constants live in the function's Proto (and therefore in LFS), not RAM.

```
function <tmp/foo.lua:10,10> (7 instructions, 28 bytes at 0x564374f72c90)
0 params, 2 slots, 1 upvalue, 0 locals, 2 constants, 0 functions
	1	[10]	GETUPVAL 	0 0	; -
	2	[10]	GETTABLE 	0 0 -1	; "FOO"
	3	[10]	GETUPVAL 	1 0	; -
	4	[10]	GETTABLE 	1 1 -2	; "BAR"
	5	[10]	ADD      	0 0 1
	6	[10]	RETURN   	0 2
	7	[10]	RETURN   	0 1
```

Unfortunately, tables themselves defined in Lua live in RAM and cannot be stored in LFS; they will be instantiated upon module load.  Depending on the number of constants referenced by each function, this could still be a net win: everyone's carrying around one pointer to one table in memory rather than many upval slots to refer to their constants, but it would need to be measured.  (And, having measured it, it would be awesome if there were something written in the documentation to guide subsequent module authors!)

With *extreme* hesitation, I wonder if it's worth pushing a very little bit of this down into C, because there we do have ROM tables available to us.  @TerryE doubtless has much more insight into this than I do.

nwf:
ETA: A truly horrifying, but nevertheless possible, way to square this particular circle would be to use **C-preprocessed Lua**, where the constants were given as `#define` statements... ick ick ick, but it'd maybe solve both problems (space and clarity) at once, at the (high!) cost of requiring the user to run `cpp` to generate their Lua.

seregaxvm:
@nwf Is Lua compiler smart enough to compile out constant values? Maybe it's just a matte of percompiling Lua source.

seregaxvm:
If we are to use cpp preprocessing, it has to be integrated into the build system. I doubt that windows users will be happy about having to install GNU toolchain in order to upload a script.

marcelstoer:
> requiring the user to run `cpp` to generate their Lua.

👎 as that would severely limit the usefulness of this module. An estimated 70% of our users are on Windows.

nwf:
We could ship both the `.lua.cpp` and the `.lua` files in the tree?  That's somewhat unpleasant, but maybe worth it for the memory savings.

seregaxvm:
Maybe it's better to write custom preprocessor in Lua? Constants may be defined with some commented keyword like
```lua
local LCD_CLEARDISPLAY = 0x80 -- !CONST!
```
This way script may be run as is, but it also may be preprocessed.

nwf:
That's a good idea, and nicely decouples the issue from this PR.  Go ahead and leave it as you have, with `local` names for constant values.

seregaxvm:
Actually I suspect that it's not too hard to detect constants automatically without any keywords

TerryE:
The whole idea of considering a c preprocessor for embedded Lua is definitely the wong way to go  IMO, and just not needed.  Accessing Upvals is cheap and the storage overhead is minimal.  In general it takes less RAM to use scalar upvals than table ones.  Also remember that binding upvals is only done once when the closure is itself bound.  In an LFS usecase, don't worry too much about the true constants and Lua code since this is in flash anyway.  If in doubt then `luac -l -l` is your friend, and the Lua VM Internals paper that I reference in my FAQ.  :smile:

nwf:
@TerryE: the particular concern is that there isn't a way to define a symbolic name for a constant in Lua.  That's the motivation for wanting either to rewrite the Lua before handing it off to luac, especially in combination with LFS or to push constants into a C rotable.  Neither option is very attractive, and I would, ordinarily, agree that neither should be considered, were it not for....

In my experiment on this PR, I found that inlining all the `local LCD_FOO = 0x04` definitions saved almost *a kilobyte* of heap, between the storage allocated for the TValues and the upvals in the closures created for the `LiquidCrystal` metatable methods.  While those TValue structures and the function closures are indeed only created once, an entire kilobyte seemed quite heavy.

That concern then spilled out more generally: should we generally be trying to minimize the number of upvals referenced by closures by referring to tables instead?  I *think* the answer is no, because unless there is truly an exorbitant number of closures, the overhead of creating the table outweighs the overhead of the upvals, but I don't know for sure.

TerryE:
> @TerryE: the particular concern is that there isn't a way to define a symbolic name for a constant in Lua...

Lua has some strengths for embedded use but unlike JavaScript and PHP, there isn't the concept of symbolic constants in the language.

> In my experiment on this PR, I found that inlining all the local LCD_FOO = 0x04 definitions saved almost a kilobyte of heap, between the storage allocated for the TValues and the upvals in the closures created ...

I accept your figure of 1Kb, which is regrettable, but then again this is only a few % of the total heap at boot.  This shouldn't be an issue with the code in LFS. The whole point of LFS is that we can move the Lua code  of RAM, so we can be a little more relaxed about creating variables.

> That concern then spilled out more generally: should we generally be trying to minimize the number of upvals referenced by closures by referring to tables instead? I think the answer is no, because unless there is truly an exorbitant number of closures, the overhead of creating the table outweighs the overhead of the upvals, but I don't know for sure.

I think that the main goals in any code should be correctness, clarity and simplicity.

seregaxvm:
This code may be used to test gpio8bit on esp32 (since sep8266 does not have enough gpio pins):
```lua
local gpio = gpio

return {
   read = gpio.read,
   write = gpio.write,
   mode = function(pin, mode)
      gpio.config{gpio={pin}, dir=mode}
   end,
   INPUT = gpio.IN,
   OUTPUT = gpio.OUT,
   HIGH = 1,
   LOW = 0,
}
```

seregaxvm:
@nwf Code and docs are updated, only tests are left. Maybe wait until #2983 before adding them?

nwf:
Thank you for all your hard work on this!  I believe this PR is good to go live, but I'll give it a day or so for others to chime in.  Yes, I think the tests will have to wait until we've decided how we want to do things; for `LiquidCrystal`, `mispec` (cf. #2984) may be able to completely cover the intended functionality (esp. since there's only one ESP and no network involved), and so we wouldn't need host-driven tests (#2983).

nwf:
Ah, one more favor to ask: can you rename your backends from (e.g.) `i2c4bit.lua` to `lc-i2c4bit.lua`?

nwf:
Er, no, one more still, I'm afraid.  While `readChar` can read the display RAM, there doesn't appear to be a way to read graphics RAM?  I think that just means another parameter to the backend's `read` function?

seregaxvm:
@nwf added missing functionality, renamed backends, fixed some 10 dot mode bugs. When this will be ready for merge I will squash the commits.

nwf:
LGTM!

marcelstoer:
> When this will be ready for merge I will squash the commits.

Thanks, but there's no need. GitHub will do that for us when we merge 😄

seregaxvm:
I've made a small program for custom character generation. I'll upload it to PyPi and update documentation soon.

image

nwf:
@seregaxvm I'm ready to push the button if you are. :)

seregaxvm:
Merge when you are ready. I don't have anything to add

PR #2984, opened 2019-12-24
Create NodeMCU test system based on gambiarra

HHHartmann:
helps fixing #2145

- [x] The changes are reflected in the documentation.

this has evolved, so I changed this description too:
Add test framework which can run tests like the following on an ESP
``` Lua
N.testasync('SEMI alarm', function(next)
  local t = tmr.create();
  local count = 0
  t:alarm(200, tmr.ALARM_SEMI, 
    function(tp)
      count = count + 1
      if count <= 1 then
        tp:start()
        return
      end
      ok(eq(count, 2), "only 2 invocations")
      next()
    end)
  ok(true, "sync end")
end)
```

Also includes first tests for the file and tmr module, mainly to have a working example.


HHHartmann:
NOTE: The readme is not yet updated.

nwf:
It'd be really nice to be able to announce the existence of our test framework by the next release.  @HHHartmann have you had time to weigh the pros and cons of mispec and gambiarra?  Would you mind doing a bit of compare-and-contrast for those of us not familiar?

nwf:
@HHHartmann Would you be opposed to just using `tests/` instead of `lua_tests/`?  I find myself wanting to write a bunch of HW tests in Lua using the as-yet-still-being-defined "NodeMCU Test Environment", which I think most naturally belongs next to the sea of TCL, which seems more like a `tests/` and less like a `lua_tests/` kind of thing.

nwf:
Having played a little with gambiarra now, I think I like it, except that we don't know how many tests we're going to run, which makes things a little sad from TAP's perspective.  We could require the tests themselves to generate the TAP plan message of "TAP 1..34" (e.g.) (which might be a perfectly OK thing to do, and is in fact kind of typical of TAP-based frameworks; it's just that because myspec build up the whole test table up front, it was easy to count).

HHHartmann:
I am also in favor of gambiarra. I already added some more functionality to it and there might be more.
It also came with tests to test itself, so refactoring and adding is safer.
The syntax is also clearer than in mispec.

I am also working on first async tests which I will commit soon.

HHHartmann:
I don't mind renaming the directory. Just thought that it matches the remaining directories containing lua code.
But as a non native speaker I don't get all of the sentence. What is TCL? I often struggle with acronyms you seem to like.

nwf:
Sorry to be dense in communication.  FWIW, your English seems better than mine, and I'm a monolingual dunce. ;)

TCL is the Tool Command Language and is used by `expect` (https://core.tcl-lang.org/expect/index), a program for automated control of other programs via text (usually their stdin/stdout).  See the `.expect` files in https://github.com/nwf/nodemcu-firmware/tree/dev-active/tests and the library `.tcl` files in https://github.com/nwf/nodemcu-firmware/tree/dev-active/tests/expectnmcu .

I think (but I'm open to suggestions) that it makes sense to put all our core tests in a single directory, regardless of their implementation language, and I, lacking creativity, chose `tests/`.  There's no real harm in a `tests/` and `lua_tests/` split, it just feels a little odd to me.

Case in point: I want to write (in Lua, and probably gambiarra at that, rather than `expect`) a test of the `adc` module.  To do this, I was going to depend on hardware as defined in https://github.com/nwf/nodemcu-firmware/blob/dev-active/tests/NodeMCU_Test_Environment.rst and so despite the implementation language, `tests/test-adc.lua` seemed more of the right place than `lua_tests/test-adc.lua` or such.

HHHartmann:
Wow your test plans are taking shape. As I see it the plan with the teo ESPs is that they are only needed for extended perifery testing. I would like to also use them for tcp testing. One being the testee and one being the known-to-work counterpart which is also steered by the test running on the first ESP.

But back to the naming issue: If  this folder is to be populated by other test files it does not make sense to name it "lua" something of course.
May be we should have subfolders for files which are for execution on (one of) the ESP8266 and the ones for the host.

nwf:
`test-net-intermodule.expect` is something of a placeholder right now; yes, it really should be doing packet exchange between the two.

HHHartmann:
I think it would be better to have expect only orchestrate the test runs and execute the tests themselves as a Lua script on the ESP. That way everyone wanting to run a test could just do that without having to set up expect at all.
Also not having to learn another language to do testing would sure help developers to accept that they have to write tests.
We should really keep the hurdle as low as possible here.

nwf:
I really, truly appreciate your point.  I think the vast majority of testing can happen on device, which is why I wrote https://github.com/nwf/nodemcu-firmware/blob/dev-active/tests/tap-driver.expect and https://github.com/nwf/nodemcu-firmware/blob/dev-active/tests/shim_gambiarra.lua and https://github.com/nwf/nodemcu-firmware/commit/28fd42e99d363af3410164bd60671d4047f9de26 .  Things like `adc` and `i2c` and `gpio` and `encoder` and `crypto` and `pixbuf` and so on can merrily live 100% in Lua and run on device and just be orchestrated, as you say, by programs on the host.  (The use of TAP, in fact, means that it should be easy to not use `expect` in some eventuality.  It's designed, on purpose, to be language-agnostic.)  I just haven't written many examples yet, but that absolutely is the goal.  I guess I can point at https://github.com/nwf/nodemcu-firmware/blob/dev-active/lua_modules/liquidcrystal/test-i2c4bit.lua and the converted https://github.com/nwf/nodemcu-firmware/blob/dev-active/lua_tests/mispec_pixbuf_1.lua if you want proof I'm not BSing when I say that I think most testing can be on the device.

But certain things (esp., the trickier modules, which are all I've written tests for, because the easy ones are easy and I know how they will work, once I know what test harness we'd like to use) like `net` and `tls` and `mqtt` and `uart` and so on need orchestration not just on one NodeMCU device, but on either two NodeMCU devices or on a NodeMCU device and programs running on a host computer.  In some cases there truly isn't any getting around that (`tls` cannot connect to another NodeMCU device and `mqtt` needs a server) while in some cases the increased power of orchestration on the host means that our tests have shorter dependencies (`uart` and `net-intermodule` can test cross-device without needing *two* *independent* cross-device signaling mechanisms).  There isn't, so far as I know, a suitable in-Lua replacement for the power of `expect`, which is why I've written those tests in TCL.

nwf:
Well, perhaps rather than all those words, an example is in order.  Here, https://github.com/nwf/nodemcu-firmware/blob/dev-active/tests/test-adc.lua, is a test of the `adc` module.  I think it conveys the flavor of how I expect on-device tests to look.

I'm worried that `gambiarra` doesn't use `node.task.post()`, so the whole program runs without any return to the SDK.  ISTR this was not true of `mispec` but perhaps is also what you were alluding to with your coroutine comment on #2145?

HHHartmann:
The test looks very nice. Especially the generation of testcases. Maybe we could add data driven tests here, Just giving an array of params and then integrate the loop into gambiarra.

But for the `node.task.post()` issue. Your assumptions are correct. The file test takes several seconds without rebooting the device. Probably due to no tcp activity during that time.

But it should be no problem instrumenting gambiarra with that. It supports async tests allready.
It the also could print the number of tests up front.

About the coroutines:
Currently tests look like this:
``` Lua
test('timer real async', function(next)
  local t = tmr.create();
  t:alarm(500, tmr.ALARM_SINGLE, function() 
             ok(true, "async end")
             next()
        end)
  ok(true, "sync ende")
end, true)
```

would change to something like this:
``` Lua
test('timer real async', function(CallbackFactory, await)
  local t = tmr.create();
  t:alarm(500, tmr.ALARM_SINGLE, CallbackFactory.timeout)
  ok(true, "sync ende")
  cbName, param1 = await()
  ok(eq("timeout", cbName))
  -- param1 is the timer which has been passed to the callback
  ok(true, "async end") next() 
end, true)
```
So especially for simpler tests this seems easier to code to me.
The test just ends at the end of the coroutine.

for fans of the next methode to be called to indicate the end of the test this can be used:

``` Lua
test('timer real async', function(CallbackFactory, await)
  next = CallbackFactory.next
  local t = tmr.create();
  t:alarm(500, tmr.ALARM_SINGLE, function() 
             ok(true, "async end")
             next()
        end)
  ok(true, "sync ende")
  await()
end, true)
```

The while test runs in a coroutine and the await switches contect and waits for any callback issued by the factory and the returns wizth the first param as the callback name and the others the parameters passed to the cb.

nwf:
I think I misunderstand the gambiarra interface; is the intent to call `test` once and use `ok` and `ko` and friends internally, or is it expected that the caller will make repeated calls to `test`?

Your examples seem like they have a good bit of boilerplate that's required for the async/callback machinery to operate; could we somehow more directly encapsulate that inside gambiarra?  Say, by making `ok` and friends yield through the user-provided function back to `test`?

HHHartmann:
The idea is to make repeated calls to test.

I think it would be a good idea to start each test with a post but it would break async testing when a callback could happen on every `ok` and `nok`. Just imagine executing an ok and a callback altering the data being checked. That would be a race condition causing the test to fail randomly.
It would however make sense to allow a timeout to the await function. So if the test requires it an `await(0)` could be issued.
I think that running all tests as async (inside a coroutine) does not make sense as it requires extra memory.

Also a failing `ok` or `nok` should fail the current test and not continue as it is now.

nwf:
I think past-me was overly optimistic about on-device testing.  Present-me went to bang out a `gpio` test in a spare minute and realized that I do not think I will be able to test `gpio.trig()` on device: doing so would require being able to have a test run, the test harness suspend its execution and wait for the trigger or a timeout, and then the test harness resume.  This strikes me as likely far easier from off-device, and so I'd reach for `expect`, but I'm open to suggestions for how to test callbacks on-device.

nwf:
I'm concerned that the async tests that call `next` leave the test function on the call stack... that could, maybe, be `return next()` instead, making it a tail call, but even then, I'd want to be convinced that we aren't growing the call stack with a series of async tests.

nwf:
Hm.  Mixing async and sync tests seems more subtle than I'd want.  I am trying to test `gpio.trig` with

```
test('gpio toggle trigger 1', function(next)
  seti2cb7(false)
  tmr.delay(100)
  gpio.trig(8, "both", function(l,w,c)
    print("TRIG 1", l, w, c)
    ok(c == 1 and l == 1)
    return next()
  end)
  seti2cb7(true)
end, true)

test('gpio toggle trigger 2', function(next)
  gpio.trig(8, "both", function(l,w,c)
    print("TRIG 2", l, w, c)
    ok(c == 1 and l == 0)
    return next()
  end)
  seti2cb7(false)
end, true)

test('gpio toggle trigger end', function(next)
  gpio.trig(8, "none")
  ok(true)
  return next()
end, true)
```

and that works if it's the end of the file, but if I replace the `gpio toggle trigger end` test with

```
test('gpio toggle trigger end', function()
  gpio.trig(8, "none")
  ok(true)
end)
```

then that one runs as soon as `gpio toggle trigger 1` yields, which isn't ideal (and, indeed, prevents the callback from firing).  I guess I understand why that has to be true, since `gambiarra` does not collect all the tests before running them but it's unfortunate.

HHHartmann:
Good to see, that you are having a look at this. I also stumbled across the fact, that the gambiarra_test file contains async tests only at the end but didn't have a closer look yet.
As you mentioned above gambiarra does not use a task.post to start the tests. Adding this would have two benefits.
1. the function would have a chance to get popped off the callstack after it executed
2. All tests would be stored in a list before executing them, regardless of async or not and get executed in the original order, one by one. That way a testrun would first execute the main program to it's end and then start the tests. This would also allow for TAP to know in advance how many tests to expect (well a statement after the last test would need to be inserted.)

I am currently fixing some other stuff and changing the ok and nok tests to abort the test instead of continuing it.

After that I will add starting the tests with a post rather than directly. So everything will be asynchronous in the end. 

Currently I am having problems if a test tails in a callback with a Lua error. They just reboot instead of failing the test. But maybe I can catch those with the new errorhandler which Terry iplemented. Will have to think about that next.
Maybe using the coroutines as I mentioned above could help fixing that too.

HHHartmann:
Added support for sync test after async and evicting old tests from callstack.
So the following code now produces expected results:
```Lua
local test = require('gambiarra')

local marker = 0
test('set marker async', function(next)
  local t = tmr.create();
  t:alarm(500, tmr.ALARM_SINGLE, function() 
    marker = "marked"
    ok(true, 'async bar')
    next()
  end)
  ok(true, 'foo')
end, true)

test('check marker sync', function()
  ok(eq(marker, "marked"), "marker check")
end)

test('check marker async', function(next)
  ok(eq(marker, "marked"), "marker check")
  next()
end, true)
```

HHHartmann:
Added catching erros and failed tests in callbacks also.
I am using the brand new `node.setonerror` functionality for this.

Next will be the integration of coroutines to run async tests as explained in https://github.com/nodemcu/nodemcu-firmware/pull/2984#issuecomment-650819656

I have it running but need to integrate it in gambiarra. Or should it stay separated?

lua_tests/gambiarra/gambiarra.lua
```diff
@@ -146,13 +147,22 @@ return function(name, f, async)
       f(handler, name, ...)
     end
 
+    local function cbError(err)
+      err = err:match(".-([^\\/]*)$") -- cut off path of filename
+      if not err:match('_*_TestAbort_*_') then
```
nwf, line 152: What's this?

HHHartmann: That is the error callback which is set in `node.setonerror(cbError)` further down which is called as new catchall errorhandler instead of rebooting the chip. Its a yet undocumented feature by Terry. Read here for more info: https://gist.github.com/TerryE/646407100785fb00c2cbddbb0e8a2b61#panic-handling

The check for '_*_TestAbort_*_' is used to distinguish failures from `ok`, `nok` or `fail` from other application filures or lua `error()` calls.

HHHartmann:
@nwf I  would like to change the interface of gambiarra to allow for separate methods for e.g. setting the output handler and also add the ability to define after-test cleanup routines and before-test prepare routines.
I'd like to hear your or everybody elses opinion on this.

the interface would then be somethink like this:

``` Lua
- local test = require('gambiarra')
+ local gambiarra = require('gambiarra')
+ local test = gambiarra.test
+ gambiarra.test("name", fn)
+ gambiarra.asynctest("name", fn)
+ gambiarra.cotest("name", fn)    -- the yet to come tests, that run in coroutines

function report(e, test, msg, errormsg)
  if e == 'begin' then
...
end
--  setting the report function, replacing the default one
- test(report)
+ gambiarra.report= report

gambiarra.cleanup = function() ... end
gambiarra.prepare = function() ... end
```

cleanup and prepare can be set as often as needed and will be valid for all tests thereafter.
Especially the cleanup part is important to clean up async tests which have pending callbacks. Imagine a tmr test failing and the timer still being armed.

nwf:
That seems like a quite reasonable API. A few thoughts:

Any reason not to run every test through `coroutine.resume()`?  Are coroutines RAM-heavy?  (I've only used them in "real computer" Lua, so heap wasn't really a concern, there.)

Given the "parse entire file into one chunk" nature of Lua's `loadfile`/`dofile` (rather than "parse file as if it were fed to the REPL"), I wonder if it's worth pondering how to simplify spreading tests across several files?  Maybe this is already sufficiently simple that it needs no new support, but it could be nice to have a worked example.

Nit: I'd prefer `gambiarra` to not be a singleton (i.e. `gambiarra = (require "gambiarra").new()` rather than just `require "gambiarra"`) so that in principle we could run several sets of tests without rebooting or having to explicitly reset the gambiarra state.

lua_tests/gambiarra/README.md
```diff
@@ -109,14 +137,17 @@ etc):
 
 Additionally, you can pass a different environment to keep `_G` unpolluted:
 
-	test(function() ... end, myenv)
+  local myenv = {}
```
nwf, line 140: Whitespace

lua_tests/gambiarra/README.md
```diff
@@ -79,22 +99,30 @@ were called with correct values.
...
 
-	test(function(event, testfunc, msg)
+	tests.report(function(event, testfunc, msg)
 		if event == 'begin' then
 			print('Started test', testfunc)
 			passed = 0
 			failed = 0
-			clock = os.clock()
 		elseif event == 'end' then
 			print('Finished test', testfunc, passed, failed, os.clock() - clock)
```
nwf, line 127: `clock` left here?

HHHartmann: removed it

HHHartmann:
> Any reason not to run every test through `coroutine.resume()`? Are coroutines RAM-heavy? (I've only used them in "real computer" Lua, so heap wasn't really a concern, there.)

That might actually make sense. It would then also be possible to map astray callbacks to their originating tests.
E.g. if I forget to stop a repeating timer in a test and it still keeps firing after the test has finished. Can't do that with async tests.

But the programming style is a lot different, so many test authors might not want to learn even more new stuff. I think keeping the entry hurdle as low as possible is a good thing to do.

NOTE: in my current commit this is not included. Things kept growing, so I decided to do a partial checkin to allow people to see the new interface. (just one feature missing)

> 
> Given the "parse entire file into one chunk" nature of Lua's `loadfile`/`dofile` (rather than "parse file as if it were fed to the REPL"), I wonder if it's worth pondering how to simplify spreading tests across several files? Maybe this is already sufficiently simple that it needs no new support, but it could be nice to have a worked example.

That could be easily done by passing the gambiarra instance around to several source files which then call the tests.
Inside gambiarra everything should get released as soon as the test has run.

But to be honest my aim is to run it out of LFS usually.

> 
> Nit: I'd prefer `gambiarra` to not be a singleton (i.e. `gambiarra = (require "gambiarra").new()` rather than just `require "gambiarra"`) so that in principle we could run several sets of tests without rebooting or having to explicitly reset the gambiarra state.

Done, you can also give a name of the run and get start and finish messages with it. I hope I didn't leave any upvalues in there.




Thinking about how much this has evolved from gambiarra I would like to also give it a name, that makes it easy to see what iti is good for.
**I'd propose to rename it to `NodeTest` or `NTest`.**
There will be a reference to the original project of course as is in gambiarra itself.

HHHartmann:
I think this is almost ready to go.
There will be some cleanup needed and some pushing around/renaming of files which I would like to get an agreement on first.

1. I would like to rename it to NTest
2. Where should the code go? a regular module or in some other place
2a. same for documentation

nwf:
I am, FWIW, unable to run the `gambiarra` selftest on my device using Lua 5.3 even with everything in LFS (sitting atop `dev` at 02e4720b3d496148235b0367d0b0292eae17988d and your `file-tests` at 715873cc2b6e4620d9de6ed67f5f91917a117b59).  What am I doing wrong?

```
dofile("gambiarra_test.lua")
> +++ Pass simple assert passes
+++ Pass simple negative assert fails
+++ Pass simple assert fails
+++ Pass simple negative assert passe
E:M 1040
E:M 2064
E:M 48
E:M 40
E:M 40
E:M 40
E:M 48
E:M 48
E:M 48
[watchdog timeout]
```

nwf:
1. Name it whatever you wish; it's your hard work, after all! :)
2. I think I'd agitate for `tests/ntests`, so that we can unify everything testing-related under `tests/`.  As to docs, maybe `tests/ntests/ntest.md` for the overview and something to wire that up into the mkdocs tree (`tests/index.md` with a pointer downwards?) so that we can add more (e.g., #2983) as time goes on.

HHHartmann:
Indeed, the memory constraints on Lua 5.3 seem to be a bit tighter.
I simplified the regexp to get the relevant stack frame from the (whole) stacktrace of the error. This is only needed if there is no explanatory message given to `ok`, `nok` and `fail`. This is expensive in terms of memory usage and has only limited benefit. Maybe I should drop it, then forcing `ok(confition, msg)` instead of having the msg optional.

It now succeeds most of the time but still prints lots of E:M nnn
If I call
``` Lua 
collectgarbage("setpause", 0)
collectgarbage("step", 100)
´´´
before the tests it always runs to the end but still prints E:M nnn s

I don't know why the E:nnn s don't lead to a failure. Is there a fallback/garbagecollect/retry in place?

File: lua_tests/gambiarra/gambiarra.lua [hunk 4]
```diff
@@ -0,0 +1,296 @@
...
+    error('_*_TestAbort_*_')
+  end
+end
+
+local function fail(handler, name, func, expected, msg)
+  if not msg then
+    -- debug.getinfo() does not exist in NodeMCU
+    -- msg = debug.getinfo(2, 'S').short_src..":"..debug.getinfo(2, 'l').currentline
+    msg = debug.traceback()
+    msg = msg:match("([^\t]*): in[^\n]*\n\t[^\n]*in function '[ROM.]*pcall'")
```
nwf, line 115: It seems odd to use `[ROM.]*` with a character class.  Is this a Lua-ism that the more apparent regex `(ROM\.)*` isn't a Pattern?

nwf (commented):
I have not given this a thorough review, sorry, but one line in the most recent diff caught my attention.

lua_tests/gambiarra/gambiarra.lua
HHHartmann: I also tried your suggestion at first but it doesn't match. So I am using this slightly too relaxed regexp.

nwf: PiL contains the surprising (to me) assertion that "Unlike some other systems, in Lua a modifier can only be applied to a character class; there is no way to group patterns under a modifier. For instance, there is no pattern that matches an optional word (unless the word has only one letter)."  So this is probably fine, but does anything go wrong if you leave off the `[ROM.]*` part, since it's being considered optional with `*`, or make it use `+` instead?

HHHartmann: The problem is that Lua 5.1 shows 'pcall' on the stack whereas Lua 5.3 shown 'ROM.pcall' in the same place.
So I tried to create a pattern that matches both.
In the meantime I remembered that Terry wrote to have the full debug implementation in Lua 5.3, so I changed the implementation back to using debug.getinfo and can remove this part of the pattern again.
Will checkin these days. I am working on freeing the result array as soon as possible instead of building up a second array holding the actual outputs and compare them to the expected at the end of the test only. Output in case of error will change for the selftest but who cares.

nwf: Thanks for explaining!

HHHartmann:
I hope it is good to go. Tested on current dev with Lua 5.1 and 5.3.
@nwf please have another look.

tests/NTest/NTest_test.lua
```diff
@@ -0,0 +1,572 @@
+local N = require('NTest')("selftest")
```
nwf, line 1: I'd almost prefer this file to be named `tests/test-ntest.lua` or something, so that `tests/NTest` was the library itself.... but maybe I'm being silly.

However, the use of `require` here along with the admonition to run NTest out of LFS would seem to mean that we also have to tell people to have `lua_examples/lfs/_init.lua` or equivalent on their devices under test.  Yes?

tests/NTest/tests/file.lua
```diff
@@ -0,0 +1,251 @@
+local N = require('NTest')("file")
```
nwf, line 1: `tests/test-file.lua`?

tests/NTest/tests/tmr.lua
```diff
@@ -0,0 +1,86 @@
+local N = require('NTest')("tmr")
```
nwf, line 1: `tests/test-tmr.lua`?

nwf (commented):
This looks really great.  Thank you for your continued hard work!

In addition to some minor point commentary below, I note that the `file` test is not currently passing on my device with Lua 5.3:
```
 ==>  fail read 9*128 bytes: size second block: different numbers expected 128 vs. 0
 ==>  fail open non existing: x: Expected to fail with Error containing "errormsg"
 ==>  fail open existing: x: Expected to fail with Error containing "errormsg"
```
is that expected?

tests/NTest/NTest_test.lua
HHHartmann: With the file naming you got exactly the spot I was unsure about.
I think that in the tests folder there will be other tests over time with other test systems, namely your #2983.
Currently it seems as if the tests can be differentiated by their filetype but what if that does not hold true anymore.
I would only expect interaction between test systems when e.g. expect calls NTest.
So I would prefer having the different test systems in different folders.

How about 
/tests/NTest/system/NTest.lua
/tests/NTest/test_file.lua   // no - as this is no lua identifier and does not permit node.LUA.whatever() to work
or may be also file_test.lua tmr_test.lua

if they are all in tests and just named test_something it is unclear to which test system they belong.

HHHartmann:
Also need to fix the luacheck errors.

tests/NTest/NTest_test.lua
nwf: Does it matter which harness (if any) they're using?  I imagine the intended flow involves pushing *all* the requisite dependencies to the device under test... though I am sensitive to the possibility of host/DUT confusion if all the files are mixed together (as they presently are in #2983... I will fix that).  How about:

* `tests/host/expectnmcu` -- TCL support libraries from #2983
* `tests/host/utils` -- additional on-host test utilities, if anything emerges as such
* `tests/host/*.*` -- files that run on the host (in whatever language, but probably shell, Lua, and/or TCL)
* `tests/dut/NTest` -- this NTest framework
* `tests/dut/utils` -- additional on-DUT test libraries or utilities, if anything ever emerges in that category
* `tests/dut/*.lua` -- things to run on the DUT, named by the module they're testing, regardless of framework.

The broader test program, then, would be to make an LFS image containing at least all of `tests/dut/**` (recursively) and `lua_examples/lfs/_init.lua` (and maybe some other in-tree utilities, as needed) and lob that over onto the device (and blank its SPIFFS) before running tests, which then get run using `dofile()` as normal.

nwf (approved):

tests/NTest/NTest_test.lua
HHHartmann: But what if a test system can be run on both, the host and sut. NTest can alreldy run its selftest on the host and as soon as #3230 (or its reimplementation after Terry is done) goes in there will be more.

Btw: replaced the hint about big code and LFS with a copy from ftpserver which contains a link to LFS but also does not mention the need of _init.lua

Also changed the layout again. See checkin.

nwf: That's fair and is a point I'd not considered.  I think I like your (implicit) proposal in the current commits to use `test/NTest/*` for NTest itself and `test/NTest_*.lua` for tests using NTest.  I'll rename my expect scripts to use a similar theme. :)

tests/NTest/NTest.lua
```diff
@@ -0,0 +1,303 @@
...
+  local function testasync(name, f)
+    testimpl(name, f, true)
+  end
+
+  local function report(f, envP)
+    if type(f) == 'function' then
+      outputhandler = f
+      env = envP or _G
+    end
+  end
```
nwf, line 252: Extremely minor and not worth not-approving the PR for: maybe rule out some dubious arguments rather than silently do nothing?
```suggestion
  local function report(f, envP)
    assert (f == nil or type(f) == 'function')
    assert (envP == nil or type(envP) == 'table')
    assert (f =~ nil or envP =~ nil)
    outputhandler = f or outputhandler
    env = envP or env
  end
```

nwf (approved):
One extremely minor thing I noticed while reading through.  If you like it as it is, I'm content to leave well enough alone.

Once again I have to thank you for all your hard work on this; NTest looks really nice.

tests/NTest/NTest.lua
HHHartmann: Good point, the check was a leftover from when the interface was only one function to activate the special functionality.
As the other methods don't have any param checks I removed them here too and just left the selective update in place.

nwf:
Is this ready to (squash) merge?

HHHartmann:
From my point of view it is ready. I'v got more ideas but that's for a later PR. First we should get some experience with this.
And Thank you for all the good feedback and ideas you had on this and your continued review and testing.

marcelstoer:
The `/tests` folder probably needs a README or index file as to get a testing overview if you navigate to the folder on GitHub? Also, maybe mention somewhere what to do about `/lua_tests`?

tests/NTest/NTest.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+You can set any of the parameters to `nil` to leave the value unchanged. 
+
+
+## Appendix
+
+This Library is for NodeMCU versions Lua 5.1 and Lua 5.3. 
+
+It is based on https://bitbucket.org/zserge/gambiarra and includes bugfixes, substantial extensions of functionality and adaptions to NodeMCU requirements.
```
marcelstoer, line 203: That is a dead end. Maybe link to https://github.com/imolein/gambiarra instead?

HHHartmann: The source you give is also based on zserges gambiarra. zserge seems to have removed the repo, so I would just link to his account https://bitbucket.org/zserge
But the repo you give has a much better readme, so we might copy some from there.

HHHartmann:
> The `/tests` folder probably needs a README or index file as to get a testing overview if you navigate to the folder on GitHub? Also, maybe mention somewhere what to do about `/lua_tests`?

The 'plan' was to convert them to NTest but as #3158 is on the way I was hoping that it would be fixed therein.
But I can convert them before that too.

pjsg:
What is unclear to me is how to actually run the tests -- are there step-by-step instructions somewhere?

nwf:
So far as I know there are no step-by-step instructions.  Eventually, I hope the answer to be a shell script on the host controlling the DUTs, and I have such a thing, more or less, somewhere between my desk and my `dev-active` branch (especially the `tap-driver.expect` program), but at the moment I think the answer is "build a LFS containing NTest and the tests you want, flash it, and then run them like any other Lua program".

PR #2985, opened 2019-12-27
Fix binary and chunked HTTP downloads. Add diffs from @anod221.

### LINKED ISSUE #2405: The HTTP module has a bug that returns a binary with a byte of \0, which ends up with the actual data that's left behind
### Test code

BUG:
The HTTP module has a bug that returns a binary with a byte of \0, which ends up with the actual data that's left behind

```Lua
http.request("http://www.dayin.la/attachment/thumb/2018-01/23/user_89580_100x100.jpg", "GET", "Range: bytes=0-100\r\n", "",  function(code, data, headers)
    if (code < 0) then
      print("HTTP request failed")
    else
      print(code, #data, string.byte(data), crypto.toBase64(data), headers["content-length"])
    end
  end)
```
data size must be is 101,but size now is 4
d2ZyAGFmZApzZGY=206	4	255	/9j/4A==	101
### NodeMCU version
Which branch are you on? If you know the Git revision then add it here as well.
nodemcu-firmware 2.2
### Hardware
Describe which ESP8266 device you use and document any special hardware setup 

HHHartmann:
Fixes #2405.

Checking the sources that @anod221 gave I created these diffs.
They look ok but I did not have a chance to test them yet.

They also fix an issue with chunked encoding replies which I also have not tested yet.

### DIFF
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -122,7 +122,7 @@ static int ICACHE_FLASH_ATTR http_chunked_decode( const char * chunked, char * d
 	 *
 	 */
 
-	return(j);
+	return(decode_size);
 }
 
 
@@ -306,6 +306,7 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
 		request_args_t	* req		= (request_args_t *) conn->reverse;
 		int		http_status	= -1;
 		char		* body		= "";
+		int		body_size = 0;
 
 		// Turn off timeout timer
 		os_timer_disarm( &(req->timeout_timer) );
@@ -411,15 +412,17 @@ static void ICACHE_FLASH_ATTR http_disconnect_callback( void * arg )
 						  body = body + 4;
 					}
 
+					body_size = req->buffer_size - (body - req->buffer);
 					if ( strcasestr( req->buffer, "Transfer-Encoding: chunked" ) )
 					{
-						int	body_size = req->buffer_size - (body - req->buffer);
-						char	chunked_decode_buffer[body_size];
+						char *chunked_decode_buffer = os_malloc(body_size);
 						os_memset( chunked_decode_buffer, 0, body_size );
 						/* Chuncked data */
-						http_chunked_decode( body, chunked_decode_buffer );
+						body_size = http_chunked_decode( body, chunked_decode_buffer );
 						os_memcpy( body, chunked_decode_buffer, body_size );
+						os_free( chunked_decode_buffer );
 					}
+					else --body_size;
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


wtarreau:
Hi!
I've just tested it now with the same method as in #3036, and it works fine, it now properly reports all downloaded bytes and does not stop at \0 anymore:
```
> http.get("http://10.9.0.22:8080/lfs.img","range: bytes=0-1023\r\n",function(code,data,headers) print(code,headers["content-length"],#data) end)
> 206   1024    1024
```

A large response still causes a panic though if I omit the range:
```
> http.get("http://10.9.0.22:8080/lfs.img","",function(code,data,headers) print(code,headers["content-length"],#data) end)
> HTTP client: Response too long (8761)
PANIC: unprotected error in call to Lua API (stdin:1: attempt to index a nil value)
```

but my understanding is that it was more or less expected anyway, and nothing prevents us from downloading large images 1kB at a time using range requests.

HHHartmann:
@wtarreau thats great news, Thaks for testing this.
I was pretty sure I wrote the message I am writing now before. Don't know why it is shown here.

In  your example ending up in a PANIC I suspect that the PANIC is either from accessing headers or data parameters to the callback which will be nil in case of an error.
The message seems to be printed by the c code. (haven't checked in the code though)

Do you have a chance to also check the chunked transfer part?

wtarreau:
On Sat, Apr 25, 2020 at 12:29:59PM -0700, Gregor Hartmann wrote:
> I was pretty sure I wrote the message I am writing now before. Don't know why
> it is shown here.

It has happened to me from time to time to lose comments when posting,
most likely due to errors when sending.

> Do you have a chance to also check the chunked transfer part?

Ah I feel ashamed, that's mainly what I wanted to test and simply forgot!
I'll do it.

Willy

marcelstoer:
@wtarreau any news?

wtarreau:
Hi Marcel. Seems like it's the second time I have to tell you I feel ashamed. I knew I had something to do last week-end and didn't remember what, given how long my todo list was and that I forgot to note it :-/  Now I've written it on my todo-list hoping I'll do it this week-end. Sorry again :-/

marcelstoer:
No worries. Don't be too hard on yourself. There are many other developers in our community - including yours truly - who could have tested this as well.

wtarreau:
Hi Marcel,

so the test is conclusive, it works. I also verified that prepending zeroes in front of the chunk sizes worked as expected:

```
server$ printf "HTTP/1.1 200 OK\r\nTransfer-encoding: chunked\r\nConnection: close\r\n\r\n1\r\nA\r\n02\r\nBC\r\n003\r\nDEF\r\n0\r\n\r\n" | nc6 --half-close -lp8181
GET /foo HTTP/1.1
Host: 10.9.0.22:8181
Connection: close
User-Agent: ESP8266
```

```
> http.get("http://10.9.0.22:8181/foo",nil,function(code,data,headers) print(code,#data,data) end)
200   6       ABCDEF
```

I had met an issue with panics when the server is not listening that caused me some gray hair, until I figured that it was my "print(#data)" which used to trigger the panic when data is nil, so everything's OK here in fact :-)

Thus for me the patch is both valid and useful, it will be very welcome!

marcelstoer:
Thanks Willy!

Issue #2958, opened 2019-11-08
Current build process only supports x86-64 hosts out of the box


TerryE:
### Expected behaviour
The NodeMCU make should work with ARM class processors such as the RPi4

### Actual behaviour
Follow the build documentation but on an RPi4.  You get this error on the make:
```
<buildroot>/tools/toolchains/esp8266-linux-x86_64-20181106.0/bin/xtensa-lx106-elf-ar: 1: <buildroot>/tools/toolchains/esp8266-linux-x86_64-20181106.0/bin/xtensa-lx106-elf-ar: Syntax error: "(" unexpected
make: *** [Makefile:337: <buildroot>/sdk/.pruned-3.0-e4434aa] Error 2
make: Leaving directory '<buildroot>'
```
### NodeMCU version
Current Master

### Hardware
RPi4
### Workarounds
The make includes the environment variable `TOOLCHAIN_ROOT` to allow the developer to specify the Xtensa toolchain path, but to simplify the build process for those who are building on a 64-bit x86 platform, the make will fetch @jmattsson's relevant prebuilt toolchain, so this make works fine most build targets.  

Our build documentation at [Linux Build Environment](/nodemcu/nodemcu-firmware/blob/master/docs/build.md#linux-build-environment) does explain that the pre-built toolchain is only supplied for 64-bit x86 platforms, but it should mention the need to set `TOOLCHAIN_ROOT` for others.

Given the increasing availability of powerful ARM-based SBCs (such as the Raspberry Pi) and Chromebooks, I feel that we should consider adding a prebuilt ARM toolchain to Johny's set.

TerryE:
I've just built the xtensa toolchain on an RPi4 and it doesn't build "out-of-the-box".  You need to enable the `deb-src` line in `/etc/apt/sources.list` and then `sudo apt-get install bison makeinfo texinfo help2man  python2.7-dev python-serial` plus a couple of other tweaks to get the toolchain to build.  I'll put up a packed tarball online tomorrow so other RPi users can avoid all of this hassle.

TerryE:
The tarball is here: https://blog.ellisons.org.uk/misc/esp8266-linux-arm_32-20191111.0.tar.xz

chathurangawijetunge:
really appreciate if you can give the steps to use "https://blog.ellisons.org.uk/misc/esp8266-linux-arm_32-20191111.0.tar.xz" with RPI

TerryE:
[...]

PR #2999, opened 2020-01-03
ESP32 Makefile: Auto-detect CPU architecture

mk-pmb:
  - Seems like Github uses the wrong (i.e. master branch's) PR template

The repo we download from only has x86_64 anyway currently, but a
download error will be much more informative than bogus errors from
an incompatible compiler.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ THIS_MK_FILE:=$(notdir $(lastword $(MAKEFILE_LIST)))
 THIS_DIR:=$(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 IDF_PATH=$(THIS_DIR)/sdk/esp32-esp-idf
 
+TOOLCHAIN_RELEASES_BASEURL:=https://github.com/jmattsson/esp-toolchains/releases/
 TOOLCHAIN_VERSION:=20181106.1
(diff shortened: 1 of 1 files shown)


HHHartmann:
I'm pretty sure @TerryE has a version for ARM architecture for raspi.

marcelstoer:
> Seems like Github uses the wrong (i.e. master branch's) PR template

Templates are available to collaborators when they are merged into the repository's default branch.

> a download error will be much more informative than bogus errors from an incompatible compiler.

Not sure I agree. While technically correct I feel that seeing a "404: Not Found" somewhere is even less obvious. If the only supported platform is Linux x86_64 anyhow we might as well ensure this in the Makefile with a precondition check.

mk-pmb:
> I feel that seeing a "404: Not Found" somewhere is even less obvious.

Well I for one would then try to "fix" the broken download URL and discover that no version for my arch is available, but the source code is. So I can then decide whether to use another computer or compile that source for the platform I tried.

mk-pmb:
I made a better error message that explicitly suggests the option of compiling the toolchain yourself.

marcelstoer:
> I'm pretty sure @TerryE has a version for ARM architecture for raspi.

It's here: #2958

PR #3010, opened 2020-01-15
document tls.cert.auth

### LINKED ISSUE #2578: tls.cert.auth should be exposed in the documentation
### Expected behavior

Documentation to include tls.cert.auth as an available function.
As indicated by issue #2576, I was able to apply both auth & ca certs to tls
in my connection to mqtt broker (which requires both) using:

tls.cert.auth(true)
tls.cert.verify(true)

### Actual behavior

Documentation excludes mention of tls.cert.auth

### Test code

The following code will work:

tls.cert.auth(true)
tls.cert.verify(true)

### NodeMCU version

Both dev and master branches don't mention  tls.cert.auth.

### Hardware

N/A

HHHartmann:
Fixes #2578.

Trying to add documentation for tls.cert.auth
Please review and tell me if it is nonsens as I don't really know much about tls.

### DIFF
--- a/app/modules/tls.c
+++ b/app/modules/tls.c
@@ -579,7 +579,6 @@ static int tls_cert_verify(lua_State *L)
   if (lua_type(L, 1) == LUA_TSTRING) {
     const char *types[2] = { "CERTIFICATE", NULL };
     const char *names[1] = { "certificate" };
-
     const char *error = fill_page_with_pem(L, &tls_server_cert_area[0], flash_offset, types, names);
     if (error) {
       return luaL_error(L, error);
--- a/app/modules/tls.c
+++ b/app/modules/tls.c
@@ -525,7 +525,8 @@ static const char *fill_page_with_pem(lua_State *L, const unsigned char *flash_m
   return NULL;
 }
 
-// Lua: tls.cert.auth(true / false | PEM data [, PEM data] )
+// Lua: tls.cert.auth(PEM data [, PEM data] )
+// Lua: tls.cert.auth(true / false)
 static int tls_cert_auth(lua_State *L)
 {
   int enable;
@@ -565,7 +566,8 @@ static int tls_cert_auth(lua_State *L)
   return 1;
 }
 
-// Lua: tls.cert.verify(true / false | PEM data [, PEM data] )
+// Lua: tls.cert.verify(PEM data [, PEM data] )
+// Lua: tls.cert.verify(true / false)
 static int tls_cert_verify(lua_State *L)
 {
   int enable;
--- a/docs/modules/tls.md
+++ b/docs/modules/tls.md
@@ -252,16 +252,16 @@ none
 
 ## tls.cert.verify()
 
-Controls the vertificate verification process when the Nodemcu makes a secure connection.
+Controls the certificate verification process when the Nodemcu makes a secure connection.
 
 #### Syntax
 `tls.cert.verify(enable)`
 
-`tls.cert.verify(pemdata)`
+`tls.cert.verify(pemdata[, pemdata])`
 
 #### Parameters
 - `enable` A boolean which indicates whether verification should be enabled or not. The default at boot is `false`.
-- `pemdata` A string containing the CA certificate to use for verification.
+- `pemdata` A string containing the CA certificate to use for verification. There can be several of these.
 
 #### Returns
 `true` if it worked.
@@ -320,6 +320,55 @@ firmware image.
 The alternative approach is easier for development, and that is to supply the PEM data as a string value to `tls.cert.verify`. This
 will store the certificate into the flash chip and turn on verification for that certificate. Subsequent boots of the nodemcu can then
 use `tls.cert.verify(true)` and use the stored certificate.
+
+## tls.cert.auth()
+
+Controls the certificate verification process when the Nodemcu authenticates against the client like when receiving a secure connection.
+
+#### Syntax
+`tls.cert.auth(enable)`
+
+`tls.cert.auth(pemdata[, pemdata])`
+
+#### Parameters
+- `enable` A boolean which indicates whether verification should be enabled or not. The default at boot is `false`.
+- `pemdata` A string containing the CA certificate to use for verification. There can be several of these.
+
+#### Returns
+`true` if it worked.
+
+Can throw a number of errors if invalid data is supplied.
+
+#### Example
+Open an mqtt client.
+```
+tls.cert.auth(true)
+tls.cert.verify(true)
+
+m = mqtt.Client('basicPubSub', 1500, "admin", "admin", 1)
+```
+For further discussion see https://github.com/nodemcu/nodemcu-firmware/issues/2576
+
+Load a certificate into the flash chip and make a request.
+
+```
+tls.cert.auth([[
+-----BEGIN CERTIFICATE-----
+CLIENT CERTIFICATE String (PEM file)
+-----END CERTIFICATE-----
+]])
+```
+
+#### Notes
+The certificate needed for verification is stored in the flash chip. The `tls.cert.auth` call with `true`
+enables verification against the value stored in the flash.
+
+The certificate can not be loaded into the flash chip at initial boot of the firmware.
+It only can be supplied by passing the PEM data as a string value to `tls.cert.auth`. This
+will store the certificate into the flash chip and turn on verification for that certificate. 
+Subsequent boots of the nodemcu can then use `tls.cert.auth(true)` and use the stored certificate.
+
+
 
 # tls.setDebug function
 
--- a/docs/modules/tls.md
+++ b/docs/modules/tls.md
@@ -323,8 +323,9 @@ use `tls.cert.verify(true)` and use the stored certificate.
 
 ## tls.cert.auth()
 
-Controls the certificate verification process when the esp authenticates against the other side 
-like when receiving a secure connection or when establishing mutual authentication like in MQTT.
+Controls the certificate proving process when the esp authenticates against the other side 
+like when establishing mutual authentication as in MQTT.  the tls.cert.auth data is used as the 
+TLS client key and certificate when the NodeMCU device is making a connection to a remote TLS server.
 
 #### Syntax
 `tls.cert.auth(enable)`
@@ -332,8 +333,8 @@ like when receiving a secure connection or when establishing mutual authenticati
 `tls.cert.auth(pemdata[, pemdata])`
 
 #### Parameters
-- `enable` A boolean which indicates whether verification should be enabled or not. The default at boot is `false`.
-- `pemdata` A string containing the CA certificate to use for verification. There can be several of these.
+- `enable` A boolean which indicates whether profing should be enabled or not. The default at boot is `false`.
+- `pemdata` A string containing the client certificate to use for profing. There second pemdata is the client private key.
 
 #### Returns
 `true` if it worked.
@@ -350,7 +351,7 @@ m = mqtt.Client('basicPubSub', 1500, "admin", "admin", 1)
 ```
 For further discussion see https://github.com/nodemcu/nodemcu-firmware/issues/2576
 
-Load a certificate into the flash chip and make a request.
+Load a certificate into the flash chip.
 
 ```
 tls.cert.auth([[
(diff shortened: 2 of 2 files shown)


nwf:
Do we know that this actually works?

dtran123:
Yes this works. If your TLS handshake doesn't take too much RAM. In my case, I have solid connections to an MQTT broker with full certification + key authentication (which I control).

docs/modules/tls.md
```diff
@@ -252,16 +252,16 @@ none
...
 ## tls.cert.verify()
 
-Controls the vertificate verification process when the Nodemcu makes a secure connection.
+Controls the certificate verification process when the Nodemcu makes a secure connection.
 
 #### Syntax
 `tls.cert.verify(enable)`
 
-`tls.cert.verify(pemdata)`
+`tls.cert.verify(pemdata[, pemdata])`
```
marcelstoer: Wrong signature.

docs/modules/tls.md
```diff
@@ -321,6 +321,55 @@ The alternative approach is easier for development, and that is to supply the PE
 will store the certificate into the flash chip and turn on verification for that certificate. Subsequent boots of the nodemcu can then
 use `tls.cert.verify(true)` and use the stored certificate.
 
+## tls.cert.auth()
+
+Controls the certificate verification process when the Nodemcu authenticates against the client like when receiving a secure connection.
```
marcelstoer: " the Nodemcu" feels odd here. Maybe also the term "client" can be ambiguous, no?

HHHartmann: Changed it. Hope it sounds better to you now. If not I'd be happy for a suggestion.

dtran123: We should give an example how one goes about configuring the Certificate + Key via tls.cert.auth()

e.g.
tls.cert.auth(
[[
-----BEGIN CERTIFICATE-----
CLIENT CERTIFICATE String (PEM file)
-----END CERTIFICATE-----
]]
,
[[
-----BEGIN RSA PRIVATE KEY-----
CLIENT PRIVATE KEY String (PEM file)
-----END RSA PRIVATE KEY-----
]]
)
tls.cert.auth(true)

Similar to the current example for tls.cert.verify()

nwf: We don't support being a TLS server, so "like when receiving a secure connection" isn't very useful (even if it would, possibly, be correct in an alternate universe in which we *did* support being a TLS server).  However, the `tls.cert.auth` data is used as the TLS **client** key and certificate when the NodeMCU device is making a connection to a remote TLS server.

docs/modules/tls.md
```diff
@@ -321,6 +321,55 @@ The alternative approach is easier for development, and that is to supply the PE
...
+
+#### Syntax
+`tls.cert.auth(enable)`
+
+`tls.cert.auth(pemdata[, pemdata])`
+
+#### Parameters
+- `enable` A boolean which indicates whether verification should be enabled or not. The default at boot is `false`.
+- `pemdata` A string containing the CA certificate to use for verification. There can be several of these.
```
nwf: This is not intended to be a CA certificate, but rather a certificate that corresponds to the client's private key, the other `pemdata` field.  There cannot, I think, be several of these.

docs/modules/tls.md
```diff
@@ -321,6 +321,55 @@ The alternative approach is easier for development, and that is to supply the PE
...
+```
+tls.cert.auth(true)
+tls.cert.verify(true)
+
+m = mqtt.Client('basicPubSub', 1500, "admin", "admin", 1)
+```
+For further discussion see https://github.com/nodemcu/nodemcu-firmware/issues/2576
+
+Load a certificate into the flash chip and make a request.
+
```
nwf: No request is made subsequent to this line.

docs/modules/tls.md
```diff
@@ -321,6 +321,55 @@ The alternative approach is easier for development, and that is to supply the PE
...
+```
+tls.cert.auth([[
+-----BEGIN CERTIFICATE-----
+CLIENT CERTIFICATE String (PEM file)
+-----END CERTIFICATE-----
+]])
+```
```
nwf: This is not a terribly meaningful invocation of the API, absent a client private key.

docs/modules/tls.md
```diff
@@ -321,6 +321,55 @@ The alternative approach is easier for development, and that is to supply the PE
...
+
+#### Notes
+The certificate needed for verification is stored in the flash chip. The `tls.cert.auth` call with `true`
+enables verification against the value stored in the flash.
+
```
nwf: Calling this verification is probably confusing; the server is doing the verifying, the client (the NodeMCU device) is doing the proving.

app/modules/tls.c
```diff
@@ -579,7 +579,6 @@ static int tls_cert_verify(lua_State *L)
   if (lua_type(L, 1) == LUA_TSTRING) {
     const char *types[2] = { "CERTIFICATE", NULL };
     const char *names[1] = { "certificate" };
-
```
nwf: Spurious?

docs/modules/tls.md
dtran123: The usecase that I have tested is when the ESP is a MQTT client connecting to a MQTT broker which enforces all three (Cert + Key + CA)

nwf: @dtran123: Just to be very clear what you mean, because there are as many as six things that might fall into "Cert + Key + CA", I think you mean that you have...

* configured a MQTT broker to enforce some TLS policy, perhaps a simple as "all clients must present TLS certificates signed by `clientCA`" (thing 1)
* installed such a certificate (thing 2) and corresponding private key (thing 3) on the ESP with `tls.cert.auth()`.
* generated a private key (thing 4) and certificate (thing 5) for the MQTT server, with the certificate perhaps signed by `serverCA` (thing 6).
* installed either the broker's certificate or `serverCA` on the ESP with `tls.cert.verify()`.

Is that right?

app/modules/tls.c
HHHartmann: Removed this to align the code for `tls_cert_verify` and `tls_cert_auth`

docs/modules/tls.md
HHHartmann: > the `tls.cert.auth` data is used as the TLS **client** key and certificate when the NodeMCU device is making a connection to a remote TLS server.

But how is this different from

## tls.cert.verify()

Controls the certificate verification process when the Nodemcu makes a secure connection.

docs/modules/tls.md
HHHartmann: Thanks for this clarification.. As I mentioned I don't really know much about this stuff. Tried to change the sentence, but probably still failed.
@nwf  you are also welcome to directly edit this PR if you like.

docs/modules/tls.md
dtran123: My mosquitto MQTT broker has the following security mosquitto.conf parameters configured: 
cafile, certfile, keyfile. Without a matching set on the ESP, the connection will fail to establish.

docs/modules/tls.md
nwf: Please consider updating the comment at https://github.com/nodemcu/nodemcu-firmware/blob/1c83f027e2e4972c66a5dd2db71d87ab3ae80a15/app/modules/tls.c#L568 as well to be less confusing.  The two signatures given here should be copied over.

docs/modules/tls.md
```diff
@@ -318,9 +318,66 @@ at `server-ca.crt` in the root of the nodemcu-firmware build tree. The build scr
...
-will store the certificate into the flash chip and turn on verification for that certificate. Subsequent boots of the nodemcu can then
+will store the certificate into the flash chip and turn on verification for that certificate. Subsequent boots of the esp can then
 use `tls.cert.verify(true)` and use the stored certificate.
 
+## tls.cert.auth()
+
+Controls the certificate proving process when the esp authenticates against the other side 
+like when establishing mutual authentication as in MQTT.  the tls.cert.auth data is used as the 
+TLS client key and certificate when the NodeMCU device is making a connection to a remote TLS server.
+
```
nwf: Perhaps just

    Controls the client key and certificate used when the ESP creates a TLS connection (for example,
    through `tls.createConnection` or `https` or `mqtt` connections with `secure = true`).

docs/modules/tls.md
```diff
@@ -318,9 +318,66 @@ at `server-ca.crt` in the root of the nodemcu-firmware build tree. The build scr
...
+#### Syntax
+`tls.cert.auth(enable)`
+
+`tls.cert.auth(pemdata[, pemdata])`
+
+#### Parameters
+- `enable` A boolean which indicates whether profing should be enabled or not. The default at boot is `false`.
+- `pemdata` A string containing the client certificate to use for profing. There second pemdata is the client private key.
```
nwf: Perhaps "A boolean, specifying whether subsequent TLS connections will present a client certificate" and "Two strings, the first containing the PEM-encoded client's certificate and the second containing the PEM-encoded client's private key."

dtran123: BTW, small typo "profing" should be "proofing"

HHHartmann: Thanks a million. That sounds really good

docs/modules/tls.md
HHHartmann: Brilliant!

HHHartmann:
Hope it is good to go now. If not please let me know.

docs/modules/tls.md
```diff
@@ -318,9 +322,65 @@ at `server-ca.crt` in the root of the nodemcu-firmware build tree. The build scr
 firmware image.
 
 The alternative approach is easier for development, and that is to supply the PEM data as a string value to `tls.cert.verify`. This
-will store the certificate into the flash chip and turn on verification for that certificate. Subsequent boots of the nodemcu can then
+will store the certificate into the flash chip and turn on verification for that certificate. Subsequent boots of the esp can then
```
marcelstoer: esp -> ESP

docs/modules/tls.md
```diff
@@ -318,9 +322,65 @@ at `server-ca.crt` in the root of the nodemcu-firmware build tree. The build scr
...
+- `enable` A boolean, specifying whether subsequent TLS connections will present a client certificate. The default at boot is `false`.
+- `pemdata` Two strings, the first containing the PEM-encoded client's certificate and the second containing the PEM-encoded client's private key.
+
+#### Returns
+`true` if it worked.
+
+Can throw a number of errors if invalid data is supplied.
+
+#### Example
+Open an mqtt client.
```
marcelstoer: mqtt -> MQTT

docs/modules/tls.md
```diff
@@ -318,9 +322,65 @@ at `server-ca.crt` in the root of the nodemcu-firmware build tree. The build scr
...
+```
+
+#### Notes
+The certificate needed for proofing is stored in the flash chip. The `tls.cert.auth` call with `true`
+enables proofing against the value stored in the flash.
+
+The certificate can not be defined at firmware build time but it can be loaded into the flash chip at initial boot of the firmware.
+It can be supplied by passing the PEM data as a string value to `tls.cert.auth`. This
+will store the certificate into the flash chip and turn on proofing with that certificate. 
+Subsequent boots of the esp can then use `tls.cert.auth(true)` and use the stored certificate.
```
marcelstoer: esp -> ESP

docs/modules/tls.md
HHHartmann: fixed all of them

docs/modules/tls.md
marcelstoer: The examples given by @dtran123 suggest otherwise and the for-loop at https://github.com/nodemcu/nodemcu-firmware/blob/9080f49df6979217e512f75d0f080a5c3520ee12/app/modules/tls.c#L497 backs this up, no?

nwf: Oh, no, right, of course the client might need to send a chain.  I retract my comment; thank you for looking.

docs/modules/tls.md
marcelstoer: How about "The client certificate the server requests to verify the callers authenticity is stored in the flash chip."

> you are also welcome to directly edit this PR if you like

Looks like you didn't set the respective flag when you created the PR. I can't edit it.

HHHartmann:
@marcelstoer found the checkbox. You should now be able to edit.

PR #3012, opened 2020-01-19
Bugfix conn metatable method call

edzius:
Here `conn` is net.socket instance, so it should be called as one.
Otherwise request is very likely to end up with crash and PANIC.

### DIFF
--- a/lua_modules/http/httpserver.lua
+++ b/lua_modules/http/httpserver.lua
@@ -89,7 +89,7 @@ do
       local buf = ""
       local method, url
       local ondisconnect = function(conn)
-	conn.on("sent", nil)
+        conn:on("sent", nil)
         collectgarbage("collect")
       end
       -- header parser


marcelstoer:
The template makes it clear that PRs have to go against `dev`. I tried to change the base branch here but yours has obviously diverged too much from upstream.

-> we can't accept the PR like that

Does this address the (one?) memory leak @hanfengcan @nwf and @HHHartmann were discussing in #2791?

HHHartmann:
It is included in #2791 too. See https://github.com/nodemcu/nodemcu-firmware/commit/331913ebec8008dc668afb6fdc79dcef068b883d

nwf:
This is obviously correct and should land.  It's a shame that #2791 is stuck and I'm sorry if that was my fault.

marcelstoer:
I don't see how that could be your fault. My proposal:

- close this PR for the reasons I mentioned
- close #2791, I agree with Gregor's final comment

Just fix the leak. I must admit that I lost track of the various changes that went into #2791. Are we talking about 1 leak or multiple?

HHHartmann:
As far as I can see this is not a memory leak but a bug. 
In #2791 the first checkin at least seems to be a leek.  There might be more (haven't checked yet).
In general I agree with your proposal above.

nwf:
"The memory leak" was never articulated in precise terms ("The $FOO and $BAR cyclically reference each other and are registered with the Lua registry", for example.)  The only setup for a leak that I know of is that `fifosock` indeed (documentedly!) creates circular references between the fifo object and the underlying socket, which is also in the Lua registry; this circularity can be broken by `nil`-ing out the `conn:on()` callbacks.  The current code attempts to do this, save for this `.`-for-`:` typo.  It is possible that it does not do so quite hard enough, in that perhaps some other callbacks must also be `nil`-ed in `ondisconnect` or `cfini` to completely strip the socket object of transitive references to the fifo object.

The first commit in #2791 `nil`-s out some additional state (`req` and `res`) which is a different kind of thing that would fix a memory leak, if that state is referenced by `conn:on()` callbacks, but I don't know exactly what's going on.  Further, given that this `.`-for-`:` bug persists in that commit, I'm skeptical that that commit was actually tested and measured improvement found.  It would be ideal if someone spent some time with the `httpserver` module and carefully examined the Lua object graph or at least looked at the evolution of `node.heap()`, but failing that, a heavy-handed approach of `nil`-ing *all* the callbacks on `conn` in both `cfini` and `ondisconnect` might be the right answer.

nwf:
ARGH.  This PR was for `master`.  That's fine, I guess...  I'll also land it on `dev`.

nwf:
Done, as 1577988d14c55eb59147f64d0d4f2fcf746c0f27.

marcelstoer:
> That's fine, I guess... I'll also land it on `dev`.

I hope so. Not sure how GitHub's gonna handle it for the next `master` drop PR. We'll see.

uDude:
I just pulled dev and I still see memory leaks using 'http-example' with 'http-server.'   about 4KB per connection.  I'm not sure the memory leaks are fixed or, perhaps I've made some silly mistake.

nwf:
@uDude Please open a new issue if there isn't one already.  We really need someone to understand the `httpserver` code and tell us what is going wrong.

PR #3013, opened 2020-01-21
Lua module for Gossip protocol

alexandruantochi:

A module based on the Gossip protocol which allows the nodes to know the state of the network without relying on a single point of failure.

https://en.wikipedia.org/wiki/Gossip_protocol

### DIFF
--- a/docs/lua-modules/gossip.md
+++ b/docs/lua-modules/gossip.md
@@ -131,11 +131,15 @@ If declared, this function will get called every time there is a `SYN` with new 
 gossip.pushGossip(data, [ip])
 
 -- remove data
-gossip.networkState[gossip.ip].data = nil
+gossip.pushGossip(nil, [ip])
 ```
 
-Send a `SYN` request outside of the normal gossip round. The IP is optional and if none given, it will pick a random node. If you want to remove the data form the network you have to manually set it to nil.
+Send a `SYN` request outside of the normal gossip round. The IP is optional and if none given, it will pick a random node.
 
+```
+!!! note
+. By calling `pushGossip(nil)` you effectively remove the `data` table from the node's network state and notify other nodes of this.
+```
 ## setRevManually()
 
 #### Syntax
--- a/docs/lua-modules/gossip.md
+++ b/docs/lua-modules/gossip.md
@@ -5,7 +5,7 @@
 | 2020-01-20 | [alexandruantochi](https://github.com/alexandruantochi) | [alexandruantochi](https://github.com/alexandruantochi) | [gossip.lua](../../lua_modules/gossip/gossip.lua) |
 
 
-This module is based on the gossip protocol and usefull for use cases when multiple controllers are on the same network and they need to know about each other's state or when a single point of failure (such as an entity that queries each controller in particular) is to be avoided. It runs using UDP, thus providing faster communication and less network noise. 
+This module is based on the gossip protocol and usefull for use cases when multiple controllers are on the same network and they need to know about each other's state. It can also be used for spreading information in the network. It runs using UDP, thus providing faster communication and less network noise. 
 
 ## Usage
 ```lua
@@ -24,6 +24,8 @@ gossip.start()
 Each controller will randomly pick an IP from it's seed list. It will send a `SYN` request to that IP and set recieving node's `state` to an intermediary state between `Up` and `Suspect`. The node that recieves the `SYN` request will compute a diff on the received networkState vs own networkState. It will then send that diff as an `ACK` request. If there is no data to send, it will only send an `ACK`. When the `ACK` is received, the sender's state will revert to `Up` and the receiving node will update it's own networkState using the diff (based on the `ACK` reply).
 
 Gossip will establish if the information received from another node has fresher data by first comparing the `revision`, then the `heartbeat` and lastly the `state`. States that are closer to `DOWN` have priority as an offline node does not update it's heartbeat.
+
+Any other parameter can be sent along with the mandatory `revision`, `heartbeat` and `state` thus allowing the user to spread information around the network.
 
 Currently there is no implemented deletion for nodes that are down except for the fact that their status is signaled as `REMOVE`.
 
@@ -119,7 +121,8 @@ Example:
   "192.168.0.53": {
     "state": 3,
     "revision": 25,
-    "heartbeat": 2500
+    "heartbeat": 2500,
+    "extra" : "this is some extra info from node 53"
   },
   "192.168.0.75": {
     "state": 0,
--- a/docs/lua-modules/gossip.md
+++ b/docs/lua-modules/gossip.md
@@ -44,9 +44,9 @@ Currently there is no implemented deletion for nodes that are down except for th
 
 There are multiple modules on the network that measure temperature. We want to know the maximum and minimum temperature at a given time and have every node display it.
 
-The brute force solution would be to query each node in particular and save the `min` and `max` values from a single point, then go back to each node and present them with the computed `min` and `max`.
+The brute force solution would be to query each node from a single point and save the `min` and `max` values, then go back to each node and present them with the computed `min` and `max`. This requires n*2 rounds, where n is the number of nodes. It also opens the algorithm to a single point of failure (the node that is in charge of gathering the data).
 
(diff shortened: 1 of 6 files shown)


marcelstoer:
This fails the CI build. Do you have access to https://travis-ci.org/nodemcu/nodemcu-firmware/builds/639717733?

alexandruantochi:
> This fails the CI build. Do you have access to https://travis-ci.org/nodemcu/nodemcu-firmware/builds/639717733?

Yep. I'll have a look to see what's wrong.

alexandruantochi:
Fixed warnings from luacheck.

marcelstoer:
> luacheck errors regarding unused variables

Yep, we have to thank @galjonsfigur for that 😃 👍

alexandruantochi:
> > luacheck errors regarding unused variables
> 
> Yep, we have to thank @galjonsfigur for that 😃 👍

I had one unused variable because of an IF statement. 

I also had 2 that I was forced to use due to the [UDPsocket:on('receive', callback);](https://nodemcu.readthedocs.io/en/master/modules/net/#netudpsocketon) function. Maybe in the future we could have a way to ignore luacheck warnings for cases like this.

HHHartmann:
I am not sure if I get the purpose correctly.
This is "only" to find other esps on the network? Would it also be possible to share some payload between the devices? Or am I missing something?

Would it make sense to identify the devices by their MAC address instead of the IP? So one would nit have to assign fixed IPs.

I like your coding style. Makes a nice separation of concerns.

Btw:
Luacheck allows for
```
-- luacheck: push no unused
local MCP23008_IODIR = 0x00
-- luacheck: pop
```

mkdocs.yml
```diff
@@ -51,6 +51,7 @@ pages:
     - 'mcp23008': 'lua-modules/mcp23008.md'
     - 'redis': 'lua-modules/redis.md'
     - 'yeelink': 'lua-modules/yeelink.md'
+    - 'gossip': 'lua-modules/gossip.md'
```
HHHartmann: Please sort alphabetically

lua_modules/gossip/gossip.lua
```diff
@@ -0,0 +1,268 @@
...
+local gossip = {};
+local constants = {};
+local utils = {};
+local network = {};
+local state = {};
+
+-- Utils
+
+utils.debug = function(message)
+    if gossip.config.debug then
```
HHHartmann: I think the coding style is 2 space indentation. Couldn't find a counter sample but I am not sure if it is documented.

alexandruantochi:
@HHHartmann it's not for finding nodes on the network, it's for knowing their real time state (online or offline) and distributing this information between them.

A payload can be sent along with the mandatory `revision`, `heartbeat` and `state` values and this should be the user's own implementation.

e.g. sending this JSON works and the receiving node will have the extra param `payload`, which will be propagated to all the nodes in the network. Once the node sends another update with another payload, this information will be propagated as the heartbeat would be > 260

```JSON
{ "type" : "ACK", "192.168.0.158" : { "heartbeat" : 260, "revision" : 1, "state": 0, "payload" : "banana" } }
```

docs/lua-modules/gossip.md
```diff
@@ -0,0 +1,131 @@
+# ESPGossip
+
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2020-01-20 | [alexandruantochi](https://github.com/alexandruantochi) | [alexandruantochi](https://github.com/alexandruantochi) | [gossip.lua](../../lua_modules/gossip/gossip.lua) |
+
+
+This module is based on the gossip protocol and usefull for use cases when multiple controllers are on the same network and they need to know about each other's state or when a single point of failure (such as an entity that queries each controller in particular) is to be avoided. It runs using UDP, thus providing faster communication and less network noise. 
+
```
HHHartmann: Most documentation includes sections like 
## require
and
## release
followed by the functions.
There are however also other modules which don't follow this style.

HHHartmann:
I alos wonder if it would make sense to allow for a callback on changed data which would be feed with `updatedNodes` from `updateNetworkState` or better with an array of these nodes (whole node or just the IP). Would also be helpful to have a sort of `push` when changing the `extra` information.

alexandruantochi:
> I alos wonder if it would make sense to allow for a callback on changed data which would be feed with `updatedNodes` from `updateNetworkState` or better with an array of these nodes (whole node or just the IP). Would also be helpful to have a sort of `push` when changing the `extra` information.

New data is expected to come in continuously because each node will increment it's hearbeat before sending a `SYN` request. The idea is good, I could add a callback function that gets the `ip` and `nodeData` as parameters after the networkState has updated. I'll put it on my todo list.

marcelstoer:
https://travis-ci.org/nodemcu/nodemcu-firmware/builds/645237431

alexandruantochi:
> https://travis-ci.org/nodemcu/nodemcu-firmware/builds/645237431

I think everyone can see the link from the check, no need to bother adding it as a comment:
image

marcelstoer:
I cannot know what you do and don't see. Previously it was suggested that non-committers wouldn't see the CI report link. Good to know this isn't the case (anymore).

alexandruantochi:
Just trying to save you some trouble. Yeah, comment can be interpreted as me being irritated, sorry about that, it was not my intention  😆

alexandruantochi:
Oh alright, I though your were adding it for me and wanted to save you some trouble.

lua_modules/gossip/gossip.lua
```diff
@@ -0,0 +1,275 @@
...
+  setConfig = utils.setConfig,
+  start = state.start,
+  setRevFileValue = state.setRevFileValue,
+  networkState = {},
+  getNetworkState = utils.getNetworkState,
+  pushGossip = network.pushGossip
+};
+
+-- unit tests
+
```
HHHartmann: Having to change the sources does not work well together with automated testing.
Using something like this together with a change in the test might help
``` Lua
-- load with loadfile("gossip.lua")("test") to run gossip_tests
if (... == "test") then
  gossip._test = {
   _gossip = gossip,
   _constants = constants,
   _utils = utils,
   _network = network,
   _state = state
  }
end
```

lua_modules/gossip/gossip_tests.lua
```diff
@@ -0,0 +1,277 @@
+-- Gossip protocol implementation tests
+-- https://github.com/alexandruantochi/
+local gossipSubmodules = require('gossip');
+
```
HHHartmann: See above for the other part of the change.
``` Lua
local gossipSubmodules = loadfile("gossip.lua")("test");

local gossip = gossipSubmodules._test._gossip;
local constants = gossipSubmodules._test._constants;
local utils = gossipSubmodules._test._utils;
local network = gossipSubmodules._test._network;
local state = gossipSubmodules._test._state;
```

lua_modules/gossip/gossip_tests.lua
```diff
@@ -0,0 +1,277 @@
...
+-- test constants and mocks
+
+local function dummy() return nil; end
+
+-- luacheck: push allow defined
+tmr = {};
+tmr.time = function() return 200; end
+sjson = {};
+sjson.decode = function(data) return data; end
+file = io;
```
HHHartmann: `file = {}` makes it run on device too

HHHartmann:
Still found no time to test this but what I saw of the code looks good. So if you don't want to wait, this could go from my side.

alexandruantochi:
Have a look first, I've learned from demos that the probability of failing is higher when somebody else is watching 😁

alexandruantochi:
@marcelstoer what are the next steps I have to take to get this merged?

marcelstoer:
Thanks for the nudge and thank you for your contribution. You don't need to do anything. Gregor already approved it a while ago. I guess he won't disapprove of me merging it then.

HHHartmann:
Sorry for not reacting earlier. I was busy preparing vacation and planned to find some time on vacation but then went ill.
So now this is out to a bigger field experiment. 
But as i said before it all looks really good. Thanks for the middle.

alexandruantochi:
> Sorry for not reacting earlier. I was busy preparing vacation and planned to find some time on vacation but then went ill.
> So now this is out to a bigger field experiment.
> But as i said before it all looks really good. Thanks for the module. .

I'm happy that you tested it and it performed well. I only had 4 modules available to run it on. Could you give me some details on your setup?

HHHartmann:
Ah sorry. I didn't test it yet. It's still on my Todo list. I will also only test with 2 modules, so this probably will not add much.

PR #3022, opened 2020-01-29
Clarify LFS size build info means capacity

mk-pmb:

Trivial clarification that the LFS build info means its capacity. For a while I was unsure about it because mine shows zero, so I was wondering if it might mean the amount used, or the start position.

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -163,7 +163,7 @@ static int node_info( lua_State* L )
       int table_index = lua_gettop(L);
       lua_pushboolean(L, BUILDINFO_SSL);
       lua_setfield(L, table_index, "ssl");
-      lua_pushnumber(L, BUILDINFO_LFS);
+      lua_pushnumber(L, BUILDINFO_LFS_SIZE);
       lua_setfield(L, table_index, "lfs_size");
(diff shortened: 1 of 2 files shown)


HHHartmann:
The branch also has to be based on dev. Maybe you can rebase it.

mk-pmb:
Base branch should be fixed now. At first I had assumed the checkbox was about the Git tree, not the Github UI.

marcelstoer:
> I had assumed the checkbox was about the Git tree, not the Github UI.

They go hand-in-hand.

HHHartmann:
This seems to be ok unless someone is parsing the output. In that case it would break that parser.

TerryE:
I missed this one sorry.  I've was just merging these changes and in doing do reviewed this `node.c` code.  @HHHartmann there are weaknesses in this approach of mining data from the make to generate runtime configuration data:
1.  This is denormalising information that is available elsewhere, for example you can enumerate the tables in `ROM` to list all included modules and other `file` and `node` API calls which give the correct runtime configuration values.
2.  The answers returned can be just _wrong_.  For example the LFS region size is based on the value in the Partition Table which can be different to the build default as this can be changed by editing the Partition Table after imaging or by the `node.setpartitiontable()` API call.

Issue #2903, opened 2019-09-08
broke what previously worked


Evonic4:
8<------------------------ BUG REPORT -----------------------------------------
### Expected behavior
>>expired

### Actual behavior
init.lua:1: attempt to call field 'create' (a nil value)

### Test code
https://nodemcu.readthedocs.io/en/master/modules/tmr/#tobjstart
```lua
local mytimer = tmr.create()
mytimer:register(5000, tmr.ALARM_SINGLE, function (t) print("expired"); t:unregister() end)
mytimer:start()
```

### NodeMCU version
2.2.1-master_20190405

### Hardware
ESP8266 device v3

8<------------------------ END BUG REPORT -------------------------------------

nwf:
It's hard to fathom how that could be true, but in any case, 20190405 was a while ago; could you try with a newer release?

HHHartmann:
Actually that is the latest release as the current one is not yet announced properly.

Evonic4:
need fix

HHHartmann:
I just checked on my device and it works (as expected)
@Evonic4 can you give more information about your firmware?
If you built from frightanic or docker please give the startup message.
Or can you please try a frightanic build (https://nodemcu-build.com/) or docker build from clean sources?

Evonic4:
and how many modules did you include, perhaps this is due to incompatibility?
I collected 16 modules
 (https://nodemcu-build.com/)

Evonic4:
still this error

NodeMCU 0.9.5 build 20150318  powered by Lua 5.1.4
Waiting ...
lua: init.lua:2: attempt to call field 'register' (a nil value)
>

nwf:
@Evonic4: No matter how many times you say "need fix", any fix will not arrive faster.  If you are so desperate for a fix, I can be hired at reasonable professional contracting rates; you're welcome to contact me out of band to arrange such.

You have not provided anyting useful in terms of feedback.  Please, for example, articulate *which* 16 modules you selected in the node builder.

`20150318` is a *very*, *very* long time ago and we certainly no longer support that version.  Are you sure you have flashed your device with the version you claimed in the original report, (`2.2.1-master`)

Evonic4:
here is the feedback to you

1. modules
cron crypto file gpio http i2c net node pcm pwm rtctime spi **tmr** uart websocket wifi.

2. last release used
https://github.com/nodemcu/nodemcu-firmware/releases
[...]

PR #3026, opened 2020-01-31
No banner

### LINKED ISSUE #3021: disable startup banner
Please add support for disable startup banner

### Workarounds
comment out `print_version(L);` in `app/lua/lua.c`

532910:
Fixes #3021.

### DIFF
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -35,7 +35,7 @@
 //         modules: file,gpio,net,node,rtctime,sntp,tmr,uart,wifi
 //  build 2020-01-27 17:39 powered by Lua 5.1.4 on SDK 3.0.2(824dc80)
 //
-// will be printed to serial console.  While it's very helpful for bug reports
+// will be printed to serial console.  While it's mandatory for bug reports
 // and good for development, it may be unwanted for non-interactive serial
 // devices.
 
--- a/app/include/user_config.h
+++ b/app/include/user_config.h
@@ -22,6 +22,26 @@
 //#define BIT_RATE_AUTOBAUD
 
 
+// At start-up firmware details like:
+//
+// NodeMCU 3.0.1.0
+//         branch:
+//         commit:
+//         release:
+//         release DTS:
+//         SSL: false
+//         build type: integer
+//         LFS: 0x0
+//         modules: file,gpio,net,node,rtctime,sntp,tmr,uart,wifi
+//  build 2020-01-27 17:39 powered by Lua 5.1.4 on SDK 3.0.2(824dc80)
+//
+// will be printed to serial console.  While it's very helpful for bug reports
+// and good for development, it may be unwanted for non-interactive serial
+// devices.
+
+//#define DISABLE_STARTUP_BANNER
+
+
 // Three separate build variants are now supported. The main difference is in the
 // processing of numeric data types.  If LUA_NUMBER_INTEGRAL is defined, then
 // all numeric calculations are done in integer, with divide being an integer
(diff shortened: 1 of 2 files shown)


TerryE:
@532910 Sergio, you need to read the contribution guidelines.   The main issues here are:

1.  I don't think that we have a consensus on the requirements so a PR is premature IMO.
2.  You need to read the contribution guidelines, and in particular that you should base your changes on a pull of the latest `dev` version and not you potentially out of date version.
3.  IMO the comment in `user_config.h`is far too verbose.

But let's get one sorted first.

marcelstoer:
I fixed 2. by changing the base from master to dev here on GitHub.

> add x inside the brackets

This was ignored so I fixed that as well BUT 

> PR is premature IMO

is still valid of course.

532910:
1. Agree!
2. I've read https://github.com/nodemcu/nodemcu-firmware/blob/master/CONTRIBUTING.md
Could you explain what have I done wrong? The steps I did:
* Forked from github web interface
```
% git clone git@github.com:532910/nodemcu-firmware
% cd nodemcu-firmware
% git checkout -b no-banner origin/dev
# modified files
% git push origin HEAD
```
* PR via github web interface

marcelstoer:
> Could you explain what have I done wrong? The steps I did:
...
PR via github web interface

When you do this you need to select the "base" branch i.e. the one you want the PR to be merged into - which is normally the same you branched from. GitHub always suggests the default branch but you need to change that to `dev`. https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request#creating-the-pull-request -> step 4

You can easily verify that because the list of commits/changes in the PR is really long if you select the wrong base branch. If you picked the wrong one you can hit the "Edit" button at the top-right and [correct it](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request#changing-the-branch-range-and-destination-repository). That's what I fixed.

nwf:
Just as a point of interest, in the non-interactive test framework I've been prototyping, I rely on the startup banner to detect when the device has booted as expected or rebooted unexpectedly.  I'm not sure why its presence would be a problem.  Moreover, I'm jittery about removing it entirely given that we get so many bug reports from people who really need to read it (#2903) -- if it were gone, there'd be no chance of them reading it, rather than merely a slim chance.

532910:
@nwf, see #3021. The banner is for users and development process. But it's bad for other devices.
For example I have a serial screen and I want no messages to be send to it without my permission.

marcelstoer:
I'm ok with this as long as it's enabled by default.

532910:
It should be enabled by default, of course.

nwf:
This PR looks fine to me (though I might suggest changing "very helpful for bug reports" to "mandatory for bug reports"), but...

You will probably be better served by changing the hardware a bit:
* a bit-bang UART on a GPIO entirely under your control,
* an AND gate and a GPIO gating one of the existing hard UARTs, or
* a dedicated UART chip (e.g. via I2C and a SC16IS740) that won't be spoken to without your permission.

532910:
> an AND gate and a GPIO gating one of the existing hard UARTs, or

Yes, that's exactly what I did.

532910:
> "very helpful for bug reports" to "mandatory for bug reports"

agree

532910:
> When you do this you need to select the "base" branch i.e. the one you want the PR to be merged into - which is normally the same you branched from.

Clear, Marcel thank you for comprehensive explanation!

PR #2996, opened 2020-01-03
Fixes to softuart module

galjonsfigur:
Fixes #2993

- [x] This PR is for the `dev` branch rather than for `master`.
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [ ] I have thoroughly tested my contribution.

I would be thankful if @mk-pmb could test it because to reasons unknown to me I couldn't reproduce the warning, even with `-Wall `option added.

PR #3029, opened 2020-01-31
Fix timing of softuart module RX

galjonsfigur:

Fix for issue mentioned in #2996 by @NicolSpies. Now ISR will only wait for stop bit only if it has to and that should fix the problem with bit shifts in some data frames. I couldn't test it with hardware yet so I would be glad if @NicolSpies (or anyone really) could test it.

### DIFF
--- a/app/modules/softuart.c
+++ b/app/modules/softuart.c
@@ -13,9 +13,12 @@
 #define SOFTUART_MAX_RX_BUFF 128
 #define SOFTUART_GPIO_COUNT 13
 
+//TODO: Overflow flag as callback function + docs
 typedef struct {
 	char receive_buffer[SOFTUART_MAX_RX_BUFF];
-	uint8_t length;
+	uint8_t buffer_first;
+	uint8_t buffer_last;
+	uint8_t bytes_count;
 	uint8_t buffer_overflow;
 } softuart_buffer_t;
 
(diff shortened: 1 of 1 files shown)


NicolSpies:
@galjonsfigur , I tested the change as before by writing 8 frames to external device and using a digital scope to manually decode the return 11 frames and comparing them to the softuart `on data` return.

**Consistent and repeatable results**
`softuart.port:write(data)` still ok.

`softuart.port:on('data',..` still the same as reported before.

before the change:
The data decoded for frames 1,3,5,7,9,.. decode correctly.
The data decoded for frames 2,4,6,8,.. decode incorrectly, data shifted 1 bit to the left.

after latest timing fix:
The data decoded for frames 1,3,7,9,.. decode correctly.
The data decoded for frames 2,4,6,8,.. decode incorrectly, data shifted 1 bit to the left.
The data decoded for frame 5 decode incorrectly, 1 of the 8 bits in the frame decoded incorrectly.

galjonsfigur:
@NicolSpies Thanks for the tests! I forgot to ask: what exact trigger method do you use? By that I mean how exactly your `softuart.port:on` code looks like. I assumed that you used `s:on("data", 1, <rest of the code>` and I don't know if that's the case. It seems like ISR just takes too long to run and loses some bits of information. I think that triggering softuart callback on every received character is not working properly. I think that there are some options to resolve this:
- **1**. Use other trigger method like triggering on a certain character or after more characters - then it should work just fine
- **2**. Use timer and volatile flag instead of task_post - I know that [Arduino Software Serial](https://github.com/plerup/espsoftwareserial) has `perform_work` function called in `loop` to make use of callbacks possible and that should be faster but I'm not sure.
- **3**. Try to optimize the ISR (for example try to do some stuff like checking trigger conditions when waiting for next bit in data frame)

I will try to check that soon as I recently got an digital oscilloscope.

NicolSpies:
@galjonsfigur , using `s:on"data",11, function(data)..` to capture the 11 bytes of incoming data, where a data frame starts with `0` start bit and ends with `1` stop bit and no parity bit. So a data frame is 10 bits at 115200 baud.

I must add the received frames have no "dead space" in between frames, so the frame stop bit and start bit of the next frame are back to back. So I agree it seems like a timing issue every second frame. An extra `0` is actually added after the start bit of alternate frames resulting in the physical right shift. I will try and slow the source device baud rate down, if it is timing it will become less critical.

NicolSpies:
@galjonsfigur , my suspicion was correct. It is a timing issue. When I slow the baud rate down one step from 115200 to 57600 it consistently decodes correctly.

galjonsfigur:
@NicolSpies 
OK - I found the problem. My code was based on [plieningerweb's work](https://github.com/plieningerweb/esp8266-software-uart) and at time I didn't understood it fully. Now I looked into that code and I realized I simplified buffer data structure and forgot about it halfway in the code :smile: . I will repair it and force-push this PR soon.

galjonsfigur:
@NicolSpies I fixed my stupid mistake in the code and tested reception on 115200 baud - it seems like it's working properly but again - your tests would be very helpful. 

I think this module can be improved quite a bit - additional callback on buffer overflow, UART modes other than 8N1, Lua examples, support for reading from buffer without callback... but that's for another time or another contributor :smile:

NicolSpies:
@galjonsfigur, great, I will implement the latest changes and test tomorrow. Fortunately I only need 8N1 to write a Modbus messages of 8 bytes and receive the Modbus replies of 7 bytes. I will thus not suffer from buffer overflow etc.

PR #3048, opened 2020-02-26
Update getting-started docs hrefs for LFS examples

### LINKED ISSUE #3045: Inconsistent href links in docs
### Expected behavior
The hrefs links in the docs appear to be inconsistent (or incorrect?). Consider the [Getting Started](https://github.com/nodemcu/nodemcu-firmware/blob/master/docs/getting-started.md#compile-lua-into-lfs-image) page where the links to the example code point to _tree/dev_. I would expect these hrefs to link to the _master_ branch.

### Actual behavior
The result of this is that _tree/master_ docs that are published point to code that is in the _dev_ branch. In the case of the [telnet example](https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua_examples/telnet/telnet.lua) the file doesn't exist.

I can't tell if this intentional or an artifact of the automation tooling. I'm happy to help fix up some links as I find problems. Thanks.

cerasmus:
Ensure that the href links under the LFS section points to the master branch on GitHub.
See issue #3045

Fixes #3045.

The build docs on RTD are done against the master branch. This PR updates the LFS examples links on the Getting Started page to point to the master branch instead of dev

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -268,10 +268,11 @@ The easiest approach is to maintain all the Lua files for your project in a sing
 
 For example to run the Telnet and FTP servers from LFS, put the following files in your project directory:
 
-* [lua_examples/lfs/_init.lua](https://github.com/nodemcu/nodemcu-firmware/tree/master/lua_examples/lfs/_init.lua).  LFS helper routines and functions.
-* [lua_examples/lfs/dummy_strings.lua](https://github.com/nodemcu/nodemcu-firmware/tree/master/lua_examples/lfs/dummy_strings.lua).  Moving common strings into LFS.
-* [lua_examples/telnet/telnet.lua](https://github.com/nodemcu/nodemcu-firmware/tree/master/lua_examples/telnet/telnet.lua).  A simple **telnet** server.
-* [lua_modules/ftp/ftpserver.lua](https://github.com/nodemcu/nodemcu-firmware/tree/master/lua_modules/ftp/ftpserver.lua).  A simple **FTP** server.
+* [lua_examples/lfs/_init.lua](../lua_examples/lfs/_init.lua).  LFS helper routines and functions.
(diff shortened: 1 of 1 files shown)


marcelstoer:
> The build docs on RTD are done against the master branch.

No, there's a separate set of docs for each branch with `master` being the default. See the branch selector in the lower-right corner of the RTD site.

https://nodemcu.readthedocs.io/en/master/
https://nodemcu.readthedocs.io/en/dev/
etc.

Hence, this change doesn't really fix anything. There is still a lot of room for inconsistencies as pointed out in #3045.

You can make them all relative so that viewing here on GitHub e.g. at https://github.com/nodemcu/nodemcu-firmware/blob/master/docs/ they will point to the correct branch. That, however, will most likely produce broken links on RTD.

marcelstoer:
> That, however, will most likely produce broken links on RTD.

I stand to be corrected. See how the source code links at the top of every module document seem to always point to the correct branch.

HHHartmann:
So the problem is that the file has been renamed and that must be corrected.

HHHartmann:
The links have to look like this:  `(../../lua_modules/http/httpserver.lua)`

cerasmus:
Ok, thanks for the feedback. I missed some of the finer details.

1. I was thinking of using relative paths to fix it also. I can make this change and test. Are you happy if I take this approach.
2. I can also do a rename on the example file and fix that mistake also. Are you happy if I do this also?

marcelstoer:
@cerasmus Thanks for improving the docs.

HHHartmann:
@cerasmus please go ahead with 1.
For 2. You mean adapting the link to the New filename? The rename took place because there are two implementations now. Maybe linking to both of them does make sense. 
Thanks for taking care!

cerasmus:
I added links that are relative to their location in .../docs. I couldn't get the relative links to work on my ReadTheDocs account but it works for yours. I assume the last commit will then be correct. Additionally, I linked to both the telnet example files.

Issue #3032, opened 2020-02-02
espconn_secure/tls: eliminate funky flash game


nwf:
### Missing feature

I would like to propose that the current `tls.cert.verify` and `tls.cert.auth` API is unfortunate.  It directly accesses particular sectors of flash and is holographically embedded within both `app/mbedtls/app/espconn_secure.c` and `app/mbedtls/app/espconn_mbedtls.c`.  I would prefer, instead, I think, either a more elaborate per-connection configuration mechanism or, failing that,
that the API be such that these two functions take Lua functions that return certificates or keys when invoked:

```
tls.cert.verify(function(ix) if ix == 1 then return file:read("ca.pem") end end)
tls.cert.auth(function(ix) ... end) -- ix 0: private key, 1: cert, 2+: chain elements
```

We can encode the current behavior by making the PEM-string forms of these functions write to 
a particular name in SPIFFS and make the boolean forms install default functions that read those files.

### Justification

At least in the test harness I'm working on, there's a fair bit of juggling of SSL keys and certs, and I'd much rather have these as files in the SPIFFS than in particular sectors of flash.

### Workarounds

Lots and lots of flash overwrites.

TerryE:
@nwk, are you aware of the platform_rcr functions (read/write/delete) ?  But that aside, I also feel on balance that it is a lot simpler to use SPIFFS as the repository for such data.

nwf:
I was not aware of the `platform_rcr` functions, no.  If we wanted to store the certs over there, I'd prefer that we expose the RCR records to Lua (perhaps as userdata rather than strings, necessarily) rather than adopt the existing, exciting flash game directly.

I admit I've stalled on actually *eliminating* the funky TLS flash goo because I haven't yet figured out a good replacement for the `Makefile`'s ability to bake in a certificate: 
* https://github.com/nodemcu/nodemcu-firmware/blob/master/Makefile#L402
* https://github.com/nodemcu/nodemcu-firmware/blob/master/tools/make_server_cert.py
* https://github.com/nodemcu/nodemcu-firmware/blob/8d091c476edf6ae2977a5f2a74bf5824d07d6183/app/modules/tls.c#L25

Because our build system doesn't, historically, construct LFS or SPIFFS images as part of its operation, there's presently no real other way to do this.  Given the existence of RCRs.... maybe we *should* expose those to Lua (at least in a RO way)... hm.  This may also be related to #3117 and #3128 and attendant platform support.

[...]

PR #3006, opened 2020-01-06
WIP: Espconn delenda est

nwf:
- [x] This PR is for the `dev` branch rather than for `master`.
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [ ] I have thoroughly tested my contribution.
- [x] The code changes are reflected in the documentation at `docs/*`.

These are addressing #3004, but there is a long way to go.  Still, open PRs early and often!

Issue #3004, opened 2020-01-05
espconn lwip wrappers actively harmful


nwf:
As far as I can tell, anyway, these things are just obstruction, rather than abstraction, over lwip.  See, for example, the damage they cause with #2987: it is neither obvious nor documented that `espconn_gethostbyname` passes the espconn structure as the callback argument, and even if it were, it's also not obvious that this could be cast back to the `lmqtt_userdata` structure of which it is the first member.  We should just be calling `dns_gethostbyname` ourselves and passing the appropriate lua object to the callback.

Am I missing something and espconn is actually useful?  If not, I think I will propose that we attempt to excise it from our tree.

nwf:
Still true, sadly.

HHHartmann:
Is it still true?

PR #3060, opened 2020-04-02
Networking rampage and accumulated fixes

nwf:

This represents the work done thus far on a few issues; #3032 and easy aspects of #3004 (xref #3006) as well as refactoring and simplification around the network-facing things I use regularly.  I've been carrying these patches locally for a while and would like to push them upstream, at least to `dev`, for others to beat about a bit and so that other things stop feeling quite so blocked on these.  Much of this has been subject to testing via the test harness in my https://github.com/nwf/nodemcu-firmware/tree/dev-active branch (recall #2983), so I feel fairly confident in having not broken anything obvious.

The `mqtt` and `tls` "stop using espconn->reverse" refactoring could be done for other network-facing modules (`http`, `coap`, and `websocket`) as well and would probably be similarly beneficial.  Then we can remove the `->reverse` field from `espconn` and make a little more progress on #3006.  I don't feel confident in changing code in the remaining ones, absent test support.

ETA: removed `espconn->reverse` from `tls`, too, and fixed a stack imbalance issue in the new TLS cert/key callback code.


nwf:
ETA, again: found and fixed a memory leak in the new `tls` cert/key callback API (just a typo, sorry).  I've adjusted my test harness to exercise the new API better and believe that this class of bug is no longer present.

marcelstoer:
Cool! I like every PR that removes more code than it adds. Can we land this on `dev` sooner rather than later?

nwf:
*I*'m happy to hit merge and deal with any fallout, but I suspect it would be better if someone else gave things a quick run-through first. ;)

marcelstoer:
I think it would very valuable (and thus welcome by the community) if we merged dev back to master in the coming weeks. It's been almost six months but there are a few things pending: https://github.com/nodemcu/nodemcu-firmware/milestone/14. So, I wouldn't mind landing it on dev soonish as it looks quite mature.

nwf:
I'm worried about the `tls` callback changes (1cc2040), which probably merit some more discussion than they've gotten (as part of #3062).  I've pulled those from this PR and will open another for the next round of rampaging.

nwf:
Argh, I didn't really mean to squash all those together.  I'm feeling jittery about more of them, now, too, but whatever, I think I can live with some egg on my face and follow-up commits to fix bugs as we find them.

PR #3064, opened 2020-04-10
More, but more subdued, rampaging for next release

nwf:

This is another round of minor scattered changes throughout the tree.  I'd like at least `tls: fix new verification API` to land because it corrects an embarrassing error on my part.

I'd appreciate someone glancing over this, especially some of the more obscure Lua bits (`__metatable`, `LUA_GCSTOP`, ...).


app/modules/net.c
```diff
@@ -857,12 +857,19 @@ static void net_dns_static_cb(const char *name, ip_addr_t *ipaddr, void *callbac
   if (ipaddr != NULL)
     addr = *ipaddr;
   else addr.addr = 0xFFFFFFFF;
-  int cb_ref = ((int*)callback_arg)[0];
-  free(callback_arg);
```
HHHartmann, line 862: This is potentially dangerous as int and pointer types do not have to have the same length.
Ah seen the assert below

app/modules/net.c
```diff
@@ -883,25 +888,24 @@ static int net_dns_static( lua_State* L ) {
...
-  if (cbref == LUA_NOREF) {
-    return luaL_error(L, "wrong callback");
-  }
-  int *cbref_ptr = calloc(1, sizeof(int));
-  cbref_ptr[0] = cbref;
   ip_addr_t addr;
-  err_t err = dns_gethostbyname(domain, &addr, net_dns_static_cb, cbref_ptr);
+
+  _Static_assert(sizeof(void *) >= sizeof(int), "Can't upcast int to ptr");
+
```
HHHartmann, line 898: Is it possible to use siomething like sizof(typeof(cbref)) to be save if cbref changes its type (to long) But maybe that's a bit too unlikely.

nwf: Sure.

nwf:
Sorry about that; pushed the wrong set of commits.  D'oh.  Anyway, here's the correct set with @HHHartmann's suggestion in place.  I'd like to merge these this weekend, so if I don't hear any objections by this evening I'll click my own button and will deal with any fallout. :)

PR #3067, opened 2020-04-10
mcp23017 Lua module

### LINKED ISSUE #3057: Add MCP23017 port expander support as a lua module
### Missing feature
Hello,
I would like to contribute a few things to this great project.

I saw that there are already issues for the MCP23017 Port Expander. Unfortunately I didn't find any Lua or C modules in the current dev branch.

Without looking into the issues, I just created a module in Lua for the MCP23017.
Now I would like to learn more and adapt the module to become part of this project.

### Justification
This module can help other IoT developers to easily provide more GPIO pins, for example to control a relays card. 

### Workarounds
I am not sure which workaround represents a good solution.

Kind regards,
Marcel

plomi-net:
Fixes #3057

Add MCP23017 port expander support as a lua module. See issue #3057


docs/lua-modules/mcp23017.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Example Script
+The example script can be found [here](../../lua_examples/mcp23017/mcp23017_example.lua)
+
+## setup()
+Configures the interface and sets the i2c pins and the address of the device.
+Automatically resets the device state (see `mcp23017.reset()`)
+
+#### Syntax
+`mcp23017.setup(address, customSCL, customSDA)`
```
nwf, line 37: I am uncomfortable with I2C device drivers that want to set up the bus, since the bus is shared by potentially multiple devices.  Perhaps take the `i2c` bus id rather than `SCL` and `SDA` and put the `i2c.setup()` call in the example.  Taking the bus id is a good idea because we might support multiple I2C busses, too.

lua_examples/mcp23017/mcp23017_example.lua
```diff
@@ -0,0 +1,112 @@
...
+    local a = mcp23017.readGPIOA()
+    print(" ")
+    print("GPIO A input states: " .. a)
+
+    local b = mcp23017.readGPIOB()
+    print("GPIO B input states: " .. b)
+    print(" ")
+
+end) then
+    print("Timer can't be created")
```
nwf, line 111: This one, too.

lua_examples/mcp23017/mcp23017_example.lua
```diff
@@ -0,0 +1,112 @@
...
+        -- print("set to low")
+        mcp23017.setPin(currentPin, mcp23017.LOW)
+        currentState = false
+    else
+        -- print("set to high")
+        mcp23017.setPin(currentPin, mcp23017.HIGH)
+        currentState = true
+    end
+end) then
+    print("Timer can't be created")
```
nwf, line 87: While, in principle, `tmr.alarm()` can return `false`, it will not do so on a freshly created timer, so this check is unnecessary.  (But good on you for erring on the side of caution!)

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,300 @@
...
+    i2c.start(MCP23017_id)
+    i2c.address(MCP23017_id, MCP23017_ADDRESS, i2c.RECEIVER)
+    local data = i2c.read(MCP23017_id, 1)
+    i2c.stop(MCP23017_id)
+    return string.byte(data)
+end
+
+-- get IO dir register
+local function getDirRegisterAddr(pin)
+    if pin > 7 then
```
nwf, line 117: This function doesn't check `> 15` unlike `convertPin` below; is that OK?

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,300 @@
...
+local function convertPin(pin)
+    if pin > 15 or pin < 0 then
+        print("The pin must be between 0 and 15")
+        return nil
+    end
+    return pin % 8
+end
+
+-- setup internal pullup
+local function setInternalPullUp(iByte)
```
nwf, line 143: It seems odd to me to couple the pull-up state across the two registers, especially as the device supports not.

I'd suggest `setInternalPullUp(regix, byte)` where `regix` is either `0` or `1` (or perhaps `"a"` or `"b"`).  Perhaps the rest of the API should be similar, referring to pins as "register index, pin index" pairs and not just "pin number"?  It would cut down on the need for `...A()` and `...B()` functions.

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,300 @@
...
+local MCP23017_INTCAPA = 0x10
+local MCP23017_INTCAPB = 0x11
+local MCP23017_OLATA = 0x14
+local MCP23017_OLATB = 0x15
+]]
+
+-- internal
+local MCP23017_id = 0
+local MCP23017_SPEED = i2c.SLOW
+local MCP23017_DEVICE_OK = false
```
nwf, line 69: Making these, and `MCP23017_ADDRESS` above, global like this means that, despite appearances (`_G[moduleName or 'mcp23017'] = M`), this module communicates with only one device and attempting to initialize a second can break the first.

Please switch to a "factory" style module, where the result is a function that yields a Lua table with methods, possibly via a metatable.

lua_examples/mcp23017/mcp23017_example.lua
plomi-net: I agree. The documentary was confusing me. See: https://nodemcu.readthedocs.io/en/master/modules/tmr/#example_5

lua_modules/mcp23017/mcp23017.lua
plomi-net: Many thanks for the advice. This is no longer necessary due to the changes to an a/b parameter.

lua_modules/mcp23017/mcp23017.lua
plomi-net: Thanks for the advice. I have now implemented this for the whole module. This reduces the size of the program code.

docs/lua-modules/mcp23017.md
plomi-net: Thanks for your comment and i agree.
I have now solved it to create an i2c setup before and submit the i2c id as parameter. 
Alternatively you can let the module do a setup for you.

lua_modules/mcp23017/mcp23017.lua
plomi-net: @nwf 
Creating the module in factory style is a very good idea.
Unfortunately I have a little bit of trouble to implement this in Lua. 

Do you have a good example where the objects are not overwritten?
Thanks a lot

nwf: Apologies for letting this fall off my radar.  An example of a "factory" module is the new LiquidCrystal module: https://github.com/nodemcu/nodemcu-firmware/blob/master/lua_modules/liquidcrystal/liquidcrystal.lua

plomi-net: @nwf thanks for your reply. that happens. Good things need a while.
i've updated the module to factory style. Your practice tip helped me a lot.
You are welcome to check if it is proper now.
many thanks.

HHHartmann:
If you have trouble rebasing your PR maybe have a look at #3174

plomi-net:
rebase doesn't working correctly in my repo. Sorry. 
I've created a new PR, #3197 
The remarks have been implemented.

Issue #2963, opened 2019-11-15
Uploading LFS.img with ESPlorer does not work anymore


HHHartmann:
### Expected behavior
 Uploading LFS image using ESPlorer should work.

### Actual behavior
The device sometimes reboots and ESPlorer displays
```
> Uploading to ESP file luac.out...Waiting answer from ESP - Timeout reached. Command aborted.
```

### Test code

This seems to be a problem for binary files only, as a 6K binary file cannot be uploaded but a 60K textfile can.

The code used by ESPlorer to receive the file looks like this:
``` Lua
_up=function(n,l,ll) local cs = 0 local i = 0 print(">".." ") uart.on("data", l, function(b) i = i + 1 file.open("Large File.txt",'a+') file.write(b) file.close() cs=0 for j=1, l do cs = cs + (b:byte(j)*20)%19 end
uart.write(0,"~~~CRC-".."START~~~"..cs.."~~~CRC-".."END~~~") if i == n then uart.on("data") end if i == n-1 and ll>0 then _up(1,ll,ll) end end,0) end file.remove("Large File.txt")

_up(270,250,145)
```
It then sends `n` blocks with the length of `l` and one last block with length `ll`.

### NodeMCU version
NodeMCU 3.0.0.0 
	branch: dev
	commit: f3044e1aff432ca48f6511554b317e0c80207d1c
	release: 2.0.0-master_20170202 +451
	release DTS: 
	SSL: false
	build type: float
	LFS: 0x0
	modules: adc,bit,dht,file,gpio,i2c,mqtt,net,node,ow,spi,tmr,uart,wifi
 build 2019-11-13 07:09 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)

### Hardware
Generic Dev board ESP8266SXD

HHHartmann:
More testing revealed that uploading of nul (0x00) bytes does not work properly.
All other char codes seem to work.

TerryE:
Gregor, thanks for this analysis.  I'll take a deep dive into this tomorrow and get back to you.  I think that this is an SDK 3.0 issue or at least the unification I've done so that the libraries can access the  Lua 5.1 and 5.3 VMs with a common API.  This big bitch with the pre SDK 3.0 runtime was that it was just far too easy to overrun the input UART, so with SDK 3.0, I've introduced the stdin and stdout pipes for buffering.  This seems to have broken something.

seregaxvm:
[...]

PR #3070, opened 2020-04-19
Correct integer types in BMP085 driver

andrmuel:

- [~] I have thoroughly tested my contribution.
⇒ I only tested it on master; current dev fails for me

### DIFF
--- a/app/modules/bmp085.c
+++ b/app/modules/bmp085.c
@@ -138,7 +138,8 @@ static int bmp085_lua_pressure_raw(lua_State* L) {
 static int bmp085_lua_pressure(lua_State* L) {
     uint8_t oss = 0;
     int32_t p;
-    int32_t X1, X2, X3, B3, B4, B5, B6, B7;
+    int32_t X1, X2, X3, B3, B5, B6;
+    uint32_t B4, B7;
 
     if (lua_isnumber(L, 1)) {
         oss = luaL_checkinteger(L, 1);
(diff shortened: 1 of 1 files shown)


nwf:
These seem sensible enough to me, but I do not have one of these devices to test.  Any of the active maintainers have one?

@andrmuel Can you elaborate on why `dev` isn't working for you?

andrmuel:
I was able to build the firmware on dev and flash it, but afterwards I was not able to flash any Lua code for testing.

In regards to testing with the BMP085: if you want to reproduce the original problem as well, that might be possible with heating it up to 30-40 deg C.

HHHartmann:
That's probably due to bug #2963. It is fixed in #2912.
It was IIRC introduced with #2836. So you could rebase your branch to a commit before that and retry.

andrmuel:
I'm not sure I understand correctly. The commit you are referring to is 14c1b8fd542beabba68646dab93f4862b364a4b3 (from Sep. 2019)? The commit before that would be c7ff86ff275c01497271b1f2542d6e750cc87ca6, and that is already part of master. It does work on the current master. What would be the benefit of testing it with an _older_ version of master (which contains less of dev)?

HHHartmann:
The changes in this PR are very local, so that I would not expect any of the changes in dev to influence the outcome of the test. 
But I think there was another upload to mentioned somewhere which still works on current dev. 
Another alternative would be to flash the entire SPIFFS partition containing all the files you need.

nwf:
Given the highly localized nature of this change, as @HHHartmann says, I've just merged it to `dev`.

PR #3072, opened 2020-04-22
esp32: Unref gpio.trig callbacks when type=INTR_DISABLE (#2880)

### LINKED ISSUE #2880: gpio.trig does not remove callback (dev-esp32)
### Expected behavior
Documentation states "trigger is disabled when [callback] omitted"
### Actual behavior
Interrupt callback is still there. The way I worked around it was to have a dummy function (il1) to switch to but it is not omitted.
### Test code
```Lua
gpio.config({gpio={22},dir=gpio.IN, pull=gpio.PULL_UP})
function il1() end
function il0() print('low') gpio.trig(22,gpio.INTR_LOW) end
gpio.trig(22,gpio.INTR_LOW,il0)
```

### NodeMCU version
dev-esp32
powered by Lua 5.1.4 on ESP-IDF v3.3-beta1-791-g890a341 on ESP-IDF v3.3-beta1-791-g890a341 on SDK IDF

### Hardware
ESP-32

tomsci:
Fixes #2880

- [x] This PR is for the `dev` branch rather than for `master`. (_actually dev-esp32_)
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well

As per my comment on the issue, this change documents how to use `gpio.INTR_DISABLE` to disable an interrupt trigger, and also ensures that the callback is unrefd (and `gpio_isr_handler_remove` is called) when that happens.

It also brings the API more in line with the esp8266 version by making `type` be optional and corrects the description of what happens when `callback` is nil. The [esp8266 docs](https://nodemcu.readthedocs.io/en/dev/modules/gpio/#gpiotrig) say:

> If the type is "none" _or omitted_ then the callback function is removed and the interrupt is disabled
> ...
> The previous callback function will be used if the function is omitted

Testing:
* Checked that calling`gpio.trig(N, gpio.INTR_DISABLE)` unrefs any existing callback (checked the output of `for k,v in pairs(debug.getregistry()) do print(k,v) end` and verified there was one fewer keys listed)
* Checked that `gpio.trig(N)` is equivalent to `gpio.trig(N, gpio.INTR_DISABLE)`
* Checked a disabled interrupt can be reenabled with another call to `gpio.trig(N, TYPE, callback)`
* Checked that changing the interrupt type with a nil callback eg `gpio.trig(N, gpio.INTR_DOWN)` preserves and keeps calling the same callback.
* Checked that calling `gpio.trig(N, gpio.INTR_DISABLE, callback)` with a non-nil callback does not cause a ref to `callback` and will not cause `callback` to be called if a subsequent call to trig with a non-zero type and a nil callback is made.

### DIFF
--- a/components/modules/gpio.c
+++ b/components/modules/gpio.c
@@ -139,7 +139,7 @@ static int lgpio_read (lua_State *L)
 static int lgpio_trig (lua_State *L)
 {
   int gpio = luaL_checkint (L, 1);
-  int intr_type = luaL_checkint (L, 2);
+  int intr_type = luaL_optint (L, 2, GPIO_INTR_DISABLE);
   if (!lua_isnoneornil (L, 3))
     luaL_checkanyfunction (L, 3);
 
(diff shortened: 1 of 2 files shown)


marcelstoer:
I'm not a firmware developer (well, not a real one at least) but this looks sane to me. Can I ask @chilipeppr or @jmattsson to have a look?

jmattsson:
I'm not in a position to verify on hardware at the moment, but on the surface this is looking good!

PR #3076, opened 2020-04-25
fix #2565 by inverting the grb2hsv argcheck condition

### LINKED ISSUE #2565: Incorrect guard in color_utils.c grb2hsv()
The argument guard for the grb2hsv() function is the opposite of what it should be. As written it rejects any non-greyscale value, when the guard should be rejecting greyscale values.

### Expected behavior 

It returns the appropriate HSV when a non-greyscale value is entered

### Actual behavior

It throws an error when a non-greyscale value is entered

### Test code
```Lua
color_utils.grb2hsv(200,200,0) -- should work (doesn't)
color_utils.grb2hsv(200,200,200) -- shouldn't work (does)
```
### NodeMCU version
There is only one revision of this file

### Hardware
No special hardware needed

Firenox89:
Fixes #2565.

The arg check condition for `color_utils.grb2hsv()`, which was intended to protect against greyscale values, did opposite by only allowing greyscale values.

After inverting check condition, I was able to use the function.

### DIFF
--- a/app/modules/color_utils.c
+++ b/app/modules/color_utils.c
@@ -217,7 +217,7 @@ static int cu_grb2hsv(lua_State *L) {
   const int r = luaL_checkint(L, 2);
   const int b = luaL_checkint(L, 3);
 
-  luaL_argcheck(L, g == r && g == b, 1, "greyscale value cannot be converted to hsv");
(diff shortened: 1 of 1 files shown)


marcelstoer:
@HHHartmann I took the liberty of adding you as a reviewer as you the had the issue self-assigned back then.

PR #3086, opened 2020-05-01
esp32: Correct level value in gpio.trig() callback (#2883)

### LINKED ISSUE #2883: GPIO.trig callback level not properly masked on dev-esp32
Make sure you read and understand http://nodemcu.readthedocs.io/en/dev/en/support/.
Use one of the two templates below and **DELETE THE REST**.

### Expected behavior
interrupt callback parameter level is 0 or 256 instead of 0 or 1
### Actual behavior
high level is 256
### Test code
```Lua
function intr(pin,level)
  print(pin,level)
  gpio.trig(21,gpio.INTR_LOW,intr)
end
gpio.config({gpio={21},dir=gpio.IN, pull=gpio.PULL_UP})
gpio.trig(21,gpio.INTR_HIGH,intr)

-->21	0
-->21	256
```
### NodeMCU version
powered by Lua 5.1.4 on ESP-IDF v3.3-beta1-791-g890a341 on ESP-IDF v3.3-beta1-791-g890a341 on SDK IDF

### Hardware
ESP-WROOM-32

tomsci:
Fixes #2883.

- [x] This PR is for the ~dev~ `dev-esp32` branch rather than for `master`.

This PR changes the `level` parameter passed to the `gpio.trig()` callback so that it matches the documentation, ie `0` or `1` rather than the current behaviour of `0` or `256`. **This is a change in behaviour** which potentially could break existing code, but it's the Right Thing from the point of view of matching the documentation and matching peoples' expectations having not used the API before. I'm not sure how such changes are normally handled, I can update the docs to note the change in behaviour if desired? As it stands I haven't changed the docs since the change is to bring the implementation in line with what the docs say should be happening.

Testing done:

```
> gpio.config({ gpio=26, dir=gpio.IN, pull=gpio.PULL_DOWN })
> gpio.trig(26, gpio.INTR_UP_DOWN, print)
-- Raised and lowered pin 26 a couple of times
> 26    1
26      0
26      1
26      0
```
(previously those `1`s were `256`)

### DIFF
--- a/components/modules/gpio.c
+++ b/components/modules/gpio.c
@@ -216,7 +216,7 @@ static void nodemcu_gpio_callback_task (task_param_t param, task_prio_t prio)
 {
   (void)prio;
   uint32_t gpio = (uint32_t)param & 0xffu;
-  int level = ((int)param) & 0x100u;
+  int level = (((int)param) & 0x100u) >> 8;
 
   lua_State *L = lua_getstate ();
   if (gpio_cb_refs[gpio] != LUA_NOREF)


marcelstoer:
> it's the Right Thing from the point of view of matching the documentation and matching peoples' expectations having not used the API before

👍

PR #3094, opened 2020-05-08
IR module (Infrared)

ecrips:

This adds a new module (with documentation) which can receive IR pulses from a suitably connected sensor (such as a TSOP382xx). It also can decode RC-5 codes.


pjsg:
Can you please include a link to the relevant hardware? I.e. what do I
search for if I want to buy one of these?

ecrips:
Sorry, searching for TSOP38238 is probably best. The code should work with any of the sensors in that family, but the TSOP38238 is the common one. I've also noticed that it might be worth spelling out that IR=InfraRed - that's mentioned in the documentation but might not be obvious otherwise - I've updated the pull request title.

pjsg:
I.e some like this ought to work:
https://www.aliexpress.com/item/32975610831.html?

I suspect that you ought to note that you need to pick an IR receiver
module that matches the PCM frequency of the transmitter....

You could also include a link to 'Standard chinese sources:
https://www.aliexpress.com/wholesale?SearchText=ir+receiver+module
<https://www.aliexpress.com/wholesale?SearchText=ir++receiver+module>'

ecrips:
> I.e some like this ought to work: https://www.aliexpress.com/item/32975610831.html?

The IR receiver will probably work, but not that the remote uses the NEC protocol which I haven't written a decoder for (RC-5 is more common here).

> I suspect that you ought to note that you need to pick an IR receiver module that matches the PCM frequency of the transmitter.... You could also include a link to 'Standard chinese sources: https://www.aliexpress.com/wholesale?SearchText=ir+receiver+module <https://www.aliexpress.com/wholesale?SearchText=ir++receiver+module>'

Yes, matching the frequency of the transmitter ensures a better signal. The TSOP38238 is available on the "standard Chinese sources": https://www.aliexpress.com/wholesale?SearchText=TSOP38238 and is what I've tested with. Other IR receiver modules are likely to work (almost all provide the same sort of signal), but I haven't tested so can't be sure.

app/modules/ir.c
```diff
@@ -0,0 +1,248 @@
...
+  int cb[2];
+  uint32_t last_time;
+  uint16_t rawbuf[RAWBUF_SIZE];
+  uint16_t rc5_data;
+  uint8_t rc5_state;
+  uint8_t rawbuf_read;
+  uint8_t rawbuf_write;
+  uint8_t pin, pin_num;
+  uint8_t setup;
+} data;
```
nwf, line 32: I know I'm going to sound like a pest, but: is it possible to avoid globals like this and encapsulate everything needed within a Lua userdata?  Could we even potentially allow for multiple IR rxers at once?

tomsci:
Turns out I've been working on [something similar](https://github.com/tomsci/nodemcu-firmware/blob/tomsci_ir_rebase/components/modules/irpwm.c), although for the esp32 so the code is conceptually similar but all the platform APIs are different. Do you think it's worth us trying to align the two, at least to the point of having the same Lua APIs in each? My attempt was more focussed on sending rather than receiving, so there's a chunk of code using the `rmt` logic block on the esp32 which I don't think the esp8266 has? Although using it for receiving I found to be more pain than it was worth, and I went with a simple GPIO ISR callback, like you.

> uses the NEC protocol which I haven't written a decoder for (RC-5 is more common here).

Where I am it's the opposite, most things I have seem to use NEC or a variant thereof :) But there's sufficient variation I'd say it's not worth trying to include an NEC decoder in the C code. I ended up doing all the decoding Lua side to account for all the weirdnesses.

fwiw I haven't found an IR device yet that the TSOP38238 couldn't decode, the biggest pain for me has been working out what the high-level protocol is.

app/modules/ir.c
ecrips: Some global data is needed for the interrupt routine (so it can find the structure with the data in). But it would (in theory) be possible to move most of this into dynamically allocated data which would allow multiple IR receivers. I have to admit removing the need for the static `rawbuf` array would be a nice improvement (reduces RAM usage when not in use). I'll have a look at how difficult it would be to do.

galjonsfigur: @ecrips Maybe it isn't the best suggestion, but you can have a look at `softuart` module where I was faced with similar problem. My solution is not ideal (as it cannot read more than one data frame simultaneously) but the method on implementing multiple receiving instances and selecting right one in ISR should be possible to apply in this code.

ecrips:
> Turns out I've been working on [something similar](https://github.com/tomsci/nodemcu-firmware/blob/tomsci_ir_rebase/components/modules/irpwm.c), although for the esp32 so the code is conceptually similar but all the platform APIs are different. Do you think it's worth us trying to align the two, at least to the point of having the same Lua APIs in each? My attempt was more focussed on sending rather than receiving, so there's a chunk of code using the `rmt` logic block on the esp32 which I don't think the esp8266 has? Although using it for receiving I found to be more pain than it was worth, and I went with a simple GPIO ISR callback, like you.

It would make sense trying to align the interface if possible. The main difference I can see is you are using negative/positive values to encode high/low pulses whereas I used the bottom bit. Your approach is probably easier to handle in Lua.

> 
> > uses the NEC protocol which I haven't written a decoder for (RC-5 is more common here).
> 
> Where I am it's the opposite, most things I have seem to use NEC or a variant thereof :) But there's sufficient variation I'd say it's not worth trying to include an NEC decoder in the C code. I ended up doing all the decoding Lua side to account for all the weirdnesses.

I initially tried doing all the IR decoding in Lua (i.e. used `gpio.trig()`) but couldn't get accurate timing, so moved all the code over to C. But having the Lua interface for decoding on Lua was actually very useful for debugging the C so I kept it knowing that it could be used for decoding other protocols in the future. NEC does seem to have more room for 'quirks' than RC-5.

> fwiw I haven't found an IR device yet that the TSOP38238 couldn't decode, the biggest pain for me has been working out what the high-level protocol is.

Yes, the only issue I've had is sensitivity to noise. E.g. some pick up quite a bit of interference from CFL bulbs.

marcelstoer:
Pending:
- rebase on dev
- @tomsci and @ecrips to agree on an API to align ESP32 and ESP8266

slorquet:
Hello

are there any plans to support IR signal transmission? for example to control a device, like an air conditioner... :p ?

thanks for any info

tomsci:
> are there any plans to support IR signal transmission

I ended up splitting my ir code into a separate external module here: https://github.com/tomsci/nodemcu-irpwm to avoid the problem that one-size-never-fits-all - my implementation was probably too specific to my requirements to suit anyone else out-of-the-box, the whole esp32 vs esp8266 API divergence doesn't seem to have gotten any better, we're all doing this in our spare time, etc.

That being said, you're welcome to look at my code for `irpwm_send()` in the above repo. However that only gets you the basics of sending IR pulses. To actually control a device you need to know its protocol and codes. I only have codes for a few specific devices I needed to support (and which I don't have any plans to publish), so I can't help you there. Not sure if ecrips's code might give you more help there or not. You may need to consult something like https://www.lirc.org/. There seem to be some promising links if you google "lirc nodemcu".

Issue #3098, opened 2020-05-10
esp32: Expose gpio_set_drive_capability API


tomsci:
I propose exposing the `gpio_set_drive_capability(gpio, strength)` API to Lua as `gpio.set_drive(gpio, strength)`. (I will submit a PR with the code shortly).

I don't think it should be an option to `gpio.config()` because the C API keeps it separate, and additionally I found myself needing to use it on a pin which I wasn't calling gpio.config on, in this case using the `rmt_` APIs but the same thing would probably apply for SPI etc.

Obviously, the reason for the proposal is because I needed to configure a pin's drive strength to something other than the default. I haven't bothered exposing the read fn `gpio_get_drive_capability()` because I couldn't think of a situation where I'd really need that from Lua - generally, it's something you just set because you know what it needs to be, it's not really a dynamic thing. But I'll add it if anyone feels strongly about it.

PR #3099, opened 2020-05-10
esp32: Expose gpio_set_drive_capability API to Lua

tomsci:
Addresses #3098.

- [x] This PR is for the ~`dev`~ `dev-esp32` branch rather than for `master`.

As discussed on the issue, this PR exposes the API to set drive strength on a GPIO pad so its maximum current output can be increased (or decreased).

### DIFF
--- a/components/modules/gpio.c
+++ b/components/modules/gpio.c
@@ -211,6 +211,17 @@ static int lgpio_write (lua_State *L)
 }
 
 
+// Lua: gpio.set_drive(gpio, gpio.DRIVE_x)
+static int lgpio_set_drive (lua_State *L)
+{
+  int gpio = luaL_checkint (L, 1);
+  int strength = luaL_checkint (L, 2);
+  luaL_argcheck (L, strength >= GPIO_DRIVE_CAP_0 && strength < GPIO_DRIVE_CAP_MAX,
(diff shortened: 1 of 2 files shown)


jpeletier:
Hi @TOMSCI. Nice one.

My only comment would be to put this as part of `gpio.setup()`, rather than having a separate function.

It could be a table parameter such as `strength` or `drive_strength`. If not specified, then it would not be configured and left as-is, so the call is backward compatible.

tomsci:
@jpeletier I wondered about adding it to `gpio.config()`, but decided against it, because `gpio.config()` maps pretty directly to the underlying Espressif SDK function `gpio_config()` so adding things to it which weren't in the underlying C API didn't feel right. In addition it's possible that you'd want to set the drive strength for a pin which you're not actually using as a GPIO, for example a SPI pin or one configured for use by the PWM or RMT hardware, so making it a parameter to `gpio.config()` would not work for that case (because you don't call `gpio.config()` in those situations). Of course having it in the `gpio` at all is a bit weird when you can use it with non-GPIO-configured pins, but that's what the C API does as well, so I guess there isn't a better place for it.

Basically, it's a separate API in the underlying C, and I think it's probably better to keep it separate in the Lua API too.

PR #3101, opened 2020-05-11
ESP8266: Simplify platform.c task handling

jpeletier:

This PR simplifies `platform.c` handling of tasks to avoid having to keep an array of callbacks. Interface is left intact and code is cleaner.


nwf:
I am not competent to review the ESP32 version of this (#3087) and only barely competent to review this ESP8266 one...

If I read this right, we were needlessly maintaining an array of objects that the underlying "not an RTOS we swear" was willing to queue for us anyway?

jpeletier:
This is an attempt to simplify the code whereupon instead of storing the callback function pointer in an array and handing over the index of that array, the handle is the pointer, thus making the code cleaner.

This has the advantage of simpler and easier to understand code and a tiny little bit of performance, but has the drawback of making things harder to debug in case somebody `task_post()` posts with a corrupt handle/pointer. Discussion is taking place in #3087

TerryE:
I am not sure what is going on here but at the moment committing this looks like it looks it would change 546 files not the few need to implement this change, and thus back out lots of other essential changes.  This would be disastrous.  We need to sort this out after we've completed this current round of core update, and possibly rebaseline and force push this again. I will add the blocked label in the meantime.

marcelstoer:
> at the moment committing this looks like it looks it would change 546 files not the few need to implement this change

The changes simply need to be rebased on the current dev.

PR #3102, opened 2020-05-11
Extend max year from 2038 to 2106 in epoch2cal()

AntonGrushin:

__Description:__
Using unsigned integer instead of signed increases amount of seconds that we can use as timestamp in rtctime.epoch2cal(stamp) function by two times.
Current max value is 2147483647 which is 19 January 2038. New value is 4294967295 which is 7 February 2106.

### DIFF
--- a/app/modules/rtctime.c
+++ b/app/modules/rtctime.c
@@ -91,10 +91,10 @@ void rtctime_deep_sleep_until_aligned_us (uint32_t align_us, uint32_t min_us)
   rtc_time_deep_sleep_until_aligned (align_us, min_us);
 }
 
-void rtctime_gmtime (const int32 stamp, struct rtc_tm *r)
+void rtctime_gmtime (const uint32 stamp, struct rtc_tm *r)
 {
   int32_t i;
-  int32_t work = stamp % (SPD);
+  uint32_t work = stamp % (SPD);
   r->tm_sec = work % 60; work /= 60;
   r->tm_min = work % 60; r->tm_hour = work / 60;
   work = stamp / (SPD);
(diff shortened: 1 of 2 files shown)


AntonGrushin:
I’m not asking for 64 bit, in this particular `rtctime.epoch2cal(stamp)` function use of signed 32 bit variable is a bit senseless, it should be unsigned. Main structure `rtc_timeval` that is used in rtctime module as a main timespamp counter is using unsigned 32 integer [too](https://github.com/nodemcu/nodemcu-firmware/blob/3d917850180f67adc7f2c6b5d00f27c152e7194c/app/include/rtc/rtctime.h#L45). I’m just asking to use same type in `rtctime.epoch2cal(stamp)`. Otherwise, we waste half of possible values.

pjsg:
OK -- but please update the documentation to say that this is *not* the same as the unix timestamp which can deal with negative values. Also, think about what to do with floats....

marcelstoer:
> @AntonGrushin Using unsigned integer instead of signed increases
> @pjsg say that this is not the same as the unix timestamp

This goes against [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), doesn't it.

However, one might argue that the 2038 limit is totally unexpected as well. Still, using unsigned integer instead of signed ones effectively just moves the window (i.e. the supported range).

pjsg:
I think that we need to think carefully about what we want to do with 2038
-- which, I would argue, is better known that the 21xx date. Maybe the
right thing is to move to 64 bit now and that is then the end of the
problem.

marcelstoer:
@pjsg I agree. I too do not support this PR.

AntonGrushin:
> @pjsg say that this is not the same as the unix timestamp 
Sure, added description.
> @marcelstoer Still, using unsigned integer instead of signed ones effectively just moves the window (i.e. the supported range).
This is true. I was just following logic that devices that use current firmware will survive till 2038 (that is only in 18 years from now), while on year 2106 they won’t be used most likely (in 86 years).
Better way is 64 bit for sure. However, this may take a while.

marcelstoer:
> Maybe the right thing is to move to 64 bit now

👍 @pjsg ok for you if we close this?

pjsg:
Yes!

Issue #3078, opened 2020-04-28
Error Handling in Lua 


TerryE:
#3075 introduced the new `luaL_callx()` to facilitate improved [Panic Handling](https://gist.github.com/TerryE/646407100785fb00c2cbddbb0e8a2b61#panic-handling).  I want to use this issue to discuss error handling in more general terms and why the focus on panic handling. 
-  **Caught Errors**.  The Lua architecture allow the application so establish an error handler via the `pcall()` and `xpcall()`, thus creating a _protected environment_ for executing a code hierachy.  In such cases the error is returned to the application and the application code itself determines how to process or log the error.
-  **Uncaught Errors**.  Any Lua code executed outside such a protected environment results in a Lua panic, which in standard Lua coding typically terminates the application.  In the case of NodeMCU, this triggers a processor restart.  

One of the main issues that catches new developers is that whilst the interactive thread established a default protected environment to catch and print errors interactively, any Lua callback runs as separate execution thread and therefore is not protected; any errors here will panic and reboot the processor.  The issue that I want to consider here is whether this is the correct behaviour for NodeMCU and if not then how we can improve this in a way which doesn't break existing applications.

We have three broad mechanisms for reporting errors in the firmware:

1.  **Print direct to the UART**.  We adopt this approach during Lua startup because this is the simplest and most robust mechanism that avoids dependencies on services that might not be yet running.
2.  **Print the error using the standard `print()` function**. Prior to SDK 3.0 releases, we always did (1) but this meant that errors generated during `node.output()` where reported only on the UART.  `print()` now sends output to the `STDOUT` which is in turn emptied to the UART or the `node.output()` reader using the [Pipe](https://nodemcu.readthedocs.io/en/dev/modules/pipe/) module, and this enables error reporting in a terminal session to occur through the session.
3. **Use a separate panic process manger** to process uncaught errors.  The problem with (2) during panics is that a simple "print error and restart" will queue the error message to STDOUT then restart the processor before the printing task has executed.  At a minimum, the restart operation should wait until the STDOUT pipe is empty (or a maximum time elapsed) before restarting the ESP.

In case (3) the default panic process should still print the error, but at least wait until the STDOUT pipe is empty or a maximum time elapsed before restarting the ESP.  We should also provide a `node.setonerror()` call to enable the application to override this default (say to add some form of error logging).

### Scope of Work

So I suggest that a good next PR would be to tidy up such error handling:
[...]

Issue #1168, opened 2016-03-17
Consistent indent style


marcelstoer:
- agree on an indent style
- document it in `CONTRIBUTING.md`
- run the code formatter (which one?) once over the entire code base?

I suggest to also persist the style in an http://editorconfig.org/ file. That's likely the most sensible approach for a group without a common editor/IDE.

TerryE:
I tend to K&R but following the Lua convention of two spaces for a logical indent.  Tabs are just too much of a bugger on different visualisation systems to be worth the saving for a 2 tab indent.  I not that the K&R convention of not indenting a single statement is quite common, and prefer 1TBS as this I've been burnt by this sort of error far too often:
```C
  if (x < 0)
    puts("Negative");
    negative(x);
  nextStatement();  if (x < 0) {
        puts("Negative");
        negative(x);
    }
```
easy to miss, eh?  But as to style, the vast bulk of of our source is the Lua core and this is written to a very consistent style, so my view is: _if we are going to encourage a unified style (which I believe that we should), then we should follow this._

As I have already started in #1165 and #1166, I think that the code base would be more maintainable if we did this sort of clean-up. I would suggest that we don't reformat imported libraries such as lwip et al, but as to the `app/modules`  and any future additions, then yes definitely you have my :+1:.

jmattsson:
Honestly, as long as the source is easily readable, I'm not fussed about 2 vs 4 vs TAB. We have more important things to work on given our limited resources, imo.

I do quite strongly prefer leaving braces off for single-statement `if` conditionals as it yields cleaner code. Of course, I tend to have a blank line after such a construct...

TerryE:
My personal  preference for one liners is to keep them on the same line as the if, while or whatever

     if (!p) return MEM_ERR;

pjsg:
The great thing about coding styles is that there are so many strongly held opinions. We went through this at work some time ago (for Java). In the end we compromised and nobody was 100% happy. We decided that the only way to not show favouritism was to adopt someone else's style guide *without* *any* changes. 

I'm not proposing that we do this here.... Well, we could adopt a style, and then not enforce it! 

My take on the examples above are 4 space indent, never use tabs, always use braces. e.g.

```
if (!p) {
    return MEM_ERR;
}
```

There was a famous security bug which was the result of missing braces after an if. The code read: (although the warning comment was not present!)

```
if ((err = SSLHashSHA1.update(&hashCtx, &serverRandom)) != 0)
[...]

PR #3111, opened 2020-05-16
Implementing optional parameter `restart` to tmr:start()

### LINKED ISSUE #3020: tmr:start() does not restart started timer
### Expected behavior
https://nodemcu.readthedocs.io/en/master/modules/tmr/#tobjstart
says that `tmr:start()` **restarts** timer. It's not true:

### Actual behavior
`tmr:start()` will **restart** timer (start it count from zero)

### Test code
```Lua
> t = tmr.create()
> t:register( 3000, tmr.ALARM_SEMI, function( t ) print( 'alarm' ) end)
> t:start()
> -- wait 2 seconds
> t:start()
> -- alarm message will be printed 3 seconds after the first t:start() call, not the second one as expected.
```

### Workaround
```Lua
t:stop()
t:start()
```

### NodeMCU startup banner
NodeMCU 3.0.1.0, build type: integer
modules: file,gpio,net,node,rtctime,sntp,tmr,uart,wifi
build 2020-01-27 17:39 powered by Lua 5.1.4 on SDK 3.0.2(824dc80)

### Hardware
ESP8266 wemos D1 (not mini)

vsky279:
Fixes #3020.

Implements the solution discussed in #3020 as wrapped up by @HHHartmann 

Also removes unecessary check of lua_newuserdata result (also in ads1115 module).

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -125,17 +125,24 @@ static int tmr_register(lua_State* L) {
 	return 0;
 }
 
-// Lua: t:start()
+// Lua: t:start( [restart] )
 static int tmr_start(lua_State* L){
 	tmr_t *tmr = (tmr_t *) luaL_checkudata(L, 1, "tmr.timer");
-  int idle = tmr->mode & TIMER_IDLE_FLAG;
+	lua_settop(L, 2);
+	luaL_argcheck(L, lua_isboolean(L, 2) || lua_isnil(L, 2), 2, "boolean expected");
+	int restart = lua_isboolean(L, 2) ? lua_toboolean(L, 2) : 0;
 
-  lua_settop(L, 1);  /* ignore any args after the userdata */
+	lua_settop(L, 1);  /* ignore any args after the userdata */
 	if (tmr->self_ref == LUA_NOREF)
 		tmr->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 
-	if(idle) {
-  	tmr->mode &= ~TIMER_IDLE_FLAG;
+	//we return false if the timer is not idle
+	int idle = tmr->mode&TIMER_IDLE_FLAG;
+	if(!(idle || restart)){
+		lua_pushboolean(L, 0);
+	}else{
+		if (!idle) {os_timer_disarm(&tmr->os);}
+		tmr->mode &= ~TIMER_IDLE_FLAG;
 		os_timer_arm(&tmr->os, tmr->interval, tmr->mode==TIMER_MODE_AUTO);
   }
 	lua_pushboolean(L, !idle); /* false if the timer is not idle */
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -56,35 +56,35 @@ static sint32_t soft_watchdog  = -1;
 static os_timer_t rtc_timer;
 
 static void alarm_timer_common(void* arg){
-  tmr_t *tmr = (tmr_t *) arg;
-  if(tmr->lua_ref > 0) {
-    lua_State* L = lua_getstate();
-    lua_rawgeti(L, LUA_REGISTRYINDEX, tmr->lua_ref);
-    lua_rawgeti(L, LUA_REGISTRYINDEX, tmr->self_ref);
-    if (tmr->mode != TIMER_MODE_AUTO) {
-    if(tmr->mode == TIMER_MODE_SINGLE) {
-      luaL_unref2(L, LUA_REGISTRYINDEX, tmr->lua_ref);
-      luaL_unref2(L, LUA_REGISTRYINDEX, tmr->self_ref);
-      tmr->mode = TIMER_MODE_OFF;
-    } else if (tmr->mode == TIMER_MODE_SEMI) {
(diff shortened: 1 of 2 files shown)


app/modules/ads1115.c
```diff
@@ -252,9 +252,6 @@ static int ads1115_lua_register(lua_State *L, uint8_t chip_id) {
         return luaL_error(L, "unexpected config value (%p) please reset device before calling this function", config_read);
     }
     ads_ctrl_ud_t *ads_ctrl = (ads_ctrl_ud_t *)lua_newuserdata(L, sizeof(ads_ctrl_ud_t));
-    if (NULL == ads_ctrl) {
-        return luaL_error(L, "ads1115 malloc: out of memory");
-    }
```
nwf, line 257: Unrelated change?  (Though I think one we should take, but not part of this PR.)

app/modules/tmr.c
```diff
@@ -184,16 +184,21 @@ static int tmr_register(lua_State* L){
 // Lua: tmr.start( id / ref )
```
nwf, line 184: `Lua: tmr.start(ref, [restart])`?  More generally, if you're wanting to tidy `tmr`, could you drop the `id` from all of the `Lua:` comments?

app/modules/tmr.c
```diff
@@ -363,7 +368,6 @@ static int tmr_softwd( lua_State* L ){
 // Lua: tmr.create()
 static int tmr_create( lua_State *L ) {
 	tmr_t ud = (tmr_t)lua_newuserdata(L, sizeof(timer_struct_t));
-	if (!ud) return luaL_error(L, "not enough memory");
```
nwf, line 366: Unrelated, but we should take it as it's part of `tmr`.

app/modules/ads1115.c
vsky279: Ok. Let's make it another PR.

app/modules/tmr.c
vsky279: Ok, let's do it.

TerryE:
Lukas, I tidied up ``tmr.c` (= quite a lot of rework) for my #3078.  If you don't mind, it's going to be a _lot_ easier to add in this couple of changes rather than having to do a three-way merge. Thanks.

vsky279:
Of course not. So you will include these few lines in your commit?

app/modules/tmr.c
```diff
@@ -125,17 +125,24 @@ static int tmr_register(lua_State* L) {
...
 }
 
-// Lua: t:start()
+// Lua: t:start( [restart] )
 static int tmr_start(lua_State* L){
 	tmr_t *tmr = (tmr_t *) luaL_checkudata(L, 1, "tmr.timer");
-  int idle = tmr->mode & TIMER_IDLE_FLAG;
+	lua_settop(L, 2);
+	luaL_argcheck(L, lua_isboolean(L, 2) || lua_isnil(L, 2), 2, "boolean expected");
+	int restart = lua_isboolean(L, 2) ? lua_toboolean(L, 2) : 0;
```
marcelstoer, line 133:
```suggestion
	int restart = lua_isboolean(L, 2);
```
You already verified it's a boolean in the line above.

app/modules/tmr.c
```diff
@@ -125,17 +125,24 @@ static int tmr_register(lua_State* L) {
...
-  	tmr->mode &= ~TIMER_IDLE_FLAG;
+	//we return false if the timer is not idle
+	int idle = tmr->mode&TIMER_IDLE_FLAG;
+	if(!(idle || restart)){
+		lua_pushboolean(L, 0);
+	}else{
+		if (!idle) {os_timer_disarm(&tmr->os);}
+		tmr->mode &= ~TIMER_IDLE_FLAG;
 		os_timer_arm(&tmr->os, tmr->interval, tmr->mode==TIMER_MODE_AUTO);
   }
```
marcelstoer, line 147:
```suggestion
	}
```

TerryE: This is an example where some of `tmr.c` uses `\t` and some uses `  ` (two spaces) as the logical indent.  But if we aren't going to fix the whole mix as Marcel suggests, then we should at least be consistent within each function.

app/modules/tmr.c
vsky279: But it can still be `nil` if default parameter is used. I have not checked that `lua_toboolean(L, 2)` returns 0 if parameter is `nil`. So 👍

app/modules/tmr.c
vsky279: There were more to be switched to tab. I prefer spaces but tab prevails. So I switched them to tab.

vsky279:
Oups. A bit of fight with @marcelstoer in commits. I hope the result is fine.

marcelstoer:
> I hope the result is fine.

As #1168 is now closed we maybe shouldn't argue about it but replacing spaces with tabs is a move in the wrong direction IMO.

I'm happy to merge this though.

vsky279:
spaces -> tabs - I only made it consistent with the rest of the file where tabs are used.
In the following PR we can replace tabs to spaces for good.

PR #3113, opened 2020-05-17
added init.lua abort script using a pin

kovianyo:
Hello,

i extended the init.lua section with an other abort script. There was no issue about this, I just thought to improve the documentation. Please use this as it fits.

### DIFF
--- a/docs/upload.md
+++ b/docs/upload.md
@@ -116,6 +116,32 @@ wifi.sta.config({ssid=SSID, pwd=PASSWORD})
 
 ```
 
+If you do not want the delay but you have a free pin, you can use that pin, pulling it to ground to abort startup. The following codes exactly that. You can put this code into `init.lua`. If the pin is not pulled to ground, it continues with executing an other lua file called `startup.lua`. Since the function is declared `local`, the memory used by this method will be freed after `init.lua` is executed.
+
+```lua
+local checkAbort = function()
+  local pin = 0 -- GPIO16, D0
+  local pinDescription = 'pin 0 (GPIO16, D0)'
+  gpio.mode(pin, gpio.INPUT)
+  print()
+  if (gpio.read(pin) == 0) then
+    print('Startup aborted (' .. pinDescription .. ' is low).')
+    return
+  else
+    local nextLuaFileName = 'startup.lua'
+    print('Starting up... (pull ' .. pinDescription .. ' to ground to abort startup)')
+    print()
+    if (file.exists(nextLuaFileName)) then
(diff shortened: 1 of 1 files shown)


marcelstoer:
Thanks for the suggested improvement. Could please code and describe in a way that we could say something like

> If you don't like the delay then...
> - replace the `startup()` function with the one below (also called startup)
> - replace `wifi_got_ip_event = function(T)` with the one below

So, people should use the one proposed but make drop-in replacements as you suggest.

kovianyo:
Thank you for taking a look on my changes.
The next thing I wanted to propose is to move the complicated wifi initailisation code somewhere else. The text above the code snippet speaks about aborting the startup, and the code is mainly about wifi initialisation. Only the timer and the startup() method from that code snippet is relevant to the satrup. It also checks for the existence of the init.lua which I don't really understand.
I indented my init.lua suggestion to be the very first code that is executed, before anything else, even wifi initialisation. I separated this satrtup aborting to a single file so it can be reused. So I would not recommed people to replace the statements in the code snippet already there. What do you think?

TerryE:
My concern with putting this type of detail into the uploads section of the documentation is that we dilute the content for those new to NodeMCU.  If essential idea of using a GPIO at startup to effect the running is a good-one and something that I sometimes do, but this particular implementation is very specific to @kovianyo's own needs and my not be a good fit for others, so a PR is the wrong construct at this state, and a documentation issue would be a better fit.

Perhaps what we need is some looser form of knowledge base in them of an FAQ organised by topic area -- say using this Github project wiki.  This needs more discussion amongst the project members

KT819GM:
I could be wrong, but after next `dev` merge `node.startupcommand` could be used and could replace everything in one-liner if we avoid declaring variables  like:

```lua
  local pin = 0 -- GPIO16, D0
```
like:
```lua
node.startupcommand("=if gpio.read(0) == 0 then print('Startup aborted')) else dofile('init.lua') end")
```
And on init.lua we have usual startup sequence. Not sure though if we shouldn't initialize pin as input with pullup before.

TerryE:
@KT819GM, yes you could add the test into your area startup command, or you could just `pcall(dofile'init.lua')` and include the GPIO test at the top of the init.lua file.

Many ways to skin a cat. 😊

marcelstoer:
> My concern with putting this type of detail into the uploads section of the documentation is that we dilute the content for those new to NodeMCU.

I see your point and I'm ok with rejecting this change for now. @kovianyo thanks again for suggesting this change.

Issue #1944, opened 2017-04-28
ESP32: dsleep time limit and wrong documentation


FrankBuss:
The documentation in docs/en/modules/node.md is wrong for dsleep, it doesn't need pin32 and pin8 connected, at least I measured GPIO16 and reset (EN) of my ESP32 with a continuity tester and it was not connected, but nevertheless the dsleep command works. And I don't know how it is planned if it will be merged with the esp8266 code, if not, then the documentation should not note something for the esp8266. And there is no second parameter for this function implemented in the dev-esp32 branch.

The other problem is the maximum sleep time. The limit of 35 minutes in the current source code is because luaL_optinteger is used in node_dsleep in components/modules/node.c. This limits the number to positive integers, which is 2^31, so 35 minutes. If I change the first line to this:

```
uint64_t us = (uint64_t) luaL_optnumber (L, 1, 0);
```

then it works even for 2 hours. Might even be a problem for the esp8266 branch. I tested it like this:

```
frank@xps8900:~$ stty -F /dev/ttyUSB0 115200 -parenb -parodd -crtscts min 1 time 0 ; date ; echo -e "node.dsleep(7200000000)\n" > /dev/ttyUSB0 ; cat /dev/ttyUSB0 | ts       
Fri Apr 28 09:46:43 CEST 2017
Apr 28 11:43:02 ets Jun  8 2016 00:22:57
Apr 28 11:43:02 
Apr 28 11:43:02 rst:0x5 (DEEPSLEEP_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
```

Not very accurate, but good enough. I guess it would work for much longer, too, because of the 64 bit value of the SDK, which is very useful, if you want to poll something, e.g. once per day. But this is for the float version of Lua, might need a comment in the docs that there is still the 35 minutes limit, if you use the int version of Lua.

PS: Might be useful to add some more functions to do a deep sleep with wake up on GPIO interrupt. The hardware can combine this with the timer, which would need a different Lua API.

jmattsson:
Ah, left-over from the ESP8266 branch. The ESP8266 does have the ~35min limitation, but we probably should allow longer sleeps on the ESP32 then. Tagging this as stuff to do as part of #1614, thanks!

tomsci:
> If I change the first line to \[use luaL_optnumber] then it works even for 2 hours

[...]

PR #3115, opened 2020-05-17
esp32: Support for GPIO wakeup in node.dsleep() (#2613)

### LINKED ISSUE #2613: ESP32: Support wakeup from deep sleep from external source
### Justification
Using `node.dsleep()` there does not seem to be any way to wake up the ESP32 with an external trigger (e.g. a button). There is `gpio.wakeup()` but that can't be used to wake up from **deep** sleep because it uses the `gpio_wakeup_enable()` API which [only works with light sleep, not deep sleep](https://github.com/espressif/esp-idf/blob/master/docs/en/api-reference/system/sleep_modes.rst#gpio-wakeup-light-sleep-only). For deep sleep there seems to be another API, [esp_sleep_enable_ext0_wakeup()](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/sleep_modes.html#_CPPv228esp_sleep_enable_ext0_wakeup10gpio_num_ti).

### Workarounds
None. Or maybe I missed something and the `esp_sleep_enable_ext0_wakeup()` API is already mapped to some NodeMCU API - in that case I'd be glad to get a pointer in the right direction.

In case I didn't miss anything: I already patched the firmware to go on with my project and added a new function `gpio.dsleep_wakeup(pin, intr_type)` that uses the IDF sleep API instead of the gpio API. If you're interested I can improve the patch (extend docs, add tests if necessary, incorporate results from discussions in this issue...) and open a PR. Of course that's a quick hack and considering there are more options to wake up an esp32 in deep sleep it might make sense to add a new `sleep` module for that.

Possibly related: #1612 #1614

tomsci:
Fixes #2613 

This PR extends `node.dsleep()` as discussed in issue #2613. I threw in a couple of other options I've been using too:

* Ability to specify integer number of seconds to allow long periods of sleep (#1944)
* Ability to keep internal pullups/pulldowns powered.
* Wake on touch support (I've tested this, it works well although with a caveat on node.bootreason)
* Ability to isolate pins (which I'm not sure if it saves power or not, but I put it in because the docs suggest someone might find it handy even I don't).

Things not in this PR:

* Support for EXT0 wakeup. As [I mentioned](https://github.com/nodemcu/nodemcu-firmware/issues/2613#issuecomment-619383164) I haven't found any need to use EXT0 since EXT1 supports everything EXT0 can do and more. So for simplicity, this PR always uses EXT1.
* Implementation of `node.bootreason()`, which is pretty much essential for many uses of deep sleep so you know why you got woken up. I have [written one](https://github.com/tomsci/nodemcu-firmware/blob/tomsci_ir_rebase/components/modules/node.c#L149) to match the existing esp8266 implementation as close as possible plus support getting GPIO info, but there is a lot of esp8266-specific stuff in the existing API and I'm not sure if we want to preserve it or simplify it. Plus I haven't tested all the other boot reasons like watchdog reset etc to see if they make sense or at all behave as expected on esp32. Whatever the consensus is on what we should do, I'll happily add to this PR, feedback welcome!
* Any support for light sleep, or changes to `gpio.wakeup()` etc. My current projects only use deep sleep so I'm only PRing code that I know works.

### DIFF
--- a/components/modules/node.c
+++ b/components/modules/node.c
@@ -1,5 +1,6 @@
 #include "module.h"
 #include "lauxlib.h"
+#include "common.h"
 #include "legc.h"
 #include "lundump.h"
 #include "platform.h"
@@ -7,6 +8,8 @@
 #include "vfs.h"
 #include "esp_system.h"
 #include "esp_log.h"
+#include "esp_sleep.h"
+#include "driver/rtc_io.h"
 #include "soc/efuse_reg.h"
 #include "ldebug.h"
 #include "esp_vfs.h"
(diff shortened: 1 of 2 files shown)


marcelstoer:
@chilipeppr do you have capacity to test & review this?

Issue #3066, opened 2020-04-10
When to drop the next master release


HHHartmann:
Sorry for opening an issue for this, but it seems right to discuss this in a visible place.

In https://github.com/nodemcu/nodemcu-firmware/pull/3060#issuecomment-610281199
@marcelstoer wrote:

> I think it would be very valuable (and thus welcome by the community) if we merged dev back to master in the coming weeks. It's been almost six months but there are a few things pending: https://github.com/nodemcu/nodemcu-firmware/milestone/14. So, I wouldn't mind landing this on dev soonish as it looks quite mature.

I would like to see some bugs fixed before we merge into master.
Uploading LFS.img with ESPlorer does not work anymore #2963
adc.readvdd33() #2925
enduser_setup "Handling POST" failure on 310faf7 and dev #2931

Especially the upload bug will leave a good part of the community without a working master.
The readvdd33 bug is a regression and I think should be fixed rather sooner than later.
They are both fixed in Terry's branch.
Also the broken enduser_setup should be addressed.

marcelstoer:
> I would like to see some bugs fixed before we merge into master.

Definitely! That's also the reason I maintain these milestones and the reference from PR to milestone (tracking yet-to-be-merged stuff).

nwf:
I would like #3064 to land as well before the next cut.

marcelstoer:
It's in the milestone, yes.

nwf:
In the interest of cutting a new release, I have dropped #2854 and #3075 from the milestone.  #2854 needs work, still, and #3075 should not be part of this cut to master, but rather the next one, since it touches so many modules (with a goal of NFC, to be clear, but just in case) and has not been given a thorough shakedown.

That means, I think, that we are blocking on @HHHartmann to decide that #3069 and #2985 should land or not.  Given @TerryE's resurrection of the Lua 5.3 branch, I'd like to make the drop to master approximately *now* so that we can then land his work on `dev` and have all of us test it out and do point fixes as needed.

HHHartmann:
As far as I am concerned  #3069 can go in as it is tested and imposes no risk at all.
 #3069 is tested only half. But with some luck it will be tested soon.

But as stated above there is at least one bug (#2963) in the current dev branch which is a showstopper and which is fixed in  #3075. 

It also seems wired to wait for this PR for 4 Months and then do the cut without it just the moment work on it continues.

marcelstoer:
[...]

Issue #2913, opened 2019-09-13
sqlite3 module not being detected


Sledmine:
Hi, i'm using the latest NodeMCU firmware build from the Cloud build service:
```
branch: master
        commit: 310faf7fcc9130a296f7f17021d48c6d717f5fb6
        release: 3.0-master_20190907
        release DTS: 201909070945
        SSL: false
        build type: float
        LFS: 0x0
        modules: color_utils,cron,file,gpio,http,net,node,pwm,sjson,sqlite3,tmr,                                                                                                                                                             uart,websocket,wifi,wifi_monitor
```

For some reason sqlite3 is not being detected as a module at all, running the example from the readthedocs.io just says "sqlite3" is nil.

```
dbtest.lua:1: attempt to index global 'sqlite3' (a nil value)
stack traceback:
        dbtest.lua:1: in main chunk
        [C]: in function 'dofile'
        stdin:1: in main chunk
```
Code used to test sqlite3:
```lua
db = sqlite3.open_memory()

db:exec[[
  CREATE TABLE test (id INTEGER PRIMARY KEY, content);

  INSERT INTO test VALUES (NULL, 'Hello, World');
  INSERT INTO test VALUES (NULL, 'Hello, Lua');
  INSERT INTO test VALUES (NULL, 'Hello, Sqlite3')
]]

for row in db:nrows("SELECT * FROM test") do
  print(row.id, row.content)
end
```

Is there something changed in this build that is messing with the module?

galjonsfigur:
`sqlite3` module was disabled due to large changes in NodeMCU Lua VM recently (see on [git blame](https://github.com/nodemcu/nodemcu-firmware/blame/310faf7fcc9130a296f7f17021d48c6d717f5fb6/app/modules/sqlite3.c#L28)) so it's impossible to build firmware with it without removing `#if 0` and corresponding `#endif` in `sqlite3.c` file. You could try using any version before commit https://github.com/nodemcu/nodemcu-firmware/commit/1990f95740e9f7d375b1157a25d0b6b508586c27 - it should at least be able to compile.

marcelstoer:
To my knowledge there has never been any consensus that want to really disable it. This `#if 0` comes as a complete surprise (at least to), I wasn't aware of it.

Sledmine:
[...]

Issue #2300, opened 2018-03-16
Should we remove SQLite3 from dev?


TerryE:
### Issue

As discussed in #2297 the **sqllite3** module compiles clean and works with trivial Lua examples, but there is a major question as to where this will every work with anything other than a trivial Lua example within the available ESP8266 RAM resources.  If it can't we used with a real Lua application, then why do we distribute this as part of dev?  Because if anyone does try to use it in anger and fails to get the application working, then we have no credible response to this sort of issue.

### Recommendation

-  Either the Original developer add a non-triival example of this libraries use to `lua_examples`, one that we can verify runs within the ESP8266 resources.
-  Or we backout this PR from dev 

Another alternative is to rehost this only for use on the ESP32.

TerryE:
I go back to my original Q on this: has _anyone_ managed to get this working on a real application?  Or even weaker statement: Has anyone managed to use a firmware build with this module included?

Picking up your point @marcelstoer, if people have downloaded unusable firmware then we are really misdirecting potential users.

TerryE:
@marcelstoer I am sorting out the ROTable support and one the library modules that is going to need quite a bit of work is the sqllite3 one.  The original contributor hasn't been seen since this was committed.  The implementation notes for the library recommend a minimum 128Kb working RAM.  I know that you said that 6% of builds in the first few weeks after it was introduced included it, but we've had no issues raised over this last year despite the fact that it won't work in a real-life application, so I doubt that anyone has actually attempted to use it.

My vote is that we drop this library on the next commit.

NicolSpies:
On [https://db-engines.com/en/ranking](url) SQLite has moved up to 10 place out of 345. It has benefits as a self-contained, server-less, zero-configuration, relational database in embedded devices. It appears to be aiming for the edge of the network. My feeling is that the ESP32 might be a better fit. So I agree to drop this but place it in "file 13" for the future. Maybe it will find a home in the future.

TerryE:
@NicolSpies, I agree that SQLite is a very popular engine and there is a set of usecases where it is a fit. However, an IoT device with under 64K total free RAM is not one.

luizfeliperj:
[...]

PR #3123, opened 2020-05-23
Remove sqlite3 module

nwf:
It's not clear that this ever worked, AFAICT nobody uses it, and it's an old version of the sqlite3 engine at this point.  Absent a maintainer, let's just get rid of it after this `master` cuts (#3066).

"Fixes" #2913 and #2300


TerryE:
If none of the other committers objects strongly, then I'll merge this in a day or so.

nwf:
How about we cut `master` (and close #3066) and then merge this?  Gives us time to back it out if someone does object.

TerryE:
I feel that I need to do some extra work checking out telnet and FTP server against Dev before we cut to master, and I also feel that this and the cut to master are independent issues.

HHHartmann:
I would also want to wait til after the master cut. That way we have our discontinuation note in the docs for one cycle. 
OTOH nobody really seems to miss it really.

marcelstoer:
> I feel that I need to do some extra work checking out telnet and FTP server against Dev before we cut to master

👍 

> I would also want to wait til after the master cut.

👍

marcelstoer:
Missing: removal of `sqlite3.md` from `mkdocs.yml`. Now fixed.

Issue #3128, opened 2020-05-26
Concerned about moving parts of modules into Lua


pjsg:
One of the nice things about nodemcu-firmware is that it allows you to build small apps very easily with just esplorer/similar. You build an image with the cloud builder, flash it, and then you can build something.

However, as modules start to be moved into Lua, this means that you can't do this stuff any more as you *need* to use the LFS to fit all the code into memory. This raises the complexity of the toy development environment significantly. 

What could be solutions to this problem? Maybe the cloud builder could build an LFS partition with all (or selected) Lua libraries that are part of nodemcu-firmware, *and* include this partition in the firmware file that the user then flashes onto their boards. 

Comments?

nwf:
I raised a similar point in the wrong repo at https://github.com/marcelstoer/docker-nodemcu-build/issues/91.  Since this is probably a better place for the conversation, I'll close that one and repeat what I said there here:

-----8<-----

It would be useful, I think, if we could pick at least in-tree lua_modules directly into a LFS image as part of the firmware bundle. This would lower the friction of jettisoning some "somewhat hairy" C modules in favor of Lua+C hybrids as I've been trying to drive us to (e.g. nodemcu/nodemcu-firmware#2819 and now nodemcu/nodemcu-firmware#3126).

However, I understand if there's hesitation to do LFS builds because people might want to add their own modules, and that opens "now I need to let people upload files" which is always a little hair-raising on the web. So...

Maybe instead of LFS, landing selected lua_modules in SPIFFS would still be useful. They could still be require-d as is, at the cost of increased memory consumption, and given @TerryE's proposal in nodemcu/nodemcu-firmware#2917, having Lua files in SPIFFS would eventually let us tell people how to compile their LFS images on-module.

-----8<-----

I know I'm being a little pushy on this, but to repeat myself: we have not demonstrated the ability to write and maintain complicated modules in C.  In many cases our modules merely appear to work or work well enough for toys but not for anything robust; memory management errors abound.  Fixes for these issues have been very slow in coming (which, I acknowledge, is partially my fault as I would rather *not* fix them in C).

[...]

Issue #2527, opened 2018-10-20
BME280 module only supports I2C?


JBFUK:
So far as I can tell it seems that the BME280 module only supports I2C whereas the BME280 itself supports both I2C and SPI.  It would be great to see SPI support added.  It would then be possible to use CS feature and run multiple devices on the same bus.

nwf:
Not to argue against SPI support, but in the interim... the BME280 already supports having two on the same bus, and you might be able to get away with an I2C mux (e.g. TCA9548A) if you need fewer than 17 of them (and don't mind having to steer the mux between samples, of course).  If #2465 lands, too, you'll be able to drive a lot more I2C busses, further multiplying the number of sensors. :)

JBFUK:
Sure, but the device supports SPI and that would be my preference as
opposed to I2C and messing around with mux etc.  I don’t see why the
library shouldn’t give me the choice to use either option.

nwf:
As I said, I didn't mean any of what I said to argue against SPI support, but unless you're willing to do the work, I suspect someone's going to close this as per #1010.

JBFUK:
I don’t disagree that this isn’t really an ‘issue’, I just couldn’t see
anywhere to post a feature/enhancement request.

marcelstoer:
The only likely candidate for enhancements to that module is @vsky279 (its creator). I'm closing this as suggested but PRs are of course more than welcome.

PR #3132, opened 2020-05-29
bme280 driver in Lua+C

### LINKED ISSUE #3126: Rewrite bme280 driver in Lua+C
In an effort to decrease the project's technical debt, I've proposed (in #3124) that the C-only `bme280` be cut into a Lua layer that drives communication and a C module that does any requisite low-level fiddling.  This will be similar to #2819 and may, incidentally, address #2527 correctly as well.  (As you can see, between that and #2241, this module has a history of being sufficiently troublesome that issues have been worked around rather than fixed.)  This rewrite will *also* eliminate yet another source of static allocation and artificial limitation in our driver layer.  All in all, deeply worth it.

I would like it very much if the result of this rewrite included a test driver, as well.  Ideally `bme280` becomes an example of how all drivers should look.  No pressure (ha!). ;)

OK, so, what does this entail, concretely?

Looking at the existing `bme280.c`, we almost immediately see that there's a bunch of goo associated with performing `i2c` transactions: `bme280_i2c_addr`, ` r8u_n`, `r8u`, and `w8u`.  These can all be directly transcoded to Lua.  The results from the `i2c` package will be Lua strings.

The `bme280_data` structure holds per-device state obtained during `setup`, which means we probably want to hold it instead in a Lua userdata object that represents the sensor.  Reading `setup`, it looks like this takes 3 or 4 I2C transactions to set up, including getting the chip ID, so I think our sensor object constructor should take 2 or 3 Lua strings (`DIG_T`, `DIG_P`, and `DIG_H`) to build itself.  Since it appears that the static values `bme280_t_fine`, `bme280_h`, and `bme280_hc` may outlive a particular measurement, they should perhaps join the `bme280_data` in the per-device userdata object.

Armed with the sensor parameters, we can look at interacting with the device: `bme280_lua_startreadout` does some i2c transactions, easily done instead in Lua, and schedules a timer, also easily done in Lua.  (Can the device not raise IRQs?)

Once it's time to read data, `bme280_lua_read` does some more i2c transactions and a lot of math, including storing "fine" constants between procedure calls.  To continue the theme, the transactions should be done in Lua, and passed to a C routine that takes the sensor userdata object and does the requisite math.  For my $0.02, the results should just land in a Lua table.

The abbreviated `bme280_lua_temp`, `bme280_lua_baro`, and `bme280_lua_humi` functions are readily achieved by having the Lua replacement for `bme280_lua_read` perform fewer transactions and call into the C function with fewer string arguments.  That function should return a table with fewer keys as a result.

For the purposes of PRing, I'd suggest placing code in

* `app/modules/bme280_math.c` for the C layer
* `lua_modules/bme280/bme280.lua` for the Lua top-level and main driver
* `lua_modules/bme280/bme280_test.lua` for the test script (ideally in mispec, I think?  See #2145).

vsky279:
Fixes #3126, #3124, #2241, 

Rewritten BME280 driver based on the discussion in #3126.
I intend to add also SPI support (#2527) which should be also relatively small modification of the Lua driver.

There is a conflict when original BME280 module is included in the build. `require('bme280')` references to the original C module and not to the new Lua driver. This can be a bit confusing. The natural solution would be to that this driver replaces the original C driver.

I assume this should be probably merged only when #3128 is implemented.

Please review the proposed commit.


vsky279:
I don't know how to make Travis CI happy when it complains that
```lua_modules/bme280/bme280.lua:22:3: accessing undefined variable bme280_math```
`bme280_math` is the new module so it should be known to the builder.

**Comment by galjonsfigur** [comment]
@vsky279 You have to add entities in `luacheck` config file that can be found in `tools/luacheck_config.lua`. You can write it yourself or use luacheck config helper(also in `tools` folder) script that will generate this for you - little note on how to use it can be found [here](https://github.com/nodemcu/nodemcu-firmware/blob/master/tools/luacheck_config_helper.lua#L4-L6).

I thought that I've written documentation on how to add new entities in luacheck config when adding new C module but turns out I was wrong - I could add this in documentation or in project wiki.

**Comment by vsky279** [comment]
@galjonsfigur Thanks. I did manual update and now it's green.

**Comment by vsky279** [comment]
#### Lua 5.3 compatibility issue - `lua_pushfstring` works differently for non-printable characters
I was using `lua_pushfstring` with `%c` formating string to return string array of configuration parameters (3 bytes). Since Lua 5.3 the string pushed to stack for e.g. value 240 is `<\240>`. Different approach using just `lua_pushstring` is needed.

I am not sure if this cannot cause some issues in other modules.

https://github.com/nodemcu/nodemcu-firmware/blob/8d091c476edf6ae2977a5f2a74bf5824d07d6183/app/lua53/lobject.c#L426

File: mkdocs.yml [hunk 1]
```diff
@@ -48,6 +48,10 @@ pages:
 - Lua Modules:
     - 'Lua modules directory': 'lua-modules/README.md'
     - 'bh1750': 'lua-modules/bh1750.md'
+<<<<<<< HEAD
+=======
+    - 'bme280':  'lua-modules/bme280.md'
+>>>>>>> Initial commit
```
nwf, line 54:
```suggestion
    - 'bme280':  'lua-modules/bme280.md'
```

PR #3134, opened 2020-05-30
Remove deprecated rc module

nwf:
It has long been superseded by the rfswitch module.

Hold as PR until after next master cut.


HHHartmann:
Shouldn't we announce a deprecation first instead of just having a module vanish without warning or explanation?

nwf:
Well, the documentation already described it as "superseded", but I'm happy to wait for the `master` drop for this to land.

marcelstoer:
Missing: removal of `rc.md` from `mkdocs.yml`. Now fixed.

nwf:
Hm... This and #3123 passed CI even with the bad `mkdocs.yml`.

marcelstoer:
Yes, but the docs build on RTD failed and that info isn't fed back to GitHub. Not sure if/how that could be achieved.

Issue #3130, opened 2020-05-28
Move to Kconfig-like build configuration


nwf:
`user_modules.h` and `user_config.h` are a little unfortunate.  Among other things, this style of configuration means:

* users express their configuration preferences by directly modifing files managed by git
* we always build everything
* we occasionally have to paw through C headers looking for particular strings with regexes (e.g., in `tools/update_buildinfo.sh`)

We should move to something like Kconfig, where we articulate the dependencies and parameters in a machine readable language and have tools generate a `.config` file that everything downstream can source easily.  Because `.config` is merely a list of assertions, with no computational content, there's less risk that things go wrong (e.g. the use of `#if 0` in `user_modules.h` will cause a discrepancy between actually included modules and the list of included modules).

We can use this to speed up builds (skipping `mbedtls`, for example, if there's no need for it, and not building modules not requested) but more significantly the improved articulation of configuration will let tools beyond C consume it and let us, for example, decide which Lua module files land in LFS/SPIFFS more readily, I think.

https://pypi.org/project/kconfiglib/ is a full Python implementation (and so works on Windows) and is already used by esp-idf upstream.

This should be very straightforward, just time consuming and invasive across the tree, with impacts on the various build tools (esp. web image builder).  Given the invasiveness, if there is consensus that this is a good idea, I propose we wait until the 5.3 work has landed, first.

TerryE:
@nwf Nathaniel, @jmattsson Johny, this also overlaps with our discussions on #1661.  What Johny and I were suggesting is to move our ESP8266 builds to an ESP-IDF (or variant) build environment with as much as the folder tree unified across the two branches.  OK, our ESP8266 builds will stil need to build an ESP8266 non-OS SDK based firmware image rather than an ESP32 RTOS based one, but the more of the core components and build processes that are shared across the two variants the better, IMO.

But quite a bit more planning needed before we start along this path, I think.

nwf:
Given that I'd like to move us in the direction of Lua+C modules sooner rather than later, I think pursueng this sooner rather than later is probably a good idea.  The alternatives all necessitate teaching builders about the dependencies anyway, right?  If we want "click button foo, get firmware image with foo.lua in LFS/SPIFFS/both and foo.c compiled in" to work.

jmattsson:
[...]

PR #3135, opened 2020-05-30
Kconfig: rename LUA_MODULE_* to NODEMCU_CMODULE_*

nwf:
As we begin to contemplate baking into firmware modules entirely written
in Lua or which are split between both Lua and C, the name LUA_MODULE_*
more obviously refers to the in-Lua bits, so rename the C ones out of
the way.

I have not yet convinced my machine to build for the esp32, but I think this is probably about right.  Still, someone checking for me would be welcome.

This is part of #3130, indirectly, at least in keeping with the idea of harmonizing the two branches.

### DIFF
--- a/components/platform/include/user_version.h
+++ b/components/platform/include/user_version.h
@@ -6,9 +6,9 @@
 #define NODE_VERSION_REVISION	0U
 #define NODE_VERSION_INTERNAL   0U
 
-#define NODE_VERSION	"NodeMCU ESP32"
+#define NODE_VERSION	"NodeMCU ESP32" " built with Docker provided by frightanic.com\n\tbranch: dev-esp32\n\tcommit: f4887bf134235c05e6c9b2efad370e6d5018f91a\n\tSSL: true\n\tmodules: -\n"
 #ifndef BUILD_DATE
-#define BUILD_DATE	  "unspecified"
+#define BUILD_DATE	  "created on 2020-05-30 18:01\n"
 #endif
 
 #define SDK_VERSION "IDF"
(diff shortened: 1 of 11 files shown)


TerryE:
Please see my discussion on #3130.  This change makes no sense to me as these are Lua runtime modules, albeit written in C.  If we really want to emphasise the "C-ness" of these then `LUA_CMODULE` would make more sense.

components/lua/lua.h
```diff
@@ -19,7 +19,7 @@
 
 
 #define LUA_VERSION	"Lua 5.1"
-#define LUA_RELEASE	"Lua 5.1.4"
+#define LUA_RELEASE	"Lua 5.1.4" " on ESP-IDF v3.3.1"
```
marcelstoer, line 22: I dislike two things here:
- static version; in both the cloud build and the Docker image I extract it dynamically, see e.g. https://github.com/marcelstoer/nodemcu-custom-build/blob/master/ESP32/before-script.sh#L68
- the ESP-IDF version is unrelated to **LUA**_RELEASE; in my scripts I only append it to have it show up in the boot message
-> I would revert this line

nwf: Whoops, that seems to have been accidentally added.  I blame my lack of `git`-fu.

nwf:
I think it's ready, but it'd be good for someone with an ESP32 environment to look over my shoulder.

components/platform/include/user_version.h
```diff
@@ -6,9 +6,9 @@
 #define NODE_VERSION_REVISION	0U
 #define NODE_VERSION_INTERNAL   0U
 
-#define NODE_VERSION	"NodeMCU ESP32"
+#define NODE_VERSION	"NodeMCU ESP32" " built with Docker provided by frightanic.com\n\tbranch: dev-esp32\n\tcommit: f4887bf134235c05e6c9b2efad370e6d5018f91a\n\tSSL: true\n\tmodules: -\n"
```
marcelstoer, line 9: Arghh, missed this one...shouldn't have ended up in this change set.

PR #3149, opened 2020-06-07
BMEXXX sensor documentation update (Lua 5.3 compatibility)

vsky279:
Fixes - no reported issue>.

BME280 and BME680 sensor modules documentation includes examples that are
- not Lua 5.3 compatible (`("%d"):format(3/2)` fails as 3/2 has no integer representation)
- were intented for integer version so includes code that is not needed in now standard flow version of the firmware.

### DIFF
--- a/docs/modules/bme280.md
+++ b/docs/modules/bme280.md
@@ -178,54 +178,23 @@ sda, scl = 3, 4
 i2c.setup(0, sda, scl, i2c.SLOW) -- call i2c.setup() only once
 bme280.setup()
 
-P, T = bme280.baro()
-print(string.format("QFE=%d.%03d", P/1000, P%1000))
-
--- convert measure air pressure to sea level pressure
-QNH = bme280.qfe2qnh(P, alt)
-print(string.format("QNH=%d.%03d", QNH/1000, QNH%1000))
-
-H, T = bme280.humi()
-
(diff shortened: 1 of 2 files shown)


pjsg:
Again, I'm not convinced that you should be removing compatibility with some versions of the firmware.

vsky279:
I don't like the current code not being compatible with Lua 5.3. Another way to make it compatible but also working in integer version is to add `math.floor`, i.e.
```Lua
local Tsgn = (T < 0 and -1 or 1); T = Tsgn*T
print(("T=%s%d.%02d"):format(Tsgn<0 and "-" or "", math.floor(T/100), T%100))
```
instead of just
```Lua
print(("T=%.2f"):format(T/100))
```
Do we need such complexity in an example?
I tend to think that if someone goes for the integer version, he/she has to build it himself, so I assume he/she knows what he's doing, knowing all the implications.

Anyway in the new BME280 driver (#3132) returned values are float so it is not integer version compatible. Should it be?

marcelstoer:
The examples in the docs should work with whatever defaults we have in the firmware (5.1 vs. 5.3). If an example only works with int or float then it should mention that clearly.

vsky279:
This PR has
\+ readability of the example and it's simplicity
\- one version for float another for integer version

It seems that consent is that compatibility is winning over simplicity (simply said) so I will close this PR.

PR #3150, opened 2020-06-07
DS18B20 Lua module - removing integer version related code

vsky279:
Fixes - no reported issue.

This PR removes code supporting integer version of the firmware. As float version is now standard I think there is no reason to have integer version support in standard volume.


pjsg:
I'm not sure that I'm happy with this -- the integer builds are not going away yet. It might make sense to have two versions -- one for float builds and one for integer builds.

vsky279:
Ok. I will add integer version of the module. I just don't like the garbage in the standard module version.

vsky279:
I think we should decide whether to merge this or close the PR. Philip @pjsg?

pjsg:
I vote to merge this one.....

Issue #2916, opened 2019-09-16
RFC: rework LED strip modules


HHHartmann:
Currently we have several LED strip modules for different Hardware.
While some only support writing out the data to the strip there are several features for the WS2812. I would like to make these features available to all led strips.

1. It would be nice to abstract the buffer used for the WS2812 to be HW independent and adapt all HW modules to be able to work with these buffers.

2. I would also like to address the different color orders of the strips. AT least RGB and GRB are needed, maybe more. The order would be given when writing to a strip, or when initializing it. (whatever we define)

3. Since usually colors are given in RGB I would like to move the color definition in `color utils Module` and `WS2812 effects Module` to RGB instead of GRB. So for the logic in the buffers we would use RGB

Or course a white channel will also be available, Just wanted to keep it short here.

I would especially like to hear your opinion on 3. since that would probably have to be a breaking change, or we could duplicate functions and then have `color_utils.hsv2grb` and `color_utils.hsv2rgb`.

### Justification
Make available the buffer manipulation routines for all LED modules, not only WS2812

### Workarounds
Do it yourself completely or recode the buffer to a string usable by the other modules.
No workaround for the `WS2812 effects Module`.

nwf:
I have not yet tested beyond "it compiles", and it is premature to PR, my `pixbuf` split.  If someone wants to grab https://github.com/nwf/nodemcu-firmware/tree/led-strip-refactor and try it out, tho', that'd be grand.  I'll try to keep it up to date when I put more time into it, but right now I'm getting up out of this chair. :)

HHHartmann:
I recently also stumbled across LED strips supporting WWA (WarmWhite/ColdWhite/Amber) so how can/should we deal with that?

 1. Ignore it, so people can still use the buffer but have to remember the mapping RGB <-> WWA
 2. Have the buffer "know" what type it is (RGB GRB RGBW WWA) and have effects and color utils check if they can handle it

nwf:
Hm; that's an interesting question.  I was, relatedly, pondering whether we should have a "generic" pixbuf that's basically just a string of bytes to be manipulated ("1 byte per LED").

How about something like this?
```
enum pixbuf_type {
  PIXBUF_TYPE_WILD /* just bytes */,
  PIXBUF_TYPE_GRB /* 3-color: green, red, blue order */,
  PIXBUF_TYPE_GRBW /* 4-color: GRB + white */,
[...]

PR #3157, opened 2020-06-12
fix a memory leak in ws2812_buffer_shift

Firenox89:
fix a memory leak in ws2812_buffer_shift by freeing the same amount of bytes we allocated before

Memory was allocated with `luaM_malloc(L, sizeof(ws2812_buffer_shift_prepare) + shift_len);`
but freed with `luaM_free(L, prepare);` this caused a leak by shift_len bytes per call.

The code I used to test with:
```lua
while (true) do
    buffer:shift(10, ws2812.SHIFT_CIRCULAR)
    ws2812.write(buffer)
    print(node.egc.meminfo())
    --coroutine.yield()
end
```
This caused a leak of 40 bytes per call before this commit, while after the numbers are stable.

### DIFF
--- a/app/modules/ws2812.c
+++ b/app/modules/ws2812.c
@@ -278,7 +278,8 @@ int ws2812_buffer_shift(lua_State* L, ws2812_buffer * buffer, int shiftValue, un
   ws2812_buffer_shift_prepare* prepare = ws2812_buffer_get_shift_prepare(L, buffer, shiftValue, shift_type, pos_start, pos_end);
   ws2812_buffer_shift_prepared(prepare);
   // Free memory
-  luaM_free(L, prepare);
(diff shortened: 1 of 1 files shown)


TerryE:
What this actually underlines in my mind is that we using Userdata for this type of usecase, then we can leave resource recovery to the Lua GC and this removes most paths for potential leakage.

nwf:
I'm not opposed to this landing in `dev` now, since a fix is a fix, but I continue to hope that #2916 and my `pixbuf` replacement is ready for the next drop to master.  https://github.com/nwf/nodemcu-firmware/blob/dev-active/app/modules/pixbuf.c already uses `userdata` for all its storage needs and eschews temporary allocations.

Issue #3122, opened 2020-05-23
Deprecate and remove ws2812 effects (and provide an in-Lua replacement)


nwf:
There is no reason whatsoever that this module should be in C; timers and ws2812 frame buffers are easily managed from Lua.  Worse, its existence complicates implementing #2916, which is obviously a good idea for the health of the entire tree.

If it were up to me, I'd just rip it out wholesale right now.  But I recognize it's not, so I wonder if someone who cares about the module would like to write Lua equivalents of each of its effects for inclusion in `lua_examples`.

If it turns out that there are operations that cannot be easily achieved from Lua, then we should augment the buffer interface until that's not true, and I would gladly add such things to my nascent #2916 implementation.

HHHartmann:
The ws2812_effects module consists of two main functionalities which are interwoven and should be separated.
1. the player, which repeatedly updates the LED strip with the newly calculated effect.
2. the effects themselves.

So there should be a player and separate effects which can be passed to the player.
It should also support all (3 I think) different Led Modules.
The effects should probably have an interface consisting of `init(params)` and `next_buffer next_step(old_buffer)` to calculate the next step.

Originally I was planning to adapt the module also, but I agree that both should be doable in Lua.

Should we sneak in a deprecation warning in the module then before the release?

@marcelstoer Do you have statistics of the usage in the cloud builder of this module?

nwf:
@HHHartmann Sneaking a deprecation warning seems like a good plan to me.  Since we don't have a replacement module ready, point at this issue?

FWIW, I've been mocking up a proposal for `lua_examples` that I'll push in just a moment.  You are more than welcome to take it and mutate it to your heart's desire. :)

HHHartmann:
I have been thinking about this. It certainly would be good to allow individual effects in Lua but on the other hand executing some effects that change many pixels individually will take around factor 100 longer to execute in Lua than in C. But then, looking at the existing effects at most the flickering effects modify several pixels, but not that many, so that might be neglected.

nwf:
If there are bulk editing combinators that would be useful to have in C, we should add them to `pixbuf`.  I'm not opposed to a "miscellaneous special purpose `pixbuf` functions" module, either (`pixbuf.morefun`?) if it's necessary.  I mostly want the timer callbacks and `pixbuf` references handled in Lua and not in C.  Actual math that's gotta go fast can, of course, remain in C.

[...]

PR #3158, opened 2020-06-13
LED strip refactor

### LINKED ISSUE #2916: RFC: rework LED strip modules

nwf:
Addresses #2916 (mostly, anyway).  Leaves #3122 to be done, and doesn't change any of the color utils libraries.

I've moved the so-called "ws2812" tests, which were really just tests of its buffer, over to the generic `pixbuf` and this now passes those tests.  While the ws2812_effects library is ported to work with `pixbuf` objects and does compile, it is operationally untested (though I don't anticipate any real breakage).

I have not actually attempted to use this to drive LEDs yet (ha) but the drivers are only lightly modified to use the new data structure, so I hope it should work fine.


docs/modules/pixbuf.md
```diff
@@ -0,0 +1,274 @@
...
+print(buffer:get(1))
+0	0	0	0
+```
+
+## pixbuf.buffer:set()
+Set the value at the given position
+
+#### Syntax
+`buffer:set(index, color)`
+
```
HHHartmann, line 54: As mentioned above, the goal should be to have the same user code for RGB and GRB buffers, so we need to swat them either here or on output. (or in a later PR?)

lua_tests/mispec_pixbuf_1.lua
```diff
@@ -133,17 +127,17 @@ describe('WS2812 buffers', function(it)
...
 
     it:should('mix with strings correctly ', function()
-        buffer1 = ws2812.newBuffer(1, 3) 
-        buffer2 = ws2812.newBuffer(1, 3) 
+        buffer1 = pixbuf.newBuffer(1, pixbuf.TYPE_GRB)
+        buffer2 = pixbuf.newBuffer(1, pixbuf.TYPE_GRB)
 
         buffer1:fill(10,22,54)
         buffer2:fill(10,27,55)
-        buffer1:mix(-257,buffer1:dump(),255,buffer2:dump())
```
HHHartmann, line 141: Did you remove the string capability from pixbuf?

lua_tests/mispec_pixbuf_2.lua
```diff
@@ -10,138 +10,133 @@ local function initBuffer(buffer, ...)
...
-        buffer1 = ws2812.newBuffer(4, 4)
-        buffer2 = ws2812.newBuffer(4, 4)
+        buffer1 = pixbuf.newBuffer(4, pixbuf.TYPE_RGBW)
+        buffer2 = pixbuf.newBuffer(4, pixbuf.TYPE_RGBW)
 
         initBuffer(buffer1,7,8,9,12)
         initBuffer(buffer2,0,0,7,8)
         buffer1:shift(2)
-        ok(equalsBuffer(buffer1, buffer2), "shift right")
+        ok(buffer1 == buffer2, "shift right")
```
HHHartmann, line 23: If we use our own == her we would need to add tests for == somewhere.
Else we test the operations with untested code (our ==)

app/modules/pixbuf.c
```diff
@@ -0,0 +1,555 @@
...
+  luaL_argcheck(L, srcLen + start - 1 <= buffer->npix, 2, "does not fit into destination");
+
+  memcpy(buffer->values + (start - 1) * channels, src, srcLen * channels);
+
+  return 0;
+}
+
+
+static int pixbuf_set_lua(lua_State *L) {
+  pixbuf *buffer = pixbuf_from_lua_arg(L, 1);
```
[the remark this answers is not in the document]
nwf: Added setRGB.

app/modules/pixbuf.h
```diff
@@ -0,0 +1,56 @@
+#ifndef APP_MODULES_PIXBUF_H_
+#define APP_MODULES_PIXBUF_H_
+
+enum pixbuf_type {
+  PIXBUF_TYPE_RGB /* 3-color: Green, Red, Blue order */,
```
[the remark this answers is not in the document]
nwf: Fixed.

app/modules/pixbuf.h
```diff
@@ -0,0 +1,56 @@
...
+  PIXBUF_TYPE_GRB /* 3-color: Green, Red, Blue order */,
+  PIXBUF_TYPE_RGBW /* 4-color: RGB + White */,
+  PIXBUF_TYPE_GRBW /* 4-color: GRB + White */,
+  PIXBUF_TYPE_WWA /* 3-color: warm White, cold White, Amber */,
+
+  PIXBUF_TYPE_IBGR /* 3-color (Blue, Green, Red) & leading Intensity */,
```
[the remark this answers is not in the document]
nwf: "BBGR" is ambiguous. :(

app/modules/pixbuf.h
```diff
@@ -0,0 +1,56 @@
...
+
+  PIXBUF_NTYPES
+};
+
+typedef struct pixbuf {
+  const size_t npix;
+  const uint8_t type;
+
+  /* FAM; true size is npix * pixbuf_channels_for(type) */
```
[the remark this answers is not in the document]
nwf: Flexible Array Member.

app/modules/pixbuf.h
```diff
@@ -0,0 +1,56 @@
...
+  PIXBUF_SHIFT_CIRCULAR
+};
+
+pixbuf *pixbuf_from_lua_arg(lua_State *, int);
+const size_t pixbuf_size(pixbuf *);
+
+// Exported for backwards compat with ws2812 module
+int pixbuf_new_lua(lua_State *);
+
+// Exported due to ws2812_effects; delete from here to HERE once that's gone
```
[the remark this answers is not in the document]
nwf: Because if they're #ifdef'd out we won't get the compiler doing type-checking for us until someone builds with WS2812_EFFECTS.  It's harmless to have the prototypes in scope even without WS2812_EFFECTS, and we need the struct definition anyway.

docs/modules/pixbuf.md
nwf: Done.

lua_tests/mispec_pixbuf_1.lua
nwf: Yes, though perhaps for somewhat specious reasoning: it wasn't documented and the test was flirting with OOM.

lua_tests/mispec_pixbuf_2.lua
nwf: Well, these *are* tests of our `==`.  But yes, maybe we should test for `~=` sometimes.  Here and elsewhere, more tests are welcome.

app/modules/apa102.c
```diff
@@ -79,9 +81,26 @@ static int apa102_write(lua_State* L) {
...
+  switch(lua_type(L, 3)) {
+  case LUA_TSTRING: {
+    size_t buf_len;
+    buf = luaL_checklstring(L, 3, &buf_len);
+    nbr_frames = buf_len / 4;
+    break;
+   }
+  case LUA_TUSERDATA: {
+    pixbuf *buffer = pixbuf_from_lua_arg(L, 3);
+    luaL_argcheck(L, buffer->type == PIXBUF_TYPE_IBGR, 3, "Pixbuf not IRGB");
```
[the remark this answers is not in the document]
HHHartmann: In the errormessage you write IRGB

app/modules/pixbuf.c
```diff
@@ -0,0 +1,555 @@
...
+
+  luaL_argcheck(L, fade > 0, 2, "fade value should be a strict positive number");
+
+  pixbuf_fade(buffer, fade, direction);
+
+  return 0;
+}
+
+static int pixbuf_fill_lua(lua_State *L) {
+  pixbuf *buffer = pixbuf_from_lua_arg(L, 1);
```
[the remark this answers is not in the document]
HHHartmann: hmm this works but it might me better to have one common function which does all the work which is called with a boolean to know whether to switch colors or not.  That way the cost for maintenance of the code would be less.

app/modules/pixbuf.c
```diff
@@ -0,0 +1,681 @@
...
+    for (size_t i = 0; i < channels; i++)
+    {
+      buffer->values[channels*led+i] = luaL_checkinteger(L, 3+i);
+    }
+  }
+
+  return 0;
+}
+
+// XXX This duplicates pixbuf_set_lua more than might be desirable
```
HHHartmann, line 438: Yes it does

app/modules/pixbuf.h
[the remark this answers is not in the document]
HHHartmann: OK google does not find it for
FAM 
computer FAM C
it FAM C
urban dictonary FAM
FAM; true size

It finds
C++ FAM

Maybe you could spell it out?

lua_tests/mispec_pixbuf_1.lua
HHHartmann: True, Maybe we should document it rather and if necessary split the tests in even more parts. As string buffers are allowed everywhere else it would be inconsistent if it wasn't here.

lua_tests/mispec_pixbuf_2.lua
HHHartmann: But still we should only test one aspect in one test and not use two untested components.

app/modules/pixbuf.c
```diff
@@ -0,0 +1,681 @@
...
+}
+
+LROT_BEGIN(pixbuf_map, NULL, LROT_MASK_INDEX | LROT_MASK_EQ)
+  LROT_TABENTRY ( __index, pixbuf_map )
+  LROT_FUNCENTRY( __eq, pixbuf_eq_lua )
+
+  LROT_FUNCENTRY( __concat, pixbuf_concat_lua )
+  LROT_FUNCENTRY( __tostring, pixbuf_tostring_lua )
+
+  LROT_FUNCENTRY( dump, pixbuf_dump_lua )
```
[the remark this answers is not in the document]
nwf: I think `dump` is mostly a debugging function, not something we expect people to use day-to-day.

lua_tests/mispec_pixbuf_2.lua
nwf: In black-box testing, it's often not possible to "only test one aspect" at a time.  Previously, this test depended on `fill` (via `initBuffer`), `shift`, and `dump`; now it depends on `fill`, `shift`, and `eq`.  I'm not opposed to having more tests, and even some that compare `==` on `dump` output, but I don't think the lack of orthogonality is new as of this diff.

lua_tests/mispec_pixbuf_1.lua
nwf: "allowed everywhere else"?

nwf:
I am deeply skeptical of the `tm1829` code; it appears to be generating a very `ws2812`-style waveform in software, but it bounces interrupts per byte, rather than per packet, which suggests to me that some fraction of the time it gets the timing wrong and the strip glitches.

Also despite the documentation claiming that its input string is in RGB order, the code does not actually change the order of any bytes, so its claim of sending GRB is probably wrong.  But the documentation may predate the conversion to not corrupt the input string, too, so...

nwf:
Blocking on #3230 since I want to be able to test pixbufs on the host.

nwf:
Eh, nevermind.  I'll just carry the local bits for host testing in my tree for a bit; no reason to block this.

HHHartmann:
ok, this might be duplicate but I can't find it anywhere:

I've been thinking about removing the *RBG methods again. The approach is a bit of a dead end.
What we want is an implementation that is for all color patterns (RGB, GRB) but atm we are stuck with the legacy GRB.

So we should have a method which configures the meaning of the params for those Methods and remove the RGB variants again.

Since people will have to touch their code anyways they might as well also configure the packet for GRB to match the legacy order. All new code could use the more commonly used RGB notation regardless of the physical layout of the LED strip.

So the default should be RGB.

We can the decide to remove the legacy GRB order at a later point or just leave it in.

So maybe add `pixbuf.setcolorchannelorder(TYPE_GRB | TYPE_RGB ) which would also work for RGBW and GRBW. We have to think about WWA though.
Maybe WWA should allways be handled raw, thus not be changed regardless of the configured order. (and not fail as it is now for the RGB variants)

nwf:
In thinking about this more myself... I'd like to remove strip type and color space consideration completely and just supply raw interfaces everywhere.  IMHO, pixbuf's job is to buffer and manipulate pixels without understanding what those pixels "mean" when displayed; its understanding of pixels should, IMHO, be limited to a very simplistic "1 to 4 channels, each between 0 and 255, inclusive".  It is currently mission-creeping towards being an entire graphics library, an endeavor I have no real desire to pursue.

I'm happy to add a `:map(f, ilo, ihi)` method which constructs a *new* `pixbuf` by applying `f(chan1, chan2, ...)` on each pixel from `ilo` (default: 0) to `ihi` (default: end of buffer) and taking `f`'s results (up to 4) as the corresponding pixel in the output buffer.  That will let users easily change channel ordering (`:map(function(g,r,b) return r,g,b end)`) or even change color spaces (`:map(color_utils.hsv2grbw)`) or perform channel projection (`:map(function(r,g,b) return r end)`) or whatever their hearts desire, without duplicating swaths of the API.

HHHartmann:
Also an approach. But I fear that it gets too slow at some point, so that only short LED strips can be used with a decent frame rate. But to be true I have no experience.

I would prefer to just remove the RBG variants as I sketched above. The functionality is in there already, only the `native` param would have to be changed to a global variable.

In any case I would have the map method to allow for a second param which passes the output buffer, also making sure that it can be the same.

Otherwise we'd get lots of objects on the heap which would have to be GCed, which does not increase speed either.

But still I would prefer to have a solution closer to what you already did. That would leave the Lua developer head more focused on the actual algorithm.

nwf:
Hm.  Still some warnings in `NTest_pixbuf` causing CI's build to fail.  Also it'd be really good to get someone actually using this code; so far I've done little more than compile and run the tests, because I've drifted off to other projects and there hasn't, seemingly, been a lot of interest.

HHHartmann:
Fixed the luacheck warnings. Now it fails in 
pixbuf.c: In function 'pixbuf_opt_from_lua_arg':
pixbuf.c:21:3: error: implicit declaration of function 'luaL_testudata' [-Werror=implicit-function-declaration]
   return luaL_testudata(L, arg, PIXBUF_METATABLE);
   ^
pixbuf.c:21:3: error: return makes pointer from integer without a cast [-Werror]
cc1: all warnings being treated as errors

nwf:
Oh, boo, `luaL_testudata` is a 5.3-only API.  In 5.1 it was called `luaL_checkudata`.

nwf:
Here's a revised version, which I think incorporates @HHHartmann's changes correctly, and sits atop #3352 to correct not building on Lua 5.1

HHHartmann:
Btw, there are scripts to execute luacheck locally for both Linux and windows hidden in the tools/travis folder

nwf:
Whoop, sorry to lose that commit in all the rebasing.  Thanks for reapplying.

HHHartmann:
Finally got around to testing this. I tested output on both channels and several effects.

the output seems to work whereas only effects without cycle work. Cycling does not work yet.

Also in the former version setting speed and delay to a float was working, now an error is thrown.

nwf:
@HHHartmann: Thanks!  Can you give me a piece of code to test, since I've never used ws2812-effects?

HHHartmann:
The first two modes work but the latter two only initialize the strip but then don't cycle it.

```Lua
  ws2812.init(ws2812.MODE_SINGLE)
  effects_buffer = pixbuf.newBuffer(9, 3)
  ws2812_effects.init(effects_buffer)
  ws2812_effects.set_color(0,255,0)
  ws2812_effects.set_mode("blink")
  ws2812_effects.set_mode("larson_scanner")
  ws2812_effects.set_mode("cycle")
  ws2812_effects.set_mode("rainbow_cycle")
  ws2812_effects.start()
```

nwf:
* The failures you mention sound like `ws2812_effects_do_shift` isn't doing the right thing... ah, I see.  `prepare_shift` doesn't do the right thing with negative positions; d'oh.  Try this new commit on for size.

* `ws2812_effects_set_speed`, before this PR, looked like
```
static int ws2812_effects_set_speed(lua_State* L) {
  uint8_t speed = luaL_checkinteger(L, 1);
```
and it still uses `luaL_checkinteger`, so I don't *think* I changed any type requirements.  `ws2812_effects_set_delay` is unaltered.

HHHartmann:
Still doesn't work. 
When you force push your changes I cannot see what changed and I need some manual steps to update the sources.

You can also use consecuteive `print(effects_buffer:dump())` to see if it shifts.

nwf:
> Still doesn't work.

Please, be more specific?  I tested

```
> ws2812.init(ws2812.MODE_SINGLE)
> b = pixbuf.newBuffer(9, 3)
> ws2812_effects.init(b)
> ws2812_effects.set_mode("rainbow_cycle")
> ws2812_effects.start()
> =b
[(66,0,100),(100,0,66),(100,0,0),(100,66,0),(66,100,0),(0,100,0),(0,100,66),(0,66,100),(0,0,100)]
> =b
[(0,100,0),(0,100,66),(0,66,100),(0,0,100),(66,0,100),(100,0,66),(100,0,0),(100,66,0),(66,100,0)]
```

and that sure looks to me like it's trying to do the right thing?

> When you force push your changes I cannot see what changed and I need some manual steps to update the sources.

Sorry, habit arising from repositories that don't squash PRs and, in general, occasionally requisite to avoid merge commits as the upstream branch (i.e., `dev`) moves forward (as, indeed, it has over the lifetime of this PR).  This is one of the pain points of `git` and long-lived branches.  Something like pijul would work better, but that's quite new.

In this case, tho', you can see the relevant changes by doing `git diff 48875d4..51fa0a8 -- app/modules/pixbuf.* app/modules/ws2812_effects.c`.

tests/NTest_pixbuf.lua
```diff
@@ -0,0 +1,297 @@
...
+
+N.test('sub', function()
+    local buffer1 = pixbuf.newBuffer(4, 4)
+    local buffer2 = pixbuf.newBuffer(4, 4)
+    initBuffer(buffer1,7,8,9,12)
+    buffer1 = buffer1:sub(4,3)
+    ok(eq(buffer1:size(), 0), "sub empty")
+
+    buffer1 = pixbuf.newBuffer(4, 4)
+    initBuffer(buffer1,7,8,9,12)
```
HHHartmann, line 255: One line missing here: test fails 
```suggestion
    buffer2 = ws2812.newBuffer(2, 4)
    initBuffer(buffer1,7,8,9,12)
```

docs/modules/pixbuf.md
```diff
@@ -0,0 +1,325 @@
...
+`buffer1` is `buffer0` and `ilo` is 1 and `ihi` is -1).
+
+#### Returns
+`buffer0`
+
+#### Example
+```
+buffer:map(function(r,g,b) return g,r,b end) -- Change channel order
+buffer:map(function(r,g,b) return g,r,b end, nil, 2, 5) -- Change channel order for a subset of pixels
+outbuf:map(function(r,g,b) return b end, inbuf, 10, 20) -- Extract one channel for a subset of pixels
```
HHHartmann, line 323: wouldn't this fail because there are less return values than channels? (unless outbuf only has 1 channel)

docs/modules/pixbuf.md
```diff
@@ -0,0 +1,325 @@
...
+ - `ilo2` This is the start of the mapped range within `buffer2`.  Negative values can be used.
+   The default is 1.
+
+`buffer0` must have `ihi - ilo + 1` pixels (which is true of the defaults, when
+`buffer1` is `buffer0` and `ilo` is 1 and `ihi` is -1).
+
+#### Returns
+`buffer0`
+
+#### Example
```
HHHartmann, line 319: After looking at the code I understand, but from the examples and description I won't be able to use this with more than a function as param.
Especially the fact that the channels of the second buffer are also passed and can have a different number.

HHHartmann (changes requested):
Still would like to have C functions to adjust channel order.
Using :map will be around 8 to 10 times slower.

But will be left for another PR then

nwf:
Thanks.  Does 74a4051 address your concerns?

nwf:
> Still would like to have C functions to adjust channel order.
> Using :map will be around 8 to 10 times slower.

A pure-C byte-order-swizzling operation is probably straightforward... what do you imagine its API should be?  That is, how do you want to specify the mapping function?

app/lua53/linit.c
```diff
@@ -48,6 +48,9 @@ extern LROT_TABLE(co_funcs);
 extern LROT_TABLE(mathlib);
 extern LROT_TABLE(utf8);
 
+extern LROT_TABLE(pixbuf_map);
+extern int luaopen_pixbuf(lua_State *);
+
```
HHHartmann, line 53: Please also see Terrys comment https://github.com/nodemcu/nodemcu-firmware/pull/3230#issuecomment-691046239

Should we have this in one PR for Lua 5.1 and Lua 5.3?

See #3230

docs/modules/pixbuf.md
```diff
@@ -0,0 +1,342 @@
...
+```
+ws2812.write(buffer1 .. buffer2)
+```
+## pixbuf.buffer:map()
+Map a function across each pixel of one, or zip a function along two,
+pixbuf(s), storing into the buffer on which it is called.
+
+#### Syntax
+`buffer0:map(f, [buffer1], [ilo], [ihi], [buffer2, [ilo2]])`
```
HHHartmann, line 300: Could we use `start` and `end` instead of `ilo` and `ihi`

HHHartmann (changes requested):
> > Still would like to have C functions to adjust channel order.
> > Using :map will be around 8 to 10 times slower.
> 
> A pure-C byte-order-swizzling operation is probably straightforward... what do you imagine its API should be? That is, how do you want to specify the mapping function?

Maybe the map method could take defined values instead of a function to select a C operation.
buf:map(RGBtoGRB, [buf1[, ilo[, ihi]]])    -- SwapRG or ToGRB might also be a naming. Base would always be RGB. Others have to use their brain to find the correct setting.

This would be nice if you have strips with different channels connected together. But it would require a second buffer which uses memory. It is also a rather exotic scenario, which leads me to 
another solution. That would be to add the parameter to the write methods.
`ws2812.write(bug, SwapRG)

To be channel neutral we could also pass ints to determin the new position of the channels.
Either as separate values or as array.
{2,1,3} would be equivalent to the above RGBtoGRB. It would also allow for 3 and 4 channels

app/lua53/linit.c
nwf: I'll back the changes to this file out and we can deal with host-side deltas later.

HHHartmann (approved):

marcelstoer:
Sorry for coming back to this but I'm confused. 

I was expecting some "Attention! This requires the pixbuf module"-style hints in the documentation of the affected modules but found none. I see that they all (unconditionally) `#include "pixbuf.h"` even though we've got `//#define LUA_USE_MODULES_PIXBUF` in the `user_modules.h`. Then I realized we don't use this anywhere.

HHHartmann:
There are several modules whose define is never used. Expect four the link magic which decides what goes on the firmware. If I'm not mistaken we always build all modules. Conditional include only is used if the behavior would change. But indeed some hints might make sense.

nwf:
Ideally, such intermodule dependencies would be handled by Kconfig (#3130).  I'm rather buried at work right now, so may I ask someone else to update the docs in the interim?

HHHartmann:
Thought about it again. 
There should be #IFDEF around the usage of pixbuf.
The hardware modules can be used without as data can be presented as stirrings too. 
Only the effects need pixbuf and ws2812

marcelstoer:
> may I ask someone else to update the docs in the interim?

That was my initial intention when I realised those hints wrt pixbuf were missing. Then I noticed that pixbuf is always compiled into the firmware as `LUA_USE_MODULES_PIXBUF` isn't used to steer that (my conclusion).

Now I don't know what the real issue is here: missing docs or missing define guards.

HHHartmann:
The build will break if one of the other modules are enabled without LUA_USE_MODULES_PIXBUF.
So this has to be fixed. 
The documentation should mention that you need to enable pixbuf of you want to use it. Except for the effects module which requires pixbuf

marcelstoer:
I think we're talking past each other, sorry.

> The build will break if one of the other modules are enabled without LUA_USE_MODULES_PIXBUF.

It doesn't, that's my point 😉 I disabled pixbuf (i.e. use the default) but enabled apa102, ws2801, ws2812, and ws2812_effects as they import `pixbuf.h`. I was stunned when the build _didn't_ fail.

There's only a single occurrence of `LUA_USE_MODULES_PIXBUF` - in the `user_modules.h`. Doesn't that mean it's always compiled into the firmware? That would explain why my build didn't fail.

HHHartmann:
Yes, also checked and was surprised that it worked. It seems that the linker just pulls in the parts it needs. So I would guess that half the pixbuf module is linked. But haven't checked yet. EDIT: it seems that the whole module is pulled in.

The trick is done in app/Makefile and the NODEMCU_MODULE macro in app/include/module.h
The macro generates a symbol which is the pulled in the firmware if enabled. Don't really understand this in detail.

I think the linker then also pulls in the required modules.

So what do we want?
The modules apa102, ws2801, ws2812 should work even when pixbuf is disabled. Just without the pixbuf interface.
Use #ifdef in the code + notes in the doc that it will work but only with strings.
The module  ws2812_effects without pixbuf will not work however. So this should fail the build I think.

PR #3159, opened 2020-06-13
cron memory management with Lua table

nwf:
Refactor `cron` internals to let the Lua GC do more of the work -- that is, store the list of scheduled cron elements as a Lua table.  While here, permit an entry's `schedule()` function to reuse the existing schedule.

### DIFF
--- a/app/modules/cron.c
+++ b/app/modules/cron.c
@@ -12,6 +12,14 @@
 #include "rtc/rtctime.h"
 #include "mem.h"
 
+static const char *CRON_ENTRY_METATABLE = "cron.entry";
+
+#if 0
+# define cron_dbg(...) printf(__VA_ARGS__)
+#else
+# define cron_dbg(...)
+#endif
+
 struct cronent_desc {
   uint64_t min;  // Minutes repeat - bits 0-59
   uint32_t hour; // Hours repeat - bits 0-23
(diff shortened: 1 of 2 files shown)


app/modules/cron.c
```diff
@@ -86,12 +85,21 @@ static int lcron_parsedesc(lua_State *L, char *str, struct cronent_desc *desc) {
...
   char *strdesc = (char*)luaL_checkstring(L, 1);
-  void *newlist;
   luaL_checktype(L, 2, LUA_TFUNCTION);
-  // Parse description
-  struct cronent_desc desc;
-  lcron_parsedesc(L, strdesc, &desc);
-  // Allocate userdata
+
+  // Grab the table of all entries onto the stack
+  lua_rawgeti(L, LUA_REGISTRYINDEX, cronent_table_ref);
```
TerryE, line 91: `cronent_table_ref``contains a list of ud's and we scan for empty slots, scan for a specific one etc.  I just wonder if we would we better off using a keyed table with the ud as the key, and just let the table access functions do all this for us.

The comment isn't what the scan does, which is scan from the start of the table looking for the first null slot.

On a separate note, just in terms of efficiency scanning up a list for a null slot must terminate (possibly by throwing and EOM) so this scan could be more efficiently written:
```C
  while(lua_rawgeti(L, -1, ix) != LUA_TNIL) { lua_pop(L, 1); ix++; }
```

nwf: Whoop, yes, the comment's stale.

I thought about using the `ud` as both the key and value, but this was a straightforward bring-over of existing logic, AIUI array-like tables are smaller in memory, and I think we should drop table entirely and just equip each cron entry with a timer as part of fixing #3160.

marcelstoer:
CI build failure

```
cron.c:95:3: error: void value not ignored as it ought to be
   while(lua_rawgeti(L, -1, ix) != LUA_TNIL) {
```

nwf:
Ah, I think that's a 5.3 vs 5.1 thing.  Let's hold on this PR until @TerryE's done the requisite backporting?  Or I could go put the call to `lua_type` back.

Issue #3141, opened 2020-06-02
Indexing a function restarts the ESP instead of reporting an error


TerryE:
### Expected behaviour
Typing `return require[1]` one standard Lua (on a PC) generates an error:
```
> return require[1]
stdin:1: attempt to index global 'require' (a function value)
stack traceback:
	stdin:1: in main chunk
	[C]: ?
> 
```
NodeMCU Lua on the ESP should do the same.
### Actual behaviour
```
> return require[1]

 ets Jan  8 2013,rst cause:2, boot mode:(3,7)
 ...
NodeMCU 3.0.0.0 built on nodemcu-build.com provided by frightanic.com
	branch: dev
	commit: 8f85c9b3e4a606f1731051af45c39302799ddb28
	release: 
	release DTS: 202005281716
	SSL: false
	build type: float
	LFS: 0x20000 bytes total capacity
	modules: encoder,file,gdbstub,gpio,net,node,sjson,tmr,uart,wifi
 build 2020-05-29 13:01 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
> return node.bootreason()
2	2	28	1073790848	0	0	100	0
>  return '0x%x' % 10737908 48
0x4000bf80
> -- This is a S/W exception in the bootROM at 0xbf80 "A load referenced a page mapped with an attribute that does not permit loads"
```
### Test code and  NodeMCU startup banner
See above

### Hardware
Wemos D1 mini

TerryE:
I've checked LUA=53 and both luac.cross variants. This only occurs in the Lua51 environment for ESP firmware builds.

TerryE:
Fixed in #3163

PR #3163, opened 2020-06-13
Implement panic call handling for all modules

### LINKED ISSUE #3078: Error Handling in Lua

TerryE:
Fixes #3078.  (Also #3141 and @KT819GM Gitter comment at  Jun 12 15:04)

- [x] I have tested my contribution.

This is a fairly large patch to the modules plus the addition of a `node.setonerror()` function so that CBs throwing an error give a proper traceback.  Apps can also override the default action of rebooting after such an error.

Note that this PR is submitted for review at this stage.  I still have some documentation to add and any changes in the light of feedback.

### DIFF
--- a/app/lua/lnodemcu.h
+++ b/app/lua/lnodemcu.h
@@ -1,5 +1,6 @@
-/* Read-only tables for Lua */
-
+/*
+ * NodeMCU extensions to Lua 5.1 for readonly Flash memory support
+ */
 #ifndef lnodemcu_h
 #define lnodemcu_h
 
@@ -16,7 +17,8 @@
 #define LOCK_IN_SECTION(s) __attribute__((used,unused,section(".lua_" #s)))
 #endif
 
-/* Macros one can use to define rotable entries */
+/* Macros used to declare rotable entries */
+
 #define LRO_FUNCVAL(v)       {{.p = v}, LUA_TLIGHTFUNCTION}
 #define LRO_LUDATA(v)        {{.p = cast(void*,v)}, LUA_TLIGHTUSERDATA}
 #define LRO_NILVAL           {{.p = NULL}, LUA_TNIL}
@@ -25,7 +27,7 @@
 #define LRO_FLOATVAL(v)      LRO_NUMVAL(v)
 #define LRO_ROVAL(v)         {{.gc = cast(GCObject *, &(v ## _ROTable))}, LUA_TROTABLE}
 
-#define LROT_MARKED           0  //<<<<<<<<<<*** TBD *** >>>>>>>>>>>
+#define LROT_MARKED          0 //<<<<<<<<<<  *** TBD *** >>>>>>>>>>>
 
 #define LROT_FUNCENTRY(n,f)  {LRO_STRKEY(#n), LRO_FUNCVAL(f)},
 #define LROT_LUDENTRY(n,x)   {LRO_STRKEY(#n), LRO_LUDATA(x)},
@@ -38,9 +40,9 @@
 #define LROT_ENTRYREF(rt)    (rt ##_entries)
 #define LROT_TABLEREF(rt)    (&rt ##_ROTable)
 #define LROT_BEGIN(rt,mt,f)  LROT_TABLE(rt); \
-  static const ROTable_entry rt ## _entries[] = {
+  static ROTable_entry rt ## _entries[] = {
 #define LROT_ENTRIES_IN_SECTION(rt,s) \
-  static const ROTable_entry LOCK_IN_SECTION(s) rt ## _entries[] = {
+  static ROTable_entry LOCK_IN_SECTION(s) rt ## _entries[] = {
 #define LROT_END(rt,mt,f)    {NULL, LRO_NILVAL} }; \
   const ROTable rt ## _ROTable = { \
     (GCObject *)1, LUA_TROTABLE, LROT_MARKED, \
@@ -64,4 +66,9 @@
 #define LROT_MASK_GC_INDEX   (LROT_MASK_GC | LROT_MASK_INDEX)
 
 /* Maximum length of a rotable name and of a string key*/
+
+#ifdef LUA_CORE
+
 #endif
+#endif
+
--- a/app/lua53/lauxlib.c
+++ b/app/lua53/lauxlib.c
@@ -649,6 +649,28 @@ LUALIB_API void luaL_unref (lua_State *L, int t, int ref) {
   }
 }
 
+
+LUALIB_API void (luaL_reref) (lua_State *L, int t, int *ref) {
+  int reft;
+/*
+ * If the ref is positive and the entry in table t exists then
+ * overwrite the value otherwise fall through to luaL_ref()
+ */
+  if (ref) {
+    if (*ref >= 0) {
+      t = lua_absindex(L, t);
+      lua_rawgeti(L, t, *ref);
+      reft = lua_type(L, -1);
+      lua_pop(L, 1);
+      if (reft != LUA_TNIL) {
+        lua_rawseti(L, t, *ref);
+        return;
+      }
+    }
+    *ref = luaL_ref(L, t);
+  }
+}
+
 /* }====================================================== */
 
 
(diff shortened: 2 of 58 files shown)


TerryE:
@dnc40085 can you please review 6cfbad4.  Lua 5.3 does not support `LUA_GLOBALSINDEX` so I had to change your code to remove this.
@galjonsfigur ditto with the platform.c GPIO code.

TerryE:
This PR has 53 files changed so there's no way that I can sensibly cover them all.  Mode of these are the same edit pattern to module code to replaces a `lua_call(L, n, 0)` lines by  `luaL_pcallx(L, n, 0)` calls.  However these are only in the case where the call is in a SDK CB.  There are cases such as where you can pass helper routines to `sjson` and these should be allowed to throw.

Also not that the last Travis build has failed on the MS Windows build of luac.cross.  I'll examine the logs an fix this on my next push.

app/lua53/lnodemcu.c
```diff
@@ -579,6 +577,10 @@ LUAI_FUNC int  luaN_reload_reboot (lua_State *L) {
 #endif
   lua_settop(L, 1);
   lua_getglobal(L, "file");
+  if (lua_isnil(L, 2)) {
+    lua_pushstring(L, "No file system mounted");
```
nwf, line 581: Mounted?  Wouldn't this be more like "file module not compiled into firmware"?

Any particular reason this is returned rather than thrown?

app/modules/node.c
```diff
@@ -21,6 +21,39 @@
...
+  static os_timer_t restart_timer = {0};
+  /* Use Lua print to print the ToS */
+  lua_settop(L, 1);
+  lua_getglobal(L, "print");
+  lua_insert(L, 1);
+  lua_pcall(L, 1, 0, 0);
+  /* One first time through set automatic restart after 2s delay */
+  if (!restart_timer.timer_func) {
+    os_timer_setfn(&restart_timer, restart_callback, NULL);
+		os_timer_arm(&restart_timer, DELAY2SEC, 0);
```
nwf, line 40: Whitespace

app/lua/lnodemcu.h
```diff
@@ -38,9 +40,9 @@
 #define LROT_ENTRYREF(rt)    (rt ##_entries)
 #define LROT_TABLEREF(rt)    (&rt ##_ROTable)
 #define LROT_BEGIN(rt,mt,f)  LROT_TABLE(rt); \
-  static const ROTable_entry rt ## _entries[] = {
+  static ROTable_entry rt ## _entries[] = {
 #define LROT_ENTRIES_IN_SECTION(rt,s) \
-  static const ROTable_entry LOCK_IN_SECTION(s) rt ## _entries[] = {
+  static ROTable_entry LOCK_IN_SECTION(s) rt ## _entries[] = {
```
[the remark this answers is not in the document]
TerryE: 'cos the `ROTable_entry` typedef is already a `const struct`.

app/modules/node.c
TerryE: Noted. Will fix

app/lua53/lauxlib.c
```diff
@@ -649,6 +649,28 @@ LUALIB_API void luaL_unref (lua_State *L, int t, int ref) {
   }
 }
 
+
+LUALIB_API void (luaL_reref) (lua_State *L, int t, int *ref) {
```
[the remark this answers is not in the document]
TerryE: I am currently drafting a NodeMCU addendum to the Lua Reference Manual which will cover all this.

marcelstoer:
Terry, did you notice this failed the Windows CI build?

-> https://travis-ci.org/github/nodemcu/nodemcu-firmware/jobs/697961546

TerryE:
> Terry, did you notice this failed the Windows CI build?

Yup see my previous post :smile:
> Also not that the last Travis build has failed on the MS Windows build of luac.cross. I'll examine the logs and fix this on my next push.

app/modules/node.c
TerryE: @nwf,
I've just done a `git diff 06aa248f bd9b4253 | gedit -` (the patch for this PR) and did a global highlight for `\t`.  This whole tab issue is a PITA.  The Lua style uses an indent depth of 2 chars and always uses two spaces to per indent.  It does use tabs (with a default of 8 chars per stop)

However modules are all over the place.  Some use tab indents (and some assume a tab stop of 4) so it is extremely easy to get the visual alignment screwed.  My vote is that we should address this as a module-wide cleanup rather piecemeal one-offs.

TerryE:
OK, the issue with the Windows build was that I updated SPIFFS and the latest version removed an MSC compatibility fix which I've put back in, and it now compiles clean.

@marcelstoer @nwf I would prefer to merge this sooner rather later as it contains a really useful enhancement and I've got other stuff backed up.  Unless anyone objects, I will do this tomorrow.

app/modules/node.c
marcelstoer: >  Some use tab indents (and some assume a tab stop of 4) so it is extremely easy to get the visual alignment screwed. 

#1168, created over 4y ago.

TerryE: Yup.  Golden oldies that still bite.

app/lua/lnodemcu.h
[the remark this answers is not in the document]
TerryE: Note that whereas ` ROTable_entry` vectors are always RO, this is one `ROTable` declaration where this isn't the case and that is for the `ROM` `ROTable`.  This is because whilst the `const` vector is collated using linker magic, its size can't easily be determined at link time, and so `lua_createrotable()` is used during Lua initialisation to build this one `ROTable` struct in RAM which points to the  ROTable_entry` vector in flash.

danjohn:
`tmr.softwd()` seems to be unstopable now by no more allowing a negative argument.
Glitch or intented?

PR #3168, opened 2020-06-18
Makes lfs/_init.lua compatible with Lua 5.3

### LINKED ISSUE #3156: LFS under Lua 5.3
I'm sorry I don't have time to study the issue myself now. So maybe I am reporting something while the issue is on the other side of the keyboard. I guess this has some easy solution.

### Expected behavior
LFS example working under Lua 5.3

### Actual behavior
```Lua
node.flashindex("_init")()
Lua error: _init.lua:55: attempt to call a nil value (global 'getfenv')
stack traceback:
	_init.lua:55: in main chunk
	stdin:1: in main chunk
	[C]: in ?
	[C]: in ?
```

### Test code
Standard [LFS example](/nodemcu/nodemcu-firmware/tree/master/lua_examples/lfs) loaded in LFS.
```Lua
node.flashindex("_init")()
```
### NodeMCU startup banner
```
	branch: watertank
	commit: 78ab92f24021bb843e611f4daac0124649304f7f
	release: 2.0.0-master_20170202 +523
	release DTS: 202006112149
	SSL: false
	build type: float
	LFS: 0x10000 bytes total capacity
	modules: enduser_setup,file,gpio,http,net,node,rtcmem,rtctime,sjson,sntp,sun,tmr,uart,wifi
 build 2020-06-11 22:06 powered by Lua 5.3.5 on SDK 3.0.1-dev(fce080e)
```

### Hardware
ESP8266 ESP-12E

vsky279:
Fixes #3156.

This commit makes [lua_examples/lfs/_init.lua](/nodemcu/nodemcu-firmware/blob/master/lua_examples/lfs/_init.lua) compatible with Lua 5.3.

### DIFF
--- a/lua_examples/lfs/_init.lua
+++ b/lua_examples/lfs/_init.lua
@@ -8,10 +8,10 @@
   module related initialisaion in this. This example uses standard Lua features to
   simplify the LFS API.
 
-  The first section adds a 'LFS' table to _G and uses the __index metamethod to
-  resolve functions in the LFS, so you can execute the main function of module
-  'fred' by executing LFS.fred(params), etc. It also implements some standard
-  readonly properties:
+  For Lua 5.1, the first section adds a 'LFS' table to _G and uses the __index
+  metamethod to resolve functions in the LFS, so you can execute the main
+  function of module 'fred' by executing LFS.fred(params), etc.
+  It also implements some standard readonly properties:
 
   LFS._time    The Unix Timestamp when the luac.cross was executed.  This can be
                used as a version identifier.
@@ -24,36 +24,44 @@
                  print(table.concat(LFS._list,'\n'))
                gives you a single column listing of all modules in the LFS.
 
+   For Lua 5.3 LFS table is populated by the LFS implementation in C so this part
+   of the code is skipped.
 ---------------------------------------------------------------------------------]]
 
 local index = node.flashindex
+local G=_ENV or getfenv()
+local lfs_t
+if _VERSION == 'Lua 5.1' then
+    lfs_t = {
+    __index = function(_, name)
+        local fn_ut, ba, ma, size, modules = index(name)
+        if not ba then
+          return fn_ut
+        elseif name == '_time' then
+          return fn_ut
+        elseif name == '_config' then
+          local fs_ma, fs_size = file.fscfg()
+          return {lfs_base = ba, lfs_mapped = ma, lfs_size = size,
+                  fs_mapped = fs_ma, fs_size = fs_size}
+        elseif name == '_list' then
+          return modules
+        else
+          return nil
+        end
+      end,
 
-local lfs_t = {
-  __index = function(_, name)
-      local fn_ut, ba, ma, size, modules = index(name)
-      if not ba then
-        return fn_ut
-      elseif name == '_time' then
-        return fn_ut
-      elseif name == '_config' then
-        local fs_ma, fs_size = file.fscfg()
-        return {lfs_base = ba, lfs_mapped = ma, lfs_size = size,
-                fs_mapped = fs_ma, fs_size = fs_size}
-      elseif name == '_list' then
-        return modules
-      else
-        return nil
-      end
-    end,
+    __newindex = function(_, name, value) -- luacheck: no unused
+        error("LFS is readonly. Invalid write to LFS." .. name, 2)
+      end,
+    }
 
-  __newindex = function(_, name, value) -- luacheck: no unused
-      error("LFS is readonly. Invalid write to LFS." .. name, 2)
-    end,
-
-  }
-
-local G=getfenv()
-G.LFS = setmetatable(lfs_t,lfs_t)
+    setmetatable(lfs_t,lfs_t)
+    G.module       = nil    -- disable Lua 5.0 style modules to save RAM
+    package.seeall = nil
+else
+    lfs_t = node.LFS
+end
+G.LFS = lfs_t
 
 --[[-------------------------------------------------------------------------------
   The second section adds the LFS to the require searchlist, so that you can
@@ -67,17 +75,8 @@ G.LFS = setmetatable(lfs_t,lfs_t)
 ---------------------------------------------------------------------------------]]
 
 package.loaders[3] = function(module) -- loader_flash
-  local fn, ba = index(module)
-  return ba and "Module not in LFS" or fn
+  return lfs_t[module]
 end
-
---[[-------------------------------------------------------------------------------
-  You can add any other initialisation here, for example a couple of the globals
-  are never used, so setting them to nil saves a couple of global entries
----------------------------------------------------------------------------------]]
-
-G.module       = nil    -- disable Lua 5.0 style modules to save RAM
-package.seeall = nil
 
 --[[-------------------------------------------------------------------------------
   These replaces the builtins loadfile & dofile with ones which preferentially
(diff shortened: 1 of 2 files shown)


lua_examples/lfs/_init.lua
```diff
@@ -8,10 +8,10 @@
   module related initialisaion in this. This example uses standard Lua features to
   simplify the LFS API.
 
-  The first section adds a 'LFS' table to _G and uses the __index metamethod to
-  resolve functions in the LFS, so you can execute the main function of module
-  'fred' by executing LFS.fred(params), etc. It also implements some standard
-  readonly properties:
+  For Lua 5.1, the first section adds a 'LFS' table to _G and uses the __index 
+  metamethod to resolve functions in the LFS, so you can execute the main 
+  function of module 'fred' by executing LFS.fred(params), etc.
```
nwf, line 13: +"For Lua 5.3, equivalent functionality is provided natively by the firmware" ?

lua_examples/lfs/_init.lua
```diff
@@ -24,36 +24,45 @@
...
+      end,
   }
 
-local G=getfenv()
-G.LFS = setmetatable(lfs_t,lfs_t)
+  G.LFS = setmetatable(lfs_t,lfs_t) 
+  --[[-------------------------------------------------------------------------------
+    You can add any other initialisation here, for example a couple of the globals
+    are never used, so setting them to nil saves a couple of global entries
+  ---------------------------------------------------------------------------------]]
```
nwf, line 61: Split this comment in twain; this, being gated by the `_VERSION` check, is not where we want to encourage users to extend `_init.lua`, but these next lines still need some commentary.

lua_examples/lfs/_init.lua
```diff
@@ -96,6 +96,4 @@ G.dofile = function(n)
   local mod, ext = n:match("(.*)%.(l[uc]a?)");
   local fn, ba   = index(mod)
```
nwf, line 97: Here (and above), `index` will not refer to `node.flashindex()` because that `local` definition is now scoped within the `_VERSION` check.  I suspect you can rewrite these to use the `LFS` table that's now guaranteed to exist (**utterly untested, almost surely wrong**)...

```
local mod, ext = n:match("(.*)%.(l[uc]a?)");
if ext ~= 'lc' and ext ~= 'lua' then return df(n) else return LFS[mod] or df(n) end
```

vsky279: :rage: that was not a good optimization...

lua_examples/lfs/_init.lua
vsky279: "For Lua 5.3 LFS table is populated by the LFS implementation in C so this part of the code is skipped." ?

lua_examples/lfs/_init.lua
vsky279: Do you mean to have this section also in the `else` part of the `_VERSION` check?
There's no need to nil `G.module` and `package.seeall` under Lua 5.3.
Or do we just modify the comment that it applies to Lua 5.1 solely?

vsky279:
How do we solve the
```
lua_examples/lfs/_init.lua:31:9: accessing undefined variable _ENV
lua_examples/lfs/_init.lua:79:10: accessing undefined variable LFS
```
luacheck error?
Will there be 2 `luacheck_config.lua`? One for Lua 5.1 and Lua 5.3?

lua_examples/lfs/_init.lua
nwf: I just mean that the encouragement "You can add any other initialisation here" is now out of place.

lua_examples/lfs/_init.lua
```diff
@@ -96,6 +96,4 @@ G.dofile = function(n)
...
   if ba or (ext ~= 'lc' and ext ~= 'lua') then return df(n) else return fn() end
-end
-
-
+end
```
marcelstoer, line 99: Missing a newline here.

nwf:
This will need further revision in light of #3176.  Blocked until that lands, I think.

vsky279:
Merging now would mean to have the example compatible with Lua 5.3. Another rework should follow #3176 imho.

nwf:
That seems like an OK plan to me. @TerryE?

TerryE:
> That seems like an OK plan to me. @TerryE?

There are +/-'s here.  Anyone that needs `_init.lua` support for 5.3 can use this code, but I am trying to get the PR which addresses #3175, #3176 and #3180 out in the next few days, and merging this now means that I will have to do a rebase before I can create this PR which is going to add to this timeline. 

BTW, it's my wife's birthday today, so I don;t want to be locked away in my study today. :smile:

lua_examples/lfs/_init.lua
```diff
@@ -24,36 +24,40 @@
...
-        return nil
-      end
-    end,
-
-  __newindex = function(_, name, value) -- luacheck: no unused
-      error("LFS is readonly. Invalid write to LFS." .. name, 2)
-    end,
-
+local G=_ENV or getfenv()
+if _VERSION == 'Lua 5.1' then
```
TerryE, line 33: We need to set `_G.LFS` to `node.LFS` on the `else` branch,  otherwise `LFS` will not resolve for the require code @ 74.  Though IMO, it would be better to hoist local `lfs_t` which is then set by the `if _VERSION` code as per 5.1 or 5.3 (the 5.1 path includes the `setmetatable()`) and move the `G.LFS = lfs_t` to the common path.  Then L74 can refer to the close upval `lfs_t`.

vsky279: This was before `node.LFS` was implemented.
I like the idea to use upvalue in `package.loaders[3]` handler.

tools/luacheck_config.lua
```diff
@@ -430,6 +430,16 @@ stds.nodemcu_libs = {
...
           }
+        },
+        LFS = {
+          read_only = true, 
+          fields = {
+            config = empty,
+            get = empty,
+            list = empty,
+            reload = empty,
+            tiem = empty
```
nwf, line 441: `time`?

tools/luacheck_config.lua
```diff
@@ -430,6 +430,16 @@ stds.nodemcu_libs = {
             MEDIUM_PRIORITY = empty,
             HIGH_PRIORITY = empty
           }
+        },
+        LFS = {
+          read_only = true, 
+          fields = {
```
nwf, line 436: Not really a question for this PR, but since you're in the neighborhood: should we suggest to advanced developers how to add entries here?  People using their own LFS might want their code to still be `luacheck`-clean?

nwf:
Sure, what's the worst that happens. ;)

tools/luacheck_config.lua
galjonsfigur: @nwf I think it would be a good idea to make some kind of documentation on how to use `luacheck` on NodeMCU code. I've written `luacheck_config_helper.lua` script when I was adding `luacheck` support but I haven't documented it anywhere besides some comments in the script itself. Should this be added to GitHub Wiki,  Extension Developer FAQ or somewhere else?

TerryE: > I think it would be a good idea to make some kind of documentation on how to use `luacheck` on NodeMCU code.

I am more of the view: what's wrong with RTFM?  If anyone wants to read up on how to use luacehck then it comes with a lot of good documentation.  May a reference to it in our FAQs.

Issue #3165, opened 2020-06-14
Additional documentation for Lua 5.3


TerryE:
We have quite a few gaps in our documentation set that moving from Lua 5.1 to Lua 5.3 underlines.  We really need to plug these before we can even consider moving the Lua 5.3 from Beta to being formally recommended.  These include:
 
-  The [Lua 5.3 Reference Manual]( https://www.lua.org/manual/5.3/) (**LRM**) is a pretty concise yet formal and complete reference for the Lua language and its implementation.  (It's still around 7,500 lines long).  We don't have any parallel formal definition of how NodeMCU Lua varies this LRM.  IMO, we need one and the easiest thing to do is to mirror the same style and structure.  I've started drafting a NodeMCU LRM.  [**This gist**](https://gist.github.com/TerryE/9834d0821bd03da12e7573154486ceaf) is a rough cut work in progress, and which I hope to finish tomorrow.

This plus the standard module documentation is the minimal set that should provide any developer with what is needed to write Lua code for the ESPs and develop C libraries.  That being said, I have been going through the LRM in detail for what must be over 5 tears and yet I am still discovering subtleties that I have missed on previous readings.

-  Our ReadTheDocs online material contains a lot of Lua 5.1 build-specific content. We need to go through this and either remove the specifics or put in a comparison table of 5.3 vs 5.1 changes.

-  Lua 5.3 is really quite big change so we could do with some detailed release notes which emphasise what has changed.
---
Lastly we have a longer term documentation objective that we need to make progress on.  I don't think that it is practical to consider this as a precursor to formally recommending our 5.3 engine.

-  The **LRM** is supplemented by a more informal Programming in Lua [**PiL**]. The 1st edition is available free [online](https://www.lua.org/pil/).  The forth edition is aligned to Lua 5.3 and available in  paperback and [eBook](https://store.feistyduck.com/products/programming-in-lua-fourth-edition-ebook) and this goes far more into techniques and worked examples.  This is a lot more developer friendly that the LRM.  Again, I think that we need to consider how we provide the additional context of how to use NodeMCU Lua, but this is a non-trivial task for a professional author -- which I am not.

I suggest in the first instance that we try to provide this material more as a community effort within the NodeMCU wiki as a collection of specific articles.  Maybe when they are sufficiently mature, all or a subset of this material could be promoted to our ReadTheDocs online material.

Thoughts? Reactions?

KT819GM:
[...]

Issue #3152, opened 2020-06-08
Emergency GC in Lua 5.3


TerryE:
### Background

Lua51 includes the eLua emergency GC, plus the various rarely used EGC tuning parameters. The default setting is `node.egc.ALWAYS` which triggers a full GC before every memory allocation so the VM spends maybe 90% of its time doing full GC sweeps.  Standard Lua 5.3 has adopted the eLua EGC but without the EGC tuning parameters.   See #2783 and #3078 as well.

### Proposed ECG implementation

NodeMCU will extend the Lua EGC with the functional equivalent of the Lua51 `ON_MEM_LIMIT` setting with a negative parameter, that is to trigger the EGC with less than a preset amount of free heap is left. Comparing heap needed to heap available is a very cheap operation as this value is maintained in a RAM memory control block by the SDK/ROM memory allocator and the `system_get_free_heap_size()` is just a getter to provide read access to this.

Adopting a lazy approach to GC results in the runtime spending far less time in the GC and a typical Lua application might run perhaps 5× faster. The risk here is that many libraries (e.g. LwIP and  MBEDTLS) make direct calls to the SDK allocator and so if there is a burst allocation demand from one of these libraries then the application can fail with an `E:M` error where it would otherwise have succeeded is the LuaGC had run with an "always GC before allocation" strategy.

Tidying up this miscellany of ways of accessing the memory allocation is going to be a total pain and not easily resourced.  However since we already use a define chain to map all of these variants back to the core `pvPortAllocXXXX()` there is nothing to stop us adding a separate allocator intercept so we can adopt a dual stage lazy GC strategy:

-  **Lua allocations**: any allocation which would leave less than `G(L)->gcmemfreeboard`  available triggers a full GC before attempting the allocation.
-  **All applications**: any allocation which fails triggers a full GC and retry before retuning a `null` allocation.

The potential issue with this is that a full GC sweep might get triggered during an allocation request in time sensitive code, but in general the various stack such as LwIP do the sensible thing and move any allocation requests forward of time critical code.  Also the potential hiccup of a GC sweep is a far more preferable course than restarting after memory exhaustion.

TerryE:
@nwf @HHHartmann @pjsg, you three are the most likely committers to eyeball this.  Feedback and discusssion welcome. :smile:

nwf:
[...]

PR #3171, opened 2020-06-20
Improved startup performance and measurement tools.

### LINKED ISSUE #3154: Reducing startup timing (dcc module)
I'm using the dcc module (from @vsky279 ) and I'm having problems getting the nodemcu platform to boot fast enough (I think) to support service mode writes. Does it work for you?

The issue is that there is only a limited amount of time between when the service mode programmer powers up the rails before the write commands are issued and the rails powered down. [Reads work fine (mostly)]

I'm intending to try setting the cpu speed high in the user_init function, getting rid of the startup banner, and using LFS (as I suspect that is faster than loading a `.lc` file). With lua53, the gc mode is already set sensibly. Are there other speedups that anybody can think of?

pjsg:
Fixes #3154 (partially)

This adds three different startup options that allow various tradeoffs during boot. In my environment, this leads to boot only needing 60% of the time that it did before. Of course, stuff like the delay of SPIFFS mounting just shuffles the time around, but it can be useful in special cases.


app/lua53/lua.c
```diff
@@ -109,6 +109,9 @@ static int docall (lua_State *L, int narg) {
 
 #ifndef DISABLE_STARTUP_BANNER
 static void print_version (lua_State *L) {
+  if (platform_rcr_get_startup_option() & STARTUP_OPTION_NO_BANNER) {
```
nwf, line 112: Maybe slightly better to just avoid calling `print_version` in the first place?

app/modules/node.c
```diff
@@ -851,6 +902,9 @@ LROT_BEGIN(node, NULL, 0)
   LROT_FUNCENTRY( compile, node_compile )
   LROT_NUMENTRY( CPU80MHZ, CPU80MHZ )
   LROT_NUMENTRY( CPU160MHZ, CPU160MHZ )
+  LROT_NUMENTRY( OPTION_NO_BANNER, STARTUP_OPTION_NO_BANNER )
```
nwf, line 905: Maybe `START_*` instead of `OPTION_*`?

app/platform/platform.h
```diff
@@ -382,4 +388,26 @@ bool platform_post(uint8 prio, platform_task_handle_t h, platform_task_param_t p
...
+} platform_count_entry_t;
+
+typedef struct {
+  int used;
+  platform_count_entry_t entries[32];
+} platform_startup_counts_t;
+
+extern platform_startup_counts_t platform_startup_counts;
+
+#define PLATFORM_STARTUP_COUNT_ENTRIES (sizeof(platform_startup_counts.entries) / sizeof(platform_startup_counts.entries[0]))
```
nwf, line 404: Split large lines with `\` escapes, here and below?

app/user/user_main.c
```diff
@@ -121,6 +121,15 @@ extern void _ResetHandler(void);
...
+    STARTUP_COUNT; 
+    int startup_option = platform_rcr_get_startup_option();
+
+    if (startup_option & STARTUP_OPTION_160MHZ) {
+      REG_SET_BIT(0x3ff00014, BIT(0));
+      ets_update_cpu_frequency(SYS_CPU_160MHZ);
+    }
+    int no_banner = startup_option & STARTUP_OPTION_NO_BANNER;
+
 #ifdef LUA_USE_MODULES_RTCTIME
```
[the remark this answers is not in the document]
pjsg: This is an interesting question -- I'm not sure how the existing switch to 160MHZ from node.setcpufreq works with respect to rtctime....

TerryE:
@pjsg, it is valuable seeing what can be achieved here and I feel that we generally need to move forward. However, I would really prefer that we finish the #3154 discussion or in fact open a proper **Startup Options** issue and agree scope and approach before we raise the PR in its final form.  Until we've done this, I really see this as a "do not merge yet" PR.

app/modules/node.c
```diff
@@ -58,6 +58,7 @@ static int node_setonerror( lua_State* L ) {
 }
 
 // Lua: startupcommand(string)
+// The lua.startup({command="string"}) should be used instead
```
nwf, line 61: Add a deprecation warning?

app/platform/platform.h
```diff
@@ -347,13 +347,19 @@ void* platform_print_deprecation_note( const char *msg, const char *time_frame);
...
 #define PLATFORM_RCR_INITSTR   0x5
+#define PLATFORM_RCR_STARTUP_OPTION   0x6
 #define PLATFORM_RCR_FREE      0xFF
 typedef union {
     uint32_t hdr;
     struct { uint8_t len,id; };
 } platform_rcr_t;
 
+#define STARTUP_OPTION_NO_BANNER        (1 << 0)
+#define STARTUP_OPTION_160MHZ           (1 << 1)
```
nwf, line 358: Since this lives nominally in our "HAL" layer (of sorts...), maybe `STARTUP_OPTION_CPU_FREQ_MAX` or something?

app/platform/platform.h
```diff
@@ -347,13 +347,19 @@ void* platform_print_deprecation_note( const char *msg, const char *time_frame);
...
+#define STARTUP_OPTION_DELAY_MOUNT      (1 << 2)
```
nwf, line 359: I might just be missing it, but I don't see the `LROT` entries I expect to see for these to be exported to Lua?

app/modules/node.c
pjsg: I add a message as a comment and also updated the documentation

app/platform/platform.h
pjsg: They aren't there -- the interface was changed as a result of the discussion, so you now say 

```
node.startup({banner=false})
```

pjsg: I could add constants for node.STARTUP_BANNER = "banner" but I'm not sure that that is really necessary.

docs/modules/node.md
```diff
@@ -564,6 +564,70 @@ node.startupcommand("=if LFS.init then LFS.init() else dofile('init.lua') end")
 ```
 
 
+## node.startupcounts()
+
```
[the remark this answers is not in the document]
pjsg: Done

app/user/user_main.c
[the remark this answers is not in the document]
pjsg: I've shuffled this around as you recommended.

app/platform/platform.h
nwf: No, indeed, that's not necessary.  Thank you for reminding me why I don't see the LROT entries. :)

pjsg:
The startup counts stuff is not compiled in by default -- I agree that most people wont want it. I built the sane interface to it because I like typing `=sjson.encode(node.startupcounts())` into esplorer! (`sjson.encode` is my goto way of displaying table structures)

app/platform/platform.c
```diff
@@ -985,6 +985,14 @@ uint32_t platform_rcr_read (uint8_t rec_id, void **rec) {
     return ~0;
 }
 
+uint32_t platform_rcr_get_startup_option() {
+  uint32_t *option_p;
```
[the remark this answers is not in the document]
pjsg: Indeed -- I'll make that change.

app/modules/node.c
```diff
@@ -631,6 +720,33 @@ static int node_random (lua_State *L) {
...
+    STARTUP_ENTRY(lineno);
+  }
+  lua_createtable(L, platform_startup_counts.used, 0);
+  for (int i = 0; i < platform_startup_counts.used; i++) {
+    const platform_count_entry_t *p = &platform_startup_counts.entries[i];
+
+    lua_createtable(L, 0, 3);
+    lua_pushliteral(L, "name");
+    lua_pushstring(L, p->name);
+    lua_settable(L, -3);
```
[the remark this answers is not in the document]
pjsg: Because I was a doofus.

app/modules/node.c
```diff
@@ -63,6 +67,91 @@ static int node_startupcommand( lua_State* L ) {
...
+        if (strcmp(key, "delay_mount") == 0) {
+          int enable = lua_toboolean(L, -1);
+          option = (option & ~STARTUP_OPTION_DELAY_MOUNT) | (enable ? STARTUP_OPTION_DELAY_MOUNT : 0);
+        }
+      }
+      lua_pop(L, 1);
+    }
+
+    if (has_entries) {
+      platform_rcr_write(PLATFORM_RCR_STARTUP_OPTION, &option, sizeof(option));
```
HHHartmann, line 119: This write is called even if no changes happened. Maybe compare to the original value.

app/modules/node.c
```diff
@@ -63,6 +67,91 @@ static int node_startupcommand( lua_State* L ) {
...
+    int has_entries = 0;
+    lua_pushnil(L);
+    while (lua_next(L, 1) != 0) {
+      if (lua_isstring(L, -2)) {
+        const char *key = lua_tostring(L, -2);
+        has_entries++;
+
+        if (strcmp(key, "command") == 0) {
+          size_t l, lrcr;
+          const char *cmd = luaL_checklstring(L, -1, &l);
```
HHHartmann, line 93: Maybe only write on changed  data. 
People might just have the setup call in their startup routine rather than just calling it once

app/modules/node.c
pjsg: Yes -- it turns out that the platform_rcr_write already checks fo 'no change' and returns quickly in that case. I didn't feel it was worth the increased code size to shave a few microseconds off this call (by duplicating the check here)

app/modules/node.c
pjsg: As above...

TerryE:
@pjsg Philip, we seem to be moving to consensus on this one.  I do face one challenge here and that is that I am almost done on preparing a PR which going to sweep to know issues with Lua 5.3 and cleaning up the 5.1 / 5.3 build compatibility in terms of APIs.  This encompasses #3152, #3165, #3165 (part), #3175, #3176 and #3180.  I know that @marcelstoer prefers one issue = one PR, but in this case the secondary Issues where usually thrown up as sub-points to the main tidy-up.  As you can imagine my PR also changes most if not all of the files that you have updated, and having to rebase against `dev` is going to be a real PITA for me if this goes in first.

So can I ask that we hold off merging this PR in until after mine lands?  I am quite willing to do the knock-on rebasing work rather than impose this on you, but it's just that its going to be less work for me this way.  Thanks Terry

pjsg:
That is fine.

app/modules/node.c
HHHartmann: Ok. Wasn't aware of that.

TerryE:
> That is fine.

Given that we've agreed to do this after #3193, I will temporarily add the blocked label.

marcelstoer:
As #3193 has since been merged I removed the "blocked" label.

docs/modules/node.md
```diff
@@ -607,6 +607,76 @@ node.startupcommand("=if LFS.init then LFS.init() else dofile('init.lua') end")
...
+Query the performance of system startup.
+Note that this is only available if
```
marcelstoer, line 613: Please wrap this with an admonition like we e.g. do for `wifi.startsmart()` https://nodemcu.readthedocs.io/en/latest/modules/wifi/#wifistartsmart. I doubt we are consistent as to the admonition _type_ (note, important, etc.).

pjsg: Fixed.

pjsg:
@HHHartmann @nwf @TerryE I think that I have addressed all of your comments....

nwf (approved):

docs/modules/node.md
```diff
@@ -607,6 +607,78 @@ node.startupcommand("=if LFS.init then LFS.init() else dofile('init.lua') end")
...
+
+#### Parameters
+
+If the argument is omitted, then no change is made to the current set of startup options. If the argument is the empty table `{}` then all options are
+reset to their default values.
+
+- `table` one or more options:
+    - `banner` - set to true or false to indicate whether the startup banner should be displayed or not.
+    - `frequency` - set to node.CPU80MHZ or node.CPU160MHZ to indicate the initial CPU speed.
+    - `delay_mount` - set to true or false to indicate whether the SPIFFS filesystem mount is delayed it is first needed or not.
```
HHHartmann, line 669:
```suggestion
    - `delay_mount` - set to true or false to indicate whether the SPIFFS filesystem mount is delayed until it is first needed or not.
```

docs/modules/node.md
```diff
@@ -607,6 +607,78 @@ node.startupcommand("=if LFS.init then LFS.init() else dofile('init.lua') end")
...
+#### Syntax
+`node.startup([{table}])`
+
+#### Parameters
+
+If the argument is omitted, then no change is made to the current set of startup options. If the argument is the empty table `{}` then all options are
+reset to their default values.
+
+- `table` one or more options:
+    - `banner` - set to true or false to indicate whether the startup banner should be displayed or not.
```
HHHartmann, line 667: Maybe indicate which values are the dafault values.

HHHartmann, line 667:
```suggestion
    - `banner` - set to true or false to indicate whether the startup banner should be displayed or not. Default: true
```

docs/modules/node.md
```diff
@@ -607,6 +607,78 @@ node.startupcommand("=if LFS.init then LFS.init() else dofile('init.lua') end")
 ```
 
 
+## node.startupcounts()
+
+Query the performance of system startup.
+
+!!! Important
+
+    This function is only available if the firmware is built with `PLATFORM_STARTUP_COUNT` defined.
```
HHHartmann, line 616: A reference to user_config.h will be helpful to the unexperienced user.

HHHartmann (changes requested):
Sorry for the late review. the rest LGTM.

docs/modules/node.md
pjsg: Done

PR #3172, opened 2020-06-20
Bump toolchain to 20190731.0

marcelstoer:
@jmattsson do you see any reason why we should _not_ bump the toolchain to latest?

Btw, I had also bumped it for `dev-esp32` but reverted that change as CI failed: https://travis-ci.org/github/nodemcu/nodemcu-firmware/builds/700433258


jmattsson:
The only reason to _not_ go with a newer toolchain is if it breaks something. Other than that I'm all for it. No point using old tools unnecessarily.

Issue #3140, opened 2020-05-31
ws2812 driver doesn't wait for UART to drain


nwf:
### Expected behavior

`ws2812.write()` should completely send all the data before returning control to Lua, so that subsequent events do not overlap the bitstream.  Alternatively, the documentation should include a suitable delay for `tmr.delay`; the answer appears (to my relatively naive eye) to be a full FIFO (at 128 bytes) being sent at 3200000 baud, which takes 300 microseconds?

### Actual behavior

`ws2812.write()` returns after the tail of the bitstream is queued to the UART FIFO, and does not stick around to wait for the FIFO to drain out the pin.

### Test code

With GPIO 3 AND-gating the ws2812 bitstream (from GPIO 4), the following sequence will not transmit the entirety of `ledfb` before the stream is interrupted:

```Lua
ws2812.init(ws2812.MODE_SINGLE)
gpio.write(3,gpio.HIGH)
ws2812.write(ledfb)
gpio.write(3,gpio.LOW)
```

### Firmware revision

Verified experimentally under, and then again by inspection of, 5135ee441aee86129c37ce78dfca02053000cc2a .

### Additional commentary

Perhaps the ws2812 driver itself should know how to do the GPIO jostling for this kind of AND-gating, as an optional feature.  It's very handy because otherwise the ESP's startup spew out its UARTs gets interpreted by the ws2812 drivers and causes annoying visual artifacts.

Perhaps @Alkorin, being the author of this (very much nicer than bitbanging) UART transmission technique, has opinions?

pjsg:
I wouldn't want that wait behavior to be the default.. if you are just
using the ws2812 library then you probably want to get the benefit of the
overlapping... Maybe add a wait option to the write method...

nwf:
I'd settle for the overlap being documented, really.  A `sync` parameter also seems like a grand option.

pjsg:
Curiously enough, I find myself wanting something very similar with some other enhancements.

My use case is that I need to be able to attach a ws2812 to an esp8266 in such a way that it *does not* illuminate during boot (this is for power consumption reasons). As far as I can tell, this means that either I have to use another gpio to gate D2, or maybe I can switch UART0 to use D8 directly. In the latter case, doing the ws2812.write synchronously would allow me to switch back to the regular TX pin for other output. 

Also, in the case that I have to gate the output signal, I probably want to be able to control whether D2 output is inverted or not. 

Any other thoughts?

[...]

PR #3177, opened 2020-06-24
uart: expose fifo depth counters

nwf:

Useful for #3140, for example.

### DIFF
--- a/app/modules/uart.c
+++ b/app/modules/uart.c
@@ -142,6 +142,27 @@ static int l_uart_write( lua_State* L )
   return 0;
 }
 
+#define DIR_RX 0
+#define DIR_TX 1
+
+static int l_uart_fifodepth( lua_State* L )
+{
+  int id  = luaL_optinteger( L, 1, 0 );
+  if ((id != 0) && (id != 1))
+    return luaL_argerror(L, 1, "Bad UART id; must be 0 or 1");
+
+  int dir = luaL_optinteger( L, 2, 0 );
+  if ((dir != DIR_RX) && (dir != DIR_TX))
+    return luaL_argerror(L, 2, "Bad direction; must be DIR_RX or DIR_TX");
+
+  int reg = READ_PERI_REG(UART_STATUS(id));
+  int rsh = reg >> ((dir == DIR_RX) ? UART_RXFIFO_CNT_S : UART_TXFIFO_CNT_S);
+  int rm  = rsh &  ((dir == DIR_RX) ? UART_RXFIFO_CNT   : UART_TXFIFO_CNT  );
+
+  lua_pushinteger(L, rm);
+  return 1;
+}
+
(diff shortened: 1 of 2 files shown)


app/modules/uart.c
```diff
@@ -142,13 +142,27 @@ static int l_uart_write( lua_State* L )
   return 0;
 }
 
+static int l_uart_fifodepth( lua_State* L )
+{
+  int id  = luaL_optinteger( L, 1, 0 );
+  int dir = luaL_optinteger( L, 2, 0 );
+
```
pjsg, line 149: Probably ought to limit id to 0 or 1 (or what is the limit on esp32?)

app/modules/uart.c
```diff
@@ -142,13 +142,27 @@ static int l_uart_write( lua_State* L )
...
+  int reg = READ_PERI_REG(UART_STATUS(id));
+  int rsh = reg >> (dir ? UART_TXFIFO_CNT_S : UART_RXFIFO_CNT_S);
+  int rm  = rsh &  (dir ? UART_TXFIFO_CNT   : UART_RXFIFO_CNT  );
+
```
pjsg, line 153: I wonder when this count goes to zero... Is it when the last byte is loaded into the UART shift register to be transmitted, or is it when the last byte has actually been transmitted? I don't think that there is much that can be done about this..... But it would nice to have a way of knowing that you could shut off the transmit pin ...

nwf: The ESP8266 technical reference is not clear, saying only "Data has been sent will be automatically cleared, at the same time transmission FIFO will be more of a vacancy."

This makes it sound like a byte is removed from the FIFO only after it has been completely shifted out the pin, which would be nice, but it's more likely that it's just been moved into a shift register and is still being tx'd when the fifo depth is observed to be zero.  Tricky; hard to measure without writing some C, I think.  At 300 baud there might be enough time to get into the TX FIFO empty IRQ and toggle a GPIO to see what the waveform looks like on a 'scope...

Speaking of, should we expose the UART FIFO IRQs through `uart.on()`?

docs/modules/uart.md
```diff
@@ -145,3 +145,17 @@ Write string or byte to the UART.
...
+
+Report the depth, in bytes, of TX or RX hardware queues associated with the
+UART.
+
+#### Syntax
+`uart.fifodepth(id, dir)`
+
+#### Parameters
+- `id` UART id (0 or 1).
+- `dir` 0 for the RX FIFO, 1 for TX FIFO.
```
HHHartmann, line 158: Could we have a label here? 
uart.TX and uart.RX

marcelstoer:
@HHHartmann the changes you requested were introduced. Ready to approve?

HHHartmann (approved):
LGTM

PR #3184, opened 2020-06-27
Fixed conflict in documentation and example for method newChannel()

kiritnarain:
It now correctly uses ledc.newChannel() and saves the return value instead of invoking ledc.setup().

Fixes no existing issue.

The previous example was incorrect because ledc.setup() does not exist and the example given for a method should use the method.


marcelstoer:
Ideally we would also get the approval of the original author @larsstenberg.

Issue #3180, opened 2020-06-24
RFC: Tidy up the Lua C API / library code split and interface


TerryE:
I first raised this point in #3176, but decided that it is sufficiently important not to get buried in the depths of another issue, so I have moved the comment here.

There all sorts of undocumented conventions involved with the Lua source file structure, (and which we sometimes don't follow).  For example:
-  All Lua directory C files are either classified as either core or library (set by defining `LUA_CORE` _or_ `LUA_LIB` respectfully, but not both).
-  The Lua core source files implement the core Lua functions and runtime.  The access other core functions by including the appropriate `lxxxx.h` header files.  Whilst core functions implement the public API, they rarely call these API functions internally and so _don't_  include the `lauxlib.h` header.  Note that almost API functions take a `lua_lock()` around the function to make reentrency safe, and in fact the Lua test configuration will barf if locks are nested. 
-  The Lua library source files only use the public API and therefore only include the `lua.h` and `lauxlib.h` headers, so no backdoors into Lua core files and data structures.
-  This is the reason for the `lstring.c`+`lstrlib.c` and  `ltable.c`+`ltablib.c`, etc. split. One does the core stuff and the other implements the Lua-callable library functions. 

The standard Lua code sticks to this split without exception, however some of the NodeMCU changes and enhancements don't obey these rules, though since I've got to grips with the Lua internals and the valid reasons for this approach, I have been backing out our abuses in the `app/lua` code.  The 'app/lua53` implementation is a lot cleaner in this regard, since I controlled all of the changes from standard.  

Doing this makes library support for both platforms a lot easier as the migration path is only defined for the public interface and there are no guarantees in respect of the core implementation, and I have therefore done the bulk of this cleanup as part of the Lua 5.3 groundwork.

In terms of the implementation of pure core NodeMCU functionality such as LFS, `app/lua/lnodemcu.c` implements the additions to the core API and `app/modules/node.c` provides the library components.  And as I discuss in #3097, this why this module doesn't belong in `modules` folder. 

Some of the tweaks in this PR will take us further along this clean-up journey.

So for example structures like the like `lua_State` are anonymous `struct *` in the public API and the internals are hidden from library functions.  Where access needs to be given to fields in the global state, for example then we should be using getters and setters implemented in `lstate.c`.  If a lirary module needs to include `lobject.h`, for example then we are doing something wrong.

TerryE:
Here are the offending exceptions:
[...]

Issue #3176, opened 2020-06-23
RFC: Tidy up node.flashindex interface


TerryE:
### Background
-  The 5.1 LFS interface loads a compiled Lua index(name) function which is essentially the following, and with the `lua_lfsindex()` function a wrapper around this.
```Lua
  local name = ...
  if name == 'mod1' then return mod1;
  if name == 'mod2' then return mod2;
  if name == 'mod3' then return mod3;
  -- ... 
  return {'mod1', 'mod2', 'mod3' --[[, ... ]]}
```
-  The 5.3 includes a ROTable which is of the form `{'mod1' = mod1 , 'mod2' = mod2,  --[[, ... ]]}`with the `lua_lfsindex()` function essentially going a `lua_rawget()` from this ROTable.

The switch in the underlying implementation was largely driven by feedback from a few users who has a _lot_ of functions in LFS and were having scaling issues with the Lua access function.  This switch is done now and I am happy with the new approach.  

The original `node.flashindex()` overloading of the function lookup and configuration fetch was a bit of a botch IMO, which is why I added the LFS pseudo-table in the `_init` example.

### Proposed tweak to this interface

We I would like to do is to tweak this in a way that tidies up the implementation but without introducing a break for standard implementations:

- `node.flashindex(name)` returns the function LFS.name if present and _`nil` otherwise_ 
- `node.flashindex(nil)` returns an extended: build timestamp,  the LFS flash and mapped addresses, the LFS length, and an array of the valid module names in the LFS. 

The difference is that `node.flashindex(invalidname)` would return `nil` instead of the extended info.

The reason for this is that in 5.3, the array of the valid module names has for be done at runtime on each call by effectively doing `for k in pairs(LFSnametable) do res[#res+1] = k end` so we should only do this when the programmer really wants the list, and hence requiring a `nil` parameter is a sensible choice here.

### Suggested change

At the moment the LFS size is LFS size _used_ if the LFS is loaded and the partition size if LFS is not.  Hence there isn't a simple way to see a used / free combination.  I think that adding the extra field so that programmers can easily do this calc would be a good idea ?

Doe anyone else have any suggestions whilst we are at this?

nwf:
For the return-`nil` change, :+1:, we'll just have to chase down places where we use the 2nd return from `node.flashindex()` instead of checking whether the first is `nil` (as in the example `_init.lua`).

[...]

Issue #3175, opened 2020-06-21
Anyone fancy playing with the Lua GC?


TerryE:
I've been looking at how the Lua (5.3) GC works on our builds.  At the moment it is pretty standard, but I want to tailor it to more our requirements.  The standard GC has two tuning settings that you can tweak with [collectgarbage()](https://www.lua.org/manual/5.3/manual.html#pdf-collectgarbage):
-  **`"setpause"`**. The current default setting is 110, which means that once the GC completes a cycle it waits until the memory use has grown by 10% before unpausing.
-  **`"setstepmul"`**.   The current default setting is 200, but I suspect that we should have this higher say 300 or 400.

The patch below adds a little instrumentation so you can see what is going on, but this could be tweaked.  Some general comments:

-  If you play, then you will see that allocating a GCObject (one that is GC collectable, such as a `string`, `proto`, `table`, `userdata`, etc.) causes a sweep check and if past the pause threshold then a (partial) sweep is made through the GCObject list to find unreferenced ones.
-  Note that freeing doesn't seem to kick off sweeps, but seems to be relying and a stable mix of allocs and frees so that the allocs trigger the sweeps.
-  It seems to me that the idea of having an exponential spacing of the pauses is really arse about face.  Unlike PCs which have poodles of RAM, we typically only have ~40-44Kb.  Instead of pausing longer between sweeps as RAM use grows, we should start off doing a sweep every 4Kb or some step, and _drop_ this interval (not increase it) as we approach running out of than so by the time we are down to the last 4Kb or so we need to switch into the Lua 5.1 which is to run `collectgabage()` every allocation.

Also note that a full GC is splil into sweeps and the full GC repeats these until no more RAM can be freed.  If you are doing GC at the end of every action function, say, then a slightly lightweight approach might be to do a `collectgarbage('step',4)` instead.

Anyway if anyone else wants to have a play and get up to speed with how this all works it would really help being able to have an informed discussion about how to tweak this for our default setup.

Thanks Terry

-----

```diff
 /*
 ** performs a basic GC step when collector is running
 */
+#ifdef LUA_USE_ESP8266 /*DEBUG*/
+extern void dbg_printf(const char *fmt, ...);
+#define CCOUNT_REG ({ int32_t r; asm volatile("rsr %0, ccount" : "=r"(r)); r;})
+#else                  /*DEBUG*/   
+#define dbg_printf(...)
+#define CCOUNT_REG 0
[...]

Issue #3170, opened 2020-06-19
Stripdebug feature


TerryE:
### Background 
I added the [LCD patch](../blob/master/docs/lcd.md) back in 2015 when I first started optimising the Lua engine for NodeMCU. The issue that this was addressing was that leaving the debug metadata in compiled code pretty much tripled the RAM required to store a given compiled function. The way to get around this was to use `node.compile()` which stripped out all debug info, but this resulted in error trackback just returning lots of "?".

LCD had three broad strategies:
-  A packed run-length encoding was adopted for line number information and this dropped the overhead for this by roughly 15×.  OK, looking up line numbers became an O(n) operation but this was only done during error tracebacks anyway.
-  A 3-level stripping scheme was introduced: no debug info; just line number info; all debug info.
-  The `node.stripdebug()` allowed you to set a default strip level during initial load.

I have yet to add this to Lua5.3 and am now doing this as part of the tidy-up activities, but since I am standardising the APIs and making this consistent across all version, I wanted to expose what I am doing here for feedback.

### Proposed Functionality

I now want to add this into Lua 5.3 to avoid loss of functionality.

-  The public `lua.h` interface is implemented by `lua_stripdebug(L, level)`.  This pops a value off the Lua stack This can be function or a positive integer (which means the function at that Lua call stack level, e.g. 1 is the parent calling function).  If this is a Lua function not in LFS, then this and any nested functions within it are stripped of debug information to the specified level.  If the value is `nil` then this sets the default strip level if `level` in the range 0 to 2 (-1 means don't change), and the default level is returned.
-  `node.stripdebug(level, [func])` is the Lua callable implementation of this, though levels here adopt the Lua convention and are 1,2,3 instead of 0,1,2 but that's for historic reasons and not worth a compatibility break.  So for example:
```Lua
  node.stripdebug(3, someFunc)  -- strip all debug info from someFunc
  node.stripdebug(2, 1)  -- strip all but line number info from the function calling this.
  node.stripdebug(1)  -- set the default to strip no debug info on load
  local strip = node.stripdebug() -- a `nil` level means "don't change" so this returns the default strip level
```
 
Also the restart default is `strip=2`, that is upvalue and local metadata are removed during load.  This mainly impacts the `debug` interface (which wasn't in our Lua 5.1 implementation anyway).  If developers need this info to retained then they need to include `node.stripdebug(1)` in their startup.

**Q**: do I add `false` / `true` for none and all stripping?

[...]

PR #3193, opened 2020-07-04
Lua 5.1 / 5.3 alignment and document

TerryE:
This is a big PR encompassing #3152, #3165 (part), #3175, #3176 and #3180.

- [x] This PR is for the `dev` branch rather than for `release`.
- [x] I have ~~thoroughly~~ tested my contribution.

### Overview of PR

This aims to sweep up most of the loose ends following the Alpha release of Lua 5.3 and as such is a _big_ PR.  Sorry but all of the interdependencies make this a difficult one to untangle.  So this encompasses:

-  #3152 **Emergency GC in Lua 5.3**.
-  #3165 **Additional documentation for Lua 5.3**.  See new `docs/` files for the NodeMCU LRM and the LFS Whitepaper.  This second is probably far too detailed for mode readers and needs culling, IMO.
-  #3170 **Implement Stripdebug feature for Lua 5.3**
-  #3176 **RFC: Tidy up node.flashindex interface**
-  #3180 **RFC: Tidy up the Lua C API / library code split and interface**.  To see how this works see for example `node.c:node_lfs_func()`

### Still TBD
-  I am still tuning the GC stuff. I will add this as a separate commit.
-  Review / implementation of `luac.cross` strip debug levels

### Caveat
Because of the size of this, I am expecting extra rework so lease don't merge until we've had this review cycle.

### DIFF
--- a/app/lua/lapi.c
+++ b/app/lua/lapi.c
@@ -152,6 +152,16 @@ LUA_API lua_State *lua_newthread (lua_State *L) {
 */
 
 
+/*
+** convert an acceptable stack index into an absolute index
+*/
+LUA_API int lua_absindex (lua_State *L, int idx) {
+  return (idx > 0 || idx <= LUA_REGISTRYINDEX)
+         ? idx
+         : cast_int(L->top - L->base) - 1 + idx;
+}
+
+
 LUA_API int lua_gettop (lua_State *L) {
   return cast_int(L->top - L->base);
 }
@@ -241,7 +251,7 @@ LUA_API void lua_pushvalue (lua_State *L, int idx) {
 
 LUA_API int lua_type (lua_State *L, int idx) {
   StkId o = index2adr(L, idx);
-  return (o == luaO_nilobject) ? LUA_TNONE : basettype(o);
+  return (o == luaO_nilobject) ? LUA_TNONE : ttnov(o);
 }
 
 
@@ -253,7 +263,7 @@ LUA_API int lua_fulltype (lua_State *L, int idx) {
 
 LUA_API const char *lua_typename (lua_State *L, int t) {
   UNUSED(L);
-  return (t == LUA_TNONE) ? "no value" : luaT_typenames[t];
+  return (t == LUA_TNONE || t >= LUA_NUMTAGS) ? "no value" : luaT_typenames[t];
 }
 
 
@@ -290,30 +300,23 @@ LUA_API int lua_rawequal (lua_State *L, int index1, int index2) {
 }
 
 
-LUA_API int lua_equal (lua_State *L, int index1, int index2) {
+LUA_API int lua_compare (lua_State *L, int index1, int index2, int op) {
   StkId o1, o2;
-  int i;
+  int i = 0;
   lua_lock(L);  /* may call tag method */
   o1 = index2adr(L, index1);
   o2 = index2adr(L, index2);
-  i = (o1 == luaO_nilobject || o2 == luaO_nilobject) ? 0 : equalobj(L, o1, o2);
+  if (o1 != luaO_nilobject && o2 != luaO_nilobject) {
+    switch (op) {
+      case LUA_OPEQ: i = luaV_equalval(L, o1, o2); break;
+      case LUA_OPLT: i = luaV_lessthan(L, o1, o2); break;
+      case LUA_OPLE: i = luaV_lessequal(L, o1, o2); break;
+      default: api_check(L, 0);
+    }
+  }
   lua_unlock(L);
   return i;
 }
-
-
-LUA_API int lua_lessthan (lua_State *L, int index1, int index2) {
-  StkId o1, o2;
-  int i;
-  lua_lock(L);  /* may call tag method */
-  o1 = index2adr(L, index1);
-  o2 = index2adr(L, index2);
-  i = (o1 == luaO_nilobject || o2 == luaO_nilobject) ? 0
-       : luaV_lessthan(L, o1, o2);
-  lua_unlock(L);
-  return i;
-}
-
 
 
 LUA_API lua_Number lua_tonumber (lua_State *L, int idx) {
@@ -366,7 +369,7 @@ LUA_API const char *lua_tolstring (lua_State *L, int idx, size_t *len) {
 
 LUA_API size_t lua_objlen (lua_State *L, int idx) {
   StkId o = index2adr(L, idx);
-  switch (basettype(o)) {
+  switch (ttnov(o)) {
     case LUA_TSTRING: return tsvalue(o)->len;
     case LUA_TUSERDATA: return uvalue(o)->len;
     case LUA_TTABLE: return luaH_getn(hvalue(o));
@@ -552,16 +555,17 @@ LUA_API int lua_pushthread (lua_State *L) {
 */
 
 
-LUA_API void lua_gettable (lua_State *L, int idx) {
+LUA_API int lua_gettable (lua_State *L, int idx) {
   StkId t;
   lua_lock(L);
   t = index2adr(L, idx);
   api_checkvalidindex(L, t);
   luaV_gettable(L, t, L->top - 1, L->top - 1);
   lua_unlock(L);
+  return ttnov(L->top - 1);
 }
 
-LUA_API void lua_getfield (lua_State *L, int idx, const char *k) {
+LUA_API int lua_getfield (lua_State *L, int idx, const char *k) {
   StkId t;
   TValue key;
   lua_lock(L);
@@ -573,20 +577,36 @@ LUA_API void lua_getfield (lua_State *L, int idx, const char *k) {
   luaV_gettable(L, t, &key, L->top);
   api_incr_top(L);
   lua_unlock(L);
+  return ttnov(L->top - 1);
 }
 
+LUA_API int lua_geti (lua_State *L, int idx, int n) {
+  StkId t;
+  TValue key;
+  lua_lock(L);
+  t = index2adr(L, idx);
+  api_checkvalidindex(L, t);
+  fixedstack(L);
+  setnvalue(&key, n);
+  unfixedstack(L);
+  luaV_gettable(L, t, &key, L->top);
+  api_incr_top(L);
+  lua_unlock(L);
+  return ttnov(L->top - 1);
+}
 
-LUA_API void lua_rawget (lua_State *L, int idx) {
+LUA_API int lua_rawget (lua_State *L, int idx) {
   StkId t;
   lua_lock(L);
   t = index2adr(L, idx);
   api_check(L, ttistable(t));
   setobj2s(L, L->top - 1, luaH_get(hvalue(t), L->top - 1));
   lua_unlock(L);
+  return ttnov(L->top - 1);
 }
 
 
-LUA_API void lua_rawgeti (lua_State *L, int idx, int n) {
+LUA_API int lua_rawgeti (lua_State *L, int idx, int n) {
   StkId o;
   lua_lock(L);
   o = index2adr(L, idx);
@@ -594,6 +614,21 @@ LUA_API void lua_rawgeti (lua_State *L, int idx, int n) {
   setobj2s(L, L->top, luaH_getnum(hvalue(o), n));
   api_incr_top(L);
   lua_unlock(L);
+  return ttnov(L->top - 1);
+}
+
+
+LUA_API int lua_rawgetp (lua_State *L, int idx, const void *p) {
+  StkId t;
+  TValue k;
+  lua_lock(L);
+  t = index2adr(L, idx);
+  api_check(L, ttistable(t));
+  setpvalue(&k, cast(void *, p));
+  setobj2s(L, L->top, luaH_get(hvalue(t), &k));
+  api_incr_top(L);
+  lua_unlock(L);
+  return ttnov(L->top - 1);
 }
 
 
@@ -612,7 +647,7 @@ LUA_API int lua_getmetatable (lua_State *L, int objindex) {
   int res;
   lua_lock(L);
   obj = index2adr(L, objindex);
-  switch (basettype(obj)) {
+  switch (ttnov(obj)) {
     case LUA_TTABLE:
       mt = hvalue(obj)->metatable;
       break;
@@ -620,7 +655,7 @@ LUA_API int lua_getmetatable (lua_State *L, int objindex) {
       mt = uvalue(obj)->metatable;
       break;
     default:
-      mt = G(L)->mt[basettype(obj)];
+      mt = G(L)->mt[ttnov(obj)];
       break;
   }
   if (mt == NULL)
@@ -720,6 +755,23 @@ LUA_API void lua_rawseti (lua_State *L, int idx, int n) {
 }
 
 
+LUA_API void lua_rawsetp (lua_State *L, int idx, const void *p) {
+  StkId o;
+  TValue k;
+  lua_lock(L);
+  api_checknelems(L, 1);
+  o = index2adr(L, idx);
+  api_check(L, ttistable(o));
+  fixedstack(L);
+  setpvalue(&k, cast(void *, p))
+  setobj2t(L, luaH_set(L, hvalue(o), &k), L->top-1);
+  unfixedstack(L);
+  luaC_barriert(L, hvalue(o), L->top-1);
+  L->top--;
+  lua_unlock(L);
+}
+
+
 LUA_API int lua_setmetatable (lua_State *L, int objindex) {
   TValue *obj;
   Table *mt;
@@ -750,7 +802,7 @@ LUA_API int lua_setmetatable (lua_State *L, int objindex) {
       break;
     }
     default: {
-      G(L)->mt[basettype(obj)] = mt;
+      G(L)->mt[ttnov(obj)] = mt;
       break;
     }
   }
@@ -901,7 +953,7 @@ LUA_API int lua_load (lua_State *L, lua_Reader reader, void *data,
 }
 
 
-LUA_API int lua_dumpEx (lua_State *L, lua_Writer writer, void *data, int stripping) {
+LUA_API int lua_dump (lua_State *L, lua_Writer writer, void *data, int stripping) {
   int status;
   TValue *o;
   lua_lock(L);
@@ -913,6 +965,30 @@ LUA_API int lua_dumpEx (lua_State *L, lua_Writer writer, void *data, int strippi
     status = 1;
   lua_unlock(L);
   return status;
+}
+
+
+LUA_API int lua_stripdebug (lua_State *L, int stripping){
+  TValue *o = L->top - 1; 
+  Proto *p = NULL;
+  int res = -1;
+  lua_lock(L);
+  api_checknelems(L, 1);
+  if (isLfunction(o)) {
+    p = clvalue(o)->l.p;
+    if (p && !isLFSobject(p) && (unsigned) stripping < 3 ) {
+    // found a valid proto to strip
+      res = luaG_stripdebug(L, p, stripping, 1);
+    }
+  } else if (ttisnil(L->top - 1)) {
+    // get or set the default strip level
+    if ((unsigned) stripping < 3)
+      G(L)->stripdefault = stripping;
+    res = G(L)->stripdefault;
+  }
+  L->top--;
+  lua_unlock(L);
+  return res;
 }
 
 
@@ -1136,15 +1212,15 @@ LUA_API const char *lua_setupvalue (lua_State *L, int funcindex, int n) {
   return name;
 }
 
-LUA_API void lua_setegcmode( lua_State *L, int mode, int limit) {
+LUA_API void lua_setegcmode ( lua_State *L, int mode, int limit) {
   G(L)->egcmode = mode;
   G(L)->memlimit = limit;
 }
 
-LUA_API void legc_set_mode(lua_State *L, int mode, int limit) {
-   global_State *g = G(L);
-
-   g->egcmode = mode;
-   g->memlimit = limit;
+LUA_API void lua_getegcinfo (lua_State *L, int *totals) {
+  if (!totals) {
+    totals[0] = G(L)->totalbytes;
+    totals[1] = G(L)->estimate;
+  }
 }
 
--- a/app/lua/lflash.c
+++ b/app/lua/lflash.c
@@ -194,20 +194,19 @@ static int loadLFS (lua_State *L);
 static int loadLFSgc (lua_State *L);
 static void procFirstPass (void);
 
-/* lua_lfsreload() and lua_lfsindex() are exported via lua.h */
+/* luaL_lfsreload() is exported via lauxlib.h */
 
 /*
  * Library function called by node.flashreload(filename).
  */
-LUALIB_API int lua_lfsreload (lua_State *L) {
+LUALIB_API void luaL_lfsreload (lua_State *L) {
   const char *fn = lua_tostring(L, 1), *msg = "";
   int status;
 
   if (G(L)->LFSsize == 0) {
     lua_pushstring(L, "No LFS partition allocated");
-    return 1;
+    return;
   }
-
 
  /*
   * Do a protected call of loadLFS.
@@ -237,7 +236,7 @@ LUALIB_API int lua_lfsreload (lua_State *L) {
     lua_cpcall(L, &loadLFSgc, NULL);
     lua_settop(L, 0);
     lua_pushstring(L, msg);
-    return 1;
+    return;
   }
 
   if (status == 0) {
@@ -257,60 +256,20 @@ LUALIB_API int lua_lfsreload (lua_State *L) {
   NODE_ERR(msg);
 
   while (1) {}  // Force WDT as the ROM software_reset() doesn't seem to work
-  return 0;
 }
 
 
-/*
- * If the arg is a valid LFS module name then return the LClosure
- * pointing to it. Otherwise return:
- *  -  The Unix time that the LFS was built
- *  -  The base address and length of the LFS
- *  -  An array of the module names in the LFS
- */
-LUAI_FUNC int lua_lfsindex (lua_State *L) {
-  int n = lua_gettop(L);
-
-  /* Return nil + the LFS base address if the LFS size > 0 and it isn't loaded */
-  if (!(G(L)->ROpvmain)) {
-    lua_settop(L, 0);
-    lua_pushnil(L);
-    if (G(L)->LFSsize) {
-      lua_pushinteger(L, (lua_Integer) flashAddr);
-      lua_pushinteger(L, flashAddrPhys);
-      lua_pushinteger(L, G(L)->LFSsize);
-      return 4;
-    } else {
-      return 1;
-    }
-  }
-
-  /* Push the LClosure of the LFS index function */
-  Closure *cl = luaF_newLclosure(L, 0, hvalue(gt(L)));
-  cl->l.p = G(L)->ROpvmain;
-  lua_settop(L, n+1);
-  setclvalue(L, L->top-1, cl);
-
-  /* Move it infront of the arguments and call the index function */
-  lua_insert(L, 1);
-  lua_call(L, n, LUA_MULTRET);
-
-  /* Return it if the response if a single value (the function) */
-  if (lua_gettop(L) == 1)
-    return 1;
-
-  lua_assert(lua_gettop(L) == 2);
-
-  /* Otherwise add the base address of the LFS, and its size bewteen the */
-  /* Unix time and the module list, then return all 4 params. */
-  lua_pushinteger(L, (lua_Integer) flashAddr);
-  lua_insert(L, 2);
-  lua_pushinteger(L, flashAddrPhys);
-  lua_insert(L, 3);
-  lua_pushinteger(L, cast(FlashHeader *, flashAddr)->flash_size);
-  lua_insert(L, 4);
-  return 5;
+LUA_API void lua_getlfsconfig (lua_State *L, int *config) {
+  if (!config)
+    return;
+  config[0] = (int) flashAddr;                   /* LFS region mapped address */
+  config[1] = flashAddrPhys;                 /* LFS region base flash address */
+  config[2] = G(L)->LFSsize;                        /* LFS region actual size */
+  config[3] = cast(FlashHeader *, flashAddr)->flash_size;  /* LFS region used */
+  config[4] = 0;                                       /* Not used in Lua 5.1 */
 }
+
+
 /* =====================================================================================
  * The following routines use my uzlib which was based on pfalcon's inflate and
  * deflate routines.  The standard NodeMCU make also makes two host tools uz_zip
@@ -483,7 +442,7 @@ void procSecondPass (void) {
 }
 
 /*
- * loadLFS)() is protected called from luaN_reload_reboot so that it can recover
+ * loadLFS)() is protected called from luaL_lfsreload() so that it can recover
  * from out of memory and other thrown errors.  loadLFSgc() GCs any resources.
  */
 static int loadLFS (lua_State *L) {
--- a/app/lua/lflash.h
+++ b/app/lua/lflash.h
@@ -43,8 +43,5 @@ typedef struct {
 } FlashHeader;
 
 LUAI_FUNC void luaN_init (lua_State *L);
-LUAI_FUNC int  luaN_flashSetup (lua_State *L);
-LUAI_FUNC int  luaN_reload_reboot (lua_State *L);
-LUAI_FUNC int  luaN_index (lua_State *L);
 #endif
 
--- a/app/lua/lnodemcu.c
+++ b/app/lua/lnodemcu.c
@@ -19,8 +19,10 @@
 #include "ltable.h"
 #include "ltm.h"
 #include "lnodemcu.h"
-#include "platform.h"
 
+//== NodeMCU lauxlib.h API extensions ========================================//
+#ifdef LUA_USE_ESP
+#include "platform.h"
 /*
 ** Error Reporting Task.  We can't pass a string parameter to the error reporter
 ** directly through the task interface the call is wrapped in a C closure with
@@ -99,8 +101,7 @@ static void do_task (platform_task_param_t task_fn_ref, uint8_t prio) {
 /*
 ** Schedule a Lua function for task execution
 */
-#include "lstate.h" /*DEBUG*/
-LUALIB_API int luaL_posttask( lua_State* L, int prio ) {            // [-1, +0, -]
+LUALIB_API int luaL_posttask( lua_State* L, int prio ) {          // [-1, +0, -]
   if (!task_handle)
     task_handle = platform_task_get_id(do_task);
 
@@ -114,6 +115,98 @@ LUALIB_API int luaL_posttask( lua_State* L, int prio ) {            // [-1, +0, 
     luaL_error(L, "Task queue overflow. Task not posted");
   }
   return task_fn_ref;
+}
+#else
+LUALIB_API int luaL_posttask( lua_State* L, int prio ) { 
+  return 0;
+} /* Dummy stub on host */
+#endif
+
+#ifdef LUA_USE_ESP
+/*
+ * Return an LFS function
+ */
+LUALIB_API int luaL_pushlfsmodule (lua_State *L) {
+  if (lua_pushlfsindex(L) == LUA_TNIL) {
+    lua_remove(L,-2);  /* dump the name to balalnce the stack */
+    return 0;          /* return nil if LFS not loaded */
+  }
+  lua_insert(L, -2);
+  lua_call(L, 1, 1);
+  if (!lua_isfunction(L, -1)) {
+    lua_pop(L, 1);
+    lua_pushnil(L);  /* replace DTS by nil */
+  }
+  return 1;
+}
+
+/*
+ * Return an array of functions in LFS
+ */
+LUALIB_API int luaL_pushlfsmodules (lua_State *L) {
+  if (lua_pushlfsindex(L) == LUA_TNIL)
+    return 0;              /* return nil if LFS not loaded */
+  lua_call(L, 0, 2);
+  lua_remove(L, -2);     /* remove DTS leaving array */
+  return 1;
+
+}
+
+/*
+ * Return the Unix timestamp of the LFS image creation
+ */
+LUALIB_API int luaL_pushlfsdts (lua_State *L) {
+  if (lua_pushlfsindex(L) == LUA_TNIL)
+    return 0;              /* return nil if LFS not loaded */
+  lua_call(L, 0, 1);
+  return 1;
+}
+#endif
+
+//== NodeMCU lua.h API extensions ============================================//
+
+LUA_API int lua_freeheap (void) {
+#ifdef LUA_USE_HOST
+  return MAX_INT;
+#else
+  return (int) platform_freeheap();
+#endif
+}
+
+#define api_incr_top(L)   {api_check(L, L->top < L->ci->top); L->top++;}
+LUA_API int lua_pushstringsarray(lua_State *L, int opt) {
+  stringtable *tb = NULL;
+  int i, j;
+
+  lua_lock(L);
+  if (opt == 0)
+    tb = &G(L)->strt;
+#ifdef LUA_USE_ESP
+  else if (opt == 1 && G(L)->ROstrt.hash)
+    tb = &G(L)->ROstrt;
+#endif
+  if (tb == NULL) {
+    setnilvalue(L->top);
+    api_incr_top(L);
+    lua_unlock(L);
+    return 0;
+  }
+
+  Table *t = luaH_new(L, tb->nuse, 0);
+  sethvalue(L, L->top, t);
+  api_incr_top(L);
+  luaC_checkGC(L);
+  lua_unlock(L);
+
+  for (i = 0, j = 1; i < tb->size; i++) {
+    GCObject *o;
+    for(o = tb->hash[i]; o; o = o->gch.next) {
+      TValue v;
+      setsvalue(L, &v, o); 
+      setobj2s(L, luaH_setnum(L, hvalue(L->top-1), j++), &v);
+    }
+  }
+  return 1;
 }
 
 LUA_API void lua_createrotable (lua_State *L, ROTable *t, const ROTable_entry *e, ROTable *mt) {
@@ -142,3 +235,22 @@ LUA_API void lua_createrotable (lua_State *L, ROTable *t, const ROTable_entry *e
   t->entry     = cast(ROTable_entry *, e);
 }
 
+#ifdef LUA_USE_ESP
+#include "lfunc.h"
+
+/* Push the LClosure of the LFS index function */
+LUA_API int lua_pushlfsindex (lua_State *L) {
+  lua_lock(L);
+  Proto *p = G(L)->ROpvmain;
+  if (p) {
+    Closure *cl = luaF_newLclosure(L, 0, hvalue(gt(L)));
+    cl->l.p = p;
+    setclvalue(L, L->top, cl);
+  } else {
+    setnilvalue(L->top);
+  }
+  api_incr_top(L);
+  lua_unlock(L);
+  return p ? LUA_TFUNCTION : LUA_TNIL;
+}
+#endif
--- a/app/lua/lstrlib.c
+++ b/app/lua/lstrlib.c
@@ -14,7 +14,6 @@
 
 #include "lauxlib.h"
 #include "lualib.h"
-#include "lnodemcu.h"
 
 /* macro to `unsign' a character */
 #define uchar(c)        ((unsigned char)(c))
@@ -141,15 +140,23 @@ static int writer (lua_State *L, const void* b, size_t size, void* B) {
 
 static int str_dump (lua_State *L) {
   luaL_Buffer b;
+  int strip, tstrip = lua_type(L, 2);
+  if (tstrip == LUA_TBOOLEAN) {
+    strip = lua_toboolean(L, 2) ? 2 : 0;
+  } else if (tstrip == LUA_TNONE || tstrip == LUA_TNIL) {
+    strip = -1;  /* This tells lua_dump to use the global strip default */
+  } else { 
+    strip = lua_tointeger(L, 2);
+    luaL_argcheck(L, (unsigned)(strip) < 3, 2, "strip out of range");
+  }
   luaL_checktype(L, 1, LUA_TFUNCTION);
   lua_settop(L, 1);
   luaL_buffinit(L,&b);
-  if (lua_dump(L, writer, &b) != 0)
-    luaL_error(L, "unable to dump given function");
+  if (lua_dump(L, writer, &b, strip) != 0)
+    return luaL_error(L, "unable to dump given function");
   luaL_pushresult(&b);
   return 1;
 }
-
 
 
 /*
--- a/app/lua53/lnodemcu.h
+++ b/app/lua53/lnodemcu.h
@@ -40,12 +40,12 @@
 #define LROT_ENTRYREF(rt)    (rt ##_entries)
 #define LROT_TABLEREF(rt)    (&rt ##_ROTable)
 #define LROT_BEGIN(rt,mt,f)  LROT_TABLE(rt); \
-  static const ROTable_entry rt ## _entries[] = {
+  static ROTable_entry rt ## _entries[] = {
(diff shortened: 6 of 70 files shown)


app/lua53/lnodemcu.h
```diff
@@ -40,12 +40,12 @@
 #define LROT_ENTRYREF(rt)    (rt ##_entries)
 #define LROT_TABLEREF(rt)    (&rt ##_ROTable)
 #define LROT_BEGIN(rt,mt,f)  LROT_TABLE(rt); \
-  static const ROTable_entry rt ## _entries[] = {
+  static ROTable_entry rt ## _entries[] = {
```
nwf, line 43: Not `const`?

nwf:
I'm afraid I have time only for a cursory inspection at the moment, but I've pulled my `dev-active` up atop this PR and have successfully run several iterations of my test harness.  It seems as though this PR also corrects some LFS bug I was half-heartedly hunting, as I am now able to load sizable LFSes (40576 bytes at the moment) that previously would result in the module being unable to find `_init`.

I am occasionally seeing #3096, I think, as well as the occasional invalid instruction panic that I haven't yet debugged, but **neither of these are regressions that should hold up this PR**.

TerryE:
I've got at least one more commit to add:
-  updates to `nodemcu-lrm.md` following review comments
-  updates to `node.c` and `node.md` following resolution of #3176 and #3200
-  updates to `lua_examples/lfs` following #3200
-  possible updates to documentation into page depending on @marcelstoer feedback.

We are really blocked on this until I get some of this feedback / input.

TerryE:
I had hoped to get some feedback on the new documentation that I've added, but alas not to be.  This PR is still blocked pending other committer input as per my previous comment.  I can't update the `node.md` documentation until we have agreement on the exact details of the `node.LFS` implementation.

marcelstoer:
> I had hoped to get some feedback on the new documentation that I've added

I had planned to make some time for that (late) tonight.

TerryE:
> I had planned to make some time for that (late) tonight.

Thanks Marcel.  Much appreciated.  Unlike the rest of you guys, I am now a gentleman of leisure / independent means, so I don't have to worry about the day job.  :smile:

docs/modules/node.md
```diff
@@ -184,8 +184,8 @@ Returns the function reference for a function in the [LFS (Lua Flash Store)](../
 
 #### Returns
 -  In the case where the LFS in not loaded, `node.flashindex` evaluates to `nil`, followed by the flash mapped base addresss of the LFS, its flash offset, and the size of the LFS.
--  If the LFS is loaded and the function is called with the name of a valid module in the LFS, then the function is returned in the same way the `load()` and the other Lua load functions do.
--  Otherwise an extended info list is returned: the Unix time of the LFS build, the flash and mapped base addresses of the LFS and its current length, and an array of the valid module names in the LFS.
+-  If the LFS is loaded and the `modulename` is a string that is the name of a valid module in the LFS, then the function is returned in the same way the `load()` and the other Lua load functions do; if there isn't a matching module then `nil` is returned.
+-  If the parameter is omitted or `nil`then an extended info list is returned: the Unix time of the LFS build, the flash and mapped base addresses of the LFS and its current length, and an array of the valid module names in the LFS.
```
marcelstoer, line 188:
```suggestion
-  If the parameter is omitted or `nil` then an extended info list is returned: the Unix time of the LFS build, the flash and mapped base addresses of the LFS and its current length, and an array of the valid module names in the LFS.
```

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+-  The [Lua 5.1 Reference Manual]( https://www.lua.org/manual/5.1/) and
+-  The [Lua 5.3 Reference Manual]( https://www.lua.org/manual/5.3/) (**LRM**)
+
+Developers using the NodeMCU environment should familiarise themselves with the 5.3 LRM.
+
+The Project provides a wide range of standard library modules written in both C and Lua to support many ESP hardware modules and chips, and these are documented in separation sections in our [online documentation](https://nodemcu.readthedocs.io/).
+
+This reference supplements LRM content and module documentation by focusing on the differences between NodeMCU Lua and standard Lua 5.3 in use.
+
+One of our goals in introducing Lua 5.3 support was to maintain the continuity of our existing C modules by ensuring that they can be successfully compiled and executed in both the Lua 5.1 and 5.3 environments.  This goal was achieved by combination of:
```
marcelstoer, line 16:
```suggestion
One of our goals in introducing Lua 5.3 support was to maintain the continuity of our existing C modules by ensuring that they can be successfully compiled and executed in both the Lua 5.1 and 5.3 environments.  This goal was achieved by a combination of:
```

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+-  enabling relevant compatibility options for standard Lua libraries;
+-  regressing some Lua 5.3 API enhancements back into our Lua 5.1 implementation, and 
+-  making some small changes to the module source to ensure that incompatible API use is avoided.
+
+Further details are given in the [Lua compatibility](#lua-compatibility) section below.  Notwithstanding this, the Project has now deprecated Lua 5.1 and will soon be moving this version into frozen support.
```
marcelstoer, line 21: > moving this version into frozen support

Do we have consensus on that yet? I for one would be happy to remove once it's been deprecated for 2-3 release cycles.

docs/nodemcu-pil.md
```diff
@@ -0,0 +1,40 @@
...
+ -  We recommend that developers make use of the full set of `luaL_` API calls to minimise code verbosity.  We have also added a couple of registry access optimisations that both simply and improve runtime performance when using the Lua registry for callback support.
+    - `luaL_reref()` replaces an existing registry reference in place (or creates a new one if needed).  Less code and faster execution than a `luaL_unref()` plus `luaL_ref()` construct.
+    - `luaL_unref2()` does the unref and set the static int hook to `LUA_NOREF`.
+
+Rather than include simple examples of module templates, we suggest that you review the modules in our GitHub repository, such as the [`utf8`](https://github.com/nodemcu/nodemcu-firmware/tree/release/app/lua53/lutf8lib.c) library.  Note that whilst all of the existing modules in `app/modules` folder compile and work, we plan to do a clean up of the core modules to ensure that they conform to best practice.
```
marcelstoer, line 40: > we suggest that you review the modules in our GitHub repository, such as the `utf8`

Feels odd to me as that's not a "module" as such.

docs/nodemcu-lrm.md
TerryE: > Do we have consensus on that yet?

I feel that the committers actively involved in this process (you, me, Nathaniel, Gregor and perhaps Philip) would like to see this sooner rather than later.  Like you, I feel that it will probably take a couple of release cycles to sort out residual gremlins and round out both the documentation and NodeMCU extras.  One side effect of going for minimal change to the Lua53 core + aligning the 5.1 and 5.3 APIs over this last year is that a lot of the nasty bits of the `eLua` implementation have been removed without loss of functionality.  

No doubt they could be with some further polishing, but with these two new reference documents we at least have a formal specification of NodeMCU Lua is and advice to new C module developers can now sensibly boil down to (i) read and understand the reference manuals and (ii) start with the Lua and modules libraries for examples.

Somewhere along this process we are going to have to grasp the nettle of merging all of these enhancement into (and integrating) the ESP32 codebase, but to be honest I would be a lot more comfortable starting this if I had committed support from one of the more ESP32-aware committers.

jmattsson:
I'm going to try to sit down with this this evening.

TerryE:
> I'm going to try to sit down with this this evening.

@jmattsson, thanks for this.  Even if you only have limited time available and can only give general feedback, I still value your perspective and depth of experience.

It terms of the overall structure of the code, you will see that all of the eLua botches are gone and we now have a proper layering of node and other libraries over `lauxlib.h` with the only API into Lua internals offered by `lua.h`.  I've backported enough Lua 5.3 extension into the Lua 5.1 API that all of the `app/modules` compile and work against both runtimes without material `#ifdef` variants.  Lua 5.3 seems to be stable as well; more RAM efficient and slightly faster than Lua 5.1.

TerryE:
> Woah Terry, for this alone I will from now on call you Sir Terry ... 

Thanks for the complement.  Much appreciated. 

> What I would appreciate if you included it in this PR are changes to the _Lua Developer FAQ_ and the _Extension Developer FAQ_. I am quite convinced that a number of paragraphs in both documents could be replaced by links to the new reference documents.

To understand the reason you need to have a scan through the companion document the [Lua 5.3 Reference Manual](https://www.lua.org/manual/5.3/) and start with this contents. This is a very dense but formal specification of the basic concepts and language is covered in §1-3 and the standard Lua libraries are in §6.  §4-5 cover the C API that is used to code C linbrary modules.

This NLR is really a supplement specifying how our NodeMCU implementation differs from the standard Lua language and and implementation as described in the LRM.  You will see that I follow the same order, prose style and layout, but only discuss the _differences_.  The reason for the seeming focus on the C API is that there very few differences for §1,2,3,6 as the Lua language and standard libraries are almost unchanged.  The bit that we have added is really the support for ROTables and the ability to move constant data structures from RAM in code space in ROM, and the extra API calls defined in `lua.h` and  `lauxlib.h` that support.

So essentially the LRM + NLR + PiL + its supplement (I haven't got a neat name for this) + the ReadTheDocs library documentation now for the first time form a complete specification of NodeMCU Lua, its libraries and how to implement new ones.

_I am not wedded to NLR. If any of you guys can come up with a more appropriate name and acronym then we can do the swap._

As to _Lua Developer FAQ_,  the _Extension Developer FAQ_, and the _LFS Whitepaper, these are getting a little dated and it's probably time for new editions removing content that overlaps with these new and more formal references.  However, I would rather do this editing cycle as a separate PR.  This PR is already big enough IMO.

TerryE:
To be honest documenting the NodeMCU delta acted as a trigger to tidy up the API a little more, because as I documented some of the calls I realised that they hadn't be done in the "Lua implementation style" and reworking the API to conform to this style actually improved the implementation.  A good example of this is the `node.LFS.__index` implementation. The original 5.1 version had lots of nasty internal Lua runtime stuff in it and this would be very different for Lua 5.3 By abstracting this and moving the appropriate functionality into the `lua.h` interface and the `luaxlib.h` one, each runtime looks after its own version specific internals and the `node.c` code is common to both.

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+Pushes an array onto the stack containing all strings in the specified strings table. If `opt` = 0 then the RAM table is used, else if `opt` = 1 and LFS is loaded then the LFS  table is used, else `nil` is pushed onto the stack.
+
+Returns a status boolean 1 = table pushed. 
+
+#### lua_stripdebug
+`  int lua_stripdebug (lua_State *L, int level);`         [-1, +0, e]
+
+This function has two modes.  A value is popped off the stack.  
+-  If this value is `nil`, then the default strip level is set to this if `level` in the range 0 to 2, otherwise the current default level is returned.
```
HHHartmann, line 140: The current default level is always returned
```suggestion
-  If this value is `nil`, then the default strip level is set to this if `level` in the range 0 to 2. Returns the current default level.
```

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+` void (lua_createrotable) (lua_State *L, ROTable *t, const ROTable_entry *e, ROTable *mt);`         [-0, +1, -]
+
+Create a RAM based `ROTable` pointing to the `ROTable_entry` vector `e`, and metatable `mt`.
+
+#### lua_debugbreak and ASSERT
+
+`  void (lua_debugbreak)(void);`
+
+`lua_debugbreak()` and `ASSERT(condtion)` are available for for development debugging.  If `DEVELOPMENT_USE_GDB` is defined then these will trigger a debugger break and evaluate a conditional assert prologue on the same.  If not, then these are effectively ignored and generate no executable code.
```
jmattsson: "condtion" -> "condition"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+#### lua_dump
+
+` int lua_dump (lua_State *L, lua_Writer writer, void *data, int strip);`         [-0, +0, –]
+
+Dumps function at the top of the stack function as a binary chunk as per LRM.  However the last argument is now an integer is in the range -1..2, rather a boolean as per standard Lua:
+- `strip`
+    - -1, use the current default strip level (which can be set by [`lua_striptdebug`](#lua_striptdebug))
```
jmattsson, line 91: "lua_striptdebug" -> "lua_stripdebug"?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+    - 1, discard Local and Upvalue debug info; keep line number info
+    - 2, discard Local, Upvalue and line number debug info
+
+The internal NodeMCU `Proto` encoding of debug line number information is typically 15× more compact than in standard Lua; the intermediate `strip=1` argument allows the removal of must of the debug information whilst retaining the ability to produce a proper line number traceback on error. 
+
+#### lua_freeheap
+
+` int (lua_freeheap) (void);`         [-0, +0, –]
+
+returns the amount of free heap available to the Lua memory allocator.
```
jmattsson, line 102: Maybe make a note that this doesn't take into account fragmentation of said available memory? I.e. the max allocatable chunk may be significantly smaller.

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Auxiliary Library Functions and Types 
+
+Note that the LRM defines an auxiliary library which contains a set of functions that assist in coding convenience and economy.  These are strictly built on top of the basic API, and are defined in `lauxlib.h`.  By convention all auxiliary functions have the prefix `luaL_` so module code should only contain `lua_XXXX()` and `luaL_XXXX()` data declarations and functions.  And since the `lauxlib.h` itself incudes `lua.h`, all C modules should only need to `#include "lauxlib.h"` in their include preambles.
+
+NodeMCU adds some extra auxiliary functions above those defined in the LRM.
+
+#### luaL_lfsreload
+
+`  int  lua_lfsreload (lua_State *L);`         [-1, +1, -]
```
jmattsson, line 173: Didn't the return type get changed to `void` above?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+`  int (luaL_rometatable) (lua_State *L, const char* tname, const ROTable *p);`         [-0, +1, e]
+
+Equivalent to `luaL_newmetatable()` for ROTable metatables.  Adds key / ROTable entry to the registry  `[tname] = p`, rather than using a new RAM table.
+
+#### luaL_unref2
+
+`  luaL_unref2(l,t,r)`
+
+This macro executes `luaL_unref(L, t, r)` and then assigns `r = LUA_NOREF`.
```
jmattsson, line 240: Handy. Can I assume it's got a `do { luaL_unref(L, t, r); r = LUA_NOREF; } while (0)` safe-guard so it's safe to use as
```
if (something)
  luaL_unref2(L, t, r);
```
?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Declaring modules and ROTables in NodeMCU
+
+All NodeMCU C library modules should include the standard header "`module.h`".  This internally includes `lnodemcu.h` and these together provide the macros to enable declaration of NodeMCU modules and ROTables within them.  All ROtable support macros are wither prefixed by `LRO_` (Lua Read Only) or in the case of table entries `LROT_`. 
+
+#### NODEMCU_MODULE
+
+` NODEMCU_MODULE(sectionname, libraryname, map, initfunc)`
+
+This macro enables the module to be statically declared and linked in the `ROM` ROTable.  The global environment's metafield `__index=ROM` hence any entries in the ROM table are resolved as read-only entries in the global environment.
```
jmattsson, line 250: "metafield" -> "metafield is set as" or some such? It took me a couple of reads to grok the meaning of the sentence as is.

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+-  `sectionname`. This is the linker section for the module and by convention this is the uppercased library name (e.g. `FILE`).  Behind the scenes `_module_selected` is appended to this section name if corresponding "use module" macro (e.g. `LUA_USE_MODULES_FILE`) is defined in the configuration.  Only the modules sections `*_module_selected` are linked into the firmware image ad those not selected are ignored.
```
jmattsson: "ad" -> "and"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+ -  `libraryname`.  This is the name of the module (e.g. `FILE`) and is the key for the entry in the `ROM` ROTable.
```
jmattsson, line 252: "FILE" -> "file"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+#### LROT_BEGIN and LROT_END
+
+`  LROT_BEGIN(rt,mt,flags)`
+`  LROT_END(rt,mt,flags)`
+These macros start and end a ROTable definition.  The three parameters must be the same in both declarations.
+-  `rt`.  ROTable name.
+-  `mt`.  ROTable's metatable.  This should be of the form `LROT_TABLEREF(tablename)` if the metatable is used and `NULL` otherwise.
+-  `flags`. The Lua VM table access routines use a `flags` field to short-circuit where the access needs to honour metamethods during access.  In the case of a static ROTable this flag bit mask must be declared statically during compile rather than cached dynamically at runtime.  Hence if the table is a metatable and it includes the metamethods `__index`, `__newindex`, `__gc`, `__mode`, `__len` or `__eq`, then the mask field should or-in the corresponding mask for example if `__index` is used then the flags should include `LROT_MASK_INDEX`.  Note than index and GC are a very common combination, so `LROT_MASK_GC_INDEX` is also defined to be `(LROT_MASK_GC | LROT_MASK_INDEX)`.
```
jmattsson, line 263: "note than" -> "note that"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+Note that ROTable entries can be declared in any order except that keys starting with "`_`" must be declared at the head of the list.  This is a pragmatic constraint for runtime efficiency.  A lookaside cache is used to optimise key searches and results in a direct table probe in over 95% of ROTable accesses.  A table miss (that is the key doesn't exist) still requires a full scan of the list, and the main source of table misses are scans for metafield values.  Forcing these to be at the head of the ROTable allows the scan to abort on reading the first non-"`_`" key.
+
+ROTables can still support other key and value types by using an index metamethod to point at an C index access function.  For example this technique is used in the `utf8` library to return `utf8.charpattern`
+
+```C
+LROT_BEGIN(utf8_meta, NULL, LROT_MASK_INDEX)
+  LROT_FUNCENTRY( __index, utf8_lookup )
+LROT_END(utf8_meta, NULL, LROT_MASK_INDEX)
+
+LROT_BEGIN(utf8, NULL, 0)
```
jmattsson, line 278: @HHHartmann you mean here, to match the `LROT_END()` below, right?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+-  The `string.dump()` `strip` parameter can take integer values 1,2,3 (the [`lua_stripdebug`](#lua_stripdebug) strip parameter + 1).  `false` is synonymous to `1`, `true` to `3` and omitted takes the default strip level.
+-  The `string` library does not offer locale support. 
+-  The 5.3 `math` library is expanded compared to the 5.1 one, and specifically:
+    - Included: ` abs`, ` acos`, ` asin`, ` atan`, ` ceil`, ` cos`, ` deg`, ` exp`, ` tointeger`, ` floor`, ` fmod`, ` ult`, ` log`, ` max`, ` min`, ` modf`, ` rad`, ` random`, ` randomseed`, ` sin`, ` sqrt`, ` tan` and ` type`
+    -  Not implemented: ` atan2`, ` cosh`, ` sinh`, ` tanh`, ` pow`, ` frexp`, ` ldexp` and ` log10`
+ -  Input, output OS Facilities (the `io` and `os` libraries) are implement for firmware builds because of the minimal OS supported offered by the embedded run-time.  The separately documented `file` and `node` libraries provide functionally similar analogues.  The host execution environment implemented by `luac.cross` does support the `io` and `os` libraries.
```
jmattsson, line 296: "are implement" -> "are not implemented"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+-  The full `debug` library is implemented less the `debug.debug()` function.  `debug.getstrings(type)`where type is one of `'ROM'`, or `'RAM'` (the default) returns a sorted array of the strings returned from the [`lua_getstrings`](#lua_getstrings) function.
```
jmattsson, line 297: Spacing around `debug.getstrings(type)` seems wonky? Should this be on a new line altogether perhaps?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+The following Compatibility modes are enabled:
+-  `LUA_COMPAT_APIINTCASTS`.  These `ìnt`cast are still used within NodeMCU modules.
+-  `LUA_COMPAT_UNPACK`.  This retains `ROM.unpack` as a global synonym for `table.unpack` 
+-  `LUA_COMPAT_LOADERS`.  This keeps `package.loaders` as a synonym for `package.seachers`
+-  `LUA_COMPAT_LOADSTRING`.  This keeps `loadstring(s)` as a synonym for `load(s)`.
+
+### New Lua 5.3 features back-ported into the Lua 5.1 API
+
+-  Table access routines in Lua 5.3 are now type `int` rather than `void` and return the type of the value pushed onto the stack.  The 5.1 routines `lua_getfield`, `lua_getglobal`, `lua_geti`, `lua_gettable`, `lua_rawget`, `lua_rawgeti`, `lua_rawgetp` and `luaL_getmetatable` have been updated to mirror this behaviour and return the type of the value pushed onto the stack.
+-  There is a general numeric comparison API function `lua_compare()` with macros for `lua_equal()` and `lua_lessthan()` whereas 5.1 only support the `==` and `<` tests through separate API calls.  5.1 has been update to mirror the 5.3. implementation.
```
jmattsson, line 321: "update" -> "updated"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+-  Lua 5.3 includes a `lua_absindex(L, idx)` which converts ToS relative (e.g. `-1`) indices to stack base relative and hence independent of further push/pop operations.  This makes using down-stack indexes a lot simpler. 5.1 has been update to mirror this 5.3 function.
```
jmattsson, line 322: "update" -> "updated" (deja vu)

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+## Introduction
+
+NodeMCU firmware is an IoT project ("the Project") which implements a Lua-based runtime for SoC modules based on the Espressif ESP8266 and ESP32 architectures.  This reference specifically addresses how the NodeMCU Lua implementation relates to standard Lua as described in the two versions of the Lua language that we currently support:
+
+-  The [Lua 5.1 Reference Manual]( https://www.lua.org/manual/5.1/) and
+-  The [Lua 5.3 Reference Manual]( https://www.lua.org/manual/5.3/) (**LRM**)
+
+Developers using the NodeMCU environment should familiarise themselves with the 5.3 LRM.
+
+The Project provides a wide range of standard library modules written in both C and Lua to support many ESP hardware modules and chips, and these are documented in separation sections in our [online documentation](https://nodemcu.readthedocs.io/).
```
nwf, line 12: "in separate sections", not "separation"

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+This reference supplements LRM content and module documentation by focusing on the differences between NodeMCU Lua and standard Lua 5.3 in use.
+
+One of our goals in introducing Lua 5.3 support was to maintain the continuity of our existing C modules by ensuring that they can be successfully compiled and executed in both the Lua 5.1 and 5.3 environments.  This goal was achieved by combination of:
+-  enabling relevant compatibility options for standard Lua libraries;
+-  regressing some Lua 5.3 API enhancements back into our Lua 5.1 implementation, and
```
nwf, line 18: Maybe "back-porting" instead of "regressing"?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+As well as providing the ability to building runtime firmware environments for ESP chipsets, the Project also offers a `luac.cross` cross compiler that can be built for common platforms such as Windows 10 and Linux, and this enables developers to compile source modules into a binary file format for download to ESP targets and loading from those targets.
+ 
+## Basic Concepts
+
+The NodeNCU runtime offers a full implementation of the [LRM §2](https://www.lua.org/manual/5.3/manual.html#2) core concepts, with the following adjustments:
+
+### Values and Types
+
+-  The firmware is compiled to use 32-bit integers and single-precision (32-bit) floats.  Address pointers are also 32-bit, and this allows all Lua variables to be encoded in RAM as an 8-byte "`TValue`" (compared to the 12-byte `Tvalue` used in Lua 5.1
```
nwf, line 31: Missing close-paren?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+-  C modules can statically declare read-only tables known as a "`ROTable`".  There are some limitations to the types for ROTable keys and value, in order to ensure that these are consistent with static declaration.  ROTables are stored in code space (in flash memory on the ESPs), and hence do not take up RAM resources.  However these are still represented by the Lua type _table_ and a Lua application can treat them during execution the same as any other read-only table.  Any attempt to write to a `ROTable` or to set its metatable will throw an error.
```
nwf, line 32: Number mismatch in "tables known as a"; perhaps `tables, known as "ROTable"-s`?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+These are implemented as per the LRM.  Note that the Lua 5.1 and Lua 5.3 language implementations are different and can introduce breaking incompatibilities when moving between versions, but this is a Lua issue rather than a NodeMCU one.
+
+### Garbage Collection
+
+All LFS functions, any string constants used in these functions, and any ROTables are stored in static code space.  These are ignored by the Lua Garbage collector (LGC).  The LGC only scans RAM-based resources and recovers unused ones.  The NodeMCU LGC has slightly modified configuration settings that increase its aggressiveness as heap usage approaches RAM capacity.
+
+### Coroutines
+
+The firmware includes a full coroutine implementation, but note that there are some slight differences between the standard Lua 5.1 and Lua 5.3 implementations.
```
nwf, line 45: Forward-ref the discussion of differences?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+The NodeMCU runtime implements a non-blocking threaded model that is similar to that of `node.js`, and hence most Lua execution is initiated from event-triggered callback (CB) routines within C library modules.  NodeMCU enables full recovery of error diagnostics from otherwise unprotected Lua execution by adding an additional auxiliary library function [`luaL_pcallx`](#luaL_pcallx).  All event-driven Lua callbacks within our library modules use `luaL_pcallx` instead of `lua_call`.  This has the same behaviour if no uncaught error occurs.  However, in the case of an error that would have previously resulted in a panic, a new SDK task is posted with the error traceback as an upvalue to process this error information.  
+
+The default action is to  print a full traceback and then trigger processor restart, that is a similar outcome as before but with recovery of the full error traceback.  However the `node.onerror()` library function is available to override this default action; for example developers might wish to print the error without restarting the processor, so that the circumstances which triggered the error can be investigated.
```
nwf, line 63: Do we want to tell our users to restart in the `onerror` CB in production environments?

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+#### LROT_xxxx_ENTRY
+
+ROTables only support static declaration of string keys and value types: C function, Lightweight userdata, Numeric, ROTable.  These are entries are declared by means of the  `LROT_FUNCENTRY`, `LROT_LUDENTRY`, `LROT_NUMENTRY`, `LROT_INTENTRY`, `LROT_FLOATENTRY` and `LROT_TABENTRY` macros.  All take two parameters: the name of the key and the value. For Lua 5.1 builds `LROT_NUMENTRY` and `LROT_INTENTRY` both generate a numeric `TValue`, but in the case of Lua 5.3 these are separate numeric subtypes so these macros generate the appropriate subtype.
+
+Note that ROTable entries can be declared in any order except that keys starting with "`_`" must be declared at the head of the list.  This is a pragmatic constraint for runtime efficiency.  A lookaside cache is used to optimise key searches and results in a direct table probe in over 95% of ROTable accesses.  A table miss (that is the key doesn't exist) still requires a full scan of the list, and the main source of table misses are scans for metafield values.  Forcing these to be at the head of the ROTable allows the scan to abort on reading the first non-"`_`" key.
```
nwf, line 269: I wonder if we should end up sorting them alphabetically, in some eventuality.  Not pertinent to this PR.

KT819GM:
Not sure, but maybe someone should note about esp32-s2. It will be ideal replacement for esp8266 (in my opinion), but at the moment it’s not supported and possibly will not be till 4.x LTS will roll out.

docs/nodemcu-lrm.md
HHHartmann: I also thought about this. Do we know the length of the array? Then we could do binary search.

app/lua/lflash.c
```diff
@@ -257,60 +256,20 @@ LUALIB_API int lua_lfsreload (lua_State *L) {
...
- */
-LUAI_FUNC int lua_lfsindex (lua_State *L) {
-  int n = lua_gettop(L);
+LUA_API void lua_getlfsconfig (lua_State *L, int *config) {
+  if (!config)
+    return;
+  config[0] = (int) flashAddr;                   /* LFS region mapped address */
+  config[1] = flashAddrPhys;                 /* LFS region base flash address */
+  config[2] = G(L)->LFSsize;                        /* LFS region actual size */
+  config[3] = cast(FlashHeader *, flashAddr)->flash_size;  /* LFS region used */
```
HHHartmann, line 268: Shouldn't there be a check if LFS is flashed at all (as in the Lua53 Version)?

TerryE:
First of all, my thanks to everyone for a super job in doing this review.  I suspect that 90% of the comments will result in an "OK will do".  In this case instead of explicitly posting this and generating a storm of these email notifications, I will just mark the comment as resolved.  In such cases, please take the "thanks, will do" implicitly.

I had emailed an ODT version of the docs to some viewers.  (Gregor replied with a marked version so his proofing comments aren't inline here.)  What I suggest I do is that as well updating the document here I save the before and after versions and do a markup compare in Google Docs.  I will send out this link by email to reviewers.

So now begins my task of reviewing and aggregating all comments.  I suspect that this will need a chunk of work.  Where appropriate I will do this inline, but will post separate comments for specific that merit wider discussion  _To be continued :smile:_

docs/nodemcu-lrm.md
TerryE: > I'm trying to schedule some work time to spend on the next step(s) on the 8266 & 32 branch merging work.

Fantastic.  I suspect that this effort is going to be resourced by the usual suspects: You, me, Nathaniel and Gregor, with Marcel doing his usual sanity check oversight.  

There are some fundamental architectural enhancements that I want to add to Lua 5.3 (such as on ESP building of LFS images, and dual LFS images) that I feel that we should deploy on the ESP8266 platform soon and get stable before we do the actual branch merge.  (See this [\#3128 comment](3128#issuecomment-638452596) and following for discussion about the advantages of a dual LFS.)

Even so (so long I can continue to dodgy this Covid bullet),  we should plan to do this within the next few months.

docs/nodemcu-pil.md
TerryE: We actually have two groups of modules:
-  the "standard" Lua ones: `bit`, `coroutine`, `debug`, `math`, `string`, `table`, `utf8`.  Even the base functions (such as `print`) are implemented as really of the module `ROM` which also includes all of the other linked modules as members.  `ROM` is linked to `_G` by its `__index` metamethod.  So `_G.ROM.print`, `_G.print`, `ROM.print`, `print` all resolve to the same print function.
- the NodeMCU ones linked in from `app/modules`

In terms of their code they are structured the same way, use the same Lua C APIs.  However, the difference was that the authors of the Lua ones really knew their way around the C API, which more than can be said for most of our module authors, who essentially copied Zeroday's coding style (written at a point when he was still leaning the C API).  Hence, in general the standard Lua module implementation are a lot tighter / neater than our modules. :scream: 

What I have done is on the earlier line "`Hence a NodeMCU C library module typically has a standard layout`" to add the explanatory qualifier "`that parallels that of the standard Lua library modules and uses the same C API to access the Lua runtime:`".

One of the thing on my / @nwf TODO list is to clean up our core modules. We can update this document linking to a cleaned up module once we've done at least one, so we don't fell ashamed when comparing them to the Lua ones. :rofl:

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+
+ROTables can still support other key and value types by using an index metamethod to point at an C index access function.  For example this technique is used in the `utf8` library to return `utf8.charpattern`
+
+```C
+LROT_BEGIN(utf8_meta, NULL, LROT_MASK_INDEX)
```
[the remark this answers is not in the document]
TerryE: It works.  Try it! :smile: I have added the extra explanatory para below the C inclusion:

Any reference to `utf8.charpattern` will call the `__index` method function (`utf8_lookup()`); this returns the UTF8 character pattern if the name equals `"charpattern"` and `nil` otherwise.  Hence `utf8` works as standard even though ROTables don't natively support a string value type.

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,335 @@
...
+  LROT_FUNCENTRY( char, utfchar )
+  LROT_FUNCENTRY( len, utflen )
+  LROT_FUNCENTRY( codes, iter_codes )
+LROT_END(utf8, LROT_TABLEREF(utf8_meta), 0)
+```
+
+### Standard Libraries
+
+-  Basic Lua functions, coroutine support, Lua module support, string and table manipulation are as per the standard Lua implementation.  However, note that there are some breaking changes in the standard Lua string implementation as discussed in the LRM, e.g. the `\z` end-of-line separator; no string functions exhibit a CString behaviour (that is treat `"\0"` as a special character).
+-  The modulus operator is implemented for string data types so `str % var` is a synonym for `string.format(str, var)` and `str % tbl` is a synonym for `string.format(str, table.unpack(tbl))`.  This python-like formatting functionality is a very common extension to the string library, but is awkward to implement with `string` being a `ROTable`.
```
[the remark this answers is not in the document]
TerryE: Ah, there are a couple of long standing bugs in standard Lua 5.1 relating to internal functions that assume that Lua strings are CStrings
-  `print()` treats any string as a CString and hence truncates any printed field at `'\0'`. (See #2363)
-  The search functions treat patterns as CStrings and will terminate the pattern at `'\0'`.  This was "fixed" by a workaround of having a `%z` character class.  IIRC, this caused a bug in one of the HTTP modules.

Both of these were fixed in Lua 5.2 and in Lua 5.3  `%z` is no longer treated as a character class.

docs/nodemcu-lrm.md
TerryE: Agreed.  Also added the following to the next point `Returns an _approximate_ estimate of the heap freed (this can be a lot higher if some strings can be garbage collected).`  In practice because of heap management overheads etc. and GC of strings this estimate can be 2-3× too small.  IMO, better to get the heap size before and after or just avoid the allocation in the first place by setting the default level beforehand.

app/lua/lapi.c
```diff
@@ -1136,15 +1212,15 @@ LUA_API const char *lua_setupvalue (lua_State *L, int funcindex, int n) {
...
   G(L)->memlimit = limit;
 }
 
-LUA_API void legc_set_mode(lua_State *L, int mode, int limit) {
-   global_State *g = G(L);
-
-   g->egcmode = mode;
-   g->memlimit = limit;
+LUA_API void lua_getegcinfo (lua_State *L, int *totals) {
+  if (!totals) {
```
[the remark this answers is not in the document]
TerryE: Whoops! Done.  Bottle of whiskey to the reviewer.

app/lua/lflash.h
```diff
@@ -43,8 +43,5 @@ typedef struct {
 } FlashHeader;
 
 LUAI_FUNC void luaN_init (lua_State *L);
```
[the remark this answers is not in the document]
TerryE: The `lua[A-KM-Z]_` prefixes are reserved for internal to Lua core calls.  This init hook is very much internal to the core and _should_ therefore have such a prefix.  Maybe it fits better in another module, but that's a secondary point.

`lflash.h` is only included in the modules that need access to these internals:  `lflash.c`, `lstate.c` and `luas_cross/lflashimg.c`

app/lua/lnodemcu.c
```diff
@@ -115,6 +116,98 @@ LUALIB_API int luaL_posttask( lua_State* L, int prio ) {            // [-1, +0,
...
+} /* Dummy stub on host */
+#endif
+
+#ifdef LUA_USE_ESP
+/*
+ * Return an LFS function
+ */
+LUALIB_API int luaL_pushlfsmodule (lua_State *L) {
+  if (lua_pushlfsindex(L) == LUA_TNIL) {
+    lua_remove(L,-2);  /* dump the name to balalnce the stack */
```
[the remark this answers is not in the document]
TerryE: You'd better email me your home address if you want that whiskey :rofl:

app/lua/lstrlib.c
```diff
@@ -141,17 +140,25 @@ static int writer (lua_State *L, const void* b, size_t size, void* B) {
 
 static int str_dump (lua_State *L) {
   luaL_Buffer b;
+  int strip, tstrip = lua_type(L, 2);
```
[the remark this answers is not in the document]
TerryE: We're in C world, not Lua.  Bro. Comma has lower precedence than assignment, though is more commonly used in `for()` constructs.

docs/nodemcu-lrm.md
TerryE: Bugger, you've caught me out.  They should match _by convention_ which isn't enforced, though if parameter 1 doesn't match we get a compile error.  Parameter 2 is only used in the END declaration.  Even so, I've updated both the doc and the the live C code.  Thanks

docs/nodemcu-lrm.md
TerryE: Correct, though in hindsight why we didn't us `io` instead of `file` seems a bad decision.

docs/nodemcu-lrm.md
TerryE: -  Now a separate indented bullet
    -   An extra function `debug.getstrings(type)` has been added; `type` is one of `'ROM'`, or `'RAM'` (the default). Returns a sorted array of the strings returned from the [`lua_getstrings`](#lua_getstrings) function.

docs/nodemcu-lrm.md
TerryE: I've used:

C modules can statically declare read-only variant of the `Table` structure type known as a "`ROTable`".

docs/nodemcu-lrm.md
TerryE: I did try this when I was first playing with different implementation strategies and benchmarked a binary chop search.  It made very little difference largely because of the effectiveness of the LRU cache and so this applies to LRU misses only.

At the moment the code does a linear scan using an inline (masked) word comparison of the first char[4] of the key and then only does a strcmp as a final check.  With a binary chop you need to so a strcmp on _every_ comparison so the binary chop only starts to win out when the table have more than 16 entries or so.  Most modules don't have large ROTables.

The time that must do a full table scan is in the case of a key miss and in practice the vast majority of these are metamethod probes which is why the keys starting with "`_`" _must_ be at the start of the table.  Doing a ROTable access used to be maybe 50× slower than a RAM Table access. Now it is something like 30% slower.  Small enough not to bother with further optimisation, IMO.

app/lua/lflash.c
TerryE: replaced by:
```C
  config[3] = (G(L)->ROstrt.hash) ? cast(FlashHeader *, flashAddr)->flash_size : 0;  
```
The used is 0 if the ROstrt is NULL.

app/lua/lstrlib.c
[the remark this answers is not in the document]
jmattsson: Oh yes, there is nothing wrong here other than my brain being further into Lua / NodeJS land than it should be while reviewing C code!

app/lua/lflash.h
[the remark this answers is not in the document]
jmattsson: Cool, I just wanted to double-check :)

app/lua/lnodemcu.c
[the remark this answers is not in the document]
jmattsson: I'm a princess and much prefer full bodied reds. But I think I'd rather have the review favour returned down the line instead! 😄

docs/lua53.md
```diff
@@ -0,0 +1,272 @@
+## Background and Objectives
+
+The NodeMCU firmware is currently based on the Lua 5.1.4 core with the `eLua` patch and other NodeMCU specific enhancements and optimisations ("**Lua51**").  This paper discusses the rebaselining of NodeMCU to the latest production Lua version 5.3.5 ("**Lua53**").  Our goals in this upgrade were:
+
+-  NodeMCU should offer a current Lua version, 5.3.5 that is as functionally complete as practical.
+
+-  Lua53 will adopt a minimum change strategy against the standard Lua source code base, that is changes to the VM and runtime system will only be made where there is a compelling reasons for any change, for example Lua53 preserves some valuable NodeMCU enhancements, for example the addition of Lua VM support for constant Lua program and data being executable directly from Flash ROM in order to free up RAM for application use to mitigate the RAM limitations of ESP-class IoT devices
```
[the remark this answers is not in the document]
TerryE: Yup I've clarified "valuable NodeMCU enhancements introduced in Lua51", and as you've mentioned in the next comment, I wrote this paper _before_ we had released Lua53 to describe what is it is going to be.  

_And thus it was_: time to remove the future tense throughout this document.

docs/lua53.md
```diff
@@ -0,0 +1,272 @@
...
+
+The test configuration has some variations from the standard suite:
+-  NodeMCU lua and luac.cross do not support dynamic loading and the related dynamic loading tests are omitted.
+-  The tests adopt the Lua compatibility modes implemented in our builds.
+-  The standard Lua VM supports the initiation of multiple VM environments and this feature is used in some tests.  Our firmware supports multiple Lua threads but only one `lua_newstate()` instance.  So the host `luac.cross` make with the `TEST=1` option set also supports multiple VM environments.
+
+This execution environment also emulates LFS loading and execution using the `-F` option to load an LFS image before running the `-e` script.  On POSIX environments this allocates the LFS region using kernel extension, a page-aligned allocator and also uses the kernel API to turning off write access to this region except during the simulated write to flash operations.  In this way unintended writes to the LFS region throw a H/W exception in a manner parallel to the ESP environment.
```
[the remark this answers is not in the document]
TerryE: TBH, I used this to debug the Lua GC and to make sure that it was skipping LFS GCobjects correctly.  Tryingto GC something in IROM0 is sudden death! :rofl:

docs/nodemcu-lrm.md
```diff
@@ -0,0 +1,331 @@
...
+
+As well as providing the ability to building runtime firmware environments for ESP chipsets, the Project also offers a `luac.cross` cross compiler that can be built for common platforms such as Windows 10 and Linux, and this enables developers to compile source modules into a binary file format for download to ESP targets and loading from those targets.
+ 
+## Basic Concepts
+
+The NodeNCU runtime offers a full implementation of the [LRM §2](https://www.lua.org/manual/5.3/manual.html#2) core concepts, with the following adjustments:
+
+### Values and Types
+
+-  The firmware is compiled to use 32-bit integers and single-precision (32-bit) floats.  Address pointers are also 32-bit, and this allows all Lua variables to be encoded in RAM as an 8-byte "`TValue`" (compared to the 12-byte `Tvalue` used in Lua 5.1
```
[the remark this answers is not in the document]
TerryE: Yup, but this is downstream -- for now.

docs/nodemcu-lrm.md
TerryE: Agreed, but its the best we have ATM.  This is also a very cheap call since this value is held as a `size_t` in DRAM0, so this is just a getter for this.

docs/nodemcu-lrm.md
TerryE: At the moment it restarts (and so never returns on a successful load, but it returns an error string (e.g. file does not exist) on error.  Personally I'd prefer it to be a `void`and to throw errors, but returning the error was the consensus when we had the discussion back when.

docs/nodemcu-lrm.md
TerryE: First confession time: no I missed this so 15-all on the C syntax competition.  So **fixed** this.

I did a quick grep-fest of the headers and also came up with:
-  4 cases of trailing `;` which shouldn't be there. **Fixed**.
-  only one `do { } while(0)` use.  My macro (because single prec fmod is missing form the xtensa lib):
```C
#define luai_nummod(L,a,b,m) do { \
  if (b==0) luaG_runerror(L,"modulo by zero"); \
  (m) = (a) - floor((a)/(b))*(b); \
  if ((m)*(b) < 0) (m) += (b); \
} while (0)
```
-  2 else stealers
-  8 bare `{ }` constructs, such as `#define lua_tmpnam(b,e)	{ e = (tmpnam(b) == NULL); }` 

The Lua headers just use the bare brace constructs and avoid the  `do { } while(0)` even though they aren't syntactically identical.  I don't think that I need to change bare brace constructs in the private headers.

docs/nodemcu-lrm.md
jmattsson: Yes, but as part of this PR it does change to a `void`. See line 202 in `lflash.c` and 290 in `lauxlib.h`.

TerryE:
After comments from @marcelstoer, I suggest that we adopt the abbreviations:
-  **LRM** for the [Lua 5.3 Reference Manual](https://www.lua.org/manual/5.3/)
-  **NRM** for the supplemental NodeMCU Reference Manual has has been created in this PR
-  **PiL** for the 4th edition of Roberto Ierusalimschy's book "Programming in Lua"
-  **PiN** for the supplemental NodeMCU reference "Programming in NodeMCU".

I have tweaked the lede sections of these documents to adopt this convention consistently.

TerryE:
I've been away from my main dev setup for a few days, so I have been coding up #3206 in my mind.  However, I will resist the scope creep of adding it here. I will update this with the review comments and let's merge this first. #3206 can quickly follow.  

I have been going though and updating the documentation, but I propose to hold off on `lfs.md` updates, since will be significantly impacted by the new functionality.  (It's quite difficult writing this when you have a 3 year-old grandson wanting a cuddle 🤣)

TerryE:
Hopefully this latest set of updates has swept up all of the review comments.  Note that I've missed hooking the  `lua53.md` paper into `mkdocs.yml`, but it is in now.  I'd like to merge this in the next da or so, so I can move onto the 2 LFS + ESP building changes.

galjonsfigur:
When I was testing Lua 5.3 LFS I found that when `node.flashreload` function gets called on non-existing file it gives misleading error message: `No LFS partition allocated` and on Lua 5.1 it gives expected error message: `LFS image file not found`. I think the fix could be added to this PR rather than opening separate issue and PR.

galjonsfigur:
Another little bug was found by updating to GCC 10 - in `lua53/linit.c`, line 78:
```
LROT_TABLE(base_func);
```
is missing `extern` attribute and that causes build to fail with message:
```
/bin/ld: .output/host/release/obj/linit.o:(.rodata+0x260): multiple definition of `base_func_ROTable'; .output/host/release/obj/lbaselib.o:(.rodata+0x2e0): first defined here
```

PR #3197, opened 2020-07-05
add mcp23017 Lua module

### LINKED ISSUE #3057: Add MCP23017 port expander support as a lua module

plomi-net:
Fixes #3057

Add MCP23017 port expander support as a lua module. See issue #3057 and old pull request #3067 
(rebase doesn't working correctly, sry)

### DIFF
--- a/docs/lua-modules/mcp23017.md
+++ b/docs/lua-modules/mcp23017.md
@@ -5,7 +5,7 @@
 | 2020-04-10 | [Marcel P.](https://github.com/plomi-net) | [Marcel P.](https://github.com/plomi-net) | [mcp23017.lua](../../lua_modules/mcp23017/mcp23017.lua) |
 
 
-This Lua module provides access to the MCP23017 module.
+This Lua module provides access to the MCP23017 I²C I/O Expander.
 
 The [MCP23017](http://ww1.microchip.com/downloads/en/devicedoc/20001952c.pdf) is a port expander and provides 16 channels for inputs and outputs. 
 Up to 8 devices (128 channels) are possible by the configurable address (A0 - A2).
@@ -40,7 +40,7 @@ Automatically resets the device state (see `mcp23017:reset()`)
 
 #### Parameter
 - `address` address for MCP23017, default: 0x20 (should be between 0x20 and 0x27)
-- `i2c_id`  id for the i2c bus connection (i2c.setup separately)
+- `i2c_id`  id for the i2c bus connection (remember to call i2c.setup before)
 
 #### Return
 `true` if device found, otherwise `false`.
--- a/docs/lua-modules/mcp23017.md
+++ b/docs/lua-modules/mcp23017.md
@@ -1,0 +1,230 @@
+# Lua MCP23017 Module for NodeMCU / ESP8266
+
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2020-04-10 | [Marcel P.](https://github.com/plomi-net) | [Marcel P.](https://github.com/plomi-net) | [mcp23017.lua](../../lua_modules/mcp23017/mcp23017.lua) |
+
+
+This Lua module provides access to the MCP23017 module.
+
+The [MCP23017](http://ww1.microchip.com/downloads/en/devicedoc/20001952c.pdf) is a port expander and provides 16 channels for inputs and outputs. 
+Up to 8 devices (128 channels) are possible by the configurable address (A0 - A2).
+
+Due to the 16 channels, 2 bytes are required for switching outputs or reading input signals. These are A and B. 
+A single pin can be set or a whole byte. 
+
+The numbering of the individual pins starts at 0 and ends with 7.
+The numbers are for each register GPIO A and GPIO B.
+
+
+!!! important
+	The module requires `i2c` and `bit` C module built into firmware.
+
+
+### Require
+```lua
+mcp = require "mcp23017"
+```
+
+## Example Script
+The example script can be found [here](../../lua_examples/mcp23017/mcp23017_example.lua)
+
+## setup()
+Configures the address of the module and tests the connection to the i2c bus. 
+The i2c id is required for an existing i2c interface, alternatively the sda and scl pins can be specified.
+Then this function will establish the connection.
+Automatically resets the device state (see `mcp23017:reset()`)
+
+#### Syntax
+`mcp23017:setup(address, i2c_id)`
+
+#### Parameter
+- `address` address for MCP23017, default: 0x20 (should be between 0x20 and 0x27)
+- `i2c_id`  id for the i2c bus connection (i2c.setup separately)
+
+#### Return
+`true` if device found, otherwise `false`.
+
+#### possible Errors
+- `MCP23017 device on address not found`
+- `MCP23017 address is out of range`
+
+#### Example
+```lua
+local mcp23017 = require "mcp23017"
+
+local address = 0x20
+local cSCL = 1
+local cSDA = 2
+local i2c_instance = 0
+
+-- setup i2c bus and create instance for mcp23017 (assigned to mcp)
+i2c.setup(i2c_instance, cSDA, cSCL, i2c.SLOW)
+local mcp = mcp23017(address, i2c_instance)
+```
+
+## setMode()
+Set the mode of a single channel. This can be OUTPUT or INPUT.
+
+#### Syntax
+`mcp23017:setMode(register, pin, mode)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+- `pin` the number to be set for the channel (0-15)
+- `mode` the mode for the channel. This can be `mcp23017.INPUT` or `mcp23017.OUTPUT`
+
+#### Return
+`true`, in case of error `nil`.
+
+#### Example
+```lua
+-- set pin 7 and 8 to output (GPA7 and GPB0) and GPB1 to input 
+mcp:setMode(mcp23017.GPA, 7, mcp23017.OUTPUT)
+mcp:setMode(mcp23017.GPB, 0, mcp23017.OUTPUT)
+mcp:setMode(mcp23017.GPB, 1, mcp23017.INPUT)
+```
+
+## setPin()
+Set the state of a single channel. This can be HIGH or LOW.
+
+#### Syntax
+`mcp23017:setMode(register, pin, state)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+- `pin` the number to be set for the channel (0-15)
+- `state` the state for the channel. This can be `mcp23017.HIGH` or `mcp23017.LOW`
+
+#### Return
+`true`, in case of error `nil`.
+
+#### Example
+```lua
+-- set pin 7 to high (GPA7)
+mcp:setPin(mcp23017.GPA, 7, mcp23017.HIGH)
+-- set pin 8 to low (GPB0)
+mcp:setPin(mcp23017.GPB, 0, mcp23017.LOW)
+```
+
+## getPinState()
+get the state for a single channel. This can be HIGH or LOW.
+
+#### Syntax
+`mcp23017:getPinState(register, pin)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+- `pin` the number for which a state is to be queried (0-15)
+
+#### Return
+`true` for HIGH, `false` for LOW, in case of error `nil`.
+
+#### Example
+```lua
+-- get the state for pin 9 (GPB1)
+print(mcp:getPinState(mcp23017.GPB, 1))
+```
+
+## reset()
+By calling this function, a safe state is established. 
+All channels are set to input.
+This function can be used for a panic program.
+
+#### Syntax
+`mcp23017:reset()`
+
+#### Parameter
+None
+
+#### Return
+None
+
+#### Example
+```lua
+-- reset the mcp23017 to startup defaults
+mcp:reset()
+```
+
+## setInternalPullUp()
+Enable or disable the internal pullup resistors.
+
+#### Syntax
+`mcp23017:setInternalPullUp(register, byte)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+- `byte` byte to set the pullup resistors
+
+#### Return
+None
+
+#### Example
+```lua
+-- enable all pullup resistors for GPA
+print(mcp:setInternalPullUp(mcp23017.GPA, 0xFF))
+-- disable all pullup resistors for GPA
+print(mcp:setInternalPullUp(mcp23017.GPA, 0x00))
+```
+
+## writeIODIR()
+Setup the mode of the channels with a whole byte.
+
+
+#### Syntax
+`mcp23017:writeIODIR(register, byte)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+- `byte` byte to set the mode for all channels for this register
+
+#### Return
+None
+
+#### Example
+```lua
+-- set all GPA to input
+print(mcp:writeIODIR(mcp23017.GPA, 0xFF))
+-- set all GPA to output
+print(mcp:writeIODIR(mcp23017.GPA, 0x00))
+```
+
+## writeGPIO()
+Setup the output state of the channels with a whole byte.
+
+#### Syntax
+`mcp23017:writeGPIO(register, byte)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+- `byte` byte to set the state for all channels for this register
+
+#### Return
+None
+
+#### Example
+```lua
+-- set all GPA to HIGH
+print(mcp:writeGPIO(mcp23017.GPA, 0xFF))
+-- set all GPA to LOW
+print(mcp:writeGPIO(mcp23017.GPA, 0x00))
+```
+
+## readGPIO()
+Read the input states of the channels with a whole byte.
+
+#### Syntax
+`mcp23017:readGPIO(register)`
+
+#### Parameter
+- `register` the side of channels (GPA or GPB)
+
+#### Return
+byte with states
+
+#### Example
+```lua
+-- get states for GPA
+print(mcp:readGPIO(mcp23017.GPA))
+```
+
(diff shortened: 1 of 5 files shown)


sonaux:
Please add mcp23017_example.lua from previous PR.

plomi-net:
> Please add mcp23017_example.lua from previous PR.
yes sure, sry

galjonsfigur:
@plomi-net Travis build is failing because of `luacheck` warnings. I've looked and it seems like you can just tell `luacheck` to ignore them by adding ` -- luacheck: no unused` to lines `129` and `151` at the end of them

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,289 @@
...
+    if (addr > 31 and addr < 40) then
+        return true
+    else
+        print("MCP23017 address is out of range")
+        return false
+    end
+end
+
+-- write byte
+function mcp23017:writeByte(registerAddr, val)
```
galjonsfigur, line 89: Similarly to functions above - those are intended to be internal functions so they shouldn't be exported. The declaration could be rewritten to:
```lua
local function writeByte(self, registerAddr, val)
```
same goes to `readByte`, `getGPIORegisterAddr`, `checkPinIsInRange`, `writeToRegister`, `readFromRegister`

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,289 @@
...
+        self:writeByte(MCP23017_DEFVALA, iByte)
+    end
+end
+
+-- set default GPIO mode
+function mcp23017:setDefaultMode(bReg, iByte)
+    self:writeByte(self:getDirRegisterAddr(bReg), iByte)
+end
+
+function mcp23017:numberToBool(val)
```
galjonsfigur, line 151: This function seems unnecessary as only values like  `OUTPUT`, `INPUT`, `HIGH` and `LOW` (so only booleans) are allowed according to module documentation.

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,289 @@
...
+function mcp23017:setInternalPullUp(bReg, iByte)
+    if bReg == self.GPB then
+        self:writeByte(MCP23017_DEFVALB, iByte)
+    else
+        self:writeByte(MCP23017_DEFVALA, iByte)
+    end
+end
+
+-- set default GPIO mode
+function mcp23017:setDefaultMode(bReg, iByte)
```
galjonsfigur, line 147: This can be inlined as it's only one line and it's called from only one function.

docs/lua-modules/mcp23017.md
```diff
@@ -0,0 +1,230 @@
+# Lua MCP23017 Module for NodeMCU / ESP8266
+
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2020-04-10 | [Marcel P.](https://github.com/plomi-net) | [Marcel P.](https://github.com/plomi-net) | [mcp23017.lua](../../lua_modules/mcp23017/mcp23017.lua) |
+
+
+This Lua module provides access to the MCP23017 module.
+
+The [MCP23017](http://ww1.microchip.com/downloads/en/devicedoc/20001952c.pdf) is a port expander and provides 16 channels for inputs and outputs.
```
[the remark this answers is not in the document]
plomi-net: @galjonsfigur Thank you very much for your advice.
This sounds logical, unfortunately I don't have a mcp23018 to test it. So I would leave it out for now.

lua_examples/mcp23017/mcp23017_example.lua
```diff
@@ -0,0 +1,110 @@
...
+
+	@version 1.0.0
+
+]]
+
+
+
+
+
+--[[
```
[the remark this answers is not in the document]
plomi-net: I removed whitespaces and empty lines.

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,289 @@
...
+local MCP23017_INTFB = 0x0F
+
+local MCP23017_INTCAPA = 0x10
+local MCP23017_INTCAPB = 0x11
+local MCP23017_OLATA = 0x14
+local MCP23017_OLATB = 0x15
+]]
+
+-- check device is available on address
+function mcp23017:checkDevice()
```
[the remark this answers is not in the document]
plomi-net: Thanks also for this advice.
I have now reprogrammed the internal function to internal.

lua_modules/mcp23017/mcp23017.lua
plomi-net: Thanks also for this advice. This help me a lot.
I have now reprogrammed the internal function to internal.

lua_modules/mcp23017/mcp23017.lua
plomi-net: This function was removed.

lua_modules/mcp23017/mcp23017.lua
plomi-net: i agree, same as above

plomi-net:
@galjonsfigur thank you for the advice with Travis build check. It is good to know. I refactor the function so the check was passed now.

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,270 @@
...
+function mcp23017:reset()
+    -- all to input
+    setDefaultMode(self, self.GPA, 0xFF)
+    setDefaultMode(self, self.GPB, 0xFF)
+end
+
+-- setup device
+function mcp23017:setup(address, i2cId)
+
+    self.address = string.format('0x%02X', address)
```
galjonsfigur, line 150: converting `address` to string isn't nessesary AFAIK - I've made small modification by changing this line to: `self.address = address` and in `checkAddress` function: 
```
local addr = tonumber(self.address)
    if (addr > 31 and addr < 40) then
```

to

``` 
if (self.address > 31 and self.address < 40) then
```

and it seems to work exactly the same

docs/lua-modules/mcp23017.md
```diff
@@ -0,0 +1,230 @@
+# Lua MCP23017 Module for NodeMCU / ESP8266
+
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2020-04-10 | [Marcel P.](https://github.com/plomi-net) | [Marcel P.](https://github.com/plomi-net) | [mcp23017.lua](../../lua_modules/mcp23017/mcp23017.lua) |
+
+
+This Lua module provides access to the MCP23017 module.
```
galjonsfigur, line 8: this sentence sounds pointless - maybe replace `MCP23017 module` with `MCP23017  I²C I/O Expander` ? I'm not a native speaker but it sounds a little bit better.

docs/lua-modules/mcp23017.md
```diff
@@ -0,0 +1,230 @@
...
+The i2c id is required for an existing i2c interface, alternatively the sda and scl pins can be specified.
+Then this function will establish the connection.
+Automatically resets the device state (see `mcp23017:reset()`)
+
+#### Syntax
+`mcp23017:setup(address, i2c_id)`
+
+#### Parameter
+- `address` address for MCP23017, default: 0x20 (should be between 0x20 and 0x27)
+- `i2c_id`  id for the i2c bus connection (i2c.setup separately)
```
galjonsfigur, line 43: You could change: 

(i2c.setup separately)

to:

Remember to call (`i2c.setup` before)

docs/lua-modules/mcp23017.md
plomi-net: @galjonsfigur Clearly, i'm not a native speaker too. I have change it.

lua_examples/mcp23017/mcp23017_example.lua
```diff
@@ -0,0 +1,70 @@
...
+]]
+-- initialize module
+local mcp23017 = require "mcp23017"
+local address = 0x20 -- the address for MCP23017 (between 0x20 and 0x27)
+local cSCL = 1 -- SCL pin = 1 = D1 / GPIO 5 (ESP8266)
+local cSDA = 2 -- SDA pin = 2 = D2 / GPIO4 (ESP8266)
+local i2cId = 0 -- i2c id
+-- setup i2c bus and create instance for mcp23017 (assigned to mcp)
+i2c.setup(i2cId, cSDA, cSCL, i2c.SLOW)
+local mcp = mcp23017(address, i2cId)
```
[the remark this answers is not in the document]
plomi-net: @galjonsfigur 
Thanks for your help. I hope you mean this error:
```
MCP23017 device on 0x20 not found
PANIC: unprotected error in call to Lua API (script4.lua:65: attempt to concatenate a nil value)
```
I added an condition to catch this error.

galjonsfigur (approved):
>Thanks for your help. I hope you mean this error:
>```MCP23017 device on 0x20 not found
>PANIC: unprotected error in call to Lua API (script4.lua:65: attempt to concatenate a nil value)
>```
>I added an condition to catch this error.

Yes - that one. I think the module is ready to be merged. The only thing that I could to is to add the note about untested possibility to work with MCP23018 IC, but other than that LGTM. Thanks for the work on this!

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,262 @@
...
+    self.i2cId = 0
+    self.address = nil
+    self.OUTPUT = false
+    self.INPUT = true
+    self.GPA = false
+    self.GPB = true
+    self.HIGH = true
+    self.LOW = false
+
+    self:setup(address, i2cId)
```
nwf, line 258: I'm not jazzed about this printing and always returning a table (that just happens to have `deviceOK` set to `false` on the case of an error).  I would prefer that `checkAddress` and `checkDevice` be moved to the example code and `setup` removed from the driver code.  This has the added benefit that all the `deviceOK` checks can also be dropped.

TerryE:
A few comments:.
1.  The return value from `require"mcp23017"`: whilst this can be anything (eg. a function or userdata), the convention is that "well behaved" modules should return a table; and other types such as function (as is this case) are unconventional.  We might want to do this on an ESP, if  the table form offers little benefit, as using a direct function saves a little RAM.  Even so, this should be clearly documented as this is non-standard usage, and it could confuse many Lua developers.
2. The old throw vs. return status debate.  Historically the initial NodeMCU module developers tended to avoid use of throw and (usually) return status -- which the application program often ignores leading to all sorts of bizarre silent errors.  This was largely because uncaught errors panicked the ESP and restarted the processor with no error diagnostics.  _This is no longer the case_.  It is also not the Lua way, and the more that I've worked with this language and platform, the more I believe in the wisdom of sticking to Lua practice.  IMO, hard errors such as using the wrong address, channel or pin indicate some coding error and should just be thrown; return statuses should only be used for soft / potentially intermittent errors.
3.  So for example `checkAddress()` and `checkDevice()` are only called once during setup, but just throwing an error makes the code simpler.   If you don't want the entry function to throw then why not just wrap the `self:setup()` in a `pcall`?  In the case of `deviceOK`, this is set up true or false only at setup.  Currently setup still returns an object with `deviceOK` set to false, but any subsequent call fails.  So why have this status at all?  Why not just return nil or throw an error if the device isn't OK and only return the object if it is OK.  This way all the `deviceOK` checks are redundant since the functions can only be called if it is OK.  And since you are using a function constructor return from the require , there also seems little point in exposing setup as a public method.
4.  `checkPinIsInRange()` **should _definitely_ throw an error** rather than just returning a `nil` as your code doesn't check this return value isn't checked, and this is just going to cause bizarre behaviour in `readByte()`, etc.
5.  A related issue is using "print" statements for error diagnostics.  This is great if you are watching your device with a UART plugged in, but many usecases will be headless.  Having printed errors is a total PITA.  At least if you throw the error, then the application can capture it with a `pcall` and the application can choose to print it, log it or return it over the WiFi.
6.  This is a matter of style, but sometimes the [Lua Ternary Operator](http://lua-users.org/wiki/TernaryOperator) form makes code simpler and shorter.  For example `getDirRegisterAddr()` could be coded as `return bReg == self.GPB and MCP23017_IODIRB or MCP23017_IODIRA`, or even in this case  `return bReg and MCP23017_IODIRB or MCP23017_IODIRA`.
7.  Using metatables or not. There are pros and cons here.  In some cases their use is practically essential: if you are using C modules and ROTables.  If your objects will be mutli-instance in most cases, etc.,  because the one metatable is shared across all instances.  However, they can also have distinct drawbacks: for example `for k,v in pairs(mcp) do print(k,v) end` does not print out your methods.  I feel that if you are going to the extra level of complexity of using a metatable then it is best to do this properly.  In this case all R/O properties should be in the metatable and _not_ the instance table, e.g. `INPUT`, `OUTPUT`, `GPA`, `GPB`, `HIGH` and `LOW`.
7.  An another example of this avoidable complication is that routines like `readbyte()` are in essence low level but you are making them class methods to you've got lots of `self` table accesses in them.  This adds code size and runtime.  Why not pass the address and i2cid in as parameters?  This allow you to bind these as lazy closures.  So, let's assume that you keep the metatable.
```Lua
local isINPUT, isGBP, isHIGH = true, true, true -- use upvals in inline code to avoid self lookups

-- The metatable initialisation becomes
local mcp23017 = {
  __index = mcp23017,  
  INPUT = isINPUT,   OUTPUT = not isINPUT,
  GPA   = not isGBP, GPB    = isGPB, 
  HIGH  = isHIGH,    LOW    = not isHIGH
}

-- Below is an example of how we could strip down the redundancy.  The other 
-- functions could either be left as class methods acquring state from self
-- or move to lazy closures with the state also passed in as upvals.

return function(address, i2cId)
  local self = setmetatable({}, mcp23017)
  if setup(self, address, i2cId) then 
    self.writeIODIR = function(self, bReg, newByte)
       writeByte(address, i2cId,                         -- upvals
                 bReg == isGBP and MCP23017_IODIRB or MCP23017_IODIRA, newByte)
    end

    self.writeGPIO = function(self, bReg, newByte)
      writeByte(address, i2cId,                          -- upvals
                bReg == isGBP and MCP23017_GPIOB or MCP23017_GPIOA, newByte)
    end

    self.readGPIO = function(self, bReg, newByte)
      readByte(address, i2cId,                           -- upvals
               bReg == isGBP and MCP23017_GPIOB or MCP23017_GPIOA)
    end
    return self
  end
  return nil
end
```

Anyway, most style-type comments are a matter of personal preference. This is intended mostly as food for thought rather than mandatory requests, but get back to me if you think that I am talking out of my arse or you don't understand my points. :smile:

marcelstoer:
@plomi-net Do you plan to address Terry's comments?

plomi-net:
> @plomi-net Do you plan to address Terry's comments?

@marcelstoer Sorry, time passes quickly. I would like to respond to the comments from terry, which are really helpful. I had already fixed a few little things.

plomi-net:
@TerryE and @nwf Thank you for your detailed comments.
I have optimized some parts and that saves about 80 lines of code. 
I agree with most of the comments.

One of your first comments is that a function is returned instead of a table. 
This is still the case, I suppose. I am a bit confused about the Lua wisdom.

Furthermore, there are certainly many advantages and disadvantages to using a metatable or allowing multi instances. The MCP23017 can theoretically be used with 8 devices that need different instances.

I look forward to your feedback and thank you for your help.

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -17,14 +17,8 @@
...
-local i2c, string = i2c, string
-local issetBit = bit.isset
-local setBit = bit.set
-local clearBit = bit.clear
-
--- metatable
-local mcp23017 = {}
-mcp23017.__index = mcp23017
+local i2c, string, issetBit, setBit, clearBit, error = i2c, string, bit.isset, bit.set, bit.clear, error
+local isINPUT, isGPB, isHIGH = true, true, true
```
nwf, line 21: These constants are probably just going to make things slower, sadly.  While I appreciate that named enumerations are better than booleans in many cases, it's probably best to just name the arguments suggestively.

lua_modules/mcp23017/mcp23017.lua
```diff
@@ -0,0 +1,184 @@
...
+        reset(address, i2cId)
+        return 1
+    end
+end
+
+return function(address, i2cId)
+    local self = setmetatable({}, mcp23017)
+
+    if setup(address, i2cId) then
+        self.writeIODIR = function(sf, bReg, newByte) -- luacheck: no unused
```
nwf, line 125: Having these functions close over address and i2cId means that there are going to be unique closures for each mcp23017 instantiated.  Any reason not to have them on the metatable and use `self.address` and `self.i2cId`?

nwf (changes requested):

lua_modules/mcp23017/mcp23017.lua
marcelstoer: Isn't premature optimization the [root of all evil](https://ubiquity.acm.org/article.cfm?id=1513451)? I personally value more legible code - that in itself is subjective - higher unless I can prove the lines in question are "too slow" (again, often subjective).

nwf: Maybe so, but `function(foo) if foo == isINPUT then ... end end` is less clear to me than `function(isINPUT) if isINPUT then ... end`.

galjonsfigur:
New version of the code works like it should (tested quickly on a breadboard with DIP version of MCP23017) It takes considerably less RAM (at least when being run from flash and not using LFS). I will test how it behaves on LFS and how much RAM usage differs between current and previous version of the module.

galjonsfigur:
The neat thing about making function that after being called returns table with all the module functions and constants is that running the module from LFS is extremely simple and requires changing line or two of code - in this particular case from:
```lua
local mcp23017 = require "mcp23017"
...
local mcp = mcp23017(address, i2cId)
```
to
```lua
local mcp = node.LFS.get("mcp23017")()(address, i2cId)
```
The difference between the previous and current version of the module in case of RAM usage when in LFS is about 100 bytes less with current version - for me a clear improvement!

plomi-net:
@galjonsfigur an improvement of 100 Bytes is amazing. How much RAM does the program use now?

PR #3202, opened 2020-07-08
ESP32: Add AM2320 support

thornley-david:

Ported almost raw from ESP8266 implementation (https://github.com/nodemcu/nodemcu-firmware/blob/release/app/modules/am2320.c).

### DIFF
--- a/components/modules/Kconfig
+++ b/components/modules/Kconfig
@@ -54,6 +54,12 @@ config LUA_MODULE_ADC
       Includes the adc module. This module provides access to the
       adc1 hardware.
 
+config LUA_MODULE_AM2320
+  bool "AM2320 module"
+  default "n"
+  help
+      Support for AM2320 temp/humidity sensor
+
 config LUA_MODULE_BIT
   bool "Bit module"
   default "n"
--- a/components/modules/am2320.c
+++ b/components/modules/am2320.c
@@ -1,0 +1,139 @@
+/*
+ * app/modules/am2320.c
+ *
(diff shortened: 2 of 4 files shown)


marcelstoer:
I'd like to land #3135 from @nwf first and then for you to rebase.

thornley-david:
No problems, this can definitely wait.

nwf:
Giving the C a quick once-over, there doesn't seem to be any reason that this should be a C module (or ESP32 specific) and not just some Lua code: the sensor communicates over I2C, there appear to be no timing requirements, and all math is on small integers.  Maybe see #3197 or the similar Lua modules (e.g., bh1750 or ds2321 or mcp23008)?

thornley-david:
@nwf, good point. I was just trying to reuse/migrate was was already there, however if there is a better way I am happy to rewrite it.

There are a couple of microsecond timing delays in the read negotiation, but they are likely not strict. To be honest, I am not aware if these can be transitioned to lua efficiently, and I was also not sure if the methodology I used was even correct when migrating (os_delay_us -> ets_delay_us). 

I'll take a look at the sample modules in detail and disband the PR if I managed to rewrite it as a lua module.

components/modules/Kconfig
```diff
@@ -54,6 +54,12 @@ config LUA_MODULE_ADC
       Includes the adc module. This module provides access to the
       adc1 hardware.
 
+config LUA_MODULE_AM2320
```
marcelstoer, line 57: In case you rebase this will have to become `NODEMCU_CMODULE_AM2320`.

PR #3203, opened 2020-07-09
add wiegand module

ccutrer:

Add a module to support Wiegand keycard readers. I first tried using the gpio module, but quickly realized that doing it in Lua was too slow to properly capture the protocol. All the bits got through, but because there are multiple data lines, and they were fast, they got lumped together and you couldn't fix the original order of the bits


app/modules/wiegand.c
```diff
@@ -0,0 +1,270 @@
...
+  }
+}
+
+static int lwiegand_close( lua_State* L)
+{
+  if(wiegand_cb_ref != LUA_NOREF) {
+    luaL_unref(L, LUA_REGISTRYINDEX, wiegand_cb_ref);
+    wiegand_cb_ref = LUA_NOREF;
+  }
+  return 0;
```
pjsg, line 219: I suspect that you ought to release the interrupt hook here.

pjsg:
I think that you probably want to handle the timer somewhat differently -- there can be significant delays between posting the task and when it gets to run. You might want to save the value of `system_get_time` in the interrupt handlers and then in the task set the timer for the appropriate time (the timeout + last bit time - current time).

app/modules/wiegand.c
ccutrer: I was trying to figure out how to do that. AFAICT, the softuart module doesn't ever unregister its hook. The rotary module does, but using a lower level function. I can't figure out the the gpio module explicitly registers their hook - it seems it's just uses an even lower level function to create a fallback hook. Any tips on the recommended way to do it?

ccutrer:
I'll definitely adjust the task posting. In terms of the timeout, I wasn't really concerned about being precise. It just needs to not try to collect the message until _at least_ that long as passed. I wanted to avoid dealing with clock wraparounds because I know they can be pitfalls. The amount of time between individual bits in a message (< 20ms) is microscopic compared to the amount of time between separate messages (I can't get my keypad to send faster than 1 message/s when I'm mashing it, and with an actual RFID card it's nearly 2s.

tools/luacheck_config.lua
```diff
@@ -717,6 +717,12 @@ stds.nodemcu_libs = {
         createClient = empty
       }
     },
+    wiegand = {
+      fields = {
+        setup = empty,
+        close = empty
+      }
+    }
```
marcelstoer, line 725: Not sure if you noticed but the CI build failed.

> Critical error: Couldn't load configuration from tools/luacheck_config.lua: syntax error (line 726: '}' expected (to close '{' at line 5) near 'wifi')

nwf:
Ah, the Wiegand protocol.  Thank you for taking a stab at doing it in Lua, but yea, this one is very timing sensitive and probably needs to be in C.

-----

However, be advised that the progenitor library, https://github.com/monkeyboard/Wiegand-Protocol-Library-for-Arduino, says in its README that

> This library is free software; you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License

I have no issue with (L)GPL (in fact, much prefer it for my own work), but the NodeMCU tree is mostly MIT.  We should perhaps have a mechanism to permit clearly documented exceptions that may change the distribution requirements for the resulting binaries (in this case, by adding an obligation to provide the wiegand module source).

app/modules/wiegand.c
pjsg: If there isn't an easy way to do this, just make sure that nothing bad happens if the interrupt hook is called when the device is closed. Also, if the device is reopened, then don't reregister the hook (or maybe the hook registration code already takes care of it). Please add a comment to say what strategy you are following in the close function.

ccutrer:
> Ah, the Wiegand protocol. Thank you for taking a stab at doing it in Lua, but yea, this one is very timing sensitive and probably needs to be in C.
> 
> However, be advised that the progenitor library, https://github.com/monkeyboard/Wiegand-Protocol-Library-for-Arduino, says in its README that
> 
> > This library is free software; you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License
> 
> I have no issue with (L)GPL (in fact, much prefer it for my own work), but the NodeMCU tree is mostly MIT. We should perhaps have a mechanism to permit clearly documented exceptions that may change the distribution requirements for the resulting binaries (in this case, by adding an obligation to provide the wiegand module source).

Good point. I copied the license lines into the top of my `wiegand.c` but I must have only been skimming because in my mind I read it as MIT :). Do you think calling it out in README.md is sufficient, or should I also call out the exception in LICENSE?

nwf:
If you don't mind, I'd like to mark this PR as blocked and raise a separate issue for the governance question?

ccutrer:
The more I think about it, the more I believe I should probably just start over from scratch. The module gets statically linked in, so technically even the LGPL would require infecting the rest of the code base with the license.

marcelstoer:
> Good point. I copied the license lines into the top of my `wiegand.c` but I must have only been skimming because in my mind I read it as MIT :). Do you think calling it out in README.md is sufficient, or should I also call out the exception in LICENSE?

That's not enough unfortunately. Unless the source can be at least dual-licensed (which wouldn't make much sense in this case) we can't accept that, sorry. I remember we have had two similar cases in the past. In one case the original author was willing to change the license of his project to something compatible with MIT. In the other case the developer contributing to NodeMCU was determined to add the new module and rewrote it "without" the source of original lib to stay clear of licensing issuesl.

ccutrer:
ok, PR updated. the Wiegand/gpio code has been completely re-written from scratch, so no more (L)GPL code. This means losing support for things besides 4 bit and 26 bit (the two types the keypad I have can send). I refactored things while I was doing it so that multiple instances can exist at the same time. While doing this, I figured out how to properly de-register my interrupt hooks (just "re-register" with a mask of 0; when registering for multiple instances it has to register a single hook for all pins at once). I also tweaked the timer code while I was in there to only delay the remaining time to get to 25ms in case the posted task takes a while to execute. Oh yeah, and another task isn't posted until the prior one runs. Phew! I learned quite a bit about the interrupt system and dbg_print while I was getting this one to work :).

marcelstoer (approved):

app/modules/wiegand.c
```diff
@@ -0,0 +1,244 @@
...
+      break;
+    if ((gpio_bits & 1) == 0)
+      continue;
+    // find the struct registered for this pin
+    volatile wiegand_t w = pins_to_wiegand_state[i];
+    if (!w) {
+      continue;
+    }
+
+  	++w->bit_count;
```
nwf, line 61: Whitespace nit

app/modules/wiegand.c
```diff
@@ -0,0 +1,244 @@
...
+{
+  uint32_t gpio_status = GPIO_REG_READ(GPIO_STATUS_ADDRESS);
+  uint32_t gpio_bits = gpio_status;
+  for(int i = 0; gpio_bits > 0; ++i, gpio_bits >>= 1) {
+    if (i == NUM_GPIO)
+      break;
+    if ((gpio_bits & 1) == 0)
+      continue;
+    // find the struct registered for this pin
+    volatile wiegand_t w = pins_to_wiegand_state[i];
```
nwf, line 56: I don't believe there's need for this to be volatile?

app/modules/wiegand.c
```diff
@@ -0,0 +1,244 @@
...
+      return true;
+  }
+  return false;
+}
+
+static void lwiegand_timer_done(void *param)
+{
+  lua_State *L = lua_getstate();
+
+  volatile wiegand_t w = (wiegand_t) param;
```
nwf, line 118: Similarly, might not need to be volatile.

app/modules/wiegand.c
```diff
@@ -0,0 +1,244 @@
...
+    lua_call(L, 2, 0);
+  }
+}
+
+static void lwiegand_cb(os_param_t param, uint8_t prio)
+{
+  wiegand_t w = (wiegand_t) param;
+  (void) prio;
+
+  *(volatile int *)&w->task_posted = 0;
```
nwf, line 137: Sorry, I am not quite sure I understand why this is being zeroed?

nwf (commented):
Sorry for not reviewing sooner.  I've one nit and some questions arising from ignorance, but this looks like it could land.

marcelstoer:
@nwf are you ready to merge this?

nwf:
Sure; let's get it some hopefully wider testing.  I, sadly, lack the hardware to do so.

PR #3209, opened 2020-07-13
Fixes dhtlib's handling of negative temperatures

emdioh:
  - There are no behaviour changes, so I didn't update the docs.

The handling of negative temperatures was wrong. The code was checking for the most significant bit, discarding it, and multiplying the resulting value by -1.
This does not work because negative numbers are 1's complement (-1 is represented as 0xFF).
I've fixed the code to use signed arithmetic and the results are now correct.

I've also factored the checksum and values decoding to their own functions, to reduce duplication.


emdioh:
Hmm there are too many changes: I'm going to redo this PR.

PR #3210, opened 2020-07-14
Dht negative

emdioh:

Negative temperatures were handled like this:
```c
if (dht_bytes[2] & 0x80)  // negative dht_temperature {
  dht_temperature = -dht_temperature;
}
```
But negative values are using 1's complement encoding, so for instance -1 is represented as 0xFF and -1 was converted to -3276
I've changed the conversion to use signed arithmetic and now it works correctly.
I've also factored the checksum and values decoding to their own functions, to reduce duplication.

### DIFF
--- a/app/dht/dht.c
+++ b/app/dht/dht.c
@@ -40,13 +40,12 @@
 #define HIGH    1
 #endif /* ifndef HIGH */
 
-#define COMBINE_HIGH_AND_LOW_BYTE(byte_high, byte_low)  (((byte_high) << 8) | (byte_low))
-
 static double dht_humidity;
 static double dht_temperature;
 
 static uint8_t dht_bytes[5];  // buffer to receive data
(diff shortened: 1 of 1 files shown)


nwf:
At a quick glance, this looks like a clear improvement.

ezbella:
Hi guys, I'm having issues with this particular commit of dht module when reading from a dht11 sensor. I was using a firmware built with a previous version of this module and everything worked fine. Issues started after building the firmware (via marcel's build service) with the latest version of this module: the same piece of code sensed the first time a value between 500 and 600, and every subsequent attempt to read the sensor throws CHECKSUM_ERROR (-1).
This is how i am reading the sensor (I've tried all of the pins):
`local status,temp,humi = dht.read11(dht11Pin)`


This is the firmware that is working **OK**:
firmware_working

And this is the firmware with issues:
firmware_not_working

PR #3222, opened 2020-07-17
Change struct & sjson to use integers. This is slightly more complex 

### LINKED ISSUE #3217: struct module uses lua_pushnumber where it ought to use lua_pushinteger
Current behavior is broken in the lua53 build.

pjsg:
Fixes #3217.

This is slightly more complex as we have to avoid casting to lua_Numbers if there a need to have integer values. We also need to deal with unsigned 32 bit integers that are not representable as a lua integer. However, this seems to work


nwf:
Color me skeptical about the use of `double` here; it's going to be implicitly converted to a `lua_Number`, and that seems like a recipe for loss of precision.

More nit-picky: please use `INT_MIN` and `INT_MAX` constants rather than hex literals.

pjsg:
The use of double was deliberate as this function returns an integer that can be much larger than a 32 bit integer (which also be unsigned). The double can hold a 53 bit integer -- but it could be that we want this to return a `int64_t` instead -- but I don't know the platform supports that.

I'll test it.

And yes, I'll fix the INT_MIN/MAX case.

nwf:
Ah, I misread the diff in my skim, sorry... `double` will probably work, but `int64_t` should work (I make heavy use of the 64-bit int types in `sntppkt` and they seem to work fine) and will be (very slightly) faster, I think.

pjsg:
Fixed sjson as well for the gross effects. Still not sure whether it uses the right number of digits or not.

app/modules/sjson.c
```diff
@@ -153,10 +153,41 @@ create_new_element(jsonsl_t jsn,
...
 
 static void push_number(JSN_DATA *data, struct jsonsl_state_st *state) {
-  lua_pushlstring(data->L, get_state_buffer(data, state), state->pos_cur - state->pos_begin);
-  LUA_NUMBER r = lua_tonumber(data->L, -1);
+  const char *start = get_state_buffer(data, state);
+  const char *end = start + state->pos_cur - state->pos_begin;
+  lua_pushlstring(data->L, start, end - start);
+  // See if there is any chance that this is an integer
+#if LUA_VERSION_NUM >= 503
+  char notint = 0;
```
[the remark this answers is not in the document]
TerryE: @pjsg TBH Philip, the only reason that I came across this was because of my work to harmonise the 5.1 and 5.3 APIs so that we don't need to have conditional `LUA_VERSION_NUM` variant code in our modules.  In general Lua hides sub types in its exposed API, so there is no difference from an API perspective between the various function subtypes: they are all classed as function.  Ditty short vs long strings, and Tables vs ROTables.   The one exception is the handling of integer vs number subtypes as there is an `lua_isinteger()` call and variant get and put API calls.  This one is a sensible extension one one that IMO is also worth back-porting into the Lua API.  If I do this, then I will also remove this variant code as the 503 version will work fine with both versions.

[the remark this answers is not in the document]
pjsg: This case is ugly -- and made worse by the fact that the input buffer is not null terminated and I was nervous that the number parsing might fall off the end of the buffer... I *suspect* that it would never happen, but I prefer to err on the side of caution.

app/modules/sjson.c
```diff
@@ -156,7 +156,7 @@ static void push_number(JSN_DATA *data, struct jsonsl_state_st *state) {
   const char *end = start + state->pos_cur - state->pos_begin;
   lua_pushlstring(data->L, start, end - start);
 #if LUA_VERSION_NUM >= 503
-  int sz = lua_stringtonumber(L, lua_tostring(L, -1));
+  int sz = lua_stringtonumber(data->L, lua_tostring(data->L, -1));
   if (sz) {
     lua_pop(data->L, 1);
```
TerryE: @psjg, You've lost me on this one.  
-   We've changed the throwing behaviour.  With 5.1 anything other than a valid number fails silently pusing 0 onto the stack.  Now we throw an error.  Do we want to do this, or should we continue the existing behaviour?
-  I can't see the added benefit of pushing the numeric string onto the stack and then immediately discarding it, when doing a `lua_stringtonumber(data->L,  start)` does the same but skips this extra stack operation.  Why not just
```C
#if LUA_VERSION_NUM == 501
  lua_pushlstring(data->L, start, end - start);
#else
  if (!lua_stringtonumber(L, start))
    lua_pushinteger(L, 0);
```
which preserves the existing behaviour

pjsg: I'd prefer to fix the 5.1 code to throw if the number cannot be parsed -- it seems to be less unfriendly than silently converting the number to 0. The original code just assumes that the LUA number parser can convert all valid JSON numbers into a value. Thinking about this some more, there are valid JSON numbers that are not representable in LUA -- but I don;'t know whether they cause parsing failures -- e.g. 1e400 which is outside the range of a float or double.

TerryE: I've just checked the behaviour.  You are right.  This uses a `lobject.o` function `luaO_str2num()` where its source states that this follows POSIX `strtod()`.  Stepping though a test case in gdb shows that this is partially incorrect in that `strtod` will stop on any non-valid delimiter, but the  `luaO_str2num()` function will return a conversion error is any tail is found after the end of the number scan.  Bottom line:  pushing the string _is_ the simplest approach.  I withdraw my comment.  Thanks.

TerryE (approved):

trev9995:
B5C27569-9C43-4185-BB01-CB6FEF9E4038

PR #3236, opened 2020-08-01
Add support to set the clock rate and have sntp use it.

### LINKED ISSUE #3235: Add support to set the clock rate and have the sntp module use it
### Justification
Currently, the clock PLL takes some time to settle down to compensate for the crystal error rate. This can take several hours. If the clock rate was saved and then restored after system boot (by LUA code), then it would start out right. The `sntp` module should start its PLL using that rate.

### Workarounds
None

pjsg:
Fixes #3235

This just adds support to `rtctime.set` to allow it to be used with the first two args as nil. In this case, if the rate is supplied as argument 3, then it is set.

The sntp module also initializes the clock PLL from the current rate.

### DIFF
--- a/docs/modules/rtctime.md
+++ b/docs/modules/rtctime.md
@@ -139,7 +139,8 @@ It is highly recommended that the timestamp is obtained via NTP (see [SNTP modul
 
 Values very close to the epoch are not supported. This is a side effect of keeping the memory requirements as low as possible. Considering that it's no longer 1970, this is not considered a problem.
 
-You specify the seconds parameter with the optional microseconds (defaults to 0). Setting the rate is independant.
+Times are specified across two arguments, seconds and microseconds. If microseconds are not specified, the default is 0 (i.e., time is exactly at the boundary between two seconds).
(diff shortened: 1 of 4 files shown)


docs/modules/rtctime.md
```diff
@@ -139,8 +139,10 @@ It is highly recommended that the timestamp is obtained via NTP (see [SNTP modul
 
 Values very close to the epoch are not supported. This is a side effect of keeping the memory requirements as low as possible. Considering that it's no longer 1970, this is not considered a problem.
 
+You specify the seconds parameter with the optional microseconds (defaults to 0). Setting the rate is independant.
```
nwf, line 142: Generally speaking, docs don't use 2nd person like this.  Separately, that's remarkably terse.  How about "Times are specified across two arguments, seconds and microseconds.  If microseconds are not specified, the default is 0 (i.e., time is exactly at the boundary between two seconds).  In addition to the time, a third argument specifies the local clock's drift rate estimate, as documented in [`rtctime.get()`](#rtctimeget).  If this argument is not given or is `nil`, the current rate estimate will not be altered, and the rate estimate may be set without changing the time by leaving the seconds and microseconds arguments `nil`."

nwf:
Delightful.  I had done something similar for #2819 and hadn't cleaved it out as its own PR.

docs/modules/rtctime.md
pjsg: Fixed.

As a side note -- this is the rate of a WEMOS D1 MINI over almost 24 hours:

image

I suspect that differences are temperature related caused by localized heating due to the amount of work being done by the CPU. I'm going to try modifying a board to use a TCXO instead and see what happens.

nwf:
I'm happy for this to land.

PR #3238, opened 2020-08-07
Makefile: don't pull binary esptool if toolchain is provided

sorki:
Only pull pre-built `esptool` when `GITHUB_TOOLCHAIN` is set.

If not set try to look-up systems `esptool.py` or bail out with an error.

This means that if `TOOLCHAIN_ROOT` is set `esptool.py` is expected to be in `$PATH`.

Related to #3239.


TerryE:
As I said on the PR, the default should remain pulling from Git, but I support the principle that this default should be overridable.

PR #3239, opened 2020-08-07
Conditional vendor SDK download

sorki:
Makes it possible to provide ESPRESSIF SDK externally instead of pulling it from GitHub.

```
GITHUB_ESPRESSIF_SDK = yes
```
is a bit weird, suggestions welcome. Marking as draft.


TerryE:
@sorki Richard, would it not be better to have this as null settings preserve the current behaviour but if `ESPRESSIF_SDK` and / or `ESPRESSIF_ESPTOOL` are defined then the relevant directory is taken for the tool root?  Ditto setting these should also suppress the relevant `clear` target?

But broadly I see no problem in the principle of this.

PR #3253, opened 2020-08-27
Fix node.random for lua 5.3 (and 64 bit integers)

### LINKED ISSUE #3250: node.random is broken wrt integer/float
The `node.random` definition has lots of cases -- sometimes it returns integers and sometimes floats. It doesn't get things right in the LUA 5.3 world.

pjsg:
Fixes #3250.

This makes node.random work for 5.1 and 5.3 dealing with nastinesses of the available defines being different. Uses pushinteger rather than pushnumber. Also works if lua_Integer is 64-bit.


app/modules/node.c
```diff
@@ -18,6 +18,18 @@
...
+#define LUA_MAXINTEGER INT_MAX
+#endif
+
+#ifndef LUA_UNSIGNED
+#if LUA_MAXINTEGER > MAX_INT
+typedef uint64_t LUA_UNSIGNED;
+#else
+typedef uint32_t LUA_UNSIGNED;
+#endif
+#endif
```
nwf, line 31: Why this hunk?

app/modules/node.c
```diff
@@ -545,16 +557,25 @@ static int node_osprint( lua_State* L )
   return 0;
 }
 
-int node_random_range(int l, int u) {
+static LUA_UNSIGNED random_value() {
+  // Hopefully the compiler is smart enought to spot the constant IF check
```
nwf, line 561: Ah, `constexpr`, one of the few things I like about C++.

nwf (commented):
Looks OK at a glance, but I'm curious about the `typedef`s up top.

app/modules/node.c
pjsg: lua5.1 doesn't have these defines. I guess that we could retrofit them into the 5.1 code as it might make other places a bit simpler.

TerryE:
Bugger, and I think that it's my bad, but there's a big overlap between this and the other PR on use of the int API. If you look at this code, then you will see that I've added unsigned variants of the API code to ensure that `uint32` types are handled correctly as well as `lua_Unsigned`.

It makes no sense two committers working on this. Why don't you just merge these PRs together?

TerryE:
Don't use `LUA_UNSIGNED` as a type. Not Lua practice. As in my mod, you need a `typedef LUA_UNSIGNED Lua_Unsigned;`

pjsg:
I have merged this into the other branch -- and I also fixed the lua_Unsigned usage.

PR #3255, opened 2020-08-28
Add missing features from Travis CI test build

### LINKED ISSUE #3254: Extend Travis CI build to test more firmware features
### Missing feature
Travis CI is compiling test build with most, but not all features (for example new I2C driver, `pmsleep` or `smartconfig` modules/features) 

### Justification
By missing some of the features in the test build, the code silently breaks sometimes and can be source of issues like (#3106 or #2952).

The fix for this would be to add another couple lines of `sed` magic to the [`pr-build.sh`](https://github.com/nodemcu/nodemcu-firmware/blob/dev/tools/travis/pr-build.sh) file to enable rest of the features on all-modules-included build. I will make the PR to add them - it shouldn't collide with any of the ongoing work and could prove useful for future PRs.

galjonsfigur:
Fixes #3254

This PR adds all missing features to the Travis CI test build, at least all those I could find.

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -437,7 +437,7 @@ void wifi_pmSleep_suspend_CB(void)
   {
     lua_State* L = lua_getstate(); // Get main Lua thread pointer
     lua_rawgeti(L, LUA_REGISTRYINDEX, wifi_suspend_cb_ref); // Push suspend callback onto stack
-    luaL_unref(L, wifi_suspend_cb_ref); // remove suspend callback from LUA_REGISTRY
+    luaL_unref(L, LUA_REGISTRYINDEX, wifi_suspend_cb_ref); // remove suspend callback from LUA_REGISTRY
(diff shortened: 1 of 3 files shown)


marcelstoer:
I support this. Can you please:
- fix the CI build
- unify the `sed` switches: ` -E -i.bak` vs. `-i.bak` vs. `-i` (haven't verified whether `-E` is actually necessary)
- fix indentation

galjonsfigur:
It should be fine now. I replaced strange `sed` parameters (`-E` option of `sed` isn't even documented anywhere...) and added small fixes as a separate commit.

marcelstoer (approved):

marcelstoer:
> `-E` option of `sed` isn't even documented anywhere

I guess that depends on your environment. On my macOS it says

> -E      Interpret regular expressions as extended (modern) regular expressions rather than basic regular expressions (BRE's).

PR #3257, opened 2020-08-29
Makefile: Read user_config by path

mk-pmb:

Disclosing the true path allows for relative includes in user_config.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -20,14 +20,14 @@ SPIFFSFILES ?= $(patsubst $(FSSOURCE)/%,%,$(shell find $(FSSOURCE)/ -name '*' '!
 # Get the filesize of /bin/0x10000.bin and SPIFFS sizing
 #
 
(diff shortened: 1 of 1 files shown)


TerryE:
As I said in the issue, I don't really want to dick around with somethig that we want to remove.

mk-pmb:
Will the removal happen soon? Otherwise, let's help people deal with the status quo until we can really remove it.

HHHartmann:
The relevant issue would be #1223? How could I include my own configs after this PR?
As I would not only have my own user_config but also user_modules and user_version would this PR suffice for that too?

mk-pmb:
> The relevant issue would be …?

I don't know if that issue is related in any way.
The issue I had, and caused me to write this patch, does not have a thread, as to me it didn't seem worth explaining. In my opinion, the opposite would be true: The current approach of concealing the filename would need a justification, and thus, documentation.

> How could I include my own configs after this PR?

I expect that all previous workflows continue to work. That is, I hope no-one has relied on using relative file paths in their config in the current scenario where these paths would be calculated based on an expressly concealed origin.

>  As I would not only have my own user_config but also user_modules and user_version would this PR suffice for that too?

I don't understand the question, sorry.
Easiest way to check would probably be to temporarily patch this into your local repo and retry.

Issue #3148, opened 2020-06-07
Supporting a reliable flash filesystem


pjsg:
I have done a bunch of fuzz testing against spiffs and there are some serious corruption issues there. Whether the LUA stack can generate the particular sequence of operations or not is unclear. I have also looked at https://github.com/ARMmbed/littlefs which is very similar to spiffs but does add support for directories. It too fell down when faced with the fuzzer, and I was hopeful that the required fixes would be forthcoming. However, no such luck yet.

It turns out that the behavior of flash memory is really nasty during power failures. If you are writing a single byte in a page (typically 256 bytes), and the power fails, then there is no guarantee on the value of the rest of the page (see my comment: https://github.com/ARMmbed/littlefs/issues/245#issuecomment-606050523 )

This makes reliable storage really quite difficult! 

If there are other filesystem candidates then I'd be prepared to connect them to my fuzzer and see what happens...

NicolSpies:
My observations over the last few years might confirm the same, before the pipe module I passed data between functions using spiffs json files. Random corruption of the saved file, rendering it unreadable, occurs with a relative low frequency. My workaround was to save two copies of the file.

nwf:
I'm hardly an expert in the space, but AIUI the move to MLC and the attendant need to whiten and ECC around bitflips has put the days of "at least we can reliably zero bits, surely" behind us.

I suppose the dance to be done is something like the ZFS top-level ring-buffered "uberblocks" (don't scream; I know ZFS is huge and is not appropriate for our space; I am talking only about a particular aspect of it), which "self-certify" their integrity via a robust checksum (IIRC fletcher4).  These are the only blocks that ZFS ever reads for which it does not know the expected checksum ahead of time; every other block pointer includes not just the device address but also the checksum of the data that should be there.  Assuming devices drain their write caches when instructed (hah), the latest uberblock that correctly self-certifies should point exclusively and transitively at stably-written sectors.

I suppose `ubifs` and `jffs2` are also too bulky to run on NodeMCU, but I'd be curious to know how well they fared against your fuzzer.

Glancing around, some other candidates may include
* https://piconomix.com/fwlib/group___p_x___l_o_g___f_s.html, though this also appears to rely on in-place zeroing of bits
* https://yaffs.net/ though this may also be too heavy-weight for us: https://yaffs.net/archives/yaffs1-memory-footprint.
[...]

PR #3260, opened 2020-08-29
Don't allocate flash to SPIFFS that is reserved by the system.

### LINKED ISSUE #3208: Wifi connection lost..... No reconnection
	branch: dev
	commit: 2fa63a1303493aa66831af5466311436d11cf827
	release: 
	release DTS: 202007071335
	SSL: false
	build type: integer
	LFS: 0x40000 bytes total capacity
	modules: file,gpio,mqtt,net,node,rtctime,sntp,tmr,uart,wifi
 build 2020-07-08 00:54 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)

I am using above build....
In long run wifi gets disconnected and does not reconnect automatically. even if i do a soft reset using  node.restart() but if i toggle power supply it reconnects.
had this issue with 4 esp-07 devices .
not sure if it is a bug or not......

pjsg:
Fixes #3208

This PR just subtracts the 20k reserved memory from the end of the flash before setting up the SPIFFS partition. This also explains why the contents of SPIFFS would get corrupted on the dev branch.

The downside of this change is that this will trigger a SPIFFS reformat (if you didn't have a fixed size spiffs). However, that is a good thing as it will no longer overwrite *important stuff*.

### DIFF
--- a/app/modules/node.c
+++ b/app/modules/node.c
@@ -750,8 +750,9 @@ static void delete_partition(partition_item_t *p, int n) {
 #define IROM0_PARTITION  (SYSTEM_PARTITION_CUSTOMER_BEGIN + NODEMCU_IROM0TEXT_PARTITION)
 #define LFS_PARTITION    (SYSTEM_PARTITION_CUSTOMER_BEGIN + NODEMCU_LFS0_PARTITION)
 #define SPIFFS_PARTITION (SYSTEM_PARTITION_CUSTOMER_BEGIN + NODEMCU_SPIFFS0_PARTITION)
+#define SYSTEM_PARAMETER_SIZE  0x3000
 
-// Lua: node.setpartitiontable(pt_settings)
+// Lua: node.setpartitiontable(ptvals)
 static int node_setpartitiontable (lua_State *L) {
   partition_item_t *rcr_pt = NULL, *pt;
   uint32_t flash_size = flash_rom_get_size_byte();
@@ -760,6 +761,7 @@ static int node_setpartitiontable (lua_State *L) {
   uint32_t n = i / sizeof(partition_item_t);
   uint32_t param[max_pt] = {SKIP, SKIP, SKIP, SKIP};
 
+/* stack 1=ptvals, 2=pt_map, 3=key, 4=ptval[key], 5=pt_map[key] */ 
   luaL_argcheck(L, lua_istable(L, 1), 1, "must be table");
   lua_settop(L, 1);
   /* convert input table into 4 option array */
@@ -771,9 +773,8 @@ static int node_setpartitiontable (lua_State *L) {
     lua_pushvalue(L, 3);  /* dup key to index 5 */
     lua_rawget(L, 2);     /* lookup in pt_map */
     luaL_argcheck(L, !lua_isnil(L, -1), 1, "invalid partition setting");
-    param[lua_tointeger(L, 5)] = lua_tointeger(L, 4);
-    /* removes 'value'; keeps 'key' for next iteration */
-    lua_pop(L, 2);  /* discard value and lookup */
+    param[lua_tointeger(L, 5)] = lua_tounsigned(L, 4);
+    lua_pop(L, 2);  /* discard value and lookup; keeps 'key' for next iteration */
   }
  /*
   * Allocate a scratch Partition Table as userdata on the Lua stack, and copy the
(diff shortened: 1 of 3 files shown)


marcelstoer:
Thanks Philip!

TerryE:
My concern here is that this area hasn't been reserved or used by the system since SDK 2.x. Since SDK 3.0, the system used the areas defined in the PT. See [user_main.c:L81-83](../blob/release/app/user/user_main.c#L81).  I did check at the time of testing this that SDK now uses the `RF_CAL`,`PHY_DATA` and `SYSTEM_PARAMETER` system partitions by erasing the old 5 pages and running a chip for a bit before checking again, but I might have missed something or some code path still uses the old locations.  So this might be work validating.  

However _if_ I am right this fix won't fix anything.  I've just checked Wemos D1 mini pro that I've been testing on for a few weeks using ` esptool --port /dev/ttyUSB0 --baud 460800 read_flash 0xfb000 0x05000 /tmp/x.bin` and this area is still all `0xFF`s.

TerryE (changes requested):
Please do not merge or restart using SDK 2.x features until we have bottomed this issue.  Thanks.

pjsg:
@TerryE What was interesting is that when this happened, the old contents of the last 12k (I was only saving 12k) did include my SSID and password, and then, when it stopped working, the SSID and password was gone.

Also I I do `wifi.sta.config({ssid="test ssid 1234", pwd="test pwd 1234"})` then those strings show up in the last 20k of flash.

*Something* is certainly writing to that area, and there is strong evidence for something reading from that area.

NicolSpies:
@pjsg, as mentioned in #3148, I have been doing continuous ongoing testing related to the SPIFFS, I also started experiencing the problem of the failing WiFi connection. Both the failures (SPIFFS related corruption & disappearing files and WiFi connection) happened at least once every 24 hours.

I have implemented your suggested change, since then none of the long time present problems have been observed so far. I will report if it changes.

TerryE:
@pjsg Philip  We have 3 reserved system partitions: `RF_CAL`,`PHY_DATA` and `SYSTEM_PARAMETER` .  We moved these to the "dead" region before the firmware, that is 0xB000 - 0xFFFF back when we did the 2.x to 3.0 upgrade; the pros and cons were debated then on the issue and PR.  I accept that I made a mistake in not including WiFi connection tests in my testing at the time and so I missed the fact that the SDK / BIOS code for the WiFi fixed the SP are as the last 3 pages in the flash, so we need to accept and implement this move.

What I don't understand is why we should also move the other two back since we _have_ validated that the SDK correctly uses the current versions.  

My suggestion is that we only move the SP partition back. We also need to check the `node.setpartitiontable()` and `nodemcu-partition.py` code to make sure that these still work.  We also need to check that this interplays well with the RFcal initialisation code ad the flash size reset.   

I will prepare and push an update along these lines, so please post back here if you disagree and then we can open this debate for resolution, before doing the change.

pjsg:
I don't have any evidence that it is using the 2 pages before the last 3
for anything. I just assumed (yes, I know) that if they screwed up the
system parameter area, they probably screwed up the other areas as well. If
those areas do seem to be mapped down low, then that is great.

galjonsfigur:
> I missed the fact that the SDK / BIOS code for the WiFi fixed the SP are as the last 3 pages in the flash, so we need to accept and implement this move.

It was very easy to miss, as Espressif stated in the [first documentation of partition table](https://github.com/espressif/ESP8266_NONOS_SDK/blob/master/documents/EN/%20Partition%20Table.md) that SYSTEM_PARAMETER partition has to be declared and can have custom address. But in "Notes" section of [README](https://github.com/espressif/ESP8266_NONOS_SDK/blob/master/README.md) they state that SYSTEM_PARAMETER partition has fixed address - very inconsistent.

TerryE:
> they probably screwed up the other areas as well. If those areas do seem to be mapped down low, then that is great

No, I don't think the other two are screwed up.  These partitions are updated as per the PT, and there is no evidence that the corresponding high pages are touched.  As @galjonsfigur points out, the documentation states: "`SYSTEM_PARTITION_RF_CAL`, `SYSTEM_PARTITION_PHY_DATA` and `SYSTEM_PARTITION_SYSTEM_PARAMETER` have to be defined in the partition table in application. _Users can change their addresses according to practical needs_."  However, the README also states that this last must be at the "correct address" (the 3 three pages of flash)

Even if we do move the SP partition back to end of flash, we still need to keep the PT valid and my code assumes that (and also validates that) the individual partitions are in ascending address and are non-overlapping, so we need to resort and update the PT to avoid side-effects.  I've made the changes and tested the load + node functions.  

`nodemcu-partition.py` doesn't crash now.  I am not really happy with this, but it is as good as it was.  I think that it is better to get this patch in before the next drop.  We can reconsider  `nodemcu-partition.py` as a low priority.

TerryE (approved):

PR #3263, opened 2020-08-30
Fixes broken `tmr.alarm`

### LINKED ISSUE #3261: tmr.create():alarm() Broken on New (dev d72ea91)
**Device :** ESP8266
```
	branch: dev
	commit: d72ea91ed01518ef0779d7aa998c5c3a9c269d71
	release: 
	release DTS: 202008291943
	SSL: false
	build type: integer
	LFS: 0x40000 bytes total capacity
	modules: file,gpio,mqtt,net,node,rtctime,sntp,tmr,uart,wifi
 build 2020-08-30 03:24 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
```

**Code:-**
```lua
if not tmr.create():alarm(5000, tmr.ALARM_SINGLE, function()
  print("hey there")
end)
then
  print("whoopsie")
end
```

**Error:-**
Lua error: 	stdin:1: bad argument #1 to 'alarm' (boolean expected)
stack traceback:
	[C]: in function 'alarm'
	stdin:1: in main chunk
	[C]: ?
	[C]: ?

vsky279:
Fixes #3261.

Fixes broken `tmr.alarm` following the commit [d72ea91](https://github.com/nodemcu/nodemcu-firmware/commit/d72ea91ed01518ef0779d7aa998c5c3a9c269d71).

The following code fragments now work correctly:
```Lua
if not tmr.create():alarm(5000, tmr.ALARM_SINGLE, function()
  print("hey there")
end)
then
  print("whoopsie")
end
```
```Lua
mytimer = tmr.create()
mytimer:register(5000, tmr.ALARM_SINGLE, function() print("hey there") end)
if not mytimer:start() then print("uh oh") end
```

### DIFF
--- a/app/modules/tmr.c
+++ b/app/modules/tmr.c
@@ -132,26 +132,29 @@ static int tmr_start(lua_State* L){
 	luaL_argcheck(L, lua_isboolean(L, 2) || lua_isnil(L, 2), 2, "boolean expected");
 	int restart = lua_toboolean(L, 2);
 
-	lua_settop(L, 1);  /* ignore any args after the userdata */
+	lua_settop(L, 1);  /* we need to have userdata on top of the stack */
 	if (tmr->self_ref == LUA_NOREF)
 		tmr->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 
-	//we return false if the timer is not idle
+	//we return false if the timer is not idle and is not to be restarted
 	int idle = tmr->mode&TIMER_IDLE_FLAG;
 	if(!(idle || restart)){
-		lua_pushboolean(L, 0);
+		lua_pushboolean(L, false);
 	}else{
 		if (!idle) {os_timer_disarm(&tmr->os);}
 		tmr->mode &= ~TIMER_IDLE_FLAG;
 		os_timer_arm(&tmr->os, tmr->interval, tmr->mode==TIMER_MODE_AUTO);
+		lua_pushboolean(L, true);
 	}
-	lua_pushboolean(L, !idle); /* false if the timer is not idle */
 	return 1;
 }
 
 // Lua: t:alarm( interval, repeat, function )
 static int tmr_alarm(lua_State* L){
 	tmr_register(L);
+	/* remove tmr.alarm's other then the 1st UD parameters from Lua stack. 
+	tmr.start expects UD and optional restart parameter. */
+	lua_settop(L, 1);
 	return tmr_start(L);
 }
 
--- a/docs/modules/tmr.md
+++ b/docs/modules/tmr.md
@@ -185,7 +185,8 @@ print( timeIt(function() tmr.ccount() end) )
 
 ### tobj:alarm()
 
-This is a convenience function combining [`tobj:register()`](#tobjregister) and [`tobj:start()`](#tobjstart) into a single call.
+This is a convenience function combining [`tobj:register()`](#tobjregister) and [`tobj:start()`](#tobjstart) into a single call. This is the reason why this method has the same parameters as `tobj:register()`.
+If `tobj:alarm()` is invoked on an already running timer the timer is stopped, new parameters are set and timer is (re)started (similar to call `tobj:start(true)`).
 
 To free up the resources with this timer when done using it, call [`tobj:unregister()`](#tobjunregister) on it. For one-shot timers this is not necessary, unless they were stopped before they expired.
 
@@ -282,7 +283,7 @@ Starts or restarts a previously configured timer. If the timer is running the ti
 - `restart` optional boolean parameter forcing to restart already running timer
 
 #### Returns
-`true` if the timer was started, `false` on error
+`true` if the timer was (re)started, `false` on error
 
 #### Example
 ```lua


app/modules/tmr.c
```diff
@@ -132,26 +132,26 @@ static int tmr_start(lua_State* L){
...
+		lua_pushboolean(L, 1);
 	}
-	lua_pushboolean(L, !idle); /* false if the timer is not idle */
 	return 1;
 }
 
 // Lua: t:alarm( interval, repeat, function )
 static int tmr_alarm(lua_State* L){
 	tmr_register(L);
+	lua_settop(L, 1); /* remove tmr.alarm's other then the 1st UD parameters from Lua stack. tmr.start expects UD and optional restart parameter. */
```
nwf, line 154: Move the comment to a block comment above the call, rather than an overlong thing to its right?

app/modules/tmr.c
```diff
@@ -132,26 +132,26 @@ static int tmr_start(lua_State* L){
...
-	//we return false if the timer is not idle
+	//we return false if the timer is not idle and is not to be restarted
 	int idle = tmr->mode&TIMER_IDLE_FLAG;
 	if(!(idle || restart)){
 		lua_pushboolean(L, 0);
 	}else{
 		if (!idle) {os_timer_disarm(&tmr->os);}
 		tmr->mode &= ~TIMER_IDLE_FLAG;
 		os_timer_arm(&tmr->os, tmr->interval, tmr->mode==TIMER_MODE_AUTO);
+		lua_pushboolean(L, 1);
```
nwf, line 146: Maybe `lua_pushboolean(L, true)` (and change the other one to `false` rather than 0?)`

nwf (commented):
Minor nits, but at a glance it LGTM.

KT819GM:
I'm still have problems, and thinking that it's related to timer. Timer will stop after some iterations (it's not constant, could be 5, could be 10). `s:on` part stays alive, as I manualy can call `s:write("1")` and it will fire response. Previous `dev` was behaving correctly. Dirty code sample for this would be:

```Lua
--read_serial
t = {0,0,0,0,0,0,0,0,0,0,0}
index = 1
s = softuart.setup(9600, 6,7)
s:on("data","\n", function(data) print('Data received, printing')
for token in string.gmatch(data, "%w+") do
    t[index] = token
    index = index + 1
    end
    print(t[2])
end)

tmr.create():alarm(2000, tmr.ALARM_AUTO, function()
print('Reading values')
index = 1
s:write("1")
    end)```

**Comment by vsky279** [comment]
Good. I cannot reproduce your issue either.

Everything works fine under Lua 5.3. However under Lua 5.1 this one does not work (no timer is fired)
```
print(tmr.create():alarm(1000, tmr.ALARM_AUTO, function() print("hey there") end));
```
while
```
t=tmr.create();print(t:alarm(1000, tmr.ALARM_AUTO, function() print("hey there") end));
```
works correctly.

This seems like a woodoo to me.

**Comment by mk-pmb** [comment]
I observed [heap usage flickering](https://gist.github.com/mk-pmb/56d6024c7f23bdcbf83a2e52abda27a2) with this fix and Lua 5.3. It was not present in my previous firmware image with Lua 5.1, but I didn't keep a backup so I can't compare. I'm trying to find a somewhat recent dev commit to rebuild a Lua 5.1 firmware and compare. However this may take a bit because somehow a lot of the late dev commits won't build for me.

**Comment by vsky279** [comment]
I think this can be due to different GC algorithm. Try to run `collectgarbage()` when your timer is running.

**Comment by mk-pmb** [comment]
Yes, thanks for explaining. Now when I collect garbage in each run, the heap use settles to a fixed number starting at the 3rd trigger.

**Comment by TerryE** [comment]
The 5.3 is GC strategy is lazier than 5.1.  5.1 effectively does a GC after every malloc.  This gives a nice low heap, but slugs runtime by maybe 3× or thereabouts.  5.3 is less frequent and sweeps up more when it does -- hence the heap flicker.  As Lukas says, you can always do a GC at the end of your timer function to force housekeeping before you sleep again.

File: app/modules/tmr.c [hunk 3]
```diff
@@ -132,26 +132,29 @@ static int tmr_start(lua_State* L){
 	luaL_argcheck(L, lua_isboolean(L, 2) || lua_isnil(L, 2), 2, "boolean expected");
 	int restart = lua_toboolean(L, 2);
 
-	lua_settop(L, 1);  /* ignore any args after the userdata */
+	lua_settop(L, 1);  /* we need to have userdata on top of the stack */
```
TerryE, line 135: This is only needed because tmr_start() is a `LUA_API` function but is also called directly by another `LUA_API` function tmr_alarm(), which is a bit naughty.  Since the alarm function is doing the naughties, then it should get the stack right, and this settop should be moved to the line immediately above the  `trm_start(L)` call in `trm_alarm()`. 

According to the documentation:
-  `start` is supposed to return `nil` if the timer is not registered and  (`true`, `mode`) otherwise. **yet in reality it only returns the boolean (was the timer started?) 
-  `alarm' is supposed `true` if the timer was started and `false` on error.

So `t:start(true)` is supposed to force a restart, but what is the behaviour of `t:alarm()` if the timer is already running?  Do you need the restart flag here as well.

Either way, the documentation should reflect the actual behaviour.

vsky279: @TerryE I already have `lua_settop` in `tmr_alarm()` at line [155-157](https://github.com/nodemcu/nodemcu-firmware/pull/3263/commits/cbf2afa7caf657e0ddcc6fddc4a2cbc070157d20#diff-ac89d4cec75b9b260a2fe9e732d2c50bR155-R157).

This one `lua_settop` is there because we need the `tmr.timer` UD on the top of the stack before eventually calling 
```
tmr->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
``` 
This gets rid off the optional argument.

Do you agree with this reasoning and approach?

nwf: Can we decouple `tmr_start()` and `tmr_alarm()` so that their `LUA_API` aspects are separate (but they might use common bits of C code encapsulated as functions)?

vsky279: Do you find this still too dirty?

I think it would not be very elegant as in `tmr_register` only one line is C code (`os_timer_setfn(&tmr->os, alarm_timer_common, tmr);`) and 16 lines relates to `LUA_API`.
It's very similar for `tmr_start`. 

To me this implementation is not so much dirty and if documented well in the code it is readable and does the job well.

TerryE: Lukas, the LUA_API entries are really intended to be invoked through `lua_call()`  If we are calling one directly then this should be clearly documented inline. _Either_ the caller or the callee should properly balance the Lua stack inbound and outbound.

Since we document the behaviour of start with the optional flag, the `tobj:alarm()` documentation should explain what happens is a `tobj` is an exisitng timer has been started.

vsky279: > the LUA_API entries are really intended to be invoked through lua_call() If we are calling one directly then this should be clearly documented inline. Either the caller or the callee should properly balance the Lua stack inbound and outbound.

This is clear. I think the proposed implementation does exactly that.

> Since we document the behaviour of start with the optional flag, the tobj:alarm() documentation should explain what happens is a tobj is an existing timer has been started.

Do you think we should add optional parameter `restart` to `tmr.alam()`? That would be probably the way giving the highest degree of freedom to user. It should be easy to be implemented + documented inline as well as in documentation.

marcelstoer: > I think this is the right behavior.

I went through the documentation and checked the obj methods again for consistency (primarily behavioral). It all makes sense _to me_.

marcelstoer (approved):
I tested all of the documented behavior on the timer obj: ✔️

Issue #3268, opened 2020-09-01
On i686 GNU/Linux, luac.cross sneakily builds invalid LFS images that fool even node.LFS.reload


mk-pmb:
### Expected behavior
* `cd app/lua/luac_cross && make LUA=53 clean all` either builds a useful executable or refuses with a link to this issue to explain why we don't support i686.
** I'd accept "nobody uses 32 bit CPUs anyway" if someone is bold enough to make that claim.
* `node.LFS.reload()` detects _before_ reboot that the reboot would fail, and thus, would not reboot at all.

### Actual behavior
When I create an LFS image with luac.cross compiled from commit b4c148e (current dev branch) with LUA=53 on Ubuntu on an x86_64 machine, it works as expected.

When I try the same on an i686 machine, it goes through all the motions with no sign that anything is wrong, and not even the pass 1 check of node.LFS.reload complains. Only after the MCU reboot it fails with `(null): invalid header in precompiled chunk`.

* [The LUA source and both resulting LFS images](http://web.archive.org/web/20200901202306if_/http://l.proggr.de/misc/proggr-dl/20/75g5la0/hello.zip), maybe it can help determine what went wrong.
* [REPL session log](https://gist.github.com/mk-pmb/97dff32db09244c7507ed54d837c42dc)

### Test code
```lua
-- -*- coding: UTF-8, tab-width: 2 -*-
return function () print('Hello.') end
```

```bash
( cd app/lua/luac_cross && make LUA=53 clean all
) && sudo mv -v luac.cross /usr/local/bin/luac-for-nodemcu
luac-for-nodemcu -f -o hello.i686.lfs hello.lua
```

### NodeMCU startup banner
See REPL log.

### Hardware
LoLin v3 ESP8266

TerryE:
If you look at the code, then you will see that we support both 32-bit and 64-bit hosts.  I understand your report that reload isn't working for an LFS image built on a 32-bit host, but at worst this is a bug and not some sort of deep-state conspiracy, so please avoid this type of wording.   You are free to use sarcasm when you are paying for my time; it really doesn't help when I am giving it pro bono.

TerryE:
[...]

PR #3269, opened 2020-09-03
LUA 5.1 luac.cross Makefile: Detect and refuse accidential invocation with LUA=53

mk-pmb:
Fixes half of #3268.

When people accidentially run the Makefile for  LUA 5.1 luac.cross with LUA=53, detect their mistake and point them to the LUA 5.1 luac.cross.

### DIFF
--- a/app/lua/luac_cross/Makefile
+++ b/app/lua/luac_cross/Makefile
@@ -15,6 +15,20 @@ CCFLAGS += -Wall
 
 TARGET = host
 
+
+# Validate LUA setting
+ifeq ("$(LUA)","")
+else ifeq ("$(LUA)","51")
+  # ok
+else ifeq ("$(LUA)","53")
+  $(error Your variable LUA="$(LUA)" looks like you probably want \
+    app/lua53/host/Makefile instead)
+else
+  $(error Unsupported value "$(LUA)" for variable "LUA", \
+    expected empty/unset (recommended) or "51")
+endif
+
+
 VERBOSE ?=
 V ?= $(VERBOSE)
 ifeq ("$(V)","1")
--- a/app/lua53/Makefile
+++ b/app/lua53/Makefile
@@ -18,6 +18,18 @@ endif
 
 STD_CFLAGS=-std=gnu11 -Wimplicit -Wall
 
+
+# Validate LUA setting
+ifeq ("$(LUA)","53")
+  # ok
+else ifeq ("$(LUA)","51")
+  $(error Your variable LUA="$(LUA)" looks like you probably want \
+    app/lua/Makefile instead)
+else
+  $(error Expected environment variable "LUA" to be "53", not "$(LUA)")
+endif
+
+
 #############################################################
 # Configuration i.e. compile options etc.
 # Target specific stuff (defines etc.) goes in here!
--- a/app/lua53/host/Makefile
+++ b/app/lua53/host/Makefile
@@ -23,6 +23,18 @@ else
   # MAKEFLAGS += --silent -w
 endif  # $(V)==1
 
+
+# Validate LUA setting
+ifeq ("$(LUA)","53")
+  # ok
+else ifeq ("$(LUA)","51")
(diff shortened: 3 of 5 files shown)


app/lua/luac_cross/Makefile
```diff
@@ -75,7 +75,11 @@ endif
 
 all: $(DEPS) $(IMAGE)
 
-$(IMAGE) : $(OBJS)
+validate_lua_var:
+	@[ "$(LUA)" != 53 ] || { echo "E: Your variable LUA='$(LUA)' looks like you probably want app/lua53/host/Makefile instead."; exit 1; }
+	@[ -z "$(LUA)" -o "$(LUA)" = 51 ] || { echo "E: Unsupported setting LUA='$(LUA)'. For LUA 5.1, unset it or set it to empty or '51'."; exit 1; }
```
HHHartmann, line 80: This check is a bit overdone since AFAICS we don't check like that anywhere else either.

HHHartmann (changes requested):

app/lua/luac_cross/Makefile
mk-pmb: AFAIR I added those error messages to give helpful messages for mistakes that I had done while trying, to save other people the effort to decipher the later failure messages. I strive to always improve user experience. If this makes other parts of the code inconsistent, I'd prefer we upgrade those other parts.

HHHartmann: Agreed. Then we should also add similar code in the lua53/host/Makefile and make sure that at least app\Makefile also fails if LUA contains something wrong.

mk-pmb: I'll try.

mk-pmb: Found two more places, and how to do the checks with Makefile commands.

app/lua53/Makefile
```diff
@@ -18,6 +18,20 @@ endif
...
+# Validate LUA setting
+ifeq ("$(LUA)","")
+else ifeq ("$(LUA)","53")
+  # ok
+else ifeq ("$(LUA)","51")
+  $(error Your variable LUA="$(LUA)" looks like you probably want \
+    app/lua/Makefile instead)
+else
+  $(error Unsupported value "$(LUA)" for variable "LUA", \
+    expected empty/unset (recommended) or "53")
```
HHHartmann, line 31: Currently only 53 is expected. we will shift the default (on unset/empty) to 53 one day, but not now.

app/lua53/host/Makefile
```diff
@@ -23,6 +23,20 @@ else
   # MAKEFLAGS += --silent -w
 endif  # $(V)==1
 
+
+# Validate LUA setting
+ifeq ("$(LUA)","")
```
HHHartmann, line 28: Same as above. Currently empty/unset means 51. This would need to be adapted when we switch the default to Lua 53

HHHartmann (changes requested):
Looks good, except for the comments.

app/lua53/Makefile
mk-pmb: I hope the default LUA version for all files inside the `lua53` directory will always be 5.3. If we need to upgrade these files, let's instead rename the directory.

app/lua53/host/Makefile
mk-pmb: Same, the path suggests 5.3 already.

app/lua53/Makefile
HHHartmann: This directory should only build if LUA=53 and not when it is empty/unset.
This will change as soon as we shift the default behaviour of the build from 51 to 53, but now it should fail if LUA is unset/empty

mk-pmb: Oh I see, you want explicit confirmation, not just clarification. That's ok with me.

HHHartmann (approved):

HHHartmann:
Thank's for your work on this. I first ran into the same pitfall.

Issue #3145, opened 2020-06-05
Addition of file.savedump(filename, func [, striplevel])


TerryE:
### Missing feature
`file.putcontents(filename, string.dump(func))` is a simple method of copying a given in-RAM or in-LFS function to an LC file in SPIFFS, however the `string.dump()` C code does
```C
  luaL_Buffer b;
  luaL_buffinit(L,&b);
  lua_dump(L, writer, &b);
  luaL_pushresult(&b);
```
and at peak this generates three copies of the function in memory:
-  the executable `Proto` hierarchy
-  the dump in the `luaL_Buffer` chunks on the Lua stack
-  their concatenation into a single TString on the stack.

This is approach is fine PC class environments but we only have ~40Kb RAM and this limits the size of function we can dump.  If we use a writer that sends directly to file, then we don't need the extra RAM to store the 2 copies of the full dump.

The main use of this function is to allow developers to dump LFS functions to LC files for #2917.

_**Note:**_

-  Because this is a `node` function, this would be available in both the Lua 5.1 and 5.3 environments.
-  It makes sense to add the `striplevel` parameter to `node.compile()` as well.

### Justification
To remove an unnecessary RAM constraint is dumping in-RAM or in-LFS functions.

### Workarounds
`file.putcontents(filename, string.dump(func))` only works for function with dumps less than 15Kb or so.  No workarounds for functions larger than this.

nwf:
> Because this is a node function, this would be available in both the Lua 5.1 and 5.3 environments.

Is `file` not available in both 5.1 and 5.3??

I think I would advocate for this to be `file.dump()` rather than `node.dumpfile()`, just so all SPIFFS/Fat functionality remains grouped together.  In terms of technical content, tho', I approve entirely.

TerryE:
OK, your suggestion of putting this function in file works well, but given that we can only have one top level function per undump, I think that this better a `file` function rather than a `file.obj` method -- that is the filename should be parameter, and `file.xxxxfile()` doesn't sound right to me so why not:
```Lua
  file.savedump(filename, func [, striplevel])
```

nwf:
I wonder... given that you have made a lovely `pipe` module for us... could we have a more generic "dump to pipe" function and pull from the pipe to the filesystem (or even elsewhere, e.g., the network) as appropriate?  Could we `undump` from a pipe similarly?  We'd need some back-pressure mechanism on `pipe` if it's not already so capable?

[...]

PR #3272, opened 2020-09-06
First cut of new LFS for initial review

### LINKED ISSUE #3206: Support for two LFS regions and on-ESP LFS building
## New feature

Additional support for two LFS regions, and the ability to both save and load (update) LFS regions on ESP.   

### Justification
Most committers and many developers would seem to want this.

### Highlights

   This enhancement will be for Lua 5.3 only as this builds upon the groundwork that I've already laid in the Lua53 implementation.
-  Nonetheless for technical reasons this will be a breaking change in terms of compiled Lua file formats.  The old formats will not be supported so any image and compiled LC files will need recompiling.  However, we will detect and throw an "old version" error if an old format file is used.
-  LFS and LC file formats will be the _same_, albeit slightly different to the existing formats (as discussed below).
-  `lua_dump()` can now take an array argument as well as a function argument at ToS.  If an array is specified then the dump stream will contain all functions in the array.
-  An extra `node.dumpfile()` function essentially does the same as `string.dump()` but writing the dump stream direct to file without needing to assemble and store the compiled content in RAM.  (This scales much better.)  Both of these functions also support the "array of functions" argument type.
-  `lua_load()` (and derived functions) now support the multi-function format, and in which case return a keyed array instead of a single function.  The array is of the form `{"name" = nameFunc, ...}`
-  `loadfile()` and `node.LFS.reload()` can take an array of filenames as well as a single file, in which case the set of files will be loaded into RAM or LFS.  In the case of loadfile this will return the `{"name" = nameFunc, ...}` array if multiple files or multi-function files are loaded.  In the case of the LFS reload, this array is not returned but is used as LFS region's index ROTable after restart.
-  Note that `node.LFS.reload()` can only process compiled files and will error on any source files.
-  An RCR option is used to determine whether  LFS1, LFS2, LFS1 + LFS2, or LFS2 + LFS1 is used for string and function resolution.  Hence a single LFS can be used; either can be used in flip-flop mode; or they can be chained in a System / App configuration.
-  When used in a chained configuration there is an inherent parent / child precedence implied.  (Let's use the terms sysLFS and appLFS for convenience. In an LFS1 + LFS2 configuration:
    - The sysLFS is loaded into LFS1 as a parent LFS.
    - The appLFS is loaded into LFS2 as a "chained" LFS and this depends on sysLFS.  The appLFS can be reloaded without invalidating the sysLFS.
    - ~~The appLFS will inherit any strings already in sysLFS `ROstrt` and hence these TStrings will not be duplicated in appLFS. ~~The `ROstrt`for the appLFS also includes all (short) TStrings in sysLFS and hence the sysLFS `ROstrt` isn't used.  The G(L) global state points to this new appLFS `ROstrt`.  From a string lookup perspective there is no difference between a single and a double LFS configuration.~~ (**Not true**.  _I do need two `ROstrt` indexes because overflow chaining depends on the table size and I can't redo these chains in the sysLFS when loading the appLFS_.)  
    - The index ROTable for the appLFS has a methamethod `__index` entry pointing to the sysLFS index ROTable.  Hence resolution across the two ROtables is "free" using standard Lua RTS array access.
    - Reloading the sysLFS will invalidate the appLFS and so it will need reloading as well
-  Change the RCR option does the expected integrity checks on any loaded LFS image and invalidates them when necessary. So for example if both LFS1 and LFS2 contain a parent LFS then the option can be toggled at will.  If it is changed to LFS1 + LFS2 then the LFS2 will be invalidated. Etc.
-   The `node.LFS.list` property has already been updated to a function which takes an optional argument: 'parent' lists only those functions in the parent LFS and likewise for 'child'; 'system' and 'application' are synonyms for these options.  Omitting the argument list all functions in both LFSs if configured as a parent / child pair.

### Example usecases

```Lua
-- Create a (child) LFS based on all LC files in SPIFFS
do
  local f,a = files.list('%.lc$'),{}
  for k in pairs(f) do a[#a+1] = k end
  node.LFS.reload(a)
end
```
```Lua
-- Create a new (child) LFS replacing one specfic function from SPIFFS
do
  local v = node.LFS.list('application')
  local a = loadfile {'mysub.lua'} -- use array form to return an array
  for _,n in ipairs(v) do
    if not a[n] then a[n] = node.LFS.get(n) end
  end
  v = {}
  for n,f in pairs(a) do v[#v+1] = f end
  node.dumpfile(v, 'lfs.img')
  node.LFS.reload('lfs.img')
end
```
### How the NodeMCU binary format differs from standard Lua 5.3

In general terms the Lua RTS dump function determinately traverses a Proto hierarchy converting all fields to a stream of binary tokens and this stream is the compiled file format.  The load executes an "undump" which does the inverse traverse recreating the Proto hierarchies.  This much is the same.  But as to why the differences:

-  File size and RAM usage is a lot more important to us so for example:
    - The default line encoding for traceback on the last code example is 192 bytes in the standard RTS but we use a run length encoding that takes 12 byte.
    -  The standard tokenisation is aimed at simplicity so an integer is represented by 1+4 byte stream.  We use a multibyte stream so in our case the constant `1` takes a single byte.

-  I have also reordered the dump walk so that compiled code on reloading can be written sequentially to the LFS region using  `spi_flash_write()` operations. 
-  When loading one or more Proto hierarchies in a file into LFS we need to add any TString constants that are not already in LFS.  I do this by maintaining in RAM a copy of what will become the `ROstrt` for the LFS.  This is a lookup that allows fast resolution against TString clashes, but in case of a clash I still need to compare the new TString against the copy in LFS to differentiate between a true match and a hash duplication.  This uses ICACHE resolved access, and so I need to flush the cache to ensure cache coherence.  I would rather do this once per file rather than once per proto.
-  Hence the dump process collects the array of TStrings used in the dump and appends this as a string vector at the end of the file. Any inline TString references use an index into this vector.
-  The NodeMCU file format includes a fixed header which includes a file CRC and the offset of this TString vector. The undumper fseeks to the TString vector and processes this first before fseeking back to the start of the file to process the Protos.  This avoids the need for two passes during dump.
-  The CRC-32 is at a fixed offset from the start of the file and can be used as an image ID, and `node.LFS.verify(image)` will check that the dump format is the current version and return this checksum.  (Special request from @HHHartmann.) It can optionally checksum the image.  The checksum is probably worth doing before reloading LFS

#### Cache coherence.

I currently do a botch to flush the ICACHE, and that is to read a sequential 32Kb address window in flash.  @jmattsson: Q: do you know a better way?

TerryE:
Fixes #3206 (part)

### Notes

-  This is a major upgrade to LFS for implement on-ESP building of LFS regions.  I have had to do some major rework of the dump / undump architecture to achieve this, and we can discuss this later. 
-  This in a initial tranche for review and comment.  I have done basic testing, but still have a lot more use cases to enumerate, so this should be considered alpha code at this stage.  I also have documentation changes and some peripheral API stuff (such as #3145) to implement.
-  I have tagged Johny and Nathaniel as reviews, as they are most familiar with the NodeMCU Lua internals.  However, any constructive input is welcome.
-  The basic functionality also works in the `luac.cross -e` environment.  Create a script containing `debug.debug()` and execute this to make a basic interpreter available.  `os` and `io` are available and there are (HOST only) base functions `lfsreload` and `lfsindex`.

### DIFF
--- a/app/lua53/host/luac.c
+++ b/app/lua53/host/luac.c
@@ -18,38 +18,35 @@
 #include <time.h>
 
 #include "lua.h"
-#include "lualib.h"
-#include "lauxlib.h"
-#include "ldebug.h"
+#include "lualib.h"       // ??
+#include "ldebug.h"       // ??
 #include "lnodemcu.h"
 #include "lobject.h"
 #include "lstate.h"
-#include "lstring.h"
+#include "lstring.h"      // ???
 #include "lundump.h"
 
-static void PrintFunction(const Proto* f, int full);
-#define luaU_print	PrintFunction
+static void PrintFunction(const Proto *f, int full);
+#define luaU_print PrintFunction
 
-#define PROGNAME	"luac.cross"	/* default program name */
-#define OUTPUT		PROGNAME ".out"	/* default output file */
+#define PROGNAME "luac.cross" /* default program name */ 
+#define OUTPUT PROGNAME ".out" /* default output file */
 
-static int listing=0;			/* list bytecodes? */
-static int dumping=1;			/* dump bytecodes? */
-static int stripping=0;			/* strip debug information? */
-static char Output[]={ OUTPUT };	/* default output file name */
-static const char* output=Output;	/* actual output file name */
-static const char* progname=PROGNAME;	/* actual program name */
-static int flash = 0;			/* output flash image */
-static lu_int32 address = 0;		/* output flash image at absolute location */
-static lu_int32 maxSize = 0x40000;	/* maximuum uncompressed image size */
-static int lookup = 0;			/* output lookup-style master combination header */
-static const char *execute;		/* executed a Lua file */
+static int listing = 0;             /* list bytecodes? */
+static int dumping = 1;             /* dump bytecodes? */
+static int stripping = 0;           /* strip debug information? */
+static char Output[] = { OUTPUT };  /* default output file name */
+static const char *output = Output; /* actual output file name */
+static const char *progname = PROGNAME; /* actual program name */
+static lu_int32 address = 0;        /* output flash image at absolute location */
+static lu_int32 maxSize = 0x40000;  /* maximuum uncompressed image size */
+static const char *execute;         /* executed a Lua file */
 char *LFSimageName;
+
 
 #define IROM0_SEG    0x40200000ul
 #define IROM0_SEGMAX 0x00100000ul
-#define IROM_OFFSET(a) (cast(lu_int32, (a)) - IROM0_SEG)
-
+#define IROM_OFFSET(a)(((lu_int32) (a)) - IROM0_SEG)
 
 static void fatal(const char *message) {
   fprintf(stderr, "%s: %s\n", progname, message);
@@ -62,7 +59,7 @@ static void cannot(const char *what) {
 }
 
 static void usage(const char *message) {
-  if ( *message == '-')
+  if (*message == '-')
     fprintf(stderr, "%s: unrecognized option '%s'\n", progname, message);
   else
     fprintf(stderr, "%s: %s\n", progname, message);
@@ -72,100 +69,101 @@ static void usage(const char *message) {
     "  -l       list (use -l -l for full listing)\n"
     "  -o name  output to file 'name' (default is \"%s\")\n"
     "  -e name  execute a lua source file\n"
-    "  -f       output a flash image file\n"
     "  -F name  load a flash image file\n"
     "  -a addr  generate an absolute, rather than "
-               "position independent flash image file\n"
+    "position independent flash image file\n"
     "           (use with -F LFSimage -o absLFSimage to "
-               "convert an image to absolute format)\n"
-    "  -i       generate lookup combination master (default with option -f)\n"
+    "convert an image to absolute format)\n"
     "  -m size  maximum LFS image in bytes\n"
     "  -p       parse only\n"
-    "  -s       strip debug information\n"
+    "  -s n     strip debug information\n"
     "  -v       show version information\n"
     "  --       stop handling options\n"
     "  -        stop handling options and process stdin\n", progname, Output);
   exit(EXIT_FAILURE);
 }
 
-#define IS(s)	(strcmp(argv[i],s)==0)
+#define IS(s)(strcmp(argv[i], s) == 0)
 
 static int doargs(int argc, char *argv[]) {
   int i;
   int version = 0;
   lu_int32 offset = 0;
+  char *p;
   if (argv[0] != NULL && *argv[0] != 0) progname = argv[0];
   for (i = 1; i < argc; i++) {
-    if ( *argv[i] != '-') {                       /* end of options; keep it */
+    if (*argv[i] != '-') {     /* end of options; keep it */
       break;
-    } else if (IS("--")) {                         /* end of options; skip it */
+    } else if (IS("--")) {    /* end of options; skip it */
       ++i;
       if (version) ++version;
       break;
-    } else if (IS("-")) {                        /* end of options; use stdin */
+    } else if (IS("-")) {     /* end of options; use stdin */
       break;
-    } else if (IS("-e")) {                  /* execute a lua source file file */
+    } else if (IS("-e")) {    /* execute a lua source file file */
       execute = argv[++i];
       if (execute == NULL || *execute == 0 || *execute == '-')
         usage("\"-e\" needs a file argument");
-    } else if (IS("-F")) {                  /* execute a lua source file file */
+    } else if (IS("-F")) {    /* execute a lua source file file */
       LFSimageName = argv[++i];
       if (LFSimageName == NULL || *LFSimageName == 0 || *LFSimageName == '-')
         usage("\"-F\" needs an LFS image file argument");
-    } else if (IS("-f")) {                                /* Flash image file */
-      flash = lookup = 1;
-    } else if (IS("-a")) {                        /* Absolue flash image file */
-      flash = lookup = 1;
+    } else if (IS("-a")) {    /* Absolue flash image file */
+//      flash = lookup = 1;
       address = strtol(argv[++i], NULL, 0);
       offset = IROM_OFFSET(address);
       if (offset == 0 || offset > IROM0_SEGMAX)
         usage("\"-a\" absolute address must be valid flash address");
-    } else if (IS("-i")) {                                          /* lookup */
-      lookup = 1;
-    }  else if (IS("-l")) {                                           /* list */
+    } else if (IS("-l")) {    /* list */
       ++listing;
-    } else if (IS("-m")) {                    /* specify a maximum image size */
-      flash = lookup = 1;
+    } else if (IS("-m")) {    /* specify a maximum image size */
+//      flash = lookup = 1;
       maxSize = strtol(argv[++i], NULL, 0);
-      if (maxSize & 0xFFF)
+      if (maxSize &0xFFF)
         usage("\"-e\" maximum size must be a multiple of 4,096");
-    } else if (IS("-o")) {                                     /* output file */
+    } else if (IS("-o")) {    /* output file */
       output = argv[++i];
-      if (output == NULL || *output == 0 || ( *output == '-' && output[1] != 0))
+      if (output == NULL || *output == 0 || (*output == '-' && output[1] != 0))
         usage("'-o' needs argument");
       if (IS("-")) output = NULL;
-    } else if (IS("-p")) {                                      /* parse only */
+    } else if (IS("-p")) {    /* parse only */
       dumping = 0;
-    } else if (IS("-s")) {                         /* strip debug information */
-      stripping = 1;
-    } else if (IS("-v")) {                                    /* show version */
+    } else if (IS("-s")) {    /* strip debug information */
+      stripping = (unsigned) strtol(argv[i+1], &p, 0);
+      if (p == argv[i+1]) {
+        stripping = 1;        /* -s 1 if default if -s applied */
+        i++;
+      }
+      if (stripping > 2)
+        usage("'-s' level must be 0 - 2");
+    } else if (IS("-v")) {    /* show version */
       ++version;
-    } else {                                                /* unknown option */
+    } else {                  /* unknown option */
       usage(argv[i]);
     }
   }
 
-  if (offset>0 && (output == NULL || LFSimageName == NULL ||
-                   execute != NULL || i != argc))
+  if (version) {
+    printf("%s\n", LUA_COPYRIGHT);
+    if (version == argc - 1) exit(EXIT_SUCCESS);
+  }
+  if (offset > 0 && (output == NULL || LFSimageName == NULL ||
+      execute != NULL || i != argc))
     usage("'-a' also requires '-o' and '-f' options without lua source files");
 
   if (i == argc && (listing || !dumping)) {
     dumping = 0;
     argv[--i] = Output;
   }
-  if (version) {
-    printf("%s\n", LUA_COPYRIGHT);
-    if (version == argc - 1) exit(EXIT_SUCCESS);
-  }
   return i;
 }
 
-static const char *corename(lua_State *L, const TString *filename, int *len) {
+static const char *corename(lua_State *L, const TString *filename, size_t *len) {
   const char *fn = getstr(filename) + 1;
   const char *s = strrchr(fn, '/');
   if (!s) s = strrchr(fn, '\\');
   s = s ? s + 1 : fn;
-  while ( *s == '.') s++;
+  while (*s == '.') s++;
   const char *e = strchr(s, '.');
   if (len)
     *len = e ? e - s : strlen(s);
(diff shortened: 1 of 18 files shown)


app/lua53/host/luac.c
```diff
@@ -72,282 +69,195 @@ static void usage(const char *message) {
...
-  for (j = 0; j < 2 ; j++) {
+  //if (address)
+  //  luaN_setabsolute(address);
+
+  do {
+    request_restart = 0;
     L = luaL_newstate();
-    if (L == NULL) fatal("not enough memory for state");
+    if (L == NULL)
+      fatal("not enough memory for state");
```
nwf, line 242: There are a great many "stylistic" changes like the above three lines of diff in these commits.  Do they increase or decrease the diff between "upstream Lua" and "nodemcu Lua"?  Most of them appear to my mind to be improvements in the code readability; if they increase the gap between "upstream" and "nodemcu", can we send them upstream? :)

app/lua53/host/luac.c
```diff
@@ -359,256 +269,286 @@ int main(int argc, char *argv[]) {
         lua_close(L);
         continue; /* and loop around once more simulating restart */
       }
-      char *err = strdup(lua_tostring(L, -1));
+      char *
```
nwf, line 272: This appears to be some violence.  What happened to the rest of the line?

app/lua53/host/luac.c
```diff
@@ -359,256 +269,286 @@ int main(int argc, char *argv[]) {
...
+    case '\\': printf("\\\\"); break;
+    case '\a': printf("\\a");  break;
+    case '\b': printf("\\b");  break;
+    case '\f': printf("\\f");  break;
+    case '\n': printf("\\n");  break;
+    case '\r': printf("\\r");  break;
+    case '\t': printf("\\t");  break;
+    case '\v': printf("\\v");  break;
+    default:
+      printf(isprint(c) ? "%c" : "\\%03d", c);
```
nwf, line 344: While "obviously" equivalent, this still takes a moment to check, so I'd advocate either for reverting to match mainline or proposing this change to mainline for their inclusion.  Many of the following changes also have the same feel.

app/lua53/lauxlib.c
```diff
@@ -768,8 +768,9 @@ static int skipcomment (LoadF *lf, int *cp) {
 
 
 LUALIB_API int luaL_loadfilex (lua_State *L, const char *filename,
-                                             const char *mode) {
+                                             const char *mode, int multi) {
   LoadF lf;
+  (void)(multi);  ////  TODO
```
nwf, line 773: Yea, I think I'm content with `multi` waiting for a later PR. :)

app/lua53/lauxlib.c
```diff
@@ -1153,6 +1154,8 @@ static int errhandler_aux (lua_State *L) {
 ** to convert into an error message which it handles in a separate posted task.
 */
 static int errhandler (lua_State *L) {
+  if (!strncmp(lua_tostring(L, -1), "!LFS reload!", sizeof("!LFS reload!") - 1))
```
nwf, line 1157: It's not really worth using `strncmp` when one of the parameters is a constant; go ahead and use `strcmp` without the repetition.  The compiler's going to inline it anyway, one imagines.

nwf (changes requested):
Some initial nits; I haven't pulled the code to play with it.

app/lua53/lauxlib.c
TerryE: Yup `memcmp` is better.  We can't use `strcmp` because the various reload messages include a qualification.

app/lua53/host/luac.c
TerryE: I will post a general discussion in #3206.  Have read and review in that light.

TerryE:
As a codicil to my #3206 comments, I do suggest that reviewers print off and go through the three main rewrites: `ldump.c`, `lundump.c`, `lnodemcu.c`

app/lua53/host/luac.c
TerryE: `luac.c` needs a lot more work largely because the unification of the LFS  and LC formats.  Given the volume of change and the previous nasty non-standard formatting of lauc.c which is nothing like the rest of the lua core codebase, I have been considering running it through a pretty print style formatter to make it more readable anyway.

TerryE:
I will be adding #3145 to this PR as this is needed for on-ESP LFS maintenance.

app/lua53/host/luac.c
```diff
@@ -72,282 +69,195 @@ static void usage(const char *message) {
...
+    if (L == NULL)
+      fatal("not enough memory for state");
     lua_pushcfunction(L, &pmain);
     lua_pushinteger(L, argc);
     lua_pushlightuserdata(L, argv);
     status = lua_pcall(L, 2, 0, 0);
+    err = (status != LUA_OK) ? strdup(lua_tostring(L, -1)) : NULL;
+    lua_gc(L, LUA_GCCOLLECT, 0);
+    lua_close(L);
+  } while (request_restart);
```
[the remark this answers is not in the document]
jmattsson: Edit: Ah, I did - it's an `extern int`. Why doesn't Github allow me to delete a pending comment? I just get a cat picture when I click the `...` now :(

app/lua53/lnodemcu.c
```diff
@@ -166,47 +183,69 @@ static void platform_s_flash_write(const void *from, lu_int32 to, lu_int32 len)
...
+  lu_int32 seed;          /* random number seed used in LFS */
+  lu_int32 timestamp;     /* timestamp of LFS build */
+  lu_int32 nROuse;        /* number of elements in ROstrt */
+  lu_int32 nROsize;       /* size of ROstrt */
+  lu_int32 oROhash;       /* offset of TString ** ROstrt hash */
+  lu_int32 protoROTable;  /* offset of master ROTable for proto lookup */
+  lu_int32 protoHead;     /* offset of linked list of Protos in LFS */
+  lu_int32 shortTShead;   /* offset of linked list of short TStrings in LFS */
+  lu_int32 longTShead;    /* offset of linked list of long TStrings in LFS */
+  short    nFiles;
```
jmattsson, line 215: How come `short` rather than an explicit 16bit type?

app/lua53/lnodemcu.c
```diff
@@ -166,47 +183,69 @@ static void platform_s_flash_write(const void *from, lu_int32 to, lu_int32 len)
...
+  short    nPVs;
+};
 
-//===================== NodeMCU lua.h API extensions =========================//
 
+typedef struct LFSflashState {
+  lua_State  *L;
+  LoadState  *S;
+  int         pass;               /* 1 or 2; also 0 implies 1 */
```
jmattsson, line 223: Should we be explicit about integer width here too (and in `nSave` below), for consistency if nothing else?

app/lua53/lundump.c
```diff
@@ -1,637 +1,876 @@
...
+
+#define FLASH_WRITE(S,b,l) S->flashWriter(S->flashWriterData, b, l)
+#define FLASH_SYNC(S) FLASH_WRITE(S, NULL, ~(size_t)0)
+#define FLASH_CURR(S) FLASH_WRITE(S, NULL, 0)
+
+/*
+** The following compression maps are used in Store() ONLY in the case of
+** 64-bit hosts to repack structs to adjust for 64-bit->32-bit points in
+** absolute mode LFS creation.  These must match the definitions in lobject.h
+*/
```
jmattsson, line 275: Thanks Terry! :)

app/lua53/lundump.c
```diff
@@ -1,637 +1,876 @@
...
+  lu_byte *dest;                              /* returns addr of dest element */
+  if (inRAM(S)) {
+    dest = cast(lu_byte *, vec) + (ndx*len);
+    if (dest != src)
+      memcpy(dest, src, len);
+#ifdef LUA_USE_HOST64
+  } else if (S->mode == MODE_LFSA && pack) {
+   /*
+    * 64-bit hosts have two LFS modes. The default LFS mode is for writing to a
+    * when writing to native LFS where the address pointers are 64 bit and any
```
jmattsson, line 316: Is there a line missing here? Or is it just a mangled comment?

jmattsson (approved):
Phew! Big stuff!
Overall looks good. See review comments for some qs and comments.
I haven't reviewed the dump/undump logic in depth - I'm not in a headspace to manage that right now, and besides I'm figuring you'll have caught any nasty surprises in your testing anyway :)

app/lua53/host/luac.c
```diff
@@ -18,38 +18,35 @@
...
-#include "lualib.h"
-#include "lauxlib.h"
-#include "ldebug.h"
+#include "lualib.h"       // ??
+#include "ldebug.h"       // ??
 #include "lnodemcu.h"
 #include "lobject.h"
 #include "lstate.h"
-#include "lstring.h"
+#include "lstring.h"      // ???
```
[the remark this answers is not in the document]
TerryE: Johny, in terms of this and all of your `luac.c` comments, I get the impression that the original `luac.c` was a "throw-in" not developed by the original Lua developers.  The coding style and layout is quite different to the core Lua code. We are getting to the stage where this is heavily modified because of our LFS, Absolute LFS and execution environment support.  I've pretty much decided to give this a complete hose down starting with a pretty print reformat, and getting rid of the some of the coding nasties.  I'll add this to the next push.  I will also reflect your general comments.

app/lua53/lnodemcu.c
TerryE: I will make these a `lu_int16` type.  The reason for the 16-bit was that I wanted to keep the header the same size if possible and need to pass and additional two integer parameters from  pass 1 to pass 2 but only had 4 bytes reserved.  Both are small integers so 2 × lu_int16 is fine.

app/lua53/lnodemcu.c
TerryE: Not really.  This an internal state structure (which is private to `nodemcu.c`.  This pattern just follows Lua coding style which uses `int` in its structures for generic integer.  The reason for the `lu_int32` declarations is that these must be 32bit.

app/lua53/lundump.c
TerryE: mangled will rewrite

app/lua53/lapi.c
```diff
@@ -1000,6 +1000,17 @@ LUA_API int lua_pcallk (lua_State *L, int nargs, int nresults, int errfunc,
   return status;
 }
 
+static void aux_load(lua_State *L, LClosure *f) {
+  if (f->nupvalues >= 1) {  /* does it have an upvalue? */
```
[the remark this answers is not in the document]
TerryE: Not really; just hoisted some existing code which is being reused.  Will use `> 0` comparison.  Unlike Lua 5.1 where upvals are initialised on load, in Lua 5.3 the first (if it exists) is assumed to be `_ENV` and is initialised.

galjonsfigur:
As I'm not that knowledgeable in Lua internals I can only test this PR. My observations:
- My GCC (10.1.1) returned some warnings when compiling this code - I don't want to paste the whole log here so I uploaded the important parts to [this Gist](https://gist.github.com/galjonsfigur/83e11d7f7f0c164e1ef1045aae1d0b15)
- The `-f` option from `luac.cross` could be depreciated first instead of being removed
- Old way of using LFS with `_init.lua`won't work even with changes from #3168 - but it's not really needed with new LFS API.

TerryE:
@galjonsfigur, thanks I will add these to the TODO list.

app/lua53/host/luac.c
```diff
@@ -72,282 +69,195 @@ static void usage(const char *message) {
...
-    "  -s       strip debug information\n"
+    "  -s n     strip debug information\n"
     "  -v       show version information\n"
     "  --       stop handling options\n"
     "  -        stop handling options and process stdin\n", progname, Output);
   exit(EXIT_FAILURE);
 }
 
-#define IS(s)	(strcmp(argv[i],s)==0)
+#define IS(s)(strcmp(argv[i], s) == 0)
```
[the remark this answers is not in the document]
TerryE: Sorry pretty print reformatting did this.  I'll go through these.

app/lua53/host/luac.c
```diff
@@ -18,38 +18,35 @@
...
+static lu_int32 maxSize = 0x40000;  /* maximuum uncompressed image size */
+static const char *execute;         /* executed a Lua file */
 char *LFSimageName;
 
+
 #define IROM0_SEG    0x40200000ul
 #define IROM0_SEGMAX 0x00100000ul
-#define IROM_OFFSET(a) (cast(lu_int32, (a)) - IROM0_SEG)
-
+#define IROM_OFFSET(a)(((lu_int32) (a)) - IROM0_SEG)
```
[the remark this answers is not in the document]
TerryE: I personally prefer `cast()` for readability.  I am not sure of the rationale here, but it is defined in `limits.h` and this is typically pathed in via `lobject.h` which is included in most of the `LUA_CORE` modules but not the `LUA_LIB` so
-  the Lua library modules (`l*lib.c`) -- i.e.the ones that interface to the core through the `lua.h` API -- **don't** use it, but 
-  the Lua core itself which access the VM internals, internal data structures, etc., do.   

Standard `luac.c` (even though it includes `object.h` for the code listing function) treats itself as non-core and therefore doesn't use cast().  Confused?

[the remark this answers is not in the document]
jmattsson: > Confused?

Yup! 😄

app/lua53/lauxlib.c
nwf: "include a qualification"?

galjonsfigur:
> Are you suggesting that this is retained as a deprecated synonym for -o?

Sorry for late reply - didn't notice it. As now `-f` option is redundant it could just print deprecation notice - that way it won't interfere with any existing scripts that use `luac.cross`

Issue #3164, opened 2020-06-14
Motion to rename master branch


nwf:
I'd like to propose that we use `release` instead of the somewhat problematic term `master`.  `dev` can stay as `dev`.

marcelstoer:
👎 

> instead of the somewhat problematic term `master`

What makes this problematic?

It would be very unusual and completely unexpected for a Git repo to _not_ have a master branch. It's certainly doable, `master` is nothing but a globally respected convention, but I'm afraid of the consequences and I see no benefit.

nwf:
https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.1

For our repository, the name is, at the very least, indicative of the wrong thing: in most git projects, the `master` branch is the one from which the releases are cut, i.e., what we call `dev`.

pjsg:
I thought that we cut releases from `master`. We do development on a long running `dev` branch that should be strictly a number of commits on top of the most recent `master` release. As I understand the normal release process is to do a monster PR from `dev` to `master` (which is fast-forward) and then tag `master` with the release version tag. 

Some repos just do all the development on `master`, but this is problematic if you want to keep `master` always working, and even more problematic if you sometimes want to cherrypick important bug fixes onto master. If there are any bug-fixes that go onto `master`, then you do a merge from `master` into `dev` sometime before you do the release.

marcelstoer:
@pjsg your observations in paragraph one are correct.

nwf:
Look, I dare say I understand both how git works and how projects in general and in particular NodeMCU historically use git's features.  Please give me *some* credit?

Our releases are just tags; tags in git are not attached to any branch, as they just point to a particular nodes in the commit DAG.  (Indeed, branch heads are also just tags; commits do not "know" they are on a branch.)  It happens that, yes, we always ensure that the so-designated commits occur on the (erroneously, problematically named) `master` branch, but we could expunge every reference to the `master` branch from git's data structures and our releases would still continue to name the existing objects.

[...]

Issue #3191, opened 2020-07-02
ftpserver mutated without adapting  documentation - breaking change


HHHartmann:
### Expected behavior
code works as documented

### Actual behavior
code fails with errormessage:

```
node.flashindex('ftpserver')().createServer("test","12345",true)
Lua error: 	...itHub\nodemcu-firmware\lua_modules\ftp\ftpserver.lua:70: attempt to index a string value
stack traceback:
	...itHub\nodemcu-firmware\lua_modules\ftp\ftpserver.lua:70: in function 'createServer'
	stdin:1: in main chunk
	[C]: ?
	[C]: ?
```

### Test code
Working code:
```Lua
node.flashindex('ftpserver')():createServer("test","12345",true)
```
Note the ':' instead of '.' before createServer

Documentation says:
### NodeMCU startup banner
This is on current dev and master branch

NodeMCU 3.0.0.0 
	branch: pr_3158
	commit: fd8ef76b0ae71dde56f0fb853faa2763520ebdea
	release: 3.0-master_20200610 +11
	release DTS: 
	SSL: false
	build type: float
	LFS: 0x40000 bytes total capacity
	modules: adc,bit,color_utils,crypto,dht,encoder,enduser_setup,file,gpio,http,i2c,l3g4200d,mdns,net,node,ow,pipe,pixbuf,pwm,rtcfifo,rtcmem,rtctime,sjson,sntp,spi,struct,tmr,uart,wifi,wifi_monitor,ws2812,ws2812_effects
 build 2020-06-26 19:39 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)

TerryE:
My bad.  Yes, this is a minor breaking change that needs documenting.  The old ftp server could only be run out of LFS which is an impediment for beginners.  I've made some tweaks so the file can be split into dynamically loaded separate LC files for top level methods and this runs fine out of RAM or the #3117 fast loader, _but_ this required you to pass the ftp table in as an arg hence the need for the object notation.

I will fix the documentation.

HHHartmann:
The bad thing is that this is on Master already. 
Do we have a policy for fixing issues on Master? 

Btw how can it be split?

TerryE:
We are talking about a few line change to the MD file. If @marcelstoer is OK with this, I'll just patch the master documentation. 

[...]

PR #3274, opened 2020-09-08
Next master drop

marcelstoer:
As said earlier I don't know whether this should land on master or release. @nwf your advice required here.


nwf:
I think we should land it on `release` and then bump `master` to match (easily done from the command line: `git push origin origin/release:master`) as we are currently keeping both labels for downstream consumers (notably RTD, right?).

marcelstoer:
Fine for me. Who's willing to approve and merge (but not squash 😉) this?

Oh, and I just noticed this will shrink our LoC significantly:  +2,942 −218,841 😄 👍

HHHartmann:
I would like to add a win luac.cross to the Abonnenten of the release. I did so for the last release but as a change which Akaik doesn't get sent out again

marcelstoer:
> nwf merged commit

Thanks.

>  I don't know whether this should land on master or release

I see we now have two tags: https://github.com/nodemcu/nodemcu-firmware/releases. We're certainly not going to write release notes twice but which one is it gonna be?

@HHHartmann sorry, I don't understand this remark.

HHHartmann:
@marcelstoer When cutting a new release we write a release note which has download links attached. Usually the source code.
To the last release I added, after the release was announced, a luac.cross.exe binary. I would like to add this right on the announcement, so that all receivers of it know that the download exists.
I built the binary locally on my machine, as I did now.
So how can we get this done. I can send it to whoever, as requested.

marcelstoer:
> So how can we get this done.

Attach it to your comment above. Once I get feedback on

> I see we now have two tags: https://github.com/nodemcu/nodemcu-firmware/releases. We're certainly not going to write release notes twice but which one is it gonna be?

(from @nwf maybe?) I can write release notes and announce the master drop. Would maybe then also allow us to close #3164?

Side note: just realized that #3191 is still open but should have been "included" before the last master drop.

nwf:
I'd go for writing the release notes on the `release` tag, and maybe add a forwarding link from the `master` tag citing #3164?

marcelstoer:
And https://github.com/nodemcu/nodemcu-firmware/#releases needs to be updated, too. I stopped publishing `master` to RTD and `/master` redirects to `/release`: https://nodemcu.readthedocs.io/en/master/

HHHartmann:
Since merging this there have been some fixes which might make sense to have in release rather sooner than later.

These are the file fix #3280 , not being able to red files of certain sizes, which I think is essential
and the two enduser_setup PRs #3275 and #3282 would be nice.

BTW: There is also a reference to master in the issue template.

marcelstoer:
@HHHartmann you may now attach your binaries to https://github.com/nodemcu/nodemcu-firmware/releases/tag/3.0-release_20200910

marcelstoer:
> BTW: There is also a reference to master in the issue template.

Sorry, didn't find it. Fortunately [GitHub uses the _default_](https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/about-issue-and-pull-request-templates#issue-templates) rather than the _master_ branch:

> Issue templates are stored on the repository's default branch

HHHartmann:
> @HHHartmann you may now attach your binaries to https://github.com/nodemcu/nodemcu-firmware/releases/tag/3.0-release_20200910

Done.
Maybe we can add it before sending out the announcement, so people will more likely spot it.
Should we add this to the docs somewhere?

HHHartmann:
> > BTW: There is also a reference to master in the issue template.
Its here: https://github.com/nodemcu/nodemcu-firmware/blob/release/.github/PULL_REQUEST_TEMPLATE.md 

marcelstoer:
Thanks, that'll eventually bubble up to the default branch.

PR #3275, opened 2020-09-10
Handle tcp fragmentation and also fix XSS problem.

### LINKED ISSUE #2931: enduser_setup "Handling POST" failure on 310faf7 and dev
### Expected behavior
Calling enduser_setup.start(), connecting trough wifi to captive portal at 192.168.4.1, scanning for AP, entering correct password will connect you to that AP.
### Actual behavior
You are returned to empty '192.168.4.1/setwifi'
some debug code on failure:

```lua
247: 	enduser_setup_check_station
270: 	status=0,chan=13
1469: 	enduser_setup_http_connectcb
1326: 	enduser_setup_http_recvcb
1043: 	enduser_setup_serve_status_as_json
870: 	enduser_setup_http_serve_header
380: 	deferred_close
360: 	handle_remote_close
247: 	enduser_setup_check_station
270: 	status=0,chan=13
1469: 	enduser_setup_http_connectcb
1326: 	enduser_setup_http_recvcb
1123: 	Handling POST
811: 	enduser_setup_http_handle_credentials
820: 	Password or SSID string not found
870: 	enduser_setup_http_serve_header
380: 	deferred_close
360: 	handle_remote_close
360: 	handle_remote_close
```
### Test code
After enabling
```C
#define LUA_USE_MODULES_ENDUSER_SETUP
```
fire
```lua
enduser_setup.start()
```
### NodeMCU version
commit 043046d2ab06fdefdae2a097c7c0f01cdb8da2fb (HEAD -> dev, origin/dev)

### Hardware
Generic ESP-07S

pjsg:
Fixes #2931.

Make sure that we accumulate the whole request before trying to execute it. Also, make sure that we escape <> characters in SSIDs so that we don't get a XSS. Also, don't allow an SSID to appear more than once in the dropdown.

I have tested by sending the request in little pieces and it appears to work. I'd like someone who has an ios device to test this and make sure that it works.


docs/modules/enduser-setup.md
```diff
@@ -6,13 +6,6 @@
 This module provides a simple way of configuring ESP8266 chips without using a
 serial interface or pre-programming WiFi credentials onto the chip.
 
-!!! attention "ATTENTION Apple users"
-
-	Due to bug [#2931](https://github.com/nodemcu/nodemcu-firmware/issues/2931) the configuration does currently not work for many Safari browsers (iOS & macOS).
-	As a **workaround** there is alternative HTML file which uses another method to transfer the login credentials. It does not support sending arbitrary additional configuration parameters and likewise no `eus_params.lua` will be written. The WiFi credentials will be stored in the ESP flash.
-
-	Just copy [enduser_setup_apple.html](../../app/modules/enduser_setup/enduser_setup_apple.html) to the ESP file system and rename it to `enduser_setup.html`.
```
marcelstoer, line 14: If we confirm that your changes fix the issue then you can delete `enduser_setup_apple.html`.

KT819GM: I've checked now on iPhone. Worked as expected.

app/modules/enduser_setup.c
```diff
@@ -1333,27 +1347,84 @@ static err_t enduser_setup_http_recvcb(void *arg, struct tcp_pcb *http_client, s
...
+  pbuf_copy_partial(p, hrb->data + hrb->length, p->tot_len, 0);
+  hrb->data[newlen] = 0;
+  hrb->length = newlen;
 
-  unsigned data_len = pbuf_copy_partial (p, data, p->tot_len, 0);
   tcp_recved (http_client, p->tot_len);
   pbuf_free (p);
 
+  // see if we have the whole request.
+  // Rely on the fact that the header shoyld not contain a null character
```
nwf, line 1397: Spelling

app/modules/enduser_setup.c
```diff
@@ -1333,27 +1347,84 @@ static err_t enduser_setup_http_recvcb(void *arg, struct tcp_pcb *http_client, s
...
+        field++;
+      }
+      if (*field == ':') {
+        size_t extra = strtol(field + 1, 0, 10);
+        if (extra + end_of_header - hrb->data > hrb->length) {
+          return ERR_OK;
+        }
+      }
+    }
+  }
```
nwf, line 1421: And if there isn't a Content-Length header?

nwf (changes requested):

pjsg:
I also fixed the bogus generate_204 so that the captive portal detection is triggered on Android and on Macos. This is a really nice usability improvement that the OS just drops you straight into a browser on the right page. No messing around. I'm also wondering if this should just out by doing the scan rather than require the user to press the button.

app/modules/enduser_setup.c
nwf: That seems fishy to me.  https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4 seems, to my very naive mind, to suggest that we should be bailing with 411 Length Required because, well, we probably require a length if this isn't a GET request, or something like that.

nwf:
Thanks.  Forgive one query about a possible pungent poisson, but surely this should land even if it's not exactly spec conformant because it's way less wrong than what we had before.

app/modules/enduser_setup.c
marcelstoer: > we should be bailing with 411 Length Required because, well, we probably require a length if this isn't a GET request

I second that.

app/modules/enduser_setup.c
```diff
@@ -105,12 +105,26 @@ static const char http_html_gzip_contentencoding[] = "Content-Encoding: gzip\r\n
 /* Externally defined: static const char enduser_setup_html_default[] = ... */
 #include "enduser_setup/enduser_setup.html.gz.def.h"
 
+#define SCAN_LISTENER_MAGIC  1
+#define HTTP_REQUEST_BUFFER_MAGIC  2
```
marcelstoer, line 109: How about adding a comment to these two to describe their purpose - or giving them an expressive name?  "magic" could mean anything. Same goes for the `magic` variables in the structs below.

app/modules/enduser_setup.c
```diff
@@ -1333,27 +1347,84 @@ static err_t enduser_setup_http_recvcb(void *arg, struct tcp_pcb *http_client, s
     return ERR_ABRT;
   }
 
+  tcp_arg_t *l = arg;
```
marcelstoer, line 1350: The legibility of this code would be improved if this thing had a real name.

marcelstoer (commented):
Thank you Philip!

I only have Apple devices to test and it does work well on them (once I remembered to set DNS back to automatic from 1.1.1.1). However, macOS and iOS never displayed this auto-captive-portal window  into which they normally load the portal page. As I sometimes see this behavior with other captive portals as well I'm not too worried. I never researched if there are any specifics the portal needs to return to trigger this window on Apple devices.

pjsg:
I reverted the change to make it trigger captive portal detection -- that will follow in its own PR with some other minor improvements.

app/modules/enduser_setup.c
pjsg: The browser can always use chunked encoding which removes the need for the content-length header. I've now added code to handle that case. I *doubt* it will happen since the post parameters are small.

I didn't want to make the code that checked whether the entire request had arrived depend on the HTTP method being used. This way, it will correctly determine whether the entire request has been received. The individual handlers can then operate on the entire request.

marcelstoer:
While you're at it could you add `<meta charset="utf-8"/>` to the HTML head to appease browsers which need a content-type declaration (or send the respective HTTP header)?

app/modules/enduser_setup.c
pjsg: Done

marcelstoer (approved):

PR #3282, opened 2020-09-14
Improve the enduser setup experience by triggering captive portal detection.

### LINKED ISSUE #3281: Improve enduser_setup user experience
### Missing feature

The enduser_setup feature should take advantage of the captive portal detection to automatically show the setup screen when the user attaches to the network.

The enduser_setup page should automatically trigger the network scan and should require entry/non-entry of the password depending on whether the authmode is 0 or not.

Additionally, it would be nice to be able to pass in the SSID to use for the AP into the `start` method.

pjsg:
Fixes #3281.

This gets rid of the generate_204 handling and now makes sure to trigger captive portal handing on (at least) Android and MacOs. Testing on other platforms would be great. This also auto triggers the search for networks as soon as the page is displayed. It also ensures that you enter a password for the network (if the network requires authentication).


marcelstoer:
As I just merged #3275 this should now be rebased on dev.

docs/modules/enduser-setup.md
```diff
@@ -137,9 +131,10 @@ Starts the captive portal.
 *Note: Calling start() while EUS is already running is an error, and will result in stop() to be invoked to shut down EUS.*
 
 #### Syntax
-`enduser_setup.start([onConnected()], [onError(err_num, string)], [onDebug(string)])`
+`enduser_setup.start([AP_SSID,] [onConnected()], [onError(err_num, string)], [onDebug(string)])`
 
 #### Parameters
+ - `AP_SSID` the (optional) SSID to use for the AP. This defaults to the ENDUSER_SETUP_AP_SSID plus the device id.
```
marcelstoer, line 137:
```suggestion
 - `AP_SSID` the (optional) SSID to use for the AP. This defaults to `SetupGadget_<device id>`.
```

app/modules/enduser_setup.c
```diff
@@ -1432,15 +1588,10 @@ static err_t enduser_setup_http_recvcb(void *arg, struct tcp_pcb *http_client, s
...
-    else if (strncmp(data + 4, "/generate_204", 13) == 0)
-    {
-      /* Convince Android devices that they have internet access to avoid pesky dialogues. */
-      enduser_setup_http_serve_header(http_client, http_header_204, LITLEN(http_header_204));
-    }
     else
     {
-      ENDUSER_SETUP_DEBUG("serving 404");
-      enduser_setup_http_serve_header(http_client, http_header_404, LITLEN(http_header_404));
+      // All other URLs redirect to / -- this triggers captive portal.
```
marcelstoer, line 1593: > redirect to /

Not anymore 😄 You changed this to `http://nodemcu.portal/`.

marcelstoer:
Tested on macOS and iOS - works perfectly! Thanks for the new SSID support in `enduser_setup.start()`.

What do you think about changing the default SSID prefix from `SetupGadget_` to something more...intuitive like `NodeMCU_`?

app/modules/enduser_setup.c
pjsg: Yes -- I can fix that. It turned out that redirecting to / did not trigger captive portal on MacOs.

pjsg:
I'd change the default string to Nodemcu -- not sure how others would feel.

app/modules/enduser_setup/enduser_setup.html
```diff
@@ -315,13 +320,24 @@ <h4 id='st'>Updating Status...</h4>
...
+                            pw.disabled = true;
+                            pw.required = false;
+                          }
 			};
 			$('#bk2').onclick = function () {
 				cur('#f1')
 			}
 			rs = to(refr, 0.5);
 			if( trying ) cur("#f3");
+                        refrAp();
```
marcelstoer, line 340:
```suggestion
                      refrAp();
```

pjsg:
I updated the default to be `NodeMCU` and I tested the captive portal detection with Windows 8 -- and it worked.

marcelstoer (approved):
Although I can't really assess the C code but this all looks sane.

PR #2269, opened 2018-02-20
Update to sdk 2.2

nwf:
Initial commit for
https://github.com/nodemcu/nodemcu-firmware/issues/2225 .

Replay patches from Espressif's repository at
https://github.com/espressif/ESP8266_NONOS_SDK between tags v2.1.0 and
v2.2.0:

	0001-sync-from-ccca00f2.patch

		Superseded by existing changes, but lines reordered in app/driver/key.c
		to minimize divergences.

	0002-sync-from-3f38ad5a.patch

		Upstream files only

	0003-Update-links.patch

		Not meaningful to NodeMCU

	0004-sync-from-01990ad0.patch
	0005-sync-from-cdf6877d.patch

		Upstream files only

	0006-sync-from-f29e744c.patch

		Upstream files only, user_interface.h override non-conflicting

	0009-feat-lwip-Move-lwip-source-code-to-third_party-folde.patch

		Merged change to lwip/app/espconn_udp.c; rest is just moves or
		appears to not apply.

	0010-feat-mbedtls-Add-mbedtls-source-code-in-third_party-.patch

		Does not apply; we use our own mbedtls

	0011-added-C-support.patch

		Merged to Makefile

	0012-feat-mbedtls-Rebuild-libmbedtls.patch

		Already applied

	0013-fix-at-Fix-some-bugs-of-AT.patch

		Upstream files only

	0014-feat-err_t-Redefine-err_t-to-s32_t.patch

		Merged to app/include/arch/cc.h and ./app/include/lwip/app/espconn.h;
		the rest is upstream files.

	0015-fix-wpa-Fix-wpa-wpa2-ptk-gtk-reinstallation-vulnerab.patch
	0016-fix-wifi-Remove-group-key-entry-before-connecting-to.patch
	0017-feat-lib-Remove-time-function-in-libmain.patch

		Upstream files only

	0018-feat-espconn-Modification-for-espconn.patch

		Merged to app/include/lwip/app/espconn.h,
		app/include/lwip/app/espconn_tcp.h, app/lwip/app/espconn.c,
		app/lwip/app/espconn_tcp.c

	0019-feat-at-Use-new-espconn_recv-to-fix-tcp-server-issue.patch
	0020-feat-examples-Update-mqtt-demo-and-auto-bin-generate.patch

		Upstream files only

	0021-wifi-Add-scan-threshold-and-dwell-time.patch
	0022-feat-wifi-Add-country-code-API.patch
[...]

PR #3284, opened 2020-09-18
Upgrade SDK to 3.0.4

### LINKED ISSUE #3008: Upgrade to latest SDK 3.0.x
https://github.com/espressif/ESP8266_NONOS_SDK/releases/tag/v3.0.4 is possibly are more stable version of what we currently use.

marcelstoer:
Fixes #3008

@TerryE I'm not sure what to do about `$RELEASE` for `latest-3.0`. Isn't "latest 3.0" the same as 3.0.4 in the else-branch?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,28 +4,21 @@
 
 TOP_DIR:=$(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 
-# SDK base version, as released by Espressif depends on the RELEASE flag
+# SDK base version, as released by Espressif depends on the SDK_RELEASE flag
 #
-# RELEASE = lastest pulls the latest V3.0.0 branch version as at the issue of this make
-# otherwise it pulls the labelled version in the SDK version's release directory
+# SDK_RELEASE = master pulls the latest changes from the NONOS_SDK master
+# branch (contains all work-in-progress commits)
 #
-ifeq ("$(RELEASE)","latest-3.0")
-  SDK_VER        := 3.0.0
-  SDK_FILE_SHA1  := NA
-  SDK_ZIP_ROOT   := ESP8266_NONOS_SDK-release-v$(SDK_VER)
-  SDK_FILE_VER   := release/v$(SDK_VER)
-else ifeq ("$(RELEASE)","master")
+ifeq ("$(SDK_RELEASE)","master")
   SDK_VER        := master
   SDK_FILE_SHA1  := NA
   SDK_ZIP_ROOT   := ESP8266_NONOS_SDK-$(SDK_VER)
   SDK_FILE_VER   := $(SDK_VER)
 else
-# SDK_VER        := 3.0
-# SDK_FILE_VER   := v$(SDK_VER)
-  SDK_FILE_VER   := e4434aa730e78c63040ace360493aef420ec267c
-  SDK_VER        := 3.0-e4434aa
-  SDK_FILE_SHA1  := ac6528a6a206d3d4c220e4035ced423eb314cfbf
-  SDK_ZIP_ROOT   := ESP8266_NONOS_SDK-$(SDK_FILE_VER)
+  SDK_VER        := 3.0.4
+  SDK_FILE_SHA1  := 24c702348ee5b9ae3ae640c1aead60c97768c881
+  SDK_ZIP_ROOT   := ESP8266_NONOS_SDK-$(SDK_VER)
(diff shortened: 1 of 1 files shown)


TerryE:
> @TerryE I'm not sure what to do about `$RELEASE` for `latest-3.0`. Isn't "latest 3.0" the same as 3.0.4 in the else-branch?

At the moment this is a bit of a mess.  With this PR, the `wget` pulls `archive/$(SDK_FILE_VER).zip` with one of three file versions:
-  `release/v3.0.0.zip` (`RELEASE=latest-3.0"`) This is a 3.0.0 release and we haven't been building against this for perhaps a year.  IMO, this is obsolete and needs removed.
-  `master.zip` (`RELEASE=master"`).  Note that master doesn't have the usage here as in our project.  Master is actually more closely aligned to our `dev` as this is where all work-in-progress commits are made. 
-  `v3.0.4.zip` (default).  This is the latest 3.0 release (and closer to our `release` staging.  This PR up this from an interim 3.0.1 commit to the tagged 3.0.4.  Note that the 3.0.x release tags are labelled versions of master. (3.0.4 was a commit on 27 May 2020; the latest master commit was on 8 June.

Whatever we do this parameter should not be called `RELEASE`, but rather something like `SDK_RELEASE`.  The default for this should be `v3.0.4` and we define the SHA1 for this default.  Any non-default should be as-is and the SHA1 should be `"NA"`.

marcelstoer:
> At the moment this is a bit of a mess.

Exactly 😞  Thanks for taking the time to describe the three options I noticed in the code but was too lazy to list.

I also wanted to dump the `RELEASE=latest-3.0` branch. Good idea to rename `RELEASE` to `SDK_RELEASE`.

nwf:
It looks like there are a lot of changes in Espressif's tree that should be applied to ours as well.  This will require a lot more than just bumping some strings in the `Makefile`, I'm afraid; it looks rather like #2269 again.  At a glace, it looks like at least these need to be ported over (I stopped reading the list when this got long enough; *this is not even an exhaustive skim*):

* https://github.com/espressif/ESP8266_NONOS_SDK/commit/cc2270a65425e7dbbf212402f00e7c3b34afd4a0
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/d3523192465a7dcefc65115fa9104a82c66afe6f
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/af2aebb01b78a82d8861b1f3520e3a06dba8eb54 (may be superseded by local dns cache management changes)
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/10977b970fc14b7e0ac0f70c3cd3ee1743843d9a will have no net consequences but should be lifted to keep our lwip fork in line
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/23a2c54762127193765bf0df1f86ee381f755840
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/06c1241f18387e9b43e95c93e3ca982cb33723f8
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/db72f79aff582ec313ea5b7fed0ff5a6475634d6
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/959a873f21f06b28fe1156f765b6ccce672465b4
* https://github.com/espressif/ESP8266_NONOS_SDK/commit/d4744e873c9a7c74ea6cd75f1378c38bcfb85cee maybe?

marcelstoer:
> This will require a lot more than just bumping some strings in the `Makefile`

I thought so...remembering from the past that SDK upgrades (even at patch level) were major undertakings.

I had the constant nagging feeling that I didn't actually know what I was doing here with this PR 🙈 I can close it or leave it open so that whoever will do the job I'm not capable of can at least grab the first commit.

nwf:
We are in a very frustrating position wrt Espressif, and it bums me out.  I'll add doing the LwIP updates to my to-do list, but don't expect progress until this weekend at the earliest.  FWIW, the `Makefile` changes do seem fine :+1:.

TerryE:
The nub of Nathaniel's comment is something that we've mentioned before: we forked LWiP, etc. a few years ago when Espressif first posted an open source stack as an alternative to the then closed version shipped with the SDK.

They subsequently switched to the OS version themselves and have done as range of bug fixes which our fork doesn't reflect. 

I am not sure that we can easily revert back to the Espressif version, but the exercise that N proposes would be a good start. It's a few days work though.

PR #3289, opened 2020-09-21
add support for sw_version in node.info()

### LINKED ISSUE #2759: node.info() missing for ESP32
on esp32
**=node.info()** command
returns:
**stdin:1: attempt to call field 'info' (a nil value)**

kit-klein:
Fixes #2759 

This PR adds basic support for the previously missing node.info() method on esp32.  It is modeled after the newer `group` properties design found in the 8266 codebase. At this time only the `sw_version` group is implemented.

I also partially ported the `update_buildinfo.sh` script to expose build info associated with the `sw_version` group.  

Autoformatter cleaned up a few rogue spaces in the .md file before I turned it off.  This is why some non-related edits exist in the docs.

### DIFF
--- a/components/modules/node.c
+++ b/components/modules/node.c
@@ -15,6 +15,7 @@
 #include "esp_vfs.h"
 #include "lnodeaux.h"
 #include "lflash.h"
+#include "user_version.h"
 
 // Lua: node.chipid()
 static int node_chipid( lua_State *L )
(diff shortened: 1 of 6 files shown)


marcelstoer:
@kit-klein thanks a lot!

> @heythisisnate reacted with heart emoji

❤️ great to see you're still following this project 😄 Just noticed you're using the ESP32 for your Alarm Panel Pro 👍 

To any committer planning on merging this: please drop me a note ahead of time; I think I need to tweak the cloud builder and the Docker image to accommodate this.

kit-klein:
@marcelstoer Happy to help! FWIW I did use the docker builder in my testing and it seemed to work fine.

marcelstoer:
Ok, I currently don't have time to check but I thought it might interfere with [how the boot message is created.](https://github.com/marcelstoer/docker-nodemcu-build/blob/master/build-esp32#L59) Or the Docker build script needs to set `USER_PROLOG` or something (like with the ESP8266 version).

kit-klein:
> Ok, I currently don't have time to check but I thought it might interfere with [how the boot message is created.](https://github.com/marcelstoer/docker-nodemcu-build/blob/master/build-esp32#L59) Or the Docker build script needs to set `USER_PROLOG` or something (like with the ESP8266 version).

Ahh yes - I think you're correct.  We modify `user_version.h` so the logic you linked is bypassed.

kit-klein:
@marcelstoer quick bump on this. I can attempt a fix in the docker builder if you'd like.

marcelstoer:
Any helping hands are welcome 👍

components/modules/node.c
```diff
@@ -149,6 +150,36 @@ static int node_dsleep (lua_State *L)
...
+  lua_setfield(L, -2, name);
+}
+
+static void add_string_field( lua_State* L, const char *s, const char *name) {
+  lua_pushstring(L, s);
+  lua_setfield(L, -2, name);
+}
+
+static int node_info( lua_State* L ){
+  const char* options[] = {"sw_version", "default", NULL};
```
marcelstoer, line 164: Do you think it would make sense to support the same options as with the ESP8266 but return an empty result for every option not yet implemented/applicable? Hhmm, maybe just a brain fart, don't know.

`const char* options[] = {"lfs", "hw", "sw_version", "build_config", "legacy", NULL};`

PR #3290, opened 2020-09-22
Further adaptation of lfs/_init.lua

### LINKED ISSUE #3278: dofile() won't execute files in Flash when LFS _init.lua is set
I have `_init.lua` in my `LFS` if i `dofile("abc.lua")` where `abc.lua` is in my flash not in `LFS` I get a error
``attempt to call a nil value in function 'dofile' ``
this use to work earlier......

nwf:
See https://github.com/nodemcu/nodemcu-firmware/issues/3278 .  Thanks to
@vsky279 and @bazooka07 for suggestions.

Fixes #3278.

### DIFF
--- a/lua_examples/lfs/_init.lua
+++ b/lua_examples/lfs/_init.lua
@@ -28,13 +28,13 @@
    of the code is skipped.
 ---------------------------------------------------------------------------------]]
 
-local index = node.flashindex
+local lfsindex = node.LFS and node.LFS.get or node.flashindex
 local G=_ENV or getfenv()
 local lfs_t
 if _VERSION == 'Lua 5.1' then
     lfs_t = {
     __index = function(_, name)
-        local fn_ut, ba, ma, size, modules = index(name)
+        local fn_ut, ba, ma, size, modules = lfsindex(name)
         if not ba then
           return fn_ut
         elseif name == '_time' then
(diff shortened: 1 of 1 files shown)


nwf:
Untested, but pushed for review.  If @chathurangawijetunge, @vsky279, and/or @bazooka07 would test this, that'd be great.

vsky279:
I will test it. Please patch it so that the CI build passes.
```Lua
- local mod, ext = n:match("(.*)%.(l[uc]a?)");
+ local mod = n:match("(.*)%.(l[uc]a?)");
```

vsky279:
I was testing this PR and started to wonder whether `dofile` and `loadfile` should not behave in similar way as `require` when a file is both in LFS and SPIFFS. In this case it should prefer the SPIFFS version. This should help debugging code.

The following patch to this PR does this thing:

```patch
-local lf, df = loadfile, dofile
+local lf, df, file_exists = loadfile, dofile, file.exists
 G.loadfile = function(n)
   local mod = n:match("(.*)%.l[uc]a?");
-  return index(mod) or lf(n)
+  if file_exists(n) then return lf(n) else return (index(mod) or lf(n)) end
 end
 
 G.dofile = function(n)
   local mod = n:match("(.*)%.l[uc]a?");
   local fn  = index(mod)
-  if fn then return fn() else return df(n) end
+  if file_exists(n) then return df(n) 
+  elseif fn then return fn() 
+  else return df(n) end
 end
```

It might seem a bit strange to call `lf` or `df` even if the file does not exists. This is intentional. It will raise the right error.

bazooka07:
checks if file_exists on the first line for every function to avoid calling to n:match. For example :
```lua
G.loadfile = function(n)
	if file_exists(n) then return lf(n) end

	local mod = n:match("%.lua$") or n:match("%.lc$") -- returns .lua or .lc or nil
	return mod and index(mod) or lf(n)
end
```
Same way for G.dofile.

About match() :
```lua
-- bad pattern !!
k = "test.luc"
print(k:match("(.*)%.l[ua]c?") -- True but wrong. return test

-- the right way. Good regex in Lua is missing
pattern1, pattern2 = "%.lua$", "%.lc$" 
for i,k in ipairs({"luc", "lua", "lc", "js", "css"}) do
   n = "test." .. k
    print(k .. ": "); print(n:match(pattern1) or n:match(pattern2))
end
[==[
results
luc: nil
lua: .lua
lc: .lc
js: nil
css: nil
]==]
```

vsky279:
Good point to call `n:match()` only when needed.

I don't think you need to match the extension precisely. What you care about is the "filename" so that you can call `index(mod)`. What do we gain by getting it more complex? Yes, call `loadfile("test.luc")` will call `test` when it is in LFS but is it really an issue?

bazooka07:
You forget the "$" in the pattern for match. And it 's not useful to catch the name. So drop (.*)

So, keep this expression:
```lua
local mod = n:match("%.lua$") or n:match("%.lc$") -- returns .lua or .lc or nil
```
Same things for "loadfile" and "dofile" in your commit

nwf:
I'm confused why you think the capture group isn't important?

nwf:
Any objection to this landing?

nwf:
Hearing no screams...

PR #3291, opened 2020-09-23
Add snprintf and remove old sprintf

### LINKED ISSUE #3287: Some numbers do not roundtrip through sjson correctly.
### Expected behavior

If `a` is a lua number then
```
a == sjson.decode(sjson.encode({a}))[1]
```

Specific example:
```
a = -4392806464851163000000000

rt_a = sjson.decode(sjson.encode({a}))[1]

if a ~= rt_a then
   print ("Different", a, rt_a, a-rt_a)
end
```

### Actual behavior

The code above prints:
```
Different	-4.392806e+24	-4.392806e+24	-5.764608e+17
```

### NodeMCU startup banner
```
	branch: float-test
	commit: 9d74cd5aa0960c10a9d8b04196d049023281e908
	release: 3.0-master_20200910 +4
	release DTS: 202009182147
	SSL: false
	build type: float
	LFS: 0x0 bytes total capacity
	modules: adc,bit,dht,enduser_setup,file,gpio,i2c,mqtt,net,node,ow,sjson,spi,tmr,uart,wifi
 build 2020-09-20 13:08 powered by Lua 5.3.5 on SDK 3.0.1-dev(fce080e)
```

I had to implement a couple of extra methods to be able to create uniformly distributed floating point values. I may well do a PR for those (simple) functions (which are under `#ifdef DEVELOPMENT_TOOLS`)

pjsg:
Fixes #3287.

This replaces the in-tree implementation of sprintf with snprintf (from https://github.com/weiss/c99-snprintf ). This is added as a submodule.

There are still places that call sprintf that ought to be calling snprintf.


pjsg:
Additionally, it won't format (in f format) numbers bigger than 10^20. This isn't really a problem (I think).

vsky279:
I tested this PR with tests that were discussed in recent PRs concerning `stdio.c`

```Lua
for i=-1,1 do print(("%.3f\t%+.3f\t% .3f"):format(0.0004*i, 0.0004*i, 0.0004*i)) end
for i=1,20 do d=(3.1405+i*0.001); print(("%.3f\t%7f\t%.20f"):format(d, d, d)) end
for i=1,10 do d=-(3.1405+i*0.001)*1e-5; print(("%.3e\t%7e\t%.20e"):format(d, d, d)) end
```
All prints as expected.

Also the sjson roundtrip test passes well even for numbers where it was failing with previous version
```Lua
a={}
for i=1,100 do a[i]=math.floor(node.random()*1e21)/(10^(math.floor(math.random()*20)-10)) end
b=sjson.encode(a)
c=sjson.decode(b)
for i,j in pairs(a) do if a[i]~=c[i] then print(("mismatch [%d]: %.20g ~= %.20g, diff: %.20g"):format(i, a[i], c[i], a[i]-c[i])) end end
```

`%d`, `%s` formats seem to work fine too.

I'd like to implement double rounding `#define SPRINTF_ROUNDING_MODE_DOUBLEROUNDING` as discussed in #3283 once this lands in `dev`.

app/lua53/luaconf.h
```diff
@@ -401,7 +401,7 @@
 @@ l_sprintf is equivalent to 'snprintf' or 'sprintf' in C89.
 ** (All uses in Lua have only one format item.)
 */
-#if !defined(LUA_USE_C89) && !defined(LUA_USE_ESP8266)
+#if !defined(LUA_USE_C89) || defined(LUA_USE_ESP8266)
```
nwf, line 404: Is `LUA_USE_C89` not defined when we're building for the ESP8266?

app/modules/node.c
```diff
@@ -643,6 +643,28 @@ static int node_writercr (lua_State *L) {
   lua_pushinteger(L, n);
   return 1;
 }
+
+static int node_int2float(lua_State *L) {
```
nwf, line 647: The changes to this file are not really part of this PR (at least, not by title), but I see no reason they shouldn't also land.

app/libc/snprintf.c
```diff
@@ -0,0 +1,2110 @@
...
+  va_end(arg);
+  return n;
+}
+
+int vsprintf (char *d, const char *s, va_list ap)
+{
+  return vsnprintf(d, 1024, s, ap);
+}
+
+#if TEST_SNPRINTF
```
nwf, line 1571: Can the test code move to its own file?

nwf (changes requested):
Mostly LGTM.

Any possibility we could rip `libmain.o`'s `eagle_lib.o` out of our builds if we're carrying around our own `v?snprintf`?  Annoyingly, there are some other symbols (`ets_strchr`, `ets_strcat`, `bit_popcount`???) in there that we'd need to add elsewhere.  Espressif continues to disappoint.

app/lua53/luaconf.h
pjsg: I'll fix that.

app/modules/node.c
pjsg: I added these as part of testing the new behavior. I can certainly rip them out (I didn't actually intend to commit them). I didn't add documentation for them. I think that they are only relevant for lua53 as well....

What would you like me to do?

app/libc/snprintf.c
pjsg: It could do -- but this file is essentially the exact code from the c99-snprintf repo -- so I'd rather leave it that way.

pjsg:
I'm try to figure out how to exclude eagle_lib.o from the linker but not
having any luck at the moment....

app/modules/node.c
nwf: I think it's fine to leave them in this PR; I was just noting.

app/libc/snprintf.c
nwf: If that's so, could we make this a `git submodule` instead?

nwf:
I appreciate you trying; let's not hold up this PR behind my crazed desire to divorce ourselves from Espressif. :)

nwf (approved):

nwf:
With the `git submodule` landed and CI happy, is this ready to land?

pjsg:
I think so. I have a followup PR to try and get rid of most/all of the
sprintf usages -- that will require more testing as some of the length
calculations are complex.

PR #3292, opened 2020-09-24
Fix lowprio

### LINKED ISSUE #3285: Wifi.eventmon scheduling
This is really a quick placeholder that should be replaced by a proper bugrep when I have time to do this, but in responding to  @KT819GM's post on [#3208](3208#issuecomment-695178385). I note that the `wifi.eventmon` code does not comply with SDK scheduling rules and hence its extended use could possibly cause the WiFi stack to fail.  Specifically [line 96](https://github.com/nodemcu/nodemcu-firmware/blob/release/app/modules/wifi_eventmon.c#L96) uses a `task_post_low()` to reschedule itself.  This means that once monitoring is initiated, there will always be at least one instance of event checker scheduled or running, thus starving out any lower priority tasks.

@dnc40085 is the maintainer of this code, but hasn't been active on the project since May 2018.  If he isn't available to look at this then this is going to be a TODO for one of the active maintainers, I think. if necessary, I can investigate further once I've merge my current in ESP LFS enhancement. 

### Test code
```Lua
-- add code here
```
### NodeMCU startup banner
You MUST include the firmware startup banner to describe the version you are using. We reserve the right to immediately close any bug that doesn't.

### Hardware
Describe which ESP8266/ESP32 device you use and document any special hardware setup

pjsg:
Fixes #3285 (partly).

This fixes (for the LUA5.3 version) the lowest priority task from `node.task.post` so that it runs *below* the priority of all the Espressif stuff. This means that you can hard reschedule at LOW_PRIO and still have the system function fine. What is less clear to me is whether we want MEDIUM to behave like this as well.....

I should probably fix the lua5.1 build as well..... I need to update the documentation.

By the way, the implementation/idea is taken from a comment on the espressif bbs and the implementation in the arduino port (which has a model of constantly running code).

### DIFF
--- a/app/lua53/lnodemcu.c
+++ b/app/lua53/lnodemcu.c
@@ -628,7 +628,6 @@ extern void lua_main(void);
 ** Task callback handler. Uses luaN_call to do a protected call with full traceback
 */
 static void do_task (platform_task_param_t task_fn_ref, uint8_t prio) {
-  //dbg_printf("do_task(%d, %d)\n", task_fn_ref, prio);
   lua_State* L = lua_getstate();
   if(task_fn_ref == (platform_task_param_t)~0 && prio == LUA_TASK_HIGH) {
     lua_main();                   /* Undocumented hook for lua_main() restart */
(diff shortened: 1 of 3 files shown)


pjsg:
Actually, this doesn't really do anything (except add the RAM_ATTR marker).

PR #3293, opened 2020-09-25
Fix platform_post by adding ICACHE_RAM_ATTR

### LINKED ISSUE #3285: Wifi.eventmon scheduling

pjsg:
Fixes #3285 (maybe).

The ICACHE_RAM_ATTR is missing from the platform_post function and this could cause all sorts of failures under high event loads.

### DIFF
--- a/app/platform/platform.c
+++ b/app/platform/platform.c
@@ -1151,7 +1151,3 @@ platform_task_handle_t platform_task_get_id (platform_task_callback_t t) {
   TQB.task_func[TQB.task_count++] = t;
   return TH_MONIKER + ((TQB.task_count-1) << TH_SHIFT);
 }
-
-bool ICACHE_RAM_ATTR platform_post (uint8 prio, platform_task_handle_t handle, platform_task_param_t par) {
-  return system_os_post(prio, handle | prio, par);
-}
--- a/app/platform/platform.c
+++ b/app/platform/platform.c
@@ -1152,6 +1152,6 @@ platform_task_handle_t platform_task_get_id (platform_task_callback_t t) {
   return TH_MONIKER + ((TQB.task_count-1) << TH_SHIFT);
 }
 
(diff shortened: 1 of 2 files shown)


nwf:
Seems alright by me.

TerryE (approved):
Good catch.  Yes, this is needed if `platform_post()` is being called from an ISR.  Given that it is a one line wrapper, the alternative would be to hoist it into `platform.h` as an inline function which would save the extra call overhead entirely.

nwf:
I think I would prefer the `static inline` wrapper in the header, FWIW.  It seems like the `or` will take fewer instructions than the call overhead. :)

pjsg:
You are exactly right -- I have now changed it to that.

PR #3298, opened 2020-09-28
Dhcp name

### LINKED ISSUE #3297: Allow the dhcp client hostname to be configured
### Missing feature
The default DHCP client hostname is set to 'ESP-xxxxxx' where the xxxxxx is the last six digits of the mac in hex. It would 
be nice to be able to configure this.

Ideally it would be possible to configure this from LUA, but the problem is that the DHCP starts before the LUA starts. This could be finessed by using a PCR -- this enhancement could probably be built off the startup option PR once that lands.

pjsg:
Fixes #3297.

This just initializes the dhcp client hostname way early in the boot process. This causes (for example) the name to appear in wifi management interfaces (like the UniFi interface).


nwf:
Can you just call `wifi_change_default_host_name` from `user_init` rather than its current `user_rf_pre_init`?

pjsg:
I've redone this.

PR #3303, opened 2020-09-30
Fix the initialization of the wifi default hostname. 

### LINKED ISSUE #3297: Allow the dhcp client hostname to be configured

pjsg:
Fixes #3297.

- N/A The code changes are reflected in the documentation at `docs/*`.

It turns out that the user_rf_pre_init is not called any more. This change really just moves the call into the user_init. However, I also cleaned up the wifi_set_default_host_name() function so it uses fewer #ifdefs

### DIFF
--- a/app/modules/wifi.c
+++ b/app/modules/wifi.c
@@ -1969,28 +1969,24 @@ void wifi_change_default_host_name(void)
   uint8 opmode_temp=wifi_get_opmode();
   wifi_set_opmode_current(STATION_MODE);
   char temp[33] = {0};//32 chars + NULL
-  uint8_t mac[6];
-  wifi_get_macaddr(STATION_IF, mac);
 
-#ifndef WIFI_STA_HOSTNAME
-  sprintf(temp, "NODE-%X%X%X", (mac)[3], (mac)[4], (mac)[5]);
-#elif defined(WIFI_STA_HOSTNAME) && !defined(WIFI_STA_HOSTNAME_APPEND_MAC)
-  if(wifi_sta_checkhostname(WIFI_STA_HOSTNAME, strlen(WIFI_STA_HOSTNAME))){
-    sprintf(temp, "%s", WIFI_STA_HOSTNAME);
-  }
-  else{
-    sprintf(temp, "NODE-%X%X%X", (mac)[3], (mac)[4], (mac)[5]);
-  }
-#elif defined(WIFI_STA_HOSTNAME) && defined(WIFI_STA_HOSTNAME_APPEND_MAC)
-  if(strlen(WIFI_STA_HOSTNAME) <= 26 && wifi_sta_checkhostname(WIFI_STA_HOSTNAME, strlen(WIFI_STA_HOSTNAME))){
-    sprintf(temp, "%s%X%X%X", WIFI_STA_HOSTNAME, (mac)[3], (mac)[4], (mac)[5]);
-  }
-  else{
-    sprintf(temp, "NODE-%X%X%X", (mac)[3], (mac)[4], (mac)[5]);
+  const char *hostname = "NODE";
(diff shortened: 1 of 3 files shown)


app/modules/wifi.c
```diff
@@ -1969,28 +1969,24 @@ void wifi_change_default_host_name(void)
   uint8 opmode_temp=wifi_get_opmode();
   wifi_set_opmode_current(STATION_MODE);
   char temp[33] = {0};//32 chars + NULL
+
+  const char *hostname = "NODE";
+#ifdef WIFI_STA_HOSTNAME
+  if (wifi_sta_checkhostname(WIFI_STA_HOSTNAME, strlen(WIFI_STA_HOSTNAME))) {
```
HHHartmann, line 1975: If wifi_sta_checkhostname fails and !defined(WIFI_STA_HOSTNAME_APPEND_MAC)
The name will be just NODE

HHHartmann (changes requested):

app/modules/wifi.c
pjsg: On reflection, I think that the possibly better approach is to handle an invalid hostname in the WIFI_STA_HOSTNAME by ignoring this entirely and reverting to the espressif default. It would simplify the code some more, and I really don't want to have this weird fallback handling.....

HHHartmann: Agreed. 
Apart from that it would be nice if it would be possible to show the failure of
wifi_sta_checkhostname to the user
Maybe just use a name like  "invalid Hostname" then.  Or mention the possibility of invalid Hostname should be described where it is defined

pjsg: I've added this

pjsg:
@HHHartmann I think that I have addressed all your requests....

app/modules/wifi.c
```diff
@@ -1969,28 +1969,33 @@ void wifi_change_default_host_name(void)
...
-#elif defined(WIFI_STA_HOSTNAME) && defined(WIFI_STA_HOSTNAME_APPEND_MAC)
-  if(strlen(WIFI_STA_HOSTNAME) <= 26 && wifi_sta_checkhostname(WIFI_STA_HOSTNAME, strlen(WIFI_STA_HOSTNAME))){
-    sprintf(temp, "%s%X%X%X", WIFI_STA_HOSTNAME, (mac)[3], (mac)[4], (mac)[5]);
-  }
-  else{
-    sprintf(temp, "NODE-%X%X%X", (mac)[3], (mac)[4], (mac)[5]);
-  }
+  int len = snprintf(temp, sizeof(temp), "%s-%02X%02X%02X", hostname, (mac)[3], (mac)[4], (mac)[5]);
+#else
+  int len = snprintf(temp, sizeof(temp), "%s", hostname);
```
HHHartmann, line 1985: If limiting the length here the wifi_sta_checkhostname below has only limited meaning.

Currently the given hostname is just cut off at 32 chars instead of generating an error.

pjsg: I think that the current code causes a stack overflow if the name is too long! I think that it is better to report an error rather than have the firmware do something surprising...

The checkhostname function is also used when an explicit call to set the wifi hostname is made. I guess that after doing this, you could switch modes to `off` and then back to `STA` to retrigger the DHCP with the new name. I don't know what else this hostname is used for.

marcelstoer (approved):

PR #3308, opened 2020-10-07
Don't free a ref of zero -- bad things happen...

### LINKED ISSUE #2940: Having memory leak problems when using gpio pulse
### Expected behavior
Memory should not leak on successive call to pulse.build
### Actual behavior
Memory leaks about 4k for every 100 calls in the sample below
### Test code
```
local function IntervalTimerRestart (timer)
    return function()
        print ('done:' .. node.heap())
        timer:start()
    end
end

local function IntervalTimerAction (actions, timer)
    return function()
        for id, a in next, actions do
            actions[id] = nil
        end
        table.insert(actions, { [7] = gpio.LOW, delay=500 })
        table.insert(actions, { [7] = gpio.HIGH, delay=500, loop=1, count=10})
        pulsar = gpio.pulse.build(actions)
        pulsar:start(IntervalTimerRestart (timer))
    end
end

gpio.mode(1, gpio.OUTPUT, gpio.PULLUP)
gpio.mode(7, gpio.OUTPUT, gpio.PULLUP)

gpio.write(1, gpio.LOW)
gpio.write(7, gpio.LOW)

node.egc.setmode(node.egc.ALWAYS)
local stepAction = {}
mytimer = tmr.create()
mytimer:alarm(1000, tmr.ALARM_SEMI, IntervalTimerAction(stepAction, mytimer))
```
### NodeMCU version
Building a local version of NodeMcu off master version 2.2.1

### Hardware
Describe which ESP8266 device you use and document any special hardware setup
Standard ESP 12 (Amica) dev board

pjsg:
Fixes #2940.

It turns out that a `luaL_unref(L, LUA_REGISTRYINDEX, 0)` has the rather surprising effect of throwing away the freelist of cells in the registry. This means that the registry then tends to grow. I wonder how many other places this happens!

### DIFF
--- a/app/modules/gpio_pulse.c
+++ b/app/modules/gpio_pulse.c
@@ -416,7 +416,9 @@ static int gpio_pulse_start(lua_State *L) {
     return luaL_error( L, "missing callback" );
   }
 
-  luaL_unref(L, LUA_REGISTRYINDEX, pulser->cb_ref);
+  if (pulser->cb_ref) {
+    luaL_unref(L, LUA_REGISTRYINDEX, pulser->cb_ref);
+  }
   pulser->cb_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 
   active_pulser = pulser;


nwf:
I'm happy to merge this small fix, but it would be nice to know that the bug doesn't happen elsewhere!

pjsg:
I was wondering whether to put a panic or log message in the lua free
method...

pjsg:
See this conversation about this issue in 2013: http://lua-users.org/lists/lua-l/2013-11/msg00175.html

TerryE (commented):
The "bad thing" is that `pulser->cb_ref` has ended up as `NULL`.  It is should either be an active reference or `LUA_NOREF` and `luaL_unref()` handles these correctly  without the need for a NULL check.  Best practice is to initialise any CB ref to `LUA_NOREF` .  We've added the `luaL_unref2()` to help simplify this coding.  My preference is to simplify rather than mask bugs elsewhere. 

**PS**.  This has fixed the bug, so moving to best practice is just another TODO tidy up.

PR #3324, opened 2020-11-08
Basic hardware test examples

nwf:
Now that `NTest` has landed (hooray) it's time to start fleshing out our test environment.  Towards that end, here are example tests of `adc` and `gpio` which rely on some modest external hardware, which is also described in this PR.  This is a subset of the work done in #2983, so it may feel familiar.

### DIFF
--- a/tests/NodeMCU_Test_Environment.rst
+++ b/tests/NodeMCU_Test_Environment.rst
@@ -1,0 +1,157 @@
+###########################
+NodeMCU Testing Environment
+###########################
+
+Herein we define the environment our testing framework expects to see when it
+runs.  It is composed of two ESP8266 devices, each capable of holding an entire
+NodeMCU firmware, LFS image, and SPIFFS file system, as well as additional
+peripheral hardware.  It is designed to fit comfortably on a breadboard and so
+should be easily replicated and integrated into any firmware validation
+testing.
+
+The test harness runs from a dedicated host computer, which is expected to have
+reset- and programming-capable UART links to both ESP8266 devices, as found on
+almost all ESP8266 boards with USB to UART adapters, but the host does not
+necessarily need to use USB to connect, so long as TXD, RXD, DTR, and RTS are
+wired across.
+
+Peripherals
+###########
+
+I2C Bus
+=======
+
+There is an I2C bus hanging off DUT 0.  Attached hardware is used both as tests
+of modules directly and also to facilitate testing other modules (e.g., gpio).
+
+MCP23017: I/O Expander
+----------------------
+
+At address 0x20.  An 16-bit tristate GPIO expander, this chip is used to test
+I2C, GPIO, and ADC functionality.  This chip's interconnections are as follows:
+
++---------+-------------------------------------------------------------------+
+| /RESET  | DUT0 reset.  This resets the chip whenever the host computer      |
+|         | resets DUT 0 over its serial link (using DTR/RTS).                |
++---------+-------------------------------------------------------------------+
+| B 0     | 4K7 resistor to DUT 0 ADC.                                        |
++---------+-------------------------------------------------------------------+
+| B 1     | 2K2 resistor to DUT 0 ADC.                                        |
++---------+-------------------------------------------------------------------+
+| B 5     | DUT1 GPIO16/WAKE via 4K7 resitor                                  |
++---------+-------------------------------------------------------------------+
+| B 6     | DUT0 GPIO13 via 4K7 resistor and DUT1 GPIO15 via 4K7 resistor     |
++---------+-------------------------------------------------------------------+
+| B 7     | DUT0 GPIO15 via 4K7 resistor and DUT1 GPIO13 via 4K7 resistor     |
++---------+-------------------------------------------------------------------+
+
+Notes:
+
+* DUT 0's ADC pin is connected via a 2K2 reistor to this chip's port B, pin 1
+  and via a 4K7 resistor to port B, pin 0.  This gives us the ability to
+  produce approximately 0 (both pins low), 1.1 (pin 0 high, pin 1 low), 2.2
+  (pin 1 high, pin 0 low), and 3.3V (both pins high) on the ADC pin.
+
+* Port B pins 6 and 7 sit on the UART cross-wiring between DUT 0 and DUT 1.
+  The 23017 will be tristated for inter-DUT UART tests, but these 
+
+* Port B pins 2, 3, and 4, as well as all of port A, remain available for
+  expansion.
+
+* The interrupt pins are not yet routed, but could be.  We reserve DUT 0
+  GPIO 2 for this purpose with the understanding that the 23017's
+  interrupt functionality will be disabled (INTA, INTB set to open-drain,
+  GPINTEN set to 0) when not explicitly under test.
+
+ESP8266 Device 0 Connections
+############################
+
++---------+---------------------------------------------------------+
+| ESP     |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 0  | Used to enter programming mode; otherwise unused in     |
+|         | test environment.                                       |
++---------+---------------------------------------------------------+
+| GPIO 1  | Primary UART transmit; reserved for host communication  |
++---------+---------------------------------------------------------+
+| GPIO 2  | [reserved for 1-Wire]                                   |
+|         | [+ reserved for 23017 INT[AB] connections]              |
++---------+---------------------------------------------------------+
+| GPIO 3  | Primary UART recieve; reserved for host communication   |
++---------+---------------------------------------------------------+
+| GPIO 4  | I2C SDA                                                 |
++---------+---------------------------------------------------------+
+| GPIO 5  | I2C SCL                                                 |
++---------+---------------------------------------------------------+
+| GPIO 6  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 7  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 8  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 9  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 10 | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 11 | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 12 |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 13 | Secondary UART RX; DUT 1 GPIO 15, I/O expander B 6      |
++---------+---------------------------------------------------------+
+| GPIO 14 |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 15 | Secondary UART TX; DUT 1 GPIO 13, I/O expander B 7      |
++---------+---------------------------------------------------------+
+| GPIO 16 |                                                         |
++---------+---------------------------------------------------------+
+| ADC 0   | Resistor divider with I/O expander                      |
++---------+---------------------------------------------------------+
+
+ESP8266 Device 1 Connections
+############################
+
++---------+---------------------------------------------------------+
+| ESP     |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 0  | Used to enter programming mode; otherwise unused in     |
+|         | test environment.                                       |
++---------+---------------------------------------------------------+
+| GPIO 1  | Primary UART transmit; reserved for host communication  |
++---------+---------------------------------------------------------+
+| GPIO 2  | [Reserved for WS2812]                                   |
++---------+---------------------------------------------------------+
+| GPIO 3  | Primary UART recieve; reserved for host communication   |
++---------+---------------------------------------------------------+
+| GPIO 4  |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 5  |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 6  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 7  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 8  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 9  | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 10 | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 11 | [Reserved for on-chip flash]                            |
++---------+---------------------------------------------------------+
+| GPIO 12 | HSPI MISO                                               |
++---------+---------------------------------------------------------+
+| GPIO 13 | Secondary UART RX; DUT 0 GPIO 15, I/O exp B 7 via 4K7   |
+|         | Also used as HSPI MOSI for SPI tests                    |
++---------+---------------------------------------------------------+
+| GPIO 14 | HSPI CLK                                                |
++---------+---------------------------------------------------------+
+| GPIO 15 | Secondary UART TX; DUT 0 GPIO 13, I/O exp B 6 via 4K7   |
+|         | Also used as HSPI /CS for SPI tests                     |
++---------+---------------------------------------------------------+
+| GPIO 16 | I/O expander B 5 via 4K7 resistor, for deep-sleep tests |
++---------+---------------------------------------------------------+
+| ADC 0   |                                                         |
++---------+---------------------------------------------------------+
+
+
--- a/tests/NTest/NTest.lua
+++ b/tests/NTest/NTest.lua
@@ -202,8 +202,8 @@ local function NTest(testrunname, failoldinterface)
 
       env.eq = deepeq
       env.spy = spy
-      env.ok = function (cond, msg1, msg2) wrap(assertok, false, cond, msg1, msg2) end
-      env.nok = function(cond, msg1, msg2) wrap(assertok, true,  cond, msg1, msg2) end
+      env.ok = function (cond, msg) wrap(assertok, false, cond, msg) end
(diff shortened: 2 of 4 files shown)


tests/NTest_gpio.lua
```diff
@@ -0,0 +1,82 @@
+-- Walk the GPIO subsystem through its paces, using the attached I2C GPIO chip
+--
+--   Node GPIO 13 (index 7) is connected to I2C expander channel B6; node OUT
+--   Node GPIO 15 (index 8) is connected to I2C expander channel B7; node IN
+
+local N = require('NTest')("gpio")
+
+N.test('setup', function()
```
HHHartmann, line 8: That looks good as an init function. I had thought about adding `init_run`, `before_each_test` and the one afterwards to NTest but this makes the `init_run` and its counterpart obsolete at least.

tests/NTest_adc.lua
```diff
@@ -0,0 +1,57 @@
...
+  N.test(fs:format(i), function()
+    setv(i)
+    print("set v", i)
+    tmr.delay(10)
+    ok(checkadc(vmin[i+1], vmax[i+1]))
+  end)
+end
+
+-- test all four stops on the way up, and the three to come back down
+for i=0,3    do mktest("%d up", i) end
```
HHHartmann, line 56: Nice way to have data tests.

tests/NTest_adc.lua
```diff
@@ -0,0 +1,57 @@
...
+local vmin = {  0, 300, 700, 1000 }
+local vmax = { 24, 400, 800, 1024 }
+
+-- Set the DAC, wait a brief while for physics, and then read the ADC
+local function mktest(fs, i)
+  N.test(fs:format(i), function()
+    setv(i)
+    print("set v", i)
+    tmr.delay(10)
+    ok(checkadc(vmin[i+1], vmax[i+1]))
```
HHHartmann, line 51: You could add a msg here to show what is failing/succeeding.
You could even move the ok to checkadc, using the string from the print statement there.

``

HHHartmann, line 51:
```suggestion
    checkadc(vmin[i+1], vmax[i+1])
```

tests/NTest_adc.lua
```diff
@@ -0,0 +1,57 @@
...
+  i2c.write(0, 0x15, v)
+  i2c.stop(0)
+end
+
+-- read out the ADC and compare to given range
+local function checkadc(min, max)
+  local v = adc.read(0)
+  print("read adc", v, min, max)
+  return v >= min and v <= max
+end
```
HHHartmann, line 38:
```suggestion
local function checkadc(min, max)
  local v = adc.read(0)
  ok(v >= min and v <= max, "read adc"..v..min..max)   -- this is pseudo code
end
```

HHHartmann (approved):
That was a quick reaction :-)
As mentioned before I like the setup of the Hardware. I would also like to use DUT0 to DUT1 tests for TCP and UDP testing. So they would have to be started at the same time.

Just some minor annotations and suggestions.

Maybe it would be good to test GPIO by directly connecting two pins. That way in case of failure there would not be so many parts involved and troubleshooting would be eased.

Maybe it would make sense to have a HW manager which knows how to set up the HW for certain scenarios.
It knows that it has to set the I/O Expander to open drain if the UART is to be used e.g.
It knows how to connect the GPIOs for GPIO testing and maybe even just gets told to set GPIO  8 to HIGH for testing interrupts.

But that is not part of this PR and might also be overdone.
That way the same test could be run on different hardware where the pins might be connected directly instead of using the expander.

HHHartmann:
BTW the tests are red (hehe last thing I added in the NTest PR)

nwf:
> That was a quick reaction :-)
> As mentioned before I like the setup of the Hardware. I would also like to use DUT0 to DUT1 tests for TCP and UDP testing. So they would have to be started at the same time.

The way I have it wired up right now, DUT1 ends up lingering in RESET until DUT0 releases it, I think.  But yes, certainly the plan is to drive both DUTs in tandem and have them flex their networking stacks.  Ideally there'd be a series of tests in which both of the DUTs 

> Just some minor annotations and suggestions.
> 
> Maybe it would be good to test GPIO by directly connecting two pins. That way in case of failure there would not be so many parts involved and troubleshooting would be eased.

Not opposed, just somewhat lacking in free pins.  Would you be equally happy with a dual-DUT GPIO test?  The two UARTs are indeed directly connected and we could do without futzing about with the I2C I/O expander in such a test.

> Maybe it would make sense to have a HW manager which knows how to set up the HW for certain scenarios.
> It knows that it has to set the I/O Expander to open drain if the UART is to be used e.g.
> It knows how to connect the GPIOs for GPIO testing and maybe even just gets told to set GPIO 8 to HIGH for testing interrupts.
>
> But that is not part of this PR and might also be overdone.

I like the idea of a "nodemcu test environment library" that knows how to do parts of the bringup; that's a solid plan.

> That way the same test could be run on different hardware where the pins might be connected directly instead of using the expander.

Expanding further to a kind of general API for test environments, plural, seems like a worthwhile goal but I'm not sure I quite see how to generalize from the singular test environment I've proposed to more than one.  But I certainly wouldn't object to seeing such a thing!

HHHartmann:
> The way I have it wired up right now, DUT1 ends up lingering in RESET until DUT0 releases it, I think. But yes, certainly the plan is to drive both DUTs in tandem and have them flex their networking stacks. Ideally there'd be a series of tests in which both of the DUTs

... ?

> 
> Not opposed, just somewhat lacking in free pins. Would you be equally happy with a dual-DUT GPIO test? The two UARTs are indeed directly connected and we could do without futzing about with the I2C I/O expander in such a test.

That would still require both DUTs to work together. Maybe DUT1 could just implement a short cirquit between the two pins. Just like I plan to just have something similar for the TCP stack. So DUT1 would be as dumb as possible and the entire test logic would run on DUT0.

> 
> > Maybe it would make sense to have a HW manager which knows how to set up the HW for certain scenarios.
> > It knows that it has to set the I/O Expander to open drain if the UART is to be used e.g.
> > It knows how to connect the GPIOs for GPIO testing and maybe even just gets told to set GPIO 8 to HIGH for testing interrupts.
> > But that is not part of this PR and might also be overdone.
> 
> I like the idea of a "nodemcu test environment library" that knows how to do parts of the bringup; that's a solid plan.
> 
> > That way the same test could be run on different hardware where the pins might be connected directly instead of using the expander.
> 
> Expanding further to a kind of general API for test environments, plural, seems like a worthwhile goal but I'm not sure I quite see how to generalize from the singular test environment I've proposed to more than one. But I certainly wouldn't object to seeing such a thing!

I would rather use it for developing new tests and not having the complete test HW available i guess.

nwf:
> > The way I have it wired up right now, DUT1 ends up lingering in RESET until DUT0 releases it, I think. But yes, certainly the plan is to drive both DUTs in tandem and have them flex their networking stacks. Ideally there'd be a series of tests in which both of the DUTs
> 
> ... ?

I, uh, apparently hit send prematurely, sorry.  "... both of the DUTs speak to each other over the network in various ways, including both connected to the same external AP and one STA connected to the other's AP or AP+STA modes." 

> > Not opposed, just somewhat lacking in free pins. Would you be equally happy with a dual-DUT GPIO test? The two UARTs are indeed directly connected and we could do without futzing about with the I2C I/O expander in such a test.
> 
> That would still require both DUTs to work together. Maybe DUT1 could just implement a short cirquit between the two pins. Just like I plan to just have something similar for the TCP stack. So DUT1 would be as dumb as possible and the entire test logic would run on DUT0.

Fair.

> > Expanding further to a kind of general API for test environments, plural, seems like a worthwhile goal but I'm not sure I quite see how to generalize from the singular test environment I've proposed to more than one. But I certainly wouldn't object to seeing such a thing!
> 
> I would rather use it for developing new tests and not having the complete test HW available i guess.

Well, as I said, I'm quite open to suggestion.  I think the simplest thing might just be to have a test environment description module which just reports what the peripheral environment / pin wiring / ... looks like, coupled with multiple `NTest_gpio-*.lua` files suitable for different configurations.  Tests finding themselves in a bad place can just skip out.

tests/NodeMCU_Test_Environment.rst
```diff
@@ -0,0 +1,157 @@
...
++---------+---------------------------------------------------------+
+| ESP     |                                                         |
++---------+---------------------------------------------------------+
+| GPIO 0  | Used to enter programming mode; otherwise unused in     |
+|         | test environment.                                       |
++---------+---------------------------------------------------------+
+| GPIO 1  | Primary UART transmit; reserved for host communication  |
++---------+---------------------------------------------------------+
+| GPIO 2  | [reserved for 1-Wire]                                   |
+|         | [+ reserved for 23017 INT[AB] connections]              |
```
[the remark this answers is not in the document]
nwf: The interrupt pins can be configured into open-drain rather than push-pull, so that should be fine, though the tests will have to remember to do it.

tests/NTest_gpio.lua
nwf: I wonder if a `N.preflight` variant on `N.test` might still be useful: sometimes if a test fails, there's no point in running any of the rest of them, and so we should just stop now.

HHHartmann: I was planning to have a collection of features which could be required in the test declaration (or the testrun) and a (configuration file)/(detection process for installed modules) which is checked before running a test.

But  preflight allows for dynamic checks.

tests/NodeMCU_Test_Environment.rst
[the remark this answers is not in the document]
HHHartmann: So that can be done by the test environment library/HW manager

HHHartmann:
How about having a WIKI page which describes our road map on this. That way it doesn't get lost in various Issues.

pjsg:
From looking at this, do you need raw ESP8266 modules, or can you build this test environment with (say) a pair of D1 Minis + supporting parts? I could fab a small number of boards to make building the test rig fairly easy.....

nwf:
Certainly the intention was to be compatible with the DIP carrier boards (e.g., D1 Mini, Feather, &c, but probably not the 8-pin modules, which also just don't have enough I/O for most peripheral testing) to make it easy for anyone to stand up a test environment paralleling the one on my desk so that the bus factor was more than 1.  For the same reason I picked an I2C I/O expander available in DIP packaging and would, ideally, continue to rely on breadboard-friendly peripherals for testing.

Fabbing some boards eventually would be great, but I think that we have not yet actually converged on what the full test environment looks like (though I hope I've made a good start between what's in tree and what's in my `dev-active` branch).  If you wanted to make a kicad schematic of the environment thus far, tho', that'd be pretty great and makes it easier to eventually fab something; I keep meaning to but not finding the time.

pjsg:
If you create yourself an account at  https://easyeda.com and then tell me your username, I can add you to the project. 

The design holds two D1 Minis, has the MCP23017, a WS2812, a position for a DS18B20 and a position for the cheap 128x64 OLED displays.

The cost of these boards is ~$30 for 10 -- this is populated with the MCP23017, the WS2812 and all the resistors. It would need female headers soldering in to hold the D1 minis, and a header for the OLED display if required. If we switched the 23017 to be a PCA9554 (8 port) or PCA9555 (16 port), then that would save $5 -- bringing the cost down to around $25/10 boards (plus shipping). 

Adding other connections and the like is easy (and remarkably cheap if the component is common).

nwf:
Hm!  That *is* surprisingly cheap!

The 9555 is a little less flexible than the 23017, but I don't think we're likely to miss the missing features.  OTOH the 9555 apparently *exists* in DIP form, but isn't made in it any more?  Perhaps there's yet a third option, or perhaps this lends increased weight to @HHHartmann's suggestion that we need a test fixture abstraction library against which the tests are written.

pjsg:
I'm going to leave it as the 23017 as he pricing on the other parts seems confusing to me....

I added:

* Two WS2812s attached to DUT1
* I2C OLED display header for DUT0 and DUT1
* DS18B20 pads for DUT0
* Brought out all the D1 Mini pins to a prototyping area
* Brought out all the GPA & INT pins from the 23017 to a prototyping area.

Rough idea:

image

nwf:
Oh nice.

Some random thoughts:

* Ah, a schematic.  Amazing.  Even if we don't spin the boards any time soon, can we land the schematic in the tree?  While less universally accessible than ASCII, it's probably also less ambiguous than the prose in the .rst doc.

* I like the prototyping area. :+1:  If it isn't (much) more expensive, could we have them drill out a solderable breadboard pattern (8-16 rows, maybe only 3 wide on a side rather than 5, two busses down the middle, say)?

* I'd been planning to have two 18B20s hanging off of DUT0 (and in fact my breadboard mockup does), so that we could test 1W bus arbitration and discovery, but it might be more fun to have two different 1W devices.  Suggestions?  (I have a Java iButton from back in the day floating around, but they're not really available these days, sadly!)

* Any possibility of automating testing the WS2812?  I haven't a clue (a color sensor IC?), but it would be nice to do.

* A naive question: is there a *minimum* distance the two 802.11 antennas have to be from each other for them to be able to talk to each other?

HHHartmann:
Wow that looks good! 

Not sure if we need that much prototyping area. Maybe the ability to solder some female headers to be able to connect to a breadboard right on both sides of the wemos and the same for the 23017

Also thought about the antenna issue and automated WS2812 tests. Isn't there a 1W color sensor?

I was wondering if we could have one DUT run a stable version of the firmware, only adapting the Lua code if needed and only have to flash the other DUT for testing. That way we could probably only connect that one to the computer, powering the whole board from that connection and only occupying one USB port.

nwf:
Headers in lieu of a prototyping area is probably also fine.

I'm not sure about testing via one DUT.  It might be possible, but I think at least some of our tests are going to want to update the firmwares on both (e.g., inter-DUT networking tests).  I don't believe there's any harm in having two USB cables to hook up, and it should be relatively straightforward to ensure a lack of ground or power loops on the board -- most peripherals connect to just one DUT and so can draw their power and ground links from that.  The three cross-DUT links right now all have current limiting resistors and besides the two should be at the same ground anyway given their USB connectors.

That said, if there's really a strong reason for only having one USB cable, we could bring DUT1's RESET and GPIO0 over to the 23017 (just as we already bring its TXD, RXD, and WAKE pins) so that DUT0 and the host could conspire to reprogram DUT1 (slowly, by using DUT0 as a bounce buffer and having it juggle its UART pins).  That does mean that DUT1 chews up five of the 23017's channels all told, but that might be OK; we still have Paris^WPort A completely free.

pjsg:
Currently I'm not bringing over the RX/TX pins from DUT1 to DUT0 -- these can't go via the 23017 as it can't emulate a UART.... Also, I guess that we would have to use the software UART to drive them....

The prototyping area just brings out each pin of DUT0 and DUT1. each pin is brought out to three holes on .1" spacing. You can solder a 16 pin 0.1" header down this strip if you want.

The 5V rails from the two boards are connected via a 1Ω resistor to prevent large loop currents. The 3.3V rails are entirely disconnected, but should be close as there are identical regulators providing that voltage.

I suspect that we could add a TSC34725 board mounted upside down over one of the ws2812s. These boards are cheap.... It would allow verification of the ws2812 driver. 

I'm also going to add another 3V3, GND, SCL, SDA header (or two) so that random other devices can be attached. E.g. BMP280 etc. It appears that most of the cheap chinese modules have the pins in that order.....

nwf:
Oh, fooi; I got wires crossed (hah) and was thinking that we already bridge DUT UARTs... and we sort of do, but alternate UART pins, which probably aren't usable by the bootstrap loader (unless someone knows some magic strapping configuration).  Maybe we could swap the connections to DUT1 GPIO15/GPIO13 to DUT1 TX/RX, but we'd probably need some way to disable the USB UARTs on DUT1.

:+1: on the 1Ω resistor between the 5V rails and the TSC34725 and the extra I2C breakouts.

pjsg:
It isn't quite clear how to cross-wire the UARTs etc. This is the current version of the board. It is somewhat larger -- about 100mm square. It includes spaces for a bunch of I2C devices which can be added or not. There is support for measuring the color of the WS2812 with a light sensor. 

image

pjsg:
I note that the DUT1 has no connection on its RST pin. This is probably required if we want DUT0 to control it.

Also, I suspect that we want to cross wire the alt UART pins on DUT0 to the real UART pins on DUT1. However, You need to pull hard on the DUT1 RX line to override the drive of the USB interface. This implies no resistors in the path. I'm not sure why the existing design has resistors and the 23017 in the path.

I've added connections from the 23017 to DUT1_RST and DUT1_D3 (which is GPIO0).

nwf:
I think we should abandon the idea of DUT0 controlling DUT1 and just use two USB ports on the host.  IMHO it's by far simpler that way.  The existing two cross connections should be more than sufficient for us to test both `gpio` and `uart` functionality.  I'm not necessarily opposed to adding more, but I don't see the point.

The resistors are in the cross-coupling path and in fact should be between all connected pairs of push/pull drivers to prevent buggy software from frying hardware: pushing a strong 5V into a strong pull to ground is going to cause something to get warm and maybe smoke.

pjsg:
I added the cross linking to the RX and TX on DUT1 -- but it is controlled by a 23017 pin. Normally it is disabled. It is designed so that programming errors will not cause smoke....

Total cost is now $35 for 10 boards + maybe $20 shipping. 

Any more capability that anybody wants? There is a significant cost increase if the board gets any larger -- it is currently 100mm x 100mm. 

image

nwf:
I think I'd rather we all had breadboards for a while before baking the test wiring into PCBs.  Obviously others are welcome to other opinions, but I think that hardware has outstripped software at this point and it would be good to actually have tests for more of the things you've mounted on that (very lovely) board.

tests/NodeMCU_Test_Environment.rst
```diff
@@ -0,0 +1,157 @@
+###########################
```
marcelstoer, line 1: Only just noticed...this is the only `.rst` doc we have. For consistency reasons we should probably convert this to markdown?

nwf: I loathe Markdown, but if someone commits the conversion I won't be too upset.

pjsg:
Time to break out the soldering iron and see how much works...

Philip

PR #3332, opened 2020-11-15
File LFS Lua module initial commit

### LINKED ISSUE #3329: RFC Arbitrary files in LFS
### Missing feature
Support for arbitrary files stored in LFS

### Justification
For current application I'd like to have SPIFFS file system free for logs or configuration files. While Lua code is stored in LFS, `init.lua` file can be there too, there is no easily available solution for an arbitrary files to be stored in LFS now. Such file can be something like `index.html`, `favicon.ico`, etc.

I had a discussion with @TerryE Terry who throw me some ideas. I'd like to hear your comments to the proposed solution.

In the linked __[gist](https://gist.github.com/vsky279/9489b1a68653d3568db4de32da9e685b)__ there are several files:
* `file_lfs.lua` - Lua module built above standard `file` module providing transparent access to files stored in LFS and SPIFFS.
* `make_resource.lua` - Lua script to be used on PC to generate `resource.lua` file from files to be embedded in LFS.
* `resource.lua` - example of `favicon.ico` file to be included in LFS

### Usage
The module is intended to replace standard `file` module. If file is present in SPIFFS then this version is preferred. If it is not present then read only access (of course) to the file in LFS is provided.
I see big advantage of this solution that no change to the existing code needs to be done. Performance of `file` function seems to be affected though this does not seem to be critical for most applications (no benchmarking has been done).

The following functions for LFS files are supported: 
* `file.exists()`
* `file.open`
* `file.read()`, `file.obj:read()`
* `file.readline()`, `file.obj:readline()`
* `file.seek()`, `file.obj:seek()`

Other functions are just passed-through to the standard `file` module.

When file is embedded in LFS, no standard userdata file structure is created. Instead a Lua object is created.

### Example
```Lua
file = require("file_lfs")

local idx = file.open("index.html") -- file not in SPIFFS but in resource.lua in LFS
print(idx:readline())
idx:seek("set", 0)
print(file.readline(idx))

idx = file.open("favicon.ico", "r") -- file not in SPIFFS but in resource.lua in LFS
print(file.read(512))
file.seek(idx, "cur", 100)
print(idx:read(16))
idx:close()

local i = file.open("init.lua") -- files in SPIFFS are accessible using same routines
print(i:readline())
```

vsky279:
Fixes #3329.

Support for arbitrary files stored in LFS

### Usage
The module is intended to overlay standard `file` module. If file is present in SPIFFS then this version is preferred. If it is not present then read only access (of course) to the file in LFS is provided.

The following functions for LFS files are supported: 
* `file.exists()`
* `file.open()`
* `file.getcontents()`
* `file.read()`, `file.obj:read()`
* `file.readline()`, `file.obj:readline()`
* `file.seek()`, `file.obj:seek()`

Other functions are just passed-through to the standard `file` module.

When file is embedded in LFS, no standard userdata file structure is created. Instead a Lua object is created.

### Example
```Lua
file = require("file_lfs")

local idx = file.open("index.html") -- file not in SPIFFS but in resource.lua in LFS
print(idx:readline())
idx:seek("set", 0)
print(file.readline(idx))

idx = file.open("favicon.ico", "r") -- file not in SPIFFS but in resource.lua in LFS
print(file.read(512))
file.seek(idx, "cur", 100)
print(idx:read(16))
idx:close()

local i = file.open("init.lua") -- files in SPIFFS are accessible using same routines
print(i:readline())
```

Also test file for the testing framework (#2983) is provided.


pjsg:
I like this -- I built something similar (though not so general purpose) for an application that I am working on. This is definitely a better approach!

HHHartmann:
@vsky279 could you please not force push? We can squash the commits when merging and the reviewer will have to read _everything_ over again and again.

vsky279:
Sure, no problem. So I will pile up commits that we squash once ready to be merged.

nwf:
I'm delighted to see `NTest` catching on! :D

lua_modules/file_lfs/file_lfs.lua
```diff
@@ -0,0 +1,102 @@
...
+
+file_lfs.exists = function (filename)
+  return (node_LFS_resource(filename) ~= nil) or file_exists(filename)
+end
+
+file_lfs.open = function (filename, mode)
+  mode = mode or "r"
+  if file_exists(filename) then
+    return file_open(filename, mode)
+  elseif node_LFS_resource(filename) and mode:find("r") then
```
HHHartmann, line 70: This should be exactly r and not allow r+

HHHartmann (changes requested):
Please also review my comments in #3329
I feel that a file enhancement should work as usual and throw errors where normal operation can not be guaranteed.
Given the following code
```lua
if file.exists("myfile") then file.rename("myfile", "myotherfile") end
print(file.getcontents("myotherfile"))
```
should work, regardless of myfile being in SPIFFS or LFS.
An alternative would be to throw a meaningful error that the operation is not possible.
Atm copy from a non existing file throws an error (I hope) but with puzzeling message.

Maybe some code to smoothly access the resource like a file might be more useful without blending with file.

lua_modules/file_lfs/make_resource.lua
```diff
@@ -0,0 +1,57 @@
...
+  local inp = string.match(a, ".*[/](.*)")
+  local content = readfile(a)
+  print(string.format("# processing %s", inp))
+
+  if content then
+    res = io.open(OUT, "a")
+    res:write(('if arg == "%s" then return %q  end\n\n'):format(inp, content))
+    res:close(file)
+  end
+end
```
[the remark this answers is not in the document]
vsky279: Please see below.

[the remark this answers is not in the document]
vsky279: This is an issue. With the structure of `resource.lua` file as it is defined now - i.e. returning file content depending on the name passed as an argument - we cannot simply list included files.
`resource.lua` can return list of included files when called w/o parameter. But this list needed to be populated manually in the file. Do you have other idea?

vsky279:
Thanks Gregor for all these comments!

>Please also review my comments in #3329

I'm sorry I missed these ones.

> open modes

I suggest the following LFS file behavior:
* "r": standard mode
* "w": opens file of the same name on SPIFFS
* "a", "r+", "w+", "a+": copy LFS file content to SPIFFS and opens with given open mode

> file.getcontents()

We already have `file.getcontents()` - this needed to be implemented for the test suite.

>file.rename()

Renaming non existing file returns `false` rather than raising error. So I guess an attempt to rename LFS file should return `false` too.
It's implemented now. But the implementation is a bit useless as calling directly `file.rename` with LFS filename (as if `file_lfs.rename` was not implemented) would return `false` anyway.

> file.stat()

 :thumbsup:


```Lua
file.open("index.html")
print(file.readline())
```

Oups. I didn't know this approach is still possible. Implemented.

lua_modules/file_lfs/file_lfs.lua
```diff
@@ -84,13 +103,36 @@ file_lfs.getcontents = function(filename)
...
+  end
+end
+
+file_lfs.stat = function(filename)
+  if node_LFS_resource(filename) ~= nil and not file_exists(filename) then
+    return {
+      name = filename,
+      size = #node_LFS_resource(filename),
+      time = {day = 1, hour = 0, min = 0, year = 1970, sec = 0, mon = 1},
+      is_hidden = false, is_rdonly = false, is_dir = false, is_arch = false, is_sys = false,
```
vsky279, line 121: Next commit will have `is_rdonly = true`

docs/lua-modules/file_lfs.md
```diff
@@ -0,0 +1,148 @@
...
+### Syntax
+```bash
+ ./make_resource.lua [-o outputfile] file1 [file2] 
+ ```
+
+### Example
+Create `resource.lua` file with all files in the `resource` directory
+```bash
+./make_resource resource/*
+```
```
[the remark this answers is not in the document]
vsky279: I think we have sufficient description in the section "## `resource.lua`" file.
I'll update a bit the description by adding
```patch
- A structure of the `resource.lua` file is simple
+ A structure of the `resource.lua` file is simple. It returns a string, i.e. file content, depending on filename parameter passed to it.
```

lua_modules/file_lfs/make_resource.lua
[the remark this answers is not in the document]
HHHartmann: no other idea. Just add a return {"filename1", "filename2" ... } if called without param. The strings are there anyways, so the LFS memory impact should be minimal.

lua_modules/file_lfs/make_resource.lua
```diff
@@ -0,0 +1,65 @@
...
+
+for _, a in pairs(larg) do
+  local inp = string.match(a, ".*[/](.*)")
+  if not inp then inp = a end
+  local content = readfile(a)
+  print(string.format("# processing %s", inp))
+
+  if content then
+    res = io.open(OUT, "a")
+    res:write(('table_insert(filelist, "%s")\n'):format(inp, content))
```
HHHartmann, line 57: This implementation will work but create the list in memory each time a resource is requested.
It would be better to build up the { "name1","name2"} string in this script and then generate code like
`if arg == nil then return { "name1","name2"} end`

lua_modules/file_lfs/file_lfs.lua
```diff
@@ -0,0 +1,155 @@
...
+      time = {day = 1, hour = 0, min = 0, year = 1970, sec = 0, mon = 1},
+      is_hidden = false, is_rdonly = true, is_dir = false, is_arch = false, is_sys = false,
+    }
+  else
+    return file_stat(filename)
+  end
+end
+
+file_lfs.list = function (pattern)
+  local filelist = file_list(pattern)
```
HHHartmann, line 129: Good thing implementing this as well.
But if I am not mistaken it hides the underlying file.list(). But maybe it is acceptable in this case.

lua_modules/file_lfs/make_resource.lua
vsky279: You're right. I was thinking about the possibility of copying blocks of the resource file into another one. Blocks as
```Lua
table_insert(filelist, "index.html")
if arg == "index.html" then return "..." end
```
But your the costs outweight the result. Adjusted as you propose.

lua_modules/file_lfs/file_lfs.lua
vsky279: You're right the underlzing `file.list()` is hidden.
I'm thinking about second parameter `file.list([pattern, [SPIFFs only]])` to make it accessbile. By default all files are listed.

vsky279:
I am not sure what's the issue with doc_check:
```
Run ./tools/check_docs_module_linkage.sh
7a8
> lua-modules/file_lfs.md
Error: Process completed with exit code 1.
```

Local doc_check does not raise any issue.

tests/NTest_file_lfs.lua
```diff
@@ -0,0 +1,103 @@
...
+file = require("file_lfs")
+
+-- check standard SPIFFS file functions
+require("NTest_file")
+
+local N = require('NTest')("file_lfs")
+
+N.test('resource.lua in LFS', function()
+  ok(node.LFS.resource~=nil, "resource.lua embedded in LFS")
+end)
```
HHHartmann, line 16: Please change the code to the following:

```Lua
-- luacheck: globals file
-- luacheck: new read globals node.LFS.resource
file = require("file_lfs")

local Nt = ...
Nt = (Nt or require "NTest")

-- check standard SPIFFS file functions
loadfile("NTest_file.lua")(Nt)

local N = Nt("file_lfs")

N.test('resource.lua in LFS', function()
  ok(node.LFS.resource~=nil, "resource.lua embedded in LFS")
end)
```

This is for two reasons:
1. we changed the tests to allow passing a preconfigured NTest instnce which must be used.
2. Please don't require tests, just dofile them.

I stumbled across this because the original version didn't run for me.

tests/NTest_file_lfs.lua
```diff
@@ -0,0 +1,103 @@
...
+  content = f:read(10)
+  ok(eq(testcontent:sub(120,129), content),"cur 10")
+
+  f:seek("cur", -50)
+  content = f:read(10)
+  ok(eq(testcontent:sub(80,89), content),"cur -50")
+
+  f:seek("end", -50)
+  content = f:read(10)
+  ok(eq(testcontent:sub(#testcontent-50,#testcontent-41), content),"end -50")
```
HHHartmann, line 98: This test fails for me

```log
 ==>  fail seek file LFS: end -50: different strings expected oot root   vs. ot root  8
```

HHHartmann (changes requested):
Sorry for the long delay.
Now again reviewed and tested and found some issues but then we should be done I think.

lua_modules/file_lfs/file_lfs.lua
```diff
@@ -0,0 +1,159 @@
...
+end
+
+file_lfs.stat = function(filename)
+  if node_LFS_resource(filename) ~= nil and not file_exists(filename) then
+    return {
+      name = filename,
+      size = #node_LFS_resource(filename),
+      time = {day = 1, hour = 0, min = 0, year = 1970, sec = 0, mon = 1},
+      is_hidden = false, is_rdonly = true, is_dir = false, is_arch = false, is_sys = false,
+    }
```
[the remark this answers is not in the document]
vsky279: Good point. Why not.

tests/NTest_file_lfs.lua
vsky279: It's interesting. This version did run for me. Anyway it's adjusted as requested.

tests/NTest_file_lfs.lua
vsky279: There was an error in the code finding the right position in testcontent (`testcontent:sub(#testcontent-50,#testcontent-41)` ). It needs to be `testcontent:sub(#testcontent+1-50,#testcontent+1-41)`.

vsky279:
Gregor, thanks for this in-deep review.  :thumbsup:
Everything should be handled now.

lua_modules/file_lfs/file_lfs.lua
vsky279: I think it is usefull. ESPlorer list all files including LFS ones.
Sometimes it's needed to list SPIFFs only files:
`l = file.list(nil, true);for k,v in pairs(l) do  print("name:"..k..", size:"..v) end`

vsky279:
I can't find anywhere what is the limitation for the string length in LFS.
53kB string is too long - LFS image that includes such string cannot be flashed.

lua_modules/file_lfs/file_lfs.lua
```diff
@@ -3,7 +3,7 @@ local FILE_READ_CHUNK = 1024
 local _file = file
 local file_exists, file_open, file_getcontents, file_rename, file_stat, file_putcontents, file_close, file_list =
   _file.exists, _file.open, _file.getcontents, _file.rename, _file.stat, _file.putcontents, _file.close, _file.list
-local node_LFS_resource = node.LFS.resource or function() end -- luacheck: ignore
+local node_LFS_resource = node.LFS.resource or function() return {} end -- luacheck: ignore
```
HHHartmann, line 6: This should only return `{}` if called without param. else `nil`

lua_modules/file_lfs/file_lfs.lua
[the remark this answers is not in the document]
HHHartmann: Either also add it to the file:stat result or adapt the documentation. I have a slight bias towards adapting the documentation.

HHHartmann:
Looking forward to this working. Will use it in my current project to speed up startup.

lua_modules/file_lfs/file_lfs.lua
vsky279: Mea culpa.

lua_modules/file_lfs/file_lfs.lua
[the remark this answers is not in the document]
vsky279: I vote also for adjusting the documentation. It is easy to test for `nil` or `false`, i.e. `~=true`, rather than to have additional cost with adjusting the original `file.stat()` result.

vsky279:
Can we merge this? I think everything has been adressed.

HHHartmann:
Yes. Sorry for the delay

PR #3334, opened 2020-11-22
DHT module bug fix

### LINKED ISSUE #3328: dht module returns checksum errors/weird values
I recently updated, now I'm getting checksum errors with the dht module.

On the old firmware, I suspect the readings are way off, more than the ±2°C tolerance which are specified for the sensor iirc. Is this normal?

In case it matters, this is a WeMos D1 mini and 3 DHT11 sensors.

### Expected behavior

NodeMCU 3.0.0.0 
	branch: master
	commit: 3d917850180f67adc7f2c6b5d00f27c152e7194c
	release: 3.0-master_20190907 +2
	release DTS: 202002231746
	SSL: false
	build type: float
	LFS: 0x0
 build 2020-03-17 02:44 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
lua: cannot open init.lua
   
> 
> for i=5,7 do print(dht.read11(i)) end
0	24	34	0	0
0	25	33	0	0
0	23	38	0	0

### Actual behavior
NodeMCU 3.0.0.0 
	branch: master
	commit: 4f6792773f93f36a7255cfd28dca7aa6c4aa9552
	release: 3.0-release_20201107
	release DTS: 202011071523
	SSL: false
	build type: float
	LFS: 0x0 bytes total capacity
 build 2020-11-12 01:20 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
cannot open init.lua: 
> for i=5,7 do print(dht.read11(i)) end
0	614.4	870.4	400	400
-1	614.4	896	400	0
-1	742.4	1203.2	400	200
> for i=5,7 do print(dht.read11(i)) end
-1	793.6	1203.2	600	200
-1	793.6	1203.2	600	200
-1	793.6	1203.2	600	200
> for i=5,7 do print(dht.read11(i)) end
-2	-999	-999	0	0
-2	-999	-999	0	0
-2	-999	-999	0	0
> for i=5,7 do print(dht.read11(i)) end
-1	793.6	1203.2	600	200
-1	793.6	1203.2	600	200
-1	793.6	1203.2	600	200
> for i=5,7 do print(dht.read11(i)) end
-2	-999	-999	0	0
-2	-999	-999	0	0
-2	-999	-999	0	0
> for i=5,7 do print(dht.read11(i)) end
-1	793.6	1203.2	600	200
-1	793.6	1203.2	600	200
-1	793.6	1203.2	600	200

vsky279:
Fixes #3328.

After #3210 DHT module started to return checksum errors after the first readout. The registers buffer was not properly zeroed.
```patch
    // EMPTY BUFFER
-   memset(dht_bytes, sizeof(uint8_t)*5, 0);
+   memset(dht_bytes, 0, sizeof(uint8_t)*5);
```

I have added some debugging stuff.

### DIFF
--- a/docs/modules/dht.md
+++ b/docs/modules/dht.md
@@ -10,6 +10,7 @@ Constants for various functions.
 
 ## dht.read()
 Read all kinds of DHT sensors, including DHT11, 21, 22, 33, 44 humidity temperature combo sensor.
+Return correct readout except for DHT12 and negative temperatures by DHT11.
 
 #### Syntax
 `dht.read(pin)`
@@ -74,8 +75,32 @@ Read DHT11 humidity temperature combo sensor.
 #### See also
 [dht.read()](#dhtread)
 
+## dht.read12()
+Read DHT12 humidity temperature combo sensor.
+
+#### Syntax
+`dht.read12(pin)`
+
+#### Parameters
+`pin` pin number of DHT12 sensor (can't be 0), type is number
+
+#### Returns
+- `status` as defined in Constants
+- `temp` temperature (see note below)
+- `humi` humidity (see note below)
+- `temp_dec` temperature decimal
+- `humi_dec` humidity decimal
+
+!!! note
+
+    If using float firmware then `temp` and `humi` are floating point numbers. On an integer firmware, the final values have to be concatenated from `temp` and `temp_dec` / `humi` and `hum_dec`.
+
+#### See also
+[dht.read()](#dhtread)
+
+
 ## dht.readxx()
-Read all kinds of DHT sensors, except DHT11.
+Read all kinds of DHT sensors, except DHT11 and DHT12. Differs from `dht.read()` only by waiting only sufficient 1 ms for sensor wake-up while `dht.read()` waits universal 18 ms.
 
 ####Syntax
 `dht.readxx(pin)`
--- a/docs/modules/dht.md
+++ b/docs/modules/dht.md
@@ -9,8 +9,8 @@ Constants for various functions.
 `dht.OK`, `dht.ERROR_CHECKSUM`, `dht.ERROR_TIMEOUT` represent the potential values for the DHT read status
 
 ## dht.read()
-Read all kinds of DHT sensors, including DHT11, 21, 22, 33, 44 humidity temperature combo sensor.
-Return correct readout except for DHT12 and negative temperatures by DHT11.
+Reads all kinds of DHT sensors, including DHT11, 21, 22, 33, 44 humidity temperature combo sensor.
+Returns correct readout except for DHT12 and negative temperatures by DHT11. Use [`dht.read12()`](#dhtread12) and  [`dht.read11()`](#dhtread11) instead. It is to use model specific read function anyway.
 
 #### Syntax
 `dht.read(pin)`
--- a/app/dht/dht.c
+++ b/app/dht/dht.c
@@ -41,7 +41,7 @@
 #define HIGH    1
 #endif /* ifndef HIGH */
 
-#define COMBINE_HIGH_AND_LOW_BYTE(byte_high, byte_low)  (((byte_high) << 8) | (byte_low))
+#define COMBINE_HIGH_AND_LOW_BYTE(byte_high, byte_low)  ((uint16_t)((byte_high) << 8) | (byte_low))
 
 static double dht_humidity;
 static double dht_temperature;
@@ -73,10 +73,15 @@ double dht_getTemperature(void)
 // DHTLIB_OK
 // DHTLIB_ERROR_CHECKSUM
 // DHTLIB_ERROR_TIMEOUT
-int dht_read_universal(uint8_t pin)
+int dht_read(uint8_t pin, dht_type type)
 {
     // READ VALUES
-    int rv = dht_readSensor(pin, DHTLIB_DHT_UNI_WAKEUP);
+    int rv = dht_readSensor(pin, 
+      type == DHT22 ? DHTLIB_DHT_WAKEUP : 
+      type == DHT11 ? DHTLIB_DHT11_WAKEUP :
+      DHTLIB_DHT_UNI_WAKEUP
+    );
+    
     if (rv != DHTLIB_OK)
     {
         dht_humidity    = DHTLIB_INVALID_VALUE;  // invalid value, or is NaN prefered?
@@ -84,16 +89,10 @@ int dht_read_universal(uint8_t pin)
         return rv; // propagate error value
     }
 
-#if defined(NODE_DEBUG)
-    int i;
-    for (i = 0; i < 5; i++)
-    {
-        NODE_DBG("%x\t", dht_bytes[i]);
-    }
-    NODE_DBG("\n");
-#endif // defined(NODE_DEBUG)
+    NODE_DBG("DHT registers: %x\t%x\t%x\t%x\t%x == %x\n", dht_bytes[0], dht_bytes[1], dht_bytes[2], dht_bytes[3], dht_bytes[4], (uint8_t)(dht_bytes[0] + dht_bytes[1] + dht_bytes[2] + dht_bytes[3]));
 
-    // Assume it is DHT11
+    // Assume it is special case of DHT11,
+    // i.e. positive temperature and dht_bytes[3] == 0 ((dht_bytes[3] & 0x0f) * 0.1 to be added to temperature readout)
     // If it is DHT11, both temp and humidity's decimal
     if ((dht_bytes[1] == 0) && (dht_bytes[3] == 0))
     {
@@ -118,14 +117,41 @@ int dht_read_universal(uint8_t pin)
         }
     }
 
-    // Assume it is not DHT11
+    // Assume it is not DHT11 special case
     // CONVERT AND STORE
     NODE_DBG("DHTxx method\n");
-    dht_humidity = (double)COMBINE_HIGH_AND_LOW_BYTE(dht_bytes[0], dht_bytes[1]) * 0.1;
-    dht_temperature = (double)COMBINE_HIGH_AND_LOW_BYTE(dht_bytes[2] & 0x7F, dht_bytes[3]) * 0.1;
-    if (dht_bytes[2] & 0x80)  // negative dht_temperature
-    {
-        dht_temperature = -dht_temperature;
+
+    switch (type) {
+      case DHT11:
+      case DHT12:
+        dht_humidity += dht_bytes[1] * 0.1;
+        break;
+      default:
+        dht_humidity = COMBINE_HIGH_AND_LOW_BYTE(dht_bytes[0], dht_bytes[1]) * 0.1;
+        break;
+    }
+
+    switch (type) {
+      case DHT11:
+        if (dht_bytes[3] & 0x80) {
+          dht_temperature = -1 - dht_temperature;
+        }
+        dht_temperature += (dht_bytes[3] & 0x0f) * 0.1;
+        break;
+      case DHT12:
+        dht_temperature += (dht_bytes[3] & 0x0f) * 0.1;
+        if (dht_bytes[2] & 0x80)  // negative dht_temperature
+        {
+            dht_temperature *= -1;
+        }
+        break;
+      default: // DHT22, DHT_NON11
+        dht_temperature = COMBINE_HIGH_AND_LOW_BYTE(dht_bytes[2] & 0x7F, dht_bytes[3]) * 0.1;
+        if (dht_bytes[2] & 0x80)  // negative dht_temperature
+        {
+            dht_temperature *= -1;
+        }
+        break;
     }
 
     // TEST CHECKSUM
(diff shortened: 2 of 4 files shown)


marcelstoer (changes requested):
Can you get rid of `DHT_DEBUG` in `dht.h`?

vsky279:
Also I was a bit suspicious about the negative values fix. It was a bit difficult to imagine that such a bug would remain undiscovered for such a long time.
So I did practical test and with **my** DHT22 the "negative values bug fix" introduced by #3210 is wrong. Negative values are calculated as something above 33000.

So the next commit is practically reverting #3210. I must admit that I did not like the "optimisation" that it introduced.
In two places it replaced simple assignment
```c
    dht_humidity    = dht_bytes[0];  // dht_bytes[1] == 0;
    dht_temperature = dht_bytes[2];  // dht_bytes[3] == 0;
```
with calling more complex function
```c
        dht_humidity    = getValue(Humidity8);
        dht_temperature = getValue(Temperature8);
```
I don't see the point. 
In two other places it replaces simple operation `high <<8 | low` with this function too.

For checksum calculation which is basically summing 2 or 4 numbers it introduces special function where 4 numbers are summed also in cases when just summing 2 is sufficient (the two other registers are 0).

The code got obfuscated. I don't see the value added. 

If any optimisation should occur then I would vote for removing `dht_read11` and `dht_read` and leave just `dht_read_universal` which covers all sensors.

marcelstoer:
I guess at this point it would interesting to also get some feedback from @emdioh (#3210's author).

nwf:
Are there `NTest`-friendly predicates that could be added to ensure that basic handling of DHT hardware is functioning?  Obviously exact sensor data will vary from test run to run, but something like "the sensor responds twice without raising CRC errors" seems plausible?  Feel free to propose something by patching https://github.com/nodemcu/nodemcu-firmware/blob/dev/tests/NodeMCU_Test_Environment.rst ; perhaps the DHT can be multiplexed on DUT0 pin 2 with its power controlled via the 23017 (say, pin B2).

emdioh:
Sorry about the late reply and sorry about messing up the memset: I wasn't checking the CRC so I did not realise the issue.

I wrote #3210 because my DHT22 was returning -3276C for -1C and I saw that the number was growing towards 0 while the temperature decreased, hence I figured the numeric encoding issue.

If this change reverts the logic to the previous state, it will break the module DHT module at least for me.

The other changes tried to encapsulate the bit mangling into a standalone function: feel free to change it if you like the previous way more.

At this point it would be interesting to understand why I am getting values encoded in a different way than vsky279 and possibly others, but I don't expect to reasonably have time to look into this before the Christmas period.

vsky279:
@emdioh The negative values issue is really strange. I can't find any example for DHT22 where negative values are treated as you suggest.

Please see [dht-sensor.cpp#L190](https://github.com/momenso/node-dht-sensor/blob/f697cf48e2e43f07eca4d507f97a7ee2164bf23b/src/dht-sensor.cpp#L190) or [Adafruit_DHT.c#L143](https://github.com/born2net/raspberry/blob/e18f838ae4c584dcddbb430426411682880e5291/Adafruit-Raspberry-Pi-Python-Code/Adafruit_DHT_Driver/Adafruit_DHT.c#L143)
```c
if (data[2] & 0x80) f *= -1;
```
or [adafruit_dht.py#227](https://github.com/adafruit/Adafruit_CircuitPython_DHT/blob/85a8f329346c074f6a0be132077269e62a19b89e/adafruit_dht.py#L227)
```python
                if buf[2] & 0x80:
                    new_temperature = -new_temperature
```

However in this library [DHT.cpp#L92](https://github.com/adafruit/DHT-sensor-library/blob/b64fe15ad0972902c9618ee4dab3e48d2b2259c2/DHT.cpp#L92) there is probably the treatment you suggest for DHT11 sensor:
```c
 switch (_type) {
    case DHT11:
      f = data[2];
      if (data[3] & 0x80) {
        f = -1 - f;
      }
      f += (data[3] & 0x0f) * 0.1;
      if (S) {
        f = convertCtoF(f);
      }
      break;
    case DHT12:
      f = data[2];
      f += (data[3] & 0x0f) * 0.1;
      if (data[2] & 0x80) {
        f *= -1;
      }
      if (S) {
        f = convertCtoF(f);
      }
      break;
    case DHT22:
    case DHT21:
      f = ((word)(data[2] & 0x7F)) << 8 | data[3];
      f *= 0.1;
      if (data[2] & 0x80) {
        f *= -1;
      }
      if (S) {
        f = convertCtoF(f);
      }
      break;
    }
```

That might be an explanation. Do you have a DHT11?

emdioh:
I just checked my order and I have bought a DHT22.
One cannot be sure of getting what bought these days, but at least the external size  shape and color match the DHT22 package.
Not sure what's going on with the data.
If I am the only one having this issue, maybe I have a fake DHT22, idk.

marcelstoer:
If it walks and talks like a DHT11 then it probably _is_ a DHT11 😉

vsky279:
I'll post a commit so you can try if the updated version works correctly with your device. I'll make sure it works correctly with mine :wink:

vsky279:
Ok, it is more complex then I thought. I'm posting a commit which:
* implements support for DHT12
* leaves only `dht_read()` C function with DHT sensor type as parameter
* Lua `dht.read()` works correctly for majority of sensors (except for DHT12 and negative temperatures measured by DHT11).

@emdioh could you please test with your sensor? Please try `dht.read11` function for freezing temperatures.

emdioh:
@marcelstoer yeah I think I got sent a fake.
@vsky279 thanks for fixing this. 
I don't have the sensor with me: I'll test it during the weekend.
Please note that the DHT11 it does not officially support negative temperatures.

ezbella:
Hi everybody! Thanks @vsky279 for your time reviewing this issue :) I wanted to know which is the best approach to make this module work while waiting for @emdioh verification. I'm generating the firmware from marcel's build service. 
Thanks in advance

vsky279:
@ezbella Hi, if you send me the configuration you need, I can build you the image with patched DHT so that you can test it.

ezbella:
@vsky279 Thanks a lot! :)
The modules are (integer version):
bit, cron, file, gpio, i2c, net, node, rtctime, tmr, uart, wifi, dht

I will test it and report ASAP.
Thanks again!

vsky279:
@ezbella Here you go!
[nodemcu_integer_c_dhtfix_20201130-1133.bin](https://gist.github.com/vsky279/a1d91805c3b80cea6a5ce61e108fee29/raw/4b791545a62f040e67ab47d1a34ad29acde820e9/nodemcu_integer_c_dhtfix_20201130-1133.bin)

```
	branch: c_dhtfix
	commit: 7cec401e9404c0e21289a87d6e75ad0a95739485
	release: 3.0-master_20200910 +45
	release DTS: 202011232310
	SSL: false
	build type: integer
	LFS: 0x0 bytes total capacity
	modules: bit,cron,dht,file,gpio,i2c,net,node,rtctime,spi,tmr,uart,wifi
 build 2020-11-30 11:34 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
```

ezbella:
@vsky279 I am still having issues with this version. The checksum error was fixed, but the sensor is throwing values higher than 500 almost all the time, but the weird thing is that sometimes, it outputs right values (see screenshot below). I have tested it with 2 different dht11 sensors that showed to function properly with the previous working version.
**This is how I am reading the sensor:**
```
local dht11Pin = 6
gpio.mode(dht11Pin, gpio.OUTPUT)
tmr.create():alarm(3000, tmr.ALARM_AUTO, function(t)
    local status,temp,humi = dht.read(dht11Pin)
    print("status: " .. status .. " - temp: " .. temp .. " - humi: " .. humi)
end )
```

**The output:**
dht11_output

ezbella:
@vsky279  I discovered that with the old working version of dht module, if I call the plain dht.read(pin) (instead of dht.read11(pin)), it throws the same high values than those of the new failing version. Perhaps this version is omitting some important dht11 handling that the previous version had.
Hope it helps!

vsky279:
@ezbella Thanks for this testing. Can you send me an email to lukas@voborsky.name so that we can discuss w/o spamming this PR?

vsky279:
Can you please check the output with enabled NODE_DEBUG?
The fw: [file-nodemcu_integer_c_dhtfix_20201130-1806-bin](https://gist.github.com/vsky279/a1d91805c3b80cea6a5ce61e108fee29#file-nodemcu_integer_c_dhtfix_20201130-1806-bin)

vsky279:
@ezbella Ezequiel, thanks for testing. There was a bug for the DHT11 readouts. Please use the `read11` function.
Also can you test negative values? If `read11` is wrong for negative values then try `read12`.

I doubt that the readouts before #3210 would be correct. It should give exactly the same result as it is giving now (temperatures above 500).

The new fw: [nodemcu_integer_c_dhtfix_20201201-0717.bin](https://gist.github.com/vsky279/a1d91805c3b80cea6a5ce61e108fee29/raw/b1ea26e6e702d0002815e2dd384219fca8d8eec0/nodemcu_integer_c_dhtfix_20201201-0717.bin)

ezbella:
I can confirm that all the readouts with DHT11 are now working properly with the last [change ](https://gist.github.com/vsky279/a1d91805c3b80cea6a5ce61e108fee29/raw/b1ea26e6e702d0002815e2dd384219fca8d8eec0/nodemcu_integer_c_dhtfix_20201201-0717.bin) :) Good job @vsky279!
I also tested negative values and decimal part and are also handled properly.

```
local dht11Pin = 6
gpio.mode(dht11Pin, gpio.OUTPUT)
tmr.create():alarm(3000, tmr.ALARM_AUTO, function(t)
    local status, temp, humi, tempdec, humidec = dht.read11(dht11Pin, DHT11)
    print("status: " .. status .. " - temp: " .. temp .. "," .. tempdec .. " - humi: " .. humi .. "," .. humidec)
end )
```

Output:
```
status: 0 - temp: 16,199 - humi: 92,0
DHT registers: 5c	0	10	2	6e == 6e
DHTxx method
status: 0 - temp: 16,199 - humi: 92,0
DHT registers: 5c	0	10	2	6e == 6e
DHTxx method
status: 0 - temp: 16,199 - humi: 92,0
DHT registers: 5c	0	10	3	6f == 6f
DHTxx method
status: 0 - temp: 16,300 - humi: 92,0
DHT registers: 5c	0	10	3	6f == 6f
DHTxx method
status: 0 - temp: 16,300 - humi: 92,0
DHT registers: 5c	0	10	3	6f == 6f
DHTxx method
status: 0 - temp: 16,300 - humi: 92,0
DHT registers: 5c	0	10	4	70 == 70
DHTxx method
status: 0 - temp: 16,399 - humi: 92,0
DHT registers: 5c	0	10	4	70 == 70
```

vsky279:
Ok, good. So this is the commit that @ezbella helped to debug. Thank you.

@emdioh Are you still willing to test?

We have now tested both DHT22 and DHT11 so I think this can land in `dev`.

docs/modules/dht.md
```diff
@@ -10,6 +10,7 @@ Constants for various functions.
 
 ## dht.read()
 Read all kinds of DHT sensors, including DHT11, 21, 22, 33, 44 humidity temperature combo sensor.
+Return correct readout except for DHT12 and negative temperatures by DHT11.
```
marcelstoer, line 13: Maybe point to the specific `read` functions for those from here then?

vsky279: Like this?

marcelstoer: Yep, thanks.

wilhelmy:
I've just updated to dev and applied this patch, I'm getting weird values with two dht11 still: 

```
> for i=5,6 do print(dht.read11(i)) end
0	24.6	30	600	0
0	26.8	29	800	0
> for i=5,6 do print(dht.read11(i)) end
0	24.7	31	699	0
0	26.6	30	600	0
```

It is nowhere near ±2°C range of 24-26°C where I am measuring and I can't really wrap my head around why one of the values is floating point while the other isn't?

marcelstoer:
[As I said](https://github.com/nodemcu/nodemcu-firmware/issues/3328#issuecomment-744714326) it's not _yet_ on `dev` 😉 This PR would otherwise be marked as merged. We would definitely appreciate more feedback - thanks for testing. However, for now you would need to build your own firmware from https://github.com/vsky279/nodemcu-firmware/tree/c_dhtfix (the source of this PR). Or use the latest binary @vsky279 offered for download in the most recent comment above.

wilhelmy:
I read what you said and then updated to dev and applied this patch.

vsky279:
@wilhelmy
The 600-800 values are decimals of the readout. This is useful when you have the integer version. There the output would be
```
> for i=5,6 do print(dht.read11(i)) end
0	24	30	600	0
0	26	29	800	0
```
You ignore it when you are on float (e.g. `status, temp, humi = read11(5)`)

I find DHT thermometers not very reliable. It also depends how often you perform the readout as it tends to heat itself and distort the readout.

It would be nice if you can test negative temperatures.

wilhelmy:
But shouldn't fields 3 and 4 only contain decimals in an integer build, and be zero on float builds? I have a float/double build (I didn't change anything on top of dev except for applying your patches and enabling sntp and rtctime), and the temp value is a floating point value with a fractional part on my system, still there's something in the third column (but not the fourth).

It has been warm lately (6°C yesterday) but I'll try with sub-zero temperatures as soon as it gets cold enough.

vsky279:
Nope. Even in float the decimals are returned as result 3 and 4. See the code, it does not make any difference:
```c
static void aux_read( lua_State *L )
{
  double temp = dht_getTemperature();
  double humi = dht_getHumidity();
  int tempdec = (int)((temp - (int)temp) * 1000);
  int humidec = (int)((humi - (int)humi) * 1000);
  lua_pushnumber( L, (lua_Float) temp ); 
  lua_pushnumber( L, (lua_Float) humi ); 
  lua_pushinteger( L, tempdec );
  lua_pushinteger( L, humidec );
}
```

I am not sure we should be waiting with the merge until it starts freezing. It's not beginning of 20th century. Can't you borrow something from your freezer and use it to cold the sensor? 😄

wilhelmy:
In order to test I attached the whole thing to a power bank and put it into my freezer. It reports -10.7 and -8.4°C and 37 and 33% humidity respectively. Hope that answers more questions than it raises ;)

marcelstoer:
Given the overwhelming positive feedback I feel this should now land.

Michal78S:
I have found problem for the DHT library working with integer firmware. For temperature above 0C everythink is OK, but below 0C happends something strange. When using this formula: 
if status == dht.OK then
    -- Integer firmware using this example
    print(string.format("DHT Temperature:%d.%03d;Humidity:%d.%03d\r\n",
          math.floor(temp),
          temp_dec,
          math.floor(humi),
          humi_dec
    ))
Got i.e. "0.-200C" for values between 0 and -1C. For -1 (and other round numbers) got "-1.0C" but for intermediate values got "-1.-400C"

I made small patch and i'm putting data to the server with formula:
string.format("%d.%03d", T, match.abs(Tdec))

It almost works, only between 0 and -1 got invertet readings, please look at printscreen:
dht_lack
 At 4:47 a.m should be -0,8C

vsky279:
Hi, it's definitely better to use the float version - some local OGs told me few years ago. 
However if you insist on decimal version you could try something like this
```Lua
print(string.format("DHT Temperature: %s%d.%03d\r\n",
   (temp_dec<0 and -1 or 1)*(temp<0 and -1 or 1)<0 and "-" or "",
   math.abs(temp),
   math.abs(temp_dec)
))
```

Michal78S:
Thanks. This code is only a small part of lua script for weather station. It registers ostdide temp, humidity, sun radiation, wind max speed and average wind. You can find it on-line here: https://pvmonitor.pl/i_user.php?idinst=10898#/pv0

I'm going to use float firmware but I have to check that other readings won't lose accuracy.

Michal78S:
I have moved to float firmvare. It almost works. Positive readings are OK, but for negative teperatures the right one sensor shows corectly but the left one gives values over 400 Celcium degrees... 0C is showing as 409.8C 
IMG_20231208_102835

I have tried dht.read() and dth.readxx(). Both gives the same. dht.read11() showing nonsense values.

Michal78S:
Hello again,
I have replaced DHT22 by SHT35. It works great despite freezing winter. Significantly better than BME280 which has problems with high humidity and malfunction soon when used outdoor. Also is easier to buy oryginal than DHT22.
Schowek02

PR #3346, opened 2020-12-12
Convert test documentation to Markdown

pjsg:

I converted the .rst to .md using an online converter and then had to clean up the table.I will be putting up another PR at some point to add the extra connections that I added on the test boards.


marcelstoer:
Thanks Philip! Again, I'm late to the party, sorry.

Could/should we rename this to README and add an overview chapter to address https://github.com/nodemcu/nodemcu-firmware/pull/2984#issuecomment-739581626?

PR #3348, opened 2020-12-13
Softuart module setup function parameters bug fix

vsky279:
Fixes no reported issue.

The following code is failing thought it was inline with documentation:
```Lua
s = softuart.setup(115200, 3, nil)
s = softuart.setup(115200, 3)
```

The first call was failing as the third argument had to be a number or none. `nil` was not acccepted (but should be).
The second call failed as `softuart` did not accept to be setup `rx` only.

### DIFF
--- a/app/modules/softuart.c
+++ b/app/modules/softuart.c
@@ -194,6 +194,7 @@ static int softuart_init(softuart_t *s)
 		}
 		return platform_gpio_register_intr_hook(mask, softuart_intr_handler);
     }
+    return 1;
 }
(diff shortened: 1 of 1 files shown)


nwf:
LGTM

galjonsfigur:
Also looks good to me - when I was making #3104 I somehow missed this. Thanks for fixing this! Those kind of bugs and regressions could easily be found by simple tests - I wish I have time to write them especially in the lights of PR like #3324

PR #3352, opened 2020-12-20
NFC: Backport luaL_testudata to Lua 5.1

nwf:
And move luaL_checkudata to it, as in 5.3.


HHHartmann:
Unfortunately `typeerror` also does not seem to be part of Lua 5.1

nwf:
OK.  Having stopped to actually read C, it looks like this is now CI green.  Opinions?

nwf:
Any objections to landing this?  It should be fairly devoid of consequence...

HHHartmann:
LGTM

PR #3354, opened 2020-12-24
Add pipeutils example

nwf:
These have been handy within the test harness for streaming files and LFS images across to the DUTs.  I am not wedded to the file location, its name, or even its exact contents, but I'd like to have them in tree for the test harness's use.

Right now, the test harness's file xfer mechanism arranges to take over the UART for the duration of the transfer and push all rx'd bytes through `debase64` and then `chunker` before committing to SPIFFS: this means that it's easy to move arbitrary data using only printable ASCII codepoints and that the FLASH is almost always given blocks far larger than the UART's rx buffer or even a small Lua string (256 bytes), minimizing wear.  The relevant glue code is landed on the module by the test harness (with fallbacks if `pipeutils` cannot be `require`-d), but could also be added to the `pipeutils` example itself if people would rather.

### DIFF
--- a/lua_examples/pipeutils.lua
+++ b/lua_examples/pipeutils.lua
@@ -1,0 +1,52 @@
+-- A collection of pipe-based utility functions
+
+-- A convenience wrapper for chunking data arriving in bursts into more sizable
+-- blocks; `o` will be called once per chunk.  The `flush` method can be used to
+-- drain the internal buffer.  `flush` MUST be called at the end of the stream,
+-- **even if the stream is a multiple of the chunk size** due to internal
+-- buffering.  Flushing results in smaller chunk(s) being output, of course.
+local function chunker(o, csize, prio)
+  assert (type(o) == "function" and type(csize) == "number" and 0 <= csize)
+  local p = pipe.create(function(p)
+              -- wait until it looks very likely that read is going to succeed
+              -- and we won't have to unread.  This may hold slightly more than
+              -- a chunk in the underlying pipe object.
+              if 256 * (#p - 2) <= csize then return nil end
+
+              local d = p:read(csize)
+              if #d < csize
+               then p:unread(d) return false
+               else o(d)        return true
+              end
+            end, prio or node.task.LOW_PRIORITY)
+
+  return {
+    flush = function() for d in p:reader(csize) do o(d) end end,
+    write = function(d) p:write(d) end
+  }
+end
+
+-- Stream and decode lines of base64, calling `o(data)` with decoded chunks or
+-- calling `e(badinput, errorstr)` on error; the error callback must ensure
+-- that this conduit is never written to again.
+local function debase64(o, e, prio)
+  assert (type(o) == "function" and type(e) == "function")
+  local p = pipe.create(function(p)
+              local eds = p:read("\n")
+              while #eds > 0 do
+                local ed
+                ed, eds = eds:match("^%s*(%S*)%s*(.*)$")
+                if #ed == 0 then break end
+                local ok, d = pcall(encoder.fromBase64, ed)
+                if ok then o(d) else e(ed, d); p = nil; return false end
+              end
+              p:unread(eds)
+              return false
+            end, prio or node.task.LOW_PRIORITY)
+  return { write = function(d) p:write(d) end }
+end
+
+return {
+  chunker = chunker,
+  debase64 = debase64,
+}
--- a/lua_examples/pipeutils.lua
+++ b/lua_examples/pipeutils.lua
@@ -6,7 +6,7 @@
 -- **even if the stream is a multiple of the chunk size** due to internal
 -- buffering.  Flushing results in smaller chunk(s) being output, of course.
 local function chunker(o, csize, prio)
-  assert (type(o) == "function" and type(csize) == "number" and 0 <= csize)
+  assert (type(o) == "function" and type(csize) == "number" and 1 <= csize)
   local p = pipe.create(function(p)
               -- wait until it looks very likely that read is going to succeed
               -- and we won't have to unread.  This may hold slightly more than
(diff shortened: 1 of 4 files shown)


lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,52 @@
+-- A collection of pipe-based utility functions
+
+-- A convenience wrapper for chunking data arriving in bursts into more sizable
+-- blocks; `o` will be called once per chunk.  The `flush` method can be used to
+-- drain the internal buffer.  `flush` MUST be called at the end of the stream,
+-- **even if the stream is a multiple of the chunk size** due to internal
+-- buffering.  Flushing results in smaller chunk(s) being output, of course.
+local function chunker(o, csize, prio)
+  assert (type(o) == "function" and type(csize) == "number" and 0 <= csize)
```
HHHartmann, line 9: Size should be at least 1

lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,52 @@
...
+              end
+            end, prio or node.task.LOW_PRIORITY)
+
+  return {
+    flush = function() for d in p:reader(csize) do o(d) end end,
+    write = function(d) p:write(d) end
+  }
+end
+
+-- Stream and decode lines of base64, calling `o(data)` with decoded chunks or
```
HHHartmann, line 29: lines of base64 _chunks_
Maybe give an example what is allowed

lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,52 @@
...
+-- calling `e(badinput, errorstr)` on error; the error callback must ensure
+-- that this conduit is never written to again.
+local function debase64(o, e, prio)
+  assert (type(o) == "function" and type(e) == "function")
+  local p = pipe.create(function(p)
+              local eds = p:read("\n")
```
HHHartmann, line 35: What if write wrote an incomplete line? The function would be called anyways and return an incomplete record. Maybe use "\n+" and check for existence of the \n.

lua_examples/pipeutils.lua
nwf: That's fair enough.

lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,54 @@
...
+-- drain the internal buffer.  `flush` MUST be called at the end of the stream,
+-- **even if the stream is a multiple of the chunk size** due to internal
+-- buffering.  Flushing results in smaller chunk(s) being output, of course.
+local function chunker(o, csize, prio)
+  assert (type(o) == "function" and type(csize) == "number" and 1 <= csize)
+  local p = pipe.create(function(p)
+              -- wait until it looks very likely that read is going to succeed
+              -- and we won't have to unread.  This may hold slightly more than
+              -- a chunk in the underlying pipe object.
+              if 256 * (#p - 2) <= csize then return nil end
```
HHHartmann, line 14: What does #p return? That seems like an implementation dependent undocumented feature.
From the manual #userdata seems to return the size of the allocated block. But this usage implies a number of blocks.

Maybe pipe should get a dedicated size method.

nwf: https://github.com/nodemcu/nodemcu-firmware/blob/4f6792773f93f36a7255cfd28dca7aa6c4aa9552/app/modules/pipe.c#L252 ; though I see it's not documented despite being implemented for Lua and not any internal reason (AFAICT).  I can add that.  Something like the following, maybe?

"The length (`__len`, `#`) of a `pipe` object is the number of records pending; this may be useful for estimating the total size of data in the `pipe`."

HHHartmann: Yes but what is a record in this case? Is it a memory chunk which contains 256 chars except for the last one? or is it something more complicated?. 
Maybe the __len implementation should be "fixed" to just return the remaining chars, not reveling the internal structure of the pipe.

HHHartmann: Reding the code it seems that the pipe object is a table which contains chunks of 256 bytes (LUAL_BUFFERSIZE), so #p would return the number of chunks in use +1 since the first index is used for the reader callback function.

The  `pipe__len` returns the used length of an individual chunk.

Maybe a __len method for the pipe itself would make sense, returning the total amount of buffered chars. 
As workaround, using the undocumented feature #p[n] (which returns the used length of chunk n) this code would return the exact amount of buffered chars. (haven't tested it, so it might not even compile)

```Lua
local len = 0
for i=2,#p,1 do
  len = len + #[i]
end
```

In general I would not want to use code which is not documented and relies on the implementation of something.

lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,54 @@
...
+
+-- Stream and decode lines of complete base64 blocks, calling `o(data)` with
+-- decoded chunks or calling `e(badinput, errorstr)` on error; the error
+-- callback must ensure that this conduit is never written to again.
+local function debase64(o, e, prio)
+  assert (type(o) == "function" and type(e) == "function")
+  local p = pipe.create(function(p)
+              local eds = p:read("\n+")
+              if eds:sub(-1) == "\n" then -- guard against incomplete line
+                while #eds > 0 do -- while there are complete lines...
```
HHHartmann, line 37: There should be only one line after reading to '\n' How would the lines look? are there several chunks on one line?
Please give an example.

nwf: The `while` loop is probably stale and dates back to when `:read()` was returning arbitrary string segments.  I think it'd be fine to remove it now that `:read()` is doing the segmentation.

lua_examples/pipeutils.lua
nwf: There's no reason to compute an exact length here; it's just meant as a heuristic to guard the slower `:read()` below this line.  I'm just trying to avoid the case of, say, character-by-character `:write()`s causing quadratically many `:read(); :unread()` while building up the chunk. 

I think it makes sense to expose the "number of chunks", even if the chunks aren't full (which can happen after, say, `unread()`s).  The worst that happens is we end up causing the `pipe` object to consolidate several partial chunks into fuller chunks, which is probably a net benefit if it does happen, so long as it only happens occasionally.

lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,54 @@
...
+
+              local d = p:read(csize)
+              if #d < csize
+               then p:unread(d) return false
+               else o(d)        return true
+              end
+            end, prio or node.task.LOW_PRIORITY)
+
+  return {
+    flush = function() for d in p:reader(csize) do o(d) end end,
```
HHHartmann, line 24: Doing it this way might lead to problems if `o` relies on being called asynchronously. But it's only an example and as such does not have to be perfect.

lua_examples/pipeutils.lua
```diff
@@ -0,0 +1,54 @@
...
+-- drain the internal buffer.  `flush` MUST be called at the end of the stream,
+-- **even if the stream is a multiple of the chunk size** due to internal
+-- buffering.  Flushing results in smaller chunk(s) being output, of course.
+local function chunker(o, csize, prio)
+  assert (type(o) == "function" and type(csize) == "number" and 1 <= csize)
+  local p = pipe.create(function(p)
+              -- wait until it looks very likely that read is going to succeed
+              -- and we won't have to unread.  This may hold slightly more than
+              -- a chunk in the underlying pipe object.
+              if 256 * (p:nrec() - 1) <= csize then return nil end
```
HHHartmann, line 14: I think a good solution would be to implement a peek method which returns true or false, depending if a matching read would return a complete block according to the params.

PR #3357, opened 2020-12-29
fix regression in luaL_typerror and Change NTest so it can run tests on the host emulating node.task.post

HHHartmann:

### Description of and rationale behind this PR

There is a regression in the luaL_typerror function which results in wrong types being print.
This PR fixes that regression

To back this fix by automatic testing it also enables the NTest system to run on the host using luac.cross.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
     - name: NTest hosttests
       run: |
         cd tests
-        cp NTest/Ntest.lua .
+        cp NTest/NTest.lua .
         ../luac.cross -e NTest_lua.lua | tee log
         grep "failed. 0" log
       shell: bash
@@ -217,7 +217,7 @@ jobs:
     - name: NTest hosttests
       run: |
         cd tests
-        cp NTest/Ntest.lua .
+        cp NTest/NTest.lua .
         ../luac.cross.exe -e NTest_lua.lua | tee log
         grep "failed. 0" log
       shell: bash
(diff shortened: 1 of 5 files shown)


nwf:
Honestly, it might be better to work out a generic host emulation of `node.task.post`?  Either in Lua or in C, replacing https://github.com/nodemcu/nodemcu-firmware/blob/4f6792773f93f36a7255cfd28dca7aa6c4aa9552/app/lua53/lnodemcu.c#L672-L674

HHHartmann:
> Honestly, it might be better to work out a generic host emulation of `node.task.post`? Either in Lua or in C, replacing
> 
> https://github.com/nodemcu/nodemcu-firmware/blob/4f6792773f93f36a7255cfd28dca7aa6c4aa9552/app/lua53/lnodemcu.c#L672-L674

It sure would. If we go the C lane we could also use things like pipe on the host. But as it has been in the Ntest_Ntest selftest already it just moved and can help us out til we have a good solution.

Do you think that is a way to go or should I just dump this PR?

nwf:
Given that this is mostly just code motion and doesn't apply to NTest when running on a NodeMCU device, let's go ahead with this for the moment.  It will take us a while to come up with a `node.task` emulation on the host.

HHHartmann:
See failed tests on the "move specific parts back to NTest_NTest" commit above and the last commit fixing them.
Running the new `tests/NTest_lua.lua` on any Lua 5.1 build will fail without this PR.

I would like to see this in the next release also,

nwf (approved):

PR #3367, opened 2021-01-07
Add check for module docs linkage in mkdocs.yml

HHHartmann:

### Description of and rationale behind this PR

It quite frequently happens that for new modules the linkage to mkdocs.yml is forgotten.
This adds a first check for documentation. More checks might be added later


.github/workflows/build.yml
```diff
@@ -240,3 +240,19 @@ jobs:
...
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-16.04 
+      
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: false
+    - name: all_docks_linked
+      run: ./tools/check_docs.sh
```
marcelstoer, line 256: If the idea is to add future checks to the _same_ file then I'm ok with `check_docs.sh`. If not, then it would have to be more specific.

tools/check_docs.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# get all linked module docs for mkdocs.yml
+grep "modules/" mkdocs.yml | sed "s/ *- .*: *'//" | sed "s/'//" | sort > /tmp/doc
+
+# get all module and lua_module *.md files
+ls docs/modules/* docs/lua-modules/* | sed "sxdocs/xx" | sort > /tmp/files
```
marcelstoer, line 7:
```suggestion
find docs/modules/*.md docs/lua-modules/*.md | sed "sxdocs/xx" | sort > /tmp/files
```

I ran this through shell-check and it complained about https://github.com/koalaman/shellcheck/wiki/SC2012

marcelstoer (changes requested):
Great idea!

tools/check_docs.sh
marcelstoer: Oh...and there's `lua-modules/README.md` 😉

HHHartmann: Saw that too, but it passes the test. If it is to be moved the test will have to be adapted

.github/workflows/build.yml
HHHartmann: Let's keep this in mind and adapt as required if more checks are added.
I could imagine a linkcheck of some kind.
I also would like to extract Lua samples and at least syntax check them, but that would go in the Lua compile step.

marcelstoer (approved):

HHHartmann:
Looks nicer now I think

Issue #2936, opened 2019-10-04
ws2812.write() fails


klauweg:
### Expected behavior
ws2812.write() function should update a connected led String with 60 Leds reliable.

### Actual behavior
In a running script with timer callbacks it does sometimes.
From interactive command prompt it does never. 

### Test code
```
ws2812.write({pin = 15, data=string.char(255,255,255, 0, 0, 0, 255,255,255,0, 0, 0, 255,255,255,0, 0, 0, 255,255,255)})
```

The resulting Output on Pin looks like this:
![grafik](https://user-images.githubusercontent.com/1049892/66244204-ac1ed580-e707-11e9-864d-b6354906ec99.png)

The gap between the two bit bursts shouldn't be there. It varies between 8 and 15 us.
There was nothing running on nodemcu. The mentioned ws2312.write() was the only command i've entered after booting.

### NodeMCU version
On branch dev-esp32
commit a4fa6c5cd3485a2bfb263e0114e94ae1bd332179

### Hardware
ESP32 doit devkit V1.0

klauweg:
placing the same function call as callback in a timer results in another waveform with later gap:
```
mytimer:alarm(700, tmr.ALARM_SINGLE, function() ws2812.write({pin = 15, data=string.char(255,255,255, 0, 0, 0, 255,255,255,0, 0, 0, 255,255,255,0, 0, 0, 255,255,255)}) end)
```
![grafik](https://user-images.githubusercontent.com/1049892/66245121-07eb5d80-e70c-11e9-8105-6d487f0dcabc.png)

Intented with the two screenshots we can see that there is not only a gap but there are bits disappeared even before the large gap.

ulab:
I am experiencing a similar problem. Controlling my WS2812b matrix works fine on NodeMCU on an ESP8266 and with FastLED/Arduino on an ESP32, but the ESP32-dev version just messes with the first two or three LEDs which made me think it's a timing issue.

landswipe:
Experienced similar issue trying to drive 5 led's. 

An interesting observation is that release/v3.3 the problem occurs all the time, however tags/v3.3.2 only occurs when wifi is enabled via wifi:start()

landswipe:
As far as the change in release/v3.3 is concerned, it looks like this commit (a5f9a0cbe4cd62f9a34689e6770ec3f6985e3d44) is the culprit.

landswipe:
[...]

PR #3368, opened 2021-01-09
More LED fixes

nwf:

See #3140 and #2936 for discussion of the ws2812 fix.  Some additional testing sure would be nice.

This is low urgency and could wait until after the next release.


HHHartmann (commented):
I'm pretty sure I have seen this diff before. Was it in the pixbuf PR originally?
Looks good as it did last time.

nwf:
I suspect it was at one point but got rebased out to keep that one as "just pixbuf" as possible.

PR #3369, opened 2021-01-09
LiquidCrystal robustness and test

nwf:

Adds an `NTest` module for LiquidCrystal.  So armed, take the opportunity to robustify communications, stripping out some dead or confusing code and more closely adhering to the manual.  Additional testing more than welcome.  This can probably wait until the next release window.

### DIFF
--- a/docs/lua-modules/liquidcrystal.md
+++ b/docs/lua-modules/liquidcrystal.md
@@ -237,6 +237,12 @@ liquidcrystal:blink(true)
 ## liquidcrystal.busy
 Get busy status of the LCD. When using GPIO backend without `rw` argument specification function does nothing.
 
+!!! note
+  At least some HD44780s and/or interfaces have been observed to count polling
+  the busy flag as grounds for incrementing their position in memory.  This is
+  mysterious, but software should restore the position after observing that the
+  busy flag is clear.
+
 #### Syntax
 `liquidcrystal.busy(self)`
 
@@ -428,6 +434,11 @@ liquidcrystal:leftToRight()
 
 ## liquidcrystal.position
 Get current position of the cursor. Position is 0 indexed. When using GPIO backend without `rw` argument specification function does nothing.
+
+!!! note
+  At least some HD44780s and/or interfaces have been observed to count reading
+  the position as grounds for incrementing their position in memory.  This is
+  mysterious, but software likely intends to restore the position anyway.
 
 #### Syntax
 `liquidcrystal.position(self)`
--- a/lua_modules/liquidcrystal/lc-i2c4bit.lua
+++ b/lua_modules/liquidcrystal/lc-i2c4bit.lua
@@ -22,27 +22,27 @@ return function(bus_args)
    -- The onus is on us to maintain the backlight state
    local backlight = true
 
-   local function send4bitI2C(value, rs_en, rw_en, read)
-      local function exchange(data, unset_read)
-         local rv = data
+   local function exchange(data, read)
+      local rv = data
+      i2c.start(busid)
+      i2c.address(busid, busad, i2c.TRANSMITTER)
+      i2c.write(busid, bit.set(data, en)) -- set data with en
+      if read then
+         i2c.start(busid) -- read 1 byte and go back to tx mode
+         i2c.address(busid, busad, i2c.RECEIVER)
+         rv = i2c.read(busid, 1):byte(1)
          i2c.start(busid)
          i2c.address(busid, busad, i2c.TRANSMITTER)
-         i2c.write(busid, bit.set(data, en))
-         if read then
-            i2c.start(busid)
-            i2c.address(busid, busad, i2c.RECEIVER)
-            rv = i2c.read(busid, 1):byte(1)
-            i2c.start(busid)
-            i2c.address(busid, busad, i2c.TRANSMITTER)
-            if unset_read then data = bit.bor(bit.bit(rs),
-                                              bit.bit(rw),
-                                              backlight and bit.bit(bl) or 0) end
-	    i2c.write(busid, bit.set(data, en))
-         end
-         i2c.write(busid, bit.clear(data, en))
-         i2c.stop(busid)
-         return rv
       end
+      i2c.write(busid, data) -- lower en
+      i2c.stop(busid)
+      return rv
+   end
+
+   local function send4bitI2C(value, rs_en, rw_en)
+      local meta = bit.bor(rs_en and bit.bit(rs) or 0,
+                          rw_en and bit.bit(rw) or 0,
+                          backlight and bit.bit(bl) or 0)
       local lo = bit.bor(bit.isset(value, 0) and bit.bit(d4) or 0,
                          bit.isset(value, 1) and bit.bit(d5) or 0,
                          bit.isset(value, 2) and bit.bit(d6) or 0,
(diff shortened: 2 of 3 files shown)


nwf:
Luacheck failures are from the shenanigans `NTest` does with `eq` and `ok`.  Suggestions?

lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
```diff
@@ -0,0 +1,116 @@
...
+  lcd:write("de")
+  lcd:cursorMove(10,2)
+  lcd:write("fg")
+  lcd:cursorMove(12,3)
+  lcd:write("hi\000")
+  lcd:cursorMove(18,4)
+  lcd:write("jk")
+end)
+
+N.test("read back 'a'", function() lcd:home()           ok(eq(0x61, lcd:read())) end)
```
HHHartmann, line 84: Tests should be self contained and not rely on the tests that run before.
If you want to print something for a test try using code like
`ok(eq(0x61, lcd:read()), "read back 'a'")`

lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
```diff
@@ -0,0 +1,116 @@
+-- Run LiquidCrystal through some basic tests
```
HHHartmann, line 1: Do we really want to place Lua-module tests beside the module and C-module tests in /tests?

docs/lua-modules/liquidcrystal.md
```diff
@@ -237,6 +237,12 @@ liquidcrystal:blink(true)
 ## liquidcrystal.busy
 Get busy status of the LCD. When using GPIO backend without `rw` argument specification function does nothing.
 
+!!! note
+  At least some HD44780s and/or interfaces have been observed to count polling
+  the busy flag as grounds for incrementing their position in memory.  This is
```
HHHartmann, line 242:
```suggestion
  At least some HD44780s and/or interfaces have been observed to increase their cursor position
  when polling the busy flag.  This is
```

Not entirely sure but I think this better expresses what is happening. I think instead of 'ground' you mean 'reason'? but again my English knowledge might not be complete here.

lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
```diff
@@ -0,0 +1,116 @@
...
+    ./lua_tests/mispec.lua \
+    ./lua_modules/liquidcrystal/liquidcrystal.lua \
+    ./lua_modules/liquidcrystal/lc-i2c4bit.lua \
+    ./lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
+
+Better to put NTest and LiquidCrystal in LFS, and then just run
+
+  TCLLIBPATH=./tests/expectnmcu expect ./tests/tap-driver.expect \
+    ./lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
+
```
HHHartmann, line 15: Maybe rather have instructions for use without expect or even better just list the required Lua files.

HHHartmann (changes requested):
Can't say anything about the changes in `lc-i2c4bit.lua` as I don't know how the displays work.
Can't test it either as I don't have a display at hand.
Maybe @matsievskiysv can have a look?

lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
nwf: I think it's cleaner, but I admit it differs from the rest of our tree.  (I think I'd rather have per-feature directories with their dependencies (e.g., `app/mqtt/lib/*`, `app/mqtt/mqtt-lua.c`) rather than one big amalgamated `app/modules` and random other directories in `app/`, with exceptions being made for things like `LwIP` that are cross-cutting.... but it's not my tree.)

If it'd make you happier, I can move this into `/tests`, probably as `NTest_liquidcrystal_i2c4bit` rather than `..._lc_...`.

docs/lua-modules/liquidcrystal.md
nwf: Your English is excellent and certainly better than my anything-not-English.  "ground" is, tragically, one of those very, very heavily overloaded terms; I was using sense 4 of https://www.merriam-webster.com/dictionary/ground , but I think you're right that it's wordy and your phrasing is better. :)

lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
nwf: Er, well, full independence is going to be tricky, but I reworked things.

docs/lua-modules/liquidcrystal.md
HHHartmann: Just out of curiosity. What is your mother tongue then since you seem to live in GB? Mine is German.

lua_modules/liquidcrystal/NTest_lc_i2c4bit.lua
HHHartmann: Its not about me being happy (well I still like to be happy) but about what makes sense. As we have a docs tree we might also have a tests tree, since it is easier to collect all the tests then. But the reason for the docs tree seems to be a technical one (not sure about that). 
From a developer perspective it would make sense to have the test with the code. At $work however we also have an extra test project sitting besides the code project.
So I think your proposal above is fine.
The filename can't get longer though. (31 chars already if I didn't miscount)

docs/lua-modules/liquidcrystal.md
nwf: Tragically I am one of those people from the USA who speak only English and, worse, I decided to study Latin for my foreign language requirements growing up. XD  Your English is *definitely* better than my German.

matsievskiysv:
I've tested the new code with [documentation example](https://nodemcu.readthedocs.io/en/release/lua-modules/liquidcrystal/#program-example). Everything works as expected.

lua_modules/liquidcrystal/lc-i2c4bit.lua
```diff
@@ -66,36 +63,45 @@ return function(bus_args)
...
       end,
       busy      = function(_)
-         local rv = send4bitI2C(0xff, false, true, true)
-         send4bitI2C(bit.bor(0x80, bit.clear(rv, 7)), false, false, false)
-         return bit.isset(rv, 7)
+         return bit.isset(send4bitI2C(0xff, false, true), 7)
       end,
       position  = function(_)
-         local rv = bit.clear(send4bitI2C(0xff, false, true, true), 7)
-         send4bitI2C(bit.bor(0x80, rv), false, false, false)
```
matsievskiysv, line 86: To be hones, I don't remember what these `send4bitI2C(...(0x80, ...` lines do. Are you sure, that they are not needed?

nwf: I think they were trying to restore the cursor position.  As I read the data sheet, which is not the clearest thing in the world, they are wrong: if the busy flag is set, no commands should be sent.

matsievskiysv: I have some fuzzy memories about screen not accepting any input unless given this command. Guess it's not needed since the example worked.

nwf: I wonder just how faithful all the HD44780 implementations floating around are.  If we find one that needs special handling -- or indeed, that the version I have is itself "special" in some way -- we can revisit.

nwf:
@matsievskiysv I'm curious, speaking of multiple implementations: if you run `:position()` twice in a row, do you see the position advance as well?

lua_modules/liquidcrystal/lc-i2c4bit.lua
matsievskiysv: No, I've tested this code on the same unit. Not sure that I want to dig deeper into this:)

PR #3370, opened 2021-01-10
Add node.sleep() to ESP32 API

### LINKED ISSUE #3179: Why is there no lightsleep() in esp32?
### Missing feature
ability to use esp32 light sleep mode is missing

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/sleep_modes.html?highlight=sleep

Only deep sleep is available.

Not sleeping at all consumes a lot of energy.

### Justification
Deep sleep can only be exited with a reboot. it is basically a shutdown with only the RTC working. Wifi connections are not maintained. This is cumbersome. Suspending code execution with simple resume would be much easier to use.

light sleep can be exited easily by gpio or uart.

### Workarounds
I do not know any workaround. maybe the freertos backend is already using light sleep when doing nothing?

Would you accept a wrapper for lightsleep in modules/node.c similar to deepsleep ?

Thank you for reading this.

tomsci:
Supporting light sleep wakeup from (non-EXT1) GPIO, UART, ULP, timer and touch

Fixes #3179

As noted in the linked issue, the esp32 branch has not until now exposed the light sleep API to Lua code, and this PR addresses that. I've tested GPIO, touch, UART and timer wakeups with a adafruit huzzah32 and verified that everything appears to work as per the documentation I've added. (I've never played with the ULP so haven't tested that, no idea if it's even possible to use it with NodeMCU). I haven't yet been able to test if my project has gained any real power saving from using the light sleep state though!

I had to add a new API `uart.wakeup()` to get UART wakeup to work, I'm not sure if I've put the code in the right place since the uart module seems to have a slightly different structure to all the others, with most functions going through a `platform_xyz()` API abstraction. I've not done that and have called `uart_set_wakeup_threshold()` directly. Let me know if I'm doing that wrong. I've also updated the docs for `gpio.wakeup()` to make it clearer it's specific to the `node.sleep()` API.

### DIFF
--- a/components/base_nodemcu/uart.c
+++ b/components/base_nodemcu/uart.c
@@ -259,6 +259,18 @@ static int uart_getconfig(lua_State* L) {
     return 4;
 }
 
+static int uart_wakeup (lua_State *L)
+{
+  uint32_t id = luaL_checkinteger(L, 1);
+  MOD_CHECK_ID(uart, id);
+  int threshold = luaL_checkinteger(L, 2);
+  esp_err_t err = uart_set_wakeup_threshold(id, threshold);
+  if (err) {
+    return luaL_error(L, "Error %d from uart_set_wakeup_threshold()", err);
+  }
+  return 0;
+}
+
 // Module function map
 LROT_BEGIN(uart)
   LROT_FUNCENTRY( setup,                      uart_setup )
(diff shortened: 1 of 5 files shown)


tomsci:
Just pushed another commit to this PR as I found an issue when using both `node.sleep()` and `node.dsleep()` - wakeup sources configured by `sleep()` could carry over into the `dsleep()` call. The fix is straightforward, to disable any existing wakeup sources on entry to `node_dsleep()` just like we do in `node_sleep()`.

jmattsson (approved):
Looks good to me while dry-reading it. I'll take your word for it working :)
And nice docs, thanks!

@marcelstoer I'll let you merge when appropriate, as I'm not really across things at the moment.

PR #3377, opened 2021-01-20
Fix ESP32 i2c timeout

### LINKED ISSUE #3011: ESP32 I2C hardware timeout due to clock stretching
### Summary
I have experienced I2C hardware timeouts due to clock stretching
I have solved the issue and thought I might share some insights with the community.

The title would spoil any attempt to write this as a high-suspense investigation report so here are the essentials.
### Situation: 
ESP-32, with the latest, standard NodeMCU firmware connected by I2C to a Bosch BNO055 IMU.
I am using the i2c module with the `i2c.SW` interface (software based bitbanging, master mode only, synchronous operation).

It's convenient for validation and testing of my I2C commands and making sure I read the BNO055 datasheet correctly :-)
With a I2C bus speed set to `i2c.SLOW` (100 kHz), the ESP32 and the BNO055 are communicating happily and reliably.

### Expectation:
The I2C bus is slow and ties down the CPU in synchronous mode.
So why not delegate the work to one of the ESP32 hardware I2C subsystems ?
This should be simple, specify `i2c.HW0/1` in `i2c.setup,` add the appropriate `i2c.transfer` call with a call back function to handle the results of the I2C operation whenever they are available.
### Issue:
It just doesn't work.
The call back function returns nonsense/random data and a false ACK flag.

Using `i2c.transfer` without a call back doesn't work better, it just fails with a Lua error and a timeout message.
That's surprising because I call `i2c.transfer` without a timeout parameter so it should default to infinite...
### Diagnostic:
The ESP32 hardware I2C interfaces have a timeout parameter and that's a good feature for reacting to I2C bus faults.
The ESP-SDK driver sets this to a default value corresponding to 8 I2C SCL clock cycle lengths.
More precisely, with a 80 Mhz ESP clock and a 100KHz I2C bus speed, this value is 8*80/0.1=6400 ESP clock cycles, i.e. 80 µS.

The BNO055 is a busy chip and can ask to slow down I2C communication by using clock stretching.
There is no documented upper bound to this so it could cause a timeout on the ESP.

You can set any I2C bus speed via `i2c.setup` and you don't have to stick to the predefined values of `i2c.SLOW` etc
A slower I2C bus speed means a higher timeout value.
In my case, an I2C bus speed of 30000 (30 Khz) does the trick and I am getting reliable I2C interaction again.
However, checking the ACK flag of the call back function is still highly recommended !
### Workaround:
Leave the firmware as it is and possibly add a cautionary note in the documentation of the i2c module.
I can do that, but read further...

#### Add a parameter to `i2c.setup` to modify the default timeout value
I find it a pity to have to use a slower I2C bus speed, just because one I2C peripheral chip may need some extra time to return results.
I would rather keep the higher bus speed and adapt the timeout value.

I have extended the `i2c.setup` function to accept a fifth, optional parameter as a (integer) timeout multiplier.
I had to reduce the bus speed by three previously to get reliable operation, so now a x4 timeout multiplier should be enough.

I can now call
`i2c.setup( i2c.HW0, pinSDA, pinSCL, i2c.SLOW, 4 )`
The timeout value is now 25 600 CPU clock cycles and I get reliable operations !

#### Adapt `i2c.setup` to return a value
By the way, `i2c.setup` does not return any value, contrary what the documentation states.
I have also adapted this function to return the I2C bus speed for the `i2c.SW` interface or the timeout value for the `i2c.HW0/1` interfaces.

#### Adapt `i2c.read` to standard I2C multibyte read practice
 By the way also, standard I2C practice for a multi byte read is to have the master (receiver) ACK'ing each byte except the last.
The ESP-SDK does this but the the glue routines from lua double-guess this with the end result of ACK'ing the last byte.
Not really according to specifications, but it still works, most likely because the I2C slave will stop sending data after seeing an bus STOP condition.

For compliance's sake, I have also adapted the `i2c.read` function to that effect.
### Conclusion
I hope this comments can help any others with similar issues.
If it is of interest to the community, I can contribute my code and the updated documentation, via, I suppose, a GitHub fork, adapt and pull request.

Let me know what you think of it.

jmdasnoy:
Fixes #3011

Fix issues as described in issue 3011
Rewrote documentation with tested real-life examples

### DIFF
--- a/docs/modules/i2c.md
+++ b/docs/modules/i2c.md
@@ -343,7 +343,7 @@ Writes send data for the master into the transmit buffer. This function returns 
 Data items can be multiple numbers, strings or lua tables.
 
 #### Syntax
-`i2c.slave.send(id, data1 [, data2[, ..., datan])`
+`i2c.slave.send(id, data1[, data2[, ..., datan]])`
 
 #### Parameters
 - `id` interface id, `i2c.HW0` or `i2c.HW1`
--- a/docs/modules/i2c.md
+++ b/docs/modules/i2c.md
@@ -2,169 +2,300 @@
 | Since  | Origin / Contributor  | Maintainer  | Source  |
 | :----- | :-------------------- | :---------- | :------ |
 | 2014-12-22 | [Zeroday](https://github.com/funshine) | [Zeroday](https://github.com/funshine) | [i2c.c](../../components/modules/i2c.c)|
+| 2020-01-17 | [jmdasnoy](https://github.com/jmdasnoy) | [jmdasnoy](https://github.com/jmdasnoy) | [i2c.c](../../components/modules/i2c.c)|
 
-This module supports different interfaces for communicating via I²C protocol. All interfaces can be assigned to arbitrary GPIOs for SCL and SDA and can be operated concurrently.
 
-- `i2c.SW` software based bitbanging, master mode only, synchronous operation
-- `i2c.HW0` ESP32 hardware port 0, master or slave mode, synchronous or asynchronous operation
-- `i2c.HW1` ESP32 hardware port 1, master or slave mode, synchronous or asynchronous operation
+This module supports 3 different interfaces for I²C communication on a ESP-32.
 
-The hardware master interfaces differ from the SW interface as the commands (start, stop, read, write) are queued up to an internal command list. Actual I²C communication is initiated afterwards using the `i2c.transfer()` function. Commands for the `i2c.SW` interface are immediately effective on the I²C bus and read data is also instantly available.
+The interface `id` can be
+- `i2c.SW` software based bitbanging, master mode only, immediate execution, synchronous operation, maximum speed 100 KHz (Standard mode)
+- `i2c.HW0` ESP32 hardware bus interface 0, master or slave mode, deferred execution, synchronous or asynchronous operation, maximum speed 1 MHz (Fast-mode Plus)
+- `i2c.HW1` ESP32 hardware bus interface 1, master or slave mode, deferred execution, synchronous or asynchronous operation, maximum speed 1 MHz (Fast-mode Plus)
 
-## i2c.address()
-Send (`SW`) or queue (`HWx`) I²C address and read/write mode for the next transfer.
+All interfaces can be used at the same time.
+Interfaces can use arbitrary GPIOs for the clock (SCL) and data (SDA) lines.
+The interfaces are not suitable for multi-master configurations.
 
-Communication stops when the slave answers with NACK to the address byte. This can be avoided with parameter `ack_check_en` on `false`.
+In master mode, I²C communication is performed by combinations of 5 basic I²C operations, provided by the functions `i2c.start()`,`i2c.stop()`,`i2c.address()`,`i2c.write()`,`i2c.read()`.
 
-#### Syntax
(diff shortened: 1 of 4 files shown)


jmdasnoy:
@nwf  thanks for your comments. Here is the PR. Best regards, jmdasnoy

docs/modules/i2c.md
```diff
@@ -2,170 +2,301 @@
...
+print( "Chip check, should succeed if chip is present and functional" )
+read_byte( bmpAddress , bmpIdRegister , check)
+```
+
+## i2c.address()
+Perform (`SW`) or enqueue (`HWx`) an I²C address operation, defining data transfer direction for the next operation (read or write).
 
 #### Syntax
-`i2c.address(id, device_addr, direction[, ack_check_en])`
+`i2c.address( id , device_addr , direction [, ack_check_en] )`
```
marcelstoer, line 135: For consistency reasons please don't change this (applies to the entire file). It's purely a matter of personal taste.

jmdasnoy: @marcelstoer Thanks for your comments. Consistency is important.
I have adapted the file accordingly.
BTW there are several style guides for  Lua. Is there any recommended one to be used here ?
Regards,
jmdasnoy

nwf:
I am unequipped to test this, but at a glance it looks good to me.

docs/modules/i2c.md
```diff
@@ -212,7 +343,7 @@ Writes send data for the master into the transmit buffer. This function returns
 Data items can be multiple numbers, strings or lua tables.
 
 #### Syntax
-`i2c.slave.send(id, data1[, data2[, ..., datan]])`
+`i2c.slave.send(id, data1 [, data2[, ..., datan])`
```
marcelstoer, line 346:
```suggestion
`i2c.slave.send(id, data1[, data2[, ..., datan]])`
```

docs/modules/i2c.md
```diff
@@ -176,7 +307,7 @@ The I²C slave mode is only available for the hardware interfaces `i2c.HW0` and
 Registers or unregisters an event callback handler.
 
 #### Syntax
-`i2c.slave.on(id, event[, cb_fn])`
+`i2c.slave.on(id, event [, cb_fn] )`
```
marcelstoer, line 310:
```suggestion
`i2c.slave.on(id, event[, cb_fn])`
```

jmdasnoy:
@nwf    @marcelstoer Thanks for finding the time to review, comment and pull in my contribution.
This is much appreciated !

I have not used the ESP32 as a I2C slave yet. When I get around to doing this, I will add a minimal working example to the documentation.

jmdasnoy:
@marcelstoer Thanks for providing the Docker container for building the firmware. I have found it very convenient to use !
Not sure if this the right place to mention it, but it has recently stopped working as documented.
The `configure-esp32` command complains about Python 2.7 deprecation but ends in error "ImportError : No module named setuptools"
My workaround is to bash into the container to run
`python -m pip --upgrade pip setuptools wheel`
After this `configure-esp32` and `build` work fine, though the upgrade is lost when I exit the container.

marcelstoer:
> Not sure if this the right place to mention it, but it has recently stopped working as documented.

https://github.com/marcelstoer/docker-nodemcu-build/ would be the right place but thanks anyway 😄

PR #3378, opened 2021-01-21
Next release

marcelstoer:
In line with the process at https://github.com/nodemcu/nodemcu-firmware/blob/release/CONTRIBUTING.md#for-collaborators


HHHartmann:
Why do we have conflicting files here?

marcelstoer:
Weird...we had a release on Sep 28, 2020 and another one on Nov 7, 2020. Yet, we're seeing commits here as far back as Sep 13, 2020 😖 ❓

nwf:
While `release` is supposed to be an ancestor of `dev`, that isn't true right now.  The last common commit between `dev` and `master` is 64bbf006898109b936fcc09478cbae9d099885a8 (i.e., tags `3.0-master 20200910` / `3.0-release 20200910`).  `release` then continues with f527ad86e4026df2de1ef1bd1f5e9ee74ccdb061 and on before getting to 4f6792773f93f36a7255cfd28dca7aa6c4aa9552 (i.e., tag `3.0-release 20201107`) while `dev` continues with c000a0894af040b5fe00d00370b030431dcc364c and has 0e88617659c0283f9ba0ed6f2ad8f90e142d7686 as its analogous commit to `3.0-release 20201107` before reaching its current head of c212b30a03280b5d14a78882ae06b7419fb26341.

The correct solution, IMHO, is to locally merge `dev` into `release` with a `theirs` merge strategy -- that is, create a new merge commit with c212b30a03280b5d14a78882ae06b7419fb26341 and 4f6792773f93f36a7255cfd28dca7aa6c4aa9552 as its parents but the content of c212b30a03280b5d14a78882ae06b7419fb26341.  Arrange that `release` and `dev` both point at this commit by performing fast-forwards, tag said comit result as `3.0-release 20210XXX` and push both branches and the new tag to github.

marcelstoer:
@nwf thanks for the analysis! Would you have time to sort this out?

PR #3382, opened 2021-01-23
Backmerge release into dev

pjsg:
This is horrible -- this is the change that you get if you back-merge release into dev. I have no idea what happened to cause these big differences. It seems that all the commits got duplicated somehow. This PR will bring them back into step.

However, I note that no actual code is changed by this merge. However, it should then clean up the merge from dev to release (it should turn into a fast-forward merge).


marcelstoer:
Well, but what is the cleanest way to resolve this? We obviously can't squash this onto `release` and creating a merge commit isn't correct either. Would we effectively change anything if we rebase and merge?

pjsg:
We could create a merge commit -- and that would get rid of the merge conflicts. Then the merge from dev back to release should be clean. 

TO validate this, I did:

```git checkout upstream/dev
git checkout -b test
git merge fix-merge-conflicts
git checkout upstream/release
git checkout -b test2
git merge test
```
And this runs clean with no conflicts. This is the process of merging this PR and then merging the dev->release PR. AT this point, release and dev are on the same commit. 

The other option is to rewind `release` back to 

`64bbf006  2020-09-04  vsky   (tag: 3.0-release_20200910, tag: 3.0-master_20200910)  Adding '+' flag to string.format` 

and then the dev -> release merge would be clean. However, this moving backwards of a branch is generally discouraged, and I don't even know how to do it in github....

nwf:
See https://github.com/nodemcu/nodemcu-firmware/pull/3378#issuecomment-770316856 for details of what's happened (though not how, as I don't know) and a proposed merge solution.

marcelstoer:
Superseded by #3393 & #3394

PR #3384, opened 2021-01-24
Doc update for bme280.lua

### LINKED ISSUE #3383: Suggestion: Add short info on how to differentiate bme280/bmp280 sensors to readthedocs
### Missing feature
Add a small paragraph to the [bme280 lua module documentation](https://nodemcu.readthedocs.io/en/release/lua-modules/bme280/) on how to differentiate between BME280 and BMP280 sensors.

The BME280 has a square shape, whereas the BMP280 has a rectangular one. [Image comparison](https://imgur.com/a/cXlS7L0)

I can make the edit and create a PR if this suggestion is accepted.

### Justification
The BMP280 sensor only supports temperature and pressure measurements.
The BME280 sensor supports temperature, pressure and humidity measurements.

The bme280 Lua module works with both of them and will happily give humidity measurements that are usually either 0 or 100 %, sometimes they are within a reasonable range (40-60) but always static.

Sellers of breakout modules sometimes confuse the two and sell one as the other. This happened to me and I spent a day debugging and trying to find out what was going on with my humidity measurements.

I suppose a check in the code that gives a warning if one tries to measure humidity with a BMP280 would work too. One possible way to find that out is to check the humidity oversampling setting/register. It can be written to in the BMP280 but won't change it's content which stays at 000. The check would unnecessarily hurt efficiency though (most of the time one will have the correct sensor) and be quite a bit of work for little payoff.

Fivefold:
Fixes #3383.

1. Changed the note about bme280_math dependency into a blue note box, like in the [bh1750 documentation](https://nodemcu.readthedocs.io/en/release/lua-modules/bh1750/)
2. Added a pagragraph on differentiating between BMP280 and BME280 sensors. It put it in an orange caution box for now, might be too visible for what's a rare problem.
3. Small fixes in bme280.setup() description

### DIFF
--- a/docs/lua-modules/bme280.md
+++ b/docs/lua-modules/bme280.md
@@ -5,11 +5,19 @@
 
 This module communicates with [BME280/BMP280 temperature/air presssure/humidity sensors](http://www.bosch-sensortec.com/bst/products/all_products/bme280) (Bosch Sensortec) through [I2C](../modules/i2c.md) interface. 
 
-Note: the module works only with the [bme280_math](../modules/bme280_math) module.
+!!! note
+
+    The module works only with the [bme280_math](../modules/bme280_math) module.
+
+!!! caution
+
+    The BMP280 only supports temperature and air pressure measurements. It will give wrong readings for humidity but no warnings or errors. Sometimes sellers of breakout boards for these sensors confuse the two and sell one as the other.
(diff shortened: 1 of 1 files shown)


marcelstoer:
Thanks for the contribution!

PR #3387, opened 2021-01-24
Add sodium.generichash() API (ESP32)

### LINKED ISSUE #3386: ESP32: Add sodium.generichash() API
### Missing feature
Expose the crypto_generichash() API as per https://doc.libsodium.org/hashing/generic_hashing to the `sodium` module that was added in https://github.com/nodemcu/nodemcu-firmware/pull/2550. I already have the code for this, PR incoming shortly.

### Justification
I found myself needing it :) As before, this is only exposing APIs already present in the native C SDK.  I've been using the code I propose to PR for several months without problems.

### Workarounds
There is already the `crypto` module which supports SHA1 and SHA2 hashes, which appears to use the Mbed TLS hashing functions, so I could have used those. It's more for consistency with the other parts of my project, which use the `sodium.crypto_box` APIs for which there is no good alternative, which drove me to want to use the Sodium hash APIs too. The Sodium APIs (based on the more modern BLAKE2b hash function) claim some performance and security improvements over older functions like SHA1/SHA2, although I have not dug much into those claims personally. The Sodium API is also a little more flexible allowing variable length hashes and convenient addition of salts to the hash (which the API calls "keys").

tomsci:
Fixes #3386.

Expose the libsodium generichash API to Lua via the `sodium` module, see #3386 for more info.

### DIFF
--- a/components/modules/crypto.c
+++ b/components/modules/crypto.c
@@ -37,7 +37,7 @@ static const algo_info_t algorithms[] = {
     { "SHA384",    64, MBEDTLS_MD_SHA384 },
     { "SHA512",    64, MBEDTLS_MD_SHA512 },
 #ifdef CONFIG_NODEMCU_CMODULE_SODIUM
-    { "BLAKE2b",     0, SODIUM_BLAKE2B }, // 0 because no fixed size for BLAKE2b hashes
+    { "BLAKE2b",   64, SODIUM_BLAKE2B }, // The hash size isn't fixed with BLAKE2b, but 64 is the maximum
 #endif
 };
 
--- a/components/modules/crypto.c
+++ b/components/modules/crypto.c
@@ -5,14 +5,18 @@
 #include "mbedtls/md.h"
 #include "module.h"
 #include "platform.h"
+#include "sodium_module.h"
 
 #define HASH_METATABLE "crypto.hasher"
+
+// Must not overlap with mbedtls_md_type_t
+#define SODIUM_BLAKE2B (-1)
 
 // algo_info_t describes a hashing algorithm and output size
 typedef struct {
     const char* name;
     const size_t size;
-    const mbedtls_md_type_t type;
+    const int type; // either a mbedtls_md_type_t or SODIUM_BLAKE2B
 } algo_info_t;
 
 // hash_context_t contains information about an ongoing hash operation
@@ -32,21 +36,21 @@ static const algo_info_t algorithms[] = {
     { "SHA256",    32, MBEDTLS_MD_SHA256 },
     { "SHA384",    64, MBEDTLS_MD_SHA384 },
     { "SHA512",    64, MBEDTLS_MD_SHA512 },
+#ifdef CONFIG_NODEMCU_CMODULE_SODIUM
+    { "BLAKE2b",     0, SODIUM_BLAKE2B }, // 0 because no fixed size for BLAKE2b hashes
+#endif
 };
 
+
+const int MAX_HASH_SIZE = 64; // Must be >= all the sizes in the above array
 
 //NUM_ALGORITHMS contains the actual number of enabled algorithms
 const int NUM_ALGORITHMS = sizeof(algorithms) / sizeof(algo_info_t);
 
-// crypto_new_hash (LUA: hasher = crypto.new_hash(algo)) allocates
-// a hashing context for the requested algorithm
-static int crypto_new_hash_or_hmac(lua_State* L, bool is_hmac) {
+static const algo_info_t* crypto_get_algo(lua_State* L, bool is_hmac)
+{
     const algo_info_t *ainfo = NULL;
     const char *algo = luaL_checkstring(L, 1);
-    const unsigned char *key = NULL;
-    size_t key_len = 0;
-    if (is_hmac)
-        key = (const unsigned char *)luaL_checklstring(L, 2, &key_len);
 
     for (int i = 0; i < NUM_ALGORITHMS; i++) {
         if (strcasecmp(algo, algorithms[i].name) == 0) {
@@ -56,7 +60,38 @@ static int crypto_new_hash_or_hmac(lua_State* L, bool is_hmac) {
     }
 
     if (ainfo == NULL) {
-        return luaL_error(L, "Unsupported algorithm: %s", algo);
+        luaL_error(L, "Unsupported algorithm: %s", algo);
+    }
+
+    if (ainfo->type == SODIUM_BLAKE2B && is_hmac) {
+        luaL_error(L, "'BLAKE2b' algorithm does not support HMAC");
+    } 
+
+    return ainfo;
+}
+
+// crypto_new_hash (LUA: hasher = crypto.new_hash(algo)) allocates
+// a hashing context for the requested algorithm
+static int crypto_new_hash_or_hmac(lua_State* L, bool is_hmac) {
+    const algo_info_t *ainfo = crypto_get_algo(L, is_hmac);
+    const unsigned char *key = NULL;
+    size_t key_len = 0;
+
+#ifdef CONFIG_NODEMCU_CMODULE_SODIUM
+    if (ainfo->type == SODIUM_BLAKE2B) {
+        lua_remove(L, 1); // algo
+        // Stack is now correct for calling l_sodium_generichash_init.
+        // Even though the underlying crypto_generichash_init() doesn't
+        // explicitly state it uses BLAKE2b (there's a separate
+        // crypto_generichash_blake2b_init API for that) in practice it is
+        // documented to do so, so I don't think it's necessary to worry about
+        // the distinction here.
+        return l_sodium_generichash_init(L);
+    }
+#endif
+
+    if (is_hmac) {
+        key = (const unsigned char *)luaL_checklstring(L, 2, &key_len);
     }
 
     // Instantiate a hasher object as a Lua userdata object
@@ -125,8 +160,8 @@ static int crypto_hash_finalize(lua_State* L) {
     // retrieve the hashing context:
     hash_context_t* phctx = (hash_context_t*)luaL_checkudata(L, 1, HASH_METATABLE);
 
-    // reserve some space to retrieve the output hash, according to the current algorithm
-    unsigned char output[phctx->ainfo->size];
+    // reserve some space to retrieve the output hash
+    unsigned char output[MAX_HASH_SIZE];
 
     int err = 0;
     // call the hash finish function to retrieve the result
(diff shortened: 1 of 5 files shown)


jmattsson:
Overall I like it, but I would love it if this was folded into the `crypto` module rather than sit off on the side. Being able to do e.g. `crypto.new_hash("blake2b", "mysalt")` would make me happy. And I could definitely see an argument for adding a convenience function of `crypto.hash("method", "data", "optional-salt")` around `crypto.new_hash("method", "optional-salt"):update("data"):finalize()`.

To support that I think it's just the `algo_info_t` that would need another member (e.g. "provider"), and then abstract the hashing context so it's not mbedtls specific.

I could also be convinced that the other two `sodium` functions should live under the `crypto` umbrella (though we'd obviously want to keep them available under the old name for some time). Simple feature toggles under the crypto module in Kconfig could enable the features as desired.

@marcelstoer @devsaurus What are your thoughts here? Am I complicating things, or keeping things simpler? :D

marcelstoer:
> Am I complicating things, or keeping things simpler?

👍 IMO it's the latter 😄 Sounds sensible to me.

> Simple feature toggles under the crypto module in Kconfig could enable the features as desired.

As the ESP32 doesn't have the scarce resources the ESP8266 has we wouldn't even need the toggles but make the features always available.

tomsci:
Sure, I can take a look at making those changes 👍

Would it be worth adding a single-call `crypto.hash(algo, data)` API (like the esp8266 branch has) to match the `sodium.generichash()`API?

I'm not sure how well the other `sodium` APIs would fit into crypto though, functions like `sodium.crypto_box.seal_open()` are pretty specific to how libsodium does things, I'm not sure attempting to map them on to generic API names in `crypto` wouldn't just make things more confusing not less - eg I could see myself thinking "I know what libsodium API I want to use, now I have to read yet more docs to figure out what `crypto` API maps on to that". We'd probably have to sit down and consider all the other API variants libsodium offers (secret boxes vs sealed boxes, etc etc) to make sure we got everything right, whereas I just exposed the subset of libsodium APIs which I needed at the time, and the others could be added later without any additional head scratching (in the sense of "how do these newly exposed APIs map onto the `crypto` APIs we've already got"?). Unless of course we wanted to go all-in on using libsodium for the esp32 `crypto` APIs that actually involve encryption? The reason I wanted to use them in the first place was the ease of use of the APIs, correctly using AES-CBC-WTF-BBQ to encrypt and sign some data is not something I'd feel confident I could get right first time :)

In short I think there's merit in retaining the separate `sodium` APIs, while at the same time I can see that making libsodium's generichash API available in the existing `crypto.new_hash()` would be handy too (with a note in the docs saying something like "see also the `sodium` module for the libsodium-style APIs").

jmattsson:
> In short I think there's merit in retaining the separate sodium APIs, while at the same time I can see that making libsodium's generichash API available in the existing crypto.new_hash() would be handy too (with a note in the docs saying something like "see also the sodium module for the libsodium-style APIs").

This sounds like a sensible compromise to me. So we're saying:
  - Include sodium's blake2b algorithm in `crypto.new_hash` *if the sodium module is enabled*
  - Ideally also `crypto.new_hmac`, but I don't recall if the current hmac setup makes that easy or not (in principle it should be easy)
  - Maybe add `crypto.hash()` for a simple all-in-one hashing small things, implemented in terms of the existing functionality
  - Keep the `sodium` module so it can be incrementally added to as/if needed

Does that sound about right? Thoughts anyone?

tomsci:
> Include sodium's blake2b algorithm in crypto.new_hash _if the sodium module is enabled_

Yep I've tested making it available when the sodium module _isn't_ enabled, and the extra code pulled in by doing so increases the ROM size by 86KB, so I think we'd definitely want to make it conditional on the module being enabled.

> Ideally also crypto.new_hmac, but I don't recall if the current hmac setup makes that easy or not (in principle it should be easy)

Looking at the libsodium code and docs, I'm not clear if blake2b (and by extension `crypto_generic_hash()`) is intended for HMACs. The equivalent (I think...) libsodium API [`crypto_auth()`](https://doc.libsodium.org/secret-key_cryptography/secret-key_authentication#purpose) uses HMAC-SHA512 internally. libsodium also uses Poly1305 for what it calls ["One-time authentication"](https://doc.libsodium.org/advanced/poly1305) but I don't think that's quite the same thing either.

That being said, the fact that libsodium's generichash API accepts an optional `key` parameter does make it _look_ a bit like an HMAC API. I haven't had enough coffee to figure that out (and possibly there isn't enough coffee in the world!). I'd feel happier not making any assumptions there and leaving the `crypto.new_hmac` APIs as-is for now.

> Maybe add crypto.hash() for a simple all-in-one hashing small things, implemented in terms of the existing functionality

👍 

> Keep the sodium module so it can be incrementally added to as/if needed

Another benefit to doing so, which I probably should've been clearer on previously, is that it makes it very easy on the other end of the communication to work out how to handle the data. You don't need to think "right, this data is a PKCS7 signature using an RSA key, what's the right API for me to decode that on this non-NodeMCU system" you merely need to think "This data was created with `sodium.crypto_box.seal()`, I need to use my local language binding for `crypto_box_seal_open()` to decode it".

tomsci:
I've taken a stab at making the discussed changes - I thought about trying to fully reconcile `hash_context_t` and `sodium_generichash_state` but since Mbed TLS uses a small struct with a heap allocation and a cleanup fn, and libsodium a larger 300 byte context struct which doesn't require explicit cleanup, I decided it wasn't a real benefit to try and merge them, and instead made `new_hash` return a different userdata type when "BLAKE2b" is requested compared to the Mbed TLS-provided algorithms. Hopefully this has also kept the changes to `crypto.c` to a manageable level.

components/modules/crypto.c
```diff
@@ -5,14 +5,18 @@
...
 
+// Must not overlap with mbedtls_md_type_t
+#define SODIUM_BLAKE2B (-1)
+
 // algo_info_t describes a hashing algorithm and output size
 typedef struct {
     const char* name;
     const size_t size;
-    const mbedtls_md_type_t type;
+    const int type; // either a mbedtls_md_type_t or SODIUM_BLAKE2B
```
jmattsson, line 19: You don't reckon it might be cleaner to have something like
```
typedef enum { CRYPTO_PROVIDER_MBEDTLS, CRYPTO_PROVIDER_SODIUM } crypto_provider_t;
typedef enum { SODIUM_ALGO_BLAKE2B } sodium_algo_t;
...
typedef struct {
  const char *name;
  const size_t size;
  const crypto_provider_t provider;
  union {
    const mbedtls_md_type_t mbedtls;
    const sodium_algo_t sodium;
  } type; // or unnamed union if the toolchain supports it
} algo_info_t;
```
It would be a bit more resistant to accidental overlap if something changes in the `mbedtls_md_type_t` values, which would appear as rather silent error (and silent errors are "teh wurst").

components/modules/crypto.c
```diff
@@ -32,21 +36,21 @@ static const algo_info_t algorithms[] = {
     { "SHA256",    32, MBEDTLS_MD_SHA256 },
     { "SHA384",    64, MBEDTLS_MD_SHA384 },
     { "SHA512",    64, MBEDTLS_MD_SHA512 },
+#ifdef CONFIG_NODEMCU_CMODULE_SODIUM
+    { "BLAKE2b",     0, SODIUM_BLAKE2B }, // 0 because no fixed size for BLAKE2b hashes
+#endif
 };
 
 
+const int MAX_HASH_SIZE = 64; // Must be >= all the sizes in the above array
```
jmattsson, line 45: No fixed size for blake2b, but 64 max? I'd say stick the 64 into the `algorithms` array entry then. Less risk of this one getting overlooked if a larger hash gets added at some later time.

components/modules/crypto.c
```diff
@@ -56,7 +60,38 @@ static int crypto_new_hash_or_hmac(lua_State* L, bool is_hmac) {
     }
 
     if (ainfo == NULL) {
-        return luaL_error(L, "Unsupported algorithm: %s", algo);
+        luaL_error(L, "Unsupported algorithm: %s", algo);
```
jmattsson, line 63: Style point - all `luaL_error()` calls should be prefixed with `return` to make it clear processing doesn't continue past this point (internally `luaL_error()` raises the exception via a `longjmp()` call, so the stack gets unwound here).

components/modules/crypto.c
```diff
@@ -56,7 +60,38 @@ static int crypto_new_hash_or_hmac(lua_State* L, bool is_hmac) {
...
+    }
+
+    if (ainfo->type == SODIUM_BLAKE2B && is_hmac) {
+        luaL_error(L, "'BLAKE2b' algorithm does not support HMAC");
```
jmattsson, line 67: `return luaL_error` please :)

jmattsson (commented):
Overall looks good!

Some comments/thoughts before I say "let's merge", as per above.

components/modules/crypto.c
tomsci: This is a now a helper function that returns `const algo_info_t*` so the usual convention can't be applied. It's a pretty small function so there isn't really much chance of confusion I don't think?

components/modules/crypto.c
tomsci: I'm in two minds as to which approach is the more likely to introduce silent errors - some subsequent code forgetting to check `provider` as well as `type`, versus the possibility that Mbed TLS in future changes how it defines `mbedtls_md_type_t`...? I picked `-1` for `SODIUM_BLAKE2B` on the basis it was pretty unlikely that `mbedtls_md_type_t` would start using that value (not to mention that it would also need to start being used by crypto.c before it'd become a problem).

In the event of a tie-break I'd be inclined to just go for the simpler solution, which doesn't introduce an additional variable and a union :)

components/modules/crypto.c
jmattsson: Ah, but if you do call `luaL_error()` in this function, then *it doesn't return*. That's been a constant source of confusion for people new to integrating with the Lua VM. The best analogy to calling `luaL_error()` is that you're effectively doing a C++ `throw`. Execution of the current function ceases, the stack gets unwound, and the error gets handled somewhere completely different up the call chain.

Not being aware of this mechanic is also an easy source of resource leakage, as a fair bit of due diligence is required to ensure any acquired resources are released prior to calling `luaL_error()`.

Because of this, the canonical way to remind everyone is to prefix all calls to `luaL_error()` with a `return`, even though the return itself never gets actioned.

components/modules/crypto.c
jmattsson: Until such a time we have more crypto libs being linked in here, I can be convinced that the simpler solution is the preferred one :)

PR #3388, opened 2021-01-25
Auto create additional partition for SPIFFS

LianYangCn:
The default partition table from `components/platform/partitions.csv`，which allocates 0x70000 byte for SPIFFS exactly.
This commit makes sure "During the first boot, the firmware creates an additional partition for SPIFFS in the remaining flash space."


jmattsson:
Oh my, I completely missed this review request, I'm so sorry!

I am not comfortable making this change the default however. Aside from the potential for reboot loops if CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED isn't set, in IDF4 we no longer have the ability to tinker with the partition table at runtime, since it's checksummed and there is no safe way of updating it on the go. I don't really want to change the default behaviour right before we're about to freeze the IDF3 branch.

Thank you for taking your time to raise the PR though, and once again my apologies for not reviewing sooner.

PR #3392, opened 2021-01-31
Add support for open networks to enduser_setup

### LINKED ISSUE #3374: enduser_setup module fails for open wifi networks
### Expected behavior
Using enduser_setup, selecting an open network (one without password) is expected to work, but results in 400 bad request error instead.

### Actual behavior
Selecting an open network result in a 400 bad request error.
This is due to the fact that the `enduser_setup.html` page will explicitly disable the password field for open networks. This results in a form that is submitted without the `wifi_password` field, which is expected to always be there in `enduser_setup.c`, in function `enduser_setup_http_handle_credentials()`, line 837:

```
if (name_str == NULL || pwd_str == NULL)
  {
    ENDUSER_SETUP_DEBUG("Password or SSID string not found");
    return 1;
  }
```

This "return 1" results in function `enduser_setup_handle_POST` replying a 400 bad request response.

All that is needed is to remove `pwd_str==NULL` from the test, and create an empty string instead when pwd_str is NULL.

### Test code
Setup a wifi network without password, then do as in the documentation:
```Lua
enduser_setup.start()
```
Select the network in the browser page. Notice that the submitted request has no `wifi_password` field. Error 400 will follow.

I don't have a working copy of nodemcu myself, so I cant submit a pull request. I hope someone can just fix this directly, as this is a simple bug.

### NodeMCU startup banner

```
	branch: release
	commit: 4f6792773f93f36a7255cfd28dca7aa6c4aa9552
	release: 3.0-release_20201107
	release DTS: 202011071523
	SSL: false
	build type: float
	LFS: 0x0 bytes total capacity
	modules: adc,bit,encoder,enduser_setup,file,gpio,i2c,mdns,mqtt,net,node,rtcmem,rtctime,sntp,tmr,uart,wifi
 build 2021-01-03 22:57 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
```

### Hardware

Esp8266 on a wemos d1 pro. Not very relevant for this bug...

blobule:
Fixes #3374.

I fixed the bug, making sure an open network (without password) is handled correctly.
I tested it and it works perfectly.

### DIFF
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -832,9 +832,14 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
   state->success = 0;
   state->lastStationStatus = 0;
 
+
   char *name_str = strstr(data, "wifi_ssid=");
   char *pwd_str = strstr(data, "wifi_password=");
-  if (name_str == NULL || pwd_str == NULL)
+
+  // in case we dont get a passwd (for open networks)
+  if (pwd_str == NULL) { pwd_str="wifi_password="; }
+
+  if (name_str == NULL)
   {
     ENDUSER_SETUP_DEBUG("Password or SSID string not found");
     return 1;
--- a/app/modules/enduser_setup.c
+++ b/app/modules/enduser_setup.c
@@ -837,11 +837,14 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
   char *pwd_str = strstr(data, "wifi_password=");
 
   // in case we dont get a passwd (for open networks)
-  if (pwd_str == NULL) { pwd_str="wifi_password="; }
+  if (pwd_str == NULL) {
+    pwd_str="wifi_password=";
+    ENDUSER_SETUP_DEBUG("No passord provided. Assuming open network");
+  }
 
   if (name_str == NULL)
   {
-    ENDUSER_SETUP_DEBUG("Password or SSID string not found");
+    ENDUSER_SETUP_DEBUG("SSID string not found");
     return 1;
   }
 
--- a/docs/modules/enduser-setup.md
+++ b/docs/modules/enduser-setup.md
@@ -54,6 +54,7 @@ Then the `eus_params.lua` file will contain the following:
 
 ```lua
 -- those wifi_* are the base parameters that are saved anyway
+-- if network is open, then there is no wifi_password
 local p = {}
 p.wifi_ssid="ssid"
 p.wifi_password="password"


blobule:
I compiled nodemcu for the first time to fix this bug, and it was a great experience. Congratulation for such a nice compiling setup.
I've been using nodemcu for years, but this is a first for me.
I hope this PR is satisfactory.

Should we add in the docs of enduser_setup a complete example of an adequate boot process (init.lua)?
Something like this...
```
print("Wating 10 seconds. Use t:stop() to cancel the init process...")
t=tmr.create()
t:alarm(10000,tmr.ALARM_SINGLE,function ()
	if wifi.sta.getip() then
		print("Wifi activated",wifi.sta.getip())
		dofile("go.lua")
	else
		print("No Wifi. Starting portal...")
		enduser_setup.start(
		  "Nodemcu setup",
		  function()
			wifi.sta.autoconnect(1)
			print("Wifi activated",wifi.sta.getip())
			dofile("go.lua")
		  end,
		  function(err, str)
		    -- schedule reboot?
		  end
		)
	end
end)
```
It would be nice to have such an example chunk for init.lua

app/modules/enduser_setup.c
```diff
@@ -832,9 +832,14 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
   state->success = 0;
   state->lastStationStatus = 0;
 
+
   char *name_str = strstr(data, "wifi_ssid=");
   char *pwd_str = strstr(data, "wifi_password=");
-  if (name_str == NULL || pwd_str == NULL)
+
+  // in case we dont get a passwd (for open networks)
+  if (pwd_str == NULL) { pwd_str="wifi_password="; }
```
HHHartmann, line 840: There should be an  ENDUSER_SETUP_DEBUG note about this.
Maybe
``` C
ENDUSER_SETUP_DEBUG("Password string not found. Setting to empty");
```

app/modules/enduser_setup.c
```diff
@@ -832,9 +832,14 @@ static int enduser_setup_http_handle_credentials(char *data, unsigned short data
...
+
+  if (name_str == NULL)
   {
     ENDUSER_SETUP_DEBUG("Password or SSID string not found");
```
HHHartmann, line 844: Please adapt the Message. "SSID string not found"

HHHartmann (changes requested):
The rest looks good.

HHHartmann:
> I compiled nodemcu for the first time to fix this bug, and it was a great experience. Congratulation for such a nice compiling setup.

Wellcome to the club!


> Should we add in the docs of enduser_setup a complete example of an adequate boot process (init.lua)?
> Something like this...
> 
> ```
> print("Wating 10 seconds. Use t:stop() to cancel the init process...")
> t=tmr.create()
> t:alarm(10000,tmr.ALARM_SINGLE,function ()
> 	if wifi.sta.getip() then
> 		print("Wifi activated",wifi.sta.getip())
> 		dofile("go.lua")
> 	else
> 		print("No Wifi. Starting portal...")
> 		enduser_setup.start(
> 		  "Nodemcu setup",
> 		  function()
> 			wifi.sta.autoconnect(1)
> 			print("Wifi activated",wifi.sta.getip())
> 			dofile("go.lua")
> 		  end,
> 		  function(err, str)
> 		    print("enduser_setup: Err #" .. err .. ": " .. str)
> 		    -- schedule reboot?
> 		  end
> 		)
> 	end
> end)
> ```
> 
> It would be nice to have such an example chunk for init.lua

The lua_examples folder would be a nice place for that. Maybe write a comment about what go.lua is (I can guess, but just to be clear)
Also set t=nil at the end
Maybe have some emphasis on waiting for Wifi too.

I think a different PR would be good for that.

marcelstoer (approved):

marcelstoer:
> The lua_examples folder would be a nice place for that....
> I think a different PR would be good for that.

Yes, and then link to it from the docs.

marcelstoer:
Oh, it's already here: #3395

marcelstoer:
@blobule could you please Gregor's comments?

HHHartmann (approved):

PR #3395, opened 2021-02-02
exemple of init.lua with wifi portal

blobule:
Fixes no specific issue,

Following a suggestion in #3374 , this is an example for a init.lua that uses a wifi portal to configure wifi if needed.
I put it inside lua_examples, but I don't know how and where to refer to this in the docs. Suggestions?

Also, I wonder if it would be interesting to provide complete examples for some common IOT setups:
- robust mqtt to light up a LED remotely
- robust mqtt sensor reading (temperature) every 10 minutes, with deep sleep in between
- robust mqtt IOT identification using [Homie](https://homieiot.github.io/)
- ...
( Here robust means that it will survive power outage, network outage, etc, and always restart properly. )
Such examples could make it much simpler to setup IOT by providing a good template.
Juste wondering if this is something that would be considered useful.

### DIFF
--- a/lua_examples/init.lua
+++ b/lua_examples/init.lua
@@ -1,0 +1,34 @@
+-- init.lua example, with wifi portal
+--
+-- Waits 10 sec to allow stopping the boot process, and get wifi activated
+-- When wifi not activated, setup a portal to select a network
+-- In all cases, wifi ends up activated and the main task main.lua is executed
+--
+-- module needed: enduser_setup
+
+print("Waiting 10sec... use t:stop() to cancel init.lua")
+t=tmr.create()
+t:alarm(10000,tmr.ALARM_SINGLE,function ()
+	t=nil -- free the timer
+	if wifi.sta.getip() then
+		print("Wifi activated",wifi.sta.getip())
+		dofile("main.lua")
+	else
+		print("No Wifi. Starting portal...")
+		enduser_setup.start(
+		  "SuperBidule",
+		  function()
+			wifi.sta.autoconnect(1)
+			print("Wifi activated",wifi.sta.getip())
+			dofile("main.lua")
+		  end,
+		  function(err, str)
+		    print("enduser_setup: Err #" .. err .. ": " .. str)
+		  end,
+		  function(str)
+		    print("Debug " .. str)
+		  end
+		)
+	end
+end)
+


pjsg:
As a stylistic point, I'd make `t` a `local` and then you don't need to set it to `nil` in order to free it.

In my code, I always start the enduser_setup -- it exits once the connection is up -- no matter whether the user configured it, or whether a previously set SSID was found.

marcelstoer:
> - [x] This PR is for the `dev` branch rather than for the `release` branch.

Now it is - I just changed the base branch.

lua_examples/init.lua
```diff
@@ -0,0 +1,34 @@
+-- init.lua example, with wifi portal
```
marcelstoer, line 1: This should IMO go into a sub directory like e.g. `lua_examples/eus/init.lua`

blobule:
To answer, @pjsg, we can't make `t` a `local` since it must be possible to type `t:stop()` during the initial 10 seconds delay. Anything local will disappear as soon as `init.lua` is done executing, so `t` must stay global. This is why `luacheck` fails, even if this is the correct code, IMO. I don' t see any way around this.

Unfortunately, `enduser_setup` will always create a wifi access points even when wifi station is already configured. The acces point will be gone in less than 10 seconds, but still I consider this less than ideal. So I prefer to check wifi status before any call to enduser_setup. Maybe `enduser_setup` should check wifi **before** setting up the access point?

I agree with @marcelstoer that this init.lua is "eus" specific. I have no problem moving it to `eus`. However, maybe we could group all examples of init.lua together... one with eus, one for waking up after deep sleep, other situations....  All I know is that when we need those examples, they are scattered around the docs, so a directory "starting up" might also work. Its up to you, I'll follow your suggeations and update the PR.

lua_examples/init.lua
blobule: I agree that this init.lua is "eus" specific. I have no problem moving it to eus. However, maybe we could group all examples of init.lua together... one with eus, one for waking up after deep sleep, other situations.... All I know is that when we need those examples, they are scattered around the docs, so a directory "starting up" might also work. Its up to you, I'll follow your suggeations and update the PR.

jmd13391:
@blobule "I don't see any way around this" -- Use a tmr.  ;-)

lua_examples/init.lua
marcelstoer: > maybe we could group all examples of init.lua together

Good idea! Do you have a naming proposal? A directory called `init_lua` and then files like e.g. `eus-init.lua`?

marcelstoer: @blobule still interested in following up on this?

HHHartmann:
You can disable checks in luacheck by adding a comment. See the luacheck documentation

PR #3396, opened 2021-02-04
DCC module update to latest NmraDcc.cpp version

### LINKED ISSUE #3025: There are two source files licensed under the GPL
I notice that there are two files `modules/am2320.c` and `driver/NmraDcc.c` that are licensed under the GPL. This is not compatible with the MIT license that the rest of the nodemcu firmware is licensed under.

I think that we either need to remove them or see if the original author will relicense them under the MIT license.

vsky279:
Updates DCC module to latest NmraDcc.cpp and fixes #3025.

This is the update of the DCC module to latest version of [https://github.com/mrrwa/NmraDcc/commits/master/NmraDcc.cpp](https://github.com/mrrwa/NmraDcc/commits/master/NmraDcc.cpp).
It preserves the original file and the port to NodeMCU Lua is done only by adding some lines. I hope this is the wise way to allow future upstream of NmraDcc. A `#define NODEMCUDCC` is used to compile NodeMCU Lua version. Any advise on the way this is done is welcome.
This should also solve all license issues (#3025) related to usage of NmraDcc.

There are two commits. The first copies the original file and the second ports it to NodeMCU Lua. I think it would be good if both commits are merged to `dev` without being squashed into one.

@pjsg Philip can you please help testing this update? At the moment I have only an Arduino based DCC controller without other DCC devices than ESP8266.

I need to test `#define NODE_DEBUG` version too.

### DIFF
--- a/app/driver/NmraDcc.c
+++ b/app/driver/NmraDcc.c
@@ -45,8 +45,59 @@
 //
 //------------------------------------------------------------------------
 
+// NodeMCU Lua port by @voborsky
+
+// #define NODE_DEBUG
+#define NODEMCUDCC
+
+#ifdef NODEMCUDCC
+  #include <stdint.h>
+  #include <stdlib.h>
+  #include <stdio.h>
+  #include "platform.h"
+  #include "user_interface.h"
+  #include "task/task.h"
+  #include "driver/NmraDcc.h"
+
+  #define BYTE_TO_BINARY_PATTERN "%c%c%c%c%c%c%c%c"
+  #define BYTE_TO_BINARY(byte)  \
+    (byte & 0x80 ? '1' : '0'), \
+    (byte & 0x40 ? '1' : '0'), \
+    (byte & 0x20 ? '1' : '0'), \
+    (byte & 0x10 ? '1' : '0'), \
+    (byte & 0x08 ? '1' : '0'), \
+    (byte & 0x04 ? '1' : '0'), \
+    (byte & 0x02 ? '1' : '0'), \
+    (byte & 0x01 ? '1' : '0')
+   
+   #define byte uint8_t
+   #define word int16_t
+
+   #define abs(a) ((a) > 0 ? (a) : (0-a))
+   
+   #define RISING GPIO_PIN_INTR_POSEDGE
+   #define FALLING GPIO_PIN_INTR_NEGEDGE
+   #define CHANGE GPIO_PIN_INTR_ANYEDGE
+   
+    static uint32_t last_time_overflow_millis;
+    static uint32_t last_system_time;
+
+    uint32_t millis() {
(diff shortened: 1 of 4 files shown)


marcelstoer:
> There are two commits. The first copies the original file and the second ports it to NodeMCU Lua. I think it would be good if both commits are merged to `dev` without being squashed into one.

That's a sensible approach, thanks.

pjsg:
This is going to be more difficult for me to test as I had made a bunch of bug fixes to the old code which it appears that I never did a PR for.... I'm going to try and rebase my stuff on top of this and see what happens....

marcelstoer:
As a FYI I am also tagging @kiwi64ajs.

vsky279:
> This is going to be more difficult for me to test as I had made a bunch of bug fixes to the old code which it appears that I never did a PR for.... I'm going to try and rebase my stuff on top of this and see what happens....

Oh, that's a pitty. Do you have it somewhere on your repository?

pjsg:
I have a new_dcc branch in my repo -- which does at least compile (it has the DCC module enabled). The couple of things that I noticed when merging in (most of) my changes was:

You are missing the interrupt disable around the copying of the DCC msg. Also, there is no reason to ifdef out all the ack processing -- in particular, my branch (tries to) support the generation of the basic ack. My problem with it is that I can't seem to get the current draw to be stable enough for the controller to detect the ack pulse when I generate it.

vsky279:
I think but I'm not sure that I have removed the ack processing because I didn't have the hw to test it so I did not bother with it. I will reintroduce it. Than I will check your repository.

vsky279:
@pjsg Philipp, I have merged your changes. 
I like the `CV_table` parameter.  I'm using table with metamethods to store CVs but did not think about having this feature directly in the module. I need to implement it in my project.

I think that at some stage we should squash this commit and the "NmraDcc port to NodeMCU Lua" commit as the first removes act functions and the second puts them back. There will be a single commit to porting NmraDcc to NodeMCU Lua so that future updates will be easier.

Enabling `#NODE_DEBUG` compiles fine but causes restarts.

pjsg:
I hope to be able to do some serious testing today.

vsky279:
I manage to test the PR with Marklin Trix 66955. Everything seems to work fine.
I think this is ready to be marged.

I remind that the first two commits should not be squashed as it will help to update NmraDcc to newer versions (2nd and 3rd are fine to be squashed).

marcelstoer:
> I remind that the first two commits should not be squashed as it will help to update NmraDcc to newer versions (2nd and 3rd are fine to be squashed).

👍  What GitHub offers us is all or nothing while merging. Can you please squash them on your end?

vsky279:
@marcelstoer and it's done :-)

PR #3399, opened 2021-02-12
Document node.setonerror()

### LINKED ISSUE #3381: Document node.setonerror()
### Expected behavior
Documentation should contain node.setonerror()

### Actual behavior
Not contained

### Test code
See https://nodemcu.readthedocs.io/en/dev/search.html?q=setonerror

See https://gist.github.com/TerryE/646407100785fb00c2cbddbb0e8a2b61#panic-handling for more Information

marcelstoer:
Fixes #3381.

### DIFF
--- a/docs/modules/node.md
+++ b/docs/modules/node.md
@@ -462,7 +462,6 @@ node.restore()
 node.restart() -- ensure the restored settings take effect
 ```
 
-
 ## node.setcpufreq()
 
 Change the working CPU Frequency.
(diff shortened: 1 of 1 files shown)


nwf:
Probably worth _strongly encouraging_ people to ensure that the callback ends with a restart.  Something has gone quite wrong and it's probably not safe to just wait for the next event (e.g., timer tick) and hope everything works out.

marcelstoer:
Very valid point, thanks! I amended the description and the snippet.

docs/modules/node.md
```diff
@@ -481,6 +480,32 @@ target CPU frequency (number)
 node.setcpufreq(node.CPU80MHZ)
 ```
 
+## node.setonerror()
+
+Use this to override the default "always restart" action if wanted; for example to write an error to a logfile or to a network syslog before restarting.
```
HHHartmann, line 485:
```suggestion
Overrides the default crash handling which always restarts the system; Can be used to write an error to a logfile or to secure connected hardware before restarting.
```

Should be in the style of the rest of the docs

nwf (approved):

HHHartmann:
I wonder if we should document how the default reset works with its 2 seconds wait before reboot.
Entirely not sure if this is rhe right place though.

marcelstoer:
Rather document too much than too little. Better to have it in the wrong place than nowhere at all.

If you propose a few sentences and could also add them to the FAQ and link to it from this function.

marcelstoer:
Gregor, do you still want to amend this or can we merge as-is?

PR #3400, opened 2021-02-12
Add note about GPL

marcelstoer:
Fixes the 2nd part of #3025 and is thus a companion PR to #3396.


nwf (approved):
Seems right to me.  (Tho': IANAL, IANYL, TINLA, HTH, HAND)

PR #3413, opened 2021-03-25
Add node.bootreason

marcelstoer:
Fixes no reported issue

Dug out from here https://www.esp8266.com/viewtopic.php?f=23&t=22314&p=90917#p90162 and integrated as-is.

### DIFF
--- a/docs/modules/node.md
+++ b/docs/modules/node.md
@@ -26,9 +26,33 @@ The second value returned is the extended reset cause. Values are:
   - 5, wake from deep sleep
   - 6, external reset
 
+Or the internal reset code of the system can be returned as extended reset cause:
+
+  - 7: NO_MEAN                 no known reason
+  - 8: POWERON_RESET           power on reset, this includes reset button connected to power-on reset
+  - 9:
+  - 10 SW_RESET                software reset digital core caused by esp-idk firmware
+  - 11: OWDT_RESET             legacy watch dog reset digital core
+  - 12: DEEPSLEEP_RESET        Deep Sleep reset digital core
+  - 13: SDIO_RESET             Reset by SLC module, reset digital core
+  - 14: TG0WDT_SYS_RESET       Timer Group0 Watch dog reset digital core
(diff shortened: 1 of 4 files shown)


marcelstoer:
@jmattsson @nwf any objections to this?

nwf:
I've no objection, but also no experience with NodeMCU on ESP32.

jmattsson:
Looks useful to me. The only potential quibble is the reuse of the PANIC code to include Lua panics as well as kernel panics (and the use of an internal API function which might change/disappear), but I think the pros likely outweigh the cons.

docs/modules/node.md
```diff
@@ -26,10 +26,34 @@ The second value returned is the extended reset cause. Values are:
...
+  - 21: EXT_CPU_RESET          for APP CPU, reseted by PRO CPU
+  - 22: RTCWDT_BROWN_OUT_RESET Reset when the vdd voltage is not stable
+  - 23: RTCWDT_RTC_RESET       RTC Watch dog reset digital core and rtc module
+
+
 In general, the extended reset cause supercedes the raw code. The raw code is kept for backwards compatibility only. For new applications it is highly recommended to use the extended reset cause instead.
 
 In case of extended reset cause 3 (exception reset), additional values are returned containing the crash information. These are, in order, EXCCAUSE, EPC1, EPC2, EPC3, EXCVADDR, and DEPC.
 
+In case of extended reset cause 19 (SW_CPU_RESET), an additional value is returned containing the number of consecutive lua panics. If the reset was caused by a call to node.restart() this value is 0; after the first panic the value is 1; if a panic reoccurs the value increments upto 15.
```
marcelstoer, line 54:
```suggestion
In case of extended reset cause 19 (SW_CPU_RESET), an additional value is returned containing the number of consecutive Lua panics. If the reset was caused by a call to node.restart() this value is 0; after the first panic the value is 1; if a panic reoccurs the value increments upto 15.
```

PR #3414, opened 2021-03-28
NFC: tweak lua53 byte extraction

nwf:
This is a first, very small, very tentative step in divorcing our Lua runtime from the ESP8266 / Xtensa architecture.  In particular, rather than using hard-coded offsets when defining byte extractor functions, we rely on the C compiler to compute offsets for us.  I've tried to make this something that might plausibly go upstream, too (and have begun work on the other direction of "meet in the middle", making a patch set against an upstream Lua 5.3, with the goal of reaching parity with `app/lua53`).

This PR should be completely cosmetic, generating identical assembler (but different DWARF, naturally).

- [x] This PR is for the `dev` branch rather than for the `release` branch.
- [x] This PR is compliant with the [other contributing guidelines](https://github.com/nodemcu/nodemcu-firmware/blob/dev/CONTRIBUTING.md) as well (if not, please describe why).
- [x] I have thoroughly tested my contribution.
- [x] The code changes are reflected in the documentation at `docs/*`.

PR #3415, opened 2021-03-29
app/lua53: catch up to lua 5.3.6

nwf:
Diff reduction to 5.3.6, mostly guided by upstream.

Tested lightly using a subset of the expect framework; enough to verify that LFS continues to function, at least!

### DIFF
--- a/app/lua53/host/liolib.c
+++ b/app/lua53/host/liolib.c
@@ -277,6 +277,8 @@ static int io_popen (lua_State *L) {
   const char *filename = luaL_checkstring(L, 1);
   const char *mode = luaL_optstring(L, 2, "r");
   LStream *p = newprefile(L);
+  luaL_argcheck(L, ((mode[0] == 'r' || mode[0] == 'w') && mode[1] == '\0'),
+                   2, "invalid mode");
   p->f = l_popen(L, filename, mode);
   p->closef = &io_pclose;
   return (p->f == NULL) ? luaL_fileresult(L, 0, filename) : 1;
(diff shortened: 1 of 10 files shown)


nwf:
Nobody here or #3414?

HHHartmann (commented):
I am a bit concerned about the luaC_objbarrier calls, as it seems they belong to GC and we have some modifications in that area.
The rest I guess is just from upstream and I guess they know what they are doing, so that's OK.

@nwf what exactly do you mean by not checking the "I tested" checkbox. Did you run the code on ESP?
Did you run the Lua test suite?

Do we need to test GC in combination with our RO elements for this?
Maybe @TerryE can help out.

nwf:
When I said "mostly guided by upstream", I meant that most of this was https://www.lua.org/work/diffs-lua-5.3.5-lua-5.3.6.html .  The `objbarrier`s are from that, FWIW.

nwf:
Ah *ha*.  After a day of debugging I have finally caught my mistake in this.  Of course it was in the `lundump.c` bit where we've diverged from upstream.  `LoadFunction` does not always (but does sometimes!) return its second argument, so it's important to re-assign, as in `p[i] = LoadFunction(S, p[i], f->source);`.

This now passes my `expect` harness's tests, modulo some local network shenanigans.  But LFS got plenty of exercise throughout the run. :)

nwf:
Will merge after #3479 lands.

marcelstoer:
> Will merge after https://github.com/nodemcu/nodemcu-firmware/pull/3479 lands.

Oh, that was a year ago 😄 Will merge now.

PR #3417, opened 2021-04-05
softuart2

sjp27:

- [x ] I have thoroughly tested my contribution.

\<Softuart2 uses hardware timer for timing, does not burn processor\>


HHHartmann:
The failed check means that you have to add the *.md file in the mkdocs.yml file. The check script could get some polish in terms of user friendliness I guess.

marcelstoer:
Why not rework the existing softuart driver rather than adding a new implementation _alongside_ the existing one?

sjp27:
My implementation is very different, did not want to cause issues for those already using the existing softuart. My solution does not do timing by burning CPU, it uses a timer, it supports full duplex, but only up to 4800 baud.

marcelstoer:
I am really not keen on having two competing modules for the same functionality in the source tree. According to the documentation your implementation is largely API-compatible with the existing one. So, in what way would your implementation cause any regressions (apart from potential bugs)?
@galjonsfigur what's your take on this?

galjonsfigur:
I looked into the code and while the idea of using Timer for TX part is definitely better than just timing using delays, it works only for low bauds. This code is essentially polling the GPIO every timer interrupt and checking the status - both for RX and TX part which don't work well for software UART, but allows full-duplex.. I don't think it would be a good idea to merge this as it is - better idea would be to use this approach in the original module, but only bellow the certain baudrate - 4800 or 2400. The way of using interrupt and reading the full frame at once works only when the frame time is short - if the ESP stays in ISRs too long, many things start to fail. 
As for TX part - it would be a good idea to use a task or timer + buffer to send each full frame for high baudrates because timing between them won't matter as long as the frames themselves are correct, instead of the current approach in original `softuart` module but then sending data becomes asynchronous and details like overflowing TX buffer have to be accounted for. Again, I think it would be a good idea to to the same as for RX part - use your code variant but only bellow certain baudrate.
And the use of Timer exclusively can be a problem for some applications - many modules use it and most of them do so using non-exclusive way. I can imagine that shared timer can introduce too much jitter to even be able to receive low baudrates.

sjp27:
I see softuart2 complementing softuart in the same way pmw2 complements pmw. Because of its different implementation it has advantages/disadvantages over the original implementation. Main advantages are - it is impacted and impacts other functionality less as it does not burn cpu, it also supports full duplex allowing simple loop back tests,  it also provides a 16X baud rate clock which is useful for infra red comms. Main disadvantages are - it does not support high baud rates and needs exclusive use of timer 1.

PR #3421, opened 2021-04-09
NodeRedux state container for ESP devices

### LINKED ISSUE #3420: Redux state container for ESP8226, ESP32
### Missing Feature: NodeRedux - state container
Redux is a predictable state container for JavaScript apps. So we can use the same concept to manage the state of ESP devices. It helps you write applications that behave consistently.

### Justification
The whole global state of your app is stored in an object tree inside a single _store_. The only way to change the state tree is to create an _action_, an object describing what happened, and _dispatch_ it to the store. To specify how the state gets updated in response to an action, you write pure _reducer_ unctions that calculate a new state based on the old state and the action.

### Workarounds
I have translated the logic of Redux in JS to Lua. You can check [here](https://github.com/shubham-sri/node-redux)

shubham-sri:
Fixes #3420


nwf:
Thank you for your contribution.

This feels very... heavy?  Enterprise-y?  for an embedded system.  Just picking on one example, constructing the first error message in `assertReducerShape`, even with an empty key `k` will consume 295 bytes of RAM, credibly between .5% and 1% of the application's initial available heap!

The more idiomatic approach in Lua is to use its OOish facilities, specifically, table lookup (`.` or the `:` sugar) and function call as message send.  While you could do that inside your store's `reducer` function to avoid the `if`/`elseif`/`else` chains, it's no less flexible or compositional to instead pass the store itself as a table to the would-be message sources and have them send messages by invoking particular fields.  The use of metatables even affords opportunity for "unknown message" handling, should that be useful.  If isolation is desired (though it is often less of a concern with the size of applications that fit in ESPs), the store methods can close over the mutable state rather than holding it in the store table itself (though this is not without its costs: the store methods become closures rather than static function references).

shubham-sri:
> reducer

Hey @nwf, thank you for the nice suggestion 

The `reducer` function is created by the end-user so for more flexibility I have to add `if`/`elseif`/`else` chains, and this same is used in [Redux](https://redux.js.org/).

And `assertReducerShape` can be optimized, I'm looking to solve it in an efficient way.

PR #3437, opened 2021-05-31
Dev esp32 add eth.set_ip() method

LianYangCn:
Add eth.set_ip() method
It can set static IP address, netmask, gateway, DNS address of the ethernet.

### DIFF
--- a/components/base_nodemcu/uart.c
+++ b/components/base_nodemcu/uart.c
@@ -259,6 +259,26 @@ static int uart_getconfig(lua_State* L) {
     return 4;
 }
 
+static int uart_wakeup (lua_State *L)
+{
+  uint32_t id = luaL_checkinteger(L, 1);
+  MOD_CHECK_ID(uart, id);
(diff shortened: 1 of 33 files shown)


jmattsson (approved):
Apologies for taking ages to review. Thanks for your contribution! I had some comments on gateway vs gw and whether the DNS setting should be part of general IP settings, but then I realised this PR matches what we already do in `wifi.sta.set_ip()`, so this is probably the best thing to do. 👍

marcelstoer:
@jmattsson Thanks for the review. I was holding back because we agreed for the `dev-esp32` branch to be locked to avoid merge conflicts down the road. Have you considered merging this into `dev-esp32-idf4` instead?

jmattsson:
I've got an equivalent almost ready to push onto the idf4 branch (since the tcpip_adapter was replaced with esp_netif in idf4). Happy to merge this onto the idf3 branch in the meantime.

PR #3457, opened 2021-08-15
Update RTD MkDocs

### LINKED ISSUE #3454: Migrate RTD/MkDocs build to Python 3
I wanted to enable the RTD docs build for @jmattsson's  `dev-esp32-idf4` branch so we can announce that branch in the docs early. The build failed with "Could not find a version that satisfies the requirement ziglang>=0.8.0". Checking [the RTD logs](https://readthedocs.org/api/v2/build/14397085.txt) made me realize we're still using Python 2 to build the docs.

Upgrading to Python 3 (in the RTD settings) worked but [the build failed again](https://readthedocs.org/api/v2/build/14397146.txt) but for a different reason. Maybe it's because the [MkDocs version](https://www.mkdocs.org/about/release-notes/#version-0173-2018-03-07) we (or RTD?) use is ancient as well.

marcelstoer:
Fixes #3454 for the IDF4 branch.

This will have to be carried over to the `dev` branch afterwards (to eventually end up on `release` as well). IMO we should skip doing this for `dev-esp32` as it'll eventually be replaced by the IDF4 branch.

@jmattsson if https://nodemcu.readthedocs.io/en/update-rtd-mkdocs/ looks fine to you then this should be good enough to land.


jmattsson:
Looks good to me from where I'm sitting!

jmattsson:
Should I delete the `update-rtd-mkdocs` branch, or do you want it around a while longer @marcelstoer ?

marcelstoer:
Thanks for asking, I removed it. I updated the RTD settings so we now have https://nodemcu.readthedocs.io/en/dev-esp32-idf4/.

jmattsson:
Thanks Marcel!

Issue #1995, opened 2017-05-30
onewire used to be more solid


joysfera:
### background information
I have a product (a thermometer) with 1-wire network of DS18B20 sensors, based on NodeMCU firmware that was originally built on 2016-05-28 from the dev branch. It worked well for several customers. 

In February I updated the firmware in my thermometer to version built on 2017-02-02 from the master branch. Suddenly the 1-wire communication of one customer's setup started to misbehave. The higher temperatures there were the more communication errors appeared. Somewhere above 80 ℃ the commucation broke completely = onewire search didn't find any connected sensors.

I admit he's got a large network of sensors (10+) with complicated wiring but his rather logical argument was that "it worked before you flashed new firmware".

### commit
It took me lot of time and effort (shipping devices back and forth with the customer, with various changed Lua files first and then NodeMCU firmware versions) to trace it down to the commit 04b86b80f6abac3d6e0bcc6ee99c4795407dd07e that changes the onewire.c implementation in two crucial things: driving bus high and timing.

Unfortunately I don't know which of these two changes break the network in the customer's case but I'd bet it's the driving bus high and not the timing (judging from the temperature based misbehavior).

### diff
The two onewire.c changes are (if I read the git diff correctly) as follows:

Originally (=before 2016-06-28) writing '1' to 1-wire bus was implemented by ```DIRECT_WRITE_HIGH``` that pulled the pin high by CPU actively. Nowadays (=since 2016-06-28) the CPU just leaves the pin floating (using ```DIRECT_MODE_INPUT```). I believe this is too weak even though I use pull-up resistor of 1800 ohms. I think the original strong pull-up by CPU made the 1-wire communication much more solid and stable in the case of large networks with complicated wiring.

### not sure
As I mentioned above the timing of 1-wire reading/writing has been changed as well by several microseconds. I don't think this is source of the problem mentioned above but I am not 100% sure.
The customer is not excited to test any more firmware updates so I cannot rule the timing change out completely but I believe it's OK. 

Naturally the only way to be sure is to bring an oscilloscope and analyze the bus errors in real-time but it's not easily possible so I thought I'd just document it here that the onewire is less stable than it used to be.

### proposal
My proposal is to return the strong CPU pull-up to onewire_write_bit, at least optionally (similarly to the power flag).

devsaurus:
Thanks for the heads-up!

Some items are unclear to me:
[...]

PR #3461, opened 2021-09-02
ow:  add alarm scans and timing tweaking 

stromnet:
- [x] I have thoroughly tested my contribution

This PR has two features:
1. Adding support for 1-wire ALARM scan, which only returns devices in ALARM condition. Basically just using a different command during search.
2. Add support to tweak the bit-banging timings.

While working with a AVR based 1-wire slave device (based on the https://github.com/stromnet/moat-owslave project), I had some timeouts & CRC issues when communicating with it from the NodeMCU, that I never had when using a PC-based LinkUSB interface with owfs. After some debugging it was revealed that the LinkUSB had much longer delays between each bit (or if it was byte? Don't remember now). Since my custom slave was a bit too slow to react in some cases, this caused issues. By adding `ow.timings` method, it is now possible to adjust these timings a bit, if necessary.

Note: I have tested these commits on top of 3.0-release_20201107 and not really dev, since I did not have time to verify the changes in newer releases at the moment. But nothing have happened in this module since that release, so should work equally fine on dev.

### DIFF
--- a/docs/modules/ow.md
+++ b/docs/modules/ow.md
@@ -118,11 +118,11 @@ Clears the search state so that it will start from the beginning again.
 Looks for the next device.
 
 #### Syntax
-`ow.search(pin, alarm_search)`
+`ow.search(pin, [alarm_search])`
 
 #### Parameters
 - `pin` 1~12, I/O index
-- `alarm_search` 1 / 0, if 0 a regular 0xF0 search is performed (default), if non-zero a 0xEC ALARM SEARCH is performed.
+- `alarm_search` 1 / 0, if 0 a regular 0xF0 search is performed (default if parameter is absent), if 1 a 0xEC ALARM SEARCH is performed.
 
 #### Returns
 `rom_code` string with length of 8 upon success. It contains the rom code of slave device. Returns `nil` if search was unsuccessful.
--- a/docs/modules/ow.md
+++ b/docs/modules/ow.md
@@ -268,13 +268,13 @@ Writes multi bytes. If `power` is 1 then the wire is held high at the end for pa
 #### See also
 [ow.depower()](#owdepower)
 
-## ow.timings()
+## ow.set_timings()
 Tweak different bit timing parameters. Some slow custom devices might not work perfectly well with NodeMCU as 1-wire master. Since NodeMCU ow module is bit-banging the 1-wire protocol, it is possible to adjust the timings a bit.
 
 Note that you can break the protocol totally if you tweak some numbers too much. This should never be needed with normal devices.
 
 #### Syntax
-`ow.timings(reset_tx, reset_wait, reset_rx, w1_low, w1_high, w0_low, w0_high, r_low, r_wait, r_delay)`
+`ow.set_timings(reset_tx, reset_wait, reset_rx, w1_low, w1_high, w0_low, w0_high, r_low, r_wait, r_delay)`
 
 #### Parameters
 Each parameter specifies number of microseconds to delay at different stages in the 1-wire bit-banging process.
@@ -297,7 +297,7 @@ A nil value will leave the value unmodified.
 #### Example
 ```lua
 -- Give 300uS longer read/write slots for slow MCU based 1-wire slave
-ow.timings(
+ow.set_timings(
 	nil, -- reset_tx
 	nil, -- reset_wait
 	nil, -- reset_rx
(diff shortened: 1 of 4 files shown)


marcelstoer:
Looks like a useful addition, thanks.

docs/modules/ow.md
```diff
@@ -264,5 +265,49 @@ Writes multi bytes. If `power` is 1 then the wire is held high at the end for pa
 #### Returns
 `nil`
 
-####See also
+#### See also
 [ow.depower()](#owdepower)
+
+## ow.timings()
```
HHHartmann, line 271: Mabe call it settimings and I am not sure if we need a getter.

docs/modules/ow.md
```diff
@@ -118,10 +118,11 @@ Clears the search state so that it will start from the beginning again.
 Looks for the next device.
 
 #### Syntax
-`ow.search(pin)`
+`ow.search(pin, alarm_search)`
```
HHHartmann, line 121: Please make it clear that alarm_search is optional.

HHHartmann (changes requested):
Looks good to me but I can't test it

docs/modules/ow.md
stromnet: Got any good examples? Or just mark as `(optional)` in Parameters section?

marcelstoer: Surround it with `[]`, see e.g. https://nodemcu.readthedocs.io/en/latest/modules/node/#nodeinfo.

stromnet: fixed!

docs/modules/ow.md
stromnet: timings or settimings, either way works. Don't see the use for a getter though, most likely this is a call you make only once during program lifetime.

HHHartmann: A getter would show the initial settings, but as they are documented that seems fine. 
I really  would like a getter be named as that though.

HHHartmann: I mean that it should be named settimings.

marcelstoer: > it should be named settimings

I generally agree. Unfortunately we have no cohesive naming strategy across our code base (in the early days nobody paid attention to that). Hence, we find snake_case, camelCase and nocase style in different modules. The least we can do is trying to be consistent _within_ a module. That would require the function to be called `set_timings` IMO.

HHHartmann: Agreed

stromnet: ow.set_timings it is, pushed!

marcelstoer: Thanks

marcelstoer (approved):

HHHartmann (approved):

stromnet:
Thanks!

TerryE:
@stromnet Johan, thanks doe this contribution.  I use OW in some of my systems -- and with up to a dozen DS18B20s on each.  This has been rock solid for four years now, and I am keen that these timing adjustments haven't broken OW for me.  I am abroad and away from my systems for the next 3 weeks.  I will review these changes and test this out, then raise an issue if needed.

TerryE:
@joysfera Petr, you might be interested in this as well.  I know from #1995 that you were still working on OW + DS18B20s a year ago.

joysfera:
> @joysfera Petr, you might be interested in this as well. I know from #1995 that you were still working on OW + DS18B20s a year ago.

@TerryE thank you for the heads up but it's been two years since I rewrote my whole project (2600 LOC) from Lua to C. The final straw that forced me to do give farewell to Lua after five years was the Lua source level incompatibility between ESP8266 and ESP32. In C (Arduino core) it is 100% "portable" and I can develop and run single source on both ESP (now possibily on four - including S2 and C3). My OneWire issues got resolved along that way as well.

stromnet:
Always good with more testers, but I'm quite confident you will not notice any issues. The defaults are the same values as before, so the only potential impact I foresee would be the additional cycles to fetch the values from RAM instead of using constants. But when we're talking busylooping several (or even tens of) microseconds, that microscopic difference is most likely not an issue. Not sure about how many cycles a typical instruction is on the ESP8266, but my guess is that we're talking nanoseconds.

I have tested it with DS18B20's as well, although only a few on the same bus.

PR #3463, opened 2021-09-07
Bump to Ubuntu 20.04

marcelstoer:
GH is shutting down 16.04 on September 20th: https://github.com/actions/virtual-environments/issues/3287


marcelstoer:
Thanks Gregor, I had the fix ready but forgot to push it.

PR #3467, opened 2021-09-24
Minor Lua fixes.

jmattsson:
- [x] I have thoroughly tested my contribution*.

Issues discovered over on the dev-esp32-idf4 branch while getting the 5.1+5.3 support across to that branch.

- Off by one error in loadLFS, leading to slight memory leak and
  potential corruption.

- Insufficient return value check in loadLFS, where uzlib may return
  one of two success conditions, one of which would result in an
  out-of-bounds access and related pain.

- One case of a side effect within a lua_assert(), leading to
  silently broken LFS image handling when compiling without asserts
  enabled, the issue showing up as module names being shuffled around.

- Incorrect encoding of TValues in LFS when 64bit numbers in use.

_*) predominantly tested on ESP32_

### DIFF
--- a/app/lua/lflash.c
+++ b/app/lua/lflash.c
@@ -473,7 +473,7 @@ static int loadLFS (lua_State *L) {
   vfs_lseek(in->fd, 0, VFS_SEEK_SET);
 
   /* Allocate the out buffers */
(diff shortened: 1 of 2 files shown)


marcelstoer:
Merge to give it some exposure...

PR #3468, opened 2021-09-24
Lua tidy-up to reduce delta against esp32 branch.

jmattsson:

These should hopefully be completely uncontentious changes, as there
are no functional changes in this commit.

There is still a delta against the esp32 branch, but to harmonise that fully
requires a bit more work.

This commit includes:

- Some LUA_USE_xxx #ifdef changes, since in many ways the ESP32 is
  closer to a host build than an ESP8266 build.

- A bunch of compiler warnings tidy-ups.

- A couple of readability/maintainability improvements (e.g. LROT_END
  definition using named member initialisation, prototype declarations
  moved to header file(s)).

- Some 64bit correctness for 64bit host builds.

### DIFF
--- a/app/lua/lbaselib.c
+++ b/app/lua/lbaselib.c
@@ -24,7 +24,7 @@
 ** model but changing `fputs' to put the strings at a proper place
 ** (a console window or a log file, for instance).
 */
-#ifdef LUA_CROSS_COMPILER
+#if defined(LUA_USE_ESP8266)
 #undef puts
 #define puts(s) printf("%s",s)
 #endif
--- a/app/lua/lnodemcu.c
+++ b/app/lua/lnodemcu.c
@@ -118,6 +118,7 @@ LUALIB_API int luaL_posttask( lua_State* L, int prio ) {          // [-1, +0, -]
 }
 #else
 LUALIB_API int luaL_posttask( lua_State* L, int prio ) { 
+  (void)L; (void)prio;
   return 0;
 } /* Dummy stub on host */
 #endif
--- a/app/lua/lnodemcu.h
+++ b/app/lua/lnodemcu.h
@@ -45,9 +45,11 @@
   static ROTable_entry LOCK_IN_SECTION(s) rt ## _entries[] = {
 #define LROT_END(rt,mt,f)    {NULL, LRO_NILVAL} }; \
   const ROTable rt ## _ROTable = { \
-    (GCObject *)1, LUA_TROTABLE, LROT_MARKED, \
-    cast(lu_byte, ~(f)), (sizeof(rt ## _entries)/sizeof(ROTable_entry)) - 1, \
-    cast(Table *, mt), cast(ROTable_entry *, rt ## _entries) };
+    .next = (GCObject *)1, .tt = LUA_TROTABLE, .marked = LROT_MARKED, \
+    .flags = cast(lu_byte, ~(f)), \
+    .lsizenode = (sizeof(rt ## _entries)/sizeof(ROTable_entry)) - 1, \
+    .metatable = cast(Table *, mt), \
+    .entry = cast(ROTable_entry *, rt ## _entries) };
 #define LROT_BREAK(rt)       };
 
 #define LROT_MASK(m)         cast(lu_byte, 1<<TM_ ## m)
@@ -65,10 +67,9 @@
 #define LROT_MASK_NEWINDEX   LROT_MASK(NEWINDEX)
 #define LROT_MASK_GC_INDEX   (LROT_MASK_GC | LROT_MASK_INDEX)
 
-/* Maximum length of a rotable name and of a string key*/
+#define LUA_MAX_ROTABLE_NAME 32  /* Maximum length of a rotable name and of a string key*/
 
 #ifdef LUA_CORE
 
 #endif
 #endif
-
--- a/app/lua/lobject.c
+++ b/app/lua/lobject.c
@@ -68,10 +68,7 @@ int luaO_log2 (unsigned int x) {
   while (x >= 256) { l += 8; x >>= 8; }
   return l + log_2[x];
 #else
- /* Use Normalization Shift Amount Unsigned:  0x1=>31 up to 0xffffffff =>0
-  * See Xtensa Instruction Set Architecture (ISA) Refman  P 462 */
-  asm volatile ("nsau %0, %1;" :"=r"(x) : "r"(x));
-  return 31 - x;
+  return 31 - __builtin_clz(x);
 #endif
 }
 
--- a/app/lua/lobject.h
+++ b/app/lua/lobject.h
@@ -28,7 +28,7 @@
 #define LUA_TUPVAL	(LAST_TAG+2)
 #define LUA_TDEADKEY	(LAST_TAG+3)
 
-#ifdef LUA_USE_ESP
+#ifdef LUA_USE_ESP8266
 /*
 ** force aligned access to critical fields in Flash-based structures
 ** wo is the offset of aligned word in bytes 0,4,8,..
--- a/app/lua53/lnodemcu.c
+++ b/app/lua53/lnodemcu.c
@@ -296,7 +296,7 @@ LUA_API void lua_getlfsconfig (lua_State *L, int *config) {
   }
 }
 
-LUA_API int  (lua_pushlfsindex) (lua_State *L) {
+LUA_API int lua_pushlfsindex(lua_State *L) {
   lua_lock(L);
   setobj2n(L, L->top, &G(L)->LFStable);
   api_incr_top(L);
@@ -309,7 +309,7 @@ LUA_API int  (lua_pushlfsindex) (lua_State *L) {
  * one upvalue that must be the set to the _ENV variable when its closure is
  * created, and as such this parallels some ldo.c processing.
  */
-LUA_API int  (lua_pushlfsfunc) (lua_State *L) {
+LUA_API int lua_pushlfsfunc(lua_State *L) {
   lua_lock(L);
   const TValue *t = &G(L)->LFStable;
   if (ttisstring(L->top-1) && ttistable(t)) {
@@ -341,7 +341,7 @@ LUA_API int  (lua_pushlfsfunc) (lua_State *L) {
 /*
  * Return an array of functions in LFS
  */
-LUALIB_API int  (luaL_pushlfsmodules) (lua_State *L) {
+LUALIB_API int luaL_pushlfsmodules(lua_State *L) {
   int i = 1;
   if (lua_pushlfsindex(L) == LUA_TNIL)
     return 0;                                 /* return nil if LFS not loaded */
@@ -357,7 +357,7 @@ LUALIB_API int  (luaL_pushlfsmodules) (lua_State *L) {
   return 1;
 }
 
-LUALIB_API int  (luaL_pushlfsdts) (lua_State *L) {
+LUALIB_API int luaL_pushlfsdts(lua_State *L) {
   int config[5];
   lua_getlfsconfig(L, config);
   lua_pushinteger(L, config[4]);
@@ -427,7 +427,9 @@ static const char *readF (lua_State *L, void *ud, size_t *size) {
 
 static void eraseLFS(LFSflashState *F) {
   lu_int32 i;
+#ifdef LUA_USE_ESP
   printf("\nErasing LFS from flash addr 0x%06x", F->addrPhys);
+#endif
   unlockFlashWrite();
   for (i = 0; i < F->size; i += FLASH_PAGE_SIZE) {
     size_t *f = cast(size_t *, F->addr + i/sizeof(*f));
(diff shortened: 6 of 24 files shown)


jmattsson:
MSVC strikes again. I'll push a fix when I have a some uninterrupted dev time.

app/lua53/lutf8lib.c
```diff
@@ -258,6 +258,7 @@ LROT_END(utf8, LROT_TABLEREF(utf8_meta), 0)
 
 
 LUAMOD_API int luaopen_utf8 (lua_State *L) {
+  (void)L;
```
TerryE, line 261: There's actually an `UNUSED(v)` macro in `llimits.h`

app/lua53/lua.h
```diff
@@ -465,6 +465,8 @@ struct lua_Debug {
   struct CallInfo *i_ci;  /* active function */
 };
 
+int lua_main(void);
```
TerryE, line 468: I hummed and harred about where to export `lua_main()`. This is roughly equivalent to `main()` in std `lua.c`.  `lua.h` is the public interface to the Lua API, and this main entry is _not_ part of the API and in general if and module writer were to attempt to call this, it would be a disaster.  This is why I left the entry _hidden_ , and declared is `extern` in the file that requires it.

app/lua53/lstate.c
```diff
@@ -42,12 +42,15 @@
 ** created; the seed is used to randomize hashes.
 */
 #if !defined(luai_makeseed)
-#if defined(LUA_USE_ESP)
+#if defined(LUA_USE_ESP8266)
```
TerryE, line 45: Isn't the `ccount` register also available on the ESP32? If so, then this should be `LUA_USE_ESP`

app/lua53/lobject.h
```diff
@@ -80,7 +80,7 @@
 ** wo is the offset of aligned word in bytes 0,4,8,..
 ** bo is the field within the word in bits 0..31
 */
-#ifdef LUA_USE_ESP
+#if defined(LUA_USE_ESP8266)
```
TerryE, line 83: Another one that also applies to the ESP32?

app/lua53/lobject.c
```diff
@@ -84,6 +78,8 @@ int luaO_ceillog2 (unsigned int x) {
   x--;
   while (x >= 256) { l += 8; x >>= 8; }
   return l + log_2[x];
+#else
+  return (x == 1) ? 0 : 32 - __builtin_clz(x - 1);
```
TerryE, line 82: Is this builtin also available for the ESP8266?  If so then we should be using it rather than a more cludgy `asm` construct.  Where can I get a list of builtins that apply to the ESP32 and ESP8266 compiler?

app/lua/lbaselib.c
```diff
@@ -24,7 +24,7 @@
 ** model but changing `fputs' to put the strings at a proper place
 ** (a console window or a log file, for instance).
 */
-#ifdef LUA_CROSS_COMPILER
+#if defined(LUA_USE_ESP8266)
```
TerryE, line 27: Isn't this a host macro?

app/lua/lnodemcu.c
```diff
@@ -118,6 +118,7 @@ LUALIB_API int luaL_posttask( lua_State* L, int prio ) {          // [-1, +0, -]
 }
 #else
 LUALIB_API int luaL_posttask( lua_State* L, int prio ) { 
+  (void)L; (void)prio;
```
TerryE, line 121: Again use `UNUSED` macro.  Actually can I make this a blanket comment relating to all `(void)var` uses in order to suppress compile warnings?  I'll drop following explicit mark-ups

app/lua/lnodemcu.h
```diff
@@ -45,9 +45,11 @@
   static ROTable_entry LOCK_IN_SECTION(s) rt ## _entries[] = {
 #define LROT_END(rt,mt,f)    {NULL, LRO_NILVAL} }; \
   const ROTable rt ## _ROTable = { \
-    (GCObject *)1, LUA_TROTABLE, LROT_MARKED, \
-    cast(lu_byte, ~(f)), (sizeof(rt ## _entries)/sizeof(ROTable_entry)) - 1, \
-    cast(Table *, mt), cast(ROTable_entry *, rt ## _entries) };
+    .next = (GCObject *)1, .tt = LUA_TROTABLE, .marked = LROT_MARKED, \
```
TerryE, line 48: This is quite a change to the ROTable layout.  The current code uses exploit the fact that a common subset of `Table` and `ROTable` fields are effectively a superset union, so the code doesn't need messy conditional selectors.  This change breaks this assumption.   I need to think through the implications.

app/lua/lobject.c
```diff
@@ -68,10 +68,7 @@ int luaO_log2 (unsigned int x) {
   while (x >= 256) { l += 8; x >>= 8; }
   return l + log_2[x];
 #else
- /* Use Normalization Shift Amount Unsigned:  0x1=>31 up to 0xffffffff =>0
-  * See Xtensa Instruction Set Architecture (ISA) Refman  P 462 */
-  asm volatile ("nsau %0, %1;" :"=r"(x) : "r"(x));
-  return 31 - x;
+  return 31 - __builtin_clz(x);
```
TerryE, line 71: Same point about use of builtins vs `asm()` as below.

app/lua/lobject.h
```diff
@@ -28,7 +28,7 @@
 #define LUA_TUPVAL	(LAST_TAG+2)
 #define LUA_TDEADKEY	(LAST_TAG+3)
 
-#ifdef LUA_USE_ESP
+#ifdef LUA_USE_ESP8266
```
TerryE, line 31: Does the ESP32 support unaligned data fetches from IRAM and IFLASH?.  If so then how many extra clocks?  The access macros add one extra register instruction to use an aligned read.  If this is still faster then why have this difference between the ESP8266 and ESP32 versions?

app/lua53/lnodemcu.c
```diff
@@ -309,7 +309,7 @@ LUA_API int  (lua_pushlfsindex) (lua_State *L) {
  * one upvalue that must be the set to the _ENV variable when its closure is
  * created, and as such this parallels some ldo.c processing.
  */
-LUA_API int  (lua_pushlfsfunc) (lua_State *L) {
+LUA_API int lua_pushlfsfunc(lua_State *L) {
```
TerryE, line 312: The standard pattern is to add the parentheses where a define macro might be encoded to force sensible compile errors here.  Why change this?

Ditto on following cases.

app/lua53/lnodemcu.c
```diff
@@ -427,7 +427,9 @@ static const char *readF (lua_State *L, void *ud, size_t *size) {
 
 static void eraseLFS(LFSflashState *F) {
   lu_int32 i;
+#ifdef LUA_USE_ESP
```
TerryE, line 430: Surely, the use of bare `printf()` statements for ESP code is generally deprecated.  We should use the appropriate equivalent.  Whatever we need a clear coding standard here.

app/lua53/lnodemcu.c
```diff
@@ -436,11 +438,13 @@ static void eraseLFS(LFSflashState *F) {
 #ifdef LUA_USE_ESP
     if (*f == ~0 && !memcmp(f, f + 1, FLASH_PAGE_SIZE - sizeof(*f)))
       continue;
+    printf(".");
```
TerryE, line 441: Same as above re use of print. The code skips the erase where it is not need and this makes flashing of smaller LFS images a lot faster.   The dots are a pretty useful visual progress bar

TerryE (changes requested):
Johny, just a quit read through. I leave it to you to wildcard the use of the `UNUSED()` macro.  I need to understand the extra meta fields on the ROTable declarations, or are this just a keyed recoding of the positional initialiser?

app/lua53/lutf8lib.c
jmattsson: Ah, I hadn't noticed that. Happy to swap over to that both here and on the esp32 branch then. Stay tuned... :)

app/lua53/lua.h
jmattsson: Hmm, on the other hand, letting others declare things `extern` elsewhere is a recipe for subtle breakage if the actual prototype _does_ change (been there, done that 😞). Maybe it should simply be dumped into its own header file that gets included by lua.c to ensure the declaration and definition match?

app/lua53/lstate.c
jmattsson: `ccount` is an Xtensa thing. The ESP32-C3 (and upcoming ESP32-H2) is a RISC-V core.

app/lua/lobject.c
jmattsson: Yeah on Xtensa `__builtin_clz()` translates to a `nsau` instruction.

app/lua/lobject.h
jmattsson: Again, this is predominantly an Xtensa vs RISC-V issue here. But yes, data memory regions are [byte accessible](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/memory-types.html). Some optimisation may be possible, but has not been a priority for me at this point - getting NodeMCU to run on all the different ESP32 SoCs has been :)

app/lua53/lnodemcu.c
jmattsson: The standard pattern is to only parenthesise function names only when `typedef`'ing function pointer types. Why would these be different? If someone `#define`s a function name to something else, surely that is on them?

app/lua53/lnodemcu.c
jmattsson: The `printf()`s aren't mine. I merely added the `#ifdef`s because having host-side LFS image generation tell me it's erasing flash was misleading at best.
(Okay, technically I believe I _moved_ one of the printfs to reduce the number of ifdefs I needed)

jmattsson:
> I need to understand the extra meta fields on the ROTable declarations, or are this just a keyed recoding of the positional initialiser?

That is purely a change from array-style initialisation (`{ 1, 2, 3 }`) to named member initialisation (`{ .x = 1, .y = 2, .z = 3}`). I had a hard time working out what got set to what when I was chasing something through this code, and the ROTable struct being composed through two levels of indirect `#define`s was doing my head in. Aside from being a bit more readable, longer term the named member initialisation is a bit safer in the face of changing structures too (also been there, done that, have the scars...).

app/lua53/lstate.c
TerryE: Yup but wouldn't we be introducing a new `LUA_USE_RISCV` category or at least review this hierarchy of USE source tailoring macros?

app/lua53/lobject.h
TerryE: As previous comment. Perhaps we need to do that review sooner rather than later.

app/lua53/lnodemcu.c
TerryE: I just feel that we should follow the standard Lua coding standards where practical

app/lua53/lnodemcu.c
TerryE: Thinking about this, we should move towards POSIX compliant source wherever practical.   The real issue here is that printing should be deprecated, except for very specific circumstances, as it will be unusual for a human or logging system to be 'listening' to the serial output, mainly development debugging. This .... progress bar is really a development tool.

marcelstoer:
Johny, any chance to revive this old PR? Reducing deltas between our branches would certainly be desirable I guess.

jmattsson:
Honestly, I think the ship has sailed at this point Marcel. Both personally and at $work we have moved on from the venerable 8266. It's been a couple of years since I last actively touched an 8266 and even then it was only in maintenance mode. Even when I prepared this PR I wasn't working with the 8266 any longer. Between that, my limited time availability, and Terry not having (had?) time/interest to skill up on the ESP32 ecosystem I don't think this will happen. We could leave the PR sitting around in the hope that changes, or we could close it. I'm happy to leave the source branch there should it ever become relevant again.

marcelstoer:
> I think the ship has sailed at this point Marcel [..] We could leave the PR sitting around in the hope that changes, or we could close it.

I fully agree and I don't mind closing this one. I asked because I see a third option: merging what's already there.

app/lua/lbaselib.c
HHHartmann: Yes it is

HHHartmann:
Especially not sure about Terry's comment in app/lua/lnodemcu.h 
So would rather not like to see at least that part in the esp8266 branch

jmattsson:
@HHHartmann I had a re-read of that comment and it does not make sense to me. There is no re-ordering, it is merely a change to using [designated initalization](https://gcc.gnu.org/onlinedocs/gcc-4.4.1/gcc/Designated-Inits.html#Designated-Inits) to avoid bugs creeping in _if_ there is reordering of actual structure _declaration_ at some point.

PR #3473, opened 2021-10-24
Add BLE GAP/GATT support to the dev-esp32 branch

### LINKED ISSUE #3469: Provide simple Bluetooth GATT/GAP interface
### Missing feature
It would be nice to have a simple way to build small bluetooth gadgets that supported GAP/GATT.

### Justification
It would be useful

### Workarounds
None.

pjsg:
Fixes #3469 

This is the first pass at BLE support. There is some more tidyup to do, and also enabling it is slightly tricky as it is built on top
of the NimBLE stack -- and I can't find a way to force enable that stack from the Kconfig. 

You can startup and shutdown the BLE stack multiple times and it doesn't obviously leak memory.

It seems to work in my test environment.

Interested in comments.


components/modules/Kconfig
```diff
@@ -14,6 +14,14 @@ menu "NodeMCU modules"
             Includes the bit module. This module provide bit manipulation
             functions on Lua numbers.
 
+    config NODEMCU_CMODULE_BLE
+        bool "BlueTooth GAP/GATT interface module"
```
marcelstoer, line 18:
```suggestion
        bool "Bluetooth GAP/GATT interface module"
```
The official term sees this as a single word.

components/modules/Kconfig
```diff
@@ -14,6 +14,14 @@ menu "NodeMCU modules"
...
+        default "n"
+	select NODEMCU_CMODULE_STRUCT
+        select BT_ENABLED
+        help
+            Includes the simple BlueTooth GAP/GATT module.
```
marcelstoer, line 23:
```suggestion
            Includes the simple Bluetooth GAP/GATT module.
```

components/modules/Kconfig
```diff
@@ -14,6 +14,14 @@ menu "NodeMCU modules"
...
+
     config NODEMCU_CMODULE_BTHCI
         bool "BlueTooth HCI interface module"
```
marcelstoer, line 26:
```suggestion
        bool "Bluetooth HCI interface module"
```

docs/modules/ble.md
```diff
@@ -0,0 +1,132 @@
+# BT HCI Module
```
marcelstoer, line 1:
```suggestion
# Bluetooth GAP/GATT Module
```

docs/modules/ble.md
```diff
@@ -0,0 +1,132 @@
...
+| Since  | Origin / Contributor  | Maintainer  | Source  |
+| :----- | :-------------------- | :---------- | :------ |
+| 2021-10-10 | [pjsg](https://github.com/pjsg) | [pjsg](https://github.com/pjsg) | [ble.c](../../components/modules/ble.c)|
+
+The BLE module provides a simple interface to allow implementation of a simple GAP/GATT server.
+This allows you to build simple gadgets that can be interrogated and controlled over BLE.
+
+## ble.init(configuration)
+
+This initializes the BlueTooth stack and starts advertising according to the data in the
```
marcelstoer, line 11:
```suggestion
This initializes the Bluetooth stack and starts advertising according to the data in the
```

docs/modules/ble.md
```diff
@@ -0,0 +1,132 @@
...
+`nil`
+
+#### Example
+```lua
+ble.advertise("foo")
+```
+
+## ble.shutdown()
+
+Shuts down the BlueTooth controller and returns it to the state where another `init` ought to work (but currently doesn't). And, at the moment, shutting
```
marcelstoer, line 60:
```suggestion
Shuts down the Bluetooth controller and returns it to the state where another `init` ought to work (but currently doesn't). And, at the moment, shutting
```

marcelstoer (dismissed):

pjsg:
I'd like to get this merged -- any more input?

marcelstoer:
As I can't assess C code I added Johny as a reviewer.

components/modules/Kconfig
```diff
@@ -14,8 +14,16 @@ menu "NodeMCU modules"
             Includes the bit module. This module provide bit manipulation
             functions on Lua numbers.
 
+    config NODEMCU_CMODULE_BLE
+        bool "Bluetooth GAP/GATT interface module"
+        default "n"
+	select NODEMCU_CMODULE_STRUCT
+        select BT_ENABLED
+        help
+            Includes the simple Bluetooth GAP/GATT module. You must enable the Nimble BLE module as well.
```
jmattsson, line 23: Can we not `select WHATEVER_NIMBLE_SYMBOL` here, rather than mention it in the help text?

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "sdkconfig.h"
+#ifdef CONFIG_NODEMCU_CMODULE_BLE
```
jmattsson, line 21: Do we need this ifdef? We won't get the full benefit of the CI build if things aren't built by default.

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+      luaL_unref(L, LUA_REGISTRYINDEX, (int) chrs->arg);
+    }
+  }
+
+  free(tofree);
+}
+
+static int
+lble_access_cb(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt *ctxt, void *arg) {
+  // Actually the only thing we care about is the arg and the ctxt
```
jmattsson: Can is suggest using `UNUSED(conn_handle); UNUSED(attr_handle);` here then?

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+  task_block->ctxt = ctxt;
+  task_block->arg = arg;
+  task_block->seqno = seqno++;
+  task_block->buffer = (char *) (task_block + 1);
+
+  if (ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR) {
+    task_block->length = OS_MBUF_PKTLEN(ctxt->om);
+    uint16_t outlen;
+    if (ble_hs_mbuf_to_flat(ctxt->om, task_block->buffer, task_block->length, &outlen)) {
+      return BLE_ATT_ERR_UNLIKELY;
```
jmattsson, line 238: Don't we leak the `task_block` if we end up in this return?

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+
+  if (!task_post(TASK_PRIORITY_HIGH, task_handle, (task_param_t) task_block)) {
+    free(task_block);
+    return BLE_ATT_ERR_UNLIKELY;
+  }
+
+  response_message_t message;
+
+  while (1) {
+    if (xQueueReceive(response_queue, &message, (TickType_t) (10000/portTICK_PERIOD_MS) ) != pdPASS) {
```
jmattsson: I'm not across Bluetooth very much, but is it safe to have such a long timeout here, in what I presume to be the Bluetooth thread?

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+      }
+    } else {
+      lua_pop(L, 1);  // Throw away the null write pointer
+    }
+    lua_pop(L, 1);	// THrow away the value
+    message.errcode = 0;
+  }
+
+cleanup:
+  lua_pop(L, 2);
```
jmattsson: I'm going to take your word for there always being 2 items on the stack here 😅
Personally I would've gone a gettop/settop to cover the just-in-case scenarios, but that' just me. Oh, and there's a case typo in the previous comment - `THrow` :)

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+  }
+  int ns = lua_rawlen(L, -1);
+
+  // -1 is the services list
+  int nc = 0;
+  for (int i = 1; i <= ns; i++) {
+    MODLOG_DFLT(INFO, "Counting -- service %d (top %d)\n", i, lua_gettop(L));
+    lua_geti(L, -1, i);
+    // -1 is now the service which should be a table. It must have a uuid
+    if (lua_type(L, -1) != LUA_TTABLE) {
```
jmattsson, line 449: `luaL_checktable()` would be a bit more canonical, no?

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+static int
+gatt_svr_init(lua_State *L) {
+    int rc;
+
+    // Now we have to build the gatt_svr_svcs data structure
+
+
+    struct ble_gatt_svc_def *svcs = NULL;
+    lble_build_gatt_svcs(L, &svcs, &notify_handles);
+    //free_gatt_svcs(L, gatt_svr_svcs);
```
jmattsson: Keep? [Yes/No]
:)

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+                desc->conn_itvl, desc->conn_latency,
+                desc->supervision_timeout,
+                desc->sec_state.encrypted,
+                desc->sec_state.authenticated,
+                desc->sec_state.bonded);
+}
+
+static int
+lble_sys_init(lua_State *L) {
+  int rc = nvs_flash_init();
```
jmattsson: Hang on, won't this interfere with our existing NVS setup over in `user_main.c`? If anything, I'd say this logic should be moved over there if we need this sort of recovery.

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+
+  luaL_argcheck(L, handle > 0 && handle <= notify_handles[0], 1, "handle out of range");
+
+  ble_gatts_chr_updated(notify_handles[handle]);
+
+/*
+  if (rc) {
+    return luaL_error(L, "Must supply a valid handle");
+  }
+*/
```
jmattsson: Looks like cruft?

components/modules/ble.c
```diff
@@ -0,0 +1,1062 @@
...
+  } else {
+    synced = READY_TO_ADVERTISE;
+  }
+
+  return 0;
+}
+
+static int lble_shutdown(lua_State *L) {
+  // It seems that shutting down the stack corrupts some critical data structures
+  // so, for now, don't allow it.
```
jmattsson: Yikes!

components/modules/ledc.c
```diff
@@ -30,6 +30,8 @@ static int lledc_new_channel( lua_State *L )
 
   ledc_timer.timer_num = opt_checkint_range(L, "timer", -1, 0, LEDC_TIMER_MAX-1);
 
+  ledc_timer.clk_cfg = LEDC_AUTO_CLK;
```
jmattsson: This one looks like a duck out of water here :)

install.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
```
jmattsson: Could we leave this in please? If there's an error things scroll past and one might very well not notice that the install failed, and then one spends ages chasing ghosts. Speaking from personal experience.

docs/modules/ble.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Characteristic table
+
+The characteristic table contains the following keys:
+
+- `uuid` The UUID of the characteristics. This can be either a 16 byte string or a 2 byte string that identifies the particular characteristic. Typically, 2 byte strings are used for well-known characteristics.
+- `type` This is the optional type of the value. It has the same value as a unpack code in the `struct` module.
+- `value` This is the actual value of the characteristic. This will be a string of bytes unless a `type` value is set.
+- `read` This is a function that will be invoked to read the value (and so does not need the `value` entry). It should return a string of bytes (unless `type` is set).
+- `write` This is a function that will be invoked to write the value (and so does not need the `value` entry). It is given a string of bytes (unless `type` is set)
```
jmattsson, line 120: So if `type` has been set, the argument to write is a number? Maybe clarify this bit slightly, please?

jmattsson (commented):
I'm not a BT expert, but on the whole this looks good!
A few comments, and a couple of fixes I'd like to see before merging, but don't wait for me to merge once you've addressed that. Let's get this landed :)

(sorry about the delay, as usual...)

components/modules/Kconfig
pjsg: Unfortunately not (at least with the version of Kconfig that Espressif uses). The BT_NIMBLE_ENABLED is a *choice* of BT_HOST -- and you can't select a choice. This is a typical example of somebody complaining about it: https://github.com/zephyrproject-rtos/zephyr/issues/6948

Since we can't modify the Espressif supplied IDF, I think we have to select it by hand. It does appear that some versions of Kconfig use the last default setting for a choice (rather than the first). Maybe I can get espressif to adopt that patch.

components/modules/ble.c
pjsg: Unfortunately yes as it includes the NIMBLE headers files and those aren't available unless configured. [Yes, this is all nasty]

components/modules/ble.c
pjsg: I'm going to reduce this to 2 seconds....

components/modules/ble.c
pjsg: Yes. I will fix.

components/modules/ble.c
pjsg: Except that this is a nested structure of tables and I wanted to indicate *which* table was missing.

components/modules/ble.c
pjsg: This is a bit worrying. I wouldn't have thought that I would have added that code without needing to -- I'll do some testing without it.

docs/modules/ble.md
pjsg: I added another paragraph below.

components/modules/ledc.c
```diff
@@ -30,8 +30,6 @@ static int lledc_new_channel( lua_State *L )
 
   ledc_timer.timer_num = opt_checkint_range(L, "timer", -1, 0, LEDC_TIMER_MAX-1);
 
-  ledc_timer.clk_cfg = LEDC_AUTO_CLK;
```
jmattsson: This looks like something that may have been added by accident to this PR?

components/modules/ble.c
jmattsson: I wonder if this is still the case on IDF5?

pjsg: This is fixed on idf5 and I have removed those comments

jmattsson:
Nice, is there anything else outstanding other than sorting out the merge issues?

docs/modules/ble.md
```diff
@@ -0,0 +1,151 @@
...
+function write_attribute(t, val)
+  -- Just store the written value in the table.
+  t.value = val
+end
+```
+
+
+### Type conversions
+
+If the `type` value converts a single item, then that will be the value that is placed into the `value` element. If it converts multiple elements, then the elements will be placed into an array that that will be placed into the `value` element.
```
HHHartmann, line 151: 'that' is repeated twice

Issue #3040, opened 2020-02-16
Our LwIP is way out of date


nwf:
Our LwIP appears to correspond to upstream's effcb90fdf29e6475614c84d327bcf2722f854b6,
a commit made in December of **2010**, but with some local patches added.  We should:

* Figure out what changes we have made to LwIP that lack upstream analogues,
  if any, and move to have those upstream.
* Move to mainline LwIP
* Remove `espconn_secure` in favor of LwIP's upstream `altcp_tls`.

rvalle:
@nwf did you look into this? I think #2931 could be related. We see sine HTTP requests failing and we can trace it down to different use of TCP segments by different browsers.

nwf:
Unfortunately, I have looked into it and LwIP straddles the boundary between the open and closed aspects of the SDK.  https://github.com/pfalcon/esp-open-sdk/issues/263 provides a reasonable, if depressing, overview of the situation.

https://github.com/d-a-v/esp82xx-nonos-linklayer looks like a sensible shim between the proprietary SDK bits and a more modern LwIP, but I have not had time to look into it much beyond theoretical discussion.

Espressif themselves have done a more modern LwIP, https://github.com/espressif/esp-lwip, but only as part of the IDF/RTOS world, so not us.  This might be further indication that living out on the NONOS branch is not a long-term survival strategy for NodeMCU, but I think we have not wanted to make the jump due to the RTOS increasing heap usage (by how much I cannot say; if someone has a link to more information handy, that'd be great).

pjsg:
To me, the https://github.com/d-a-v/esp82xx-nonos-linklayer repo looks like the right code to start with. I especially like that they have managed to get ipv6 working -- and while I don't expect nodemcu to support it as a result of this migration, it then would at least be feasible -- it is mostly the lua interfaces that would need to be changed.

TerryE:
I endorce @nwf and @jsqg working on this but I won't have much bandwidth to support this until I have this next tranche of LFS changes in and working.  BTW, as I noted elsewhere I've got some yard-works going on at the moment and keeping an eye on this is soaking up a lot of my free time for this coming week.

PR #3476, opened 2021-11-18
Fix IGMP timer

stromnet:

When sending traffic from a non-local multicast sender, via an multicast router, it was only working sporadically. For some reason the router was dropping the subscriptions every now and then, and thus stopped forwarding the multicast traffic.
I suspect this could affect L2-devices doing IGMP-snooping too.

The IGMP Queries had a max-response-timeout of 100, but in packet logs this was clearly not respected by the device. Sometimes minutes passed between reports. After a lot of digging and printfs (oh btw, `os_printf` does not work. had to hack around to redefine it to `dbg_printf`), I found root cause: `LWIP_RAND()` aka `r_rand()` is defined as `int` (well, in  app/include/lwip/igmp.h it's actually not defined at all, in sdk/esp_iot_sdk_v3.0-e4434aa/third_party/include/lwip/igmp.h it IS defined. Not sure which one is used?).

Anyway, signed means negative. and `% 100` of negative value, casted to `u16_t` is potentially > 100. Experiment here: https://godbolt.org/z/vzWGEqzzx

This fix ensures that we actually cap timer to `max_time` and thus reply within the routers expected interval.

Another fix is to ensure we define LWIP_RAND() as (u32), like defined here:
https://www.nongnu.org/lwip/2_0_x/group__compiler__abstraction.html#ga77370c377781ee7489e30eaf772ea05a
but.. again, not sure about all lwip files, what goes where and so on..

Upstream has a lot of other fixes, but not sure lwip and nodemcu lifecycles integrate..
https://git.savannah.nongnu.org/cgit/lwip.git/tree/src/core/ipv4/igmp.c#n692

### DIFF
--- a/app/lwip/core/ipv4/igmp.c
+++ b/app/lwip/core/ipv4/igmp.c
@@ -731,7 +731,7 @@ igmp_start_timer(struct igmp_group *group, u8_t max_time)
 if(max_time == 1)
   group->timer = 1;
 else
-  group->timer = (LWIP_RAND() % (max_time - 1)) + 1;
+  group->timer = (u16_t) ((unsigned int)LWIP_RAND() % (max_time - 1)) + 1;
 }
 
 /**


marcelstoer:
Thanks for the fix.

> Upstream has a lot of other fixes, but not sure lwip and nodemcu lifecycles integrate..

Pops up every now and then, e.g. here #3040.

PR #3481, opened 2021-12-21
Consider RMT channel limitations at allocation time.

### LINKED ISSUE #3480: ws2812.write always fails on ESP32-C3
### Expected behavior

The write function should emit the requested sequence on one of the pins.

### Actual behavior

On the ESP32-C3 the function always fails with "set gpio for RMT driver failed".
Looks like the function ```platform_rmt_allocate``` does not distinguish between RX and TX channels. This is important for the ESP32-C3 as it can only do TX on RMT channels 0 and 1, Channels 2 and 3 can only do RX. Currently channel 3 is allocated for a TX session, which is rejected by the SDK.

"The RMT has four channels numbered from zero to three. The first half (i.e. Channel 0 ~ 1) channels can only be configured for transmitting, and the other half (i.e. Channel 2 ~ 3) channels can only be configured for receiving."
https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/rmt.html

Oh, and I have some patch for this. Unfortunately there seems to be more trouble with RMT on the ESP32-C3 as only the reset sequence is visible after changing it. But its a start...

### Test code

```Lua
> ws2812.write({pin=8, data=string.char(255,0,0)})
E (264122) rmt: rmt_set_gpio(529): RMT CHANNEL ERR
E (264122) rmt: rmt_config(685): set gpio for RMT driver failed
E (264122) rmt: rmt_wait_tx_done(1158): RMT CHANNEL ERR
E (264122) rmt: rmt_driver_uninstall(943): No RMT driver for this channel
Lua error: 	stdin:1: sending failed
stack traceback:
	[C]: in function 'write'
	stdin:1: in main chunk
	[C]: ?
	[C]: ?
```
### NodeMCU startup banner

NodeMCU ESP32 build unspecified powered by Lua 5.1.4 [5.1-doublefp] on IDF v4.4-dev-2594-ga20df743f1

### Hardware
ESP32-C3

docbacardi:
The ESP32-C3 is limited to TX on channel 0-1 and RX on channel 2-3.

Fixes #3480 .

### DIFF
--- a/components/platform/include/platform_rmt.h
+++ b/components/platform/include/platform_rmt.h
@@ -14,14 +14,14 @@
  * @brief Allocate an RMT channel.
  *
  * @param num_mem Number of memory blocks.
- * @param is_tx Mode of the channel, !=0 allocates a TX channel, ==0 an RX channel.
+ * @param mode Mode of the channel, RMT_MODE_TX allocates a TX channel, RMT_MODE_RX an RX channel.
  *
  * @return
  *     - Channel number when successful
  *     - -1 if no channel available
  *
  */
-int platform_rmt_allocate( uint8_t num_mem, uint8_t is_tx );
+int platform_rmt_allocate( uint8_t num_mem, rmt_mode_t mode );
 
 /**
  * @brief Release a previously allocated RMT channel.
--- a/components/platform/include/platform_rmt.h
+++ b/components/platform/include/platform_rmt.h
@@ -14,13 +14,14 @@
  * @brief Allocate an RMT channel.
  *
  * @param num_mem Number of memory blocks.
+ * @param is_tx Mode of the channel, !=0 allocates a TX channel, ==0 an RX channel.
  *
  * @return
  *     - Channel number when successful
  *     - -1 if no channel available
  *
  */
-int platform_rmt_allocate( uint8_t num_mem );
+int platform_rmt_allocate( uint8_t num_mem, uint8_t is_tx );
 
 /**
  * @brief Release a previously allocated RMT channel.
--- a/components/platform/onewire.c
+++ b/components/platform/onewire.c
@@ -113,8 +113,8 @@ static int onewire_rmt_init( uint8_t gpio_num )
   }
 
   // acquire an RMT module for TX and RX each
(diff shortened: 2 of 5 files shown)


components/platform/include/platform_rmt.h
```diff
@@ -14,13 +14,14 @@
...
  * @param num_mem Number of memory blocks.
+ * @param is_tx Mode of the channel, !=0 allocates a TX channel, ==0 an RX channel.
  *
  * @return
  *     - Channel number when successful
  *     - -1 if no channel available
  *
  */
-int platform_rmt_allocate( uint8_t num_mem );
+int platform_rmt_allocate( uint8_t num_mem, uint8_t is_tx );
```
jmattsson, line 24: Could we please have an enum with `RMT_TX`/`RMT_RX` instead of a non-descript `uint8_t` for the `is_tx` parameter? It will help readability in the places it gets used.

components/platform/onewire.c
```diff
@@ -113,8 +113,8 @@ static int onewire_rmt_init( uint8_t gpio_num )
   }
 
   // acquire an RMT module for TX and RX each
-  if ((ow_rmt.tx = platform_rmt_allocate( 1 )) >= 0) {
-    if ((ow_rmt.rx = platform_rmt_allocate( 1 )) >= 0) {
+  if ((ow_rmt.tx = platform_rmt_allocate( 1, 1 )) >= 0) {
+    if ((ow_rmt.rx = platform_rmt_allocate( 1, 0 )) >= 0) {
```
jmattsson, line 117: `platform_rmt_allocate(1, RMT_TX)` would be easier to follow.

jmattsson (changes requested):
Thanks for both pointing out the issue and providing a PR!
Other than one minor request about naming, this looks good. I'm in no position to test it myself, but the changes look sane so happy to approve :)

docbacardi:
Back at work. :) I agree, an enum makes it a lot easier. Lucky me found something ready-made in the SDK.

jmattsson (approved):
Looks good to me, thanks!

PR #3486, opened 2021-12-30
The next round of test work

nwf:

Here's a slice through the test framework I've had in my back pocket.  Of importance:
- it introduces, and demonstrates initial minimal use of, a `NTestEnv` module (and associated file in SPIFFS); this will let us flexibly extend the notion of "test environment" without requiring that everything move in lockstep.
- the MQTT test program demonstrates the real power of the "harness" approach, IMHO
- the top-level runner (`tests/all.sh`) shows how the whole thing hangs together

Thoughts more than welcome.


tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
...
+  -- this isn't great, but it's what we've got
+  local mstr = node.info('build_config').modules
+  return mstr:match("^"..m..",") -- first module in the list
+      or mstr:match(","..m.."$") -- last module in the list
+      or mstr:match(","..m..",") -- somewhere else
+end
+
+-- Determine if a given Lua module is available for require-ing.
+function NTE.hasL(m)
+  for _, l in ipairs(package.loaders) do
```
HHHartmann, line 17: Why not use node.LFS.get()

tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
+-- A series of convenient utility functions for use with NTest on NodeMCU
+-- devices under test (DUTs).
+
+local NTE = {}
+
+-- Determine if the firmware has been built with a given module
+function NTE.hasC(m)
+  -- this isn't great, but it's what we've got
+  local mstr = node.info('build_config').modules
```
HHHartmann, line 9: you could also just check if the name is available in the root namespace (and is a rom table maybe)

tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
...
+  return mstr:match("^"..m..",") -- first module in the list
+      or mstr:match(","..m.."$") -- last module in the list
+      or mstr:match(","..m..",") -- somewhere else
+end
+
+-- Determine if a given Lua module is available for require-ing.
+function NTE.hasL(m)
```
HHHartmann, line 16: I'd prefer `hasLua`. On a host system (with more memory) I would even call it `hasLuaModule`

tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
...
+  for _, l in ipairs(package.loaders) do
+    if type(l(m)) == "function" then return true end
+  end
+  return false
+end
+
+-- Look up one (or more) feature keys in our configuration file,
+-- returning the associated value (or a map from keys to values).
+function NTE.getFeat(...)
```
HHHartmann, line 25: How about `getFeatures`. The plural indicates that several params can be given at once

tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
...
+      end
+    }
+  })
+  local cfgf = file.open("testenv.conf", "r")
+
+  if cfgf == nil then return {} end -- no features if no config file
+
+  local cstr
+  repeat
+    cstr = cfgf:read(); decoder:write(cstr or "")
```
HHHartmann, line 57: Does it make sense to have smaller chunks than the default 1024 bytes?
Should we call an explicit garbage collect from time to time?

tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
...
+  until (not cstr or #cstr == 0)
+  cfgf:close()
+  local givenFeats = decoder:result()
+  assert (type(givenFeats) == "table", "Malformed configuration file")
+
+  local res = {}
+  for _, v in ipairs(reqFeats) do
```
HHHartmann, line 64: Maybe this should be done only on request.
I imagine a test that behaves differently on existence of a feature.
Maybe have on of
- a separate function that errors on missing features
- a parameter that steers the behavior
- a separate function that is called with the result of `getFeat` and the list of required features to error out

tests/utils/NTestEnv.lua
```diff
@@ -0,0 +1,72 @@
...
+  --   "feat2" : ...
+  -- }
+  local reqFeats = { ... }
+  local decoder = sjson.decoder({
+    metatable = {
+      checkpath = function(_, p)
+        if #p > 1 then return true end            -- something we're keeping
+        if #p == 0 then return true end           -- root table
+        local thisFeat = p[1]
+        assert (type(thisFeat) == "string")
```
HHHartmann, line 43: not sure about json syntax but are there other keys than strings allowed?
w3schools states: "In JSON, keys must be strings, written with double quotes"
so this check can be removed

tests/NTest_adc_env.lua
```diff
@@ -4,21 +4,23 @@
...
 
 -- TODO: Use the mcp23017 module in the main tree rather than hand-coding
 -- the commands
 
 N.test('setup', function()
-  -- Configure the ADC
+  -- Ensure that we're on DUT 0
+  assert(NTE.getFeat('DUT') == 0, "Not on DUT 0")
+
+  -- Configure the ADC (this implicitly checks for having the ADC module)
```
HHHartmann, line 16: Early checks with distinct error message are a nice thing to have. So since we now have respective methods we should use them. Maybe even have such checks for basic modules as sjson before using them in `NTestEnv`

tests/NTest_adc_env.lua
```diff
@@ -4,21 +4,23 @@
...
   if adc.force_init_mode(adc.INIT_ADC)
   then
     node.restart()
     error "Must reboot to get to ADC mode"
   end
 
-  -- Configure the I2C bus
+  -- Configure the I2C bus (again, implicitly testing...)
```
HHHartmann, line 23: again ... ;-)

tests/utils/NTestEnv.lua
HHHartmann: Ah, I see the difference. Your method will also find modules in SPIFFS.
But will it not load all available modules,  that way exhausting memory if SPIFFS is not cleaned out?
But I am no expert in loaders.

HHHartmann (changes requested):
Good approach. Happy to see things going on.
For extra points you could add a test for `NTestEnv` itself

Maybe it would make sense to allow for several parameters in hasLua and hasC.

tests/HardwareTestHarness.md
```diff
@@ -94,22 +64,34 @@ LED is connected to the INT pin and so you can disable the LED under software co
...
 5V voltage rail. 
 
+Note: Not yet used in the test harness.
+
 #### DHTxx
 
 On DUT1 pin D6/GPIO 12 there is a connection to a position for a DHTxx device. The silk screen indicates the
 orientation of the device.
 
+Note: Not yet used in the test harness.
```
HHHartmann, line 81: you mean the software test harness? This is the documentation of the HWTH

tests/HardwareTestHarness.md
```diff
@@ -94,22 +64,34 @@ LED is connected to the INT pin and so you can disable the LED under software co
...
+
 #### DS18B20
 
 There are two positions for DS18B20s -- one with the VCC pin connected and one without. The data pin is
 connected to DUT1 pin D5/GPIO 14.
 
+Note: Not yet used in the test harness.
+
+Note: It would make sense to augment the 1-Wire testing facility to
+include bus-drive power, perhaps via the MCP23017, especially if we ever
```
HHHartmann, line 91: As states above there are already 2 devices possible, one without VCC connected which then would be bus-driven

tests/HardwareTestHarness.md
```diff
@@ -94,22 +64,34 @@ LED is connected to the INT pin and so you can disable the LED under software co
...
 orientation of the device.
 
+Note: Not yet used in the test harness.
+
 #### DS18B20
 
 There are two positions for DS18B20s -- one with the VCC pin connected and one without. The data pin is
 connected to DUT1 pin D5/GPIO 14.
 
+Note: Not yet used in the test harness.
```
HHHartmann, line 88: meaning no tests available yet?

tests/HardwareTestHarness.md
```diff
@@ -48,43 +48,13 @@ Peripherals
...
-
-At address 0x20. An 16-bit tristate GPIO expander, this chip is used to
-test I2C, GPIO, and ADC functionality. This chip's interconnections are
-as follows:
-
-MPC23017 | Purpose
----------|--------------------------------------------------------------
-/RESET   |DUT0 reset. This resets the chip whenever the host computer resets DUT 0 over its serial link (using DTR/RTS).
-B 0      |4K7 resistor to DUT 0 ADC.
-B 1      |2K2 resistor to DUT 0 ADC.
```
HHHartmann, line 67: again B2,3 and 4 are not described in the readme

tests/HardwareTestHarness.md
```diff
@@ -121,61 +103,9 @@ There are three positions for I2C devices with other pin orders. Each of these p
...
 Breadboard Area
 ===============
 
 All the pins on each D1 Mini and the A port of the MCP23017 are brought out to a breadboard 
 area. This can be used to solder components and/or wires, or even a header could be soldered
 on to transfer all the signals to a conventional breadboard. 
-
-
-ESP8266 Device 0 Connections
-----------------------------
```
HHHartmann, line 134: The description here is more exact and exceeds the spec in the readme, so it should be kept.
e.g.; D4/5 of DUT 1 is not described at all.
Also in general the setup might evolve in the readme but this document describes a hardware version, which will remain unchanged (else it will be a v2)

HHHartmann (changes requested):
forgot to check where the deleted content was moved to

PR #3488, opened 2022-01-01
Added mkdir/rmdir support to file module.

jmattsson:

When working with SD cards formatting with FAT, having the ability to create and remove directories become important. This PR exposes the IDF's `mkdir` and `rmdir` functions via the `file` module, since there is no standard Lua interface for it.

### DIFF
--- a/components/modules/file.c
+++ b/components/modules/file.c
@@ -11,6 +11,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 #include <dirent.h>
+#include <errno.h>
 
 static const char *default_fs_label =
   ((CONFIG_NODEMCU_DEFAULT_SPIFFS_LABEL &&
@@ -116,6 +117,29 @@ static int file_fsinfo( lua_State* L )
 }
 
 
+static int file_mkdir( lua_State *L )
+{
+  const char *name = luaL_checkstring(L, 1);
+  unsigned mode = luaL_optint(L, 2, 0777);
+  if (mkdir(name, mode) != 0) {
+    return
+      luaL_error(L, "failed to create directory '%s'; code %d", name, errno);
+  }
+  return 0;
+}
+
+
+static int file_rmdir( lua_State *L )
+{
+  const char *name = luaL_checkstring(L, 1);
+  if (rmdir(name) != 0) {
+    return
+      luaL_error(L, "failed to remove directory '%s'; code %d", name, errno);
+  }
+  return 0;
+}
+
+
 // Module function map
 LROT_BEGIN(file, NULL, 0)
   LROT_FUNCENTRY( list,      file_list )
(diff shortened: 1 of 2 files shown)


components/modules/file.c
```diff
@@ -116,6 +117,30 @@ static int file_fsinfo( lua_State* L )
 }
 
 
+static int file_mkdir( lua_State *L )
+{
+  const char *name = luaL_checkstring(L, 1);
+  unsigned mode = luaL_optint(L, 2, 0777);
+  int ret = mkdir(name, mode);
+  if (ret != 0)
```
marcelstoer: You know I'm not a firmware programmer 😉 but I'd write this like so:

```c++
if (mkdir(name, mode) != 0) {
  return luaL_error(L, "failed to create directory '%s'; code %d", name, errno);
}
return 0;
```

- you use `if`/`else` here but only `if` in `file_rmdir` (inconsistent style)
- no need for the local `ret` variable
- `{}` even for single line `if` blocks, seems consistent with the other code in `file.c`

jmattsson: You keep claiming not to be one, yet the quality I've seen from you surpasses that of many a professional firmware programmer I've come across.

All fair points; I've updated accordingly, thanks.

Oh, and when you're ready, let's do the `dev-esp32` -> `dev-esp32-idf3-final` + `dev-esp32-idf4` -> `dev-esp32` renames. As far as I'm aware the idf4 branch is production ready (or at least I hope so, seeing as we've been shipping devices based off it for a while at $work 😆).

marcelstoer: > Oh, and when you're ready, let's do the `dev-esp32` -> `dev-esp32-idf3-final` + `dev-esp32-idf4` -> `dev-esp32` renames.

Oh absolutely, I am not opposed to that. The fact that neither https://github.com/marcelstoer/docker-nodemcu-build/issues/101 nor https://github.com/marcelstoer/nodemcu-custom-build/issues/41 are ready, doesn't mean the progress of _this_ project should be held hostage.

jmattsson:
Somehow a year has already passed. @marcelstoer I'm going to both merge this and do the discussed branch renaming now; I've dawdled long enough, and I'm going to feel silly working on IDF5 support on a branch named IDF4 :D
Also CC @pjsg a heads-up. I think we're the only active members on ESP32 at the moment?

marcelstoer:
Thanks Johny. Can you also close the "migrate to IDF 4" tickets. IIRC there must be more than one.

I am really out of touch with the development here. Hence, I might need some help with https://github.com/marcelstoer/docker-nodemcu-build/issues/101 and https://github.com/marcelstoer/nodemcu-custom-build/issues/41. Probably won't take long until the support emails start coming in now 😄

jmattsson:
I don't see any such tickets here. Do you have a link?

marcelstoer:
Looks like #3397 is the only one out of those labeled with "ESP32": https://github.com/nodemcu/nodemcu-firmware/issues?q=is%3Aopen+is%3Aissue+label%3AESP32

PR #3489, opened 2022-01-02
docs: node: point at telnet module now

nwf:
See 8d091c476edf6ae2977a5f2a74bf5824d07d6183

### DIFF
--- a/docs/modules/node.md
+++ b/docs/modules/node.md
@@ -340,7 +340,8 @@ Submits a string to the Lua interpreter. Similar to `pcall(loadstring(str))`, bu
 ```lua
 sk:on("receive", function(conn, payload) node.input(payload) end)
 ```
-See the `telnet/telnet.lua` in `lua_examples` for a more comprehensive example.
+See [the telnet lua module](../lua_modules/telnet.md) for a worked
+example.
 
 #### See also
 [`node.output()`](#nodeoutput)
@@ -420,7 +421,8 @@ Redirects the Lua interpreter to a `stdout` pipe when a CB function is specified
 
 #### Example
 
-See the `telnet/telnet.lua` in `lua_examples` for a more comprehensive example of its use.
+See [the telnet lua module](../lua_modules/telnet.md) for a worked
+example.
 
 #### See also
 [`node.input()`](#nodeinput)
--- a/docs/modules/node.md
+++ b/docs/modules/node.md
@@ -340,7 +340,7 @@ Submits a string to the Lua interpreter. Similar to `pcall(loadstring(str))`, bu
 ```lua
 sk:on("receive", function(conn, payload) node.input(payload) end)
 ```
-See [the telnet lua module](../lua_modules/telnet.md) for a worked
+See [the telnet lua module](../lua-modules/telnet.md) for a working
 example.
 
 #### See also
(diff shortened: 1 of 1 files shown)


docs/modules/node.md
```diff
@@ -340,7 +340,8 @@ Submits a string to the Lua interpreter. Similar to `pcall(loadstring(str))`, bu
 ```lua
 sk:on("receive", function(conn, payload) node.input(payload) end)
 ```
-See the `telnet/telnet.lua` in `lua_examples` for a more comprehensive example.
+See [the telnet lua module](../lua_modules/telnet.md) for a worked
```
marcelstoer: Either "`telnet` Lua" or "Telnet Lua", isn't it?

docs/modules/node.md
```diff
@@ -420,7 +421,8 @@ Redirects the Lua interpreter to a `stdout` pipe when a CB function is specified
 
 #### Example
 
-See the `telnet/telnet.lua` in `lua_examples` for a more comprehensive example of its use.
+See [the telnet lua module](../lua_modules/telnet.md) for a worked
```
HHHartmann: working example

HHHartmann: Unfortunately the link has to be lua-modules, not with underscore. Don't know how we got that in. Or we rename the docs/lua-modules folder.

marcelstoer:
Just fixed the links (from `_` to `-`).

HHHartmann (approved):

HHHartmann:
OK. This doesn't work in https://nodemcu.readthedocs.io/en/dev/modules/node/#nodeinput
The link gets expanded to
https://github.com/nodemcu/nodemcu-firmware/tree/dev/lua-modules/telnet/
Should be
https://nodemcu.readthedocs.io/en/dev/lua-modules/telnet/

marcelstoer:
Oh, just realized we have the same issue on the BME280 math page.

Caused by this https://github.com/nodemcu/nodemcu-firmware/blob/release/docs/js/extra.js#L46

HHHartmann:
Staring at the code not knowing much about js it might be that adding a "/" to the relativePath might do the job. 
I assume that the dots are evaluated as wildcats matching any character. Thus effectively matching the intended start of "../../dev" but also "../lua-modules"

marcelstoer:
With the docs a lot of stuff isn't as it may seem - and much of it stems from the requirement of having links work both while browsing GitHub and inside the docs.

- In [the original `node.md`](https://raw.githubusercontent.com/nodemcu/nodemcu-firmware/dev/docs/modules/node.md) with have `See [the telnet lua module](../lua-modules/telnet.md) for...` -> relative link 1 level up
- His works when browsing GitHub
- When this is built into the HTML structure by RTD/MkDocs this becomes `<p>See <a href="../../lua-modules/telnet/">the telnet lua module</a> for a working` -> relative link 2 levels up
  - This is basically fine (MkDocs ensures all relative links are consistent) as we want to go from
  https://nodemcu.readthedocs.io/en/dev/modules/node/index.html#nodeinput
  to 
  https://nodemcu.readthedocs.io/en/dev/lua-modules/telnet/index.html
  - It clashes, however, with all links that already have `../../` _in the original_ `.md`. This is used in links to `.c` source files for example: `[node.c](../../app/modules/node.c)`. MkDocs realizes those links point beyond the docs structure and keeps them as-is (rather than adding another level).
 - Our JavaScript code can't tell the two apart and applies the "fix" to both.

Side note: in JavaScript the pattern in `String.prototype.replace()` can be a string or a RegExp; we use it with a string.

PR #3496, opened 2022-01-12
Give the SPI module a chance of working...

### LINKED ISSUE #3494: spi in dev-esp32-idf4 seems broken
### Expected behavior
```
device:transfer({address=1, rxlen=1, txdata=string.char(0)})
```
ought to work
### Actual behavior

It actually says
```
spi_master: check_trans_valid(687): rx length > tx length in full duplex mode
```

### Root cause

It appears that the code in `spi_master.c` is broken:

in `lspi_device_transfer` the following code is missing the line that I've marked

```
    trans.cmd  = opt_checkint( L, "cmd", 0 );
    trans.addr = opt_checkint( L, "addr", 0 );
    //
    rx_len = opt_checkint( L, "rxlen", 0 );
    //
    trans.flags |= opt_checkbool( L, "addr_mode", false ) ? SPI_TRANS_MODE_DIOQIO_ADDR : 0;
    //
    lua_getfield( L, stack, "mode" );
    trans.flags |= options_flags[ luaL_checkoption( L, -1, options[0], options ) ];
    lua_pop(L, 1);       //     MISSING LINE  ---------------------------------------------------
    //
    data_len = 0;
    data = opt_checklstring( L, "txdata", "", &data_len );
```

Without this line, the txdata doesn't get fetched from the right place (as the lua_getfield pushes an element onto the stack).

This leads to a more general question -- does anybody have working spi code? I'm trying to talk to a device but it doesn't appear to work. I may have to break out the logic analyzer....

pjsg:
Fixes #3494.

There were a number of things broken about the `spi` support.

* The parameter to `transfer` for the address was actually `addr` rather than `address`
* If specified, the `cs` line has to be configured as a gpio and set to the inactive value
* The `txdata` argument to `transfer` didn't work.

This PR fixes all of the above and it appears to work (at least it drives the one device that I have tried it on!)

### DIFF
--- a/docs/modules/spi.md
+++ b/docs/modules/spi.md
@@ -148,7 +148,7 @@ device:transfer(txdata)
 `trans` table containing the elements of the transaction:
 
 - `command` data for command phase, amount of bits was defined during device creation, optional
-- `address` data for address phase, amount of bits was defined during device creation, optional
+- `addr` data for address phase, amount of bits was defined during device creation, optional
 - `txdata` string of data to be sent to the device, optional
 - `rxlen` number of bytes to be received, optional
 - `mode` optional, one of
--- a/docs/modules/spi.md
+++ b/docs/modules/spi.md
@@ -147,7 +147,7 @@ device:transfer(txdata)
 #### Parameters
 `trans` table containing the elements of the transaction:
 
-- `command` data for command phase, amount of bits was defined during device creation, optional
+- `cmd` data for command phase, amount of bits was defined during device creation, optional
 - `addr` data for address phase, amount of bits was defined during device creation, optional
 - `txdata` string of data to be sent to the device, optional
 - `rxlen` number of bytes to be received, optional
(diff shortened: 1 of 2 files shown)


docs/modules/spi.md
```diff
@@ -148,7 +148,7 @@ device:transfer(txdata)
 `trans` table containing the elements of the transaction:
 
 - `command` data for command phase, amount of bits was defined during device creation, optional
-- `address` data for address phase, amount of bits was defined during device creation, optional
+- `addr` data for address phase, amount of bits was defined during device creation, optional
```
marcelstoer, line 151: Is this the string literal used here?

https://github.com/nodemcu/nodemcu-firmware/blob/1b98a42002e2988897daaca81edf853e56edbf6e/components/modules/spi_master.c#L102-L103

If so then `command` above should be renamed to `cmd`.

marcelstoer:
>  requested a review from marcelstoer

Sorry Philip, I wish I _did_ understand that code. I guess @jmattsson should take a look instead.

jmattsson:
Other than the documentation issue Marcel picked up on, this looks good as far as I can see. Feel free to merge once that command->cmd has been adressed!

PR #3499, opened 2022-01-19
Fix build for WSL users

ilkka:
This PR just puts the argument to one `echo` in quotes, in order to prevent the shell from erroring out in case the argument (the user's `$PATH`) contains things like parentheses, which is quite likely in case of WSL.


HHHartmann (approved):
I assume that the echo command fails for certain pathes with spaces
The fix doesn't break any of our standard builds, so it seems to be ok.

pjsg (approved):

PR #3509, opened 2022-02-23
Use a translator callback to convert GRB data to RMT pulses.

### LINKED ISSUE #3508: ws2812.write only emits the reset signal on ESP32-C3 but not the data
### Expected behavior

The ```ws2812.write``` function should emit the reset signal and the data.
   
### Actual behavior

On the ESP32-C3 only the reset signal is generated. The data is not generated.

I could narrow down the problem to the function ```ws2812_fill_memory_encoded``` which writes new items to the RMT memory, but nothing appears on the outside. To be honest I have no idea what is going wrong, but I have a workaround. The new SDK offers a translator callback to generate new RMT items on the fly from input data. I have a patch for this.

### Test code
This should burn a hole in your retina using the onboard RGB LED of the ESP32-C3-DevKitM-1:
```Lua
ws2812.write({pin = 8, data = string.char(255,0,0)})
```
### NodeMCU startup banner
NodeMCU ESP32 build unspecified powered by Lua 5.1.4 [5.1-doublefp] on IDF v4.4-beta1-284-ge5892a7286

### Hardware
ESP32-C3

docbacardi:
Fixes #3508.

This patch replaces the custom ISR in ws2812.write with a translator callback.
Furthermore it introduces 2 small improvements:
1) If the hardware supports it, all channels in the write request are placed in a group. This moves the start of all RMT sequences much closer together. On my hardware I observed >=50ns .
2) New optional parameters can set the duration of the reset signal and the duration for the high and low phase of the 0 and 1 symbols. This allows to adapt the timings to many other RGB LEDs.

### DIFF
--- a/components/modules/ws2812.c
+++ b/components/modules/ws2812.c
@@ -11,6 +11,13 @@
 #define SHIFT_LOGICAL  0
 #define SHIFT_CIRCULAR 1
 
+// The default bit H & L durations in multiples of 100ns.
+#define WS2812_DURATION_T0H 4
+#define WS2812_DURATION_T0L 7
+#define WS2812_DURATION_T1H 8
+#define WS2812_DURATION_T1L 6
+// The default reset duration in multiples of 100ns.
+#define WS2812_DURATION_RESET 500
 
 typedef struct {
   int size;
@@ -33,6 +40,7 @@ static void ws2812_cleanup( lua_State *L, int pop )
 // ws2812.write({pin = 4, data = string.char(255, 0, 0, 255, 255, 255)}) first LED green, second LED white.
 static int ws2812_write( lua_State* L )
 {
+  int type;
   int top = lua_gettop( L );
 
   for (int stack = 1; stack <= top; stack++) {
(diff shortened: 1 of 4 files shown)


pjsg:
This looks plausible to me -- I need to get some ws2812s attached to an esp32 to test.....

docbacardi:
The module name might have become a little bit misleading. It is not ws2812 specific, but works with a lot of other RGB LEDs. I did my tests with a single SK6812 on the ESP32-C3 devkit. Another unknown LED strip from some home improvement store is on the way.

HHHartmann:
But there are other led strips which will not work,  so that would be as misleading.

jmattsson:
@docbacardi @pjsg any chance of seeing what needs tweaking to make this compatible with the current IDF5 branch? I'm out of my depth with the RMT stuff.

docbacardi:
I will try and port this. Also a good chance to move to the new RMT driver.

docbacardi:
Please find the patch here: #3629 

The migration to the new RMT code in IDF5 would affect more modules. One step after another...

jmattsson:
Is this PR still relevant with #3629 having been merged?

jmattsson:
Closing as superseded.

PR #3514, opened 2022-04-30
Working version of the websocket support

pjsg:
Fixes #<GITHUB_ISSUE_NUMBER>.

This is the websocket enhancement to the httpd implementation. Unfortunately, it doesn't actually work due to a bug in the IDF, but there is a fix, and I'm hopeful that it will get merged soon. https://github.com/espressif/esp-idf/issues/8596


pjsg:
It still has debugging statement that I will remove once it works well.

pjsg:
I'm having problems in deciding what the lifetime of the ws object should be. I suspect that it isn't right at the moment. It deserves more thought.

jmattsson:
Any further thoughts on the `ws` object lifetime? Is the current state good enough?

pjsg:
After more thought, I've fixed the lifetime problems. Also I got rid of a memory leak. This is good to go.

docs/modules/httpd.md
```diff
@@ -256,6 +256,66 @@ end
...
+
+#### Returns
+nil
+
+#### Example
+```lua
+httpd.start({
+  webroot = "<static file prefix>",
+  max_handlers = 20,
+  auto_index = httpd.INDEX_NONE + httpd.INDEX_ROOT + httpd.INDEX_ALL,
```
jmattsson: I think you want `||` rather than `+` here; they're meant to be a choice, not bits to be set together :)

jmattsson:
This turned out to be more effort than I'd expected, so I'm glad you tackled it instead of me :D

I haven't tried it out or done an in-depth code review, but if you reckon it's ready to go, I'm happy to go with that. I left one comment about the docs that you might want to check before merging, but other than that, it'd be good to get this functionality in!

docs/modules/httpd.md
```diff
@@ -76,7 +76,7 @@ configured.
 httpd.start({
   webroot = "<static file prefix>",
   max_handlers = 20,
-  auto_index = httpd.INDEX_NONE || httpd.INDEX_ROOT || httpd.INDEX_ALL,
+  auto_index = httpd.INDEX_ALL,
```
jmattsson, line 79: This removes the mention of the other index modes, I can't say I'm in favour of that 🙃

pjsg: It turns out that these are actually enums and shouldn't be or'ed together. Worse, the example uses || which is the logical or. Thus, this expression returns the first non-zero value!!

jmattsson: Yes, logical OR, because you need to choose one of them. What's the correct syntax for that in this case, if not that? (genuine question, not a snarky comment)

PR #3519, opened 2022-05-28
Ensure flash address is set before obtaining memory mapped address.

h2zero:

Before this change the struck member add_phys would contain random data and the call to spi_flash_phys2cache
could return an incorrect memory address.


jmattsson:
Yikes, I'm surprised this hasn't bitten me before!
Thanks for the fix!

PR #3520, opened 2022-05-28
Fix node.compile fwrite size check.

h2zero:

The return value from fwrite was being checked against the size of the data rather than the number of bytes written.
This caused node.compile() to falsely return failure.


jmattsson:
Nice catch, thanks!

PR #3521, opened 2022-05-28
Explicitly set RMT config::flags.

h2zero:

If this variable is not set in the config it may contain random data and set the clock source to the REF_CLK
resulting in timing errors, the DHTx, DS18B20 and ws2812 devices to not communicate.

### DIFF
--- a/components/platform/dht.c
+++ b/components/platform/dht.c
@@ -86,6 +86,7 @@ static int dht_init( uint8_t gpio_num )
     rmt_rx.channel = dht_rmt.channel;
     rmt_rx.gpio_num = gpio_num;
     rmt_rx.clk_div = 80;  // base period is 1us
+    rmt_rx.flags = 0;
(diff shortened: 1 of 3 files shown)


jmattsson:
We should probably either `memset(ptr, 0, size)` these structs, or make use of automatic zero-fill of partially initialised structs, but I'm happy to get this set of minimal fixes in now.
Thanks for the PR!

h2zero:
I think all the other parameters are set, this is just a new one for IDF v4.x. There are also macros for `RMT_DEFAULT_CONFIG_TX` and `RMT_DEFAULT_CONFIG_RX` that could be used here.

jmattsson:
@h2zero Oooh, that would be even more future-proof! If you want to PR it, just tag me as a reviewer and I can get it merged. I'm a bit overloaded with stuff at the moment, so I only see things that directly mention me.

PR #3525, opened 2022-06-16
embed_lfs.sh: handle riscv.attributes in esp32c3 map file

jasaw:
esp32c3 map file contains `.riscv.attributes` that causes `embed_lfs.sh` to fail.
```
 .riscv.attributes
                0x0000000000001360       0x24 esp-idf/embedded_lfs/libembedded_lfs.a(lua.flash.store.reserved.S.obj)
```

This PR fixes that.

@jmattsson Please review.

### DIFF
--- a/tools/embed_lfs.sh
+++ b/tools/embed_lfs.sh
@@ -15,7 +15,7 @@ if [ ! -f "${LUAC_CROSS}" ]; then
 	exit 1
 fi
 
-LFS_ADDR_SIZE=$(grep -E "0x[0-9a-f]+[ ]+0x[0-9a-f]+[ ]+esp-idf/embedded_lfs/libembedded_lfs.a\(lua.flash.store.reserved.S.obj\)" "${MAP_FILE}" | grep -v -w 0x0 | tr -s ' ')
+LFS_ADDR_SIZE=$(grep -E "0x[0-9a-f]+[ ]+0x[0-9a-f]+[ ]+esp-idf/embedded_lfs/libembedded_lfs.a\(lua.flash.store.reserved.S.obj\)" "${MAP_FILE}" | grep -v -w 0x0 | grep -v -w 0x24 | tr -s ' ')
 if [ -z "${LFS_ADDR_SIZE}" ]; then
 	echo "Error: LFS segment not found. Use 'make clean; make' perhaps?"
 	exit 1


jmattsson:
Thanks! We could probably optimise it to be a single `grep -v -w -e 0x0 -e 0x24` but this will do.

HHHartmann:
Hey guys, wasn't aware that we have such a feature.
Would it be complicated to port it to the esp8266 branch?

jmattsson:
Oh gosh, I haven't looked at ESP8266 stuff in a couple of years now!
I expect it would be doable, but a bit of work. On the ESP32 side we're depending on the IDF's ability to embed random data into the firmware, which would need to be replicated. The LFS support code inside Lua should hopefully port straight over.

Embedding an LFS image is a two-step process:
  1. Build the firmware once with a dummy file of the correct size.
  2. Parse the resulting map file to get the correct addresses to pass to luac when generating the LFS image, and then rebuild the firmware using the LFS image instead of the dummy file.
  
For the 8266 we could probably get away without doing a full re-link and instead just write the LFS image into the correct address in the binary (on the esp32 the app is checksummed, so not an option there).

I'm happy to answer questions, but I don't have any surplus time/energy myself right now.

HHHartmann:
Asking questions was what I was hoping for. 
So the embedded LFS can be overwritten on chip? Or will there be  a second LFS region available?
Or is it just fixed at buildtime and there will be no further updating possible?

jmattsson:
This is fixed at build time. It's a way of combining C firmware and Lua firmware into a single binary for easy OTA upgrades. No support for replacing the embedded LFS at runtime (again, on the ESP32 that's not doable since the firmware is checksum and possibly also signed). To upgrade, you upgrade the full firmware. On the ESP32 you get A/B firmware upgrades with rollback support out of the box/IDF. At $work on the 8266 we're using a [dedicated boot loader](https://github.com/DiUS/nodemcu-firmware/tree/dius-2.x/nodemcu-boot) to do A/B firmware slots with rollback support. A long time ago I raised a PR about whether there was any interest in mainlining that, but there didn't seem to be. There was also talk about using raburton's rBoot, but again, insufficient interest in the end.

I know Terry was thinking about the possibility of supporting multiple LFS volumes, but I don't think it went beyond the thinking phase.

PR #3529, opened 2022-07-18
Support sending strings with multi-byte-wide SPI

stuartpb:
The iteration in the loop in the prior version of spi.send assumes that the value of databits is the width of  `char` when sending any string value.

- [ ] I haven't thoroughly tested my contribution - it's a simple enough change that, if it compiles and passes existing tests, it works (it only fixes behavior that was previously broken).
- [x] The code changes are reflected in the documentation at `docs/*`. (This only fixes the function to more closely match the expected documented behavior.)

### DIFF
--- a/app/modules/spi.c
+++ b/app/modules/spi.c
@@ -140,13 +140,21 @@ static int spi_send_recv( lua_State *L )
     else
     {
       luaL_Buffer b;
+      size_t datachars = spi_databits[id] / 8;
 
+      if (spi_databits[id] % 8 != 0) {
+        return luaL_error( L, "attempt to send string with non-character-divisible databits" );
+      }
+      if (datalen % datachars != 0) {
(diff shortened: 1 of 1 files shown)


stuartpb:
There should probably be further tests with this to ensure that multi-char values from strings get read and sent properly - there could be endianness issues with the string layout.

Alternately, trying to send a string value with a databits value of anything other than 8 could be an error, or ignore the `databits` value and always use a width of 8 for strings, transmitted character-by-character (this override should probably be documented in that case).

jmattsson:
I don't see how `platform_spi_send( id, spi_databits[id], pdata[ i ] );` would work with >8 bits? The `pdata[i]` would only pass 8 bits of data through to the function, given that `pdata` is of `const char *` type. Am I missing something?

PR #3531, opened 2022-07-29
threads

nedoskiv:
Hello,

Sorry if my question is not for here, I search thru docs, and cannot find anything about threads support in nodemcu

Is it done and I missed it, if not is it planned to be done?


marcelstoer:
https://nodemcu.readthedocs.io/en/release/support/

PR #3532, opened 2022-07-31
Add support for the builtin USB CDC console on the ESP32C3 [WIP]

### LINKED ISSUE #3526: Support the USB CDC device in ESP32-C3
### Missing feature
If you are using the USB CDC device in the ESP32-C3 for the console, then it doesn't work. This feature would add support for using the CDC device as the console. Ideally, it would also appear as a uart so that the `uart` family of functions would also work. However, that raises the issue of what *number* the CDC device should be. The other two uart devices also exist. Either we use 0 for the CDC device (if it is configured) and push the real uarts down to 1 & 2, or we do something else. Opinions wanted.

Yes, I will implement this.

### Justification
It makes some cheap boards usable.

### Workarounds
You have to use an external Serial-USB chip as a workaround.

pjsg:
Fixes #3526.

If the USB CDC console is configured, then it appears as the default console for the Lua REPL. It also appears as UART 0 and bumps the other uart(s) down one. The baud rate cannot be configured as there is no actual serial line.

When configured this way, stdin and stdout are connected to this device. Unfortunately, async reading is not possible at the moment. Thus the input side is handled by doing blocking reads in a seperate thread, and then posting the data into the Lua environment. Happily this is very similar to how the uart code works.


pjsg:
What is puzzling me at the moment is how the command line echo works. In particular, I'm not getting the linefeeds echoed.

pjsg:
This got a whole bunch of other stuff merged into it.

PR #3538, opened 2022-08-29
uart: feed rx data to both lua console and on-data callback

jasaw:
UART on-data example usage with `run_input` set to 1:
```
uart.on(0, "data", 0,
  function(data)
    print("receive from uart:", data)
end, 1)
```

According to documentation for `run_input`:
>If 1, input from UART will go into Lua interpreter, **_and_** run.

Currently, UART data either go into Lua interpreter **_or_** on-data callback. This PR fixes that.

### DIFF
--- a/components/modules/uart.c
+++ b/components/modules/uart.c
@@ -10,6 +10,10 @@
 
 static lua_State *gL = NULL;
 
+bool uart_has_on_data_cb(unsigned id){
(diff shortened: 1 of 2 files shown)


jasaw:
@jmattsson I can't assign you as a reviewer, so I just ping you here instead.

jmattsson:
This looks like a needed alignment to the docs (and I suspect this might be my bug in the first place - apologies).

PR #3539, opened 2022-08-29
platform: platform_uart_setup handles null ptr

jasaw:
`uart_pins_t* pins` input parameter is supposed to accept a null pointer as well, but the code dereferences the `pins` pointer without checking whether `pins` is null or not.

This PR fixes it.

@jmattsson Please review.


jmattsson:
Looks rather needed, thanks!

PR #3540, opened 2022-08-29
uart: do not use uninitialized uart pins config

jasaw:
In some scenario, uart LUA module calls `platform_uart_setup` with uninitialized `uart_pins_t` pins config data structure. The result is random failure when calling `uart.setup` from LUA.

This PR fixes it.

@jmattsson Please review.


jmattsson:
Yikes, nice catch.

PR #3541, opened 2022-08-29
Slight clean up of uart.setup code

jasaw:
The current code is a little confusing to read and doesn't return error when given an invalid input.
There are 2 code paths here:
1. uart port is our console. In this case, we accept an optional `echo` parameter. If `echo` is not provided, we leave the existing `input_echo` flag unchanged.
2. uart port is **_NOT_** our console. We accept an optional `pins config table` parameter.

This PR cleans up the code a bit to make it clearer.

This PR makes changes on top of PR #3540

@jmattsson Please review.

### DIFF
--- a/components/modules/uart.c
+++ b/components/modules/uart.c
@@ -140,12 +140,10 @@ static int uart_setup( lua_State* L )
   databits = luaL_checkinteger( L, 3 );
   parity = luaL_checkinteger( L, 4 );
   stopbits = luaL_checkinteger( L, 5 );
-  if(id == CONFIG_ESP_CONSOLE_UART_NUM){
-    if (!lua_isnoneornil(L, 6)) {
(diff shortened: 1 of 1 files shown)


marcelstoer:
Since you have `if (!lua_isnoneornil(L, 6)) {` in both the `if` and the `else` branch I don't see why this couldn't be the primary condition to check (i.e. guarding the if-else with it).

jmattsson:
That's a fair point @marcelstoer. It would probably be a bit nicer.
I resolved the merge conflict that appeared when I merged the other patches, but I'll hold off merging this one until @jasaw has had a chance to comment.

jasaw:
@marcelstoer @jmattsson I've moved `if (!lua_isnoneornil(L, 6)) {` to the outside of the if-else block as suggested.

jmattsson:
Cheers!

PR #3545, opened 2022-08-30
platform uart: really feed uart data to both console and on-data cb

jasaw:
I made a mistake yesterday on the PR #3538 where first byte of uart data is fed to the LUA console and 2nd byte to the on-data callback.
It should have been feeding all bytes to both LUA console and on-data callback.

Sorry my bad.

This PR loops all the uart bytes for the console first, then loops all the uart bytes again for the on-data callback.

@jmattsson Please review.


jmattsson:
Well whoops, I too should've caught that yesterday!

PR #3548, opened 2022-10-07
Improve a simple HTTP server example in README.

stawiski:
Hi,

The simple HTTP server example which is the first thing people see going to the main page of the repo could be slightly improved with couple of simple changes like adding `Connection: close` or `Content-Length`.

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,11 @@ srv = net.createServer(net.TCP)
 srv:listen(80, function(conn)
 	conn:on("receive", function(sck, payload)
 		print(payload)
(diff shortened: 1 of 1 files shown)


marcelstoer:
That's a well-meant but dangerous "improvement", see https://nodemcu.readthedocs.io/en/latest/modules/net/#note.

stawiski:
> That's a well-meant but dangerous "improvement", see https://nodemcu.readthedocs.io/en/latest/modules/net/#note.

Ok thanks, that's an easy fix to combine it to one `send`. Will post an update.

PR #3552, opened 2022-10-23
Allow LUA_INIT_STRING to be set in menuconfig

tomsci:
- [x] This PR is for the `dev-esp32-idf4` branch rather than for the `release` branch.
- [x] This PR is compliant with the [other contributing guidelines](/CONTRIBUTING.md) as well
_The old option wasn't documented anywhere I could find, so I haven't added anything other than describing the new option in Kconfig._

With the build system changes in the idf4 branch, it's no longer possible to set `LUA_INIT_STRING` using `MORE_CFLAGS="-DLUA_INIT_STRING=..."`. So instead, make it a menuconfig option.

```
(Top) → Component config → Lua configuration

    Lua version (Lua 5.1)  --->
[*] Integer-only build
    Core Lua modules  --->
    Lua compilation  --->
    Boot command (init.lua from SPIFFS)  --->
```

which expands to:

```
(Top) → Component config → Lua configuration → Boot command

(X) init.lua from SPIFFS
( ) init module from LFS
( ) Custom
```

This removes the previous hard-coded setting when CONFIG_NODEMCU_EMBED_LFS is defined - meaning anyone relying on that setting to also configure LUA_INIT_STRING, will need to rerun `make menuconfig` to select `init module from LFS`. I couldn't seem to find a good way of expressing "if CONFIG_NODEMCU_EMBED_LFS is enabled default to LFS otherwise init.lua" in Kconfig without making the setting really weird.

### DIFF
--- a/components/lua/Kconfig
+++ b/components/lua/Kconfig
@@ -132,4 +132,35 @@ menu "Lua configuration"
       select NODEMCU_CMODULE_UART
       select LUA_BUILTIN_DEBUG
 
+    choice LUA_INIT_STRING
+        prompt "Boot command"
+        default LUA_INIT_STRING_INIT_LUA
+        help
+            Command to run on boot. This can be a .lua file, an LFS module, or
+            any valid Lua expression. By default init.lua is loaded and run
+            from the SPIFFS filesystem.
+        config LUA_INIT_STRING_INIT_LUA
+            bool "init.lua from SPIFFS"
+        config LUA_INIT_STRING_INIT_LFS
+            bool "init module from LFS"
+        config LUA_INIT_STRING_CUSTOM
+            bool "Custom"
+    endchoice
+
+    config LUA_INIT_STRING_CUSTOM_STRING
+        string "Custom boot command" if LUA_INIT_STRING_CUSTOM
+        default ""
+        help
+            Run a custom command on boot.
+
+            Specify @filename.lua to load "filename.lua" from SPIFFS.
+            Specify node.LFS.get('foo')() to load the module "foo" from LFS.
+            Or specify any other valid Lua expression to execute that on boot.
+
+    config LUA_INIT_STRING
+        string
+        default "@init.lua" if LUA_INIT_STRING_INIT_LUA
+        default "node.LFS.get('init')()" if LUA_INIT_STRING_INIT_LFS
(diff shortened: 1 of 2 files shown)


tomsci:
@jmattsson ping, is this something you'd want to take onto the dev-esp32-idf4 branch?

jmattsson:
At first glance this looks fine. I'm curious as to what your use case is though? When I did the initial port I considered whether LUA_INIT_STRING should go in Kconfig, but I couldn't think of a scenario where it would be useful.

tomsci:
I tend to have all my Lua code in LFS as a separate partition, not embedded in the ROM image, so I can flash it independently for faster testing of Lua-only changes. Which means I also hit all the woes of absolute addressing when generating the LFS, but that's a separate issue! I used to put a bootstrap init.lua on SPIFFS which loaded LFS but eventually ditched that to simplify device setup (and because I wasn't seeing any benefits to having it), which I did by overriding LUA_INIT_STRING in MORE_CFLAGS. So now MORE_CFLAGS is no more, I needed some other way to do it. I used to have more Lua code on SPIFFS (especially when testing) but again eventually stopped doing that and moved everything to LFS to make the flashing process a bit more streamlined (and to avoid accidentally forgetting to update one Lua file on SPIFFS, etc). I also found that loading Lua files from SPIFFS to take significantly longer than from LFS (not surprising, really), which was another factor in moving to a LFS-only setup, because it was materially impacting boot up time and I had a scenario where speedy boot from deep sleep was important.

Also at the moment I'm working with an esp32s2 with no uart and only [my hacks](https://github.com/nodemcu/nodemcu-firmware/issues/3526#issuecomment-1279974405) to get the native USB to work, which means nodemcu-uploader doesn't work, meaning I can't easily deploy an init.lua to the SPIFFS even if I wanted to.

I've never needed to set the init string to anything other than "init from SPIFFS" or "init from LFS", but I figured I'd leave the option open in case there was anyone out there already using something whacky for LUA_INIT_STRING.

jmattsson:
Fair enough! I'll see if I can take this for a spin and reassure myself that everything including embedded LFS still works :)
I mean, it _looks_ like it should, but I've screwed up recent code reviews here so I'm trying to be more diligent.

jmattsson (approved):

jmattsson:
I didn't find anything broken, so it's in! Thanks @tomsci!

PR #3560, opened 2022-12-04
Don't error from node.sleep() when a sleep reject occurs

tomsci:

In IDFv4 there's a new return value from esp_light_sleep_start which needs handling in node.sleep(). It's not really documented as far as I can see, but in the IDF source it's termed a "sleep reject" meaning (I think) that the conditions for wakeup already apply so the board doesn't actually ever go to sleep. You see it if you try to sleep with a GPIO low wakeup, and the GPIO is already low, for example. We really don't care about this nuance (compared to sleeping then immediately waking up), so the change is to just ignore the (undocumented) return value from esp_light_sleep_start and treat it the same as slept-then-woke-up, rather than erroring.

Now I think about it some more writing this PR, given it's not documented, maybe we should just return a boolean from node.sleep() indicating whether the sleep succeeded or not. Never error, don't make assumptions about the return value other than zero/nonzero, and get rid of the ESP_ERR_INVALID_STATE check which doesn't appear to actually be used any more?

### DIFF
--- a/components/modules/node.c
+++ b/components/modules/node.c
@@ -282,6 +282,13 @@ static int node_sleep (lua_State *L)
   int err = esp_light_sleep_start();
   if (err == ESP_ERR_INVALID_STATE) {
     return luaL_error(L, "WiFi and BT must be stopped before sleeping");
+  } else if (err == 1) {
+    // Not documented, and not a esp_err_t, but '1' here means a sleep reject
+    // occurred. Usually because a GPIO is already in a state that would cause
+    // an immediate wakeup (in IDFv3 this situation would just cause an
+    // immediate wakeup and return 0). This isn't really an error condition and
+    // we should just treat it as OK. esp_sleep_get_wakeup_cause appears to
+    // still return the correct value when this happens.
   } else if (err) {
     return luaL_error(L, "Error %d returned from esp_light_sleep_start()", err);
   }


components/modules/node.c
```diff
@@ -282,6 +282,13 @@ static int node_sleep (lua_State *L)
   int err = esp_light_sleep_start();
   if (err == ESP_ERR_INVALID_STATE) {
     return luaL_error(L, "WiFi and BT must be stopped before sleeping");
+  } else if (err == 1) {
```
pjsg, line 285: Can you use ESP_ERR_SLEEP_REJECT here? This return code is documented. There is also ESP_ERR_SLEEP_TOO_SHORT_SLEEP_DURATION which maybe should be handled in the same way. 

It also isn't clear what the right thing to do is -- it might be surprising to some developers that `sleep` might not sleep.

jmattsson:
While I like not having errors thrown needlessly, I don't see how we could find out the correct return value in the reject case?

pjsg:
In some sense, the two error cases are 'system didn't sleep because the timeout happened or the wakeup condition was already satisfied'. In some senses this is the same as 'it timed out, or the wakeup condition was triggered'

jmattsson:
We currently return the wakeup cause from this function, but I guess we could use `ESP_SLEEP_WAKEUP_UNDEFINED` for the sleep-rejected case? Not sure whether that's cleaner that the current behaviour though?

PR #3563, opened 2022-12-14
ESP32: Support for external components & modules

jmattsson:
Addresses comments in #3397

With the switch to IDFv4, the old `extmod` support became defunct, and we've lacked an ability to easily have external modules. This PR builds on modifications used at my $work, where we do have that need. The functional change is simply taking `EXTRA_COMPONENT_DIRS` from the environment and injecting it into the IDF, letting it do it's thing (since extra component support is already baked into it). Additionally I've included a template here as a quick-start, and added some basic documentation to hopefully give people the information needed to get off the ground.

I also fixed up some old esp866'isms that I found in the nodemcu-pil.md documentation.

@tomsci this is for you. @jpeletier you may want to have a look at this; it is not as neat as your `extmod` stuff, but on the other hand it leaves the heavy lifting to the IDF and is likely to stay supported longer.

Feedback and improvements welcome.


extcomp-template/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  set(mod_opt "CONFIG_NODEMCU_CMODULE_${module_ucase}")
+  if (${${mod_opt}})
+    list(APPEND extmods_enabled ${module_ucase})
+  endif()
+endforeach()
+message("Including the following modules: ${extmods_enabled}")
+
+foreach(mod ${extmods_enabled})
+  target_link_libraries(${COMPONENT_LIB} "-u ${mod}_module_selected1")
+endforeach()
```
tomsci: Could we abstract this out into a `extmod_declare_module(modname)` cmake helper function? The way the `modname_module_selected1` macro name is derived from the source file names has caught me out before and it'd be nice not to have to expose this to external modules.

jmattsson: Maybe? I have to admit that I don't know how scoping works across different components in the IDF and whether it's possible to expose a function in one for use in another one. I'm pretty handy with a regular Makefile, but CMake it an entirely different kettle of fish that I'm not comfortable with yet.

Looking at the way I've implemented this, I'm also wondering whether it's possible to instead iterate over the sdkconfig variables that start with `CONFIG_NODEMCU_CMODULE_` instead of source files...

tomsci:
Overall I think the approach is a good one 👍  Do you have any suggestion for how to deal with having multiple EXTRA_COMPONENT_DIRS set? I have two separate external modules to include (each is a separate git repo) each with a separate Kconfig file, and as I've initially set them up based on the example, I have 2 separate "External modules" menu entries under "Component config". Which works, but is inelegant. I guess maybe don't put a `menu "External modules"` in my Kconfigs and let those 2 modules' config options both be directly under "Component config"?

jmattsson:
Hmm, that is a good point. I forgot that Kconfig doesn't merge menus with the same name. Right now I can't think of an elegant way around it (that isn't fighting against the IDF structure lots). What you suggest is probably the least ugly option.

extcomp-template/CMakeLists.txt
tomsci: I had a go at making a `extmod_declare_module()` helper function here https://github.com/jmattsson/nodemcu-firmware/compare/extcomp...tomsci:nodemcu-firmware:tomsci-extcomp and I think it simplifies things nicely. I tested it with the external modules I use (eg https://github.com/tomsci/nodemcu-irpwm/tree/tomsci_extmodv2) and it seems to work, enabling and disabling them in the menuconfig correctly adds and removes symbols from the resulting ROM image. What do you think? (I experimented with also passing in module name separately from config name in the same way the `NODEMCU_MODULE()` macro does, but ultimately decided that it isn't needed, and probably won't ever be needed in the future either, so there wasn't any benefit to doing so).

> I'm also wondering whether it's possible to instead iterate over the sdkconfig variables that start with CONFIG_NODEMCU_CMODULE_ instead of source files...

It seems like sdkconfig.cmake does declare a variable `CONFIGS_LIST` which includes every sdkconfig variable name, so you could iterate that and match for the names of interest. For the built-in modules though I don't think it's that much of a problem, and maybe not worth the effort? Something to keep in mind for the future maybe, if the modules stuff ever needs another rework?

jmattsson: Okay, that definitely looks cleaner! I'd say the function should probably go into a `project_include.cmake` file in the `components/modules/` directory though, as it relates to module building.

I wonder if it would work to do the filtering of the `CONFIGS_LIST` once in the regular module directory, and then (if necessary) have external modules just depend on "modules"; whether that would get the `-u XYZ_module_selected1` onto the linker command line in the right place...

jmattsson:
Alright, @tomsci I've pushed an updated version here, heavily based on your suggestions. After thinking it through, I agree that trying to use the same mechanism for the NodeMCU `modules` directory and external modules is not worth it (and possibly even a bad idea).

I've added two functions - `extmod_register_conditional(MODNAME)` and `extmod_register_unconditional(MODNAME)`. The former for when there is a Kconfig option for toggling the module on/off, and the latter for when there isn't. The mere action of listing the module dir in  `EXTRA_COMPONENT_DIRS` could be considered enough for some some people to want a module automatically enabled, so it seemed worth providing the option.

jmattsson:
Merging in the absence of further feedback. Happy to refine later if needed.

PR #3568, opened 2023-02-08
Add support for using multiple memory slots in rmt setup

### LINKED ISSUE #3567: The rmt peripheral on ESP32 only can capture 64 pulses
### Expected behavior

When you capture signals with the `rmt` module, you ought to be able to get more than 64 pulses.

### Actual behavior

It is limited to 64.

### Root Cause

It turns out that this is the default when you use the rmt driver from the idf. It is possible to configure more memory to be used by the hardware device.

pjsg:
Fixes #3567 .

I added the slots parameter to configure the amount of memory that is used by the sender or receiver. This can be up to 8 on the regular ESP32, but is smaller on other platform. An error will be thrown if the request cannot be honored.


components/modules/rmt.c
```diff
@@ -112,6 +112,13 @@ static int lrmt_txsetup(lua_State *L) {
       config.flags |= RMT_CHANNEL_FLAGS_INVERT_SIG;
     }
     lua_pop(L, 1);
+
+    lua_getfield(L, 3, "slots");
+    int slots = lua_tointeger(L, -1);
```
jmattsson, line 117: Won't this raise an error if `slots` is `nil`? Consider using `luaL_optinteger()`

components/modules/rmt.c
```diff
@@ -155,6 +162,13 @@ static int lrmt_rxsetup(lua_State *L) {
       config.rx_config.idle_threshold = threshold;
     }
     lua_pop(L, 1);
+
+    lua_getfield(L, 3, "slots");
+    int slots = lua_tointeger(L, -1);
```
jmattsson, line 167: Same comment as above - this doesn't look like slots is optional?

jmattsson (changes requested):
Other than the comments around optionality, looks good to me 👍

components/modules/rmt.c
pjsg: lua_tointeger returns 0 if the value cannot be converted to an integer and this is checked in the next line

components/modules/rmt.c
pjsg: lua_tointeger returns 0 if the value cannot be converted to an integer and this is checked in the next line

jmattsson (approved):

PR #3600, opened 2023-05-01
Fix i2c.setup() for pins >= 32

tomsci:

`pin_bit_mask` in i2c_sw_master.c is a 64bit value and the current code shifts an `int` sized `1`, so doesn't work for any pin numbers >= 32 because it's only shifting a 32-bit value. The upshot is software i2c doesn't work (on the default pins) on esp32s2 ~~and later~~ and probably some others too.

This change fixes that, by using a 64-bit `1` meaning the `pin_bit_mask` gets correctly set even for pin numbers above 32.

Tested on adafruit feather esp32s2 using pins 41 and 42 (Sda and Scl)

### DIFF
--- a/components/driver_i2c/i2c_sw_master.c
+++ b/components/driver_i2c/i2c_sw_master.c
@@ -112,7 +112,7 @@ void i2c_sw_master_gpio_init(uint8_t sda, uint8_t scl)
 
   gpio_config_t cfg;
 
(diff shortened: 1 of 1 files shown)


tomsci:
Any chance you could look at this one while you're here @jmattsson? :D

jmattsson:
Yep! Thanks for the fix!

PR #3601, opened 2023-05-01
Fix ws2812 on later esp32 variants

tomsci:

This PR is a refactor of ws2812 module to get it to work on esp32s2 and later variants, and also to remove the (seemingly) fragile dependencies on RMT register layouts, since it seems like those change on every new SoC. Instead of using the RMT registers directly with a custom ISR, use only the published RMT APIs and the default ISR.

The specific thing that was broken on the S2 was that the threshold event bits start at bit 12 of the ISR status register, not bit 24 as the old code assumed, and I think it would probably be broken on the S3 as well because the memory buffers there are only 48 words not 64. I couldn't get the custom ISR to work right even after fixing those issues but I didn't bother digging to the bottom of that because it was apparent by then that a refactor was the better approach.

Instead use `rmt_write_sample` and `rmt_translator_init` which mean we don't have to override the default RMT ISR (something which the docs say to avoid doing, and I can see why) and still get all the same buffer refilling goodness. This code should be much more resilient against future esp32 SoC changes because it doesn't rely on any of the SoC-specific details, except for a couple of assumptions about the size of the buffers which seem unlikely to ever be broken, but just in case I've put some `_Static_assert`s in that will explode if those assumptions ever do stop being true.

I haven't been able to thoroughly end-to-end test the complex case, where the ISR fires to repeatedly refill the buffers, because I don't own any large strips of neopixels (which would necessitate the larger buffers). What I have tested is the single LED case (using the single neopixel built-in to the adafruit esp32-s2 feather), which now works, plus I've manually put debug tracing in and tested some longer inputs, and from the logging it looked like the right stuff was happening.

If anyone has an esp32s2 and some neopixel strips and can test this, that'd be great. Otherwise I guess this might have to wait until I cave and buy some myself!

@jmattsson fyi, I think you touched ws2812.c last 😃

### DIFF
--- a/components/platform/platform_rmt.c
+++ b/components/platform/platform_rmt.c
@@ -24,19 +24,19 @@ static bool rmt_channel_check( uint8_t channel, uint8_t num_mem )
   return rmt_channel_check( channel-1, num_mem-1);
 }
 
-#if defined(CONFIG_IDF_TARGET_ESP32C3)
 int platform_rmt_allocate( uint8_t num_mem, rmt_mode_t mode )
-#else
-int platform_rmt_allocate( uint8_t num_mem, rmt_mode_t mode __attribute__((unused)))
-#endif
 {
   int channel;
   int alloc_min;
   int alloc_max;
   uint8_t tag = 1;
 
-#if defined(CONFIG_IDF_TARGET_ESP32C3)
-  /* The ESP32-C3 is limited to TX on channel 0-1 and RX on channel 2-3. */
+#if SOC_RMT_TX_CANDIDATES_PER_GROUP == SOC_RMT_CHANNELS_PER_GROUP
+  alloc_min = 0;
+  alloc_max = RMT_CHANNEL_MAX - 1;
+#else
+  /* On platforms where channels cannot do both TX and RX, the TX ones always
+     seem to start at index 0, and are then followed by the RX channels. */
   if( mode==RMT_MODE_TX ) {
     alloc_min = 0;
     alloc_max = SOC_RMT_TX_CANDIDATES_PER_GROUP - 1;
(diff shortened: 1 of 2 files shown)


jmattsson (approved):
Oooh, this looks good! I don't own any ws2812 devices myself, I'm just the person who ends up crying whenever I update the IDF and something RMT related ends up broken again, so I can't do any additional testing. The code _looks_ good though.

I'm inclined to merge this even without having stress tested it fully, considering the uncertain state the current version is in anyway. Moving away from direct RMT access and using the supported API sounds like a good move!

Any chance of getting a similar PR for the IDFv5 branch? Pretty please? 🥺

tomsci:
> Any chance of getting a similar PR for the IDFv5 branch?

I'm maxed out fixing issues with esp32s2 and esp32s3 support on idfv4 at the moment, sorry. But I'll take a look if I find the time.

Will the idfv5 branch be caught up from `dev-esp32`, or is the expectation that they'll essentially fork until the current `dev-esp32` branch is retired? ie what happens to bugfixes etc that get merged to `dev-esp32` now?

jmattsson:
Very good questions. Too many fixes, and there'll be merge issues for sure. I was hoping to not have to sit on v5 for too long, but I've also had lukewarm response to my call for testing.

Like you, I'm short on available time to keep on top of everything myself :-/

tomsci:
> I'm inclined to merge this even without having stress tested it fully, considering the uncertain state the current version is in anyway. Moving away from direct RMT access and using the supported API sounds like a good move!

@jmattsson I've now tested this with animating a 144-LED strip on both esp32 and esp32s3 so I'm pretty happy the buffer filling logic is solid. I also found an issue in platform_rmt_allocate on the S3 (I was testing on S2 before...) so I've made a similar fix there to not rely on ifdefing for a specific silicon variant and instead infer the platform's capabilities based on the definitions of the various SOC_ macros.

jmattsson:
@tomsci Nice! I'm happy to merge this if you think it's ready now? I'll take your word for the C3 ifdef change to the SOC_ macros :)

tomsci:
On C3 SOC_RMT_TX_CANDIDATES_PER_GROUP=2 and SOC_RMT_CHANNELS_PER_GROUP=4 so yeah the code will be doing the same as before on that platform. The difference should only be on non-C3 platforms where TX_CANDIDATES_PER_GROUP isn't the same as CHANNELS_PER_GROUP, which on right now is the S3 and H2 but importantly, this change should not need revisiting when new platforms are introduced, providing espressif don't rename the SOC_ macros or redefine the order of TX and RX channels, of course.

> happy to merge this if you think it's ready now

Please do, no reason not to that I can think of!

jmattsson:
Thank you kindly!

PR #3623, opened 2024-01-30
Upgrade to IDF v5.1.2

jmattsson:
Just an IDF upgrade to the latest release.
Minor code tweaks to deal with deprecated symbols and `const` changes, but nothing particularly exciting.


pjsg (approved):
Looks good to me

PR #3629, opened 2024-02-04
Update rmt translator

### LINKED ISSUE #3628: Support the RGB LED used on the ESP32-C3-DevKitM-1
### Missing feature
The ESP32-C3-DevKitM-1 has an SK6812 RGB LED which does not respond to the WS2812 module.
The timings in the WS2812 module are currently hard-coded. If they could be defined by the user somehow, it would be possible to use the module also for other RGB LEDs.

This continues the discussion in #3509 .

Please note that I do not have any evil plans to rename the module.

### Justification
It would be great if the peripherals on a widely available ESP dev kit could be used right away with nodemcu.

### Workarounds
No workarounds, but I have a patch for the current dev-esp32 branch.

docbacardi:
Fixes #3628 .

This set of patches adds optional parameters to the WS2812 module to set the duration of the reset signal and the duration for the high and low phase of the 0 and 1 symbols. This allows to control many other RGB LEDs besides the WS2812.
The main intention of this patch is to control the RGB LED of the ESP32-C3 dev kit.

### DIFF
--- a/components/platform/ws2812.c
+++ b/components/platform/ws2812.c
@@ -35,19 +35,11 @@
 
 #undef WS2812_DEBUG
 
-// If either of these fails, the reset logic in ws2812_sample_to_rmt will need revisiting.
-_Static_assert(SOC_RMT_MEM_WORDS_PER_CHANNEL % 8 == 0,
-  "SOC_RMT_MEM_WORDS_PER_CHANNEL is assumed to be a multiple of 8");
-_Static_assert(SOC_RMT_MEM_WORDS_PER_CHANNEL >= 16,
-  "SOC_RMT_MEM_WORDS_PER_CHANNEL is assumed to be >= 16");
-
 // divider to generate 100ns base period from 80MHz APB clock
 #define WS2812_CLKDIV (100 * 80 /1000)
-// bit H & L durations in multiples of 100ns
-#define WS2812_DURATION_RESET (50000 / 100)
 
-// This is one eighth of 512 * 100ns, ie in total a bit above the requisite 50us
-const rmt_item32_t ws2812_rmt_reset = { .level0 = 0, .duration0 = 32, .level1 = 0, .duration1 = 32 };
+// This is 512 * 100ns, which is a bit above the requisite 50us
+const rmt_item32_t ws2812_rmt_reset = { .level0 = 0, .duration0 = 256, .level1 = 0, .duration1 = 256 };
 
 // descriptor for a ws2812 chain
 typedef struct {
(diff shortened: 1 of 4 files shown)


components/platform/ws2812.c
```diff
@@ -32,112 +32,180 @@
...
+      {
+        ucBitPos = ptContext->bitpos;
+
+        /* Each bit of the input data is converted into one RMT item. */
+
+        pucData = (const uint8_t*)src;
+        /* Get the current byte. */
+        ucData = pucData[cnt_in] << ucBitPos;
+
+        while( cnt_in<src_size && cnt_out<wanted_num )
```
pjsg, line 93: I suspect that there is an issue here if the `cnt_out<wanted_num` fails when not on a byte boundary. The original code validated that the entire byte would fit before it tried to encode it. There isn't a way to signal back to the rmt library that only part of a byte has been encoded.

docbacardi: There is a good way to signal back that a byte was not completely processed: do not increase ```translated_size```.
The output parameter ```translated_size``` of the translator function reports how many bytes were processed. If a byte was not completely processed, it should not be considered in ```translated_size```. Additional information for the decoding process like the bit position can be stored in the translator context. Please note that the number of consumed raw bytes and the number of produced ```rmt_item32_t``` elements are not forced to have a fixed relationship of 1:8. The documentation has a note that it should be like this, but it is not a strict requirement (see [here](https://docs.espressif.com/projects/esp-idf/en/v4.3/esp32/api-reference/peripherals/rmt.html#_CPPv415sample_to_rmt_t)). The current routine inserts a reset signal with 8 ```rmt_item32_t``` elements without consuming a single raw byte for it (see [here](https://github.com/nodemcu/nodemcu-firmware/blob/10e378394861b8261eac4366658dad069aa12f63/components/platform/ws2812.c#L100)), and it is working apparently well.

I'm afraid my commit messages did not give a good explanation why I would like to change an implementation which is doing a good job. Playing around with the translator callback I found out that the transmission of RMT pulses stops if the ```wanted_num``` parameter is not met.

Example: I want to send 12 raw bytes which would give 96 ```rmt_item32_t``` elements after translation. The first call to the translator function has ```wanted_num``` set to 48, but my test code decides to provide only 40. The transmission stops after these pulses. The translator function is never called for the remaining 56 bytes of raw data. This was tested with an ESP32-C3.

Fortunately all observed values for the ```wanted_num``` parameter were multiples of 8. But I did not find a guarantee for it. That's why I changed the translator function. Am I too paranoid?

pjsg: Thanks -- I missed the saving of the bitpos in the context.

docs/modules/ws2812.md
```diff
@@ -44,6 +52,10 @@ ws2812.write({pin =  4, data = string.char(255, 0, 0, 255, 0, 0)},
              {pin = 14, data = string.char(0, 255, 0, 0, 255, 0)}) -- turn the two first RGB leds to green on the first strip and red on the second strip
 ```
 
+```lua
+ws2812.write({pin = 8, reset = 800, t0h = 3, t0l = 9, t1h = 6, t1l = 6, data = string.char(1, 0, 0)}) -- turn the SK6812 RGB led on the ESP32-C3-DevKitM-1 to green
```
pjsg: Does this work -- I would have thought that using 255 as the intensity rather than 1. Also, what is the color ordering for the SK6812? You set the first value to get green, which implies that it isn't RGB.

docbacardi: The SK6812 is a very bright LED. A value of 1 is clearly visible. Setting it to 255 was painful for my eyes.

You are correct, the color ordering is really green - red - blue, see end of page 6 in the data sheet [here](https://cdn-shop.adafruit.com/product-files/1138/SK6812+LED+datasheet+.pdf). Thanks, the patch is updated.

However the WS2812 has the same color ordering (see [here](https://cdn-shop.adafruit.com/datasheets/WS2812.pdf), page 5). The APA104 is the only LED in the list using RGB color ordering (see [here](https://www.ledlightinghut.com/files/APA104%20Datasheet.pdf), page5). The current [documentation](https://github.com/nodemcu/nodemcu-firmware/blob/10e378394861b8261eac4366658dad069aa12f63/docs/modules/ws2812.md?plain=1#L14) uses the term "RGB" or "RGBW" for all of them:
```
Send data to up to 8 led strip using its native format which is generally Green,Red,Blue for RGB strips and Green,Red,Blue,White for RGBW strips.
```

If you agree that this is a bit confusing I will open a new ticket for it.

pjsg: Yeah -- that makes sense

PR #3638, opened 2024-02-18
Fifo fixes

HHHartmann:
Fixes problems I had during development.

_To be completed below: Description of and rationale behind this PR._

This contains fixes for problems I found while using fifo.

### DIFF
--- a/lua_modules/fifo/fifo.lua
+++ b/lua_modules/fifo/fifo.lua
@@ -1,6 +1,6 @@
 -- A generic fifo module.  See docs/lua-modules/fifo.md for use examples.
 
-local tr, ti = table.remove, table.insert
+local tableRemove, tableInsert = table.remove, table.insert
 
 -- Remove an element and pass it to k, together with a boolean indicating that
 -- this is the last element in the queue; if that returns a value, leave that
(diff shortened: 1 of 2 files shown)


lua_modules/fifo/fifo.lua
```diff
@@ -17,29 +17,33 @@ local tr, ti = table.remove, table.insert
...
+  if #q > 0 then
+    local new, again = k(q[1], #q == 1)
+    if new == nil then
+      tableRemove(q, 1)
+    else
+      q[1] = new
+    end
+    if again then
+      return dequeue(q, k)
+    end -- note tail call
```
nwf, line 30: I guess I'd never had reason to both replace the top element of the fifo and want to be called back on the thing I just pushed.  That does seem a little odd to want, but I agree that this patch is probably a better thing to do if the `k`onsumer asks for it.

lua_modules/fifo/fifosock.lua
```diff
@@ -58,7 +58,15 @@ local function wrap(sock)
     -- Either that was a function or we've hit our coalescing limit or
     -- we didn't ship above.  Ship now, if there's something to ship.
     if s ~= nil then
-      if sslan == 0 then sock:send(s) else sock:send(ssla .. s) end
+      if sslan == 0 then
+        if #s > 0 then
```
nwf, line 62: Sure, avoiding `send`-ing 0-length things seems like a good idea.  And while `return`-ing `ns or nil, true` below means we won't call `gc()` this time, we will probably do so "soon" since we're about to be called back by the `fifo` logic.

nwf (approved):
LGTM.  Some commentary just to make sure I understand the changes.

marcelstoer:
@HHHartmann as it's been approved, do you want to merge this? Or do you still plan to add changes given @nwf comments?

lua_modules/fifo/fifo.lua
HHHartmann: I did these changes around two years ago, but the code runs better with it.
IIRC it was that I forced code to send a whole file into the httpserver.lua module including giving the size non chunked but at the beginning and so had some uh, well "interesting" usings of the whole thing.
Should have changed the api instead of hiding it behind specific parametrs, but well.
It just stopped sending in certain cases and I tracked it down to having to return what I just pushed.

HHHartmann:
Nope, not planning to add changes.

PR #3639, opened 2024-02-24
Fix auto create release.yml

HHHartmann:
Fixes #<GITHUB_ISSUE_NUMBER>.

for the prepare release note workflow to work it has to run on specific os versions


marcelstoer:
I'm not sure if this fix is correct or if not fixing it would cause any issues. However, fix at least brings this on par with what we have in the build script at https://github.com/nodemcu/nodemcu-firmware/blob/dev/.github/workflows/build.yml.

marcelstoer (approved):

PR #3651, opened 2024-05-11
node.chipid() returns "0x_unknown" for ESP32xx chips

serg3295:

The reason for this change is that the [nodemcu-tool](https://github.com/serg3295/NodeMCU-Tool/tree/dev-esp32) utility expects the `node.chipid()` command to exist in the Lua environment.
However, ESP32c6|s2|h2 chips do not have a chipid, so `node.chipid() `function will return `0x_unknown` for these chips.

### DIFF
--- a/components/modules/node.c
+++ b/components/modules/node.c
@@ -155,10 +155,10 @@ static int node_bootreason( lua_State *L)
   return 2;
 }
 
-#if defined(CONFIG_IDF_TARGET_ESP32)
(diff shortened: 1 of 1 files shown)


marcelstoer:
> ESP32c6|s2|h2 chips do not have a chipid, so `node.chipid() `function will return `0x_unknown`

Shouldn't the docs in `node.md` mention this? There's a - now outdated - comment in #3652.

PR #3680, opened 2025-05-12
Zero out ledc timer config in newChannel

nfreya:
ledc.newChannel was broken for me on `dev-esp32`:

```
> pwm = ledc.newChannel({
>>   gpio = 4,
>>   bits = ledc.TIMER_10_BIT,
>>   mode = ledc.HIGH_SPEED,
>>   timer = ledc.TIMER_0,
>>   channel = ledc.CHANNEL_0,
>>   frequency = 1000,
>>   duty = 512,
>> })
 (421647) ledc: ledc_timer_del(589): LEDC is not initialized
Lua error: 	stdin:1: timer configuration failed code 259
stack traceback:
	[C]: in function 'ledc.newChannel'
	stdin:1: in main chunk
	[C]: in ?
	[C]: in ?
```

espressif/esp-idf#15806 indicates this is due to a (presumably) new field, zeroing out the struct fixes it.

### DIFF
--- a/components/modules/ledc.c
+++ b/components/modules/ledc.c
@@ -20,7 +20,7 @@ static int lledc_new_channel( lua_State *L )
   luaL_checktable (L, 1);
 
   /* Setup timer */
-  ledc_timer_config_t ledc_timer;
+  ledc_timer_config_t ledc_timer = {};
 
   ledc_timer.duty_resolution = opt_checkint_range (L, "bits", LEDC_TIMER_13_BIT, 0, LEDC_TIMER_BIT_MAX-1);
 


jmattsson:
Hi and thank you for reporting and the PR! 😊

Yeah it looks like there's a new `deconfigure` field that's been added. I deliberated on whether it would be better to explicitly add that, or go with the zero-initialisation approach. I like the clarity of explicitly listing what's being set, but on the other hand, zero-init is good, clean practice 👍. Plus it's hopefully more future proof in case new fields are added again.

jmattsson:
Housekeeping note: I can't get the checks to build as it picks up the config from the PR branch, not the target branch; I've manually applied the PR on top of latest dev-esp32 on a personal branch, and had CI run that. Everything's happy there.

Issue #2970, opened 2019-11-24
Issues with http module when accessing an HTTPS API endpoint.


dankirkham:
### Expected behavior
I was able to run a successful GET request exactly once. It hasn't worked again. I am getting two error codes, `E:M 544` and `46`, and cannot determine what they mean.

### Actual behavior
```
E:M 544
HTTP client: Disconnected with error: 46
HTTP client: Connection timeout
```

### Test code
```Lua
local http = require("http")

API_APP = "redacted"
API_ENDPOINT = "https://graph-na04-useast2.api.smartthings.com/api/smartapps/installations/" .. API_APP
API_TOKEN = "redacted"

headers = "Authorization: Bearer " .. API_TOKEN .. "\r\n"
http.get(API_ENDPOINT .. "/switches", headers, function (code, data)
    print(data)
end)
```

### NodeMCU version
```
NodeMCU 3.0.0.0 built on nodemcu-build.com provided by frightanic.com
	branch: master
	commit: 310faf7fcc9130a296f7f17021d48c6d717f5fb6
	release: 3.0-master_20190907
	release DTS: 201909070945
	SSL: true
	build type: integer
	LFS: 0x0
	modules: file,gpio,http,net,node,sjson,tmr,uart,wifi,tls
 build 2019-11-24 14:43 powered by Lua 5.1.4 on SDK 3.0.1-dev(fce080e)
```

### Hardware
NodeMCU v2 (HiLetgo on Amazon)

marcelstoer:
SSL and HTTP are an endless source of frustration. If look through the (closed) issues here you'll find numerous examples of constellations that don't work. It's either memory constraints or lack of support for certain TLS features (ciphers, SNI, etc.).

You can try using the lower-level TLS as per https://nodemcu.readthedocs.io/en/latest/modules/tls/#tlssocketon but there is some doubt you'll have better luck. According to https://www.ssllabs.com/ssltest/analyze.html?d=smartthings.com&s=54.183.45.39 smartthings.com uses SNI.

Issue #1484, opened 2016-09-04
Support TLS SNI or handle error gracefully


marcelstoer:
### Expected behavior
Either support [TLS SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) or at least handle errors gracefully.

Someone wanted to provide a fix for axTLS at https://github.com/dsheets/axtls/issues/2 but it's still open. Can our firmware at least handle the error?

### Actual behavior
Crash.

> client handshake start.
> client handshake ok!
> client's data invalid protocol
> Error: invalid protocol message
> Fatal exception 28(LoadProhibitedCause):
> epc1=0x4000df1b, epc2=0x00000000, epc3=0x00000000, excvaddr=0x01000000, depc=0x00000000
> 
>  ets Jan  8 2013,rst cause:2, boot mode:(3,6)

### Test code
```Lua
http.get("https://nodemcu-build.com/", nil, function(code, data)
    if (code < 0) then
      print("HTTP request failed")
    else
      print(code, data)
    end
  end)
```
Other than its use of SNI the certificate for the domain is fine: https://www.ssllabs.com/ssltest/analyze.html?d=nodemcu-build.com

### NodeMCU version
Latest `dev`.

### Hardware
NodeMCU devkit V2/1.0.

jmattsson:
I'm getting the definite impression we need to get some more focus on the shift to mbedTLS. I'd do it if I had time for it, but at the moment I do not.

marcelstoer:
I'll close this. https://github.com/nodemcu/nodemcu-firmware/issues/1485#issuecomment-246176969 showed that the cause is unrelated to SNI. https://nodemcu-build.com/jumbotron-narrow.css works just fine.

PR #3685, opened 2025-08-28
TLS: add SNI support for secure connections

caiohamamura:
Probably fixes most TLS (https) issues #2970 #1707 #1484.

This will add SNI support for secure connections by providing these modifications:

 - [`ssl_options`](https://github.com/caiohamamura/nodemcu-firmware/blob/547ed87695839d8024ec5077fcf3126568fa2405/app/include/sys/espconn_mbedtls.h#L97): Add a `char *hostname` field
 - [`espconn_secure_set_hostname`](https://github.com/caiohamamura/nodemcu-firmware/blob/547ed87695839d8024ec5077fcf3126568fa2405/app/mbedtls/app/espconn_secure.c#L185C8-L185C50): function to provides an interface to set the `ssl_client_options.hostname` . 
 - Within `httpclient.c` [`http_dns_callback`](https://github.com/caiohamamura/nodemcu-firmware/blob/547ed87695839d8024ec5077fcf3126568fa2405/app/http/httpclient.c#L495) function: before connecting through `espconn_secure_connect` it will set [`espconn_secure_set_hostname( hostname )`](https://github.com/caiohamamura/nodemcu-firmware/blob/547ed87695839d8024ec5077fcf3126568fa2405/app/http/httpclient.c#L536)
 - During [`mbedtls_msg_config`](https://github.com/caiohamamura/nodemcu-firmware/blob/547ed87695839d8024ec5077fcf3126568fa2405/app/mbedtls/app/espconn_mbedtls.c#L617): it will now call `mbedtls_ssl_set_hostname` given `ssl_client_options.hostname` is provided, which will then address SNI.

### DIFF
--- a/app/mbedtls/app/espconn_mbedtls.c
+++ b/app/mbedtls/app/espconn_mbedtls.c
@@ -626,6 +626,20 @@ static bool mbedtls_msg_config(mbedtls_msg *msg)
 	ret = mbedtls_ssl_setup(&msg->ssl, &msg->conf);
 	lwIP_REQUIRE_NOERROR(ret, exit);
 
+	/* Add hostname for SNI support */
+	if (ssl_client_options.hostname != NULL) {
+		os_printf("Configuring SNI for hostname: %s\n", ssl_client_options.hostname);
+        ret = mbedtls_ssl_set_hostname(&msg->ssl, ssl_client_options.hostname);
+		if (ret == 0) {
+     	   os_printf("SNI hostname configured successfully\n");
+    	} else {
+	        os_printf("SNI hostname configuration failed: -0x%x\n", -ret);
+	    }
+        lwIP_REQUIRE_NOERROR(ret, exit);
+    } else {
+		os_printf("Hostname was not set!\n");
+	}
+
 	/*Initialize the RNG and the session data*/
 	ret = mbedtls_ctr_drbg_seed(&msg->ctr_drbg, mbedtls_entropy_func, &msg->entropy, "client", 6);
 	lwIP_REQUIRE_NOERROR(ret, exit);
--- a/app/http/httpclient.c
+++ b/app/http/httpclient.c
@@ -533,6 +533,7 @@ static void ICACHE_FLASH_ATTR http_dns_callback( const char * hostname, ip_addr_
 #ifdef CLIENT_SSL_ENABLE
 		if ( req->secure )
 		{
+			espconn_secure_set_hostname( hostname );
 			espconn_secure_connect( conn );
 		}
 		else
--- a/app/include/lwip/app/espconn.h
+++ b/app/include/lwip/app/espconn.h
@@ -516,6 +516,15 @@ extern sint8 espconn_secure_send(struct espconn *espconn, uint8 *psent, uint16 l
 extern sint8 espconn_secure_sent(struct espconn *espconn, uint8 *psent, uint16 length);
 
 /******************************************************************************
+ * FunctionName : espconn_secure_set_hostname
+ * Description  : set hostname for SNI (Server Name Indication) support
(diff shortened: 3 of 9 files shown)


caiohamamura:
I just pushed commits for extending SNI support for MMQT and Websockets 4bf7acdcd7b3494992ca9a393392973274528563

marcelstoer:
Sorry, please rebase on `dev` - my bad.

caiohamamura:
Rebase done!

app/mbedtls/app/espconn_mbedtls.c
```diff
@@ -626,6 +626,20 @@ static bool mbedtls_msg_config(mbedtls_msg *msg)
 	ret = mbedtls_ssl_setup(&msg->ssl, &msg->conf);
 	lwIP_REQUIRE_NOERROR(ret, exit);
 
+	/* Add hostname for SNI support */
+	if (ssl_client_options.hostname != NULL) {
+		os_printf("Configuring SNI for hostname: %s\n", ssl_client_options.hostname);
```
jmattsson, line 631: Is there a way to disable these logging statements in this function? If they're always enabled then they could interfere with normal operations if the console uart is also used to talk to a peripheral.
(I'll add that I'm not across the 8266 side these days, so if these already require an ifdef to be enabled somewhere to have an effect, please ignore me 🙃).

jmattsson (commented):
This looks like a great addition!
Just one comment around some debug prints, but otherwise a big thumbs up from me 👍 (with the caveat I haven't looked at the 8266 branch in years 😅)

app/mbedtls/app/espconn_mbedtls.c
caiohamamura: Actually I don't know exactly why, but it seems that os_printf is being controlled by `#def DEVELOP_VERSION`. I just know it empirically because when I disable DEVELOP_VERSION nothing is printed out to Serial.

That makes sense, because there are other functions in this same file using os_printf, that's why I chose it.

## PS
Oh, and thank you @jmattsson for your support on my PR. Too bad you are not using 8266 lately. I myself almost gave up NodeMCU Lua framework for Arduino framework just because I couldn't comunicate with modern HTTPS APIs (almost a must these days). But then I couldn't just abandon such a great project and I knew this had a huge impact in NodeMCU Lua framework adoption and decided to contribute to this nice project. I know NodeMCU is not being adopted as much as ESP32 these days, but in Brazil it is still quite popular just because it is cheaper and just a few dollars already hurts us really bad here.

marcelstoer: > Actually I don't know exactly why

Isn't it because of this? https://github.com/nodemcu/nodemcu-firmware/blob/c8faff28e7e1676c7d14ece13e2cbb293860337e/app/user/user_main.c#L340-L342

`DEVELOP_VERSION` controls `NODE_DEBUG`?

caiohamamura: So is it safe, or do you think I should remove some messages there?

jmattsson:
Thanks, looks safe then! 👍

marcelstoer:
@caiohamamura This is gold, thank you!

I just went through a pile of old HTTPS issues (you linked to some of them) and tested the URLs we were testing back then. Just sending HTTP GET to them through our `http` module. The only ones that still don't work:

- https://dziadalnfpolx.cloudfront.net/
- https://clients5.google.com/pagead/drt/dn/
- https://raw.githubusercontent.com/espressif/esptool/master/MANIFEST.in

The output is always the same and I'm sure it's totally unrelated to what this PR fixes. Didn't investigate further, though.

```
HTTP client: Disconnected with error: -114
HTTP client: Connection timeout
HTTP request failed
```

PR #3687, opened 2025-09-02
Allow windows compilation using official espressif toolchain (msys2)

caiohamamura:
- [X] This PR is compliant with the [other contributing guidelines](/CONTRIBUTING.md) as well.

## Summary

This PR improves Windows compilation support for NodeMCU-Firmware by addressing missing headers, inconsistent function names, and toolchain differences between Windows and Linux.

### Makefile: 
 - Correctly handles Windows vs. Linux toolchain download and expansion.
 - Allows implicit functions.
 - Defines WIN32.
 - Defines stricmp as a macro for strcasecmp.

### cc.h:
 - Adds a check to avoid redefining BYTE_ORDER when already defined in Windows toolchain.

### socket.h:
 - Adds safeguard to avoid redefining timeval if already included.

### locale support:
 - Defines __locale_c_type_ptr function since locale/ctype support is missing in Windows toolchain.

### Headers
 - Adds missing headers required by the Windows toolchain.

## Building in windows

These fixes allow building NodeMCU-Firmware with the official windows toolchains provided by espressif: <https://docs.espressif.com/projects/esp8266-rtos-sdk/en/latest/get-started/windows-setup.html>. 

Follow these steps:

1. Download and extract the "Windows all-in-one toolchain & MSYS2": <https://dl.espressif.com/dl/esp32_win32_msys2_environment_and_toolchain-20181001.zip>
2. Open the mingw32 version (mingw32.exe)
3. Clone this fork: `git clone https://github.com/caiohamamura/nodemcu-firmware --branch=windows_compilation --depth=1 --recurse-submodules --shallow-submodules`
4. Change to directory and run make: `cd nodemcu-firmware && make`

## Motivation

Besides allowing other developers to use Windows for developing and eventually contributing to this repository, I am writing a VSCode/VSCodium extension for developing for nodemcu, allowing to manage the firmware compilation, selection of modules, flashing and compiling/loading lua code. So I need this repository to be compile ready cross-platform wise for the extension to also work cross-platform.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,19 +80,16 @@ jobs:
       with:
         submodules: true
 
-    - uses: msys2/setup-msys2@v2
-      with:
-        update: true
-        cache: true
-        install: >-
-          make
-          gcc
-          python
-          python-pip
-          unzip
+    - name: Download ESP32 MSYS2 Toolchain
+      run: |
+        Invoke-WebRequest -Uri "https://media.githubusercontent.com/media/nodemcu/espressif-sdk-archive/refs/heads/master/esp32_win32_msys2_environment_and_toolchain-20181001.zip" -OutFile "esp32_toolchain.zip"
 
+    - name: Extract Toolchain
+      run: |
+        Expand-Archive -Path "esp32_toolchain.zip" -DestinationPath "D:\" -Force
+    
     - name: Setup Python venv
-      shell: msys2 {0}
+      shell:  D:\msys32\usr\bin\bash.exe --login {0}
       run: |
         python -m venv .venv
         . .venv/bin/activate
@@ -100,7 +97,7 @@ jobs:
         pip install pyserial
 
     - name: Build firmware
-      shell: msys2 {0}
+      shell:  D:\msys32\usr\bin\bash.exe --login {0}
       run: |
         . .venv/bin/activate
 
@@ -117,13 +114,13 @@ jobs:
         fi
 
     - name: Copy required DLLs
-      shell: msys2 {0}
+      shell:  D:\msys32\usr\bin\bash.exe --login {0}
       run: |
         ldd luac.cross.exe | awk '{print $3}' | grep -vi "/c/windows" | xargs -I{} cp -v {} . || true
         ls *.dll
 
     - name: Check build items
-      shell: msys2 {0}
+      shell:  D:\msys32\usr\bin\bash.exe --login {0}
       run: |
         ls -ll bin/0x00000.bin bin/0x10000.bin luac.cross.exe
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,87 @@ jobs:
         name: luac.cross_${{ matrix.lua_ver }}_${{ matrix.numbers }}
         path: luac.cross
 
+  build_win_msys2:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - lua_ver: 51
+            numbers: float
+          - lua_ver: 51
+            numbers: integral
+          - lua_ver: 53
+            numbers: float
+          - lua_ver: 53
+            numbers: 64bit
+
+    env:
+      LUA: ${{ matrix.lua_ver }}
+      MSYSTEM: MINGW32
+      CHERE_INVOKING: 1
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+
+    - uses: msys2/setup-msys2@v2
+      with:
+        update: true
+        cache: true
+        install: >-
+          make
+          gcc
+          python
+          python-pip
+          unzip
+
+    - name: Setup Python venv
+      shell: msys2 {0}
+      run: |
+        python -m venv .venv
+        . .venv/bin/activate
+        python -m pip install --upgrade pip
+        pip install pyserial
+
+    - name: Build firmware
+      shell: msys2 {0}
+      run: |
+        . .venv/bin/activate
+
+        EXTRA_FLAGS=""
+        if [ "${{ matrix.numbers }}" = "integral" ]; then
+          EXTRA_FLAGS="-DLUA_NUMBER_INTEGRAL"
+        elif [ "${{ matrix.numbers }}" = "64bit" ]; then
+          EXTRA_FLAGS="-DLUA_NUMBER_64BITS"
+        fi
+        make EXTRA_CCFLAGS=$EXTRA_FLAGS
+        
+        if [ "${{ matrix.numbers }}" = "integral" ]; then
+          mv -f luac.cross.int.exe luac.cross.exe
+        fi
+
+    - name: Copy required DLLs
+      shell: msys2 {0}
+      run: |
+        ldd luac.cross.exe | awk '{print $3}' | grep -vi "/c/windows" | xargs -I{} cp -v {} . || true
+        ls *.dll
+
+    - name: Check build items
+      shell: msys2 {0}
+      run: |
+        ls -ll bin/0x00000.bin bin/0x10000.bin luac.cross.exe
+
+    - name: Upload luac.cross
+      if: ${{ success() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: luac.cross_${{ matrix.lua_ver }}_${{ matrix.numbers }}_win
+        path: |
+          luac.cross.exe
+          *.dll
+
 # The files in the 'msvc' folder would need to be updated to be compatible with VS 2022.
 # Sample job (that failed): https://github.com/nodemcu/nodemcu-firmware/actions/runs/17306868602/job/49131917376#step:3:153
 # => disable the Windows jobs for now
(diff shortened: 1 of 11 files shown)


caiohamamura:
In addition this will also allow to more throughly test and CI with a windows github action relying in MSYS2.

caiohamamura:
Now I added compilation and tests support for windows using the official toolchain<https://docs.espressif.com/projects/esp8266-rtos-sdk/en/latest/get-started/windows-setup.html>.

## Note: question?

Would it be better if we provided a github release for the windows toolchain just in case espressif stop giving support?

Makefile
```diff
@@ -286,21 +282,24 @@ endif
...
 sdk_pruned: sdk_extracted toolchain $(TOP_DIR)/sdk/.pruned-$(SDK_VER)
 
-ifdef GITHUB_TOOLCHAIN
-  TOOLCHAIN_ROOT := $(TOP_DIR)/tools/toolchains/esp8266-linux-x86_64-$(TOOLCHAIN_VERSION)
-
+ifdef TOOLCHAIN_URL
 toolchain: $(TOOLCHAIN_ROOT)/bin $(ESPTOOL)
 
-$(TOOLCHAIN_ROOT)/bin: $(TOP_DIR)/cache/toolchain-esp8266-$(GCCTOOLCHAIN).tar.xz
+$(TOOLCHAIN_ROOT)/bin: $(TOP_DIR)/cache/toolchain-esp8266-$(GCCTOOLCHAIN).$(TOOLCHAIN_EXT)
```
HHHartmann, line 288: This seems to be a debugging leftover

marcelstoer:
> Would it be better if we provided a github release for the windows toolchain just in case espressif stop giving support?

I support that idea. IIRC there are other tools for which (at some point but not anymore?) we kept or keep local copies.

caiohamamura:
It needs the MSYS2 for compiling, so we need both:

- MSYS2:  https://dl.espressif.com/dl/esp32_win32_msys2_environment_and_toolchain-20181001.zip
- Toolkit: https://dl.espressif.com/dl/xtensa-lx106-elf-gcc8_4_0-esp-2020r3-win32.zip

marcelstoer:
> IIRC there are other tools for which (at some point but not anymore?)...

I just remembered: https://github.com/nodemcu/espressif-sdk-archive

As those two Win tools are ~750MB I suggest we add them both to our `espressif-sdk-archive` project.

caiohamamura:
Will you create the releases there? Then if you could provide me the links to update this PR.

marcelstoer:
GitHub doesn't allow artifacts >100MB in repositories without Git LFS. Hence, that repository now uses LFS to track those two files. Their raw location is as follows:

- https://media.githubusercontent.com/media/nodemcu/espressif-sdk-archive/refs/heads/master/esp32_win32_msys2_environment_and_toolchain-20181001.zip
- https://media.githubusercontent.com/media/nodemcu/espressif-sdk-archive/refs/heads/master/xtensa-lx106-elf-gcc8_4_0-esp-2020r3-win32.zip

However, the workflow checkout action or similar allow you reference their repository location without hard-coding the raw URL.

.github/workflows/build.yml
```diff
@@ -56,95 +56,86 @@ jobs:
         path: luac.cross
 
 
-  build_luac_cross_win:
```
HHHartmann: Nice to see that we will finally have a test for the mingw build

but please do not remove the visualstudio build of the tools.
It can be built with VS and does not need mingw

same for NTest_win.

Maybe call the new sections something with cygwin like `NTest_cygwin`.

HHHartmann: ok, got it. that's the ones not working currently.
So I will have a look and re add them when I found the solution.
Will start with the work by @marcelstoer on this

marcelstoer: Yeah, it's currently a bit confusing. On `release` I disabled the Windows jobs: https://github.com/nodemcu/nodemcu-firmware/blob/release/.github/workflows/build.yml. `dev` is 1 commit behind and still has the Windows jobs. In this PR @caiohamamura kinda "hijacked" them 🤓 . Thanks for helping to sort this out.

caiohamamura: So should I leave the previous tests commented out?

caiohamamura:
> GitHub doesn't allow artifacts >100MB in repositories without Git LFS. Hence, that repository now uses LFS to track those two files. Their raw location is as follows:
> 
> * https://media.githubusercontent.com/media/nodemcu/espressif-sdk-archive/refs/heads/master/esp32_win32_msys2_environment_and_toolchain-20181001.zip
> * https://media.githubusercontent.com/media/nodemcu/espressif-sdk-archive/refs/heads/master/xtensa-lx106-elf-gcc8_4_0-esp-2020r3-win32.zip
> 
> However, the workflow checkout action or similar allow you reference their repository location without hard-coding the raw URL.

Done, I have updated the urls in both the Makefile and the github action.

.github/workflows/build.yml
HHHartmann: Maybe the best solution for now

marcelstoer: I don't remember why I thought handling it that way was a good idea (it wasn't), sorry. We could cherry-pick the `release` commit into the `dev` branch at the right position, rebase this one onto `dev` and proceed from there.
@HHHartmann feel free to execute this if you agree.

HHHartmann: ok, finally executed this. @caiohamamura please rebase this PR on the updated dev which contains the commented out windows tests.

caiohamamura:
Rebase done

.github/workflows/build.yml
```diff
@@ -59,95 +59,83 @@ jobs:
 # Sample job (that failed): https://github.com/nodemcu/nodemcu-firmware/actions/runs/17306868602/job/49131917376#step:3:153
 # => disable the Windows jobs for now
 
-  # build_luac_cross_win:
```
HHHartmann, line 62: First let me apologize for not coming back to this PR earlier.
The Idea of merging the commit of @marcelstoer was to leave the commented out part in, so it can be fix later to also build with VisualStudio. Your new build method should be separate.
Possibly with your changes concerning the header It will also be possible to move the VisualStudio Build to a newer version.

app/include/xtensa/corebits.h
HHHartmann, line 1: shouldn't these files be taken from the toolchain?
Or if the toolchain does not have the files, at least be included after the toolchain version (and then effectively be ignored due to the include guard)

caiohamamura: Hi @HHHartmann , I think this will already work, because the search order is the same order of the `-I` flags, as SDK includes are included right after the SDK_overrides, they will be the default, unless not found then it will search the other paths. As I could see from commands here, it looks like the app/include is the last PATH to be searched.

```
xtensa-lx106-elf-gcc  -I ../../sdk-overrides/include -I /home/caioh/src/nodemcu-firmware/sdk/esp_iot_sdk_v3.0-e4434aa/include -D_C_TYPES_H_ -O2 -ffunction-sections -fno-jump-tables -fdata-sections -fpack-struct=4 -Wno-error=implicit-function-declaration -Dstricmp=strcasecmp -DWIN32 -g -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals  -DESP_INIT_DATA_DEFAULT="\"/home/caioh/src/nodemcu-firmware/sdk/esp_iot_sdk_v3.0-e4434aa/bin/esp_init_data_default_v05.bin\""  -DLWIP_OPEN_SRC  -std=gnu11 -Wimplicit -I ../libc -I ../lua -I ../platform -I ../spiffs -I ../ -I ../include  -o .output/eagle/debug/obj/user_main.o -c user_main.c
```

The `-I ../include` is the least priority here when I run `make -n` to check.

.github/workflows/build.yml
caiohamamura: Done, now the commits won't mess with the commented out windows builds.

app/include/xtensa/corebits.h
HHHartmann: The files are part of the toolchain and that path is included at the end. Use make -v to see the full truth.
At least in the toolchain used for linux the files are included.

But given the current state of the project it might just be ok to have them duplicated here.

@marcelstoer what do you think? Should we accept the PR or find another place for the files and only include them for the windows build?

marcelstoer: Thanks for asking Gregor. Assuming I got this right, then IMO these header files shouldn't be maintained here but in the toolchain package for Windows we keep at https://github.com/nodemcu/espressif-sdk-archive.

HHHartmann: That would make sense. Please also note that the Linux toolchain is fetched from https://github.com/jmattsson/esp-toolchains.
But I have no idea how that compares/relates.

jmattsson: You're welcome to drag that repo in under the nodemcu project too, though it really only needs the 8266 compiler included I suspect. On the esp32 side the compilers and tools are all handled by the Espressif IDF automatically these days and pulled from their servers.

HHHartmann: So who can adjust the toolchain package.
Should we adapt the original package or should we introduce a patch package which just includes the two files

marcelstoer: I am not sure I understand the full implications of this change, but I'd say we put the two `.h` files into a patch package and give it an expressive name. I feel https://github.com/nodemcu/espressif-sdk-archive is still the right place to host them/it (we firmware maintainers have full access).

caiohamamura: Oh I got what I've done wrong. The issue was that I used the wrong toolchain version. The 8.4.0 is actually for newer than >=3.0 SDK, I should have used the version 4.8.5. I did a throughly test now and it works <https://dl.espressif.com/dl/xtensa-lx106-elf-win32-1.22.0-88-gde0bdc1-4.8.5.tar.gz>. But I need to actually review everything and push the commits.

I also created a [`cmake`](https://github.com/caiohamamura/nodemcu-firmware/tree/cmake) based version as building system that can handle everything as Makefile is currently handling (except for flash targets and spiffs images, although the spiffs binary itself is being built), I made it work and compile in multiple platforms using github actions but I actually never flashed those into a real device, only for my machine to double check. As soon as I have more time I will do that.

Thanks guys for supporting, I was actually inactive for a while because of other projects, now I am currently teaching IoT so I plan to become active again and, why not, try to engage some students into NodeMCU community to give it a new breath. With the ESP32 on the table I don't know if I should focus efforts into ESP32 though. 

Maybe we could do an effort to make a single source of truth and work with ESP32 and ESP8266 `#defines #ifdef` , though I don't know how difficult would it be as I never actually looked at the ESP32 codebase. But it would force us to implement on both platforms whenever we update one of them.

jmattsson: @caiohamamura A shared source tree for both the old 8266 and the current ESP32 series is not really feasible I'm afraid. The ESP32s are running FreeRTOS and the SDK (IDF) for it looks completely different. Many years ago I did do a NodeMCU port for the 8266 based on the old Espressif FreeRTOS SDK for the 8266, but we ended up with far less available RAM compared to the nonOS SDK, so it didn't go anywhere.

I haven't had much time or energy to spend on NodeMCU for some time now, so I'd certainly welcome a hand on the ESP32 side. There are a couple of modules that need to be updated to switch from the deprecated legacy drivers to the current ones. If you're looking for something to poke at and have the hardware to test with, they might be a good starting point.

Also, the documentation for the IDF is quite good, so getting up to speed is hopefully pretty easy.

But if you prefer to work on 8266, that's totally cool too!

PR #3706, opened 2026-05-30
Fix sdk pruning

### LINKED ISSUE #3705: SDK pruning rule uses the wrong `ar`
### Expected behavior

The rule for `$(TOP_DIR)/sdk/.pruned-$(SDK_VER)` should prune using the `ar` supplied by the toolchain, not any `ar` ambiently available in `${PATH}`.

### Actual behavior

[If the version of GNU make is <= 3.81](https://stackoverflow.com/a/77925167), then the rule for `$(TOP_DIR)/sdk/.pruned-$(SDK_VER)` does not use the `ar` supplied by the toolchain, instead using whatever is ambiently available, which may be incompatible or missing.

### Test code
On a system with make <= 3.81 (e.g. MacOS), from a shell where `ar` is not available on `${PATH}`:

```shell
make
```

This is a build-time bug so is independent of NodeMCU firmware version.

(Copied from https://github.com/georgevdd/nodemcu-firmware/issues/4)

georgevdd:
Fixes #3705.

The main `Makefile` adjusts `PATH` to be able to find the toolchain's executables, but commands that require those executables must be executed in subshells otherwise they will not see the updated value of `PATH`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,29 +55,32 @@ jobs:
         name: luac.cross_${{ matrix.lua_ver }}_${{ matrix.numbers }}
         path: luac.cross
 
+# The files in the 'msvc' folder would need to be updated to be compatible with VS 2022.
+# Sample job (that failed): https://github.com/nodemcu/nodemcu-firmware/actions/runs/17306868602/job/49131917376#step:3:153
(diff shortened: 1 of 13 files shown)


georgevdd:
Accidentally made a PR for the release branch

PR #3707, opened 2026-05-30
Fix SDK pruning to use toolchain's `ar`.

georgevdd:
Fixes #3705.

The main `Makefile` adjusts `PATH` to be able to find the toolchain's executables, but (for GNU make <= 3.81, including that shipped with MacOS) commands that require those executables must be executed in subshells otherwise [they will not see](https://stackoverflow.com/a/77925167) the updated value of `PATH`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -333,9 +333,11 @@ $(TOP_DIR)/sdk/.extracted-$(SDK_VER): $(TOP_DIR)/cache/$(SDK_FILE_VER).zip
 $(TOP_DIR)/sdk/.pruned-$(SDK_VER):
 	rm -f $(SDK_DIR)/lib/liblwip.a $(SDK_DIR)/lib/libssl.a $(SDK_DIR)/lib/libmbedtls.a
 	$(summary) PRUNE libmain.a libc.a
-	echo "$(PATH)"
-	$(AR) d $(SDK_DIR)/lib/libmain.a time.o
-	$(AR) d $(SDK_DIR)/lib/libc.a lib_a-time.o
+	# On Make <=3.81 (such as MacOS provides) these commands need to be run
+	# in shells rather than directly by Make, so that $(AR) is found using
+	# the updated `PATH` that is computed above.
+	(exec $(AR) d $(SDK_DIR)/lib/libmain.a time.o)
+	(exec $(AR) d $(SDK_DIR)/lib/libc.a lib_a-time.o)
 	touch $@
 
 $(TOP_DIR)/cache/$(SDK_FILE_VER).zip:


marcelstoer:
> need to run in subshells so as to be able to observe the updated PATH

Hhmm, not sure this assessment is actually correct. Make passes all exported variables as environment variables to every shell it spawns for recipe lines, regardless of whether they're wrapped in `(…)` or not. I suspect the actual cause of the build failure is something else.

@jmattsson how about declaring a new `TOOLCHAIN_AR` and use that one instead of `$(AR)`?

georgevdd:
Here's a `Makefile` that demonstrates the phenomenon:

```make
PATH := ${PATH}:additional_path

additional_path/tool:
	mkdir -p $$(dirname $@)
	echo '#!/bin/sh' >> $@
	echo 'echo tool was invoked with path $${PATH}' >> $@
	chmod ug+x $@

x: additional_path/tool
	$(info PATH is ${PATH})
	(tool)
```

If I remove the `()` from the final line, `make x` fails with `make: tool: No such file or directory`.

The reason as I understand it is that, while `make` does pass the updated `PATH` as an environment variable to commands that it spawns (as you point out) it _does not_ use that updated `PATH` to do its own lookup. By executing `${AR}` in a subshell, the command is looked up on the updated `PATH` and not the one that `make` itself is using.

georgevdd:
[Further digging](https://stackoverflow.com/a/77925167) reveals the likely reason why we seem to be disagreeing: I'm using `make` version 3.81 but you're not on Mac OS and are using a later version.

marcelstoer:
> but you're not on Mac OS 

I am 🤓 However, I never build NodeMCU on macOS.

georgevdd:
I've updated this PR and the issue to which it relates, to be clear about which versions of GNU make are affected.

jmattsson:
I can confirm the behaviour here on my mac with the old GNU Make 3.81. The exact same Makefile runs as expected on my Linux workstation with GNU Make 4.4.1.
```
$ make x
PATH is /opt/homebrew/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Library/Apple/usr/bin:/opt/homebrew/bin:additional_path
tool
make: tool: No such file or directory
make: *** [x] Error 1
```

At first I thought the cause seemed to be the invoked shell not having an up to date command hash for some odd reason. If instead of `(tool)` I change the line to `hash -r; tool` that also works. Digging even deeper, I found that on 4.4.1, `.SHELLFLAGS` is set to `-c`, but is empty on 3.81. Explicitly setting it to `-c` on 3.81 did _not_ fix things.

Looking around, I found this article: https://stackoverflow.com/questions/77019756/make-not-finding-executable-added-to-path-in-makefile
which lays out the full story and says this:

> The problem is a bug or limitation of Gnu Make 3.81, which is the version that Apple ships with its OS (/usr/bin/make). The behavior was changed in version 3.82, and it appears (in my limited testing) that version 3.82 and later execute the OP's script without error. However, 3.81 was the last version to be distributed under GPLv2, and Apple has shown a firm resistance to shipping anything licensed under GPLv3 (such as make 3.82 or bash v4), so I would never expect Apple to ever ship an updated version of Gnu Make. If they were to ship a newer make utility, I would expect them to use a different fork of Make altogether, as they did by replacing bash with zsh as their default terminal shell.

> In Gnu Make 3.81 (Apple's make), changes to PATH affect the shell, but do not affect make itself. If a recipe is a straightforward invocation of a command, make does not launch a shell and have the shell look for it and launch it, make launches the command itself, searching for it using the $PATH it was launched with. In this scenario, your changes to PATH are ignored.

> The solution, if you want to be compatible with the default version of make you will find on Apple computers, is to force make to launch a shell. That can be as simple as chaining lines together using ; \ or using the : no op with &&:

So a sub-shell isn't technically required as a workaround (and is the least efficient). Regardless of which approach we pick from (likely not a comprehensive list):
- `(tool)`
- `hash -r; tool`
- `:&& tool`
- `:; tool`

we should update the comment to say that this odd syntax is due to needing a workaround for GNU Make 3.81, to save people from head scratching down the line (and inadvertently backing out the fix).

marcelstoer:
Thanks Jade for documenting a few options. I don't have a strong preference but would go with `:&& tool` (and the comment of course) if it were my project.

georgevdd:
So, @jmattsson is seeing the same as me (Make v3.81 or earlier doesn't use the `PATH` Make variable to look up commands for execution) and managed to find the same Stack Overflow comment that I did. The Stack Overflow comment notes: "The solution [...] is to force make to launch a shell."

I'm not sure I follow @jmattsson 's remark that "a sub-shell isn't technically required as a workaround (and is the least efficient)." I think the argument is that writing `(tool)` would cause Make to launch a shell in order to interpret the parentheses, which would in turn create a subshell. Have I got that right?

I've updated my test `Makefile` to emit `SHLVL` as well as `PATH` from the `tool` process. It shows 3 for `(tool)` and for `:; tool` but I can get 2 with `(exec tool)` so maybe that's best

jmattsson:
I'd forgotten about the existence of SHLVL! Neat detective work there 👍
I'm surprised at the results, but don't doubt your findings. Did you also try `exec tool`, or did make consider that trivial?

Since I'm not really active on this branch any more, I'll defer to Marcel's preferences 😊

marcelstoer (approved):

georgevdd:
`exec tool` gives the same results as `(tool)`: lookup respects `PATH` and `SHLVL` is 2.

Of the two I prefer `(tool)` because it is clearer that the resulting process won't replace the `make` process itself.

georgevdd:
I've updated this change to use our findings and to make the comment clearer. I've also taken the liberty of removing an `echo "$(PATH)"` that I suspect is a legacy of some earlier attempt to investigate the same phenomenon.

